PR #17, opened 2020-10-29
add pyarrow integration for python crate

houqp:


python/src/lib.rs
```diff
@@ -26,13 +26,28 @@ impl DeltaTable {
...
-    pub fn files(&self) -> PyResult<Vec<String>> {
-        Ok(self._table.get_files().to_vec())
+    #[args(full_path = "false")]
+    pub fn files(&self, full_path: bool) -> PyResult<Vec<String>> {
+        let list = self._table.get_files().to_vec();
+        if full_path {
+            Ok(list
+                .iter()
+                .map(|fpath| {
+                    Path::new(&self._table.table_path)
```
rtyler, line 36: I did a lot of this last weekend, getting `String` from `Path` is super annoying

rtyler (approved):

python/src/lib.rs
houqp: yes, utf8 fun.

PR #27, opened 2020-10-31
add time travel to python binding

houqp:
also throw in code cleanup to remove use on unwrap

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -21,17 +21,17 @@ bytes = "0.5.3"
 log = "0"
 regex = "1"
 chrono = "0"
-crossbeam = "0"
 rusoto_core = "0.43"
 rusoto_s3 = "0.43"
 datafusion = { version = "2", optional = true }
+crossbeam = { version = "0", optional = true }
 # NOTE: disable rust-dataframe integration since it currently doesn't have a
 # version published in crates.io
 # rust-dataframe = {version = "0.*", optional = true }
 
 [features]
 rust-dataframe-ext = []
-datafusion-ext = ["datafusion"]
+datafusion-ext = ["datafusion", "crossbeam"]
 
 [dev-dependencies]
 utime = "0.3"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,6 @@ jobs:
     - name: Check
       run: cargo clippy
     - name: Build
-      run: cargo build --verbose
+      run: cargo build
     - name: Run tests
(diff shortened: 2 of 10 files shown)


rust/Cargo.toml
```diff
@@ -21,17 +21,19 @@ bytes = "0.5.3"
 log = "0"
 regex = "1"
 chrono = "0"
-crossbeam = "0"
 rusoto_core = "0.43"
 rusoto_s3 = "0.43"
+pyo3 = { version = "0", features = ["extension-module"], optional = true }
```
rtyler, line 26: Why does this dependency need to come into the root crate? I'm sure there's a reason but would you please comment in the file why this exists here? (cause it feels leaky)

rust/src/delta.rs
```diff
@@ -79,6 +79,21 @@ pub enum DeltaTableError {
...
 
+#[cfg(feature = "python")]
+pub mod python {
+    extern crate pyo3;
+    use self::pyo3::create_exception;
+    use self::pyo3::exceptions::PyException;
+
+    create_exception!(deltalake, PyDeltaTableError, PyException);
+
+    impl From<super::DeltaTableError> for pyo3::PyErr {
```
rtyler, line 90: I don't understand enough of what's going on here to tell why this trait implementation must exist in this crate versus the python crate :confused: _

houqp: i tried to do this within python crate, but couldn't get it to compile because `DeltaTableError` and `pyo3::PyErr` are both defined in foreign crates. To define a From trait, either the destination or the end needs to defined in the current crate, which is not the case for python create :(

rtyler: I figured it might be something like this.

I think we might need to try to find a different approach here unfortunately. Would the `deltalake` crate be able to define `DeltaLakeError` as a trait and use an `impl DeltaTableError` where it's sending errors around, such that the Python crate would implement its own custom error on top?

My main concern here is that dependency features in Cargo workspaces are shared such that I don't believe the Ruby crate would be able to be built _without_ the `python` feature (I haven't yet checked this though).

I cannot remember exactly, but I think [this issue in Cargo](https://github.com/rust-lang/cargo/issues/4463) was the one I originally looked at.

houqp: I moved the error chaining into python crate using `map_err`. The code is a little bit more verbose, but i think it's worth the tradeoff to remove a language specific feature from the main crate.

rtyler (approved):
Definitely cleaner, good solution :+1:

PR #28, opened 2020-11-02
Update LICENSE

zsxwing:
Copy `LICENSE.txt` from https://github.com/delta-io/delta and remove the `Apache spark` part since this project doesn't contain codes from `Apache Spark`.


LICENSE.txt
```diff
@@ -0,0 +1,179 @@
+Copyright (2020) The Delta Lake Project Authors.  All rights reserved.
```
rtyler, line 1: (I'm obviously not a lawyer but have spent tons of time working on licensing around various projects.)

From my perspective, this is not a valid copyright declaration since there is no entity "The Delta Lake Project Authors" which can hold copyrights. Furthermore contributors to delta-io/delta have no copyrights on this codebase and vice versa.

Since the Apache license doesn;t require strict attribution, the copyright header is really only useful for informing tools and lawyers who might own the rights here. Since we're not doing copyright assignment, I think the following would probably cover our bases for the future (lifted from the Jenkins core license file):


```suggestion
Copyright (2020) QP Hou and a number of other contributors.  All rights reserved.
```

(I think @houqp's name should remain since he's the original creator of this software)

zsxwing: @rtyler Thanks for the feedback. I have little knowledge regarding licensing.

@tdas since you added this to the Delta Lake core project, could you clarify what are the requirements regarding licensing from The Linux Foundation?

tdas: to be honest I do not have authoritative knowledge either, I have followed what our lawyers suggested us to do. this requires more conversation without our lawyers. let me know look into it.

rtyler (approved):
This has been hanging around long enough, I've committed my suggested rephrase, but I'm merging the whole of it.

PR #32, opened 2020-11-09
add delta-inspect binary

houqp:


rust/src/bin/delta-inspect.rs
```diff
@@ -0,0 +1,68 @@
+extern crate anyhow;
+extern crate deltalake;
+
+use clap::{App, AppSettings, Arg};
+
+fn main() -> anyhow::Result<()> {
+    let matches = App::new("Delta table inspector")
+        .version("1.0")
```
rtyler, line 8:
```suggestion
        .version(env!("CARGO_PKG_VERSION"))
```

This should pin the version of the tool to the cargo version

houqp: wouldn't it be better to decouple the cli version from the library version? for example, if we are to introduce a breaking cli UX change, we wouldn't want to bump the major crate version just for that right?

rtyler: Up to you to be honest. Since this is a tool inside of this cargo workspace, it's pinned to the Rust library anyways at least as far as functionality goes.

Coupling them together for now gives us one less version number to futz with at release time.

rtyler (approved):
Either way @houqp goes on versions, this is fine

rust/src/bin/delta-inspect.rs
houqp: ok, switched to use crate version, we can decouple the version if we run into issue with that in the future.

PR #34, opened 2020-11-28
Add Azure storage backend; feature gate both azure and s3 backends

### LINKED ISSUE #30: Add Azure support
It would be great to have support for Azure as well as S3 here. The Rust Azure SDK recently moved from https://github.com/MindFlavor/AzureSDKForRust to https://github.com/Azure/azure-sdk-for-rust and hasn't yet released anything to crates.io using the new crate names, but I imagine it won't be too long!

### LINKED ISSUE #31: Feature gate some modules/dependencies
Specifically the S3 (and Azure, if #30 is implemented) functionality could be feature gated fairly easily.

This would also be required for #22 because currently Rusoto depends on `hyper`, which won't compile to WASM (https://github.com/rusoto/rusoto/issues/1708 would be the likely eventual solution there).

sd2k:
The Azure storage backend currently requires the `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY` environment variables to be set, since I'm not aware of an equivalent of Rusoto's `CredentialProvider` for Azure. This should probably be changed though since it's non-obvious - suggestions welcome!

Closes #30, closes #31.


rust/Cargo.toml
```diff
@@ -22,18 +22,28 @@ bytes = "0.5.3"
 log = "0"
 regex = "1"
 chrono = "0"
-rusoto_core = "0.43"
-rusoto_s3 = "0.43"
+
+# Azure
+azure_core = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true}
```
houqp, line 27: given that the rust sdk is in active development, it's more likely for it to break our azure build if we don't pin to a specific commit.

houqp (commented):
great change overall! could you also update github action to make sure s3 and azure features are linted and tested in CI?

houqp:
for credential injection, i think it's fine to start with `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY` as long as it's documented. once upstream sdk standardized the credential management flow, we can switch to that.

rust/Cargo.toml
sd2k: Makes sense - done.

sd2k:
I've updated this to:
1. pin the version of the Azure SDK for now, since it's unreleased
2. accept `AZURE_STORAGE_SAS` as well as `AZURE_STORAGE_KEY`
3. add a very basic test to fetch the 'simple' Delta table from a storage account I created
4. switch to using a tokio `Runtime` instead of trying to use `futures::executor` - it looks like the Azure SDK requires tokio to be used.

I think it's in a better place now!

sd2k:
I was wondering about some of the APIs: since functions aren't async we need to create a new runtime every time a `StorageBackend` performs an operation. Since a large number of operations depend on this it means we're blocking a lot of times when several things could probably be done concurrently (e.g. downloading all the log files for a table). Would you be averse to async-ifying some of these functions? Happy to have a crack at it if so.

houqp (approved):

houqp:
@sd2k with regards to async-ifying everything, it's the top item on the todo list: https://github.com/delta-io/delta.rs/blob/main/TODO#L1. So definitely feel free to take a stab at it ;)

PR #35, opened 2020-12-03
Asyncify APIs

sd2k:
This PR makes the `StorageBackend` trait methods async (using `async_trait`), and changes the implementations for file/Azure/S3 to be async. The `list_obj` methods now returns a `Stream<Item = Result<ObjectMeta, StorageError>>` instead of an iterator.

This change ends up propagating all the way back to the main entrypoints, since those functions do quite a lot of IO to get data from the storage backend. Not 100% how that could affect the Python APIs yet - perhaps I should look into it before this is merged - but I thought it could be worth pushing this up as-is.

I don't have any way of testing this against S3 directly unfortunately; it would be good to get tests for a large number of objects to test the continuation token code.

### DIFF
--- a/rust/src/storage/azure.rs
+++ b/rust/src/storage/azure.rs
@@ -119,7 +119,7 @@ impl StorageBackend for ADLSGen2Backend {
     {
         debug!("Listing objects under {}", path);
         let obj = parse_uri(path)?.into_adlsgen2_object()?;
-        let account_name = self.client.account().clone();
+        let account_name = self.client.account();
         let stream = self
             .client
             .list_blobs()
--- a/rust/src/storage/azure.rs
+++ b/rust/src/storage/azure.rs
@@ -91,8 +91,7 @@ impl StorageBackend for ADLSGen2Backend {
             .with_container_name(obj.file_system)
             .with_blob_name(obj.path)
             .finalize()
-            .await
-            .unwrap();
+            .await?;
         let modified = properties
             .blob
             .last_modified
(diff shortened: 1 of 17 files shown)


sd2k:
:man_facepalming: I just realised I've been running tests from inside the `rust` directory and there's loads of clear failures for the Python/Ruby tests! I'll take a look at that tomorrow.

sd2k:
Waiting for https://github.com/Azure/azure-sdk-for-rust/pull/112 before the Azure feature will compile.

rtyler (approved):

rust/src/storage/azure.rs
```diff
@@ -89,75 +79,73 @@ impl Default for ADLSGen2Backend {
...
-                .expect("Last-Modified should never be None for committed blobs"),
-        })
+        let properties = self
+            .client
+            .get_blob_properties()
+            .with_container_name(obj.file_system)
+            .with_blob_name(obj.path)
+            .finalize()
+            .await
+            .unwrap();
```
houqp, line 95: AzureError from trait has already been implemented for StorageError, I think we should be able to replace the unwrap here with a`?`?

houqp (commented):
tested S3 with some of our tables, works as expected :)

rust/src/storage/azure.rs
sd2k: Good spot! Fixed.

sd2k:
Think this is ready for another review, if you're happy with the usage of tokio in the Python/Ruby libs!

houqp:
Great work, thanks @sd2k !

PR #48, opened 2020-12-28
avoid duplicate mac os pypi release

houqp:
10.15 an 11.0 targets the same platform macosx_10_7, so we only need to
run release with 10.15.


rtyler (approved):
:shrug:

PR #49, opened 2020-12-28
statically link openssl for python wheel

houqp:


rtyler (approved):

rtyler:
If you're actively working on this, I would recommending making a note in the readme with your next pull request. A statically built openssl while convenient will have security implications for users.

houqp:
Security concern was the main reason I went with dynamic linking initially. But after further research, I learned that for manylinux python wheels, we are supposed to use static linking for libraries like openssl instead: https://www.python.org/dev/peps/pep-0513/#id48. I also double checked couple other projects, they all use static linking for manylinux wheels, for example: https://cryptography.io/en/latest/installation.html#static-wheels. Basically if a manylinux wheels dynamically links against openssl, it won't work on all linux distributions due to unstable abi.

rtyler (approved):

PR #51, opened 2020-12-30
support table stats

houqp:


rust/src/action.rs
```diff
@@ -48,6 +48,67 @@ fn gen_action_type_error(action: &str, field: &str, expected_type: &str) -> Acti
     ))
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum ColumnValueStat {
+    Column(HashMap<String, ColumnValueStat>),
+    Value(serde_json::Value),
```
houqp, line 55: leaving the value as generic json value for now since the implementation is not released in open-source code base and the official spec is ambiguous on what types are supported. We can refactor this once reference implementation is released.

rtyler (approved):
I'm on the fence about the serde errors bubbling up through this. Either way, merge away

Issue #42, opened 2020-12-12
S3 storage backend should take the region as a parameter


rtyler:
Right now us-east-2 is hard-coded. Not sure what the right API will be here, but the region should be user-defined in some fashion

houqp:
Based on the look up order in https://docs.rs/rusoto_credential/0.45.0/src/rusoto_credential/lib.rs.html#368-373, we can try to first read the region from `AWS_REGION` env var, if not found use https://docs.rs/rusoto_credential/0.45.0/rusoto_credential/struct.ProfileProvider.html#method.region_from_profile to load it from profile. And lastly throw an error if region cannot be resolved.

houqp:
I think this could be a useful feature to add to upstream trait: https://docs.rs/rusoto_credential/0.45.0/rusoto_credential/trait.ProvideAwsCredentials.html

rtyler:
@houqp I was referring to [this line](https://github.com/delta-io/delta.rs/blob/main/rust/src/storage/s3.rs#L32) which I don't fully understand how the credentials interacts with that setting on the `S3Client`

houqp:
it uses https://docs.rs/rusoto_credential/0.45.0/rusoto_credential/struct.DefaultCredentialsProvider.html, which is basically chained provider. there is also a method that lets you override what provider to use: https://rusoto.github.io/rusoto/rusoto_s3/struct.S3Client.html

kevinglasson:
Should it just mimic the [S3Client API](https://rusoto.github.io/rusoto/rusoto_s3/struct.S3Client.html) API and have:

pub fn new(region: Region) -> S3Client

pub fn new_with<P, D>(
    request_dispatcher: D,
    credentials_provider: P,
    region: Region
) -> S3Client

?

houqp:
@kevinglasson the code you pasted still requires users to manually specify a region and it is what we are using at the moment: https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/s3.rs#L32.

The issue here is implementing a way to automatically resolve the region based on credential chain so it works just like other tools using AWS sdks.

kevinglasson:
@houqp ahh I think I misunderstood, sorry.

I personally think it makes sense for the user to have to specify the region when creating a storage backend, seeing as the bucket they're supposedly targeting should already exist.

[...]

PR #52, opened 2021-01-05
use default region for S3StorageBackend (#42)

kevinglasson:
Addresses #42

Using Region::default() instead of a specific hard coded region.

This will allow the user to set the region using one of AWS' config files **HOWEVER** the region (as far as I understand) is irrelevant when working with s3:// URI's and buckets in general because their names are globally unique.


houqp:
Thanks @kevinglasson ! time to release a new version of the python binding with this fix :D

PR #61, opened 2021-01-10
make snapshot parsing more forward compatible

### LINKED ISSUE #54: PyDeltaTableError: Invalid action record found in log: Invalid action field: Unexpected field name for add action: partitionValues_parsed
I'm getting the below error when trying to read a delta table from ADLS Gen 2:
`PyDeltaTableError: Invalid action record found in log: Invalid action field: Unexpected field name for add action: partitionValues_parsed`

When I try and read the delta table using spark sql, I have no issues. Any ideas what could be causing this?

houqp:
Do not return unrecoverable error on new record fields from future
delta table versions. Fix #54.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -660,6 +660,7 @@ dependencies = [
  "clap 3.0.0-beta.2",
  "crossbeam",
  "datafusion",
+ "env_logger",
  "futures",
  "log",
  "parquet",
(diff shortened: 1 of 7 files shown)


rtyler (approved):
Personally, I would rather these be user-visible warnings rather than just debug statements. since they may indicate inaccurate results when using delta-rs. Not important though, merge at will.

PR #64, opened 2021-01-11
release python binding with maturin 0.9.0 release

houqp:
upstream release have caught up on my patches, this should speed up release pipeline


rtyler (approved):
:tada:

Issue #72, opened 2021-01-17
File not found exception for a delta table stored on cloud storage


danielhaviv:
Passing a list of files to pyarrow.dataset.dataset is treated as a list of local files, quoting the [docs](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.dataset.html#pyarrow.dataset.dataset):
```
List of file paths:

    Create a FileSystemDataset from explicitly given files. The files must be located on the same filesystem given by the filesystem parameter. Note that in contrary of construction from a single file, passing URIs as paths is not allowed.

```

Reading tables off of cloud storage fails with a file not found exception.
To fix this, `dataset` should be called with the filesystem parameter and the file paths should be only the keys and not the full URIs.

I'll submit a PR.

Issue #69, opened 2021-01-16
Call to DeltaTable from Python


fvaleye:
Hello guys,

I am using `delta-rs` from Python and it seems that calling `DeltaTable()` blocks after calling it once.
I tested with DeltaTable on AWS S3.

**Context information:**
```
deltalake="==0.2.0" and deltalake="==0.1.4" installed via pip 20.3.3
Python 3.7.8
```

**Scenario:**
```
table_list = ["s3://bucket-test/delta_table1",  "s3://bucket-test/delta_table2"]
for table in table_list:
    delta_table = DeltaTable(table)
    print(delta_table.files())
```

The first call to `DeltaTable()` with `files()` worked and the files are printed. But on the second call, it blocks my application.
I tried to use it without the loop, with sequential calls, but it is the same.

Is it normal?

Thank you for your help :)

rtyler:
I was able to quickly reproduce this, it has nothing to do with multiple tables, doing an iteration across a list with the same bucket hangs on the second iteration as well.

```
>>> table_list = ['s3://deltars/simple', 's3://deltars/simple']
>>> for table in table_list:
...     print('hello')
...     t = deltalake.DeltaTable(table)
...     print(t.files())
```


@houqp  I have a feeling that there is something that's not being properly deallocated here. I would write a proper integration test, but I cannot get pyo3 to build on my local (openSUSE) machine :unamused:

rtyler:
My hunch is that the tokio runtime that's being created per DeltaTable object is not being dropped properly in the iteration.

PR #73, opened 2021-01-17
Fixed issue #72

danielhaviv:


python/deltalake/__init__.py
```diff
@@ -23,7 +24,14 @@ def load_version(self, version: int) -> None:
...
     def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
-        return dataset(self._table.file_paths(), format="parquet")
+        paths = [urlparse(curr_file) for curr_file in self._table.file_paths()]
+
+        # Decide based on the first file, if the file is on cloud storage or local
+        if paths[0].netloc:
+            keys = [curr_file.path for curr_file in paths]
+            return dataset(keys, filesystem=paths[0].scheme + "://" + paths[0].netloc)
+        else:
+            return dataset(self._table.file_paths(), format="parquet")
```
houqp, line 34: for local files, self._table.file_paths() is invoked twice, could you save it as a local var at the beginning of the function and reuse that local var here?

python/deltalake/__init__.py
```diff
@@ -23,7 +24,14 @@ def load_version(self, version: int) -> None:
         self._table.load_version(version)
 
     def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
-        return dataset(self._table.file_paths(), format="parquet")
+        paths = [urlparse(curr_file) for curr_file in self._table.file_paths()]
+
+        # Decide based on the first file, if the file is on cloud storage or local
+        if paths[0].netloc:
+            keys = [curr_file.path for curr_file in paths]
+            return dataset(keys, filesystem=paths[0].scheme + "://" + paths[0].netloc)
```
houqp, line 32: Good practice to avoid string concatenation using `+`

```suggestion
            return dataset(keys, filesystem=f"{paths[0].scheme}://{paths[0].netloc}")
```

houqp (approved):
Good catch :+1:

python/deltalake/__init__.py
danielhaviv: Done

rtyler:
The build error looks to be unrelated, and might be a similar problem to that which is described in #69 but manifesting through our integration tests.

```
running 2 tests
test s3::test_s3_simple_golden ... ok
test s3::test_s3_simple ... FAILED

failures:

---- s3::test_s3_simple stdout ----
thread 's3::test_s3_simple' panicked at 'dispatch dropped without returning error', /home/tyler/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.13.9/src/client/conn.rs:234:35
```


I'll let @houqp merge this if he's satisfied with the changes.

houqp:
#69 is a bug in python binding, while this test is rust only, so it's something else. I will merge this PR for now. Integration test should be fixed in a separate PR.

houqp:
Thanks @danielhaviv for the contribution!

danielhaviv:
What are the release cycles (or when should I expect to see the fix in PyPi)?

houqp:
we don't have a set release cadence, I was hoping to cut a new release after #75 gets merged.

PR #75, opened 2021-01-17
Avoid tokio runtime blocking table load from python binding

### LINKED ISSUE #69: Call to DeltaTable from Python

houqp:
fix #69

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -63,9 +63,9 @@ jobs:
     # - name: Run Rust tests
     #   run: cargo test
 
-    - name: Install maturin
+    - name: Install matruin
       run: |
-        cargo install --git https://github.com/PyO3/maturin.git --rev 98636cea89c328b3eba4ebb548124f75c8018200 maturin
+        pip install maturin==0.9.0
 
     - name: Run Python tests
       run: |
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-python"
-version = "0.2.0"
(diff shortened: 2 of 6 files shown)


houqp:
FYI @sd2k , interestingly, upgrading tokio to 1.0 also fixes the blocking issue without requiring creating a new runtime for every async call.

rtyler (commented):
Rather than one tokio runtime per DeltaTable, this is one runtime per invocation of anything? 

What would this mean if I had tried to use a `DeltaTable` in a multithreaded Python application?

houqp:
That's correct, one runtime per invocation. this approach is basically copied from https://github.com/rusoto/rusoto/blob/1e1fd8357cc59f8619c3277c5e2ff1f941ec1ee3/rusoto/signature/src/stream.rs#L139.

I have added a test for concurrent read from multiple threads in python.

rtyler (approved):

PR #76, opened 2021-01-18
bump tokio to 1.0

houqp:
datafusion tests are expected to fail until we pick up https://github.com/apache/arrow/pull/9230.

### DIFF
--- a/rust/src/storage/azure.rs
+++ b/rust/src/storage/azure.rs
@@ -3,14 +3,19 @@
 //! This module is gated behind the "azure" feature. Its usage also requires
 //! the `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY` environment variables
 //! to be set to the name and key of the Azure Storage Account, respectively.
+use std::error::Error;
+use std::sync::Arc;
 use std::{env, fmt, pin::Pin};
 
+use azure_core::errors::AzureError;
 use azure_core::prelude::*;
-use azure_storage::{client, key_client::KeyClient, Blob, ClientEndpoint, Container};
+use azure_storage::clients::{
+    AsBlobClient, AsContainerClient, AsStorageClient, ContainerClient, StorageAccountClient,
+};
 use futures::stream::{Stream, TryStreamExt};
 use log::debug;
 
-use super::{parse_uri, ObjectMeta, StorageBackend, StorageError};
+use super::{parse_uri, ObjectMeta, StorageBackend, StorageError, UriError};
 
 /// An object on an Azure Data Lake Storage Gen2 account.
 #[derive(Debug, PartialEq)]
@@ -40,7 +45,8 @@ impl<'a> fmt::Display for ADLSGen2Object<'a> {
 /// This uses the `dfs.core.windows.net` endpoint.
 #[derive(Debug)]
 pub struct ADLSGen2Backend {
-    client: KeyClient,
+    account: String,
+    container_client: Arc<ContainerClient>,
 }
 
 impl ADLSGen2Backend {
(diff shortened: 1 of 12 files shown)


rust/src/storage/azure.rs
```diff
@@ -120,18 +156,18 @@ impl StorageBackend for ADLSGen2Backend {
     {
         debug!("Listing objects under {}", path);
         let obj = parse_uri(path)?.into_adlsgen2_object()?;
-        let account_name = self.client.account();
+        self.validate_container(&obj)?;
+
         let stream = self
-            .client
+            .container_client
             .list_blobs()
```
houqp, line 163: @sd2k the new sdk's list_blob response builder now takes a reference to container client (see  https://github.com/Azure/azure-sdk-for-rust/blob/master/sdk/storage/src/blob/container/requests/list_blobs_builder.rs#L11), which is causing the following error:

```
rror[E0515]: cannot return value referencing temporary value
   --> rust/src/storage/azure.rs:155:9
    |
130 |           let stream = self
    |  ______________________-
131 | |             .storage_client
132 | |             .as_container_client(obj.file_system)
    | |_________________________________________________- temporary value created here
...
155 |           Ok(Box::pin(stream))
    |           ^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

This is because the current Azure Storage backend is not tied to a specific container, which forces us to create new method local container client on every method call.

I don't have a good solution to address this off the top of my head, so I changed azure backend's `new` interface to take container name as argument so we can store the container client as part of the struct. This fixes the build for now, but we will get into the same problem with blob client if we are to convert `get_obj` to return a stream in the future. Let me know if you any suggestion on how to better handle this upgrade.

sd2k: @houqp Ah, that isn't fun at all. I was hoping we could solve this by returning a separate struct (e.g. `ListBlobsStream`) which wraps the existing stream and the container, and implements `Stream<Item = Result<ObjectMeta, StorageError>>`. I think this should be possible somehow, but I'm running into a similar issue as you when playing around with implementing it... do you know any `Stream` experts? :D

rtyler (approved):
This change looks fine to me.

However it's unclear how long it will take that Arrow pull request to be merged. I don't think we should merge this change until it has. If it's going to take a way, I would rather pin our development against a new arrow development ref.

Part of the problem with pinning to development refs is that we cannot cut releases to crates.io.  And if that arrow change is not coming in until after 3.0.0 and "[has a few months to stabilize](https://github.com/apache/arrow/pull/9230#pullrequestreview-570046965)" then we will basically be unable to cut a crates.io release for a number of months here.

rust/src/storage/azure.rs
houqp: ha, good call, i think returning a struct that implements `Stream` is the way to go :)

houqp:
Yeah, I am thinking if tokio 1.0 upgrade is not included in arrow 3.0 release, we can cut a new crate release with 3.0 first before merging this PR.  Then point datafusion to development ref.

rtyler:
Based on our meeting earlier this week, marking this as "on-hold" since we're effectively blocked on Arrow adopting Tokio 1.0 until we can feasibly make this change

houqp:
based on my failed arrow 3.0 bump attempt at https://github.com/delta-io/delta-rs/pull/80, i am now leaning towards tracking arrow head again until they hit 4.0.

rtyler:
The datafusion stuff is still broken it looks like, don't have time at the moment to investigate whether upstream is broken still or not. 

Moving the football slightly further though

houqp:
That's great, i will take it from there :D

houqp:
Once https://github.com/apache/arrow/pull/9369 is merged, we should be ready to go.

Issue #40, opened 2020-12-12
DeltaTable likely not handling log_path in a platform-safe manner


rtyler:
Lots of DeltaTable code is doing a primitive "join" (e.g. `format!("{}/{}", log_path, other)`) when creating paths. In the cases where this path is going to be a file system path, this isn't a safe way to create paths.

@houqp I think we're going to have to find some way to handle these path creations better than just munging strings.

rtyler:
OOOOF, I think this is very much related to #39 `find_latest_check_point_for_version` is doing path regular expressions which are in no possible way going to work as expected on Windows :facepalm:

houqp:
yep, apparently the code was not written with windows in mind :P should replace with `std::path::Path`.

houqp:
having a second thought here. when using the azure or s3 backend in windows, do we get unix path or windows path returned from APIs? If it's consistent with native platform path, then we are good, otherwise, we might need to move all the path handling logic into file storage backend instead.

sd2k:
I've just tried running the E2E Azure test on Windows and it passes. The docs also state that the 'path' in the URI should be forward slash delimited: https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri.

houqp:
Thanks @sd2k for the test! Then it looks like what we need to do is to extend the storage backend trait (https://github.com/apache/arrow/tree/master/rust/parquet) to add methods for path component join and split. Then only use use `std:path::Path` in file storage backend to accommodate for platform specific fs path.

houqp:
as a note for implementer, https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L497 also needs to be updated to stop using platform specific `Path`.

houqp:
@fvaleye just saw my note in this PR, looks like we missed `get_file_paths` in your refactoring :)

fvaleye:
It seems that I forgot to include this in my PR. Thank you for noticing it, I will submit another PR for refactoring this :)

roeap:
A core feature of `object_store` is to provide a common (and extensively tested) abstraction around paths in an object store across all backends and supported OSs. So I believe this should be properly handled now :).

PR #78, opened 2021-01-19
Handle trailing slash in S3 URI

### LINKED ISSUE #62: Trailing slashes cause S3 binding to throw NotATable
The difference between `deltalake::open_table("s3://mybucket/path/")` and `deltalake::open_table("s3://mybucket/path")` is that the former throws an error, the latter succeeds.

IMHO Delta-rs shouldn't be so silly about redundant slashes.

fvaleye:
Proposal for dealing with trailing whitespace in S3 URI.

Fix #62

### DIFF
--- a/rust/src/storage/mod.rs
+++ b/rust/src/storage/mod.rs
@@ -7,10 +7,8 @@ use futures::Stream;
 #[cfg(feature = "azure")]
 use azure_core::errors::AzureError;
 
-#[cfg(feature = "s3")] use {
-    rusoto_core::RusotoError,
-    std::borrow::Cow
-};
+#[cfg(feature = "s3")]
+use rusoto_core::RusotoError;
 
 #[cfg(feature = "azure")]
 pub mod azure;
@@ -111,24 +109,12 @@ pub fn parse_uri<'a>(path: &'a str) -> Result<Uri<'a>, UriError> {
                             return Err(UriError::MissingObjectBucket);
                         }
                     };
-                    let raw_key = match path_parts.next() {
+                    let key = match path_parts.next() {
                         Some(x) => x,
                         None => {
                             return Err(UriError::MissingObjectKey);
                         }
                     };
-
-                    let key = match raw_key.contains('/') {
-                        false => Cow::Borrowed(raw_key),
-                        true => {
-                            let key_cleaned = raw_key.split('/')
-                                                     .filter(|p| !p.is_empty())
-                                                     .collect::<Vec<& str>>()
-                                                     .join("/");
-                            Cow::Owned(key_cleaned)
-                        }
-                    };
-
 
                     Ok(Uri::S3Object(s3::S3Object { bucket, key }))
                 } else {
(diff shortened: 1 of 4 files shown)


rust/src/storage/mod.rs
```diff
@@ -109,13 +111,25 @@ pub fn parse_uri<'a>(path: &'a str) -> Result<Uri<'a>, UriError> {
...
+                    let key = match raw_key.contains('/') {
+                        false => Cow::Borrowed(raw_key),
+                        true => {
+                            let key_cleaned = raw_key.split('/')
+                                                     .filter(|p| !p.is_empty())
+                                                     .collect::<Vec<& str>>()
+                                                     .join("/");
+                            Cow::Owned(key_cleaned)
+                        }
+                    };
```
houqp, line 130: We can trust rust compiler to do the right optimization and simplify this block to:

```suggestion
                    let key =  raw_key.split('/')
                                      .filter(|p| !p.is_empty())
                                      .collect::<Vec<& str>>()
                                      .join("/");
```

rust/src/storage/s3.rs
```diff
@@ -14,7 +15,7 @@ use super::{parse_uri, ObjectMeta, StorageBackend, StorageError};
 #[derive(Debug, PartialEq)]
 pub struct S3Object<'a> {
     pub bucket: &'a str,
-    pub key: &'a str,
+    pub key: Cow<'a, str>,
```
houqp, line 18: Instead of creating a new String and hold a reference to it through Cow, it is more idiomatic to just use the String type here.

Alternatively, we could decide to not handle redundant cases like `s3://foo///bar`, which is the case for AWS cli. The we can keep using `&'a str` here by taking a reference to the original `&str` through https://doc.rust-lang.org/std/primitive.str.html#method.trim_left_matches.

rust/src/storage/mod.rs
fvaleye: Right, it is simpler and more readable!

rust/src/storage/s3.rs
fvaleye: I could keep the redundant test case and refactoring the code with the String type. Or do you prefer to keep the `&'a str`, delete the redundant test case and find another solution for removing the `"//"`?

houqp: I am leaning more towards not handling redundant slashes in delta.rs and leave it to the AWS S3 SDK.

fvaleye: Maybe we could just use https://doc.rust-lang.org/std/primitive.str.html#method.trim_end_matches for `log_path` variable when creating the DeltaTable?

rtyler: The original issue that I filed @houqp  was around _trailing_ slashes in #62. I don't know how the Delta bindings for Scala treat this, but it's silly to me that `s3://delta/simple` is a table but `s3://delta/simple/` is not.

houqp: @fvaleye my bad for posting the wrong link in my previous comment, trim_end_matches is the the api to use here.

fvaleye: I added `trim_end_matches` when creating the DeltaTable to keep using `&'a str` in the S3Bucket and leave the S3 URI to AWS S3 SDK. The error was raised in S3Storage because the checkpoint couldn't be read because of "s3://delta/simple//_delta_log" S3 URI.

houqp: The path normalization needs to happen at the beginning of https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L526. Ideally it needs to be done in a platform agnostic way based on storage backends at use, but we can leave that refactoring to https://github.com/delta-io/delta-rs/issues/40.

houqp (approved):

houqp:
Thanks @fvaleye !

rust/src/storage/s3.rs
fvaleye: I will look into the #40 for level up in Rust :)

houqp: That would be great, while you work on that, i recommend also changing the code to do the path normalization at the beginning of `DeltaTable::new` function so `table_path: table_path.to_string()` also gets the consistent value without trailing slash.

PR #79, opened 2021-01-20
Refactoring - trailing slash in table_path

fvaleye:
Refactoring of trailing slash in `table_path`


rtyler:
Test failure looks entirely unrelated, that stupid dispatch error again. Otherwise looks reasonable to me, but I'll let @houqp be the final decider here :smile_cat:

houqp (approved):

houqp:
yeah, we need to fix that flaky integration test :P i haven't been able to reproduce it locally yet. maybe it's already fixed in https://github.com/delta-io/delta-rs/pull/76 :P

PR #80, opened 2021-01-26
bump arrow to 3.0

houqp:
Now that I think about it, we will not be able to publish to crates.io anyway even with the stable arrow version bump due to dependency on azure sdk, which doesn't have a creates.io release yet.


houqp:
Never mind, datafusion from arrow 3.0 release also introduced a breaking change which made it not possible to build a TableProvider using `ParquetExec` due to introduction of `ParquetPartition`. I think we should just skip arrow 3.0 and keep going with the git refs for now until 4.0 is out. We can add a note in readme that consumption of deltalake crate should go through git ref at the moment.

Issue #39, opened 2020-12-11
Test fails on Windows


thovoll:
Run cargo test

Output:

running 2 tests
test read_delta_2_0_table_without_version ... ok
test read_delta_2_0_table_with_version ... FAILED

---- read_delta_2_0_table_with_version stdout ----
thread 'read_delta_2_0_table_with_version' panicked at 'assertion failed: `(left == right)`
  left: `["part-00000-7c2deba3-1994-4fb8-bc07-d46c948aa415-c000.snappy.parquet", "part-00001-c373a5bd-85f0-4758-815e-7eb62007a15c-c000.snappy.parquet", "part-00000-cb6b150b-30b8-4662-ad28-ff32ddab96d2-c000.snappy.parquet"]`,
 right: `["part-00000-cb6b150b-30b8-4662-ad28-ff32ddab96d2-c000.snappy.parquet", "part-00000-7c2deba3-1994-4fb8-bc07-d46c948aa415-c000.snappy.parquet", "part-00001-c373a5bd-85f0-4758-815e-7eb62007a15c-c000.snappy.parquet"]`', rust\tests\read_delta-0-2-0_test.rs:67:5

The arrays contain the same elements, but in a different order.

rtyler:
I had spun up a dev windows instance a couple weeks ago, I'll take a crack at this sometime soon

rtyler:
Alright, I've got a VM where I have reproduced this failure. Will have a fix in the next available free timeslot :laughing:

rtyler:
Alright, so #40 is the same issue as this. Basically the `find_latest_check_point_for_version` function is not identifying _any_ checkpoint file for version three of this table.

On Linux, it identifies this path as the checkpoint

`Ok(ObjectMeta { path: "./tests/data/delta-0.2.0/_delta_log/00000000000000000003.checkpoint.parquet", modified: 2020-10-30T18:57:13.473315930Z })`

On Windows this entry looks something like this:

`Ok(ObjectMeta { path: "./tests/data/delta-0.2.0/_delta_log\\00000000000000000003.checkpoint.parquet", modified: 2020-12-12T02:02:36.008358300Z })`


I think the right solution here is to use `std::path::Components` for the look up of the checkpoint rather than the wacky regular expressions that currently exist there. Unfortunately, I'm not fresh enough tonight to brainpower this problem, so maybe over the weekend :smile:

houqp:
[...]

PR #81, opened 2021-02-01
Refactoring Path Management in storage_backend

fvaleye:
Proposal for better handling path in a platform-safe manner (#39, #40)

### DIFF
--- a/rust/src/storage/file.rs
+++ b/rust/src/storage/file.rs
@@ -1,4 +1,5 @@
 use std::pin::Pin;
+use std::path::Path;
 
 use chrono::DateTime;
 use futures::{Stream, TryStreamExt};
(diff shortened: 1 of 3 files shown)


rust/src/storage/file.rs
```diff
@@ -17,6 +18,19 @@ impl FileStorageBackend {
...
+        &self,
+        path: &str,
+        path_to_join: &str,
+    ) -> String {
+        let new_path = Path::new(path);
+        new_path
+            .join(path_to_join)
+            .into_os_string()
+            .into_string()
+            .unwrap()
```
rtyler, line 31: It's hard for me to ever imagine what sequence of events could lead to this `into_string()` call failing, but I will defer to @houqp's judgement on whether `join_path` should propagate its Error

rtyler (commented):
By and large looks good to me, thanks for the pull request!

rust/src/storage/file.rs
houqp: i think it should be fairly safe here since we only have `&str` as inputs.

houqp (approved):
LGTM!

PR #84, opened 2021-02-02
Add basic transaction protocol for writes

xianwill:
This PR adds a basic transaction protocol for executing writes with `delta-rs` (See https://github.com/delta-io/delta-rs/discussions/67).

The primary core code change is an API for creating a delta transaction from a delta table instance and using the transaction to commit actions.

To validate the transaction API and lead into higher-level APIs, this branch also includes a beefy test file ([rust/tests/write_exploration.rs](https://github.com/delta-io/delta-rs/pull/84/commits/ebb39cd446a3eaef65dcf4b55007341cc47a9d26#diff-e4f413a2dddc60ba52a720746b834a03a2dd21d2d2e90aa83190ebf84c37ad41) requires "load diff") for testing the API design with a writer in play. Ultimately, something like this DeltaWriter should probably exist in `delta-rs` but the struct used in test is very naive and just for validating the tx api.

As mentioned in open meet, my idea is that the tx apis should become crate internal and have higher level wrapping apis exposed to end users (such as an "UpdateCommand" api) as `delta-rs` evolves. So the basic API provided here for txs should eventually be pub (crate) only, and we should rely on those higher level commands to ease the end-user API.

Also worth noting - the tx API added here relies on a `put_obj` method added to storage backend, but leaves every storage backend unimplemented other than `file.rs`. `s3` backend will require a guard (i.e. DynamoDb) to fail if the version file exists already.


rust/src/delta.rs
```diff
@@ -412,6 +415,37 @@ impl DeltaTable {
...
+        match self.get_last_checkpoint().await {
+            Ok(last_check_point) => {
+                let should_restore = self.last_check_point.is_none() || 
+                    self.last_check_point.unwrap().version != last_check_point.version;
+
+                if should_restore {
+                    self.last_check_point = Some(last_check_point);
+                    self.restore_checkpoint(last_check_point).await?;
+                    self.version = last_check_point.version + 1;
+                }
```
houqp, line 433: If we derive `Eq` and `PartialEq` trait for `CheckPoint`, then we can rewrite the this block of code without unwrap as:

```suggestion
               if self.last_check_point.is_none()
                   || self.last_check_point == Some(last_check_point)
               {
                    self.last_check_point = Some(last_check_point);
                    self.restore_checkpoint(last_check_point).await?;
                    self.version = last_check_point.version + 1;
                }
```

rust/src/delta.rs
```diff
@@ -512,6 +546,28 @@ impl DeltaTable {
             .collect()
     }
 
+    pub fn get_partition_cols(&self) -> Result<&Vec<String>, DeltaTableError> {
```
houqp, line 549: looks like this function is not used?

rust/src/delta.rs
```diff
@@ -613,6 +673,127 @@ impl std::fmt::Debug for DeltaTable {
...
+            _ => DeltaTransactionError::Storage { source: error },
+        }
+    }
+}
+
+impl From<DeltaTableError> for DeltaTransactionError {
+    fn from(error: DeltaTableError) -> Self {
+        DeltaTransactionError::DeltaTable { source: error }
+    }
+}
```
houqp, line 707: you can avoid this kind of pass through wrapper code by using `from` attribute defined in thiserror:

```rust
    #[error("DeltaTable interaction failed: {source}")]
    DeltaTable {
        #[from]
        source: DeltaTableError,
    },
```

rust/src/delta.rs
```diff
@@ -613,6 +673,127 @@ impl std::fmt::Debug for DeltaTable {
...
+    }
+
+    async fn next_attempt_version(&mut self) -> Result<DeltaDataTypeVersion, DeltaTransactionError> {
+        self.delta_table.update().await?;
+        Ok(self.delta_table.version + 1)
+    }
+}
+
+fn max_retry_commit_attempts() -> u32 {
+    std::env::var("DELTA_MAX_RETRY_COMMIT_ATTEMPTS").map_or(10000000, |s| s.parse::<u32>().unwrap())
```
houqp, line 794: It's better to let application decide how they want to read config value at runtime, through config file, env var, etc. In delta.rs, we can make this configurable by making it a config value of the struct itself. This also avoids global side effects and makes it possible to set different max commit attempts for different tables if needed.

rust/src/storage/file.rs
```diff
@@ -59,4 +60,23 @@ impl StorageBackend for FileStorageBackend {
             })
         })))
     }
+
+    async fn put_obj(&self, path: &str, obj_bytes: &[u8]) -> Result<(), StorageError> {
+        let dir = std::path::Path::new(path);
+        let dir = dir.parent().unwrap();
```
houqp, line 66: we should propagate the error instead of doing unwrap here. basically avoid unwrap at all costs :)

rust/src/storage/file.rs
```diff
@@ -59,4 +60,23 @@ impl StorageBackend for FileStorageBackend {
...
+
+        fs::create_dir_all(dir).await.map_err(|err| StorageError::IO{ source: err })?;
```
houqp, line 68: I think the map_err is not needed?

```suggestion
        fs::create_dir_all(dir).await?;
```

rust/tests/write_exploration.rs
```diff
@@ -0,0 +1,462 @@
+extern crate chrono;
```
houqp, line 1: is this file for demo purpose only? or do you expect it to be merged as part of the patch set?

rust/src/delta.rs
```diff
@@ -613,6 +673,127 @@ impl std::fmt::Debug for DeltaTable {
...
+        // try to commit in a loop in case other writers write the next version first
+        let version = self.try_commit_loop(actions).await?;
+
+        self.delta_table.update().await?;
+
+        Ok(version)
+    }
+
+    async fn try_commit_loop(&mut self, actions: &[Action]) -> Result<DeltaDataTypeVersion, DeltaTransactionError> {
+        let max_retries = max_retry_commit_attempts();
```
houqp, line 748: What's the semantic of retry_commit_attempts? Is it the count of number of max commit actions? or retry count? I think number of max commit actions should equal to retries + 1.

rust/src/delta.rs
```diff
@@ -613,6 +673,127 @@ impl std::fmt::Debug for DeltaTable {
...
+        let mut attempt_number: u32 = 0;
+
+        loop {
+            if attempt_number > max_retries {
+                return Err(DeltaTransactionError::CommitRetriesExceeded);
+            }
```
houqp, line 754: nit pick, we can move this to the bottom of the loop after incremented attempt_number. the first check entering the loop is an no op because it will always be true.

rust/src/delta.rs
```diff
@@ -613,6 +673,127 @@ impl std::fmt::Debug for DeltaTable {
...
+        //     IsolationLevel::SnapshotIsolation
+        // } else { 
+        //     IsolationLevel::Serializable
+        // };
+
+        // TODO: create a CommitInfo action and prepend it to actions.
+
+        // try to commit in a loop in case other writers write the next version first
+        let version = self.try_commit_loop(actions).await?;
+
```
houqp, line 741: perhaps worth adding a note for future optimization, after the commit, we already have the new version and transaction log in memory, so we don't need to call update to fetch it from remote store.

rust/src/delta.rs
```diff
@@ -613,6 +673,127 @@ impl std::fmt::Debug for DeltaTable {
...
+        }
+    }
+
+    async fn try_commit(&mut self, actions: &[Action]) -> Result<DeltaDataTypeVersion, DeltaTransactionError> {
+        // Serialize all actions that are part of this log entry.
+        let jsons: Vec<String> = actions.iter()
+            .map(|a| {
+                serde_json::to_string(a).unwrap()
+            }).collect();
+        let log_entry = jsons.join("\n");
```
houqp, line 772: looks like this is something that could be assembled by the caller instead? I can see a case where we might need to retry a commit with the same set of actions in append only mode due to version conflicts, in that case, we can avoid doing the same json serialization again.

also, we should propagate `serde_json::to_string` error to the caller :)

rust/tests/write_exploration.rs
xianwill: i was planning to merge it as a guidepost for future iterations of write support, but I'm happy to remove it if you think that is better.

rust/src/delta.rs
xianwill: ah. good thought. perhaps a `DeltaTransactionOptions` to pass to `new` then in case we have other configs to pass in in the future.

rust/src/delta.rs
xianwill: Hmmm. This is a nice idea, but do you think the semantics are okay? Would all callers be okay by defining `version` as the metric of equality even when compared instances may have different `size` and `parts`? Ideally, it would not be possible to get into such a state I suppose.

rust/src/delta.rs
xianwill: fixed

rust/src/delta.rs
xianwill: neat. TIL... fixed.

rust/src/delta.rs
xianwill: fixed

rust/src/storage/file.rs
xianwill: sort of fixed - used an `if let ...` guard to avoid the unwrap

rust/src/delta.rs
xianwill: you're absolutely right. fixed.

rust/src/delta.rs
xianwill: fixed

rust/src/delta.rs
xianwill: fixed

rust/src/delta.rs
xianwill: good call. added note.

rust/src/delta.rs
xianwill: I have a local commit that uses `version` as equality metric if you think the semantics are okay. I can push if we are on the same page and agree on semantics. As it stands, I do not see any usages that would corrupt the checkpoint in a way that invalidates the version equality check.

Contents are:

```
impl PartialEq for CheckPoint {
    fn eq(&self, other: &Self) -> bool {
        self.version == other.version
    }
}

impl Eq for CheckPoint {}
```

rust/src/delta.rs
```diff
@@ -613,6 +662,143 @@ impl std::fmt::Debug for DeltaTable {
...
+        self.delta_table.update().await?;
+
+        Ok(version)
+    }
+
+    async fn try_commit_loop(&mut self, log_entry: &[u8]) -> Result<DeltaDataTypeVersion, DeltaTransactionError> {
+        let mut attempt_number: u32 = 0;
+
+        loop {
+            let commit_result = self.try_commit(log_entry).await;
```
rtyler, line 768: This is offering no means for the caller to figure out why the commit is failing which I imagine is going to lead to some frustration down the road.

@houqp how would you feel about at a minimum logging an error or debug level message about the error attached to `commit_result`?

rust/src/delta.rs
```diff
@@ -613,6 +662,143 @@ impl std::fmt::Debug for DeltaTable {
...
+
+            if commit_result.is_ok() {
+                return commit_result;
+            }
+
+            attempt_number += 1;
+
+            if attempt_number > self.options.max_retry_commit_attempts + 1 {
+                return Err(DeltaTransactionError::CommitRetriesExceeded);
```
rtyler, line 777: Related to the above comment, perhaps in the case of retries being exhausted, the `CommitRetriesExceeded` could carry an `inner` Error out from `try_commit`?

Basically I'm thinking of the remote storage backends where commit errors are likely to be much more common, and how the caller could be left in the dark without propagating this information out.

rust/src/delta.rs
```diff
@@ -613,6 +662,143 @@ impl std::fmt::Debug for DeltaTable {
...
+            _ => DeltaTransactionError::Storage { source: error },
+        }
+    }
+}
+
+pub struct DeltaTransactionOptions {
+    max_retry_commit_attempts: u32,
+}
+
+const DEFAULT_DELTA_MAX_RETRY_COMMIT_ATTEMPTS: u32 = 10000000;
```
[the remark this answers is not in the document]
xianwill: It is in fact 10 million retries (not ms). Feels a bit on the large side. I chose this value since it is the default used in the reference implementation. Open to other considerations.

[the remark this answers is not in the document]
xianwill: I did go ahead and commit your suggested change fwiw.

rust/src/delta.rs
```diff
@@ -7,7 +7,7 @@ use std::io::{BufRead, BufReader, Cursor};
 
 use chrono::{DateTime, FixedOffset, Utc};
 use futures::StreamExt;
-use parquet::errors::ParquetError;
+use parquet::{errors::ParquetError};
```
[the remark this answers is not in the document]
xianwill: fixed

rust/src/delta.rs
xianwill: @rtyler, regarding this comment and the one above, I have [decomposed the error types and added an inner to CommitRetriesExceeded](https://github.com/delta-io/delta-rs/pull/84/files#diff-3b5c80af9e5527fbeaa6dad151a835f0733a26e61d0ba0471c58f243a29dfa99R666).

I'll be happy to add error/warn logging as well, but will wait for @houqp to chime in since you addressed him directly.

rust/src/delta.rs
```diff
@@ -527,6 +572,10 @@ impl DeltaTable {
         }
     }
 
+    pub fn create_transaction(&mut self, options: Option<DeltaTransactionOptions>) -> DeltaTransaction {
```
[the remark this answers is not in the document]
xianwill: Yeah, I definitely get your opinion here. And since I already have a `pub new` on the `DeltaTransaction` I could just delete this helper. I don't have any huge master plan behind it, its just a common helper API I'm used to seeing in other contexts. 

I'll wait a bit before making the change in case anyone else wants to comment.

rust/src/action.rs
```diff
@@ -762,3 +762,18 @@ mod tests {
         );
     }
 }
+
+pub enum SaveMode {
```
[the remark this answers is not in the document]
xianwill: sprouted some rust docs in https://github.com/delta-io/delta-rs/pull/84/commits/38fed27af6a2bf065f22b3c9fbb2a03b25041ce5

rust/src/delta.rs
xianwill: newly committed changes cut across a few different comments. might be best to start from https://github.com/delta-io/delta-rs/pull/84/files#diff-3b5c80af9e5527fbeaa6dad151a835f0733a26e61d0ba0471c58f243a29dfa99R796 for re-review.

rust/src/delta.rs
xianwill: newly committed changes cut across a few different comments. might be best to start from https://github.com/delta-io/delta-rs/pull/84/files#diff-3b5c80af9e5527fbeaa6dad151a835f0733a26e61d0ba0471c58f243a29dfa99R796 for re-review.

rust/src/delta.rs
houqp: yeah, i think this is a reasonable performance optimization, i can't think of a case why size and parts would be different for the same version other than bugs.

rust/src/delta.rs
houqp: yeah, I think we should break the loop and propagate the error back to caller immediately if the error cannot be resolved by optimistic retry. For example, `storage.put_obj` call in `try_commit` can fail for network or authentication error, which will always fail on retry. Another example would be unresolvable transaction conflicts, which will require the caller discard the transaction entirely and redo the compute from scratch.

For errors like `"the next version you picked is already committed another writer"`, we can safely retry without logging any error because this is expected and normal behavior.

I do think it's valuable to add a debug logging in all cases, whether the error can be resolved by optimistic retry or not.

Lastly, as a future improvement, I think it would be cool to return retry count as part of the result to the caller so that can be collected and sent to whatever monitoring system being used. This way, application owner can keep an eye on whether the system is spending too much time on doing busy retries.

rust/src/delta.rs
```diff
@@ -613,6 +662,175 @@ impl std::fmt::Debug for DeltaTable {
...
+            delta_table,
+            options: options.map_or_else(|| DeltaTransactionOptions { 
+                max_retry_commit_attempts: DEFAULT_DELTA_MAX_RETRY_COMMIT_ATTEMPTS 
+            }, |o| o),
+        }
+    }
+
+    /// Commits the given actions to the delta log.
+    /// This method will retry the transaction commit based on the value of `max_retry_commit_attempts` set in `DeltaTransactionOptions`.
+    pub async fn commit_all(&mut self, actions: &[Action], _operation: Option<DeltaOperation>) -> Result<DeltaDataTypeVersion, DeltaTransactionError> {
```
houqp, line 754: After thinking more this, I agree with @nevi-me that it would be handy to maintain the list of staged actions within the transaction struct itself. I can see for more complex transactions, users would do some non-trivia computation, produce some actions, then do so other non-trivia computations and produce other actions. It would be a mental burden for them to having to keep track of all the actions needed to commit at the very end v.s. being able to "queue" those actions as the flow of application logic goes on.

If we remain all the staged actions in an internal vector and add a `commit` method to transaction struct, then `commit_all` can be simplified to a helper method that adds more action to the vector and calls `commit`. So users can still use this interface for simple transactions.

I can see why you didn't want to maintain an internal vector though, it's extra work to deal with action struct clone or reference passing :P But it looks like it should be pretty straight forward to derive `Clone` for all the action types, then we don't need to deal with lifetime and references. Given this is not a hot code path, I think doing clone on every action would be a reasonable compromise.

rust/src/delta.rs
xianwill: @houqp - I have added [this commit](https://github.com/delta-io/delta-rs/pull/84/commits/f46833da900c959dfa93b938d425e79b06842c6b) which fails fast on anything other than an optimistic concurrency retry based on version number and also adds `debug` logging to all commit failure cases. The pattern matching approach I have taken does undo your previous [mention of holding the attempt number check 'til the end](https://github.com/delta-io/delta-rs/pull/84#discussion_r570000729) since pattern match needs to start with most restrictive check first. Let me know you think the single pattern match is worth the trade-off or would prefer additional lines to optimize. I personally tend to prefer fewer code lines when possible.

rust/src/delta.rs
```diff
@@ -613,6 +662,175 @@ impl std::fmt::Debug for DeltaTable {
...
+
+impl<'a> DeltaTransaction<'a> {
+    /// Creates a new delta transaction.
+    /// Holds a mutable reference to the delta table to prevent outside mutation while a transaction commit is in progress.
+    /// Transaction behavior may be customized by passing an instance of `DeltaTransactionOptions`.
+    pub fn new(delta_table: &'a mut DeltaTable, options: Option<DeltaTransactionOptions>) -> Self {
+        DeltaTransaction { 
+            delta_table,
+            options: options.map_or_else(|| DeltaTransactionOptions { 
+                max_retry_commit_attempts: DEFAULT_DELTA_MAX_RETRY_COMMIT_ATTEMPTS
```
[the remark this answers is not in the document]
xianwill: Another neat TIL. Thanks. Implemented this approach in https://github.com/delta-io/delta-rs/pull/84/commits/f46833da900c959dfa93b938d425e79b06842c6b#diff-3b5c80af9e5527fbeaa6dad151a835f0733a26e61d0ba0471c58f243a29dfa99R735.

rust/src/delta.rs
xianwill: Your comment above sounds like a nice middle-ground to me. For now, what I have done is to rename `commit_all` to `commit_with` in https://github.com/delta-io/delta-rs/pull/84/commits/e288fc065eb55695cf88676f09f5625390d37667. `commit_all` as a name "sounds" exclusive to me such that callers might expect _all and only actions passed in the call_ to be committed. `commit_with` sounds more inclusive to me as a method that would include actions stored within the tx to be committed along with actions passed to the method. Basically just trying to future-proof the naming a bit.

I agree your suggestion should be pretty straightforward to implement. I would be tempted to implement it immediately, but I feel like it is a semantic I would like to spend more time with outside of laying the foundation.

rust/tests/write_exploration.rs
xianwill: Excellent. I personally think it is good and fine to run this test in CI.

rust/src/delta.rs
xianwill: Also - I added a note to potentially include `num_retries` in the error object later per your suggestion.

rust/src/delta.rs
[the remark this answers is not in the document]
xianwill: Your mention @houqp is exactly what I was shooting for in the implementation - support either client code path. I am very friendly to @rtyler's suggestion as well though - abstractions can get leaky and cause problems down the road if they are not necessary and I do believe it is good to be very deliberate and explicit when defining your APIs, but I'm just so used to seeing this style of helper method elsewhere that it almost feels like a de facto standard helper method that should come out of the box. I don't have a super strong opinion myself regarding whether the helper should be included in the first commit myself. One thing worth mentioning is that it would be super easy to remove this helper for now and add it back later to be a bit "non-committal". Other than that, I will leave it to you guys to decide if this helper should be included in the initial API. Will watch for further comments.

rust/src/delta.rs
xianwill: Nice. pushed code change.

rust/src/delta.rs
houqp: I am also for achieving the same goal with least amount of code :)

This is just my personal preference: I find the logic a little bit easier for to follow with a single match branch and move the max retry check into the branch as a nested if block. Another benefit with single branch is we would never have to worry about restrictiveness of match branches going forward, although I think the compiler will likely complain if an incorrect order results in unreachable code.

houqp (approved):
Great work @xianwill ! LGTM :) @rtyler you want to take a final look?

houqp:
@xianwill also don't forget to fix the linter error found by clippy :)

rust/src/delta.rs
```diff
@@ -613,6 +669,192 @@ impl std::fmt::Debug for DeltaTable {
...
+}
+
+impl<'a> DeltaTransaction<'a> {
+    /// Creates a new delta transaction.
+    /// Holds a mutable reference to the delta table to prevent outside mutation while a transaction commit is in progress.
+    /// Transaction behavior may be customized by passing an instance of `DeltaTransactionOptions`.
+    pub fn new(delta_table: &'a mut DeltaTable, options: Option<DeltaTransactionOptions>) -> Self {
+        DeltaTransaction { 
+            delta_table,
+            options: options.map_or_else(DeltaTransactionOptions::default, |o| o),
```
houqp, line 760: If there is no transformation needed, we can use `unwrap_or` instead:

```suggestion
            options: options.unwrap_or(DeltaTransactionOptions::default),
```

xianwill: Nice. Thanks.

rtyler:
I won't have the time to give this serious consideration this week, if @houqp approves, just merge it IMO (after conflicts are corrected obviously :smile_cat: )

xianwill:
@houqp - I do not see any merge conflicts, but I also do not have permission to merge. I have fixed last mentioned clippy error already fwiw.

Issue #82, opened 2021-02-01
race condition in s3 tests


houqp:
Our S3 test is failing randomly at the moment. Based on my quick glance, the specific error message we are getting is usually caused by access of global hyper client from multiple threads. However, that doesn't seem to be the case for us because we are not using a global S3 client. Perhaps rusoto is doing something funny with hyper? One thing I noticed is if I only run one of the S3 tests, it passes reliably.

xianwill:
@houqp - I spent a few minutes to to look into this, and as per your theory, rusoto does seem to use [a shared singleton hyper client](https://github.com/rusoto/rusoto/blob/master/rusoto/core/src/client.rs#L28) as per your mentioned theory.

I'm not sure how we might accommodate/work-around this, or if this might be the final nail (in addition to the announcement that  [rusoto is in maintenance mode](https://github.com/rusoto/rusoto#maintenance-status)) to migrate to a different s3 client. 

I would be interested in looking into this further time permitting, but my available time for this specific problem is somewhat limited. So for now, I will leave this comment here for further development.

xianwill:
@houqp + @rtyler - I have submitted https://github.com/delta-io/delta-rs/pull/87 as a workaround for CI. I do still have concerns for multi-threaded processes using delta-rs that might experience the same problem. For now, this is a lo-pro issue aside from CI from my perspective since I expect to only have a single S3 writer thread per-process for my production use case.

houqp:
relevant upstream bug report: https://github.com/rusoto/rusoto/issues/1467

xianwill:
@houqp - since CI is resolved by https://github.com/delta-io/delta-rs/pull/87, should we resolve this issue or keep it open for upstream changes in https://github.com/rusoto/rusoto/issues/1467?

PR #86, opened 2021-02-16
Add serial test execution for s3_test suite

xianwill:
As described in https://github.com/delta-io/delta-rs/issues/82, rusoto uses a shared global hyper client that causes problems in parallel test execution. 

This PR adds [serial_test](https://docs.rs/serial_test/0.4.0/serial_test/) as a dev dependency and tags each function in `s3_test` as `#[serial]` to workaround the test execution context.

Ultimately, this _could_ mask problems when supporting multi-threaded delta-rs users, but we don't know how that landscape will shape yet, so perhaps this is a good solution in the interim to fix CI for now.


xianwill:
Closing to remove accidental file commits

PR #90, opened 2021-02-22
Add schema in Python bindings

fvaleye:
Add bindings for getting the delta table schema in Python: `DELTA` and `ARROW` formats.


houqp:
@fvaleye this is ready for review right? or are you still working on it?

fvaleye:
@houqp Yes, it is ready for review!

python/deltalake/__init__.py
```diff
@@ -36,3 +57,18 @@ def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
 
     def to_pyarrow_table(self) -> pyarrow.Table:
         return self.to_pyarrow_dataset().to_table()
+
+    def to_pyarrow_schema(self) -> pyarrow.Schema:
```
houqp, line 61: since we are not converting a delta table to  pyarrow schema here, i think `pyarrow_schema` would have been a better method name.

python/deltalake/__init__.py
```diff
@@ -36,3 +57,18 @@ def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
...
+        format = DeltaTableSchemaFormat.ARROW
+        delta_fields = [
+            DeltaTableField(
```
houqp, line 64: Looks like the `DeltaTableField` is not being used at all? Why not just invoke `pyarrow.field` directly in this list comprehension loop?

python/deltalake/schema.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class DeltaTableSchema:
+    def __init__(self, fields: List[DeltaTableField]):
+        self.fields = fields
+
+    def __str__(self):
+        for field in self.fields:
+            print(field)
+        return ""
```
houqp, line 40: `__str__` method needs to return the representation as a string, not print it out in stdout :)

python/deltalake/schema.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __repr__(self):
+        return self.__str__()
+
+
+class DataType:
+    @classmethod
+    def from_delta_json(cls, json_dict: Dict[str, Any]):
```
houqp, line 48: could you annotate the method return type as well?

rust/src/schema.rs
```diff
@@ -22,6 +23,11 @@ impl SchemaTypeStruct {
     pub fn get_fields(&self) -> &Vec<SchemaField> {
         &self.fields
     }
+
+    pub fn to_json(&self) -> Value {
```
houqp, line 27: I don't think this method is needed, `Deserialize` trait + serde_json should handle this automatically for us. So you should be able to do something like this:

```rust
let s = SchemaTypeStruct{};
let json_s: serde_json::Value = s;
```

rust/src/schema.rs
```diff
@@ -54,11 +60,20 @@ impl SchemaField {
     pub fn get_metadata(&self) -> &HashMap<String, String> {
         &self.metadata
     }
+
+    pub fn to_json(&self) -> Value {
```
houqp, line 64: same here, we can leverage serde to avoid having to implement these `to_json` methods.

python/src/lib.rs
```diff
@@ -60,6 +104,28 @@ impl RawDeltaTable {
     pub fn file_paths(&self) -> PyResult<Vec<String>> {
         Ok(self._table.get_file_paths())
     }
+
+    pub fn schema(&self, format: &str) -> PyResult<DeltaTableSchema> {
+        let schema = match &format[..] {
+            "ARROW" => {
+                <ArrowSchema as From<&deltalake::Schema>>::from(self._table.schema().unwrap())
```
houqp, line 111: we should propagate the error through `PyResult` instead of calling unwraps. In general, `unwraps` and `panic` should be avoided.

houqp (changes requested):
Good change overall 👍  I will need some more time to think about what's the best way to handle conversion between Delta Schame and pyarrow schemas. I have a feeling that it would be cleaner if we add an extra `arrow_schema` method instead of passing format as string argument.

python/deltalake/__init__.py
fvaleye: That's right, I will change the method_name to `pyarrow_schema`

python/deltalake/__init__.py
fvaleye: The `DeltaTableField` is used for extracting the Arrow data type in JSON format given by the `schema()` method.

The method `from_json()`(available in Rust) is not available in Python for Arrow (https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html).

The main idea for the binding between Python and Rust was to use the `JSON` format for the data type binding. It avoids creating all the pyo3 structures for the binding of all the structures in the delta schema.

python/deltalake/schema.py
fvaleye: Yes, forgot about that one :)

python/deltalake/schema.py
fvaleye: Done!

fvaleye: I revert the custom class typing because the import needed was not working for all the Python versions.

houqp:
@fvaleye sent a PR to your fork at https://github.com/fvaleye/delta-rs/pull/1/files to simplify the design a bit, please review and comment there :)

houqp (approved):
Thanks for adding the tracking ticket :)

Issue #91, opened 2021-02-24
Handle the decimal data type in the arrow format conversion


fvaleye:
When the delta table has a `decimal(_,_)` column type, it is not handled by the ArrowSchema conversion (https://github.com/delta-io/delta-rs/blob/main/rust/src/delta_arrow.rs#L42). We could use pattern matching for extracting precision and scale for generating an `ArrowDataType::Decimal(_,_)` value.

houqp:
Yep, that's a good idea :+1:

fvaleye:
It is resolved by #92, I close the issue!

PR #92, opened 2021-02-25
Handle decimal type in arrow schema

fvaleye:
Handle decimal type on a delta table schema #91

### DIFF
--- a/rust/src/delta_arrow.rs
+++ b/rust/src/delta_arrow.rs
@@ -1,4 +1,5 @@
 use regex::Regex;
+use lazy_static::lazy_static;
 use arrow::datatypes::{
     DataType as ArrowDataType, Field as ArrowField, Schema as ArrowSchema, TimeUnit,
 };
@@ -40,7 +41,9 @@ impl From<&schema::SchemaDataType> for ArrowDataType {
     fn from(t: &schema::SchemaDataType) -> Self {
         match t {
             schema::SchemaDataType::primitive(p) => {
-                let decimal_regex = Regex::new(r"\((\d{1,2}),(\d{1,2})\)").unwrap();
+                lazy_static! {
+                    static ref DECIMAL_REGEX: Regex = Regex::new(r"\((\d{1,2}),(\d{1,2})\)").unwrap();
+                }
                 match p.as_str() {
                     "string" => ArrowDataType::Utf8,
                     "long" => ArrowDataType::Int64, // undocumented type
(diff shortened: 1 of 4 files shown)


rust/src/delta_arrow.rs
```diff
@@ -47,8 +49,12 @@ impl From<&schema::SchemaDataType> for ArrowDataType {
                     "byte" => ArrowDataType::Int8,
                     "float" => ArrowDataType::Float32,
                     "double" => ArrowDataType::Float64,
-                    "boolean" => ArrowDataType::Boolean,
```
nevi-me, line 50: Why does your change remove "boolean" and "binary"?

fvaleye: Oh, good catch! We don't want to change this :) I fix it.

rust/src/delta_arrow.rs
```diff
@@ -39,6 +40,7 @@ impl From<&schema::SchemaDataType> for ArrowDataType {
     fn from(t: &schema::SchemaDataType) -> Self {
         match t {
             schema::SchemaDataType::primitive(p) => {
+                let decimal_regex = Regex::new(r"\((\d{1,2}),(\d{1,2})\)").unwrap();
```
houqp, line 43: we should wrap this with lazy static to repetitive compilation: https://docs.rs/regex/1.4.3/regex/#example-avoid-compiling-the-same-regex-in-a-loop

houqp (changes requested):

houqp:
The rest looks good to me 👍 . I also filed https://github.com/delta-io/delta-rs/issues/93 as a follow up item.

houqp:
Given that there is no mention of decimal type in the official delta spec: https://github.com/delta-io/delta/blob/master/PROTOCOL.md, could you share where see it being used?

nevi-me:
> Given that there is no mention of decimal type in the official delta spec: https://github.com/delta-io/delta/blob/master/PROTOCOL.md, could you share where see it being used?

I think it's an omission on the document. I'll check this with a test file, and revert back.

fvaleye:
@houqp I tested my branch (https://github.com/delta-io/delta-rs/pull/90) on our production Delta tables and an error was raised: 

```
exception:
thread '<unnamed>' panicked at 'unexpected delta schema type: decimal(20,2)', rust/src/delta_arrow.rs:62:25
  File "/Users/florianvaleye/Documents/workspace/delta-rs/python/venv/lib/python3.9/site-packages/deltalake/__init__.py", line 43, in schema
    for field in self._table.schema(format=format.value).schema
pyo3_runtime.PanicException: unexpected delta schema type: decimal(20,2)

```

Example of a field with decimal type in the fields section of the delta table transaction:
`{"name":"price","type":"decimal(20,2)","nullable":true,"metadata":{}},`

houqp:
awesome, in that case, @fvaleye you should also send a PR to https://github.com/delta-io/delta/blob/master/PROTOCOL.md :D

houqp (approved):

PR #96, opened 2021-02-28
Fix the decimal type in Python schema.

fvaleye:
- The method `type_for_alias()` of pyarrow is not available for the decimal type, I changed the conversion by creating the decimal128 type directly
- Add test on the decimal conversion for arrow schema

### DIFF
--- a/rust/tests/delta_arrow_test.rs
+++ b/rust/tests/delta_arrow_test.rs
@@ -1,0 +1,14 @@
+extern crate deltalake;
+use arrow::datatypes::DataType as ArrowDataType;
+
+#[test]
+fn test_arrow_from_delta_decimal_type() {
+    let precision = 20;
+    let scale = 2;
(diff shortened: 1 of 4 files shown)


rust/tests/delta_arrow_test.rs
```diff
@@ -0,0 +1,14 @@
+extern crate deltalake;
+use arrow::datatypes::DataType as ArrowDataType;
+
+#[tokio::test]
+async fn test_arrow_from_delta_decimal_type() {
```
houqp, line 5: Looks like there is no async function involved, so better define it as a sync test. It will get us a much more readable stack trace.

```suggestion
#[test]
fn test_arrow_from_delta_decimal_type() {
```

houqp (commented):
Good catch

rust/tests/delta_arrow_test.rs
fvaleye: Done!

houqp (approved):

PR #104, opened 2021-03-03
Add proper StorageError::NotFound error handling for s3 HEAD request

### LINKED ISSUE #103: StorageError::NotFound is not returned by head_obj on missing object
`StorageError::NotFound` error is not produced by error handlers https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/mod.rs#L222 due to upstream issue https://github.com/rusoto/rusoto/issues/716 where `HeadObjectError::NoSuchKey` is not returned as expected for missing object.

I reproduced issue only with `head_obj` call. It seems that rusoto tries to parse response body for HEAD call, however it is empty

mosyp:
Fixes #103.

In upstream issue users report that GET does not return `GetObjectError::NoSuchKey`, however that wan't the case for me when I tested it. It makes sense this to be reproducible only for HEAD calls, since rusoto has shared parsers for both which parses response body, however there's none for HEAD. Hence, response code is used to determine Not Found response

### DIFF
--- a/rust/src/storage/mod.rs
+++ b/rust/src/storage/mod.rs
@@ -222,6 +222,9 @@ impl From<RusotoError<rusoto_s3::HeadObjectError>> for StorageError {
             RusotoError::Service(rusoto_s3::HeadObjectError::NoSuchKey(_)) => {
                 StorageError::NotFound
             }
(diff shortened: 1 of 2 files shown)


xianwill (approved):
LGTM

rtyler (approved):

PR #107, opened 2021-03-05
Add missing fields in Remove Action.

fvaleye:
With Delta 0.8 version, the RemoveFile has 4 additional fields (https://github.com/delta-io/delta/blob/master/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L276):
- extendedFileMetadata: true or false (for protocol compatibility reasons)
- partitionValues
- size
- tags

It is not yet available in the PROTOCOL specification (https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file), I will submit a PR for it.

I tested this code with Delta Table v0.8 version tables but it could worth adding a table with the v0.8 version for the test

@xianwill: you may look at this change if it changes something in the low_level transaction writing


rust/tests/read_delta-0-2-0_test.rs
```diff
@@ -24,6 +24,10 @@ async fn read_delta_2_0_table_without_version() {
             path: "part-00000-512e1537-8aaa-4193-b8b4-bef3de0de409-c000.snappy.parquet".to_string(),
             deletionTimestamp: 1564524298213,
             dataChange: false,
+            extendedFileMetadata: None,
+            partitionValues: None,
+            tags: None,
+            size: None
```
houqp, line 30: you can define a default trait for remove action to reduce boilerplates here if you want.

houqp (approved):
+1 on adding 0.8 test table

rust/tests/read_delta-0-2-0_test.rs
fvaleye: Thanks for the tip!

houqp (approved):

houqp:
:tada:

Issue #93, opened 2021-02-26
Get rid of panics and unwraps in Delta Schema to Arrow Schema conversion code


houqp:
To remove all panics and unwraps, we need to change the `Try` trait implementation to `TryFrom` trait so we can propagate the error back to caller through Result.

PR #108, opened 2021-03-07
Refactoring of ArrowSchema conversion

fvaleye:
Refactoring of ArrowSchema, using `TryForm` trait and raise an `ArrowError` when there is an error in the conversion of the Schema.
Related to the issue: #93

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -5,6 +5,7 @@ use std::collections::HashMap;
 use std::fmt;
 use std::io::{BufRead, BufReader, Cursor};
 
+use arrow::error::ArrowError;
 use chrono::{DateTime, FixedOffset, Utc};
 use futures::StreamExt;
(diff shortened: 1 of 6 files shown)


rust/src/delta.rs
```diff
@@ -60,6 +61,11 @@ pub enum DeltaTableError {
         #[from]
         source: ParquetError,
     },
+    #[error("Failed to convert into Arrow schema: {}", .source)]
```
houqp, line 64: Not a blocker and can be addressed as follow up PRs. I think we might need a more specific error variant here because in the future, other part of the code path could trigger ArrowError as well which is not necessarily schema related, when this happens, the error message could be misleading.

houqp (approved):
Nice clean up :+1:

rtyler (approved):
Nice cleanup :+1:

PR #116, opened 2021-03-11
add crates.io and api doc link to rust workspace readme

houqp:
# Description

Add crates.io and api doc link to rust workspace readme.


rtyler (approved):
Not going to be much use until we get another proper release, but a good idea nonetheless :smile_cat:

PR #126, opened 2021-03-15
Make transaction log commit with filesystem storage backend to be atomic

### LINKED ISSUE #94: Transaction log commit with filesystem storage backend is not atomic
We should write out the new transaction log into a temporary file first, then use atomic file rename to create the new transaction log instead. Current implementation writes directly to the new transaction log path, which could lead to log corruption if process crashes before completing the write.

mosyp:
# Description

This PR makes transaction log commit by writing file in `put_obj` first in temporary file and once that successful then uses atomic rename (with checking on destination existence) to move temporary file into final commit file.

# Related Issue(s)
Fixes #94

# Documentation
- Linux `renameat2` https://man7.org/linux/man-pages/man2/rename.2.html
- Mac `renamex_np` https://www.unix.com/man-page/mojave/2/renamex_np/


rust/src/storage/file.rs
```diff
@@ -58,26 +65,110 @@ impl StorageBackend for FileStorageBackend {
...
     async fn put_obj(&self, path: &str, obj_bytes: &[u8]) -> Result<(), StorageError> {
-        let dir = std::path::Path::new(path);
-
-        if let Some(d) = dir.parent() {
-            fs::create_dir_all(d).await?;
-        }
+        let tmp_file = self
+            .root
+            .join("_tmp_writes")
+            .join(Uuid::new_v4().to_string());
```
mosyp, line 71: I chose the relative path within table instead of env tmp (`/tmp`) because `rename(2)` will fail if source and destination will be at different partitions

rust/src/storage/file.rs
```diff
@@ -58,26 +65,110 @@ impl StorageBackend for FileStorageBackend {
...
-            .create_new(true)
-            .write(true)
-            .open(path)
-            .await
-            .map_err(|e| match e.kind() {
-                std::io::ErrorKind::AlreadyExists => StorageError::AlreadyExists(path.to_string()),
-                _ => StorageError::Io { source: e },
-            })?;
+        create_parent_dirs(&tmp_file).await?;
+        create_parent_dirs(Path::new(path)).await?;
```
mosyp, line 74: Not a big fan of calling this on each time. I'm thinking of moving `put_obj` into helper function and use a wrapper with error handling to catch missing/uncreated parent dirs and create them once and when needed instead of calling create_parent_dirs each time. Wdyt?

rust/src/storage/file.rs
```diff
@@ -58,26 +65,110 @@ impl StorageBackend for FileStorageBackend {
...
+        flags: c_uint,
+    ) -> c_int;
+
+    #[cfg(target_os = "macos")]
+    fn renamex_np(from: *const c_char, to: *const c_char, flags: c_uint) -> c_int;
+}
+
+fn rename(from: &str, to: &str) -> Result<(), StorageError> {
+    let from_s = CString::new(from).unwrap();
+    let to_s = CString::new(to).unwrap();
```
mosyp, line 114: is it safe to call unwrap here?
`?` operator didn't work since the `From` for returned `Err` is missing. Should I leave it as is or create new `From` impl

rust/src/storage/file.rs
```diff
@@ -58,26 +65,110 @@ impl StorageBackend for FileStorageBackend {
...
-                std::io::ErrorKind::AlreadyExists => StorageError::AlreadyExists(path.to_string()),
-                _ => StorageError::Io { source: e },
-            })?;
+        create_parent_dirs(&tmp_file).await?;
+        create_parent_dirs(Path::new(path)).await?;
 
+        let mut f = fs::File::create(&tmp_file).await?;
         f.write(obj_bytes).await?;
 
+        rename(tmp_file.to_str().unwrap(), path)?;
```
mosyp, line 79: `to_str` on path returns Option, leave it as is with `unwrap` or there's fancy workaround here?

mosyp (commented):
I've got several concern here
- If `rename` fails due to concurrent commits, it'll start over again with incremented version which means that we'll have it writes tmp file again and again. However the impact will be on higher level within `try_commit_loop` which is out of backend storage lvl. 
- Current impl does not remove tmp files once it fails, but this should be added/revised once previous concern is resolved

rust/src/storage/file.rs
houqp: You can avoid this unwrap by changing signature of rename to accept `AsRef<Path>`, basically something like [this Rust playground example](https://play.rust-lang.org/?code=%23!%5Ballow(unused)%5D%0Afn%20main()%20%7B%0Ause%20std%3A%3Apath%3A%3A%7BPath%2C%20PathBuf%7D%3B%0A%0Alet%20path%20%3D%20Path%3A%3Anew(%22%2Ftmp%2Ffoo.txt%22)%3B%0Aassert_eq!(path.with_file_name(%22bar.txt%22)%2C%20PathBuf%3A%3Afrom(%22%2Ftmp%2Fbar.txt%22))%3B%0A%0Alet%20path%20%3D%20Path%3A%3Anew(%22%2Ftmp%22)%3B%0Aassert_eq!(path.with_file_name(%22var%22)%2C%20PathBuf%3A%3Afrom(%22%2Fvar%22))%3B%0A%7D&edition=2018).

Then inside rename function convert Path to OsString, then to CString.

rust/src/storage/file.rs
houqp: In theory it's safe, but I think it's better to propagate the error back to caller. You don't have to implement `From` trait for `NulError`, you can also just apply `map_err` with a named or anonymous function.

rust/src/storage/file.rs
```diff
@@ -58,26 +65,110 @@ impl StorageBackend for FileStorageBackend {
...
+fn rename(from: &str, to: &str) -> Result<(), StorageError> {
+    let from_s = CString::new(from).unwrap();
+    let to_s = CString::new(to).unwrap();
+
+    unsafe {
+        cfg_if::cfg_if! {
+            if #[cfg(target_os = "linux")] {
+                if renameat2(0, from_s.as_ptr(), 0, to_s.as_ptr(), RENAME_NOREPLACE) == 0 {
+                   return Ok(())
+                }
```
houqp, line 121: we should probably use something like https://crates.io/crates/errno to check for the error code and return a generic non-retryable error, otherwise it can get stuck in retry loop on stuff like permission error. I believe `EEXIST` is the error code that should be mapped to `StorageError::AlreadyExists`.

rust/src/storage/file.rs
mosyp: yeap, that's more I like it

rust/src/storage/file.rs
mosyp: yeap, make sense

rust/src/storage/file.rs
mosyp: Done that in https://github.com/delta-io/delta-rs/pull/126/commits/44805f34b88b09a466b5502ee295a4aa184f4044#diff-eed9f66f54ec53b6ccce2167c4fb49e27e075536e5d46390dfcd3fc92da18d2bL112, but still needed some map_err within rename, which was better. 

However I found that there's `join_path` which works with `String`s, and I needed `Path` only for joining. So basically reverted all changes for `Path`/`PathBuf` and using plain string sinstead.

rust/src/storage/file.rs
mosyp: Done that, howevevr I've introduced new `Generic` error with plain text https://github.com/delta-io/delta-rs/pull/126/files#diff-8e70f44f4dac2ac3e0e0e1d3d3ebfcad6032c979eab15a5b5c97984753cfd546R163, wdyt?

rust/src/storage/file.rs
mosyp: Well, will test that with linux later, but in mac it's always `-1` for me. I feel like the safest way (at least for mac) is to test on `exists` call once `-1` is received so we won't get an inf loop

p.s. `-1` was received in mac for each case
- b already exists
- b has uncreated parent dirs
- a is missing

mosyp: ugh, didn't know that, from docs https://www.unix.com/man-page/mojave/2/renamex_np/
```
     A 0 value is returned if the operation succeeds, otherwise rename() returns -1 and the global variable errno indicates the reason for the failure
 ```    
yeap, https://crates.io/crates/errno worth a try, otherwise I figured how it works now

Comment (mosyp) [reply rust/src/storage/file.rs]: addressed that in https://github.com/delta-io/delta-rs/pull/126/commits/c8dd3823fc338ef29a65245679c82a9ad45b0587

Comment (mosyp) [reply rust/src/storage/file.rs]: also moved rename into separate module as it has grown big

File: rust/src/storage/rename.rs [hunk 6]
```diff
@@ -0,0 +1,110 @@
...
+
+pub fn rename(from: &str, to: &str) -> Result<(), StorageError> {
+    let ret = unsafe {
+        let from = to_c_string(from)?;
+        let to = to_c_string(to)?;
+        platform_specific_rename(from.as_ptr(), to.as_ptr())
+    };
+
+    if ret != 0 {
+        std::fs::remove_file(from).unwrap_or(());
```
houqp, line 26: i feel like it would be less surprise to let caller handle the clean up so semantically it's the same as the rename system call.

rust/src/storage/file.rs
```diff
@@ -58,26 +63,14 @@ impl StorageBackend for FileStorageBackend {
...
-                std::io::ErrorKind::AlreadyExists => StorageError::AlreadyExists(path.to_string()),
-                _ => StorageError::Io { source: e },
-            })?;
+        let tmp_file_name = format!("{}.temporary", Uuid::new_v4().to_string());
+        let tmp_path = self.join_path(&self.root, &tmp_file_name);
 
+        let mut f = fs::File::create(&tmp_path).await?;
         f.write(obj_bytes).await?;
 
+        super::rename::rename(&tmp_path, path)?;
```
houqp, line 72: Since this rename is specific to fs backend, it would be better to namespace it within the file module so it's self contained. To do that, we can change `file.rs` into the following structure:

```
src/storage/file/mod.rs
src/storage/file/rename.rs
```

then from within `file/mod.rs`, we can import the rename code using `crate::storage::file::rename`.

rust/src/storage/file.rs
houqp: looks good 👍

rust/src/storage/rename.rs
mosyp: current design creates tmp file on each try in atomic loop, so if there's concurrent workers it is expected that some of the attemps might now work and hence thos file will remain.

mosyp: But I feel like we can add remove_file only in EEXIST case

houqp: Right, i think it's fine to leave out temp files, which we will handle later with a gc routine implemented specifically for each storage backend. There are many ways for the writer to leave out dangling temp files, for example, any crash or pause between temp file being created to rename syscall being completed. Moving the temp file remove logic into rename syscall wrapper won't reduce the chance of that happening.

I also think the old behavior you had is more correct :) The file removal should happen on all error cases as long as the writer is not crashed to reduce the dangling temp files we are leaving in the table folder. The writer should still try to be a good citizen when things are in full control :)

Essentially we are implementing a cross platform rename syscall warper for UNIX systems here, so keeping the semantic the same would lead to less surprise to the users. In general, clean up should be handled by the creator of the file because it needs to have file clean code anyway for other failure cases, for example: temp file was created successfully, but file write failed in the middle due to out of disk space.

houqp:
The rest looks good to me 👍  just need to fix that build issue :D

rust/src/storage/file.rs
```diff
@@ -58,26 +63,14 @@ impl StorageBackend for FileStorageBackend {
...
-            .open(path)
-            .await
-            .map_err(|e| match e.kind() {
-                std::io::ErrorKind::AlreadyExists => StorageError::AlreadyExists(path.to_string()),
-                _ => StorageError::Io { source: e },
-            })?;
+        let tmp_file_name = format!("{}.temporary", Uuid::new_v4().to_string());
+        let tmp_path = self.join_path(&self.root, &tmp_file_name);
 
+        let mut f = fs::File::create(&tmp_path).await?;
```
houqp, line 69: So I found this bug wile verifying a draft version of our S3 write design using TLA+. Although the chance of getting a UUID conflict is very very low, when it happens, we will corrupt the transaction log. We need to make sure the temporary files are created without overriding any existing files. This means we should reuse the original `OpenOptions::new().create_new(true)` logic to create the tmp file to guard against this unlikely event.

In the S3 write implementation, we will need to leverage dynamodb to guarantee uniqueness of temporary S3 keys as well.

rust/src/storage/rename.rs
mosyp: > I also think the old behavior you had is more correct :) The file removal should happen on all error cases 

yeap, but the thing is when we get EEXIST this is expected behavior but rather than an error, since we expected other workers to commit versions concurrently 

> For example: temp file was created successfully, but file write failed in the middle due to out of disk space

This I agree. As mentioned before the clean-up after EEXIST is like using java try-catch for program flow, it's discouraged  but is ok when there's no alternative. 

Anyway, I'm fine with deleting `remove` temp file lines at all, if my statement above does not convince you :)

rust/src/storage/file.rs
mosyp: yeap, makes sense, will do

mosyp: @houqp did that, but also added a loop inside put_obj when creating tmp file, because if we propagate AlreadyExists back to the caller, then the ty_commit will call this function with incremented version, which actual is not that we want

mosyp: meaning that we address uuid conflict, not version number

rust/src/storage/file/mod.rs
```diff
@@ -58,25 +65,29 @@ impl StorageBackend for FileStorageBackend {
...
+        let tmp_path = self.join_path(&self.root, &tmp_file_name);
 
-        if let Some(d) = dir.parent() {
-            fs::create_dir_all(d).await?;
+        // run this in loop in case tmp file with tmp_file_name already exists
+        loop {
+            let mut rf = fs::OpenOptions::new()
+                .create_new(true)
+                .write(true)
+                .open(&tmp_path)
```
houqp, line 76: shouldn't we retry with new temp path here?

rust/src/storage/file/mod.rs
```diff
@@ -58,25 +65,29 @@ impl StorageBackend for FileStorageBackend {
...
+                .await;
+            match rf {
+                Ok(mut f) => {
+                    f.write(obj_bytes).await?;
+                    break;
+                },
+                Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
+                    continue;
```
houqp, line 84: we should also add a max retry, i don't think infinitely looping for uuid conflicts will happen in reality, but could happen due to bugs in our code.

rust/src/storage/file.rs
houqp: 👍

rust/src/storage/rename.rs
houqp: > yeap, but the thing is when we get EEXIST this is expected behavior but rather than an error, since we expected other workers to commit versions concurrently

I agree.

> This I agree. As mentioned before the clean-up after EEXIST is like using java try-catch for program flow, it's discouraged but is ok when there's no alternative.

Yep, catch all error is usually an anti-pattern. I think it just happened that we do need to perform clean up for all the error cases here :( On the other hand, it is a common practice in Linux kernel code to perform this kind of clean up using gotos. Similarly for use of `defer` in golang.

Perhaps it's better to model it as transaction rollback. Creating temp file is one of the step in the transaction. If the transaction failed at the end for whatever error, we need to rollback the full transaction. On transaction commit or abort, all the steps should either all happen or not happen.

Reducing side effect of the function is another way we can justify this behavior. If this function returns, part of the contract we provide should be no dangling temporary file is left on the filesystem.

If leaving those temp files around in error cases provides value to the users, then I think it's worth keeping them. But I don't think it is the case here?

rust/src/storage/file/rename.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[cfg(target_os = "linux")]
+extern "C" {
+    fn renameat2(
+        olddirfd: libc::c_int,
+        oldpath: *const libc::c_char,
+        newdirfd: libc::c_int,
+        newpath: *const libc::c_char,
+        flags: libc::c_uint,
+    ) -> libc::c_int;
```
mosyp, line 15: @houqp I figured out linux error, path is required to be absolute, but in tests it is given as relative. 
One way to achive that is to call `fs::canonicalize` within `FileStorageBackend::new` constructor, wdyt?

rust/src/storage/rename.rs
mosyp: > But I don't think it is the case here?

yeah I feel like not.

Overall the transactional approach it is? e.g. to delete tmp file no matter what error

rust/src/storage/file/mod.rs
mosyp: yeap, I'm just not sure how to actually test it since we need to mock UUID call somehow to reproduce an error

rust/src/storage/file/mod.rs
mosyp: yeah :)

rust/src/storage/file/rename.rs
houqp: yep, that's a good idea :+1:

rust/src/storage/file/mod.rs
houqp: i feel like a best effort on preventing infinite loop here is good enough, we don't have to write test for it. I also don't have a good idea on how to patch the return value of uuid here without changing the storage backend struct.

rust/src/storage/rename.rs
houqp: yeah, that's what I am leaning towards, unless we come up with a good use for those dangling temp files.

rust/src/storage/file/mod.rs
```diff
@@ -58,26 +65,43 @@ impl StorageBackend for FileStorageBackend {
...
+    backend: &FileStorageBackend,
+    path: &str,
+    obj_bytes: &[u8],
+) -> Result<String, StorageError> {
+    let mut i = 0;
 
-        // use `OpenOptions` with `create_new` to create the file handle.
-        // this will force ErrorKind::AlreadyExists if the file already exists.
-        let mut f = fs::OpenOptions::new()
+    while i < 5 {
```
mosyp, line 81: Now I'm a little bit concerning regarding this magic number, how do we usually go about that?

mosyp: leave it as is or move into some static const?

xianwill: I think moving it to a const is a good idea. For optimistic concurrency retries in `DeltaTransaction`, we also provide an [options argument](https://github.com/delta-io/delta-rs/blob/715be9a497ec63b9bcd7dbdae7eba58f15e62049/rust/src/delta.rs#L732-L772), but that might be overkill in the case of the file storage backend.

xianwill:
> I've got several concern here
> 
> * If `rename` fails due to concurrent commits, it'll start over again with incremented version which means that we'll have it writes tmp file again and again. However the impact will be on higher level within `try_commit_loop` which is out of backend storage lvl.
> * Current impl does not remove tmp files once it fails, but this should be added/revised once previous concern is resolved

@mosyp - Regarding 1 you mentioned above, I'll create a separate issue for that problem. Ultimately, we need to expose a `rename_obj` on the `StorageBackend` so `DeltaTransaction` can handle it the same way across storage backends. I.e.

```
* write temp file
* optimistic commit loop:
  * atomic rename
  * if version already exists
    * do conflict resolution if update, update temp file
  * increase version
```

mosyp:
@xianwill yes, that's a good idea to move it out of this scope. I feel like once we're done with s3 atomic rename, then we're safe to create some abstraction which will work for both fs and s3

xianwill:
Posted https://github.com/delta-io/delta-rs/issues/135

houqp (approved):
great work @mosyp !

houqp:
@xianwill want to take a final look?

xianwill:
LGTM! Great work @mosyp!

PR #132, opened 2021-03-18
Support EKS(K8s) service account authentication for S3 backend

ryanlyn:
# Description
The standard `ChainProvider` used in `rusoto` does not support credentials from a web token file by default. This authentication method is only relevant in the context of the `S3StorageBackend` and when running the client libraries in pods on EKS (that themselves are attached to Service Accounts with relevant IAM roles). 

I've tested this change on an actual cluster and it works. 

# Documentation
* `WebIdentityProvider`: https://docs.rs/rusoto_sts/0.45.0/rusoto_sts/struct.WebIdentityProvider.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -610,7 +610,9 @@ dependencies = [
  "regex",
  "reqwest",
  "rusoto_core",
+ "rusoto_credential",
  "rusoto_s3",
+ "rusoto_sts",
  "serde",
  "serde_json",
  "serial_test",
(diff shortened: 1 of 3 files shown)


ryanlyn:
cc  @rtyler  @houqp , I can't seem to add any reviewers 😅 . Could I get a look at this please?

houqp (approved):

houqp:
Thanks @ryanlyn for the fix!

PR #136, opened 2021-03-18
Duplicate files after update

### LINKED ISSUE #127: Duplicate files after update
Hi,

After playing with the `update()` function for #124 I noticed that the number of files in one of my test runs was off.

I was able to reproduce the behavior in my [branch](https://github.com/delta-io/delta-rs/compare/main...dispanser:duplicate-files-after-update). The test loads version 0 of a data set, and then updates to the newest version. The expected outcome is that `get_files()` produces the same set of files as when directly loading the newest version. However, the file that was already in version 0 appears twice in the list.

`update()` is the only call site to `restore_checkpoint` that does not start empty, and I believe it should start from an empty state when loading a checkpoint, as checkpoints are typically self-contained. Instead, it applies the actions read from the checkpoint to the previously stored state.

I may be misunderstanding how checkpoints are supposed to work, but the provided dataset has been produced by a very simple spark function (delta 0.8.0), and I guess that defines a compatibility baseline.

dispanser:
closes #127 

Attempting to follow the advice given by @houqp in #127 , created a separate `DeltaTableState` struct that has a (derived) `Default` instance, keeping all the things that need a reset on restore from checkpoint in there.

I'm not confident about which of the `DeltaTable` fields should move over to `DeltaTableState` -- typically this is an indication of an immature abstraction / separation.

This is the first time working with rust for me, so opening a daft PR and expect major changes :).


rust/Cargo.toml
```diff
@@ -24,6 +24,7 @@ regex = "1"
 chrono = "0"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 lazy_static = "1"
+pretty_assertions = "0"
```
houqp, line 27: should move this to dev-dependencies section if it's only used in tests

rust/src/delta.rs
```diff
@@ -178,22 +178,28 @@ impl From<StorageError> for LoadCheckpointError {
...
+
     // metadata
     // application_transactions
     storage: Box<dyn StorageBackend>,
 
-    files: Vec<String>,
-    app_transaction_version: HashMap<String, DeltaDataTypeVersion>,
-    commit_infos: Vec<Value>,
     current_metadata: Option<DeltaTableMetaData>,
     last_check_point: Option<CheckPoint>,
```
houqp, line 204: i think field being mutated in `process_action` method should go into the state struct. then we can probably just pass in a mut reference of state into `process_action` instead of self.

houqp (commented):
looking good overall, thanks for the refactor!

rust/Cargo.toml
dispanser: done, thanks

rust/src/delta.rs
dispanser: done. Looks a bit clearer now, separating the meta data (path, backend, ...) from the content.

Had to introduce getters for `min_reader_version` and `min_writer_version`, which looks a bit odd.

Anything else? Squashing?

houqp (approved):

rust/src/delta.rs
houqp: For future reference, you can leave the incremental commits in the PR if you prefer, i will just do a squash merge at the end.

houqp:
Thanks @dispanser for your first PR!

PR #137, opened 2021-03-19
add join_paths for more efficient path join with multiple components

houqp:
# Description
Add join_paths for more efficient path join with multiple components.

Also added document for StorageBackend trait.

### DIFF
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -1,4 +1,4 @@
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::pin::Pin;
 
 use chrono::DateTime;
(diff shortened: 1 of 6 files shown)


rust/src/storage/file/mod.rs
```diff
@@ -36,6 +36,13 @@ impl StorageBackend for FileStorageBackend {
             .unwrap()
     }
 
+    fn join_paths(&self, paths: &[&str]) -> String {
+        let mut iter = paths.iter();
+        let mut path = Path::new(iter.next().unwrap_or(&"")).to_path_buf();
```
fvaleye, line 41: Maybe it is just me in the morning, but it took me a bit of time to understand this implementation.

What do you think about using PathBuf declaration directly (with from or a mut variable)?
`let mut path = PathBuf::new();`
`let path = PathBuf::from(string);`

houqp: good suggestion 👍

fvaleye (approved):

PR #138, opened 2021-03-19
Fix Delta <> Arrow schema timestamp

nevi-me:
# Description

This fixes an incorrect schema mapping for `timestamp` from delta to arrow

# Documentation

The protocol talks about timestamp being a microsecond precision value with no time zone. The Arrow `Time64` only stores the time portion of a timestamp, so the mapping was incorrect.

### DIFF
--- a/rust/src/delta_arrow.rs
+++ b/rust/src/delta_arrow.rs
@@ -95,7 +95,7 @@ impl TryFrom<&schema::SchemaDataType> for ArrowDataType {
                     }
                     "timestamp" => {
                         // Microsecond precision timestamp without a timezone.
-                        Ok(ArrowDataType::Time64(TimeUnit::Microsecond))
+                        Ok(ArrowDataType::Timestamp(TimeUnit::Microsecond, None))
                     }
                     s => Err(ArrowError::SchemaError(format!(
                         "Invalid data type for Arrow: {}",


fvaleye (approved):

nevi-me:
@fvaleye I have a suspicion that the delta protocol is incorrect on the representation of dates, but I need to check with some test data first. The delta `date` should be mapped to `arrow::DataType::Date32`, as Spark uses similar.

I'll address that separately though

fvaleye:
> @fvaleye I have a suspicion that the delta protocol is incorrect on the representation of dates, but I need to check with some test data first. The delta `date` should be mapped to `arrow::DataType::Date32`, as Spark uses similar.
> 
> I'll address that separately though

Ok thank you for letting me know!

I think that Delta uses DateType:
> DateType: Represents values comprising values of fields year, month and day, without a time-zone.

It worth testing it and have another PR on this.

nevi-me:
> DateType: Represents values comprising values of fields year, month and day, without a time-zone.

That's the odd thing for me, because Spark says that the date type is stored as days since/before epoch. https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/types/DateType.scala#L36

Even with the timestamp type, I haven't seen the y,m,d,h,m,s format anywhere.

Issue #135, opened 2021-03-18
Implement rename_obj (atomic rename) API in storage backend


xianwill:
As discussed in https://github.com/delta-io/delta-rs/discussions/89, for multi-writer support, the S3 storage backend requires an atomic rename operation that uses a distributed lock lease to ensure that log files are not overwritten by other writers. 

Ultimately, the high-level logic of a `DeltaTransaction` commit should abstract over storage backends and look like:

```
* write temp file
* optimistic commit loop:
  * atomic rename
  * if version already exists
    * do conflict resolution if update, update temp file
  * increase version
```

Usage of the distributed lock should use a two-phase-commit and a compare-and-swap (i.e. conditional write) approach to guard the file write. The two-phase-commit is necessary so other writers can detect a stale transaction, determine state and continue or repair the previous transaction state. The compare-and-swap is required so a writer that has acquired the lock can fail in case it paused without knowing and exceeded the lock lease timeout. 

Further discussion about the implementation (including a proposal for a DynamoDb schema that represents a lock) should continue in  https://github.com/delta-io/delta-rs/discussions/89 before this issue is taken on.

houqp:
As a side note for implementation, it looks like we need two types of consistency guarantees where:

1. for creating the temp object, we need put if not exists
2. for renaming the temp object, we need rename if not exists

PR #140, opened 2021-03-20
fix windows build

houqp:
# Description

Fix windows build, which was blocking us from releasing a windows
python client. Added mac and window build to the build matrix.

# Related Issue(s)

Also move rename temp file clean up logic into put_obj method in
preparation for moving the temp file creation logic into optimistic
delta commit loop. See https://github.com/delta-io/delta-rs/issues/135.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -8,6 +8,7 @@ use std::io::{BufRead, BufReader, Cursor};
 use arrow::error::ArrowError;
 use chrono::{DateTime, FixedOffset, Utc};
 use futures::StreamExt;
+use lazy_static::lazy_static;
 use log::debug;
 use parquet::errors::ParquetError;
 use parquet::file::{
@@ -290,23 +291,22 @@ impl DeltaTable {
         &self,
         version: DeltaDataTypeVersion,
     ) -> Result<Option<CheckPoint>, DeltaTableError> {
-        let mut cp: Option<CheckPoint> = None;
-        let root = self.storage.join_path(r"^*", "_delta_log");
-        let regex_checkpoint = self
-            .storage
-            .join_path(&root, r"(\d{20})\.checkpoint\.parquet$");
-        let regex_checkpoint_parts = self
-            .storage
-            .join_path(&root, r"(\d{20})\.checkpoint\.\d{10}\.(\d{10})\.parquet$");
-        let re_checkpoint = Regex::new(&regex_checkpoint).unwrap();
-        let re_checkpoint_parts = Regex::new(&regex_checkpoint_parts).unwrap();
+        lazy_static! {
+            static ref CHECKPOINT_REGEX: Regex =
+                Regex::new(r#"^*[/\\]_delta_log[/\\](\d{20})\.checkpoint\.parquet$"#).unwrap();
+            static ref CHECKPOINT_PARTS_REGEX: Regex = Regex::new(
+                r#"^*[/\\]_delta_log[/\\](\d{20})\.checkpoint\.\d{10}\.(\d{10})\.parquet$"#
+            )
+            .unwrap();
+        }
 
+        let mut cp: Option<CheckPoint> = None;
         let mut stream = self.storage.list_objs(&self.log_path).await?;
 
         while let Some(obj_meta) = stream.next().await {
             // Exit early if any objects can't be listed.
             let obj_meta = obj_meta?;
-            if let Some(captures) = re_checkpoint.captures(&obj_meta.path) {
+            if let Some(captures) = CHECKPOINT_REGEX.captures(&obj_meta.path) {
                 let curr_ver_str = captures.get(1).unwrap().as_str();
                 let curr_ver: DeltaDataTypeVersion = curr_ver_str.parse().unwrap();
                 if curr_ver > version {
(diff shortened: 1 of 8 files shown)


houqp:
cc @mosyp and @xianwill since this is related to #135.

rust/src/storage/file/mod.rs
```diff
@@ -74,7 +76,12 @@ impl StorageBackend for FileStorageBackend {
     async fn put_obj(&self, path: &str, obj_bytes: &[u8]) -> Result<(), StorageError> {
         let tmp_path = create_tmp_file_with_retry(self, path, obj_bytes).await?;
 
-        rename::rename(&tmp_path, path)
+        if let Err(e) = rename::atomic_rename(&tmp_path, path) {
+            fs::remove_file(tmp_path).await.unwrap_or(());
```
houqp, line 80: switched to async rename

rust/src/delta.rs
```diff
@@ -290,23 +291,22 @@ impl DeltaTable {
...
-        let re_checkpoint = Regex::new(&regex_checkpoint).unwrap();
-        let re_checkpoint_parts = Regex::new(&regex_checkpoint_parts).unwrap();
+        lazy_static! {
+            static ref CHECKPOINT_REGEX: Regex =
+                Regex::new(r#"^*[/\\]_delta_log[/\\](\d{20})\.checkpoint\.parquet$"#).unwrap();
+            static ref CHECKPOINT_PARTS_REGEX: Regex = Regex::new(
+                r#"^*[/\\]_delta_log[/\\](\d{20})\.checkpoint\.\d{10}\.(\d{10})\.parquet$"#
+            )
+            .unwrap();
+        }
```
houqp, line 301: turns out we can't use `join_path` to build regex query in a cross-platform way because the separator `/` from windows is a regex escape character.

rust/src/storage/file/rename.rs
```diff
@@ -1,61 +1,81 @@
...
 
-        return Err(StorageError::Io {
-            source: custom_io_error(format!("{}", e)),
-        });
+#[cfg(unix)]
+mod imp {
+    use super::*;
+    use std::ffi::CString;
+
+    #[cfg(target_os = "linux")]
```
rtyler, line 23: Linux specific code in the unix-specific implementation? I'm quite confused by why these libc functions need to be mapped in? There's a high likelihood this could break on BSDs, non-glibc implementations.

(I haven't tested it of course, so you may know something I don't)

Isn't there a crate that abstracts this behavior which doesn't require the `unsafe` below?  E.g. https://docs.rs/crate/atomicwrites/0.3.0


Of course, if you've already gone down this path and this is the best possible option, then so be it, I would appreciate some commentary in the code explaining why "this must be"

rust/src/storage/file/rename.rs
```diff
@@ -1,61 +1,81 @@
...
+        Ok(())
     }
-}
 
-fn to_c_string(p: &str) -> Result<CString, StorageError> {
-    CString::new(p).map_err(|e| StorageError::Generic(format!("{}", e)))
+    unsafe fn platform_specific_rename(from: *const libc::c_char, to: *const libc::c_char) -> i32 {
+        cfg_if::cfg_if! {
+            if #[cfg(target_os = "linux")] {
+                renameat2(libc::AT_FDCWD, from, libc::AT_FDCWD, to, RENAME_NOREPLACE)
```
rtyler, line 66: Reading the manpage for this function this little bit jumped out to me

> RENAME_NOREPLACE **requires support from the underlying filesystem.**

Emphasis mine of course. I don't have a good alternative suggestion, but just would like to express my worry that there couplings to underlying system implementation details that we cannot clearly communicate to users here.

rust/src/storage/file/rename.rs
houqp: platform specific atomic rename needs to be implemented per platform unfortunately, so BSD and non-glibc users will just need to send PRs for the platforms they want to use, just like windows.

I looked into atomicwrites before, their implementation is not ideal. It requires 3 system calls to achieve what can be done in 1 system call, which is the implementation we have here. I also don't think their implementation is correct, because their implementation started with a hardlink function call, which based on the std rust doc, it doesn't error out if destination path already exists.

The unsafe code we have here is scoped to only system call invocation, which is basically what happens when you call `std::fs::rename`. So even though atomicwrites has less unsafe code on the surface, it's just hidden under the std lib implementation.

rust/src/storage/file/rename.rs
houqp: good call, added rust doc for filesystem back to clarify that.

rtyler (approved):
Looks good to me, I think @fvaleye or @mosyp should take a look before we merge

rust/src/storage/file/rename.rs
mosyp: @rtyler 
> why these libc functions need to be mapped in

Because they're missing in libc crate and PR for introducing them is still in open

mosyp: well https://github.com/rust-lang/libc/pull/2116#issuecomment-801511926 seems like they're merged this a couple of days ago

mosyp (approved):
lgtm!

rust/src/storage/file/rename.rs
houqp: ha nice, filed https://github.com/delta-io/delta-rs/issues/143 as follow up

PR #147, opened 2021-03-22
Feature/partitions filter

fvaleye:
# Description
Add the functionality of filtering partitions while reading a partitioned DeltaTable.
The allowed operations are: `=, !=, in, not in`.
The supported type for the value is `String` but we could provide other types for enabling other operators `>, <, >=, <=` for different types.

The partition filters is a ` List` and the filters are evaluated one-by-one with an `AND` operator.
The partition filter could be a `Tuple(key, operation, value)` for operator `=` and `!=` operators
The partition filter could be a `Tuple(key, operation, List<values>)` for `in` and `not in` operators

The Python bindings are also added.

# Documentation
Inspired by https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html


houqp:
this looks neat 👍  I will take a closer look later today.

rust/src/partitions.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct DeltaTablePartition<'a> {
+    pub key: &'a str,
+    pub value: &'a str,
+}
+
+impl<'a> DeltaTablePartition<'a> {
+    pub fn filter_accept_partition(&self, filter: &PartitionFilter<&'a str>) -> bool {
```
houqp, line 66: nitpick, how about naming it as `match_filter`? The `partition` suffix here sounds redundant.

rust/src/partitions.rs
```diff
@@ -0,0 +1,100 @@
...
+        if filter.partition_key != self.key {
+            return true;
```
houqp, line 68: shouldn't we return false if partition key doesn't match?

python/src/lib.rs
```diff
@@ -63,8 +70,36 @@ impl RawDeltaTable {
             .map_err(PyDeltaTableError::from_raw)?)
     }
 
-    pub fn files(&self) -> PyResult<Vec<String>> {
-        Ok(self._table.get_files().to_vec())
+    pub fn files(
```
houqp, line 73: i recommend creating a separate method for this like you did with the rust struct for better readability and to avoid runtime option matching.

rust/src/delta.rs
```diff
@@ -561,6 +569,34 @@ impl DeltaTable {
...
 
+    pub fn get_files_by_partitions(
+        &self,
+        filters: Vec<PartitionFilter<&str>>,
+    ) -> Result<Vec<String>, DeltaTableError> {
+        let partitions_number = match &self.state.current_metadata {
+            Some(metadata) if !metadata.partition_columns.is_empty() => {
+                metadata.partition_columns.len()
+            }
+            _ => return Err(DeltaTableError::LoadPartitions),
```
houqp, line 580: the error message thrown by this error is not accurate, because this match branch could be reached when table struct is created, but having loaded any transaction log yet.

perhaps a better way to deal with this is to invoke `self.get_metadata()?` .

rust/src/delta.rs
```diff
@@ -561,6 +569,34 @@ impl DeltaTable {
...
+        };
+        let files = self
+            .state
+            .files
+            .iter()
+            .filter(|f| {
+                f.splitn(partitions_number + 1, self.storage.get_separator())
```
houqp, line 587: nitpick, should call `self.storage.get_separator()` outside of the loop once.

rust/src/partitions.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl<'a> TryFrom<&'a str> for DeltaTablePartition<'a> {
+    type Error = DeltaTableError;
+
+    fn try_from(partition: &'a str) -> Result<Self, DeltaTableError> {
+        match partition.find('=') {
+            Some(position) => Ok(DeltaTablePartition {
+                key: &partition[0..position],
+                value: &partition[position + 1..],
```
houqp, line 93: what happens if `=` is the last char?

rust/tests/read_delta_partitions_test.rs
```diff
@@ -0,0 +1,102 @@
...
+        partition_key: month_key,
+        partition_value: deltalake::PartitionValue::Equal(month_12),
+    };
+    let partition_year_22_filter = deltalake::PartitionFilter {
+        partition_key: year_key,
+        partition_value: deltalake::PartitionValue::Equal(year_2022),
+    };
+
+    assert_eq!(
+        partition_2021.filter_accept_partition(&partition_year_2021_filter),
```
houqp, line 91: method being tested here doesn't match the test name `test_filters_accept_partition`

python/src/lib.rs
fvaleye: 👍

rust/src/delta.rs
```diff
@@ -561,6 +569,34 @@ impl DeltaTable {
         }
     }
 
+    pub fn get_files_by_partitions(
+        &self,
+        filters: Vec<PartitionFilter<&str>>,
```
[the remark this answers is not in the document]
fvaleye: 👍

rust/src/delta.rs
fvaleye: I agree! I will separate the error handling into two separate calls:

1. No metadata error 
2. No partition column found (feel free to tell me if the error message is relevant)

rust/src/delta.rs
fvaleye: Yep right!

I changed this by using the ` /` separator because it works in all cases except one.
The only case that could raise an error is a  `Delta Table` partitioned **created** on a Windows machine but I don't know how to test/detect it. A `DeltaTable` could be created in S3 and read on a local Windows machine...

rust/src/partitions.rs
```diff
@@ -0,0 +1,100 @@
...
+    Equal(T),
+    NotEqual(T),
+    In(Vec<T>),
+    NotIn(Vec<T>),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PartitionFilter<'a, T> {
+    pub partition_key: &'a str,
+    pub partition_value: PartitionValue<T>,
```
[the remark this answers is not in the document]
fvaleye: 👍

rust/src/partitions.rs
fvaleye: 👍

rust/src/partitions.rs
fvaleye: The filters are applied with a `AND` operator on all the partition keys. So we could apply a filter with a key on a different partition key. The only case where we might be wrong is to have a filter key that doesn't exist inside the partitions keys. I will check this case before applying the filters.

> Example:
> partition_columns: "year=2020" and "month="12"
> filter: ("year"="2020") is true on the partition column "year" and also true by default with "month".
> the `all()` method is valid.

If we had:

> partition_columns: "year=2020" and "month=12"
> filter: ("year"="2020") is true on the partition column "year" and also false by default with "month".
> the `all()` method would be invalid.

rust/src/partitions.rs
fvaleye: A `DeltaTable` uses the Hive Partitioning (https://cwiki.apache.org/confluence/display/Hive/Tutorial).
I would say it is impossible... but we could add a unit test!

rust/tests/read_delta_partitions_test.rs
```diff
@@ -0,0 +1,102 @@
...
+        value: "2020",
+    };
+    let partition_2019 = deltalake::DeltaTablePartition {
+        key: "year",
+        value: "2019",
+    };
+    let year_key = "year";
+    let month_key = "month";
+    let year_2020 = "2020";
+    let month_12 = "12";
```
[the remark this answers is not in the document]
fvaleye: 👍

rust/tests/read_delta_partitions_test.rs
fvaleye: 👍

rust/src/partitions.rs
houqp: yeah, i am more concerned about error cases like typos in partition names. it looks like a more robust way to evaluate the partition predicates would be to evaluate each single filter on the full partition path, instead of each single partition component. i.e. we would need to model partition as a vector of key value pairs, then check if a filter matches any one of the key pair. And finally, we can have the evaluation logic written as:

```
for filter in all filters:
    any(filter.match(kv) for kv in partition_components )
```

rust/src/partitions.rs
houqp: yep, i don't expect this to error out in the happy path. Just want to make sure we are not resulting a hard runtime cash when processing invalid input :) It would be better to throw out an Error to caller in this case.

houqp (commented):
the rest looks good to me 👍  very useful feature to speed up the query.

rust/src/partitions.rs
fvaleye: Totally agree with you! Let me add an unit-test on it!

rust/src/partitions.rs
fvaleye: Ok, I will see how I can refactor this!

rust/src/partitions.rs
```diff
@@ -85,12 +85,16 @@ impl<'a> TryFrom<&'a str> for DeltaTablePartition<'a> {
     type Error = DeltaTableError;
 
     fn try_from(partition: &'a str) -> Result<Self, DeltaTableError> {
-        match partition.find('=') {
-            Some(position) => Ok(DeltaTablePartition {
+        let separator = "=";
+        match (
+            partition.find(separator),
+            partition.matches(separator).count(),
```
houqp, line 91: nitpick, the [split](https://doc.rust-lang.org/std/primitive.str.html#method.split) method will probably work better for this use-case so we won't need to search the string twice for the same separator.

fvaleye: Oh! But do we need to check if there is only one `=` in `partition`? Or just ignoring it by using `split` method (and get the wrong value)?

houqp: depends on your use-case, if we only want to split on the first `=`, then we can use [splitn](https://doc.rust-lang.org/std/primitive.str.html#method.splitn).

fvaleye: Oh got it!

python/deltalake/table.py
```diff
@@ -18,6 +18,27 @@ def version(self) -> int:
...
+    def files_by_partitions(self, partition_filters: List[Tuple]) -> List[str]:
+        """
+        Partitions which do not match the filter predicate will be removed from scanned data.
+        Predicates are expressed in disjunctive normal form (DNF), like [("x", "=", "a"), ...].
+        DNF allows arbitrary boolean logical combinations of single partition predicates.
+        The innermost tuples each describe a single partition predicate.
+        The list of inner predicates is interpreted as a conjunction (AND), forming a more selective and multiple
+        partition predicates.
+        Each tuple has format: (key, op, value) and compares the key with the value.
+        The supported op are: =, !=, in and not in.
```
rtyler, line 30:
```suggestion
        The supported op are: `=`, `!=`, `in`, and `not in`.
```

I found this comment a bit confusing since `and` is frequently used in predicate expressions, I think I've quoted the right things here :laughing:

python/tests/test_table_read.py
```diff
@@ -17,6 +17,50 @@ def test_read_simple_table_by_version_to_dict():
...
+        "year=2021/month=12/day=4/part-00000-6dc763c0-3e8b-4d52-b19e-1f92af3fbb25.c000.snappy.parquet",
+        "year=2021/month=4/day=5/part-00000-c5856301-3439-4032-a6fc-22b7bc92bebb.c000.snappy.parquet",
+    ]
+    partition_filters = [("day", "not in", ["3", "20"]), ("year", "=", "2021")]
+    assert dt.files_by_partitions(partition_filters=partition_filters) == [
+        "year=2021/month=12/day=4/part-00000-6dc763c0-3e8b-4d52-b19e-1f92af3fbb25.c000.snappy.parquet",
+        "year=2021/month=4/day=5/part-00000-c5856301-3439-4032-a6fc-22b7bc92bebb.c000.snappy.parquet",
+    ]
+    partition_filters = [("invalid_operation", "=>", "3")]
+    with pytest.raises(Exception):
```
rtyler, line 53: I could just be misunderstanding some of the Python code here, but would it be possible for the binding to raise a typed exception that indicates that the filter contains an invalid operand?

If this is not how other underlying errors from the Rust layer are mapped into the Python binding, please ignore my noise :smile:

rtyler (commented):
I'm looking forward to this merging, great work @fvaleye

python/tests/test_table_read.py
fvaleye: We need to provide more information on the exceptions raised by Rust in the Python tests 👍

python/deltalake/table.py
fvaleye: 👍

rust/src/partitions.rs
fvaleye: Hum, I refactored this part based on your suggestion.

But it changes the way we apply the filters by using an `OR` operator instead of `AND`.
`(filter_1), (filter_2)` we keep partitions that matched the `filter_1` and `filter_2` by iterating over the filters.

It may not be a problem but I found easier to apply and `AND` operator with a list of filters and we may use a list of list for the OR operator. WDYT?

fvaleye: > This means a partition path needs to match all the filters.
My implementation missed that part and was "a partition path needs to match one of the filters".

It's ready!

rust/src/partitions.rs
```diff
@@ -0,0 +1,101 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PartitionValue<T> {
+    Equal(T),
+    NotEqual(T),
+    In(Vec<T>),
+    NotIn(Vec<T>),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PartitionFilter<'a, T> {
```
houqp, line 14: typically in query engine implementations, predicates/filters are modeled as structs with 3 fields `key, op, value` or `left, op, right`. we are merging `op` with `value/right` here :)

but i think this is something minor we can refactor later, no need to block the merge.

houqp (approved):
Great work!

rust/src/partitions.rs
fvaleye: Interesting, let me know if I need to change it!

houqp: I don't think we need to refactor this in this PR since I think it's something minor and I have already approved it : ) But it is definitely a good change to introduce in future follow ups.

houqp:
@fvaleye clippy fixed merged into master, looks like it's rebase time :)

fvaleye:
Hum, same error with the Window build... 😞 it might be related to fetching the git history.

houqp:
@rtyler you want to take a final look?

houqp:
@fvaleye now that the PR has been merged, do you want to try releasing a new pypi package? Here are the steps:

* bump python create version in cargo.toml
* create a new github release at https://github.com/delta-io/delta-rs/releases

This should trigger the automated pypi release github action pipeline.

Issue #146, opened 2021-03-22
Run python binding tests in manylinux envirionment


houqp:
We should run tests in the environment that our abi3 wheels targets.

PR #148, opened 2021-03-23
Fix python releases

### LINKED ISSUE #145: Deltalake 0.4.3 (and 0.4.2) breaks our CI
Hi Again :)

We tried to upgrade our base code with deltalake 0.4.3 (also with 0.4.2) but we got an error during our CI process (when launching unit tests).

The error is 
`dags/utils/delta_lake_utils.py:4: in <module>
    from deltalake import DeltaTable, Schema
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/deltalake/__init__.py:1: in <module>
    from .deltalake import RawDeltaTable, rust_core_version
E   ImportError: /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/deltalake/deltalake.abi3.so: undefined symbol: renameat2
 `

Everything is working well with deltalake 0.4.1 do you have any idea about this issue ?

houqp:
# Description

Avoid linking to renameat2 glibc symbol for old glibc versions. This should unblock python release. Delta write is not supported in python at the moment, so this should have no impact to python users.

Will add workarounds for old glibc and linux kernel in a separate follow up PR.

# Related Issue(s)

close #145 and #146.

### DIFF
--- a/rust/build.rs
+++ b/rust/build.rs
@@ -1,0 +1,56 @@
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+mod platform_cfg {
+    // glibc version is taken from std/sys/unix/os.rs
+    pub fn glibc_version() -> (usize, usize) {
+        use regex::Regex;
+        use std::process::Command;
+
+        let output = Command::new("ldd")
+            .args(&["--version"])
+            .output()
+            .expect("failed to execute ldd");
+        let output_str = std::str::from_utf8(&output.stdout).unwrap();
+
+        let version_reg = Regex::new(r#"ldd \(.+\) ([0-9]+\.[0-9]+)"#).unwrap();
+        if let Some(captures) = version_reg.captures(output_str) {
+            let version_str = captures.get(1).unwrap().as_str();
+            parse_glibc_version(version_str).unwrap()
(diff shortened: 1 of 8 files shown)


houqp:
also cc @mosyp @xianwill

rust/build.rs
```diff
@@ -0,0 +1,56 @@
...
+        use regex::Regex;
+        use std::process::Command;
+
+        let output = Command::new("ldd")
+            .args(&["--version"])
+            .output()
+            .expect("failed to execute ldd");
+        let output_str = std::str::from_utf8(&output.stdout).unwrap();
+
+        let version_reg = Regex::new(r#"ldd \(.+\) ([0-9]+\.[0-9]+)"#).unwrap();
```
fvaleye, line 14: Maybe you can add a unit-test for extracting the regex with an example of the glibc version?

houqp: build.rs doesn't support unit test, let me pull those code into its own crate.

houqp: @fvaleye moved that to a different crate and added tests.

fvaleye (approved):
Perfect 👍

PR #151, opened 2021-03-24
docs: update readme badges

houqp:
# Description

~~Ever wonder why our CI badge is always green?~~

EDIT: never mind, looks we were covered by github repo redirect :D


rtyler (approved):
I thought we were just that good

fvaleye (approved):
delta-rs never failed!

PR #153, opened 2021-03-24
Add dynamodb lock implementation

mosyp:
# Description
This PR adds dynamo distributive lock ported from https://github.com/awslabs/amazon-dynamodb-lock-client

Note that not every feature is ported from java lock client to easy the integration. 

In addition, this port also introduces additional feature, it is the extending a timeout duration on each time exiting lock changes its rvn. Meaning that there's other clients trying to acquire a lock and they were more successful, so instead retuning  None (due to timeout as in original java version) this implementation extends the wait duration by adding another `lease_duration` to the try acquire loop.

### DIFF
--- a/build/setup_localstack.sh
+++ b/build/setup_localstack.sh
@@ -23,11 +23,9 @@ function wait_for() {
 wait_for "S3" "aws s3api list-buckets --endpoint-url=$ENDPOINT"
 
 echo "Uploading S3 test delta tables..."
-aws s3 rm s3://deltars --recursive --endpoint-url=$ENDPOINT > /dev/null 2>&1
-aws s3api delete-bucket --bucket deltars --endpoint-url=$ENDPOINT
-aws s3api create-bucket --bucket deltars --endpoint-url=$ENDPOINT
-aws s3 sync /data/golden s3://deltars/golden/ --endpoint-url=$ENDPOINT > /dev/null
-aws s3 sync /data/simple s3://deltars/simple/ --endpoint-url=$ENDPOINT > /dev/null
+aws s3api create-bucket --bucket deltars --endpoint-url=$ENDPOINT > /dev/null 2>&1
+aws s3 sync /data/golden s3://deltars/golden/ --delete --endpoint-url=$ENDPOINT > /dev/null
+aws s3 sync /data/simple s3://deltars/simple/ --delete --endpoint-url=$ENDPOINT > /dev/null
 
 wait_for "DynamoDB" "aws dynamodb list-tables --endpoint-url=$ENDPOINT"
 
--- a/build/setup_localstack.sh
+++ b/build/setup_localstack.sh
@@ -5,22 +5,40 @@ export AWS_ACCESS_KEY_ID=test
 export AWS_SECRET_ACCESS_KEY=test
 export ENDPOINT=http://localstack:4566
 
-retries=10 #seconds
-until aws s3api list-buckets --endpoint-url=$ENDPOINT > /dev/null 2>&1
-do
-  if [ "$retries" -lt "0" ]; then
-    echo 'S3 is still offline after 10 retries';
-    exit 1;
-  fi
-  echo 'waiting on S3 to start...'
-  sleep 5
-  retries=$((retries - 1))
-done
 
-echo 'S3 is running'
+function wait_for() {
+  retries=10
+  until eval $2 > /dev/null 2>&1
+  do
+    if [ "$retries" -lt "0" ]; then
+      echo "$1 is still offline after 10 retries";
+      exit 1;
+    fi
+    echo "Waiting on $1 to start..."
+    sleep 5
+    retries=$((retries - 1))
+  done
+}
 
+wait_for "S3" "aws s3api list-buckets --endpoint-url=$ENDPOINT"
+
+echo "Uploading S3 test delta tables..."
+aws s3 rm s3://deltars --recursive --endpoint-url=$ENDPOINT > /dev/null 2>&1
+aws s3api delete-bucket --bucket deltars --endpoint-url=$ENDPOINT
 aws s3api create-bucket --bucket deltars --endpoint-url=$ENDPOINT
-aws s3 sync /data/golden s3://deltars/golden/ --endpoint-url=$ENDPOINT
-aws s3 sync /data/simple s3://deltars/simple/ --endpoint-url=$ENDPOINT
+aws s3 sync /data/golden s3://deltars/golden/ --endpoint-url=$ENDPOINT > /dev/null
+aws s3 sync /data/simple s3://deltars/simple/ --endpoint-url=$ENDPOINT > /dev/null
 
-echo localstack is configured!
+wait_for "DynamoDB" "aws dynamodb list-tables --endpoint-url=$ENDPOINT"
+
+echo "Creating DynamoDB test lock table..."
+aws dynamodb delete-table --table-name test_table --endpoint-url=$ENDPOINT > /dev/null 2>&1
+aws dynamodb create-table --table-name test_table --endpoint-url=$ENDPOINT \
+    --attribute-definitions \
+        AttributeName=key,AttributeType=S \
+    --key-schema \
+        AttributeName=key,KeyType=HASH \
+    --provisioned-throughput \
+        ReadCapacityUnits=10,WriteCapacityUnits=10 > /dev/null
+
+echo Localstack is configured!
--- a/rust/src/storage/s3/dynamodb.rs
+++ b/rust/src/storage/s3/dynamodb.rs
@@ -129,7 +129,7 @@ impl DynamoLock {
     pub async fn acquire_lock(&self) -> Result<LockItem, DynamoError> {
         let mut state = AcquireLockState {
             client: self,
-            active_lock: None,
+            cached_lock: None,
             started: Instant::now(),
             timeout_in: Duration::from_millis(self.opts.sleep_millis),
         };
@@ -295,7 +295,7 @@ pub fn attr<T: ToString>(s: T) -> AttributeValue {
 
 struct AcquireLockState<'a> {
     client: &'a DynamoLock,
-    active_lock: Option<LockItem>,
+    cached_lock: Option<LockItem>,
     started: Instant,
     timeout_in: Duration,
 }
@@ -317,13 +317,13 @@ impl<'a> AcquireLockState<'a> {
             }
             Some(existing) => {
                 // there's existing lock and it's out first attempt to acquire it
-                if self.active_lock.is_none() {
+                if self.cached_lock.is_none() {
                     // first we store it, extend timeout period and try again later
-                    return self.set_new_active_lock_and_fail(existing);
+                    return self.set_new_cached_lock_and_fail(existing);
                 }
 
                 // there's existing lock and we've already tried to acquire it, let's try again
-                let active = self.active_lock.as_ref().unwrap();
+                let active = self.cached_lock.as_ref().unwrap();
                 let active_rvn = &active.record_version_number;
 
                 // let's check store rvn against current lock from dynamo
@@ -339,21 +339,17 @@ impl<'a> AcquireLockState<'a> {
                 } else {
                     // rvn doesn't match, meaning that other worker acquire it before us
                     // let's change active lock with new one and extend timeout period
-                    self.set_new_active_lock_and_fail(existing)
+                    self.set_new_cached_lock_and_fail(existing)
                 }
             }
         }
     }
 
-    fn set_new_active_lock_and_fail(&mut self, lock: LockItem) -> Result<LockItem, DynamoError> {
-        self.extend_timeout_by(lock.lease_duration);
-        self.active_lock = Some(lock);
+    fn set_new_cached_lock_and_fail(&mut self, lock: LockItem) -> Result<LockItem, DynamoError> {
+        self.timeout_in = Duration::from_secs(self.timeout_in.as_secs() + lock.lease_duration);
+        self.cached_lock = Some(lock);
 
         Err(DynamoError::ConditionalCheckFailed)
-    }
-
-    fn extend_timeout_by(&mut self, seconds: u64) {
-        self.timeout_in = Duration::from_secs(self.timeout_in.as_secs() + seconds)
     }
 
     async fn upsert_new_lock(&self) -> Result<LockItem, DynamoError> {
--- a/rust/src/storage/s3/dynamodb.rs
+++ b/rust/src/storage/s3/dynamodb.rs
@@ -1,0 +1,427 @@
+use std::collections::HashMap;
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+
+use rusoto_core::RusotoError;
+use rusoto_dynamodb::*;
+use uuid::Uuid;
+
+pub struct DynamoLock {
+    client: DynamoDbClient,
+    opts: Options,
+}
+
+#[derive(Clone, Debug)]
+pub struct Options {
+    pub partition_key_value: String,
+    pub table_name: String,
+    pub owner_name: String,
+    pub lease_duration: u64,
+    pub sleep_millis: u64,
+}
+
+#[derive(Clone, Debug)]
+pub struct LockItem {
+    pub owner_name: String,
+    pub record_version_number: String,
+    pub lease_duration: u64,
+    pub is_released: bool,
+    pub data: Option<String>,
+    pub lookup_time: u128,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum DynamoError {
+    #[error("Dynamo table not found")]
+    TableNotFound,
+
+    #[error("Conditional check failed")]
+    ConditionalCheckFailed,
+
+    #[error("DynamoDB item has invalid schema")]
+    InvalidItemSchema,
+
+    #[error("Could not acquire lock for {0} sec")]
+    TimedOut(u64),
+
+    #[error("Put item error: {0}")]
+    PutItemError(RusotoError<PutItemError>),
+
+    #[error("Update item error: {0}")]
+    UpdateItemError(#[from] RusotoError<UpdateItemError>),
+
+    #[error("Get item error: {0}")]
+    GetItemError(RusotoError<GetItemError>),
+
+    #[error("Delete item error: {0}")]
+    DeleteItemError(#[from] RusotoError<DeleteItemError>),
+}
+
+pub const PARTITION_KEY_NAME: &str = "key";
+pub const OWNER_NAME: &str = "ownerName";
+pub const RECORD_VERSION_NUMBER: &str = "recordVersionNumber";
+pub const IS_RELEASED: &str = "isReleased";
+pub const LEASE_DURATION: &str = "leaseDuration";
+pub const DATA: &str = "data";
+
+mod expressions {
+    pub const ACQUIRE_LOCK_THAT_DOESNT_EXIST: &str = "attribute_not_exists(#pk)";
+
+    pub const PK_EXISTS_AND_IS_RELEASED: &str = "attribute_exists(#pk) AND #ir = :ir";
+
+    pub const PK_EXISTS_AND_RVN_MATCHES: &str = "attribute_exists(#pk) AND #rvn = :rvn";
+
+    pub const PK_EXISTS_AND_OWNER_RVN_MATCHES: &str =
+        "attribute_exists(#pk) AND #rvn = :rvn AND #on = :on";
+
+    pub const UPDATE_IS_RELEASED_AND_DATA: &str = "SET #ir = :ir, #d = :d";
+
+    pub const UPDATE_IS_RELEASED: &str = "SET #ir = :ir";
+}
+
+mod vars {
+    pub const PK_PATH: &str = "#pk";
+    pub const RVN_PATH: &str = "#rvn";
+    pub const RVN_VALUE: &str = ":rvn";
+    pub const IS_RELEASED_PATH: &str = "#ir";
+    pub const IS_RELEASED_VALUE: &str = ":ir";
+    pub const OWNER_NAME_PATH: &str = "#on";
+    pub const OWNER_NAME_VALUE: &str = ":on";
+    pub const DATA_PATH: &str = "#d";
+    pub const DATA_VALUE: &str = ":d";
+}
+
+mod options {
+    pub const PARTITION_KEY_VALUE: &str = "DYNAMO_LOCK_PARTITION_KEY_VALUE";
+    pub const TABLE_NAME: &str = "DYNAMO_LOCK_TABLE_NAME";
+    pub const OWNER_NAME: &str = "DYNAMO_LOCK_OWNER_NAME";
+    pub const LEASE_DURATION: &str = "DYNAMO_LOCK_LEASE_DURATION";
+    pub const SLEEP_MILLIS: &str = "DYNAMO_LOCK_SLEEP_MILLIS";
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        fn str_env(key: &str, default: String) -> String {
+            std::env::var(key).unwrap_or(default)
+        }
+
+        fn u64_env(key: &str, default: u64) -> u64 {
+            std::env::var(key)
+                .ok()
+                .and_then(|e| e.parse::<u64>().ok())
+                .unwrap_or(default)
+        }
+
+        Self {
+            partition_key_value: str_env(options::PARTITION_KEY_VALUE, "delta-rs".to_string()),
+            table_name: str_env(options::TABLE_NAME, "delta_rs_lock_table".to_string()),
+            owner_name: str_env(options::OWNER_NAME, Uuid::new_v4().to_string()),
+            lease_duration: u64_env(options::LEASE_DURATION, 20),
+            sleep_millis: u64_env(options::SLEEP_MILLIS, 1000),
+        }
+    }
+}
+
+impl DynamoLock {
+    pub fn new(client: DynamoDbClient, opts: Options) -> Self {
+        Self { client, opts }
+    }
+
+    pub async fn acquire_lock(&self) -> Result<LockItem, DynamoError> {
+        let mut state = AcquireLockState {
+            client: self,
+            active_lock: None,
+            started: Instant::now(),
+            timeout_in: Duration::from_millis(self.opts.sleep_millis),
+        };
+
+        loop {
+            match state.try_acquire_lock().await {
+                Ok(lock) => return Ok(lock),
+                Err(DynamoError::ConditionalCheckFailed) => {
+                    if state.has_timed_out() {
+                        return Err(DynamoError::TimedOut(state.started.elapsed().as_secs()));
+                    }
+                    tokio::time::sleep(Duration::from_millis(self.opts.sleep_millis)).await;
+                }
+                Err(e) => return Err(e),
+            }
+        }
+    }
+
+    pub async fn get_lock(&self) -> Result<Option<LockItem>, DynamoError> {
+        let output = self
+            .client
+            .get_item(GetItemInput {
+                consistent_read: Some(true),
+                table_name: self.opts.table_name.clone(),
+                key: hashmap! {
+                    PARTITION_KEY_NAME.to_string() => attr(self.opts.partition_key_value.clone())
+                },
+                ..Default::default()
+            })
+            .await?;
+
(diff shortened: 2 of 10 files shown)


rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+fn attr<T: ToString>(s: T) -> AttributeValue {
+    AttributeValue {
+        s: Some(s.to_string()),
+        ..Default::default()
+    }
+}
+
+impl From<RusotoError<GetItemError>> for DynamoError {
```
xianwill, line 296: You should be able to use the `#[from]` attribute from `thiserror` to skip this implementation

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,311 @@
...
+                }
+                Some(ref l) if l.is_released => {
+                    println!("upsert_released_lock");
+                    self.upsert_released_lock(&new_lock).await
+                }
+                Some(ref existing) => match lock_trying_to_be_acquired {
+                    None => {
+                        println!("lock_trying_to_be_acquired None");
+                        lock_trying_to_be_acquired = existing_lock.clone();
+                        dur_to_wait = Duration::from_secs(5 + existing.lease_duration);
```
xianwill, line 102: The value you're setting `dur_to_wait` to here smells funny to me. From [LockItem::new](https://github.com/delta-io/delta-rs/pull/153/files#diff-b4ae09784de4beabd0a09649f76c108935196dfde86adf592a0b9bd95832eed9R261), I'm expecting lease duration to be the number of seconds to acquire the lock for. 

Would something like `now - (item.lookup_time + item.lease_duration)` be a good wait time?

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,311 @@
...
+                        std::thread::sleep(Duration::from_secs(1));
```
xianwill, line 103: I'm wondering, would something like `now - (item.lookup_time + item.lease_duration)` be a good wait time for these sleeps instead of one second?

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,311 @@
...
+    #[error("Get item error: {0}")]
+    GetItemError(RusotoError<GetItemError>),
+
+    #[error("Put item error: {0}")]
+    PutItemError(RusotoError<PutItemError>),
+}
+
+impl DynamoLock {
+    pub fn new() -> Self {
+        let client = DynamoDbClient::new(Region::default());
```
houqp, line 72: would be better to refactor and reuse the same provider construction logic used in S3 client: https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/s3.rs#L59-L74.

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,311 @@
...
+                    }
+                    Some(ref prev)
+                        if existing.record_version_number == prev.record_version_number =>
+                    {
+                        if prev.is_expired() {
+                            println!("lock_trying_to_be_acquired Some/eq/is_expired");
+                            self.upsert_expired_lock(prev, &new_lock).await
+                        } else {
+                            println!("lock_trying_to_be_acquired Some/eq/");
+                            std::thread::sleep(Duration::from_secs(1));
```
houqp, line 114: should use async sleep like https://docs.rs/tokio/1.4.0/tokio/time/fn.sleep.html in async method

rust/src/storage/s3/dynamodb.rs
mosyp: @houqp Btw, does it make sense to share the same client as well? (`rusoto_core::Client`)

houqp: probably better to keep them separate unless we get performance benefit out of that.

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,396 @@
...
+            vars::PK_PATH.to_string() => PARTITION_KEY_NAME.to_string(),
+            vars::RVN_PATH.to_string() => RECORD_VERSION_NUMBER.to_string(),
+            vars::OWNER_NAME_PATH.to_string() => OWNER_NAME.to_string(),
+            vars::IS_RELEASED_PATH.to_string() => IS_RELEASED.to_string(),
+        };
+        let mut values = hashmap! {
+            vars::IS_RELEASED_VALUE.to_string() => attr("1"),
+            vars::RVN_VALUE.to_string() => attr(&lock.record_version_number),
+            vars::OWNER_NAME_VALUE.to_string() => attr(&lock.owner_name),
+        };
```
mosyp, line 170: @houqp there's several other places where hashmap is created with k/v `k.to_string() -> y.to_string()` and `k.to_string -> attr(v)`

For the sake of clearer code, does it make sense to just use `k -> v` and then apply some mapping function `convert(hashmap!{k -> v})`;
does performance penalty (of creating second hash map) worth it?

houqp: We can get what you want with zero overhead by creating our own helper macros, for example:

```rust
macro_rules! string_hashmap {
    ($($key:expr => $value:expr,)+) => { hashmap!($($key.to_string() => $value),+) };
}

fn main() {
    let mut m = string_hashmap!{
        "a" => 1,
        "b" => 2,
    };
    m.insert("c".to_string(), 3);

    dbg!(m);
}
```

output:

```
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 2.50s
     Running `target/debug/playground`
[src/main.rs:31] m = {
    "b": 2,
    "c": 3,
    "a": 1,
}
```

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,396 @@
...
+            }
+            Some(existing) => {
+                // there's existing lock and it's out first attempt to acquire it
+                if self.active_lock.is_none() {
+                    // first we store it, extend timeout period and try again later
+                    return self.set_new_active_lock_and_fail(existing);
+                }
+
+                // there's existing lock and we've already tried to acquire it, let's try again
+                let active = self.active_lock.as_ref().unwrap();
```
houqp, line 295: Do you know what's reason for storing the current active lock as `self.active_lock` here? why not just check for expiration of current active lock every time and acquire the lock if it's expired?

mosyp: `lookup_time` is generated on `get_lock`, so each existing lock will have refreshed `lookup_time`, and yes, we won't get proper calculation of `timeout_in` without it

mosyp: but storing it as `active_lock` we'll be able to test against very first access and increase timeout to it's `lease_duration`

mosyp: however when exiting_lock rvn will be different to active_lock rvn, this will signal us that someone else acquired it before us, so we store new active_lock and increase timeout

rust/src/storage/s3/dynamodb.rs
xianwill: neat

mosyp:
@xianwill @houqp this is ready for review 
(however it contains some commits from https://github.com/delta-io/delta-rs/pull/175 to make it build)

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
```
houqp, line 326: You can probably get rid of the unwrap here by using a match assign pattern:

```suggestion
                let active = match self.active_lock.as_ref() {
                    Some(active) => active,
                    None => {
                        // first we store it, extend timeout period and try again later
                        return self.set_new_active_lock_and_fail(existing);
                    }
                };
```

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
+    pub const DATA_PATH: &str = "#d";
+    pub const DATA_VALUE: &str = ":d";
+}
+
+mod options {
+    pub const PARTITION_KEY_VALUE: &str = "DYNAMO_LOCK_PARTITION_KEY_VALUE";
+    pub const TABLE_NAME: &str = "DYNAMO_LOCK_TABLE_NAME";
+    pub const OWNER_NAME: &str = "DYNAMO_LOCK_OWNER_NAME";
+    pub const LEASE_DURATION: &str = "DYNAMO_LOCK_LEASE_DURATION";
+    pub const SLEEP_MILLIS: &str = "DYNAMO_LOCK_SLEEP_MILLIS";
```
houqp, line 98: could you add doc strings to all the public interfaces? this is something we will be enforcing after https://github.com/delta-io/delta-rs/pull/156 is merged.

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
+                // there's no lock, we good to acquire it
+                Ok(self.upsert_new_lock().await?)
+            }
+            Some(existing) if existing.is_released => {
+                // lock is released by a caller, we good to acquire it
+                Ok(self.upsert_released_lock(existing.data).await?)
+            }
+            Some(existing) => {
+                // there's existing lock and it's out first attempt to acquire it
+                if self.active_lock.is_none() {
```
houqp, line 320: Just my own opinion, but i think `self.cached_lock` would have been a better name here.

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    fn set_new_active_lock_and_fail(&mut self, lock: LockItem) -> Result<LockItem, DynamoError> {
+        self.extend_timeout_by(lock.lease_duration);
+        self.active_lock = Some(lock);
+
+        Err(DynamoError::ConditionalCheckFailed)
+    }
+
+    fn extend_timeout_by(&mut self, seconds: u64) {
```
houqp, line 355: nitpick, this function is only used in `set_new_active_lock_and_fail`. unless you are planning to use it in other places, it would have been better to just unroll and inline it within that function, in general less indirection makes the code easier to follow.

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
+use std::collections::HashMap;
```
houqp, line 1: This file should probably be named like dynamodblock or something like that. Once it gets into a stable state, we should release it as a standalone crate :)

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
+            active_lock: None,
+            started: Instant::now(),
+            timeout_in: Duration::from_millis(self.opts.sleep_millis),
+        };
+
+        loop {
+            match state.try_acquire_lock().await {
+                Ok(lock) => return Ok(lock),
+                Err(DynamoError::ConditionalCheckFailed) => {
+                    if state.has_timed_out() {
```
houqp, line 141: i am curious when will we ever hit this branch, it looks like if the lock is not expired, we won't reach timeout because lock lookup time is set before we extend timeout_in. when the lock is expired, we will get the new lock and refresh timeout_in with the release duration.

houqp (commented):
In the dynamodb lock module, could you add a simple module level doc to describe the design from a high level? A rough overview of what item attributes are created and how we leverage these attributes to maintain different states of the lock should be suffice if you don't have time to write a detailed one :)

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    #[error("Delete item error: {0}")]
+    DeleteItemError(#[from] RusotoError<DeleteItemError>),
+}
+
+pub const PARTITION_KEY_NAME: &str = "key";
+pub const OWNER_NAME: &str = "ownerName";
+pub const RECORD_VERSION_NUMBER: &str = "recordVersionNumber";
+pub const IS_RELEASED: &str = "isReleased";
+pub const LEASE_DURATION: &str = "leaseDuration";
```
houqp, line 63: have you looked into using item ttl to manage lock expiration instead? i am assuming we are manually managing the expiration here using lease_duration because that's what the reference java implementation did?

houqp: the benefit of ttl is it could simplify the design, where expired locks will get deleted in dynamodb automatically so we can delete all the expiration checking logic in our code. if a lock item exists in dyanmodb, it's a valid and active lock, otherwise we are safe to create a new lock.

rust/src/storage/s3/dynamodb.rs
mosyp: yeah, it was used previously in two places but no anymore so that's makes total sense to inline it

rust/src/storage/s3/dynamodb.rs
mosyp: wdyt of placing this file, should it be under s3 module, or under storage itself?

build/setup_localstack.sh
```diff
@@ -0,0 +1,44 @@
...
+    echo "Waiting on $1 to start..."
+    sleep 5
+    retries=$((retries - 1))
+  done
+}
+
+wait_for "S3" "aws s3api list-buckets --endpoint-url=$ENDPOINT"
+
+echo "Uploading S3 test delta tables..."
+aws s3 rm s3://deltars --recursive --endpoint-url=$ENDPOINT > /dev/null 2>&1
```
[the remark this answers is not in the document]
mosyp: nope, not by default, but I just figured that there's `--delete` flag for it so I should use that

rust/src/storage/s3/dynamodb.rs
houqp: i think s3 would be better since it's only needed for s3 for now. when we release it as a standalone crate, we can relocate it to the repo root or create a new github repo for it.

mosyp: yeap, makes sense

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
+            })
+            .await?;
+
+        Ok(LockItem {
+            owner_name: self.opts.owner_name.clone(),
+            record_version_number: rvn,
+            lease_duration: self.opts.lease_duration,
+            is_released: false,
+            data,
+            lookup_time,
```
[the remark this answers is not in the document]
mosyp: that's how it is in the origin version, I believe they address the clock skew here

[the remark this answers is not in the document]
mosyp: yeap, that's the case https://github.com/awslabs/amazon-dynamodb-lock-client#how-we-handle-clock-skew

rust/src/storage/s3/dynamodb.rs
```diff
@@ -0,0 +1,427 @@
...
+        })
+    }
+}
+
+impl LockItem {
+    fn is_expired(&self) -> bool {
+        if self.is_released {
+            return true;
+        }
+        now_millis() - self.lookup_time > (self.lease_duration as u128) * 1000
```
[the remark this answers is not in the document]
mosyp: yes, that's correct see https://github.com/delta-io/delta-rs/pull/153#discussion_r606129584

rust/src/storage/s3/dynamodb.rs
mosyp: btw, is it possible to just copy paste some of the docs from java origin impl?

rust/src/storage/s3/dynamodb.rs
mosyp: that's why we need a cached/active lock, if rvn of retrieved existing lock matches the one we cached then we simply try again later, hence lookup_time will remain the same and eventually we'll get `has_timed_out()` to be true. Otherwise, if rvn of retrieved lock changed, then either the existing lock has been refreshed (we don't have a function for it now) or the owner has died

rust/src/storage/s3/dynamodb.rs
mosyp: ttl is not reliable since it could be up to one day between dynamodb triggers deletion and item ttl's passes. That's what originally is used in rust's dynalock and as it turned out was not right way for our case.  Also we use update_item to set is_released=true (instead of deleting) for the lock releasing which allows us to store some arbitrary metadata within lock item such us latest delta log version or kafka offsets for kafka-delta-ingest

rust/src/storage/s3/dynamodb.rs
houqp: yes, that's perfectly fine as long as they are applicable. we copy stuff from delta spec into our doc string as well :)

rust/src/storage/s3/dynamodb.rs
houqp: ha, for some reason i had the impression that dynamodb honors TTL in a synchronous and consistent fashion, that's unfortunate :(

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -0,0 +1,478 @@
...
+    opts: Options,
+}
+
+impl DynamoDbLockClient {
+    /// Creates new DynamoDB lock client
+    pub fn new(client: DynamoDbClient, opts: Options) -> Self {
+        Self { client, opts }
+    }
+
+    /// Attempts to acquire a lock until it either acquires the lock or the error is found.
```
houqp, line 204: could you expand a little bit more on what are the categories of errors that would break the loop?

rust/src/storage/s3/dynamodb.rs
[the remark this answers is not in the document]
houqp: Sorry for the confusion, let me try to clarify my original question with more context. The lookup time is recorded before we set the relative lease duration time in dynamodb. The way I understand this is lease duration starts counting after the item in dymoadb is created. However, by recording the lookup time before we send the dynamodb request, we are kicking off the expiration timer earlier than we should, which could cause false positive expiration detection.

I noticed this when I was looking at our `get_lock` implementation, where we record lookup time after the dynamodb request. This might result in longer longer time to detect expiration (false negative), but at least we can guarantee correctness. I took a quick look at `updateItemAndStartSessionMonitor` in the java implementation, they also record the lookup time after the dynamodb request.

rust/src/storage/s3/dynamodb.rs
houqp: Since we set the lock lookup time slightly earlier before extending the timeout, wouldn't it hit `cached.is_expired()` branch first before we return to do the timeout check? This branch calls `upsert_expired_lock`, which extends the timeout again before returning to the try loop. Wouldn't this make the `state.has_timed_out` check pass?

After thinking more on this, it looks like we could hit the timeout branch in the following scenario:
1. try_acquire_lock returns an active lock
2. cached lock lookup time set to now
3. timeout_in extended by lease time
4. lock expired
5. conditional lock upsert failed due to another concurrent writer got the lock first
6. back to try loop with `ConditionalCheckFailed` raised
7. timeout check failed, return timeout error

In this scenario, we exit the loop after trying to acquire the lock once. Is this an expected behavior?
In your doc string, you added `Attempts to acquire a lock until it either acquires the lock or the error is found.`, do we consider concurrent lock acquisition failures to be a failure that would break the loop?

houqp:
@xianwill I think we are almost ready to merge, you want to take a final look?

rust/src/storage/s3/dynamodb.rs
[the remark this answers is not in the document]
mosyp: > The lookup time is recorded before we set the relative lease duration time in dynamodb 

This is where item is put in dynamo https://github.com/delta-io/delta-rs/pull/153/files#diff-cdce0c557ea383d6919df0813d5e8f29bf38aa2efa60dcda197b94386f102b57R322-R327, e.g. `lookup_time` is local value and not stored in dynamodb. Moreover, none of the possible timestamps are on dynamodb.

Regarding your next assumption 
> This might result in longer time to detect expiration
yes, that is true, although with that we guarantee that the lock will not be acquire _before_ the expiration under any condition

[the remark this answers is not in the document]
mosyp: We might revisit that by ensuring that there won't be clock skew and work with timestamp in dynamodb but with less safe environment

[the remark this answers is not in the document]
mosyp: However we talk here only for non expected scenarios as when the lock is not released, which means that worker which held it had died. We might tune lease_duration later once we establish possible fault scenarios

rust/src/storage/s3/dynamodb.rs
mosyp: Okay, so I get deterministic TimedOut for one of the concurrent calls (imagine 2 workers try to acquire already expired lock) when using unit tests. However when I tested it manually everything worked. This is because in unit tests they're started in parallel however when running manually there's time between those calls, and if that time is more than a `sleep_millis` (1 sec) then everything is okay, however is less (unit tests) it fails with `TimedOut`. 

That is because 2nd, the failed worker already reaches its timeout before one more call to the dynamodb to find out that the lock has been changed. 

I think I should stop reinventing a wheel here and proceed with original impl, since they return `None` in the situation  above, and leave to the caller decide on what to do next, either fail or continuing  on trying. Also, if we'll go with `Option<LockItem>` as a result of `acquire_lock` we'll get 1-to-1 match with original version which which will allow as to move dynamodb_lock into separate library.

mosyp: @houqp
>Is this an expected behavior?

Also very good catch on this since I was sure otherwise until wrote a unit tests

rust/src/storage/s3/dynamodb_lock.rs
mosyp: I'll rewrite a documentation once change the return type to Option<LockItem> so it'll be 1:1 with original's

rust/src/storage/s3/dynamodb.rs
mosyp: @houqp I've updated are updated and are the same as in origin's since both impl are the same now

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -0,0 +1,510 @@
...
+    /// return the lock to the caller. If it does see the lock, it will note the lease
+    /// expiration on the lock. If the lock is deemed stale, then this will acquire and return it.
+    /// Otherwise, if it waits for as long as `additional_time_to_wait_for_lock` without acquiring
+    /// the lock, then it will a [`DynamoError::TimedOut].
+    ///
+    /// Note that this method will wait for at least as long as the `lease_duration` in order
+    /// to acquire a lock that already exists. If the lock is not acquired in that time,
+    /// it will wait an additional amount of time specified in `additional_time_to_wait_for_lock`
+    /// before giving up.
+    pub async fn acquire_lock(&self) -> Result<LockItem, DynamoError> {
```
mosyp, line 242: @houqp Okay so there's no `Option` yet as in original's but here's some thoughts. There's 2 expected ways of exiting a acquire loop, [someone else acquired the lock](https://github.com/awslabs/amazon-dynamodb-lock-client/blob/master/src/main/java/com/amazonaws/services/dynamodbv2/AmazonDynamoDBLockClient.java#L532)  or [provisioned throughput exceeded](https://github.com/awslabs/amazon-dynamodb-lock-client/blob/master/src/main/java/com/amazonaws/services/dynamodbv2/AmazonDynamoDBLockClient.java#L538).
However they use single exception for those cases and a new method `public Optional<LockItem> tryAcquireLock` to try-catch on that and return None if `LockNotGrantedException` is caught.  

We might however still use `TimedOut` and new one for "provisioned throughput exceeded" to pattern match for `Option<LockItem>`.

wdyt?

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -0,0 +1,510 @@
...
+    /// Error that indicates the condition in the DynamoDB operation could not be evaluated.
+    /// Mostly used by [`DynamoDbLockClient::acquire_lock`] to handle unsuccessful retries
+    /// of acquiring the lock.
+    #[error("Conditional check failed")]
+    ConditionalCheckFailed,
+
+    /// Error that is returned by [`DynamoDbLockClient::release_lock`] when the given lock
+    /// has been already expired and could not be released.
+    #[error("Lock is expired")]
+    LockIsExpired,
```
mosyp, line 113: Is this naming ok?

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -0,0 +1,510 @@
...
+
+    /// The required field of [`LockItem`] is missing in DynamoDB record or has incompatible type.
+    #[error("DynamoDB item has invalid schema")]
+    InvalidItemSchema,
+
+    /// Error returned by [`DynamoDbLockClient::acquire_lock`] which indicates that the lock could
+    /// not be acquired for more that returned number of seconds.
+    #[error("Could not acquire lock for {0} sec")]
+    TimedOut(u64),
```
mosyp, line 122: and this one?

rust/src/storage/s3/dynamodb.rs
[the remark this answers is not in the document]
houqp: > > This might result in longer time to detect expiration
>
> yes, that is true, although with that we guarantee that the lock will not be acquire before the expiration under any condition

Yep, what i meant in my previous reply was I think this behavior is OK since it guarantees correctness. However, this doesn't seem to be the behavior for `upsert_item` because local timer might expire sooner than the configured lease duration. Basically I am proposing that we do:

```rust
        // dyamodb put_item request
        self.client.put_item(...);

        Ok(LockItem {
            owner_name: self.opts.owner_name.clone(),
            record_version_number: rvn,
            lease_duration: self.opts.lease_duration,
            is_released: false,
            data,
            now_millis(),
        })
```

Instead of what we have right now:

```rust
        let lookup_time = now_millis();
 
        // dyamodb put_item request
        self.client.put_item(...);

        Ok(LockItem {
            owner_name: self.opts.owner_name.clone(),
            record_version_number: rvn,
            lease_duration: self.opts.lease_duration,
            is_released: false,
            data,
            lookup_time,
        })
```

This should make the local timer behavior consistent with our `get_lock` and upstream's `updateItemAndStartSessionMonitor` implementation.

rust/src/storage/s3/dynamodb_lock.rs
houqp: yep, i think your design is much better than the reference implementation :) I am not a fan of exception anyway, not to mention a single catch all exception :P

rust/src/storage/s3/dynamodb_lock.rs
houqp: yep, can probably be further simplified to `LockExpired`.

rust/src/storage/s3/dynamodb_lock.rs
houqp: LGTM 👍

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -0,0 +1,510 @@
...
+            _ => DynamoError::GetItemError(error),
+        }
+    }
+}
+
+impl From<RusotoError<UpdateItemError>> for DynamoError {
+    fn from(error: RusotoError<UpdateItemError>) -> Self {
+        match error {
+            RusotoError::Service(UpdateItemError::ConditionalCheckFailed(_)) => {
+                DynamoError::LockIsExpired
```
houqp, line 161: The conditional put could fail due to other constraint failures right? For example rvn or data field not matched?

rust/src/storage/s3/dynamodb.rs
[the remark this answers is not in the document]
mosyp: oh wow, that's true

rust/src/storage/s3/dynamodb_lock.rs
mosyp: Okay yes, this one is questionable. Because reference implementation returns a boolean value where false catches all of the cases, basically when condition is failed. I'll change it that way.

Also this `From` is invalid because if we'd add other UpdateItem this'll be broken and misleading. I'll move this sort of pattern matching into `release_lock` function

xianwill (approved):
Nice work!

houqp (approved):
Good work! I might be biased, but I think @mosyp 's implementation is better and more readable than the java implementation.

On top of that, we have Rust compiler watching our back to make sure there would be no data race condition bug when used in a mulit-threaded setting. It's going to become the best dynamodb lock implementation when we move it into a standalone crate :)

PR #158, opened 2021-03-25
Fix/cargo clippy

fvaleye:
# Description
Fix clippy errors

### DIFF
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -148,7 +148,7 @@ mod tests {
             .put_obj(new_file_path.to_str().unwrap(), b"hello")
             .await
         {
-            panic!(format!("Expect put_obj to return Ok, got Err: {:#?}", e));
+            panic!("Expect put_obj to return Ok, got Err: {:#?}", e)
         }
 
         // second try should result in already exists error
--- a/rust/src/storage/file/rename.rs
+++ b/rust/src/storage/file/rename.rs
@@ -119,8 +119,8 @@ mod tests {
                     }
                 }
             }
-            Err(e) => panic!(format!("expect std::io::Error, got: {:#}", e)),
-            Ok(()) => panic!("expect rename to fail with Err, but got Ok"),
+            Err(e) => panic!("expect std::io::Error, got: {:#}", e),
+            Ok(()) => panic!("{}", "expect rename to fail with Err, but got Ok"),
         }
 
         // successful move A to C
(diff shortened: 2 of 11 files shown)


rust/src/storage/file/mod.rs
```diff
@@ -148,7 +148,8 @@ mod tests {
             .put_obj(new_file_path.to_str().unwrap(), b"hello")
             .await
         {
-            panic!(format!("Expect put_obj to return Ok, got Err: {:#?}", e));
+            let error_message = format!("Expect put_obj to return Ok, got Err: {:#?}", e);
+            panic!("{}", error_message);
```
houqp, line 152: would be better to rewrite it as `panic!("Expect put_obj to return Ok, got Err: {:#?}", e)`

rust/src/storage/file/rename.rs
```diff
@@ -119,8 +119,11 @@ mod tests {
                     }
                 }
             }
-            Err(e) => panic!(format!("expect std::io::Error, got: {:#}", e)),
-            Ok(()) => panic!("expect rename to fail with Err, but got Ok"),
+            Err(e) => {
+                let error_message = format!("expect std::io::Error, got: {:#}", e);
+                panic!("{}", error_message)
```
houqp, line 124: same here, we can just use `panic!` like `format!`.

rust/src/storage/file/mod.rs
fvaleye: 👍

rust/src/storage/file/rename.rs
fvaleye: 👍

fvaleye:
@houqp wait for the CI but it is fixed for the error in Windows build (wrong space at the end of the filename 🤦 )

houqp (approved):
the windows build error looks weird, what's up with that?

fvaleye:
> the windows build error looks weird, what's up with that?

The file `config.yml ` in ISSUE_TEMPLATE merged in `main` contains a space at the end.
It is fixed in this PR but I think that the CI failed because it tries to fetch the current main branch...

houqp:
i see, that makes sense

houqp:
@fvaleye Looks like the windows build is still failing on master, so it's something different.

houqp:
oh, guess we need to delete the invalid config.yml file as well?

fvaleye:
You are right! I fixed it 😃

PR #159, opened 2021-03-26
Documentation/python binding

fvaleye:
# Description
Add documentation (Docstrings) in Python Binding


houqp:
@rtyler looks like s3 credential is broken, we are getting 403 errors for all s3 tests.

python/deltalake/table.py
```diff
@@ -10,16 +10,31 @@
...
     def files(self) -> List[str]:
+        """
+        Get the .parquet files of the DeltaTable.
+        :return: list of the .parquet files referenced for the current version of the DeltaTable
+        """
         return self._table.files()
 
     def files_by_partitions(self, partition_filters: List[Tuple]) -> List[str]:
         """
+        Get the files that matched after applying a list of partitions filters.
```
houqp, line 37:
```suggestion
        Get the files that match a given list of partitions filters.
```

python/deltalake/table.py
```diff
@@ -68,4 +106,8 @@ def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
             return dataset(file_paths, format="parquet")
 
     def to_pyarrow_table(self) -> pyarrow.Table:
+        """
+        Read the DeltaTable with the PyArrow Table format.
```
houqp, line 110:
```suggestion
        Build a PyArrow Table using data from the DeltaTable.
```

python/deltalake/table.py
```diff
@@ -45,18 +63,38 @@ def files_by_partitions(self, partition_filters: List[Tuple]) -> List[str]:
...
     def pyarrow_schema(self) -> pyarrow.Schema:
+        """
+        Get the current schema of the DeltaTable with the Parquet PyArrow format.
+        :return: the current Schema with the Parquet PyArrow format
+        """
         return pyarrow_schema_from_json(self._table.arrow_schema_json())
 
     def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
+        """
+        Read the DeltaTable with the PyArrow dataset format.
```
houqp, line 95: Using build to suggest that the resulting Dataset is self-contained and has no reference to the original delta table anymore.

```suggestion
        Build a PyArrow Dataset using data from the DeltaTable.
```

houqp (approved):
LGTM 👍

PR #160, opened 2021-03-26
Enable date column support for delta -> arrow schema conversion.

dispanser:
# Description

Enable support for `primitive(date)` delta schema type with datafusion / arrow.

Noticed the other day that I wasn't able to load a delta table with a `date` column using datafusion. The same table loaded just fine using the parquet-based code paths (ignoring delta), so I assumed that there must be a way.

The proposed fix is just a one-line change in the schema conversion from delta to arrow.


houqp (approved):
interesting, based on the description in delta spec, i thought it's stored as a string :P

dispanser:
> interesting, based on the description in delta spec, i thought it's stored as a string :P

Sorry for not consulting the specs first; I guess sometimes it's better to show up uninitiated :-).

I'm looking forward for some more write support, so that we can stop adding more and more varieties of delta tables into `tests/data` and move to roundtrip property-based tests.

PR #166, opened 2021-03-28
Fix/add timestamp type

fvaleye:
# Description
Add the `timestamp` type for PyArrow schema with the default unit `microsecond` for the DeltaTable timestamp type (DeltaTable timestamp don't specify the unit for the timestamp type).


houqp (approved):

houqp:
@fvaleye looks like your local main branch is out of sync :P

PR #169, opened 2021-03-29
support custom endpoint url for S3 backend

### LINKED ISSUE #163: Support minio and other S3 API compatible services
# Description

To support S3 compatible services like minio, we need to provide a way for users to specify custom endpoint url through environment variables or other means.

https://docs.rs/rusoto_core/0.46.0/rusoto_core/enum.Region.html#aws-compatible-services has the example on how this can be done by setting a custom region object when initiating the S3 client.

https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/s3.rs#L146 is where we create the region object in delta-rs, so we just need to change the code to detect the endpoint url override there.

houqp:
# Description

S3 compatible services like minio and localstack need to be accessed by setting a custom endpoint url. This PR adds support for setting the endpoint url through `AWS_ENDPOINT_URL` env var. The env var name is picked based on aws cli argument `--endpoint-url`.

# Related Issue(s)

- closes #163

### DIFF
--- a/rust/src/storage/s3.rs
+++ b/rust/src/storage/s3.rs
@@ -143,7 +143,16 @@ pub struct S3StorageBackend {
 
 impl S3StorageBackend {
     pub fn new() -> Self {
-        let client = create_s3_client(Region::default()).unwrap();
+        let region = if let Ok(url) = std::env::var("AWS_ENDPOINT_URL") {
+            Region::Custom {
+                name: "custom".to_string(),
+                endpoint: url.clone(),
+            }
+        } else {
+            Region::default()
+        };
+
+        let client = create_s3_client(region).unwrap();
         Self { client }
     }
 }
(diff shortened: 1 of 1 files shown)


fvaleye (approved):

fvaleye (commented):
LGTM, after fixing clippy errors!

rust/src/storage/s3.rs
```diff
@@ -143,7 +143,16 @@ pub struct S3StorageBackend {
 
 impl S3StorageBackend {
     pub fn new() -> Self {
-        let client = create_s3_client(Region::default()).unwrap();
+        let region = if let Ok(url) = std::env::var("AWS_ENDPOINT_URL") {
+            Region::Custom {
+                name: "custom".to_string(),
+                endpoint: url.clone(),
```
fvaleye, line 149:
```suggestion
                endpoint: url,
```

PR #171, opened 2021-03-30
Ignore the S3 acceptance tests until we get a better acceptance test

rtyler:
I may just need to recycle these keys


houqp:
there are python tests as well.

houqp:
also https://github.com/delta-io/delta-rs/pull/168 should help troubleshoot issues like this :)

rtyler:
@houqp would you mind doing the skipping of the python tests, I don't have a working local Python development env for this

PR #182, opened 2021-04-04
tla+ spec for dynamodb lock

houqp:
# Description

Spec is model checked without pause and crash.

Adding crash and pause to the modeling requires enabling timer in the
spec (i.e. setting TIME_TICK_UNIT > 0). This will fail the model check
because lock with expiration cannot gurantee correctness without use of
fenching token.

### DIFF
--- a/tlaplus/README.md
+++ b/tlaplus/README.md
@@ -1,0 +1,7 @@
+This folder contains TLA+ proofs for various subsystems of the project including:
+
+* Distributed Dynamodb lock
+* S3 multi-writer design
+
+To model check these specs, you need to use [TLA+
+toolbox](https://lamport.azurewebsites.net/tla/toolbox.html).
--- a/tlaplus/dynamodblock.tla
+++ b/tlaplus/dynamodblock.tla
@@ -1,0 +1,633 @@
+---------------------------- MODULE dynamodblock ----------------------------
+
+EXTENDS TLC, Integers, Sequences, FiniteSets
+
+CONSTANT NULL
+    , WRITERS  \* number of writers defined as model values set
+    , LOOP_COUNT  \* how many loop iteration each writer should go through
+    \* set TIME_TICK_UNIT to 0 to disable lock expiration check, > 1 to enable it
+    , TIME_TICK_UNIT
+
+ASSUME Cardinality(WRITERS) > 0
+ASSUME TIME_TICK_UNIT \in Nat
+ASSUME LOOP_COUNT \in Nat \ {0}
+
+(*--algorithm dynamodblock
+variables
+    shared_counter = 0,
+    time_now = 0,
+    active_writer = Cardinality(WRITERS),
+
+    \* local record version is only guaranteed to be unique within worker process, not globally
+    local_record_version = [ x \in WRITERS |-> 0 ],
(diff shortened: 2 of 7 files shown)


houqp:
also cc @mosyp who is the author of the dymodb lock.

The spec is probably not very readable for those who hasn't worked with TLA+ before. If you really want to read it, only read lines before `\* BEGIN TRANSLATION (chksum(pcal) = "e6415912" /\ chksum(tla) = "ffc65bb2")`. The rest are transpiled TLA+ code from the PlusCal algorithm code defined in the comment.

tlaplus/dynamodblock.tla
```diff
@@ -0,0 +1,633 @@
...
+        /\ cached_lock_lookup_time_ = [ self \in ProcSet |-> NULL]
+        /\ cached_lock_duration_ = [ self \in ProcSet |-> NULL]
+        /\ cached_lock_version = [ self \in ProcSet |-> NULL]
+        (* Procedure CriticalSection *)
+        /\ local_count_value = [ self \in ProcSet |-> NULL]
+        /\ stack = [self \in ProcSet |-> << >>]
+        /\ pc = [self \in ProcSet |-> CASE self \in WRITERS -> "writerloop"
+                                        [] self = TIME_TICK_UNIT -> "tick"]
+
+cond_put(self) == /\ pc[self] = "cond_put"
```
rtyler, line 255: ohgod


This looks like Pascal and LaTeX joined forces with brainfuck

rtyler (commented):
Uh, please add a README or something for this new folder :scream_cat:

tlaplus/dynamodblock.tla
houqp: Yeah, it's heavily inspired by LaTex, one of the features it provides is rendering the spec in a pdf just like latex :(

mosyp:
Wow, that is impressive diffs, how to run these specs?

houqp:
@mosyp you will need to download tla+ toolbox and load the project to run it. See download link in the readme.

tlaplus/README.md
```diff
@@ -0,0 +1,7 @@
+This folder contains TLA+ proofs for various subsystems of the project including:
+
+* Distributed Dynamodb lock
+* S3 multi-writer design
+
+To model check these specs, you need to use [TLA+
+toolbox](https://lamport.azurewebsites.net/tla/toolbox.html).
```
xianwill, line 7: Might be worth mentioning [this confusing behavior](https://github.com/tlaplus/tlaplus/issues/350) when opening a spec in the toolbox in the README.

xianwill (approved):

Alexander-N:
Hi,
I landed here because I was searching for TLA+ applications and thought I'll share a few things which may be helpful:
* TLA+ specs can be checked from the command line, either [directly](https://github.com/lemmy/BlockingQueue) or with [other](https://github.com/pmer/tla-bin) [tools](https://github.com/hwayne/tlacli).
* To make checking the spec more straightforward I would also add a cfg file, for example as in https://github.com/fpaxos/fpaxos-tlaplus.
* If you can't get used to the latex syntax, it's possible to use VSCode and configure it so that it will be hidden: https://github.com/alygin/vscode-tlaplus/wiki/Fonts

PR #185, opened 2021-04-05
Bump Python crate version to v.0.4.5

fvaleye:
# Description
Bump Python crate version to `0.4.5`

### DIFF
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-python"
-version = "0.4.4"
+version = "0.4.5"
 authors = ["Qingping Hou <dave2008713@gmail.com>"]
 homepage = "https://github.com/delta-io/delta-rs"
 license = "Apache-2.0"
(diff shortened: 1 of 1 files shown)


fvaleye:
@houqp While preparing the `0.4.5` release for the Python bindings, I noticed that the new version of numpy has dropped the support of Python 3.6: `Python version >= 3.7 required.`.

Do we also need to drop the support of Python 3.6?

houqp:
@fvaleye depends on whether we have to use the new version of numpy. As far as I can tell, we are not pinning numpy versions at the moment, meaning we should be able to work with new and old numpy versions.

houqp:
ok, see the build failure now :) we can update the python version in our CI/CD pipeline to 37, 36 users would just have to manually pin numpy versions for now.

There is a way to specify dependency version constraints based on host python versions, but I will need some time to look it up.

houqp:
I think this is the syntax `numpy<1.20.0;python_version<="3.6"`, @fvaleye could you give that a try in Cargo.toml?

fvaleye:
> I think this is the syntax `numpy<1.20.0;python_version<="3.6"`, @fvaleye could you give that a try in Cargo.toml?

It worked! Thx 👍

houqp (approved):

PR #191, opened 2021-04-13
Add Metadata in Python bindings

### LINKED ISSUE #190: Support for Table metadata with python
# Description
Adds the ability to retrieve Table metadata like description/comment, right now its only possible for columns

**Use Case**
I am developing a POC that reads the delta from s3 and saves it on Glue, the only missing feature is to save table comment/description to Glue because there's no way to retrieve that information using delta-rs

```
dt = DeltaTable("<PATH>")
dt.metadata()
```

fvaleye:
# Description
Add the metadata in the Python bindings.

Metadata:
- id
- name
- description
- configuration
- partition_columns

# Related Issue(s)
- closes #190

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -4,8 +4,60 @@ from urllib.parse import urlparse
 import pyarrow
 from pyarrow.dataset import dataset
 
-from .deltalake import RawDeltaTable
+from .deltalake import RawDeltaTable, RawDeltaTableMetaData
 from .schema import Schema, pyarrow_schema_from_json
+
+
+class Metadata:
+    """ Create a Metadata instance. """
+
+    def __init__(self, table: RawDeltaTable):
+        self._metadata = table.metadata()
+
+    @property
+    def id(self):
+        """ Return the unique identifier of the DeltaTable. """
+        return self._metadata.id
+
+    @property
+    def name(self):
+        """ Return the user-provided identifier of the DeltaTable. """
+        return self._metadata.name
+
+    @property
+    def description(self):
+        """ Return the user-provided description of the DeltaTable. """
+        return self._metadata.description
+
+    @property
+    def configuration(self):
+        """ Return the DeltaTable properties. """
+        return self._metadata.configuration
+
+    @property
+    def partition_columns(self):
+        """ Return an array containing the names of the partitioned columns of the DeltaTable. """
+        return self._metadata.partition_columns
+
+    def __str__(self) -> str:
+        return (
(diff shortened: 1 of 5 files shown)


python/src/lib.rs
```diff
@@ -42,6 +43,20 @@ struct RawDeltaTable {
     _table: deltalake::DeltaTable,
 }
 
+#[pyclass]
+struct RawDeltaTableMetaData {
```
houqp, line 47: Curious what's the reason for wrapping this struct in python instead of just exposing it as `Metadata`. Are you concerned about extensibility from python side? If so, we could subclass `Metadata` with `RawDeltaTableMetaData`, that why, we won't need to copy all the attributes again in the `__init__` method.

fvaleye: Good question!

I started by working directly with the `Metadata` structure without having a Python class. And I didn't feel comfortable manipulating a Python object without knowing anything about its structure (documentation, debug, attributes & methods). So, I changed my mind by creating the Python class and I was wondering if it might also be interesting to add the documentation on the Python properties also.

houqp (approved):

python/deltalake/table.py
```diff
@@ -4,10 +4,41 @@
 import pyarrow
 from pyarrow.dataset import dataset
 
-from .deltalake import RawDeltaTable
+from .deltalake import RawDeltaTable, RawDeltaTableMetaData
 from .schema import Schema, pyarrow_schema_from_json
 
 
+class Metadata:
```
fvaleye, line 11: @houqp: I was wondering if I could change this class (for documentation purposes) to something like this but I need your opinion on that:

```
class Metadata:
    """ Create a Metadata instance. """

    def __init__(self, table: RawDeltaTable):
        self._metadata = table.metadata()

    @property
    def id(self):
        """ Return the unique identifier of the DeltaTable. """
        return self._metadata.id

    @property
    def name(self):
        """ Return the user-provided identifier of the DeltaTable. """
        return self._metadata.name

    @property
    def description(self):
        """ Return the user-provided description of the DeltaTable. """
        return self._metadata.description

    @property
    def configuration(self):
        """ Return the DeltaTable properties. """
        return self._metadata.configuration

    @property
    def partition_columns(self):
        """ Return an array containing the names of the partitioned columns of the DeltaTable. """
        return self._metadata.partition_columns
```

houqp: sounds like a good idea :+1:

fvaleye: Ok I will add this in this PR :)

houqp (approved):

PR #194, opened 2021-04-13
Change the default timestamp to nanosecond precision for Delta <> Arrow schema

fvaleye:
# Description
The default timestamp type for parquet files is INT96, which is with nanosecond precision

# Related Issue(s)
- https://github.com/delta-io/delta/issues/643


houqp (approved):

houqp:
@fvaleye worth starting the same discussion you asked in that ticket in the delta lake dev mailing list, which seems to be more actively monitored than github.

PR #197, opened 2021-04-15
Unify s3 and fs try_commit behavior

mosyp:
# Description
Before introducing S3 locking we need to match its behaviour with filesystem backend. 
Also, temporary files are now created before commit loop, to avoid unnecessary creation of tmp files at each try. 

I've also made simple_commit to be tested against s3 backend, so with these we have first s3 writer test

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -213,7 +213,6 @@ pub struct DeltaTable {
 
     last_check_point: Option<CheckPoint>,
     log_path: String,
-    commit_path: String,
     version_timestamp: HashMap<DeltaDataTypeVersion, i64>,
 }
 
@@ -223,9 +222,9 @@ impl DeltaTable {
         self.storage.join_path(&self.log_path, &version)
     }
 
-    fn commit_path(&self, token: &str) -> String {
-        let path = format!("_commit_{}", token);
-        self.storage.join_path(&self.commit_path, &path)
+    fn tmp_commit_log_path(&self, token: &str) -> String {
+        let path = format!("_commit_{}.json", token);
+        self.storage.join_path(&self.log_path, &path)
     }
 
     fn get_checkpoint_data_paths(&self, check_point: &CheckPoint) -> Vec<String> {
@@ -665,7 +664,6 @@ impl DeltaTable {
         storage_backend: Box<dyn StorageBackend>,
     ) -> Result<Self, DeltaTableError> {
         let log_path_normalized = storage_backend.join_path(table_path, "_delta_log");
-        let commit_path_normalized = storage_backend.join_path(table_path, "_prepared_commits");
         Ok(Self {
             version: 0,
             state: DeltaTableState::default(),
@@ -673,7 +671,6 @@ impl DeltaTable {
             table_path: table_path.to_string(),
             last_check_point: None,
             log_path: log_path_normalized,
-            commit_path: commit_path_normalized,
             version_timestamp: HashMap::new(),
         })
     }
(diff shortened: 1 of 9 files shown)


rust/src/delta.rs
```diff
@@ -658,13 +665,15 @@ impl DeltaTable {
         storage_backend: Box<dyn StorageBackend>,
     ) -> Result<Self, DeltaTableError> {
         let log_path_normalized = storage_backend.join_path(table_path, "_delta_log");
+        let commit_path_normalized = storage_backend.join_path(table_path, "_prepared_commits");
         Ok(Self {
```
mosyp, line 669: I;m still not confident whether it is a right way to do it.
E.g. before, we stored tmp files for fs within a root dir, but now, since we plan to introduce a recovery option, it'd be handy to store these files within its separate directory. However, since rename for fs is atomic, we might not need that. So does it make sense to move this into storage level code?

rust/src/storage/file/mod.rs
```diff
@@ -85,84 +84,52 @@ impl StorageBackend for FileStorageBackend {
...
-    }
-}
-
-const CREATE_TEMPORARY_FILE_MAX_TRIES: i32 = 5;
-
-async fn create_tmp_file_with_retry(
-    backend: &FileStorageBackend,
-    path: &str,
-    obj_bytes: &[u8],
-) -> Result<String, StorageError> {
```
mosyp, line 105: there's no retry with these changes anymore since they didn't suite well with new approach, we might reconsider it's later

rust/src/storage/mod.rs
```diff
@@ -180,6 +180,12 @@ pub enum StorageError {
         source: rusoto_core::RusotoError<rusoto_s3::PutObjectError>,
     },
     #[cfg(feature = "s3")]
+    #[error("Failed to delete S3 object: {0}")]
+    S3Delete(#[from] rusoto_core::RusotoError<rusoto_s3::DeleteObjectError>),
+    #[cfg(feature = "s3")]
+    #[error("Failed to copy S3 object: {0}")]
+    S3Copy(#[from] rusoto_core::RusotoError<rusoto_s3::CopyObjectError>),
+    #[cfg(feature = "s3")]
```
mosyp, line 188: It is different to the s3 errors above, I feel like with this approach it's cleaner looking and easier to pattern match, however it differs from other s3 errors, should I put it back, OR should I change other, existing errors

rust/src/delta.rs
```diff
@@ -914,23 +924,36 @@ impl<'a> DeltaTransaction<'a> {
...
+            .storage
+            .put_obj(&commit_path, log_entry)
+            .await?;
+
+        Ok(commit_path)
+    }
+
+    async fn try_commit(
+        &mut self,
+        commit_file: &str,
```
houqp, line 944: nitpick, would be good to use a different variable name to make the naming more consistent. Below we have log_path, which is also a path with string type, so we should use the same suffix here. For example, something like `tmp_log_path`.

houqp (commented):
the rest looks good to me :+1:

rust/src/delta.rs
mosyp: yeah, makes sense

rust/src/delta.rs
mosyp: yeap, let's not bring new concept, e.g. new dir, and place within table_path, later we'll se how that'll work for us

mosyp: actually, on the second thought I've placed them under log_path, because a) it so connected to log rather then files, b) iirc data file names are arbitary hence we won't have a name collision if someone would try to name a file `_commit_...`

houqp (approved):

xianwill (approved):
LGTM

PR #199, opened 2021-04-16
Fix fs parent dir creation and tx version log update

xianwill:
# Description

This commit contains two drive by bug fixes I encountered when upgrading to the latest delta-rs commit.

1. Ensure parent dir exists within fs backend put_obj
2. Update the latest tx version for txn commits when reading log entries.

The behavior I observed that led to these two bug fixes:

1. For partitioned tables, file backend failed on rename since partition folder does not exist on first write.
2. The delta metadata did not contain the latest txn version after calling update on the delta table. So log always contained txn version of 1.

### DIFF
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -87,6 +87,12 @@ impl StorageBackend for FileStorageBackend {
     async fn put_obj(&self, path: &str, obj_bytes: &[u8]) -> Result<(), StorageError> {
         let tmp_path = create_tmp_file_with_retry(self, path, obj_bytes).await?;
 
+        let dir = std::path::Path::new(path);
+
+        if let Some(d) = dir.parent() {
+            fs::create_dir_all(d).await?;
+        }
+
         if let Err(e) = rename::atomic_rename(&tmp_path, path) {
             fs::remove_file(tmp_path).await.unwrap_or(());
             return Err(e);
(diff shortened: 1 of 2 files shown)


rust/src/storage/file/mod.rs
```diff
@@ -87,6 +87,12 @@ impl StorageBackend for FileStorageBackend {
     async fn put_obj(&self, path: &str, obj_bytes: &[u8]) -> Result<(), StorageError> {
         let tmp_path = create_tmp_file_with_retry(self, path, obj_bytes).await?;
 
+        let dir = std::path::Path::new(path);
+
+        if let Some(d) = dir.parent() {
+            fs::create_dir_all(d).await?;
+        }
+
```
mosyp, line 95: Yeap, I've notice that too, but my recent PR changes the way put_obj and rename_obj should work, please see https://github.com/delta-io/delta-rs/pull/197/files#diff-ce6ba0c9d799f33c7c8ec4b5d7f36e1c03c8451390190a348b530caf838c89dfR87-R88

Because current version of put_obj of file brings ambiguity, e.g. it's not actual put_obj but the whole ceremony

mosyp: But I'm happy to merge your changes and I'll update my PR then

xianwill:
Closing since @mosyp has a PR for the parent dir fix already. I will submit a separate PR for the transasction log fix.

PR #201, opened 2021-04-16
Update app tx version when processing action

xianwill:
# Description

This is a drive-by fix.

I encountered a bug when using the delta txn version from kafka-delta-ingest. The prior `process_action` implementation does not update the existing map entry for tx version when re-processing actions. This change resolves that.


houqp (approved):
good catch 👍  could you also write a simple test?

xianwill:
> good catch 👍 could you also write a simple test?

will do

houqp:
> @houqp - While writing the test, I noticed that process_action was contained within the DeltaTable scope even though it doesn't have a self ref in the sig. Feels more clear to hoist the code into the outer scope to me

I totally agree!

PR #202, opened 2021-04-17
Update arrow deps

nevi-me:
# Description

This updates the arrow deps to the latest commit (https://github.com/apache/arrow/commit/9c1e5bd19347635ea9f373bcf93f2cea0231d50a). 

There's the following changes:
* `TableProvider` signature in datafusion now takes a limit
* Parquet writer returns statistics (more relevant for kafka-delta-ingest)

# Documentation

No changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,27 +97,6 @@ checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 [[package]]
 name = "arrow"
 version = "4.0.0-SNAPSHOT"
-source = "git+https://github.com/apache/arrow.git?rev=05b36567bd8216bec71b796fe3bb6811c71abbec#05b36567bd8216bec71b796fe3bb6811c71abbec"
-dependencies = [
- "cfg_aliases",
- "chrono",
- "csv",
- "flatbuffers",
- "hex",
- "indexmap",
- "lazy_static",
- "lexical-core",
- "num",
- "rand 0.7.3",
- "regex",
- "serde",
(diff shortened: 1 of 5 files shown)


houqp (dismissed):

houqp:
looks like there are two minor breaking api changes in datafusion.

nevi-me:
> looks like there are two minor breaking api changes in datafusion.

hmm, didn't see them with tests, maybe I'm not doing something right. I'll have a look at them

nevi-me:
> looks like there are two minor breaking api changes in datafusion.

thanks, looks like I hadn't run tests with the datafusion feature

nevi-me:
@houqp I'm unable to resolve the Python error. Seems like the relevant function is expecting `std::convert::Infallible` from the arrow conversion, but I can't see what change would have made this to start failing.

Would you be able to help me out with this?

houqp:
@nevi-me I found the issue, it's because the arrow crate is also pinned with git hash in `python/Cargo.toml`, so cargo is installing two different versions of arrow, which resulted in the following error:

```
`the trait `std::convert::From<&deltalake::Schema>` is not implemented for `arrow::datatypes::Schema`
```

Really need that 4.x arrow release :P

houqp:
Looks like datafusion tests need to be updated to use the new table provider interface as well.

nevi-me:
This is now all good @houqp

houqp (approved):

houqp:
looks like there might be some kind of race bug in our table read tests in windows platform, it was failing due to incorrect output, but a restart fixed it.

PR #204, opened 2021-04-19
Add dry_run vacuum command

fvaleye:
# Description
Add a Vacuum dry run command for listing files no longer referenced by a Delta table and are older than the retention threshold.

# Documentation
https://docs.delta.io/0.8.0/delta-utility.html#remove-files-no-longer-referenced-by-a-delta-table

### DIFF
--- a/python/tests/test_table_read.py
+++ b/python/tests/test_table_read.py
@@ -17,6 +17,36 @@ def test_read_simple_table_by_version_to_dict():
     assert dt.to_pyarrow_dataset().to_table().to_pydict() == {"value": [1, 2, 3]}
 
 
+def test_vacuum_dry_run_simple_table():
+    table_path = "../rust/tests/data/delta-0.2.0"
+    dt = DeltaTable(table_path)
+    retention_periods = 169
+    assert dt.vacuum(retention_periods) == [
+        "../rust/tests/data/delta-0.2.0/part-00000-512e1537-8aaa-4193-b8b4-bef3de0de409-c000.snappy.parquet",
+        "../rust/tests/data/delta-0.2.0/part-00000-b44fcdb0-8b06-4f3a-8606-f8311a96f6dc-c000.snappy.parquet",
+        "../rust/tests/data/delta-0.2.0/part-00001-185eca06-e017-4dea-ae49-fc48b973e37e-c000.snappy.parquet",
+        "../rust/tests/data/delta-0.2.0/part-00001-4327c977-2734-4477-9507-7ccf67924649-c000.snappy.parquet",
+    ]
+
+    retention_periods = -1
+    with pytest.raises(Exception) as exception:
+        dt.vacuum(retention_periods)
+    assert str(exception.value) == "The retention periods should be positive."
+
+    retention_periods = 167
+    with pytest.raises(Exception) as exception:
+        dt.vacuum(retention_periods)
+    assert (
+        str(exception.value)
+        == "Invalid retention period, retention for Vacuum must be greater than 1 week (168 hours)"
+    )
+
+    retention_periods = 167
+    with pytest.raises(Exception) as exception:
+        dt.vacuum(retention_periods, dry_run=False)
+    assert str(exception.value) == "Only Vacuum with dry_run is available."
+
+
 def test_read_partitioned_table_metadata():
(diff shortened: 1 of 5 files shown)


rust/src/delta.rs
```diff
@@ -605,6 +608,29 @@ impl DeltaTable {
...
     }
 
+    /// Run the dry run of the Vacuum command on the Delta Table: list files no longer referenced by a Delta table and are older than the retention threshold.
+    pub fn vacuum_dry_run(&self, retention_hours: u64) -> Result<Vec<String>, DeltaTableError> {
+        let retention_checked = match retention_hours {
+            _ if retention_hours < 168 => {
+                return Err(DeltaTableError::InvalidVacuumRetentionPeriod)
+            }
+            _ => retention_hours,
+        };
```
houqp, line 618: looks like the match can be simplified to a if block if we use rentention_hours directly after the check:


```suggestion
        if retention_hours < 168 {
            return Err(DeltaTableError::InvalidVacuumRetentionPeriod);
        }
```

rust/src/delta.rs
```diff
@@ -605,6 +608,29 @@ impl DeltaTable {
         self.state.min_writer_version
     }
 
+    /// Run the dry run of the Vacuum command on the Delta Table: list files no longer referenced by a Delta table and are older than the retention threshold.
+    pub fn vacuum_dry_run(&self, retention_hours: u64) -> Result<Vec<String>, DeltaTableError> {
+        let retention_checked = match retention_hours {
+            _ if retention_hours < 168 => {
```
houqp, line 614: Could you copy the warning text from the official doc to explain the recommended minimal 7 days retention magic number?

houqp (approved):
LGTM! @xianwill @mosyp could you take a look at this as well?

rust/src/delta.rs
```diff
@@ -109,6 +110,8 @@ pub enum DeltaTableError {
     PartitionError { partition: String },
     #[error("Invalid partition filter found: {}.", .partition_filter)]
     InvalidPartitionFilter { partition_filter: String },
+    #[error("Invalid retention period, retention for Vacuum must be superior than 7 weeks.")]
```
rtyler, line 113:
```suggestion
    #[error("Invalid retention period, retention for Vacuum must be greater than 1 week (168 hours)")]
```

I believe this was a typo

python/tests/test_table_read.py
```diff
@@ -17,6 +17,36 @@ def test_read_simple_table_by_version_to_dict():
...
+    with pytest.raises(Exception) as exception:
+        dt.vacuum(retention_periods)
+    assert str(exception.value) == "The retention periods should be positive."
+
+    retention_periods = 167
+    with pytest.raises(Exception) as exception:
+        dt.vacuum(retention_periods)
+    assert (
+        str(exception.value)
+        == "Invalid retention period, retention for Vacuum must be superior than 7 weeks."
```
rtyler, line 41:
```suggestion
        == "Invalid retention period, retention for Vacuum must be greater than 1 week (168 hours)"
```
See comment below

rust/tests/read_delta_test.rs
```diff
@@ -187,3 +189,33 @@ async fn read_delta_8_0_table_with_partitions() {
...
+async fn vacuum_delta_8_0_table() {
+    let table = deltalake::open_table("./tests/data/delta-0.8.0")
+        .await
+        .unwrap();
+
+    let retention_hours = 169;
+
+    assert_eq!(
+        table.vacuum_dry_run(retention_hours).unwrap(),
+        vec![format!["./tests/data/delta-0.8.0{}part-00001-911a94a2-43f6-4acb-8620-5e68c2654989-c000.snappy.parquet", MAIN_SEPARATOR]]
```
rtyler, line 203: Since there's already `/` in the path, I'm curious what the objective for the `format!` with `MAIN_SEPARATOR` is?

fvaleye: It seems that the Windows CI test step doesn't like the `/` in the path 😆

houqp: we could use `join_paths` to construct the path here right? https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/mod.rs#L266

rtyler (approved):

Issue #97, opened 2021-03-01
Implement vacuum command


xianwill:
delta-rs should have a "vacuum table" utility analogous to the one provided by the [open source Spark Delta Lake implementation](https://docs.delta.io/latest/delta-utility.html#remove-files-no-longer-referenced-by-a-delta-table). This utility is useful for cleaning up old files that are no longer referenced by the delta log (e.g. files rewritten by merge statements, optimize command etc.).

See the [VacuumCommand](https://github.com/delta-io/delta/blob/40182f3de69ca083afd48ef522897519d36e5fab/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala) in the open source implementation for reference.

rtyler:
@fvaleye I think this is actually done right? I'm not clear what work we have left to do

fvaleye:
> @fvaleye I think this is actually done right? I'm not clear what work we have left to do

Yes, it is already implemented! Hum, we need to improve the tests suite: https://github.com/delta-io/delta-rs/issues/227

MironAtHome:
Как насчет поправить документацию?
Наверное стоит подновить с вот этого
![image](https://user-images.githubusercontent.com/18360184/140049963-8e265274-7faf-4c8c-b70d-325501400845.png)
на вот это?
![image](https://user-images.githubusercontent.com/18360184/140050299-8bddf34b-6994-4be0-953d-200724b3621a.png)

mrk-its:
@rtyler @fvaleye It looks like there are still two serious issues with vacuum implementation:
* vacuum lists all files in dataset using `StorageBackend.list_objs`. The problem is this function returns all files (including these in subdirectories) on `s3` backend and `gcs` backend (althrough I'm not sure about gcs). On `file` and `azure` backends this function  lists only first-level files (without recursing to subdirectories). 
* vacuum ignores files not referenced by delta log at all (so not included on DeltaTableState.files() and DeltaTableState.all_tombstones() lists)

wjones127:
Resolved by #669.

PR #207, opened 2021-04-25
Add Vacuum command in DeltaTable

fvaleye:
# Description
Add Vacuum on the Delta table

# Related Issue(s)
#97

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1,3 +1,5 @@
+//! Delta Table read and write implementation
+
 // Reference: https://github.com/delta-io/delta/blob/master/PROTOCOL.md
 
 use std::cmp::Ordering;
@@ -29,8 +31,10 @@ use super::storage;
 use super::storage::{StorageBackend, StorageError, UriError};
 use uuid::Uuid;
 
+/// Metadata for a checkpoint file
 #[derive(Serialize, Deserialize, Debug, Default, Clone, Copy)]
 pub struct CheckPoint {
+    /// Delta table version
     version: DeltaDataTypeVersion, // 20 digits decimals
     size: DeltaDataTypeLong,
     parts: Option<u32>, // 10 digits decimals
@@ -765,6 +769,8 @@ impl DeltaTable {
         Ok(tombstones)
     }
 
+    /// Return table schema parsed from transaction log. Return None if table hasn't been loaded or
+    /// no metadata was found in the log.
     pub fn schema(&self) -> Option<&Schema> {
         self.state.current_metadata.as_ref().map(|m| &m.schema)
     }
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -610,9 +610,8 @@ impl DeltaTable {
         self.state.min_writer_version
     }
 
-    /// Run the dry run of the Vacuum command on the Delta Table: list files no longer referenced by a Delta table and are older than the retention threshold.
-    /// We do not recommend that you set a retention interval shorter than 7 days, because old snapshots and uncommitted files can still be in use by concurrent readers or writers to the table. If vacuum cleans up active files, concurrent readers can fail or, worse, tables can be corrupted when vacuum deletes files that have not yet been committed.
-    pub fn vacuum_dry_run(&self, retention_hours: u64) -> Result<Vec<String>, DeltaTableError> {
+    /// List files no longer referenced by a Delta table and are older than the retention threshold.
+    fn get_stale_files(&self, retention_hours: u64) -> Result<Vec<String>, DeltaTableError> {
         if retention_hours < 168 {
             return Err(DeltaTableError::InvalidVacuumRetentionPeriod);
         }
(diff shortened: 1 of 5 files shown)


rust/src/delta.rs
```diff
@@ -627,8 +626,45 @@ impl DeltaTable {
...
+        let mut deleted_tombstones = vec![];
+        for tombstone in tombstones {
+            match self
+                .storage
+                .delete_obj(&self.storage.join_path(&self.table_path, &tombstone.path))
+                .await
+            {
+                Ok(_) => deleted_tombstones.push(tombstone.clone()),
+                Err(StorageError::NotFound) => deleted_tombstones.push(tombstone.clone()),
+                Err(_) => continue,
```
houqp, line 658: I feel like we should either propagate the error back to caller immediately or add a `warn`/`error` log here instead of continue without any feedback.

rust/src/delta.rs
```diff
@@ -627,8 +626,45 @@ impl DeltaTable {
...
+            }
+        }
+
+        self.state
+            .tombstones
+            .retain(|tombstone| !deleted_tombstones.contains(tombstone));
+        //TODO: Commit the new state
```
fvaleye, line 665: @houqp `update` the table is the right method to use for updating the DeltaTable state?

rust/src/delta.rs
```diff
@@ -610,9 +610,8 @@ impl DeltaTable {
         self.state.min_writer_version
     }
 
-    /// Run the dry run of the Vacuum command on the Delta Table: list files no longer referenced by a Delta table and are older than the retention threshold.
-    /// We do not recommend that you set a retention interval shorter than 7 days, because old snapshots and uncommitted files can still be in use by concurrent readers or writers to the table. If vacuum cleans up active files, concurrent readers can fail or, worse, tables can be corrupted when vacuum deletes files that have not yet been committed.
-    pub fn vacuum_dry_run(&self, retention_hours: u64) -> Result<Vec<String>, DeltaTableError> {
+    /// List files no longer referenced by a Delta table and are older than the retention threshold.
+    fn gc(&self, retention_hours: u64) -> Result<Vec<&action::Remove>, DeltaTableError> {
```
rtyler, line 614:
```suggestion
    fn get_stale_files(&self, retention_hours: u64) -> Result<Vec<&action::Remove>, DeltaTableError> {
```

`gc` conveys "garbage collection" to me, as in an _active_ collection of garbage like in the JVM.

I propose `get_stale_files` to match more with `get_files` and `get_files_by_partitions`

fvaleye: +1

rust/src/delta.rs
xianwill: not QP level authority - but yes - this is what we use to update table state after committing a new tx as well.

fvaleye: **Updated** - now you are 😄

rust/src/delta.rs
```diff
@@ -627,8 +629,45 @@ impl DeltaTable {
...
+                Ok(_) => deleted_tombstones.push(tombstone.clone()),
+                Err(StorageError::NotFound) => deleted_tombstones.push(tombstone.clone()),
+                Err(err) => return Err(DeltaTableError::StorageError { source: err }),
+            }
+        }
+
+        self.state
+            .tombstones
+            .retain(|tombstone| !deleted_tombstones.contains(tombstone));
+        self.update().await?;
```
houqp, line 668: I don't think we need to update the state here since vacuum doesn't involve any transaction commit?

rust/src/delta.rs
```diff
@@ -627,8 +629,45 @@ impl DeltaTable {
             .get_tombstones()
             .iter()
             .filter(|tombstone| tombstone.deletionTimestamp < delete_before_timestamp)
+            .collect())
```
houqp, line 632: looking at the code in https://github.com/delta-io/delta/blob/34002d3349dcb2cac38fcc44b89d584fffa63751/core/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala#L95, looks the right behavior here should be listing all data files pass retention date in a table, then subtract that with files being tracked in delta tables (including removed files) to get to the final list of files to delete.

fvaleye: You are right! 

Vacuum steps are the following:
1. List all files inside the DeltaTable root path
2. Filter by subtracting the files still in use and tombstones that don't reach the retention threshold
3. Delete the rest

rust/src/delta.rs
fvaleye: Fixed it!

rust/src/delta.rs
```diff
@@ -610,9 +610,11 @@ impl DeltaTable {
...
     }
 
-    /// Run the dry run of the Vacuum command on the Delta Table: list files no longer referenced by a Delta table and are older than the retention threshold.
-    /// We do not recommend that you set a retention interval shorter than 7 days, because old snapshots and uncommitted files can still be in use by concurrent readers or writers to the table. If vacuum cleans up active files, concurrent readers can fail or, worse, tables can be corrupted when vacuum deletes files that have not yet been committed.
-    pub fn vacuum_dry_run(&self, retention_hours: u64) -> Result<Vec<String>, DeltaTableError> {
+    /// List files no longer referenced by a Delta table and are older than the retention threshold.
+    fn get_stale_files(
+        &self,
+        retention_hours: u64,
+    ) -> Result<Vec<&action::Remove>, DeltaTableError> {
```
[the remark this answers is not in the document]
fvaleye: Refactor this, tell me what you think!

rust/src/delta.rs
```diff
@@ -707,6 +706,69 @@ impl DeltaTable {
             .collect::<Vec<String>>())
     }
 
+    /// Whether a path should be hidden for delta-related file operations, such as Vacuum.
+    /// Names of the form partitionCol=[value] are partition directories, and should be
+    /// deleted even if they'd normally be hidden. The _db_index directory contains (bloom filter)
+    /// indexes and these must be deleted when the data they are tied to is deleted.
+    fn is_hidden_directory(&self, path_name: &str) -> Result<bool, DeltaTableError> {
```
rtyler, line 713: The lack of unit tests here make me uneasy, considering the importance not deleting data that shouldn't be deleted :laughing: 

For the purposes of vacuum is it insufficient to just delete files ending in `.parquet` :thinking:

rust/src/delta.rs
```diff
@@ -707,6 +706,69 @@ impl DeltaTable {
...
+                        path_name.starts_with(
+                            &self.storage.join_path(&self.table_path, partition_column),
+                        )
+                    }),
+        )
+    }
+
+    /// Run the Vacuum command on the Delta Table: delete files no longer referenced by a Delta table and are older than the retention threshold.
+    /// We do not recommend that you set a retention interval shorter than 7 days, because old snapshots and uncommitted files can still be in use by concurrent readers or writers to the table. If vacuum cleans up active files, concurrent readers can fail or, worse, tables can be corrupted when vacuum deletes files that have not yet been committed.
+    pub async fn vacuum(
```
rtyler, line 738: This function is clean, nice work!

rust/src/delta.rs
fvaleye: I agree! We should add integration tests to validate the different scenarios.

rtyler (approved):
I will file an issue for testing around vacuum

rust/src/delta.rs
```diff
@@ -707,6 +706,69 @@ impl DeltaTable {
...
+        &mut self,
+        retention_hours: u64,
+        dry_run: bool,
+    ) -> Result<Vec<String>, DeltaTableError> {
+        let tombstones_path = self.get_stale_files(retention_hours)?;
+
+        let mut tombstones = vec![];
+        let mut all_files = self.storage.list_objs(&self.table_path).await?;
+        while let Some(obj_meta) = all_files.next().await {
+            let obj_meta = obj_meta?;
```
houqp, line 748: I think it would be better to strip table path from `obj_meta.path` right after here and only work with object path like how it's done in the scala implementation, this way, we can avoid all the `self.storage.join_path` calls, which can get expensive if we are dealing with 100k+ objects.

rust/src/delta.rs
```diff
@@ -707,6 +706,69 @@ impl DeltaTable {
...
+    }
+
+    /// Run the Vacuum command on the Delta Table: delete files no longer referenced by a Delta table and are older than the retention threshold.
+    /// We do not recommend that you set a retention interval shorter than 7 days, because old snapshots and uncommitted files can still be in use by concurrent readers or writers to the table. If vacuum cleans up active files, concurrent readers can fail or, worse, tables can be corrupted when vacuum deletes files that have not yet been committed.
+    pub async fn vacuum(
+        &mut self,
+        retention_hours: u64,
+        dry_run: bool,
+    ) -> Result<Vec<String>, DeltaTableError> {
+        let tombstones_path = self.get_stale_files(retention_hours)?;
```
houqp, line 743: nitpick, i think `expired_tombstones` would be a better name here so readers won't fix this up with valid tombstones that are not supposed to be deleted.

houqp (commented):
sorry for the late review, I added some performance related suggestions :)

I didn't know there is a `_delta_index` directory, this is pretty cool.

PR #212, opened 2021-04-27
Add dynamodb locking for s3 backend rename calls

mosyp:
# Description

# Related Issue(s)

# Documentation


rtyler:
Sorry @mosyp the rustdocs work that  did has brought some conflicts in that you'll need to resolve here :frowning_face:

houqp (commented):
I think it's better to not bring locking to the main optimistic concurrency commit loop. The commit loop should work as long as rename operation is atomic. See the commit flow documented in https://github.com/delta-io/delta-rs/issues/135. Then we just leave it to each individual storage backend to decide how they want to enforce atomicity for the rename operation.

In fact, cloud storage for filesystem, HDFS, azure and GCP all supports atomic renames natively. S3 is the only special case here. So that's another reason why we want to scope this distributed locking logic into s3 backend's rename method. Even within S3, there are other ways to support multi-writer without using distributed locks, see https://github.com/delta-io/delta-rs/discussions/89#discussioncomment-387261.

mosyp:
@houqp That is reasonable but it conflicts (imo) with. some of the logic in https://github.com/delta-io/delta-rs/discussions/89#discussioncomment-606534, let's move our discussion into that comment section.

P.S. I'll move locking inside s3 rename and after that we'll decide what to do with repairing commits within lock

rust/src/storage/s3/mod.rs
```diff
@@ -315,34 +356,10 @@ impl StorageBackend for S3StorageBackend {
...
 
     async fn rename_obj(&self, src: &str, dst: &str) -> Result<(), StorageError> {
-        debug!("rename s3 object: {} to {}...", src, dst);
-
-        match self.head_obj(dst).await {
-            Ok(_) => return Err(StorageError::AlreadyExists(dst.to_string())),
-            Err(StorageError::NotFound) => (),
-            Err(e) => return Err(e),
+        match self.lock_client {
+            None => self.unsafe_rename_obj(src, dst).await?,
```
xianwill, line 360: @houqp - @mosyp and I were discussing earlier, the combination of this bit and the silent default to `unsafe` makes us very nervous about folks running multiple workers accidentally corrupting the delta log by not providing an env var for `AWS_S3_LOCKING_PROVIDER`. 

I am thinking we should force configuration up front in some way for single or multi worker mode so we can fail fast if the locking provider is not configured in multi worker mode. Do you have any thoughts or preferences on a friendly/discoverable way to do this?

rust/src/storage/s3/mod.rs
```diff
@@ -163,14 +165,53 @@ impl S3StorageBackend {
...
+        let client = create_s3_client(region.clone())?;
+        let lock_client = try_create_lock_client(region)?;
+
+        Ok(Self {
+            client,
+            lock_client,
+        })
+    }
+
+    async fn unsafe_rename_obj(
```
xianwill, line 177: IMHO - I think we should log a `warn!` anytime this method is called.

rust/src/storage/s3/mod.rs
mosyp: We might set AWS_S3_LOCKING_PROVIDER to dynamodb by default, so if users want to use single-worker env they have to specify it explicitly

rust/src/storage/s3/mod.rs
mosyp: I reuse it in both locked/non-locked envs, so i feel like we should find something else here, or even change function name to be less misleading

xianwill: Ah true. Maybe a warn in https://github.com/delta-io/delta-rs/pull/212/files#diff-87e43fff59e072c5dbd99295f15fbf86d5a0e53f6586689191b2c61e07ace91aR417 on startup without a locking provider would be better.

rust/src/storage/s3/mod.rs
xianwill: Nice. I like the idea of defaulting the lock provider and requiring opt-in for single-worker mode. API-wise - I think an opt-in for single-worker mode might should come at a code level as an Option<> to the delta table factory functions. Relying on env vars would still make me feel nervous. Thinking about the difference between reader vs writer semantics is interesting though... I have no idea what the future of multi-reader looks like for delta-rs and whether this option would eventually be useful at a table level for readers.

houqp: yep, +1 for explicit opt in for disabling dynamodb locking in s3 backend. another thing to consider is for read only users, we shouldn't force dynamodb on them. Perhaps by default, we would assuming dynamodb lock for s3, but if dynamodb is not configured, we would just throw an error in all write operations. That way, it still works transparently for read only consumers.

mosyp: yeah, makes sense, will roll out new changes to address your comments guys

houqp: Another annoying thing for configuring backend is we don't really have an option to pass in storage backend specific option at the moment. The only thing that controls how backends are initialized is through uri scheme parsing, which is too limited. We will likely need to refactor that part of the code to make it more flexible.

For now, perhaps it's best to skip support for single writer configuration for now and move on with dynamodb only to unblock the project. We don't need single writer for our own use-cases anyway. Once we get stuff working end to end, we can come back to refactor the code to add single writer support.

rust/src/storage/s3/mod.rs
```diff
@@ -362,13 +379,113 @@ impl StorageBackend for S3StorageBackend {
...
+    dst: &str,
+) -> Result<(), StorageError> {
+    let lock_client = match s3_backend.lock_client {
+        Some(ref lock_client) => lock_client,
+        None => {
+            return Err(StorageError::S3Generic(
+                "dynamodb locking is not enabled".to_string(),
+            ))
+        }
+    };
```
mosyp, line 450: @xianwill @houqp So what I've came up with that is throwing an error when users call rename (used only in commit loop currently) but lock_client is none, which is either dynamodb not configured or its cargo feature turned off

rust/src/storage/s3/mod.rs
mosyp: Single worker will not be possible right now, but I left it as is, e.g. users still need to explicitly configure it, which is ok, since. they have to specify dynamodb table at least. Also I like the idea of moving configs from envs into delta table options somehow, or into _main args_. But this is something that could be done later, once we have our mvp

houqp (approved):
LGTM overall other than the two minor comments.

@mosyp @xianwill @rtyler with the current design, curious what's your thought on removing the dynamodb feature flag and make the dynamodb client a hard dependency as part of the s3 feature. The current design already makes dynamodb optional when its feature flag is enabled, i.e. pure s3 readers and single writer can access s3 tables without setting up dynamodb by not providing the dynamodb config. There are two benefits for doing this:

* Avoids accidental unsafe production write. If a user forgot to compile with dynamodb flag, all dynamodb related config will just be ignored silently without any error or warning. Users might be tricked into thinking they are performing safe concurrent write.
* Reduces code complexity, we can remove a bunch of conditional compilation code and decouple all locking related logic into dynamodb lock module, which can be released as a standalone production ready dynamodb lock crate.

Considering the downside is minor code size increase, this seems like a good tradeoff.

xianwill (approved):

rust/src/storage/s3/mod.rs
```diff
@@ -362,13 +379,113 @@ impl StorageBackend for S3StorageBackend {
...
+    pub lease_duration: u64,
+    /// Tells whether or not the lock was marked as released when loaded from DynamoDB.
+    pub is_released: bool,
+    /// Optional data associated with this lock.
+    pub data: Option<String>,
+    /// The last time this lock was updated or retrieved.
+    pub lookup_time: u128,
+}
+
+///  TODO
```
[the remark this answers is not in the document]
mosyp: to not forget put some docs, thanks :)

rust/src/storage/s3/mod.rs
```diff
@@ -362,13 +379,113 @@ impl StorageBackend for S3StorageBackend {
...
+                }
+            }
+        }
+    }
+    let rename_result = s3_backend.unsafe_rename_obj(src, dst).await;
+    let release_result = lock_client.release_lock(&lock).await;
+
+    rename_result?;
+
+    if !release_result? {
```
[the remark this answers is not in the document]
mosyp: so the code is 
```rust
    let rename_result = s3_backend.unsafe_rename_obj(src, dst).await;
    let release_result = lock_client.release_lock(&lock).await;

    rename_result?;
```

e.g. before unwrapping `rename_result` possible error, I'd like to release a lock, that's the best alternative for `try-finally` I could came up with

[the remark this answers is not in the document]
mosyp: @houqp I'm merging this as is, if you have any other ideas please let me know and I'll address them with new PR

mosyp:
>  Considering the downside is minor code size increase, this seems like a good tradeoff.
@houqp @xianwill 

Since most of the users would like to have safe and distributive writes I feel like this is what we should do. I might address that within my next PR if there's no objections.

AFAIK only @rtyler was asking regarding having it optional way before we started working on it

rust/src/storage/s3/mod.rs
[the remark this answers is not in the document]
houqp: oh, haha, my bad, i misread release_result as rename_result since they kind of looked similar :P

houqp:
yeah, I feel like as long as we make it optional we should be good, i.e. not forcing every s3 users to setup and configure dynamodb.

hustnn:
Is dynamodb compulsory for rename now?

houqp:
@hustnn it's not required if you only have a single writer and don't need concurrent write support.

hustnn:
> @hustnn it's not required if you only have a single writer and don't need concurrent write support.

@houqp I found it throws exception if lock client doesn't exist. I removed it in my local branch https://github.com/hustnn/delta-rs/commit/23836cdaf2fc2120b1d1f4c7e55d87badca6d1a2

houqp:
@hustnn that's not expected, could you send a PR to fix that?

hustnn:
ok

PR #213, opened 2021-04-27
Refactor and clean up the README

rtyler:
This makes the README a bit more clear on what our project status is.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,3 +1,5 @@
+:toc: macro
+
 = delta-rs
 
 image:https://github.com/delta-io/delta-rs/workflows/build/badge.svg[Build Status,link=https://github.com/delta-io/delta-rs/actions]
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -6,20 +8,70 @@ image:https://img.shields.io/badge/docs-rust-blue.svg?style=flat-square[Docs,lin
...
 
-It can also act as the basis for native bindings in other languages such as Python, Ruby or Golang.
+| Filter files with partitions
+| :heavy_check_mark: 
+| :heavy_check_mark: 
+|
+
+| Vacuum (delete stale files)
+| link:https://github.com/delta-io/delta-rs/issues/97[#97]
+|
```
houqp, line 50: we can add the same link for python as well since @fvaleye already added the wrapper, once 97 is merged, python binding will get the fully implementation automatically.

```suggestion
| link:https://github.com/delta-io/delta-rs/issues/97[#97]
```

README.adoc
```diff
@@ -6,20 +8,70 @@ image:https://img.shields.io/badge/docs-rust-blue.svg?style=flat-square[Docs,lin
...
+* Local file system
+* AWS S3
+* Azure Data Lake Storage Gen 2
+* :x: Google Cloud Storage see link:https://github.com/delta-io/delta-rs/issues/56[#56]
+
+.Support features
+|===
+| Operation/Feature | Rust | Python | Ruby
+
+| Read table metadata
```
houqp, line 38: Maybe better to just change this to `Read table` to avoid confusion between delta table metadata attribute, which is different from the files being tracked.

We also support streaming reading in Rust, so we can list that as a separate row if you think it's worth mentioning.

houqp (approved):

README.adoc
rtyler: Please make what suggestions you deem suitable :smile:

README.adoc
```diff
@@ -6,20 +8,70 @@ image:https://img.shields.io/badge/docs-rust-blue.svg?style=flat-square[Docs,lin
...
+| :heavy_check_mark: 
+| :heavy_check_mark: 
+| :heavy_check_mark:
```
houqp, line 41: Maybe something like this:

```suggestion
| Read table
| :heavy_check_mark: 
| :heavy_check_mark: 
| :heavy_check_mark:

| Stream table update
| :heavy_check_mark: 
|
|
```

PR #215, opened 2021-04-28
[Fix] Python table to_pyarrow_dataset: add support for hive style par…

malonsocasas:
…titionning when reading a table

# Description
When reading a partitioned table with `.to_pyarrow_dataset()` method, the partition columns are always set to `null`.

For example, if my table has one parquet file with one row `{value: 14}`: 
`s3://my-bucket/my-database/my-table/year=2020/month=11/day=23/part-1234.parquet`

The following: `DeltaTable("s3://my-bucket/my-database/my-table/").to_pyarrow_dataset()` will return one row with content:
`{ value: 14, year: null, month: null, day: null }`

The current fix makes pyarrow handle the partitioning when creating a dataset.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -3,7 +3,7 @@ from urllib.parse import urlparse
 
 import os
 import pyarrow
-from pyarrow.dataset import dataset
+from pyarrow.dataset import dataset, partitioning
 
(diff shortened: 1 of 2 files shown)


fvaleye:
LGTM! Thanks for fixing it @malonsocasas

fvaleye (approved):

PR #221, opened 2021-04-28
Allow to_pyarrow_table() to take an optional list of partitions

### LINKED ISSUE #220: DeltaTable.to_pyarrow_table() should allow specification of optional partition filters
# Description

**Use Case**

For partitioned tables, it is quite helpful to be able to specify a partitions before generating the Table in memory. Something like the following would be ideal:

```python
dt  = DeltaTable('s3://delta/delta-rs/delta-0.8.0-partitioned')
df = dt.to_pyarrow_table(partitions=['year=2020'])
```

**Related Issue(s)**

rtyler:
This will help ensure that Python is only loading what's necessary

Fixes #220

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -183,13 +183,23 @@ class DeltaTable:
         """
         return pyarrow_schema_from_json(self._table.arrow_schema_json())
 
-    def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
+    def to_pyarrow_dataset(
+        self, partitions: Optional[List[Tuple]] = None
+    ) -> pyarrow.dataset.Dataset:
         """
         Build a PyArrow Dataset using data from the DeltaTable.
 
+        :param partitions: A list of partition filters, see help(DeltaTable.files_by_partitions) for filter syntax
         :return: the PyArrow dataset in PyArrow
         """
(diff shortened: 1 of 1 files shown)


python/deltalake/table.py
```diff
@@ -183,13 +183,15 @@ def pyarrow_schema(self) -> pyarrow.Schema:
         """
         return pyarrow_schema_from_json(self._table.arrow_schema_json())
 
-    def to_pyarrow_dataset(self) -> pyarrow.dataset.Dataset:
+    def to_pyarrow_dataset(self, partitions: Optional[List[Tuple]] = []) -> pyarrow.dataset.Dataset:
```
houqp, line 186: Although we are not mutating the partitions here, it's best to avoid the anti-pattern of using mutable object as the default in python, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

rtyler: I first hit that with Python 2.3, It's been over a decade and they still haven't fixed that bug? :facepalm:

houqp: this is part of the language design unfortunately, so it probably will never get "fixed" :) at least it won't change in python 3 because it will break existing code.

houqp (approved):
LGTM!

Issue #218, opened 2021-04-28
Pyarrow error on data-array-primitives golden table


rtyler:
# Environment

**Delta-rs version**: bd5f431f5a000ccd3d09b4b981fb0c802536403e

**Environment**:
- **Cloud provider**:
- **OS**: Linux/amd64
- **Other**:

***
# Bug

Similar to #217 but with a different error which is interesting. This was with the [data-array-primitives](https://github.com/delta-io/connectors/tree/master/golden-tables/src/test/resources/golden/data-reader-array-primitives) set.

**What happened**:

```python
from deltalake import DeltaTable
from pprint import pprint

if __name__ == '__main__':
    dt = DeltaTable('s3://delta/golden/data-reader-array-primitives')
    print(f'Table version: {dt.version()}')
    for f in dt.files():
        print(f' - {f}')

    df = dt.to_pyarrow_table().to_pandas()
    pprint(df)
```

This lead to:

```
% python query-demo.py
Table version: 0
 - part-00000-182665f0-30df-470d-a5cb-8d9d483ed390-c000.snappy.parquet
 - part-00001-2e274fe7-eb75-4b73-8c72-423ee747abc0-c000.snappy.parquet
Traceback (most recent call last):
  File "query-demo.py", line 10, in <module>
    df = dt.to_pyarrow_table().to_pandas()
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/table.py", line 225, in to_pyarrow_table
    return self.to_pyarrow_dataset().to_table()
  File "pyarrow/_dataset.pyx", line 372, in pyarrow._dataset.Dataset.to_table
  File "pyarrow/_dataset.pyx", line 2266, in pyarrow._dataset.Scanner.to_table
  File "pyarrow/error.pxi", line 122, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 107, in pyarrow.lib.check_status
pyarrow.lib.ArrowTypeError: fields had matching names but differing types. From: as_array_int: list<element: int32> To: as_array_int: list<item: int32>
```

**What you expected to happen**:

I was kind of hoping this would just work :laughing: 

**How to reproduce it**:

**More details**:

fvaleye:
We are using pyarrow.list_ (https://arrow.apache.org/docs/python/generated/pyarrow.list_.html) for building a pyarrow list and the structure is like this: `list<item: int32>`.
[...]

PR #223, opened 2021-04-28
Bump the version of pyarrow needed for Python bindings

fvaleye:
# Description
The version of pyarrow must be >= 4 for the Python bindings

# Related Issue(s)
#218


houqp (approved):
ha, so it's a breaking change from arrow 4.x release?

fvaleye:
> ha, so it's a breaking change from arrow 4.x release?

Yep, because I was with the pyarrow 3.x in my local Python environment and it was the same issue.

rtyler (approved):

Issue #217, opened 2021-04-28
Cannot convert data-reader-map golden table to pyarrow


rtyler:
# Environment

**Delta-rs version**: bd5f431f5a000ccd3d09b4b981fb0c802536403e

**Environment**:
- **Cloud provider**:
- **OS**: Linux/amd64
- **Other**:

***
# Bug

When running the following simple code against the golden data set:'s [data-readers-map](https://github.com/delta-io/connectors/tree/master/golden-tables/src/test/resources/golden/data-reader-map) table

```python
from deltalake import DeltaTable
from pprint import pprint

if __name__ == '__main__':
    dt = DeltaTable('s3://delta/golden/data-reader-map')
    print(f'Table version: {dt.version()}')
    for f in dt.files():
        print(f' - {f}')

    df = dt.to_pyarrow_table().to_pandas()
    pprint(df)
```

**What happened**:

```
% python query-demo.py
Table version: 0
 - part-00000-d9004e55-077b-4728-9ee6-b3401faa46ba-c000.snappy.parquet
 - part-00001-3d30d085-4cde-471e-a396-12af34a70812-c000.snappy.parquet
Traceback (most recent call last):
  File "query-demo.py", line 10, in <module>
    df = dt.to_pyarrow_table().to_pandas()
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/table.py", line 225, in to_pyarrow_table
    return self.to_pyarrow_dataset().to_table()
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/table.py", line 213, in to_pyarrow_dataset
    schema=self.pyarrow_schema(),
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/table.py", line 184, in pyarrow_schema
    return pyarrow_schema_from_json(self._table.arrow_schema_json())
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/schema.py", line 289, in pyarrow_schema_from_json
    arrow_fields = [pyarrow_field_from_dict(field) for field in schema_json["fields"]]
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/schema.py", line 289, in <listcomp>
    arrow_fields = [pyarrow_field_from_dict(field) for field in schema_json["fields"]]
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/schema.py", line 275, in pyarrow_field_from_dict
    pyarrow_datatype_from_dict(field),
[...]

PR #226, opened 2021-04-29
Add valueContainsNull for the MapType in DeltaTable schema

fvaleye:
# Description
Add valueContainsNull for the MapType in DeltaTable schema


rtyler:
Is this for #217 or something different?

fvaleye:
> Is this for #217 or something different?

It's different. I still can't find why it doesn't work for the `pyarrow_schema()` method.
But I faced also these errors when using the same dataset for the `schema()` method. It works now 😄

houqp (approved):

PR #228, opened 2021-04-30
add "element" as the arrow list field name

nevi-me:
# Description

When converting delta array type to arrow list, the arrow list should have a name in its field.
Spark and Parquet use "element" as the field name, thus I have used that.

# Related Issue(s)

None

# Documentation

"element" comes from https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#nested-types


nevi-me:
CC @xianwill

houqp (approved):

Issue #234, opened 2021-05-06
python package metadata not being uploaded to PyPI


houqp:
# Environment

**Delta-rs version**: all

**Binding**: python

***
# Bug

**What happened**:

The metadata section in https://pypi.org/project/deltalake/ is missing all the python package metadata defined in cargo.toml.

Possibly something to do with matruin packaging or upload.

**What you expected to happen**:

Have supported python version to show up i the metadata section.

fvaleye:
It works better now after upgrading the maturin version!

PR #243, opened 2021-05-14
Add requires for maturin in Python bindings

fvaleye:
# Description
Add requires for maturin in Python bindings

# Related Issue(s)
#234


houqp (approved):

houqp:
Thanks for taking care of this @fvaleye !

Issue #45, opened 2020-12-18
read table and file stats


houqp:
collect and calculate table and file stats from transaction log actions and checkpoints

houqp:
partially fixed by https://github.com/delta-io/delta-rs/pull/51. the remaining work is to track and surface file stats through a public interface.

PR #245, opened 2021-05-17
Exposing file stats

rtyler:
# Description

This change will refactors `DeltaTableState` to carry out the `Add` struct from the action module, this will allow the `DeltaTable` or `DeltaTableState` structs to expose stats more readily to their consumers.

# Related Issue(s)

Todo #45


houqp (approved):
should also bump the version to 0.4?

PR #248, opened 2021-05-18
Adding a primitive and high-level writer interface

rtyler:
# Description

This pull request is a work in progress, pending more complete riffing on the [deltalake_ext.rs](https://github.com/delta-io/kafka-delta-ingest/blob/main/src/deltalake_ext.rs).

The objective here is that the primitive API will just handle writing some bytes to storage and the subsequent `add` transaction, whereas the high-level writer API will manage buffering of rows for a parquet file.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -816,6 +816,64 @@ impl DeltaTable {
         DeltaTransaction::new(self, options)
     }
 
+
+    /// Create a new add action and write the given bytes to the storage backend as a fully formed
+    /// Parquet file
+    pub async fn add_file(&mut self, bytes: &Vec<u8>, partitions: Option<Vec<WritablePartitionValue>>) -> Result<i64, DeltaTransactionError> {
+        let path = self.generate_parquet_filename(partitions);
+        let storage_path = self
+            .storage
+            .join_path(&self.table_path, &path);
+
+        debug!("Writing a parquet file to {}", &storage_path);
+        self.storage.put_obj(&storage_path, &bytes)
+            .await
+            .map_err(|source| DeltaTransactionError::Storage { source })?;
+
+        // Determine the modification timestamp to include in the add action - milliseconds since epoch
+        // Err should be impossible in this case since `SystemTime::now()` is always greater than `UNIX_EPOCH`
+        let modification_time = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+        let modification_time = modification_time.as_millis() as i64;
+
+        let add = action::Add {
+            path,
+            size: bytes.len() as i64,
+            partitionValues: HashMap::default(),
+            partitionValues_parsed: None,
+            modificationTime: modification_time,
+            dataChange: true,
+            stats: None,
+            stats_parsed: None,
+            tags: None,
+        };
+
+        let mut tx = self.create_transaction(None);
+        let actions = vec![Action::add(add)];
+        let version = tx.commit_with(&actions, None).await?;
+
+        debug!("Committed Delta version {}", version);
+
+        Ok(version)
+    }
+
+    fn generate_parquet_filename(&self, partitions: Option<Vec<WritablePartitionValue>>) -> String {
+        let mut path_parts = vec![];
+        /*
+         * The specific file naming for parquet is not well documented including the preceding five
+         * zeros and the trailing c000 string
+         *
+         */
+        path_parts.push(format!("part-00000-{}-c000.snappy.parquet", Uuid::new_v4()));
+
+        if let Some(partitions) = partitions {
+            for partition in partitions {
+                path_parts.push(format!("{}={}", partition.name, partition.value));
+            }
+        }
+
+        self.storage.join_paths(&path_parts.iter().map(|s| s.as_str()).collect::<Vec<&str>>())
+    }
+
     /// Create a new Delta Table struct without loading any data from backing storage.
     ///
     /// NOTE: This is for advanced users. If you don't know why you need to use this method, please
@@ -1283,6 +1341,12 @@ pub async fn open_table_with_ds(table_path: &str, ds: &str) -> Result<DeltaTable
 /// Returns rust create version, can be use used in language bindings to expose Rust core version
 pub fn crate_version() -> &'static str {
     env!("CARGO_PKG_VERSION")
+}
+
+/// A partition value for writing files
+pub struct WritablePartitionValue {
+    name: String,
+    value: String,
 }
 
 #[cfg(test)]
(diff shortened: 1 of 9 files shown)


rust/src/writer.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn data(&self) -> Vec<u8> {
+        self.cursor.data()
+    }
+
+    fn close(&mut self) -> Result<parquet_format::FileMetaData, DeltaTableError> {
+        self.writer.close().map_err(|source| DeltaTableError::ParquetError { source })
+    }
+}
+
+struct BufferedJSONWriter {
```
xianwill, line 53: Nice! I like this as a more conventional API to write to for clients.

```suggestion
pub struct BufferedJSONWriter {
```

rust/src/delta.rs
```diff
@@ -1038,6 +1047,7 @@ impl Default for DeltaTransactionOptions {
 #[derive(Debug)]
 pub struct DeltaTransaction<'a> {
     delta_table: &'a mut DeltaTable,
+    actions: Vec<crate::action::Action>,
```
houqp, line 1050: I believe Action has already been imported at the top?
```suggestion
    actions: Vec<Action>,
```

rust/src/delta.rs
```diff
@@ -1048,15 +1058,93 @@ impl<'a> DeltaTransaction<'a> {
...
+
+        // Determine the modification timestamp to include in the add action - milliseconds since epoch
+        // Err should be impossible in this case since `SystemTime::now()` is always greater than `UNIX_EPOCH`
+        let modification_time = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+        let modification_time = modification_time.as_millis() as i64;
+
+        self.actions.push(Action::add(action::Add {
+            path,
+            size: bytes.len() as i64,
+            partitionValues: HashMap::default(),
```
houqp, line 1112: do we need to populate the partition values here as well?

rust/src/writer.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /**
+     * Return the total Values pending in the buffer
+     */
+    pub fn count(&self, partitions: &WriterPartition) -> Option<usize> {
+        if let Some(buffer) = self.buffer.get(&partitions) {
+            return Some(buffer.len());
+        }
+        None
```
houqp, line 59: nit, we can probably simplify this to:
```suggestion
        self.buffer.get(&partitions).map(|b| b.len())
```

rust/src/writer.rs
```diff
@@ -0,0 +1,254 @@
...
+    schema: arrow::datatypes::SchemaRef,
+    partitions: Vec<String>,
+    txns: Vec<Txn>,
+}
+
+impl BufferedJSONWriter {
+    /**
+     * Attempt to construct the BufferedJSONWriter, will fail if the table's metadata is not
+     * present
+     */
```
houqp, line 35: nit, worth considering standardizing all doc string comments to `///` and `//!`, should save you two top and bottom lines as well. this seems to be the norm in other rust projects.

rust/src/writer.rs
```diff
@@ -0,0 +1,254 @@
...
+            let count = self.count(&partitions).unwrap_or(0);
+            let mut value_iter = InMemValueIter::from_vec(&values);
+            let decoder = Decoder::new(self.schema.clone(), count, None);
+
+            let record_batch = decoder
+                .next_batch(&mut value_iter)
+                .map_err(|source| DeltaTableError::ArrowError { source })?;
+
+            if record_batch.is_none() {
+                return Ok(());
```
houqp, line 123: I think this will abort and exit the flush method immediately, is it expected?

rust/src/writer.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        let mut dtx = self.table.create_transaction(None);
+        dtx.add_actions(
+            self.txns
+                .drain(0..)
+                .map(|t| crate::action::Action::txn(t))
+                .collect(),
+        );
+
+        dtx.commit(None).await?;
```
houqp, line 140: we need to add file actions before commit as well?

houqp (commented):
overall looks good 👍

rust/src/writer.rs
rtyler: :rofl: I missed the most important part, `add_file`

rust/src/writer.rs
rtyler: I think this is super tedious and whatever "other rust projects do" doesn't convince me much unless there is a lint to enforce it :smile: 

I'll change it anyways, but I'm registering my protest :smiling_imp:

houqp: haha, I would agree with the tedious claim if you didn't put a `*` at the beginning of every line ;)

houqp (approved):

rust/src/delta.rs
```diff
@@ -1048,15 +1058,100 @@ impl<'a> DeltaTransaction<'a> {
...
+    }
+
+    fn generate_parquet_filename(&self, partitions: Option<Vec<(String, String)>>) -> String {
+        let mut path_parts = vec![];
+        /*
+         * The specific file naming for parquet is not well documented including the preceding five
+         * zeros and the trailing c000 string
+         *
+         */
+        path_parts.push(format!("part-00000-{}-c000.snappy.parquet", Uuid::new_v4()));
```
nfx, line 1138: typical hadoop parquet has a UUID, that is shared between all files in the batch. e.g. this useful in debugging.

instead of hardcoded `00000`, i'd ask to do an incrementing counter.

and instead of hard-coding `snappy`, i'd advice to take it from the compression engine used in `ParquetBuffer .. WriterProperties`.

houqp: @nfx to summarize, this is the scheme we should be using for the partition name right?

`part-{COUNTER}-{BATCH_UUID}-c000.ext`.

Do you know what is the `c000` segment for?

> and instead of hard-coding snappy, i'd advice to take it from the compression engine used in ParquetBuffer .. WriterProperties.

I also agree with this one :+1:

nfx: @houqp c000 might be the bucket number, but I don’t remember from top of my head. Will look once open a laptop next week.

i was writing custom parquet writer 4 years ago and researched it a bit then. It’s referenced from classes referenced in InsertIntoHadoopFsRelation in Apache Spark, for sure.. or at least not too far deep from it.

Main part: file naming won’t mean much for reading and writing (unless bucketed). It’s just for troubleshooting when something goes terribly wrong and you have to manually recover from fs. Probably won’t happen with delta 😜

rtyler: The counter stuff I'll file a ticket for later

PR #252, opened 2021-05-18
Add commit_version method to DeltaTransaction

xianwill:
# Description

This PR adds a `commit_version` method to `DeltaTransaction` which takes a version parameter from the caller. 

Some clients (i.e. https://github.com/delta-io/kafka-delta-ingest) include `txn` actions when creating delta log entries that may require additional conflict resolution with external systems before incrementing the delta log version. The `commit_version` method propagates version conflict errors back to the caller immediately so they may apply conflict resolution steps before attempting another commit.

This PR is complete - but leaving it in a draft status per @rtyler's [notice](https://delta-users.slack.com/archives/C013LCAEB98/p1621350036003900) that we may preemptively avoid some conflicts by incorporating this into [his PR](https://github.com/delta-io/delta-rs/pull/248).

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1212,9 +1212,7 @@ fn log_bytes_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
         jsons.push(json);
     }
 
-    let log_entry = jsons.join("\n");
-
-    Ok(log_entry)
+    Ok(jsons.join("\n"))
 }
 
 fn process_action(
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1056,18 +1056,10 @@ impl<'a> DeltaTransaction<'a> {
         // TODO: create a CommitInfo action and prepend it to actions.
 
         // Serialize all actions that are part of this log entry.
-        let mut jsons = Vec::<String>::new();
-
-        for action in additional_actions {
-            let json = serde_json::to_string(action)?;
-            jsons.push(json);
-        }
-
-        let log_entry = jsons.join("\n");
-        let log_entry = log_entry.as_bytes();
+        let log_entry = log_bytes_from_actions(additional_actions)?;
 
         // try to commit in a loop in case other writers write the next version first
-        let version = self.try_commit_loop(log_entry).await?;
+        let version = self.try_commit_loop(log_entry.as_bytes(), None).await?;
 
         // NOTE: since we have the log entry in memory already,
         // we could optimize this further by merging the log entry instead of updating from storage.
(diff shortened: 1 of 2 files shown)


rust/src/delta.rs
```diff
@@ -1153,6 +1180,19 @@ impl<'a> DeltaTransaction<'a> {
     }
 }
 
+fn log_bytes_from_actions(actions: &[Action]) -> Result<String, serde_json::Error> {
```
rtyler, line 1183: I find it awkward to have a private function just floating in the top level namespace when it's really only something needed/used by the functions within the `DeltaTransaction`. 

I suggest moving this up in to there.

rust/src/delta.rs
```diff
@@ -1153,6 +1180,19 @@ impl<'a> DeltaTransaction<'a> {
...
+    let mut jsons = Vec::<String>::new();
+
+    for action in actions {
+        let json = serde_json::to_string(action)?;
+        jsons.push(json);
+    }
+
+    let log_entry = jsons.join("\n");
+
+    Ok(log_entry)
```
rtyler, line 1193:
```suggestion
    Ok(jsons.join("\n"))
```

rust/src/delta.rs
```diff
@@ -1130,10 +1151,16 @@ impl<'a> DeltaTransaction<'a> {
     async fn try_commit(
         &mut self,
         tmp_log_path: &str,
+        version: Option<DeltaDataTypeVersion>,
```
rtyler, line 1154: I think it's worth making version non-optional here, that way the commit_with could just do:

* Fetch the next attempt version
* Call commit_version

That way the bulk of the work goes into commit_version. That may also obviate the need for the  `log_bytes_with_actions` function as well

rust/src/delta.rs
xianwill: This is a super interesting style point. I always have an internal battle in cases like this. The rule I've been following (privately made rule) up to now has been - "if my fn doesn't have a self ref - it should go into an outer scope since it does not depend on self state" - implication - all struct "methods" should be true methods that have an actual dependency on self state.

I don't disagree with your opinion necessarily, but I'd love to discover a hard rule to base placement decisions like this on rather than a smell which is a bit more ambiguous.

xianwill: If wonder if we were to create a `mod` for `delta_transaction` and this `fn` lived in that `mod` but still out side of the `impl DeltaTransaction` if this might be more palatable to both of us.

rust/src/delta.rs
xianwill: FYI @rtyler - I implemented your suggestion to make `version` non-optional in this method and the implication that `commit_version` doesn't need the optimistic concurrency loop at all and also fixed naming of `log_bytes_from_actions` (it doesnt return bytes lol). However, rather than funnel `commit_with` down to a call of `commit_version`, I left the separation in place so that call stacks now look like:

```
commit_version
log_entry_from_actions
try_commit
```

```
commit_with
log_entry_from_actions
try_commit_loop
try_commit
```

My reasoning here is that `try_commit_loop` shouldn't have to serialize the  same set of log actions on every iteration of the loop until succeeding, so rather than re-use `commit_version`, I think its better if both `commit_version` and `commit_with` serialize the log entry independently and funnel down separate paths to the `try_commit` call.

mosyp (approved):
LGTM, is it still in draft mode or we're ready to merge it?

xianwill:
> LGTM, is it still in draft mode or we're ready to merge it?

@rtyler will be incorporating the changes into https://github.com/delta-io/delta-rs/pull/248. LMK if this becomes a blocker for your work on `kafka-delta-ingest` @mosyp - I think we're likely okay to wait a bit for now.

rtyler:
I think this is fine to bring into `main` as is @xianwill

rust/src/delta.rs
```diff
@@ -1153,6 +1189,17 @@ impl<'a> DeltaTransaction<'a> {
     }
 }
 
+fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Error> {
+    let mut jsons = Vec::<String>::new();
```
houqp, line 1193: you can actually write it in a more efficient way by writing serialized action json directly into a string buffer, that way we can avoid the vector join at the end.

PR #253, opened 2021-05-18
Add get_file_paths_by_partitions in Python bindings

fvaleye:
# Description
Add `get_file_paths_by_partitions()` method in Python bindings


houqp (approved):
you may want to bump python binding version in cargo.toml in this commit as well to prep for the new release.

rtyler (approved):
Nice! I didn't even notice that the Python binding kinda already did a version of this :smile_cat:

PR #256, opened 2021-05-22
FIX #249 Rename camelCased fields to snake_case

### LINKED ISSUE #249: Use #[serde(rename_all = "camelCase")] for structs to avoid camelCase in the Rust actions
# Description

The `action.rs` file allows camelCase instead of the Rust conventional snake case, I suspect because when it was originally written the struct attribute  `#[serde(rename_all = "camelCase")]`

Filing this for later, it should be relatively straightforward but will likely require a fair bit of refactoring of the callsites within delta-rs

nfx:
* This initial commit adds #[serde(rename_all = "camelCase")] for all structs to avoid camelCase in the Rust actions
* StatsParsed is skipped for now and would be fixed in the following commits

This pull request is supposed to address #249


rust/src/action.rs
```diff
@@ -112,22 +112,25 @@ impl ColumnCountStat {
...
+    pub max_values: HashMap<String, ColumnValueStat>,
     /// The number of null values for all columns.
-    pub nullCount: HashMap<String, ColumnCountStat>,
+    pub null_count: HashMap<String, ColumnCountStat>,
 }
 
 /// File stats parsed from raw parquet format.
 #[derive(Debug, Default)]
 pub struct StatsParsed {
+    // somehow #[serde(rename_all = "camelCase")] didn't j=wor just for this one..
```
houqp, line 132: `j=wor` a typo?

houqp (commented):
Thanks for the cleanup!

rust/src/action.rs
nfx: Yes... why I didn’t review line by line before committing...

so with this struct i had a problem - I could not add rename_all here, because compiler was complaining about it being out-of-scope. This is one of those rust features that are not fully loaded into my head yet.

I’ll look into it again later.

nfx:
@houqp comments applied, please review

houqp:
@nfx there are some remaining build issues, could you take a look into them?

nfx:
@houqp there seem to be caching issues on the github side 🤷‍♂️ errors look like a dangling artifact. can you re-trigger this one?..

nfx:
@houqp i think i've found the problem, let's trigger the build again.

houqp (approved):

PR #258, opened 2021-05-22
optimize vacuum operation

houqp:
# Description

Avoid repeatedly creating strings and vectors at runtime.

Turn O(n) path lookup into O(1) with Hashset.

Early exit in stale file check loop.

# Related Issue(s)

Follow up on #207.


houqp:
@fvaleye I noticed we only vaccum data files, not commit logs and indexes, is this expected?

rust/src/delta.rs
```diff
@@ -672,6 +689,15 @@ impl DeltaTable {
         self.get_files_iter().collect()
     }
 
+    /// Returns file names present in the loaded state in HashSet
+    pub fn get_file_set(&self) -> HashSet<&str> {
```
fvaleye, line 693: nitpick: since we are returning file paths, we should rename the method, proposition:
```suggestion
    pub fn get_file_paths_set(&self) -> HashSet<&str> {
```

rust/src/delta.rs
```diff
@@ -765,33 +782,43 @@ impl DeltaTable {
...
         while let Some(obj_meta) = all_files.next().await {
             let obj_meta = obj_meta?;
-            let is_not_valid_file = !self.get_file_paths().contains(&obj_meta.path);
-            let is_valid_tombstone = tombstones_path.contains(&obj_meta.path);
-            let is_not_hidden_directory = !self.is_hidden_directory(&obj_meta.path)?;
-            if is_not_valid_file && is_valid_tombstone && is_not_hidden_directory {
-                tombstones.push(obj_meta.path);
+            let rel_path = self.rel_path(&obj_meta.path)?;
+
+            if valid_files.contains(rel_path) // file is still being tracked in table
```
fvaleye, line 795: 👍 for the performance improvement by using Hashset!

rust/src/delta.rs
```diff
@@ -1287,9 +1315,8 @@ pub fn crate_version() -> &'static str {
 
 #[cfg(test)]
 mod tests {
-    use super::action;
-    use super::action::Action;
-    use super::{process_action, DeltaTableState};
+    use super::*;
+    use pretty_assertions::assert_eq;
```
fvaleye, line 1319: ❤️

rust/src/delta.rs
houqp: file path is defined as `self.storage.join_path(&self.table_path, &add.path` right? this method returns `add.path`.

houqp:
@rtyler @fvaleye @mosyp @xianwill refactored the code to use uri instead of path wherever applicable. This is due to some of the code path is more efficient to work with path directly, for example vacuum path matching. While others are better suited for URIs, for example object read/write. URI is a superset of path, but the current code base has been mixing both of them in various places.

All public methods have been renamed to use uri instead of path as well. Only breaking change is the `table_path` named argument in python's table init method. This should not be a problem is downstream consumers are initializing tables without argument names, which is what we do in all our sample code.

fvaleye (approved):

xianwill (approved):

rtyler:
@houqp ya got some conflicts here, but we can merge once that's resolved

PR #261, opened 2021-05-23
Add initial Rust release CI 

### LINKED ISSUE #238: Automate rust crates release with Github Action
# Description

Same setup with our pypi release, setup a pipeline that will be triggered with tag pushes in the format of `rust-vx.y.z`. The pipeline will run `cargo publish`.

FrancisMurillo:
# Description

Whenever a tag `rust-vx.y.z` (like `rust-v0.4.0`) is pushed, this workflow will publish via `cargo publish`.

GitHub Secrets configuration:
- `CARGO_REPOSITORY_TOKEN` - Token used in `cargo login`

**BLOCKER: Since `cargo publish` requires all dependencies to be published, this will fail until [azure-sdk-for-rust](https://github.com/Azure/azure-sdk-for-rust) is stable or released.**

Also, I noticed in the `rust-v3.0.0/rust/Cargo.lock` file that `azure_core` and `azure_storage` is commented so this seems acknowledged:

```toml
# Azure
# reqwest = { version = "0", optional = true }
# azure_core = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e" }
# azure_storage = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e", features = ["blob", "account", "adls_gen2"] }
```

We could create another `Cargo.release.toml` to hide the `azure` dependencies for the meantime but seems unfulfilling.

# Related Issue(s)

- closes #238 

# Documentation

Links:
- https://doc.rust-lang.org/cargo/reference/publishing.html#github-permissions
- https://github.com/katyo/publish-crates

### DIFF
--- a/.github/workflows/rust_release.yml
+++ b/.github/workflows/rust_release.yml
@@ -1,0 +1,46 @@
+name: Release to Cargo
+
+on:
+  push:
+    tags: [ 'rust-v*' ]
+
+defaults:
+  run:
+    working-directory: ./rust
+
+jobs:
+  validate-release-tag:
+    name: Validate git tag
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: compare git tag with cargo metadata
+        run: |
+          PUSHED_TAG=${GITHUB_REF##*/}
+          CURR_VER=$( grep version Cargo.toml | head -n 1 | awk '{print $3}' | tr -d '"' )
+          if [[ "${PUSHED_TAG}" != "rust-v${CURR_VER}" ]]; then
+            echo "Cargo metadata has version set to ${CURR_VER}, but got pushed tag ${PUSHED_TAG}."
+            exit 1
+          fi
+
+  release-crate:
+    needs: validate-release-tag
+    name: Release crate
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - uses: actions-rs/cargo@v1
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        with:
+          command: publish
+          args: --manifest-path ./rust/Cargo.toml


.github/workflows/rust_release.yml
```diff
@@ -0,0 +1,44 @@
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: publish
+          args: --manifest-path ./rust/Cargo.toml --token ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
houqp, line 44: would be better to inject the token using `CARGO_REGISTRY_TOKEN` instead. I don't know if github action would be smart enough to hide the token from output for us.

rust/Cargo.toml
```diff
@@ -29,8 +29,8 @@ lazy_static = "1"
 
 # Azure
 reqwest = { version = "0", optional = true }
-azure_core = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e" }
-azure_storage = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e", features = ["blob", "account", "adls_gen2"] }
+azure_core = { version = "0.1.0", git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e" }
+azure_storage = { version = "0.1.0", git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e", features = ["blob", "account", "adls_gen2"] }
```
houqp, line 33: We can leave the azure dependencies as is for now. During releases, we will merge main back into the current stable release branch, which has azure dependencies commented off. This should be good enough to unblock the release :)

houqp (commented):
Thanks @FrancisMurillo ! Really appreciate the help :)

.github/workflows/rust_release.yml
FrancisMurillo: I did try it on my fork and it GitHub does hide secrets from the logs:
https://github.com/FrancisMurillo/delta-rs/runs/2652351497?check_suite_focus=true#step:4:1

rust/Cargo.toml
FrancisMurillo: Understood. Reverted the explicit versions from the azure dependencies.

.github/workflows/rust_release.yml
FrancisMurillo: Ah okay... `cargo publish` specifies CARGO_REGISTRY_TOKEN as an option. Removed `--token` option in favor of env

rust/Cargo.toml
```diff
@@ -29,7 +29,7 @@ lazy_static = "1"
 
 # Azure
 reqwest = { version = "0", optional = true }
-azure_core = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e" }
+azure_core = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev =  "536da42ebefd411feff8ba6a0965865e2741267e" }
```
houqp, line 32: nitpick, could you help remove the extra space here to help avoid potential merge conflicts?

rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake"
-version = "0.4.0"
+version = "0.4.11"
```
houqp: curious what's the reason for going from 0 to 11 here?

houqp (approved):

rust/Cargo.toml
FrancisMurillo: Sorry... still testing the env var in GitHub actions... will revert it back to 0.4.0 before finalizing

rust/Cargo.toml
FrancisMurillo: Oops... didn't notice that.

FrancisMurillo:
Okay... the `env` works as well with `--token`
https://github.com/FrancisMurillo/delta-rs/runs/2652574466?check_suite_focus=true#step:4:1

Reverting version to the original version (0.4.0) to mark as finalized

Issue #205, opened 2021-04-19
API to return full table history


houqp:
# Description

**Use Case**

See https://docs.delta.io/0.8.0/delta-utility.html#retrieve-delta-table-history

nfx:
i'll look into it

Smurphy000:
@nfx Are you still working this? I noticed the PR that is up when I was looking for a way to get table history but it has been a few months.

xuky76:
@houqp I think this might be able to be closed out 

# Related issue

- closed in PR #428 

# Documentation
[Docs](https://docs.rs/deltalake/latest/deltalake/delta/struct.DeltaTable.html#method.history)

PR #265, opened 2021-05-24
[DRAFT] Started exposing delta history in scope of #205

nfx:
* Started exposing delta history in scope of #205
* added parsing for CommitInfo
* exposed Vec<CommitInfo> on DeltaTable implementations for Rust & Python
* added some unit tests here and there
* more things and tests to follow in the following commits


rust/src/delta.rs
```diff
@@ -1316,4 +1328,17 @@ mod tests {
...
+
+    macro_rules! aw {
+        ($e:expr) => {
+            tokio_test::block_on($e)
+        };
+      }
+
+    #[test]
+    fn opens_table_with_history() {
+        let mut table = aw!(open_table("tests/data/delta-0.8.0")).unwrap();
```
houqp, line 1340: you should be able to write an async test like this: https://github.com/delta-io/delta-rs/blob/main/rust/tests/read_delta_test.rs#L9-L13

nfx:
current state of PR implementation may not be fully efficient, though. E.g. it would rely on loading CommitInfo's to literally every table. I'll probably refactor tx log into a dedicated `DeltaLog` struct. cc @houqp @rtyler

houqp:
history api has been added in https://github.com/delta-io/delta-rs/pull/428, thanks @nfx for taking a stab at it nonetheless!

PR #271, opened 2021-05-27
Prototype support for Delta Sharing

rtyler:
See [Delta Sharing](https://delta.io/sharing) for more information.

Delta Sharing throws our `StorageBackend` concept out the window a bit since it doesn't act like a primitive storage backend but rather it's almost like the `_delta_log` data is served directly over HTTP.

I think the right place to bring this in is within `DeltaTable`, I'm not expecting we're going to have multiple "things" like this, contrasted to multiple object stores.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2144,6 +2144,7 @@ dependencies = [
  "pin-project-lite",
  "rustls",
  "serde",
+ "serde_json",
  "serde_urlencoded 0.7.0",
  "tokio",
  "tokio-native-tls",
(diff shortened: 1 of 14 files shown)


rtyler:
The work that I still need to do is to implement some errors for attempting to writes or other unsupported operations on a `DeltaTable` with Delta Sharing

rtyler:
Closing this out after the [discussion in Slack](https://delta-users.slack.com/archives/C013LCAEB98/p1622128287073800)

Unfortunately finding the right API layer for supporting Delta Sharing is still up in the air. The Delta Sharing specification omits a _lot_ of the power and flexibility of Delta Lake and at present I'm not terribly interested in building a "little brother" API with less capabilities to `DeltaTable`.

Issue #106, opened 2021-03-03
Implement delta log checkpointing


xianwill:
JSON delta log entries should be consolidated into a parquet snapshot periodically so that log readers can deserialize the log quickly. The spark delta reference implementation [consolidates json log entries into parquet checkpoint snapshots](https://github.com/delta-io/delta/blob/7899c47dd6594128d80db341bcb8d89ef62a9b78/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala) after every 10th JSON log formatted log entry. This behavior is also mentioned under the [Checkpoints section in the delta protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#checkpoints). delta-rs should offer a utility function to create parquet checkpoint snapshots similarly. An MVP implementation should minimally provide a parquet checkpoint utility function so that callers can schedule this appropriately.

PR #280, opened 2021-06-02
Add checkpoint writer

xianwill:
# Description

This PR adds support for writing checkpoints.

This PR is ready with the exception of the missing Arrow Map feature discussed in [this slack thread](https://delta-users.slack.com/archives/C013LCAEB98/p1622663872101400). I would like to merge as is to get the code in and follow up with a fix afterwards.

I haven't implemented an invocation pattern in this PR. I want to take a look at [using S3 event triggers with lambda](https://delta-users.slack.com/archives/C013LCAEB98/p1622163699080100?thread_ts=1622152820.079500&cid=C013LCAEB98) instead of the [design proposed originally](https://github.com/delta-io/delta-rs/discussions/273). 

TBH though, I'd like to leave the invocation to a different PR and merge this one first. If we merge this API, callers may create their own invocation strategy.

For the PR proper, I have a couple of design points I'm especially interested in getting feedback on:

1. I had a hard time deciding where to hang the methods that actually _do_ the checkpoint. The design proposed initially in https://github.com/delta-io/delta-rs/discussions/273 adds them to `DeltaTable`. When it came time to implement, I tucked them under the `CheckPointWriter` instead, cuz I just had a funny smell in my nose. Easily movable and would love opinions from others.
2. I ended up creating this type in `schema.rs` called `DeltaLogSchemaFactory` that merges the table specific schema into common schema fields . I went back and forth in my mind between readable/maintainable code vs efficient code on how to structure this and err'ed on the side of readable/maintainable. My existing implementation is based on the thought that common delta log schema fields are more readable and maintainable if written using the `json` macro and deserialized from there. Also - I struggled with mod placement. I've written this code a few different ways already so it will be easy to restructure if you have strong opinions that make the decision easier.

# Related Issue(s)

* closes https://github.com/delta-io/delta-rs/issues/106
* https://github.com/delta-io/delta-rs/discussions/273
* https://github.com/apache/arrow-rs/issues/395
* https://github.com/apache/arrow-rs/issues/396

**NOTE** I also have this [dirty pre-draft branch](https://github.com/xianwill/delta-rs/compare/checkpoints...xianwill:checkpoints-lambda) I'm working on to add a lambda executable for checkpoints after this PR.


rust/src/delta.rs
```diff
@@ -1355,6 +1455,223 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
...
+
+        debug!("Writing to checkpoint parquet buffer...");
+
+        while let Some(batch) = json_reader.next()? {
+            writer.write(&batch)?;
+        }
+
+        let _ = writer.close()?;
+
+        info!("Finished writing checkpoint file.");
```
mosyp, line 1656: Since it's info, do you mind adding the file name/path to this log

rust/src/delta.rs
```diff
@@ -1355,6 +1455,223 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
...
+            last_checkpoint_path,
+            storage,
+            schema_factory,
+        }
+    }
+
+    /// Creates a new checkpoint at the specified version.
+    /// NOTE: This method loads a new instance of delta table to determine the state to
+    /// checkpoint.
+    pub async fn create_checkpoint_for_version(
```
mosyp, line 1536: I might be missing it, but is there any usage within delta-rs or users are expected to call it from within outside?

rust/src/delta.rs
xianwill: Whoops, that log statement is incorrect actually. The file isn't in scope here. This is just completion of the buffer. Fixed that and changed it to debug. The file info statement is [here](https://github.com/delta-io/delta-rs/pull/280/files#diff-3b5c80af9e5527fbeaa6dad151a835f0733a26e61d0ba0471c58f243a29dfa99R1569).

xianwill:
**NOTE** I also have this [dirty pre-draft branch](https://github.com/xianwill/delta-rs/compare/checkpoints...xianwill:checkpoints-lambda) I'm working on to add a lambda executable for checkpoints after this PR.

rust/src/delta.rs
xianwill: For internal usage - see also https://github.com/delta-io/delta-rs/pull/280#issuecomment-856884265

rust/src/delta.rs
xianwill: <grumble />. apparently that link doesn't work. currently line 1569 of `delta.rs`.

houqp:
sent a PR for reducing clones: https://github.com/xianwill/delta-rs/pull/1

rust/src/schema.rs
```diff
@@ -160,3 +160,487 @@ impl Schema {
...
+                            metadata: HashMap::new(),
+                            r#type: SchemaDataType::r#struct(SchemaTypeStruct {
+                                r#type: "struct".to_string(),
+                                fields: partition_fields.clone(),
+                            }),
+                        };
+                        fields.push(partition_values_parsed);
+                    }
+
+                    if !non_partition_fields.is_empty() {
```
houqp, line 350: looks like we are missing `numRecords` here?

rust/src/schema.rs
```diff
@@ -160,3 +160,487 @@ impl Schema {
...
+/// Factory for creating a Delta log schema for a specific table schema.
+/// REF: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#checkpoint-schema
+pub struct DeltaLogSchemaFactory {
+    common_fields: HashMap<String, Vec<SchemaField>>,
+}
+
+impl DeltaLogSchemaFactory {
+    /// Creates a new DeltaLogSchemaFactory which can be used to create Schema's representing the
+    /// Delta log for specific tables.
+    pub fn new() -> Self {
```
houqp, line 185: it looks like this `new` method only creates static values? If so, we could simplify it by declaring these values two lazy statics values:

* a vector of `SchemaField` structs for all fields other than add. 
* fields for add schemafield

Then we can simplify the code by replacing all of this factory struct with a single `delta_log_schema_for_table` function that does the following:

* clone the lazy static vector of non-add `SchemaField`s
* generate add SchemaFIeld based on provided table schema and partition columns
* append generated add SchemaField to cloned SchemaField vector

rust/src/delta.rs
```diff
@@ -1355,6 +1455,223 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
...
+            size,
+            parts: None,
+        };
+
+        let file_name = format!("{:020}.checkpoint.parquet", version);
+        let checkpoint_path = self.storage.join_path(&self.delta_log_path, &file_name);
+
+        info!("Writing checkpoint to {:?}.", checkpoint_path);
+        self.storage
+            .put_obj(&checkpoint_path, &parquet_bytes)
```
houqp, line 1571: for naming consistency sake, since https://github.com/delta-io/delta-rs/pull/258#issuecomment-850925010, we use uri to refer to object reference that contains both scheme,  host and path, for example `s3://bucket/key`. path suffix is used for variables that only hosts the path part. For example `key`.

So I think a better name here would be `checkpoint_uri`.

rust/src/delta.rs
```diff
@@ -1355,6 +1455,223 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
...
+
+        self.create_checkpoint_from_state(version, table.get_state())
+            .await
+    }
+
+    /// Creates a new checkpoint at the specified version from the given DeltaTableState.
+    pub async fn create_checkpoint_from_state(
+        &self,
+        version: DeltaDataTypeVersion,
+        state: &DeltaTableState,
```
houqp, line 1550: The version, state and fields in self here is basically all the fields in the `DetalTable` struct, I feel like it would be simpler if we just make this a function that takes a `DeltaTable` struct reference as input or a method of the `DeltaTable` struct. This way, we won't need to define a dedicated `CheckPointWriter` struct anymore.

Or are you thinking that there is a valid use-case where one might want to manually construct delta table states and versions without through through `DeltaTable` struct?

rust/src/delta.rs
```diff
@@ -1355,6 +1455,223 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
...
+            .put_obj(
+                self.last_checkpoint_path.as_str(),
+                last_checkpoint_content.as_bytes(),
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    fn parquet_bytes_from_state(
```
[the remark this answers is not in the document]
xianwill: This is really slick and exactly what I wanted. The chaining is fighting me a bit on type mismatch. I'll play a bit more but might yell for help.

rust/src/schema.rs
```diff
@@ -160,3 +160,487 @@ impl Schema {
...
+        // TODO: map<string, string> is not supported by arrow currently.
+        // See:
+        // * https://github.com/apache/arrow-rs/issues/395
+        // * https://github.com/apache/arrow-rs/issues/396
+
+        let meta_data_fields = json!([
```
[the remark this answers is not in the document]
xianwill: heh. in fact, i wrote this out with `SchemaField`s initially, but I felt like it was quite noisy. Using `json!` macro it reads just like the docs in delta protocol spec and I thought that was nicer for maintainability. I am happy to change back if you think it is not worth the deserialization.

rust/src/delta.rs
xianwill: I was thinking about a use case where an application might want to run data writes and checkpoints on separate threads. This was actually related to my initial design in https://github.com/delta-io/delta-rs/discussions/273 that included a channel for writing checkpoints from a separate thread in the same process while leaving the table instance available for further write actions. Theoretically, one might have a primary write thread, after every 10th commit, clone the state, send it as part of the message to the channel and keep using the same table on the primary write thread. I think we are going to use lambda checkpoints which are decoupled at the process level at this point though, so it is not as important to maintain this API. To your point though, even with the design in https://github.com/delta-io/delta-rs/discussions/273, we could clone the `DeltaTable`. Idk - given that thought process, whats your feedback?

houqp: Ha makes sense. Now I see where you are coming from. Based on this, I am cool with this design. The idea of cloning `DeltaTable` also sounds like a good idea. I will leave that decision to you :)

rust/'
```diff
@@ -0,0 +1,324 @@
+//! Implementation for writing delta checkpoints.
```
houqp, line 1: this filename exposed that you are a vim user :D

rust/src/schema.rs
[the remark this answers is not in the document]
houqp: yeah, I think it's not worth incurring the extra runtime overhead here. i feel you that creating schemas that way manually using arrow api requires a lot of boilerplates. I think this usability issue can be addressed by introducing better apis in upstream arrow crates.

rust/src/delta.rs
xianwill: I thought about this a little more, and for public consumption, I prefer the two methods we have. We have an api that takes a version parameter, and opens a delta table at that version internally. We also have an api that takes version and state in case the caller wants to pass in memory state that they have in hand already. I like that these two APIs are very explicit about exactly what version and state will be checkpointed. An API that takes a DeltaTable would use the version and state on the DeltaTable instance so it does not communicate the content of the checkpoint that will be created as explicitly IMHO.

rust/'
xianwill: lol whoops. :flushed:

rust/src/delta.rs
xianwill: fixed. also noticed a couple other uris i was calling paths and fixed those.

rust/src/schema.rs
xianwill: this relates to https://github.com/delta-io/delta-rs/pull/280#discussion_r647880588 as well. i moved the schema creation into `delta_arrow.rs`, skipped both json and delta schema entirely and went straight to arrow and included the lazy statics as recommended.

rust/src/schema.rs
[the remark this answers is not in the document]
xianwill: got it. fixed. see comment in https://github.com/delta-io/delta-rs/pull/280#discussion_r649263222.

rust/src/delta.rs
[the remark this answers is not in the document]
xianwill: fixed. thanks for pairing with me on this one. lots of fun. hopefully we both learned something. i know i did.

rust/src/delta.rs
```diff
@@ -1355,6 +1455,223 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
     Ok(jsons.join("\n"))
 }
 
+/// Error returned when the CheckPointWriter is unable to write a checkpoint.
+#[derive(thiserror::Error, Debug)]
+pub enum CheckPointWriterError {
```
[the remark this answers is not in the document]
xianwill: moved to `checkpoints.rs`

rust/src/delta.rs
```diff
@@ -279,6 +319,61 @@ struct DeltaTableState {
...
+    pub fn new(
+        tombstones: Vec<action::Remove>,
+        files: Vec<action::Add>,
+        commit_infos: Vec<Value>,
+        app_transaction_version: HashMap<String, DeltaDataTypeVersion>,
+        min_reader_version: i32,
+        min_writer_version: i32,
+        current_metadata: Option<DeltaTableMetaData>,
+    ) -> Self {
+        Self {
```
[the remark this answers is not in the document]
xianwill: removed.

rust/src/schema.rs
```diff
@@ -160,3 +160,487 @@ impl Schema {
...
+                    "keyType": "string",
+                    "valueType": "string",
+                    "valueContainsNull": true,
+                },
+                "nullable": true,
+                "metadata": {},
+            }*/
+        ]);
+
+        let remove_fields = json!([
```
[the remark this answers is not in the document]
xianwill: fixed

rust/src/schema.rs
```diff
@@ -160,3 +160,487 @@ impl Schema {
...
+            { "name": "minReaderVersion", "type": "integer", "nullable": true, "metadata": {} },
+            { "name": "minWriterVersion", "type": "integer", "nullable": true, "metadata": {} }
+        ]);
+
+        let txn_fields = json!([
+            { "name": "appId", "type": "string", "nullable": true, "metadata": {} },
+            { "name": "version", "type": "long", "nullable": true, "metadata": {} }
+        ]);
+
+        let add_fields = json!([
```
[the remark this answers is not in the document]
xianwill: fixed

rust/src/schema.rs
xianwill: fixed

rust/src/delta.rs
```diff
@@ -1355,6 +1455,223 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, serde_json::Erro
...
+            .await
+    }
+
+    /// Creates a new checkpoint at the specified version from the given DeltaTableState.
+    pub async fn create_checkpoint_from_state(
+        &self,
+        version: DeltaDataTypeVersion,
+        state: &DeltaTableState,
+    ) -> Result<(), CheckPointWriterError> {
+        // TODO: checkpoints _can_ be multi-part... haven't actually found a good reference for
```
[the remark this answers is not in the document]
xianwill: Filed https://github.com/delta-io/delta-rs/issues/288

rust/src/schema.rs
```diff
@@ -160,3 +160,487 @@ impl Schema {
...
+        let log_schema = factory
+            .delta_log_schema_for_table(&table_schema, partition_columns.as_slice())
+            .unwrap();
+
+        assert_eq!("struct", log_schema.r#type);
+        assert_eq!(5, log_schema.get_fields().len());
+
+        for f in log_schema.get_fields().iter() {
+            match f.get_name() {
+                "txn" => {
```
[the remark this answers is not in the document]
xianwill: Filed https://github.com/delta-io/delta-rs/issues/287

rust/src/checkpoints.rs
```diff
@@ -0,0 +1,196 @@
...
+            .current_metadata()
+            .ok_or(CheckPointWriterError::MissingMetaData)?;
+        let jsons: Vec<Result<serde_json::Value, ArrowError>> = vec![
+            action::Action::protocol(action::Protocol {
+                min_reader_version: state.min_reader_version(),
+                min_writer_version: state.min_writer_version(),
+            }),
+            action::Action::metaData(action::MetaData::try_from(current_metadata.clone())?),
+        ]
+        .into_iter()
```
houqp, line 153: I was able to get `once` working using the following code:

```diff
-        let jsons: Vec<Result<serde_json::Value, ArrowError>> = vec![
-            action::Action::protocol(action::Protocol {
-                min_reader_version: state.min_reader_version(),
-                min_writer_version: state.min_writer_version(),
-            }),
-            action::Action::metaData(action::MetaData::try_from(current_metadata.clone())?),
-        ]
-        .into_iter()
+        let jsons = std::iter::once(action::Action::protocol(action::Protocol {
+            min_reader_version: state.min_reader_version(),
+            min_writer_version: state.min_writer_version(),
+        }))
+        .chain(std::iter::once(action::Action::metaData(
+            action::MetaData::try_from(current_metadata.clone())?,
+        )))
```

houqp (commented):
the rest looks good to me 👍

rust/src/delta.rs
```diff
@@ -40,6 +40,21 @@ pub struct CheckPoint {
     parts: Option<u32>, // 10 digits decimals
 }
 
+impl CheckPoint {
+    /// Creates a new checkpoint from the given parameters.
+    pub(crate) fn new(
```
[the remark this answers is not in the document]
xianwill: the struct fields on CheckPoint are private and i moved the writer to a separate mod so i needed this to construct it.

rust/src/checkpoints.rs
```diff
@@ -0,0 +1,196 @@
...
+                .map(|(app_id, version)| {
+                    action::Action::txn(action::Txn {
+                        app_id: app_id.clone(),
+                        version: *version,
+                        last_updated: None,
+                    })
+                }),
+        )
+        .map(|a| serde_json::to_value(a).map_err(ArrowError::from))
+        .collect();
```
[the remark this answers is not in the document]
xianwill: fixed

rust/src/checkpoints.rs
xianwill: fixed. super cool - i learned a lot about iterators thanks to this series of comments. :bow:

houqp (approved):

houqp (approved):

PR #289, opened 2021-06-12
Disable default features for rusoto

rtyler:
This allows clients to use rustls or native-tls. In my case I need to use rustls
so that I can build Lambdas more easily

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -57,7 +57,7 @@ async-trait = "0.1"
 rust-dataframe-ext = []
 datafusion-ext = ["datafusion", "crossbeam"]
 azure = ["azure_core", "azure_storage", "reqwest"]
-s3-native-tls = ["rusoto_core/native-tls", "rusoto_credential", "rusoto_s3/native-tls", "rusoto_sts/native-tls", "rusoto_dynamodb/native-tls", "maplit"]
+s3 = ["rusoto_core/native-tls", "rusoto_credential", "rusoto_s3/native-tls", "rusoto_sts/native-tls", "rusoto_dynamodb/native-tls", "maplit"]
 s3-rustls = ["rusoto_core/rustls", "rusoto_credential", "rusoto_s3/rustls", "rusoto_sts/rustls", "rusoto_dynamodb/rustls", "maplit"]
 
 [build-dependencies]
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -33,11 +33,11 @@ azure_core = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = t
 azure_storage = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "536da42ebefd411feff8ba6a0965865e2741267e", features = ["blob", "account", "adls_gen2"] }
 
 # S3
-rusoto_core = { version = "0.46", optional = true }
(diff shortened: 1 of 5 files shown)


rust/Cargo.toml
```diff
@@ -57,7 +57,8 @@ async-trait = "0.1"
 rust-dataframe-ext = []
 datafusion-ext = ["datafusion", "crossbeam"]
 azure = ["azure_core", "azure_storage", "reqwest"]
-s3 = ["rusoto_core", "rusoto_credential", "rusoto_s3", "rusoto_sts", "rusoto_dynamodb", "maplit"]
+s3-native-tls = ["rusoto_core/native-tls", "rusoto_credential", "rusoto_s3/native-tls", "rusoto_sts/native-tls", "rusoto_dynamodb/native-tls", "maplit"]
```
houqp, line 60: how about leaving `s3` feature to native-tls to keep it backwards compatible?

rtyler: :shrug: I don't have a strong opinion either way, but when did we start caring bout backwards compatibility? :laughing:

houqp: In this case, it's just a simple one line change from our end to avoid unnecessary surprises downstream :) This should also help fix all the builds as well.

houqp (approved):

houqp:
@rtyler could we squash intermediate commits before merge going forward? Intermediate broken commits makes it harder to do the following :)

* trace down reasons or context for a specific code change using git blame. When a logical self contained change is broken up into multiple intermediate commits, git blame will only get you one of the commits. Having all related changes contained in a single commit with well written commit message has saved my ass many times when working on complex projects.
* trace down bugs using bisect
* revert a logically self-contained feature/bugfix
* draft changelog based on commit history

I understand the need for keeping rebase merge enabled in github because sometimes one might want to merge multiple logically self contained changes in a single PR. In that case, the PR author should manually squash related commits after the review is done and before the rebase merge.

Here are some examples of intermediate commits in our main branch tree:

* https://github.com/delta-io/delta-rs/commit/debaeced6fdd2a8709c5d8f3209651fa7ba5be45
* https://github.com/delta-io/delta-rs/commit/d760449d6d59af593982dee92d2a76f48e7529e5
* commits around https://github.com/delta-io/delta-rs/commit/6c95b60541c792ec3c63b7a4dd4f4d7186eee2fc are all for adding transaction commit support
* https://github.com/delta-io/delta-rs/commit/21aa8912f239d0c0ed9741eb7e1144b57cbc44eb

rtyler:
@houqp If the objective is to ensure that no single commit cannot build (across what spectrum of our tests?) then I think we need to have a much broader discussion with the other committers since that has a pretty big impact on development.

houqp:
yes, let's discuss this in our next bi-weekly sync :) The goal for me is less about enforcing that every single commit has to build and pass all tests, but more about avoid unnecessary intermediate commits to make the commit history clean, easy to follow and revert if needed. 

For example, running git blame on https://github.com/delta-io/delta-rs/blob/d922b311b4ff697f5e0c964ddc7572324d2d3c01/python/src/lib.rs#L103 gives me a commit with commit message `Integrate the review comments` without any context on why this change is needed and what are the other related commits that I need to look into.

I would be curious what impact this would have to other committers. As long as PRs are self-contained, which is the case most of the time, the only change would be clicking the squash merge instead of rebase merge button.

PR #291, opened 2021-06-13
Fix generated parquet filename

viirya:
# Description

The current parquet filename is appended with partition values, e.g. `part-0000-xxxx-c000.snappy.parquet/part1=value1/part2=value2`. Seems it doesn't follow existing conventions as `part1=value1/part2=value2/...` .

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1287,13 +1287,15 @@ impl<'a> DeltaTransaction<'a> {
          * zeros and the trailing c000 string
          *
          */
-        let mut path_parts = vec![format!("part-00000-{}-c000.snappy.parquet", Uuid::new_v4())];
+        let mut path_parts = vec![];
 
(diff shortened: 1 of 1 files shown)


houqp (approved):
Good catch, thanks for adding tests as well!

houqp:
@viirya looks like there are minor formatting errors: https://github.com/delta-io/delta-rs/pull/291/checks?check_run_id=2815237004, could you format the code with https://github.com/rust-lang/rustfmt?

viirya:
Thanks @houqp. Fixed the format.

houqp (approved):

viirya:
Oh, seems the test fails on windows because of path separator. I will fix it.

houqp:
Turns out our auto merge setting allows merge with broken windows CI. can't tell if that's a feature or a bug :)

PR #292, opened 2021-06-13
Fix path check of parquet_filename

viirya:
# Description

Newly added test `parquet_filename` failed on windows-2019 CI:

---- delta::tests::parquet_filename stdout ----
thread 'delta::tests::parquet_filename' panicked at 'assertion failed: parquet_filename.contains(\"col1=a/col2=b/part-00000-\")', rust\src\delta.rs:1610:9

This tries to fix it.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1607,6 +1607,10 @@ mod tests {
             (String::from("col2"), String::from("b")),
         ];
         let parquet_filename = txn.generate_parquet_filename(Some(partitions));
-        assert!(parquet_filename.contains("col1=a/col2=b/part-00000-"));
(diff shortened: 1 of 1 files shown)


viirya:
Actually I'm not sure if path separator is the issue on windows-2019. Because I checked `join_paths`, it do `.join("/")` on the path elements. So seems it should simply return a string `col1=a/col2=b/part-00000...` no matter on windows or not.

Wait for CI.

viirya:
The CI passed. cc @houqp

houqp (approved):

viirya:
Thanks @houqp!

PR #293, opened 2021-06-14
Add lambda function for checkpoints

xianwill:
# Description

Add a lambda function for creating delta checkpoints,

@rtyler + @houqp - Is this is an appropriate package to include in delta-rs?

# Documentation

[delta-checkpoint README](https://github.com/xianwill/delta-rs/blob/checkpoint-lambda/aws/delta-checkpoint/README.md)


aws/delta-checkpoint/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "delta-checkpoint"
```
houqp, line 2: i would add `lambda` as suffix or prefix to the name unless we plan to add other types of checkpoint daemon in the same package in the future.

houqp (approved):
really cool. looks good to me, I am cool with keeping this in the delta-rs repo.

houqp (approved):

aws/delta-checkpoint/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+async fn func(event: Value, _: Context) -> Result<(), CheckPointLambdaError> {
+    process_event(&event).await
+}
+
+async fn process_event(event: &Value) -> Result<(), CheckPointLambdaError> {
+    let (bucket, key) = bucket_and_key_from_event(&event)?;
+    let (path, version) = table_path_and_version_from_key(key.as_str())?;
+    let table_uri = table_uri_from_parts(bucket.as_str(), path.as_str())?;
+
+    if version % 10 == 0 && version != 0 {
```
rtyler, line 37: I assume this is just to replicating the behavior in the Spark checkpoint writer? Assuming so, I think a comment along those lines here, and a mention in the README would be beneficial

aws/delta-checkpoint/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    let func = handler_fn(func);
+    lambda_runtime::run(func).await?;
+    Ok(())
+}
+
+async fn func(event: Value, _: Context) -> Result<(), CheckPointLambdaError> {
+    process_event(&event).await
```
rtyler, line 29: Since this is propagating the `CheckPointLambdaError`up to the lambda runtime caller, will that actually put an error message out into the Lambda logs? 

I think it would be good to check, otherwise failures will be completely opaque to the user.

Also, since there is no (pretty) env logger init here, there's no way that i can think of for the log messages to actually get to the Lambda output logs.

aws/delta-checkpoint/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+async fn process_event(event: &Value) -> Result<(), CheckPointLambdaError> {
```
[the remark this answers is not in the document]
xianwill: Yeah, looking at the event structure, there are a ton of fields that I don't need. I figured just plucking the JSON out would be faster to avoid the deserialization. I'm happy to change this if you think its advisable.

aws/delta-checkpoint/src/main.rs
xianwill: Ahhh. Errors do show, but my other logs were not. Ty for the pointer to pretty_env_logger. Fixed.

aws/delta-checkpoint/src/main.rs
xianwill: Comment added

```
    // Checkpoints are created for every 10th delta log commit.
    // Follows the reference implementation described in the delta protocol doc.
    // https://github.com/delta-io/delta/blob/master/PROTOCOL.md#checkpoints.

```

houqp (approved):

rtyler (approved):

PR #294, opened 2021-06-20
Added .to_pandas to deltalake python

bramrodenburg:
# Description
Many users that work in Python use pandas in their daily work. Adding a `.to_pandas` method makes it super easy for users to read in a Delta table into a pandas dataframe. 

# Related Issue(s)
-

# Documentation
- Added documentation to the `.to_pandas` method. Will update the docs if the proposed change looks fine.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -1,10 +1,11 @@
 import os
 import warnings
 from dataclasses import dataclass
-from typing import Any, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, List, Optional, Tuple
 from urllib.parse import urlparse
 
-import pandas as pd
+if TYPE_CHECKING:
+    import pandas as pd
 import pyarrow
 from pyarrow.dataset import dataset, partitioning
 
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from typing import Any, List, Optional, Tuple
 from urllib.parse import urlparse
 
+import pandas as pd
 import pyarrow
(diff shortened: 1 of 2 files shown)


python/deltalake/table.py
```diff
@@ -4,6 +4,7 @@
 from typing import Any, List, Optional, Tuple
 from urllib.parse import urlparse
 
+import pandas as pd
```
houqp, line 7: pandas is an optional dependency, this would break some of the downstream code, we need to wrap this import with try except.

bramrodenburg: Mmm.. though you would still need it for the type checking (`def to_pandas() -> pd.DataFrame`) and in the unit tests. Would you have any suggestions how to incorporate that?

houqp: that's a good call, have you tried writing it out as a string: `"pd.DataFrame"`? mypy should still be smart enough to pick it up.

bramrodenburg: Tried that out, but `mypy` still complains because pandas is not available during the lint check. Added a `#ignore` to suppress it for now.

houqp:
Ignoring the type check there works for me :) I will merge this after https://github.com/delta-io/delta-rs/pull/296 to address the new clippy error.

houqp (approved):

adityamcodes:
So if we're sourceing from a delta table in a bronze table in delta format and then converting to panda, using pandas transforms and reverting back to delta lake for write into a silver table (for example), does that delta table still store & retain the pandas transforms in the history of the delta table?

wjones127:
@adityamcodes To ask a question, could you please open an issue or discussion rather than commenting on an old pull request?

PR #295, opened 2021-06-22
Make dynamodb lease_duration optional to support non-expirable locks

mosyp:
# Description
This PR brings the support of non-expirable locks. Such behaviour might be desired by external systems which interact with concurrent writer, e.g. in order to pause them etc. However there's the risk of delta-rs writer might expire such lock. With these changes, when the active lock has lease_duration as None, then the lock client treats such lock as non expirable and will not expire it. 

I've also introduced an optional `do_not_timeout_on_non_expirable_locks` config with default to `true`. If `true` then the acquire lock function will not fail with timeout when its reached so (the timeout is config's `lease_duration` + `additional_time_to_wait_for_lock`)

### DIFF
--- a/rust/src/storage/s3/dynamodb_lock.rs
+++ b/rust/src/storage/s3/dynamodb_lock.rs
@@ -77,7 +77,12 @@ impl LockItem {
         if self.is_released {
             return true;
         }
-        now_millis() - self.lookup_time > (self.lease_duration as u128) * 1000
+        match self.lease_duration {
+            None => false,
+            Some(lease_duration) => {
+                now_millis() - self.lookup_time > (lease_duration as u128) * 1000
+            }
+        }
     }
 }
 
@@ -296,11 +301,17 @@ impl DynamoDbLockClient {
             .await?;
 
         if let Some(item) = output.item {
-            let lease_duration = get_string(item.get(LEASE_DURATION))?
-                .parse::<u64>()
-                .map_err(|_| DynamoError::InvalidItemSchema)?;
+            let lease_duration = {
+                match item.get(LEASE_DURATION).and_then(|v| v.s.clone()) {
+                    None => None,
+                    Some(v) => Some(
+                        v.parse::<u64>()
+                            .map_err(|_| DynamoError::InvalidItemSchema)?,
+                    ),
+                }
+            };
 
-            let data = item.get(DATA).and_then(|r| r.s.as_ref()).cloned();
+            let data = item.get(DATA).and_then(|r| r.s.clone());
 
             return Ok(Some(LockItem {
                 owner_name: get_string(item.get(OWNER_NAME))?,
@@ -400,7 +411,7 @@ impl DynamoDbLockClient {
         Ok(LockItem {
             owner_name: self.opts.owner_name.clone(),
             record_version_number: rvn,
-            lease_duration: self.opts.lease_duration,
+            lease_duration: Some(self.opts.lease_duration),
             is_released: false,
             data: data.map(String::from),
             lookup_time: now_millis(),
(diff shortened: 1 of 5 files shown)


rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -439,8 +462,14 @@ struct AcquireLockState<'a> {
 }
 
 impl<'a> AcquireLockState<'a> {
+    /// If lock is expirable (lease_duration is set) and is do_not_timeout_on_non_expirable_locks
+    /// then this function returns `true` if the elapsed time sine `started` is reached `timeout_in`.
     fn has_timed_out(&self) -> bool {
-        self.started.elapsed() > self.timeout_in
+        self.started.elapsed() > self.timeout_in && {
+            let is_non_expirable = self.cached_lock.is_some()
+                && self.cached_lock.as_ref().unwrap().lease_duration.is_none();
```
houqp, line 470: have you tried this?

```rust
let non_expirable = if let Some(cached_lock) = self.cached_lock {
    cached_lock.lease_duration.is_none()
} else {
    false
};
```

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -42,6 +45,8 @@ pub struct Options {
     pub refresh_period: Duration,
     /// The amount of time to wait in addition to `lease_duration`.
     pub additional_time_to_wait_for_lock: Duration,
+    /// If set, then `try_acquire_lock` will not fail with timeout if the active lock is non expirable.
+    pub do_not_timeout_on_non_expirable_locks: bool,
```
houqp, line 49: under what circumstance would one want to set this to false? i am thinking whether we can just get rid of this option entirely if we don't have a use-case for it at the moment.

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -400,7 +423,7 @@ impl DynamoDbLockClient {
         Ok(LockItem {
             owner_name: self.opts.owner_name.clone(),
             record_version_number: rvn,
-            lease_duration: self.opts.lease_duration,
+            lease_duration: Some(self.opts.lease_duration),
```
houqp, line 426: so upsert a nonexpirable lock will make it expirable? what's the use-case for this?

rust/src/storage/s3/dynamodb_lock.rs
mosyp: Makes sense, let's not over engineer here

rust/src/storage/s3/dynamodb_lock.rs
mosyp: Yes, I figured that this one results on more code, however if that's preferable convention I'll stick with that

rust/src/storage/s3/dynamodb_lock.rs
mosyp: There's no use case for non-expirable lock in delta-rs. I was thinking of adding that, e.g `acquire_non_expirable_lock` but since we don't have a use case for that then left it as is

rust/src/storage/s3/dynamodb_lock.rs
houqp: yeah, i think this is slightly better because it avoids a redundant enum match.

mosyp: Oh I see, ok

rust/src/storage/s3/dynamodb_lock.rs
houqp: got it, I am mostly concerned about a future client accidentally invoked one of the upsert method to corrupt the lock.

how does KDI acquire a nonexpirable lock to block writes from spark in this case?

mosyp: > concerned about a future client accidentally invoked one of the upsert method to corrupt the lock.

It's private so should not be possible. And other functions that calls it will succeed unless the lock is owned.

> how does KDI acquire a nonexpirable lock to block writes from spark in this case?

It's not, we plan to do that within proprietary code

houqp: this looks good based on our offline discussion, looks like it is not going cause problem because optimize will just ignore the lease attribute entirely and never acquire an expired lock from its end.

xianwill (approved):

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -439,8 +450,17 @@ struct AcquireLockState<'a> {
...
+    /// if the elapsed time sine `started` is reached `timeout_in`.
     fn has_timed_out(&self) -> bool {
-        self.started.elapsed() > self.timeout_in
+        self.started.elapsed() > self.timeout_in && {
+            let non_expirable = if let Some(ref cached_lock) = self.cached_lock {
+                cached_lock.lease_duration.is_none()
+            } else {
+                false
+            };
+            !non_expirable
```
houqp, line 462: I would avoid the double negative here to make the code more readable:

```suggestion
            let expirable = if let Some(ref cached_lock) = self.cached_lock {
                cached_lock.lease_duration.is_some()
            } else {
                true
            };
            expirable
```

houqp (approved):

PR #297, opened 2021-06-22
fix python docs build

houqp:
also extend typechecks to python tests


houqp:
looks like our localstack s3 integration test setup is broken :(

mosyp (approved):

PR #298, opened 2021-06-29
[fix] s3 code not compiled when using s3-rustls

rdettai:
# Description
When using the s3-rustls feature instead of s3, the s3 backend module is not compiled.

# Reproduce
Import deltalake with:
`deltalake = { git = "https://github.com/delta-io/delta-rs/", rev = "9597c3073fcf5202adf1723cb7a2cdc108820da3", features = ["s3-rustls"] }`

Then try `storage::parse_uri("s3://xxx/yyy")`. It will return an `InvalidScheme` error.

# Related Issue(s)
None


houqp (approved):
good catch

PR #301, opened 2021-07-05
Checkpoint map support

nevi-me:
# Description

This adds the map type to the checkpoint schema, pending completion of the feature in arrow-rs.

# Related Issue(s)

TODO

# Documentation

TODO

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "8b5ace29ee3216de37c0546865ad08edef58b0f9e76838ed8959a84a990e58c5"
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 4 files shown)


nevi-me:
@xianwill I've temporarily pointed this to the `map-support` on my arrow fork. I expect some tests to fail (datafusion might not compile because of different arrow versions), but the core Rust tests with default features pass, showing that the map support works.

We can update the PR when the arrow-rs work is complete and a new version is released, but I'm hoping this unblocks your progress in the interim.

xianwill:
Superseded by https://github.com/delta-io/delta-rs/pull/349

PR #305, opened 2021-07-08
Remove commit_version in favor of public prepare_commit/try_commit transaction API

mosyp:
# Description
This PR brings low-level transaction API into public. Once it's merged we'll end up with 3 ways of writing data into delta:
1. Low-level transaction API: The `prepare_commit` and `try_commit` function. First creates the temporary file and second is trying to rename it with given version. That's what is going to be used by kafka-delta-ingest. Users of this will have to handle the commit loop by themself
2. High-level transaction API: The `commit` function. It actually reuses the API above and provide it's own commit loop.
3. Wirter API in the `writer.rs`.

We had to get rid of `commit_version` as it's produced a unhandled temporary file in case of commit conflict. Moreover, since the user didn't have the control over the temp file, it's been created on each try. With new API, having the `PreparedCommit` object, one might use it for each retry without the need to duplicates files in end storage.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1142,12 +1142,15 @@ impl Default for DeltaTransactionOptions {
 
 /// Object representing a delta transaction.
 /// Clients that do not need to mutate action content in case a transaction conflict is encountered
-/// may use the `commit_with` method and rely on optimistic concurrency to determine the
+/// may use the `commit` method and rely on optimistic concurrency to determine the
 /// appropriate Delta version number for a commit. A good example of this type of client is an
 /// append only client that does not need to maintain transaction state with external systems.
-/// Clients that may need to do conflict resolution if the Delta version changes should use the `commit_version`
-/// method and manage the Delta version themselves so that they can resolve data conflicts that may
-/// occur between Delta versions.
+/// Clients that may need to do conflict resolution if the Delta version changes should use
+/// the `prepare_commit` and `try_commit` methods and manage the Delta version themselves so
+/// that they can resolve data conflicts that may occur between Delta versions.
+///
+/// Please not that in case of non-retryable error the temporary commit file such as
+/// `_delta_log/_commit_<uuid>.json` will orphaned in storage.
 #[derive(Debug)]
 pub struct DeltaTransaction<'a> {
     delta_table: &'a mut DeltaTable,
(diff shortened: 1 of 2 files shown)


rust/src/delta.rs
```diff
@@ -1116,52 +1103,13 @@ pub enum DeltaTransactionError {
     },
 }
 
-/// Error that occurs when a single transaction commit attempt fails
-#[derive(thiserror::Error, Debug)]
-pub enum TransactionCommitAttemptError {
```
mosyp, line 1121: I have removed this as it was not different to `DeltaTransactionError` at all and introduced additional mapping/layers

rust/tests/simple_commit_test.rs
```diff
@@ -85,7 +83,8 @@ mod simple_commit_fs {
 
         let mut tx1 = table.create_transaction(None);
         tx1.add_actions(tx1_actions());
-        let result = tx1.commit_version(1, None).await.unwrap();
+        let commit = tx1.prepare_commit(None).await.unwrap();
+        let result = tx1.try_commit(&commit, 1).await.unwrap();
```
mosyp, line 87: That's how the low-level transaction API is going to be used instead of `commit_version`

houqp (approved):
LGTM!

rust/src/delta.rs
```diff
@@ -1329,64 +1277,89 @@ impl<'a> DeltaTransaction<'a> {
...
-        let version = self.try_commit(&tmp_commit_uri, version).await?;
+        // move temporary commit file to delta log directory
+        // rely on storage to fail if the file already exists -
+        self.delta_table
+            .storage
+            .rename_obj(
+                &commit.uri,
+                &self.delta_table.commit_uri_from_version(version),
+            )
+            .await?;
```
rtyler, line 1331: In the cases where an error is not retry-able or recoverable, if this operation fails then the temporary file gets left on the file system in perpetuity correct?

I don't see any problem with that at this point, but if that's the case, I think the Rustdocs should be updated to document which cases might result in a temporary commit file orphaned in storage, and the types of file names (presuming `_commit_blah.json` is a temp file name) that an admin could then go in later and clean up

xianwill (dismissed):

rust/src/delta.rs
mosyp: is that going to work for us? https://github.com/delta-io/delta-rs/pull/305/files#diff-3b5c80af9e5527fbeaa6dad151a835f0733a26e61d0ba0471c58f243a29dfa99R1152-R1153

rtyler (approved):

PR #307, opened 2021-07-09
Introduce try_commit_transaction in DeltaTable

mosyp:
While introducing changes from #305 in kafka-delta-ingest, I figured out that it was not enough. Because in previous custom commit loop, the transaction was being created on each retry which was okay. But now, since we try to use the same transaction for each try in order to avoid duplicates of tmp files, we also have to call `table.update()` before each try to get the latest version. However the `DeltaTransaction` holds mutable reference to `DeltaTable` so you cannot call an update on it.  So there's 2 possible ways on achieving that:
- Make `delta_table` public in `DeltaTransaction`. This will allow us to access to table object within transaction to trick borrow checker, this repeats the behaviour in delta-rs commit_loop as it's implemented within `DeltaTransaction` self scope. I don't like that because it breaks encapsulation and nothing stop a user to modify it and change to w/e comes up.
- Move `try_commit` in `DeltaTable`. That's sounds more reasonable and that's what this PR brings. Once `prepare_commit` is called, there's no more need of `DeltaTransaction` since the file is written, and hence we may safely drop it. Having that, we end up with `PreparedCommit` object which has only tmp file uri without mutable state and everyone's happy. 

I've also replicated the kafka-delta-ingest usage in `test_low_level_tx_api` so we won't break it in future

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -972,6 +972,29 @@ impl DeltaTable {
         DeltaTransaction::new(self, options)
     }
 
+    /// Tries to commit a prepared commit file. Returns `DeltaTransactionError::VersionAlreadyExists`
+    /// if the given `version` already exists. The caller should handle the retry logic itself.
+    /// This is low-level transaction API. If user does not want to maintain the commit loop then
+    /// the `DeltaTransaction.commit` is desired to be used as it handles `try_commit_transaction`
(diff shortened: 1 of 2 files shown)


rtyler (approved):

houqp (approved):
neat solution to make borrow checker happy :)

PR #306, opened 2021-07-09
Expose delta statistics as utility function

Dandandan:
# Description
This exposes the file statistics of the files in a table.

This also adds some integration testing against a delta dataset.

# Related Issue(s)

Closes #247

# Documentation

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file

PR #308, opened 2021-07-09
Expose delta statistics as utility function #306, start using statistics for Datafusion `TableProvider` 

### LINKED ISSUE #247: Incorporate test data with per-file statistics
# Description

None of the golden data sets nor ours have any per-file statistics in their delta logs.

**Use Case**

**Related Issue(s)**

#246

Dandandan:
# Description
This starts using the provided metadata in delta transactions for the table provider, instead of the rowgroup.
This can be much faster, especially for a large number of files.

This also adds some integration testing against a delta dataset.
This exposes the file statistics of the files in a table.

# Related Issue(s)

Closes #247

# Documentation

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -827,7 +827,7 @@ impl DeltaTable {
     }
 
     /// Returns statistics for files, in order
-    pub fn get_stats(&self) -> Result<Vec<Option<Stats>>, DeltaTableError> {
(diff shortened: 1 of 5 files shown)


houqp (approved):
Much better, thanks @Dandandan !

PR #309, opened 2021-07-10
DataFusion: Expose, calculate number of records in TableProvider

Dandandan:
# Description
Adds implementation of `statistics` to table provider to return total number of rows.

This helps in applying cost-based optimization (join reordering), and optimization with a direct  `COUNT(*)`.
I also removed the unneeded crossbeam dependency.

# Related Issue(s)

# Documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -462,64 +462,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


houqp (approved):
Nice! What do you think about returning stats through an iterator instead of a materialized vector? The file list could get pretty big for large tables.

Dandandan:
> Nice! What do you think about returning stats through an iterator instead of a materialized vector? The file list could get pretty big for large tables.

Yes, might make sense for very large tables. But I guess it only starts mattering at the thousands, 10s of thousands of files. Not having to read the parquet metadata should already be the biggest saving though!

houqp:
oh yeah, avoid unnecessary parquet file read is definitely the biggest gain for sure :)

PR #311, opened 2021-07-11
Fix incorrect latest version

viirya:
# Description
The `get_latest_version` method looks incorrect to me, by comparing with the logic in other latest version related methods.

The correct logic looks like to be: 1. get last check point version from `get_last_checkpoint`. 2. try to load log with `version = version + 1` until failing to load one and the the latest version is `version = version - 1`.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -547,7 +547,7 @@ impl DeltaTable {
 
     async fn get_latest_version(&mut self) -> Result<DeltaDataTypeVersion, DeltaTableError> {
         let mut version = match self.get_last_checkpoint().await {
-            Ok(last_check_point) => last_check_point.version,
+            Ok(last_check_point) => last_check_point.version + 1,
             Err(LoadCheckpointError::NotFound) => {
                 // no checkpoint, start with version 0
                 0


rust/src/delta.rs
```diff
@@ -547,7 +547,7 @@ impl DeltaTable {
 
     async fn get_latest_version(&mut self) -> Result<DeltaDataTypeVersion, DeltaTableError> {
         let mut version = match self.get_last_checkpoint().await {
-            Ok(last_check_point) => last_check_point.version,
+            Ok(last_check_point) => last_check_point.version + 1,
```
viirya, line 550: Not sure if I misunderstand it. But by comparing with other related methods, this looks incorrect to me.

viirya:
cc @houqp

rust/src/delta.rs
houqp: I think this is a valid optimization. It should save us one unnecessary object head call because for a valid delta table, the version stored in the checkpoint should always point to a valid commit object.

houqp (approved):
LGTM :+1: 

While going through the code, I noticed we missed one edge case handling where for an empty table directory, this method would return `-1`.

houqp:
Filed https://github.com/delta-io/delta-rs/issues/312.

PR #313, opened 2021-07-12
Introduce update_incremental

### LINKED ISSUE #128: Default to avoid checkpoint restore in update method
Follow up for https://github.com/delta-io/delta-rs/issues/125#issuecomment-800516633.

When pulling in table updates, It would be more efficient to just load all the all the new versions commit by commit instead of rebuilding the state from the latest checkpoints. Checkpoints are mostly useful for bootstrapping purpose. Running tables are usually not too far behind from the latest commits.

mosyp:
Not sure that that fully "fixes" the #128, but that's still the improvement, especially for writes, as you don't load checkpoints on each write/retry.

Also note attention to the function rename `apply_logs_after_current_version` -> `apply_logs_from_current_version`. I found after misleading as it's first applies the current version and then increments. That was not acceptable for update_incremental function since the current version is already applied. Hence, without creating any new wheels I figured to reuse that function but then, within update_incremental, I'm incrementing the table version so the next one will be applied. There's the case when  there's no new version after current, but `apply_logs_from_current_version` already handles that by reverting the version to one back so we're good there.

upd. Closes #128.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -635,8 +635,7 @@ impl DeltaTable {
     /// incrementally applying each version since current.
     pub async fn update_incremental(&mut self) -> Result<(), DeltaTableError> {
         self.version += 1;
-        self.apply_logs_from_current_version().await?;
-        Ok(())
+        self.apply_logs_from_current_version().await
     }
 
     async fn apply_logs_from_current_version(&mut self) -> Result<(), DeltaTableError> {
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -602,12 +602,13 @@ impl DeltaTable {
             }
         }
 
-        self.apply_logs_after_current_version().await?;
+        self.apply_logs_from_current_version().await?;
 
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


xianwill (dismissed):

rust/src/delta.rs
```diff
@@ -625,12 +626,20 @@ impl DeltaTable {
...
         Ok(())
     }
 
-    async fn apply_logs_after_current_version(&mut self) -> Result<(), DeltaTableError> {
+    /// Updates the DeltaTable to the most recent state committed to the transaction by
+    /// incrementally applying each version since current.
+    pub async fn update_incremental(&mut self) -> Result<(), DeltaTableError> {
+        self.version += 1;
+        self.apply_logs_from_current_version().await?;
+        Ok(())
```
houqp, line 639:
```suggestion
        self.apply_logs_from_current_version().await
```

houqp (approved):
Nice, I think we can close #128 with this PR :)

houqp (approved):

PR #321, opened 2021-07-18
Add total_byte_size to datafusion statistics

viirya:
# Description

This patch adds `total_byte_size` to datafusion statistics

# Related Issue(s)

# Documentation

PR #322, opened 2021-07-18
rename get_actions to get_active_add_actions

houqp:
# Description

`get_actions` might give reader the wrong impression that it will return all action variants.

Follow up on https://github.com/delta-io/delta-rs/pull/321.


houqp:
FYI @viirya .

viirya (approved):
lgtm

fvaleye (approved):

PR #327, opened 2021-07-19
Add max and min values to Statistics

viirya:
# Description

This patch adds max and min values to Statistics.

### DIFF
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -167,8 +167,7 @@ impl TableProvider for delta::DeltaTable {
                                                         if left_larger_than_right(
                                                             old_max_value.clone(),
                                                             max_value.clone(),
-                                                        )
-                                                        {
+                                                        ) {
                                                             Some(old_max_value)
                                                         } else {
                                                             Some(max_value)
@@ -191,8 +190,7 @@ impl TableProvider for delta::DeltaTable {
                                                         if left_larger_than_right(
                                                             min_value.clone(),
                                                             old_min_value.clone(),
-                                                        )
-                                                        {
+                                                        ) {
                                                             Some(old_min_value)
                                                         } else {
                                                             Some(min_value)
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -161,21 +161,19 @@ impl TableProvider for delta::DeltaTable {
                                             let old_stats = stats.clone();
                                             let max_value = to_scalar_value(x.as_value()?);
 
-                                            match max_value {
-                                                Some(max_value) => match old_stats.max_value {
-                                                    Some(old_max_value) => {
-                                                        if left_larger_than_right(
-                                                            old_max_value.clone(),
-                                                            max_value.clone(),
-                                                        ) {
-                                                            Some(old_max_value)
-                                                        } else {
-                                                            Some(max_value)
-                                                        }
+                                            match (max_value, old_stats.max_value) {
+                                                (Some(max_value), Some(old_max_value)) => {
+                                                    if left_larger_than_right(
+                                                        old_max_value.clone(),
+                                                        max_value.clone(),
+                                                    ) {
+                                                        Some(old_max_value)
+                                                    } else {
+                                                        Some(max_value)
                                                     }
-                                                    _ => Some(max_value),
-                                                },
-                                                _ => old_stats.max_value,
+                                                }
+                                                (Some(max_value), None) => Some(max_value),
+                                                (None, old) => old,
                                             }
                                         },
                                     ),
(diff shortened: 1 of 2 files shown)


houqp:
Recommend installing a rust-fmt plugin for the editor you use so you won't need to manually fix formatting ever again :)

houqp:
also cc @Dandandan

rust/src/delta_datafusion.rs
```diff
@@ -163,9 +213,51 @@ impl TableProvider for delta::DeltaTable {
...
 }
 
-fn stat_to_val(stat: &ColumnCountStat) -> Option<usize> {
-    match stat {
-        ColumnCountStat::Value(val) => Some(*val as usize),
-        ColumnCountStat::Column(_) => None,
+fn to_scalar_value(stat_val: &serde_json::Value) -> Option<datafusion::scalar::ScalarValue> {
+    if stat_val.is_number() {
+        if stat_val.is_i64() {
+            Some(ScalarValue::from(stat_val.as_i64()?))
```
houqp, line 219: I think the `is_type()` method call can be avoided if we reuse the option from `as_type` method:

```suggestion
        if let Some(val) = stat_val.as_i64() {
            Some(ScalarValue::from(val))
```

rust/src/delta_datafusion.rs
```diff
@@ -147,12 +153,56 @@ impl TableProvider for delta::DeltaTable {
...
+                                                        ) {
+                                                            Some(old_min_value)
+                                                        } else {
+                                                            Some(min_value)
+                                                        }
+                                                    }
+                                                    _ => Some(min_value),
+                                                },
+                                                _ => old_stats.min_value,
+                                            }
```
houqp, line 202: nitpick, you could reduce nested level and make the code more readable by matching on two values at the same time:

```suggestion
                                            match (min_value, old_stats.min_value) {
                                                (Some(min_value), Some(old_min_value)) => {
                                                    if left_larger_than_right(
                                                        min_value.clone(),
                                                        old_min_value.clone(),
                                                    ) {
                                                        Some(old_min_value)
                                                    } else {
                                                        Some(min_value)
                                                    }
                                                }
                                                (Some(min_value), None) => Some(min_value),
                                                (None, old) => old,
                                            }
```

rust/src/delta_datafusion.rs
```diff
@@ -163,9 +213,51 @@ impl TableProvider for delta::DeltaTable {
...
+        }
+        ScalarValue::Int32(Some(v)) => {
+            let i_right = i32::try_from(right).unwrap();
+            v > i_right
+        }
+        ScalarValue::Int64(Some(v)) => {
+            let i_right = i64::try_from(right).unwrap();
+            v > i_right
+        }
+        _ => unimplemented!("Unimplemented for {:?}", left),
```
houqp, line 261: nitpick, could we make this error message more explicit for easier debugging?

```suggestion
        _ => unimplemented!("Scalar value comparison unimplemented for {:?} and {:?}", left, right),
```

houqp (changes requested):

houqp (approved):
LGTM, I will give @Dandandan some time to review before merging this in. Thanks @viirya !

viirya:
Thank you @houqp!

rust/src/delta_datafusion.rs
```diff
@@ -163,9 +209,53 @@ impl TableProvider for delta::DeltaTable {
...
+fn to_scalar_value(stat_val: &serde_json::Value) -> Option<datafusion::scalar::ScalarValue> {
+    if stat_val.is_number() {
+        if let Some(val) = stat_val.as_i64() {
+            Some(ScalarValue::from(val))
+        } else if let Some(val) = stat_val.as_u64() {
+            Some(ScalarValue::from(val))
+        } else {
+            stat_val.as_f64().map(ScalarValue::from)
+        }
+    } else {
```
Dandandan, line 221: Probably some more easy to add are string / boolean.

viirya: Yea, just wondering if max/min values of string/boolean are kept in statistics too.

rust/src/delta_datafusion.rs
```diff
@@ -163,9 +209,53 @@ impl TableProvider for delta::DeltaTable {
     }
 }
 
-fn stat_to_val(stat: &ColumnCountStat) -> Option<usize> {
-    match stat {
-        ColumnCountStat::Value(val) => Some(*val as usize),
-        ColumnCountStat::Column(_) => None,
+fn to_scalar_value(stat_val: &serde_json::Value) -> Option<datafusion::scalar::ScalarValue> {
```
Dandandan, line 212: I think to be complete we have to pass the (arrow) datatype as well to this function, for example to parse/handle timestamps properly.

viirya: Hm, for comparison purpose, I think using few basic primitive types are more easy. Maybe we can convert the compared `ScalarValue` to correct `ScalarValue` after comparison, e.g. `ScalarValue::Int64` to `ScalarValue::TimestampSecond` if it is Timestamp Second, etc.

viirya: How about addressing this in another PR?

Dandandan: Yeah perfectly fine to handle this in a follow up PR.

FYI, There was some other approach of using the aggregation code from DataFusion over here:
https://github.com/apache/arrow-datafusion/pull/719

Might also make sense to extend the ScalarValue a bit at DataFusion side  (e.g. derive `Ord`)

Dandandan (approved):

rust/src/delta_datafusion.rs
houqp: +1 on extending ScalarValue in Datafusion wherever applicable for better code reuse. Implementing `Ord` trait there is a great idea.

viirya: Sounds good, the suggestion. I will find some time to see if I can implement `Ord` there.

PR #333, opened 2021-07-22
Fix consecutive checkpoints

mosyp:
Some of the code within `checkpoint_add_from_state` is not executed because of excessive check. This PR fixes it.

UPD. This is dependent on map support, so keeping this as DRAFT until the former is merged

### DIFF
--- a/rust/src/checkpoints.rs
+++ b/rust/src/checkpoints.rs
@@ -254,7 +254,7 @@ fn checkpoint_add_from_state(
 ) -> Result<Value, ArrowError> {
     let mut v = serde_json::to_value(action::Action::add(add.clone()))?;
 
-    if !add.partition_values.is_empty() && add.partition_values_parsed.is_none() {
+    if !add.partition_values.is_empty() {
         let mut partition_values_parsed: HashMap<String, Value> = HashMap::new();
 
         for (field_name, data_type) in data_types.iter() {
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -102,7 +102,8 @@ impl StorageBackend for FileStorageBackend {
             fs::create_dir_all(parent).await?;
         }
         let mut f = fs::OpenOptions::new()
-            .create_new(true)
+            .create(true)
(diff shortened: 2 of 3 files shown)


rust/src/storage/file/mod.rs
```diff
@@ -102,7 +102,8 @@ impl StorageBackend for FileStorageBackend {
             fs::create_dir_all(parent).await?;
         }
         let mut f = fs::OpenOptions::new()
-            .create_new(true)
+            .create(true)
+            .truncate(true)
             .write(true)
             .open(path)
             .await?;
```
mosyp, line 109: This was failing with already exists for fs because of `_last_checkpoint` overwrite. I was thinking about creating new function like `replace_obj` but then, in out s3 backend, the `put_obj` replaces exists by default and there's no workaround for it. So instead, I've updated this to be the same semantically as in s3

cc @houqp @xianwill

houqp: I think your change introduces the right behavior, only `rename_obj` is supposed to check for already exists error.

xianwill (approved):
LGTM. @mosyp - I think we should merge this to main then reverse integrate it into our map support branches.

rust/src/checkpoints.rs
```diff
@@ -254,7 +254,7 @@ fn checkpoint_add_from_state(
 ) -> Result<Value, ArrowError> {
     let mut v = serde_json::to_value(action::Action::add(add.clone()))?;
 
-    if !add.partition_values.is_empty() && add.partition_values_parsed.is_none() {
```
houqp, line 257: should we file an issue to track the work of supporting parsed partition values for better performance?

rust/src/storage/file/mod.rs
xianwill: Sure, I'll add an issue.

rust/src/checkpoints.rs
xianwill: Sure, I'll add an issue.

xianwill (approved):

PR #334, opened 2021-07-24
added warning message if the detected glibc version is < 2.28

Smurphy000:
# Description
Added a warning message to assist in debugging since atomic renames has only been implemented for glibc version >= 2.28

## Example output
image

### DIFF
--- a/rust/build.rs
+++ b/rust/build.rs
@@ -8,7 +8,7 @@ mod platform_cfg {
         } else {
             let more_info = "https://docs.rs/deltalake/latest/deltalake/storage/file/struct.FileStorageBackend.html";
             println!(
-                "cargo:warning=glibc versions < 2.28 are currently unsupported, glibc version found {}.{}. For more information: {}", 
+                "cargo:warning=glibc version >= 2.28 is required for performing commits to local file system, glibc version found {}.{}. For more information: {}", 
                 ver.major, ver.minor, more_info
             );
         }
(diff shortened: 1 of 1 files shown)


Smurphy000:
Looking for some input on wording for the warning message, and if the info link should even be included here.
Thoughts?

rust/build.rs
```diff
@@ -5,6 +5,12 @@ mod platform_cfg {
         let ver = glibc_version::get_version().unwrap();
         if ver.major >= 2 && ver.minor >= 28 {
             println!("cargo:rustc-cfg=glibc_renameat2");
+        } else {
+            let more_info = "https://docs.rs/deltalake/0.4.0/deltalake/storage/file/struct.FileStorageBackend.html";
```
houqp, line 9: would be better to point the doc url to latest version to pick up changes automateically.

```suggestion
            let more_info = "https://docs.rs/deltalake/latest/deltalake/storage/file/struct.FileStorageBackend.html";
```

rust/build.rs
```diff
@@ -5,6 +5,12 @@ mod platform_cfg {
...
+            println!(
+                "cargo:warning=glibc versions < 2.28 are currently unsupported, glibc version found {}.{}. For more information: {}",
```
houqp, line 11: for better clarification, glibc <2.28 is supported if users only need to perform read operations. >=2.28 is needed if users need to perform any write using filesystem backend.

Perhaps something along the line of:

```suggestion
                "cargo:warning=glibc version >= 2.28 is required for performing commits to local file system, glibc version found {}.{}. For more information: {}", 
```

houqp (approved):

PR #336, opened 2021-07-26
Convert scalar value to correct type based on arrow data type.

viirya:
# Description

This is a follow up of #327 for the comment https://github.com/delta-io/delta-rs/pull/327#discussion_r672030078.
For the max/min column scalar values, we convert it to correct `ScarlarValue` based on arrow datatype.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -229,11 +229,11 @@ impl TableProvider for delta::DeltaTable {
                             max_value: col_states
                                 .max_value
                                 .as_ref()
-                                .map(|scalar| correct_scalar_value_type(scalar.clone(), &dt)),
+                                .and_then(|scalar| correct_scalar_value_type(scalar.clone(), &dt)),
                             min_value: col_states
                                 .min_value
                                 .as_ref()
-                                .map(|scalar| correct_scalar_value_type(scalar.clone(), &dt)),
+                                .and_then(|scalar| correct_scalar_value_type(scalar.clone(), &dt)),
                             distinct_count: col_states.distinct_count,
                         }
                     })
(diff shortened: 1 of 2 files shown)


viirya:
I tried to add a test on date type with `delta-0.8.0-date` table. But the query like `SELECT max(date), min(date) FROM date_table` throws an error:

```
Error: Internal("Unsupported CAST from Date32 to Utf8")
```

Seems the aggregation on date types is not ready?

I can look at it later.

houqp:
> Seems the aggregation on date types is not ready?

Yes, datafusion only supports string, numeric and timestamp types for min max aggregation at the moment. We need to update `physical_plan/expressions/min_max.rs` and `physical_plan/aggregates.rs` to add date support.

rust/src/delta_datafusion.rs
```diff
@@ -224,6 +257,67 @@ fn to_scalar_value(stat_val: &serde_json::Value) -> Option<datafusion::scalar::S
...
+    value: datafusion::scalar::ScalarValue,
+    field_dt: &ArrowDataType,
+) -> datafusion::scalar::ScalarValue {
+    match field_dt {
+        ArrowDataType::Int64 => {
+            let raw_value = i64::try_from(value).unwrap();
+            ScalarValue::from(raw_value)
+        }
+        ArrowDataType::Int32 => {
+            let raw_value = i64::try_from(value).unwrap() as i32;
```
houqp, line 270: Is there any reason why we prefer downcasting from i64 to i32 over direct `i32::try_from(value)` call?

houqp (dismissed):

houqp:
FYI @Dandandan

rust/src/delta_datafusion.rs
```diff
@@ -224,6 +257,67 @@ fn to_scalar_value(stat_val: &serde_json::Value) -> Option<datafusion::scalar::S
...
+        }
+        ArrowDataType::Timestamp(TimeUnit::Microsecond, None) => {
+            let raw_value = i64::try_from(value).unwrap();
+            ScalarValue::TimestampMicrosecond(Some(raw_value))
+        }
+        ArrowDataType::Timestamp(TimeUnit::Millisecond, None) => {
+            let raw_value = i64::try_from(value).unwrap();
+            ScalarValue::TimestampMillisecond(Some(raw_value))
+        }
+        _ => unimplemented!(
```
Dandandan, line 313: Better to return a `None` here and leave the statistic value out instead of panicking.

houqp: If we are to return a None here, then I think it would be good to log an error message with `log::error!` so we don't leave unexpected performance degradation unnoticed.

rust/src/delta_datafusion.rs
viirya: `i32::try_from` call only applies on an input of `ScalarValue::Int32`. For number scalar value, we all use `ScalarValue::Int64` to compare to get max/min values. That's said here the input is `ScalarValue::Int64`.

viirya:
> > Seems the aggregation on date types is not ready?
> 
> Yes, datafusion only supports string, numeric and timestamp types for min max aggregation at the moment. We need to update `physical_plan/expressions/min_max.rs` and `physical_plan/aggregates.rs` to add date support.

Thanks for the info. I may take some time looking at it later.

viirya:
@houqp @Dandandan Thanks for the review. Is this good to go?

Dandandan (approved):

rust/src/delta_datafusion.rs
houqp: ha, i see. perhaps in the future, we could change `to_scalar_value` to do the scalar value type cast based on field type so we can avoid this intermediate 64 bits conversion overhead.

houqp (approved):

PR #342, opened 2021-07-31
Low level create table

Smurphy000:
# Description
This is the current implementation for a low level create table api

# Looking for feedback
- I think the new() method for DeltaTableMetaData needs to take in format (just defaulted to parquet to get things working)
- I think the create() method should return an updated version of the DeltaTable instead of nothing at the moment

### DIFF
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -364,11 +364,8 @@ pub struct Format {
 
 impl Format {
     /// Allows creation of a new action::Format
-    pub fn new(provider: String, options: Option<HashMap<String,String>>) -> Self {
-        Self {
-            provider,
-            options
-        }
+    pub fn new(provider: String, options: Option<HashMap<String, String>>) -> Self {
+        Self { provider, options }
     }
 
     /// Return the Format provider
@@ -382,7 +379,7 @@ impl Default for Format {
     fn default() -> Self {
         Self {
             provider: "parquet".to_string(),
-            options: Default::default()
+            options: Default::default(),
         }
     }
 }
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -363,6 +363,7 @@ pub struct Format {
 }
 
 impl Format {
+    /// Allows creation of a new action::Format
     pub fn new(provider: String, options: Option<HashMap<String,String>>) -> Self {
         Self {
             provider,
@@ -370,6 +371,7 @@ impl Format {
         }
     }
 }
+
 /// Action that describes the metadata of the table.
 /// This is a top-level action in Delta log entries.
 #[derive(Serialize, Deserialize, Debug, Default, Clone)]
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -614,7 +614,7 @@ impl Remove {
 
 /// Action used by streaming systems to track progress using application-specific versions to
 /// enable idempotency.
-#[derive(Serialize, Deserialize, Debug, Default)]
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct Txn {
     /// A unique identifier for the application performing the transaction.
@@ -667,7 +667,7 @@ impl Txn {
 
 /// Action used to increase the version of the Delta protocol required to read or write to the
 /// table.
-#[derive(Serialize, Deserialize, Debug, Default)]
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct Protocol {
     /// Minimum version of the Delta read protocol a client must implement to correctly read the
@@ -712,7 +712,7 @@ impl Protocol {
 
 /// Represents an action in the Delta log. The Delta log is an aggregate of all actions performed
 /// on the table, so the full list of actions is required to properly read a table.
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum Action {
     /// Changes the current metadata of the table. Must be present in the first version of a table.
     /// Subsequent `metaData` actions completely overwrite previous metadata.
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -370,6 +370,11 @@ impl Format {
             options
         }
     }
+
+    /// Return the Format provider
+    pub fn get_provider(self) -> String {
+        self.provider
+    }
 }
 
 /// Action that describes the metadata of the table.
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -354,7 +354,7 @@ impl Add {
 }
 
 /// Describes the data format of files in the table.
-#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Format {
     /// Name of the encoding for files in this table.
     provider: String,
@@ -374,6 +374,16 @@ impl Format {
     /// Return the Format provider
     pub fn get_provider(self) -> String {
         self.provider
+    }
+}
+
+// Assuming this is a more appropriate default than derived Default
+impl Default for Format {
+    fn default() -> Self {
+        Self {
+            provider: "parquet".to_string(),
+            options: Default::default()
+        }
     }
 }
 
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -362,6 +362,14 @@ pub struct Format {
     options: Option<HashMap<String, String>>,
 }
 
+impl Format {
+    pub fn new(provider: String, options: Option<HashMap<String,String>>) -> Self {
+        Self {
+            provider,
+            options
+        }
+    }
+}
 /// Action that describes the metadata of the table.
 /// This is a top-level action in Delta log entries.
 #[derive(Serialize, Deserialize, Debug, Default, Clone)]
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -354,7 +354,7 @@ impl Add {
 }
 
 /// Describes the data format of files in the table.
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct Format {
     /// Name of the encoding for files in this table.
     provider: String,
@@ -679,7 +679,7 @@ impl Txn {
 
 /// Action used to increase the version of the Delta protocol required to read or write to the
 /// table.
-#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+#[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct Protocol {
     /// Minimum version of the Delta read protocol a client must implement to correctly read the
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -207,6 +207,29 @@ pub struct DeltaTableMetaData {
     pub configuration: HashMap<String, String>,
 }
 
+impl DeltaTableMetaData {
+    fn new(
+        name: Option<String>, 
+        description: Option<String>, 
+        schema: Schema, 
+        partition: Vec<String>, 
+        config: HashMap<String, String>
+    ) -> Self {
+        //use inputs
+        
+        Self {
+            id: Uuid::new_v4().to_string(), //create a new GUID
+            name: name,
+            description: description,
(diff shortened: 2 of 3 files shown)


rust/src/delta.rs
```diff
@@ -1034,6 +1057,30 @@ impl DeltaTable {
...
     }
 
+    pub async fn create(
+        &mut self, 
+        metadata: DeltaTableMetaData, 
+        protocol: action::Protocol
+    ) -> Result<(), DeltaTableError> {
+        // need to check if table already exists so that protocal and metaData arent overwritten on second create
+        
+        let meta = action::MetaData::try_from(metadata).unwrap();
```
houqp, line 1067: we should propagate the error back to caller here instead of using unwrap :)

rust/src/delta.rs
```diff
@@ -1034,6 +1057,30 @@ impl DeltaTable {
...
+        let mut transaction = self.create_transaction(None);
+
+        //add commit info action
+        transaction.add_actions(
+            vec![
+                Action::protocol(protocol), 
+                Action::metaData(meta)
+            ]
+        );
+        transaction.commit(None).await.unwrap();
```
houqp, line 1078: it might be better to use lower level transaction API here because `commit` calls `try_commit_loop`, which is mostly designed for committing transactions into existing tables. For example, in the commit loop, it has a retry logic to keep retrying the commit with newer versions. This is not a valid action for table creation because we should only try to commit with version `0`. If we got an already exists error, we should just abort and return an error back to caller.

`try_commit_transaction` is probably what you want to use here.

rust/src/action.rs
```diff
@@ -362,6 +362,14 @@ pub struct Format {
     options: Option<HashMap<String, String>>,
 }
 
+impl Format {
+    pub fn new(provider: String, options: Option<HashMap<String,String>>) -> Self {
```
fvaleye, line 366: You can add [doc comments](https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments) later on the `new` method.

rust/src/delta.rs
```diff
@@ -207,6 +207,29 @@ pub struct DeltaTableMetaData {
...
+        name: Option<String>, 
+        description: Option<String>, 
+        schema: Schema, 
+        partition: Vec<String>, 
+        config: HashMap<String, String>
+    ) -> Self {
+        //use inputs
+        
+        Self {
+            id: Uuid::new_v4().to_string(), //create a new GUID
```
fvaleye, line 221: 👍  
Delta uses `java.util.UUID.randomUUID().toString` to create the [GUID](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#change-metadata): https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L350

rust/src/delta.rs
```diff
@@ -207,6 +207,29 @@ pub struct DeltaTableMetaData {
...
+            name: name,
+            description: description,
+            format: action::Format::new("parquet".to_string(), None),
+            schema: schema,
+            partition_columns: partition,
+            created_time: Utc::now().timestamp(),  //create a timestamp for current timestamp
```
fvaleye, line 227: The timestamp is in millisecond precision: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#change-metadata, https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L358.

You could use this method: https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#method.timestamp_millis

rust/src/delta.rs
```diff
@@ -1034,6 +1057,30 @@ impl DeltaTable {
         })
     }
 
+    pub async fn create(
```
fvaleye, line 1060: You can add doc comments later on the `create` method.

rust/src/delta.rs
```diff
@@ -1577,4 +1624,48 @@ mod tests {
...
+            ]
+        );
+
+        let deltamd = DeltaTableMetaData::new(
+            None, 
+            None, 
+            test_schema, 
+            vec![], 
+            HashMap::new()
+        );
```
fvaleye, line 1655: It might be really helpful to test what would be the generated value of `DeltaTableMetadata` attributes: `id`, `created_time`, `format` 😄

rust/src/delta.rs
Smurphy000: are you saying that this looks correct for generating a new guid?

rust/src/delta.rs
Smurphy000: ill add some asserts, thanks. How would you mock out file creation, so that when test run, files are just created somewhere? Or should a test include the write action, but just remove the files at the end?

rust/src/delta.rs
houqp: Yes, it could be helpful to add this reference in the code as a comment.

rust/src/delta.rs
Smurphy000: I have tried it now with the `try_commit_transaction` and it works well, however, I want to update the DeltaTable, I currently am not generating a _last_checkpoint (should that happen after a creation?) and my DeltaTable does not have a DeltaTableState, not sure how to create one from inside this function

rust/src/delta.rs
houqp: You can perform the write in a temp dir, then remove the directory at the end of the test. For example: https://github.com/delta-io/delta-rs/blob/a7328d6644688c79d618285aa4bd4ab728159340/rust/src/storage/file/mod.rs#L132

rust/src/delta.rs
Smurphy000: what method would you use to propagate the error back up? I tried the `?` but I am getting compile errors from that. I have seen a lot of ways to check with like `match` and stuff, but is there a preferred way, such as `unwrap_or_else` or possibly another method?

houqp: Yes, once the from trait for the error you want to propagate is implemented for `DeltaTableError`, you will be able to use `?`. 

We use `thiserror` crate to automatically implement error conversion for `DeltaTableError` at: https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L70`

rust/src/delta.rs
houqp: You don't need to worry about checkpoint here, we only create checkpoint for every 10 commits and it will be handled by https://github.com/delta-io/delta-rs/tree/main/aws/delta-checkpoint externally.

You don't need to create a new table state here since it's created in `DeltaTable::new` already. You can mutate it using `self.state` by applying the committed actions with `process_action`.

rust/src/delta.rs
Smurphy000: is unwrapping the try_from conversion from DeltaTableMetaData -> MetaData actually result in an error that would be considered a DeltaTableError? Ive never used the `thiserror` create and having trouble conceptualizing where the Error needs to be implemented to allow the `?` operator to be valid, sorry

houqp: Unwrap on a Result causes the program to crash immediately. You would want to put `?` on the `Result` itself, not the return from unwrap. So basically replace unwrap with `?`. `thiserror` is just a helper library that reduces boilerplate code to implement `From` trait for a Result's error enum variant. What the `?` operator does is it uses that `From` trait implementation to convert the Error enum variant to the error type used by the enclosing function, in this case `DeltaTableError`.

rust/src/delta.rs
Smurphy000: I'm trying to update the tables state inside the `create()` but after the transaction is committed. Since I am trying to use the same actions that were just committed to use in the `process_action()`, it results in an error due to `actions: Vec<>` already being used. Should I add the `copy` or `clone` trait for the `transaction actions` or would there be a better way to handle this?

rust/src/delta.rs
```diff
@@ -1034,6 +1058,34 @@ impl DeltaTable {
...
+            Action::protocol(protocol), 
+            Action::metaData(meta)
+        ];
+        
+        let mut transaction = self.create_transaction(None);
+        transaction.add_actions(actions);
+
+        // Need better error handling here
+        let prepared_commit = transaction.prepare_commit(None).await.unwrap();
+        self.try_commit_transaction(&prepared_commit, 0).await.unwrap();
```
Smurphy000, line 1081: These two lines are still using `unwrap()`, they return `DeltaTransactionError` on Err, but it has `From` implemented on `DeltaTableError`. When I tried to add a `From` for `DeltaTableError` to `DeltaTransactionError` I got some compile issue because of a cycle. Wondering how I can handle possible errors here better?

rust/src/delta.rs
houqp: Yes, you need to clone those actions since `add_action` takes ownership of the input.

rust/src/delta.rs
houqp: This is a good question. I feel like the the error wrapping might have been done in the wrong direction. I think it makes more sense to make `DeltaTransactionError` an enum variant of `DeltaTableError`, not the other way around. @xianwill @rtyler @mosyp what do you think?

houqp:
@Smurphy000 could you help manually resolve the comments you already addressed with follow up commits? That will make it a little bit easier for us to do follow up reviews :)

rust/src/delta.rs
Smurphy000: Maybe if I return a `DeltaTransactionError` here instead of a `DeltaTableError`?

houqp: Yeah, returning `DeltaTransactionError` here will work, but it just feels a little bit awkward because some of the errors are not related to transactions. Let's do this for now, we can manage the `DeltaTableError` change later as a follow up task.

houqp: created a follow up task https://github.com/delta-io/delta-rs/issues/346

rust/src/delta.rs
```diff
@@ -207,6 +207,31 @@ pub struct DeltaTableMetaData {
...
+    ///
+    pub fn new(
+        name: Option<String>, 
+        description: Option<String>, 
+        format: Option<action::Format>,
+        schema: Schema, 
+        partition: Vec<String>, 
+        config: HashMap<String, String>
+    ) -> Self {
+        // https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L350
```
houqp, line 220: Could use more context on the comment ;)

```suggestion
        // Reference implementation uses uuid v4 to create GUID:
        // https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L350
```

rust/src/delta.rs
```diff
@@ -1034,6 +1059,31 @@ impl DeltaTable {
...
 
+    /// Create a DeltaTable with version 0 given the provided MetaData and Protocol
+    pub async fn create(
+        &mut self,
+        metadata: DeltaTableMetaData,
+        protocol: action::Protocol,
+    ) -> Result<(), DeltaTransactionError> {
+        let meta = action::MetaData::try_from(metadata)?;
+
+        // TODO add commit info action
```
houqp, line 1070: do you mind filing a follow up issue for this?

rust/src/delta.rs
```diff
@@ -1577,4 +1627,75 @@ mod tests {
...
+        assert_eq!(dt.state.files.len(), 0);
+
+        // assert new _delta_log file created in tempDir
+        let table_path = Path::new(&dt.table_uri);
+        assert!(table_path.exists());
+
+        let delta_log = table_path.join("_delta_log");
+
+        assert!(delta_log.exists());
+        assert!(delta_log.join("00000000000000000000.json").exists());
```
houqp, line 1693: could you also add an assert to check for the content of the commit?

houqp (commented):
Looks good to me overall, great work @Smurphy000 !

rust/src/delta.rs
Smurphy000: I will, do you know where I can find some more info related to commit info? Is that something that still hasn't been fully implemented in delta-rs?

houqp: everything you want to know about the delta spec is documented at https://github.com/delta-io/delta/blob/master/PROTOCOL.md#commit-provenance-information. It's modeled as a json object in the current code base: https://github.com/delta-io/delta-rs/blob/main/rust/src/action.rs#L720

houqp (approved):
LGTM!

houqp:
Thank you @Smurphy000, I hope you had a good time hacking on rust and delta ;)

PR #351, opened 2021-08-05
Fix checkpoint and timestamp bugs

xianwill:
# Description

This PR fixes several blocking issues related to checkpoint data and schema correctness.

One especially important note is that I'm switching from nanoseconds to microseconds for the default arrow timestamp unit. Neither actually works well for read scenarios because the parquet timestamp could be one of various types (https://github.com/delta-io/delta/issues/643), but for writes, microseconds definitely looks like the best bet for broad support and good precision.

Note: I'm merging this to the writer-map-support branch rather than main because the required arrow and parquet crate refs aren't ref'ed by datafusion yet.

Additional note: this also picks up @mosyp's fix to the fs backend that was merged to main in https://github.com/delta-io/delta-rs/pull/376

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "8b5ace29ee3216de37c0546865ad08edef58b0f9e76838ed8959a84a990e58c5"
 
 [[package]]
 name = "addr2line"
-version = "0.14.1"
+version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a55f82cfe485775d02112886f4169bde0c5894d75e79ead7eafe7e40a25e45f7"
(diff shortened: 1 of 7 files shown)


mosyp:
@xianwill I had to edit your PR because you were picking a bug that is fixed in recent main.
 I don't know what's going on but it seems like it's a broken PR. I've reset you're branch to the latest upstream and added your changes on top of it with a single commit but github PR shows diffs and commits that should not be here, as well as conflicts

mosyp:
It's a PR to `delta-io:writer-map-support` ugh. Too early, need my coffee :D

mosyp (approved):
Alright, I've messed up with a PR. But bug fixes are looking good!

PR #354, opened 2021-08-06
Modify partition_values field type in Add/Remove actions.

### LINKED ISSUE #348: Read table failed when containing null partition values
# Environment

**Delta-rs version**: 0.5.1

**Binding**: Rust/Python

**Environment**:
- **OS**: CentOS 7.6
- **Other**: Spark 3.0.1 + delta 0.7.0

***
# Bug

**What happened**:

When reading partitioned delta table containing null values, the program throws an exception.

Should be able to read table containing null partition values.

**How to reproduce it**:

```
spark = pyspark.sql.SparkSession.builder.getOrCreate()
sdf = spark.createDataFrame([['A', 1], [None, 2]], schema=['k', 'v'])
delta_path = 's3a://tmp/test_partition_with_null'
s3_path = 's3://tmp/test_partition_with_null'

sdf.write.format("delta").mode("overwrite").option("overwriteSchema", "true").save(delta_path)
dt = DeltaTable(s3_path)  # throw PyDeltaTableError!
```

The error message:

```
In [15]: dt = DeltaTable(path)
PyDeltaTableError                         Traceback (most recent call last)
<ipython-input-15-2195f07dec1c> in <module>
----> 1 dt = DeltaTable(path)

/home/guandata/.pyenv/versions/3.7.9/lib/python3.7/site-packages/deltalake/table.py in __init__(self, table_uri, version)
     74         :param version: version of the DeltaTable
     75         """
---> 76         self._table = RawDeltaTable(table_uri, version=version)
     77         self._metadata = Metadata(self._table)
     78

PyDeltaTableError: Failed to apply transaction log: Invalid JSON in log record
```

It works fine when using Spark:

```
In [16]: sdf1 = spark.read.format('delta').load('s3a://dolphinscheduler/tmp/test_partition_on_null')

In [17]: sdf1.show()
+----+---+
|   k|  v|
+----+---+
|null|  2|
|   A|  1|
+----+---+
```

zijie0:
# Description
Modify partition_values to type HashMap<String, Option<String>> according to the delta spec.

# Related Issue(s)

- closes #348 

# Documentation

Detailed discussion here: https://groups.google.com/g/delta-users/c/ZzzrtNRqRyk


houqp:
Thank you @zijie0 ! Looks good from a quick glance, I will take a closer look later today.

zijie0:
@houqp I'm not sure whether I need to add some write tests. I tried locally changed following test: https://github.com/delta-io/delta-rs/blob/main/rust/tests/write_exploration.rs#L422 with empty string value:

```
// test data set #2 - updates
let json_rows = vec![
    json!({ "id": "D", "value": 148, "modified": "" }),
    json!({ "id": "E", "value": 150, "modified": "" }),
    json!({ "id": "F", "value": 152, "modified": "" }),
];
```

The generated files seems to be the same with "native" delta.

rust/src/checkpoints.rs
```diff
@@ -293,25 +293,38 @@ fn checkpoint_add_from_state(
     Ok(v)
 }
 
-fn typed_partition_value_from_string(
-    string_value: &str,
+fn typed_partition_value_from_option_string(
```
houqp, line 296: perhaps we can simplify this by wrapping the original `typed_partition_value_from_string` implementation in `typed_partition_value_from_option_string` like so:

```rust
match string_value {
    Some(s) => if s.is_empty() {
        Ok(Value::Null)  // empty string should be deserialized as null
    } else {
        typed_partition_value_from_string(s, date_type)
    },
    None => Ok(Value::Null),
}
```

rust/tests/write_exploration.rs
```diff
@@ -175,12 +175,13 @@ impl DeltaWriter {
...
-                        .ok_or(DeltaWriterError::MissingPartitionColumn {
-                            col_name: k.to_string(),
-                        })?;
+                let partition_value = partition_values
+                    .get(k)
+                    .ok_or(DeltaWriterError::MissingPartitionColumn {
+                        col_name: k.to_string(),
+                    })?
+                    .clone()
+                    .unwrap_or("__HIVE_DEFAULT_PARTITION__".to_string());
```
houqp, line 184: This reminded me that we are doing partition filtering wrong in `get_files_by_partitions`. We are parsing Action.path to  reconstruct the partition values, but instead, we should be using `Action.partition_values` or `Action.partition_values_parsed` directly.

I will file a separate issue to track this.

rust/src/action.rs
```diff
@@ -150,7 +173,7 @@ pub struct Add {
     /// The size of this file in bytes
     pub size: DeltaDataTypeLong,
     /// A map from partition column to value for this file
-    pub partition_values: HashMap<String, String>,
+    pub partition_values: HashMap<String, Option<String>>,
```
houqp, line 176: It occurred to me that the problem you found has a much bigger impact than just the partition_values field, it also applies to Add.tags, Remove.tags, MetaData.format.options, MetaData.configurations.

Let me know if you prefer to address them or only focus on partition_values in this PR. If you prefer the latter, I can file an issue to track the fix for the other fields.

rust/tests/write_exploration.rs
houqp: cc @fvaleye https://github.com/delta-io/delta-rs/issues/358.

houqp:
> @houqp I'm not sure whether I need to add some write tests.

@zijie0 it would be great if you could help write some write tests to cover the cases of both empty strings and actual null values. So perhaps something like below:

```
    json!({ "id": "D", "value": 148, "modified": "" }),
    json!({ "id": "E", "value": 150, "modified": null }),
```

rust/src/action.rs
zijie0: Hey QP, I went through the spec doc but didn't found situations that those structs could also contain null values in the map. Is that because Map[String, String] allows null value in Scala, so that we also need to make the change to HashMap<String, Option<String>>?

rust/src/action.rs
```diff
@@ -47,6 +47,29 @@ fn populate_hashmap_from_parquet_map(
...
+    for j in 0..pmap.len() {
+        map.entry(
+            keys.get_string(j)
+                .map_err(|_| "key for HashMap in parquet has to be a string")?
+                .clone(),
+        )
+        .or_insert(Some(
+            values
+                .get_string(j)
+                .map_err(|_| "key for HashMap in parquet has to be a string")?
```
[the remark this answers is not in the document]
zijie0: Fixed.

rust/src/checkpoints.rs
zijie0: Modified.

rust/src/action.rs
houqp: That's correct, based on the discussion in the dev list, all data types are annotated using scala syntax, so all `Map[String, String]` should map to `HashMap<String, Option<String>>` in rust.

houqp: btw, I am perfectly fine with us addressing these other fields in an other PR if you want to keep the scope of this PR small. The fix for partition_values itself is already a  big improvement :)

rust/tests/write_exploration.rs
```diff
@@ -453,10 +458,54 @@ async fn smoke_test() {
...
+    let remove = create_remove(remove_path);
+
+    transaction.add_actions(vec![Action::add(add), Action::remove(remove)]);
+    // commit the transaction
+    transaction.commit(None).await.unwrap();
+
+    let files = delta_table.get_files();
+    assert_eq!(files.len(), 1);
+    assert_eq!(
+        files[0].split("/").next().unwrap(),
```
houqp, line 505: one minor ask, could we check the partition_values field instead to make sure it's null? delta doesn't care whether we have partition value in the file path or not, so it's perfectly valid to not have `modified=__HIVE_DEFAULT_PARTITION_` in the path.

houqp:
The rest looks good to me, thank you @zijie0 for discovering this big can of worms :)

rust/src/action.rs
zijie0: Let me take a look on this :)

rust/tests/write_exploration.rs
zijie0: Good point! I will change the assertion.

zijie0:
It is weird, `cargo clippy` passed on my local machine. Could you help me on this build failure? @houqp

zijie0:
Sorry, forgot to rebase on main branch...

houqp (approved):

houqp:
Thank you @zijie0 !

PR #355, opened 2021-08-06
Google cloud

### LINKED ISSUE #56: DeltaTable support for GCS
Can I read DeltaTable that resides in GCS like below

dt = DeltaTable("gs://<path>")

blogle:
# Description
This pull request adds a storage backend enabling support for Google Cloud Storage. 
- Read support
- [x] `head_obj`
- [x] `get_obj`
- [x] `list_objs`

- Write Support
- [x] `put_obj`
- [x] `rename_obj`
- [x] `delete_obj`

# Related Issue(s)
- Closes #56


rust/src/storage/gcs/util.rs
```diff
@@ -0,0 +1,158 @@
...
+                request,
+                scope_hash,
+                ..
+            } => {
+                let (parts, body) = request.into_parts();
+                let read_body = std::io::Cursor::new(body);
+                let new_request = http::Request::from_parts(parts, read_body);
+
+                let req = convert_request(new_request, &backend.client).await?;
+                //.context("failed to create token request")?;
```
houqp, line 24: we should delete these context comment if they are not needed anymore.

houqp (approved):
Looking good from a quick glance. @blogle could you also update the `Supported backends` section in the project root readme to mark gcs as supported?

mosyp (approved):
LGTM, thank you!. 
Although I recommend to add local development capabilities with docker as we've done with localstack [here](https://github.com/delta-io/delta-rs/blob/main/docker-compose.yml) and [here](https://github.com/delta-io/delta-rs/blob/main/build/setup_localstack.sh).

I've not personally worked with google cloud but with quick google I found they provide docker image for testing as in https://cloud.google.com/run/docs/testing/local.

This way we can prevent behavioural differences between backend and avoid future bug. 

Btw, we can merge it as is not and add it in separate PR or do it here. Both works for me well

blogle:
> LGTM, thank you!.

:bow: We are currently trying to make a determination as to whether we should build some new data infrastructure around BigQuery or gcs+delta lake, so this work is at least somewhat motivated by self interest.
 
> Although I recommend to add local development capabilities with docker as we've done with localstack [here](https://github.com/delta-io/delta-rs/blob/main/docker-compose.yml) and [here](https://github.com/delta-io/delta-rs/blob/main/build/setup_localstack.sh).
> 
> I've not personally worked with google cloud but with quick google I found they provide docker image for testing as in https://cloud.google.com/run/docs/testing/local.
> 
> This way we can prevent behavioural differences between backend and avoid future bug.
> 
> Btw, we can merge it as is not and add it in separate PR or do it here. Both works for me well

I definitely agree it would be ideal to have a local test harness for developing against google cloud storage. Unfortunately to the best of my knowledge Google does not expose such a mock/service fake. While they do have test versions for pub/sub, bigtable, etc (IIRC) I haven't seen such thing for gcs. The link you referenced is for google cloud run, which is essentially their version of aws fargate and used for deploying containers on managed infrastructure. There is a 3d party repo https://github.com/fsouza/fake-gcs-server which might be feasible, but its not clear at the surface level if it has 1:1 compatibility with google cloud storage or has the same semantics. It might be worth investigating in a follow on PR

houqp (approved):
I agree, if the tooling for gcs mocking is not as mature as what s3 has, then we could tackle it as a follow up task.

mosyp (approved):
Ok, works for me then

blogle:
Sounds like we are good to merge? I don't have write access to the repository, can someone do the honors of hitting the merge button?

houqp:
Thanks @blogle ! welcome to the party :)

PR #356, opened 2021-08-07
pub DeltaTable method to retrieve table configurations

### LINKED ISSUE #353: public DeltaTable method to access table config
# Description

**Use Case**

Looks like there is no public method declared on the `DeltaTable` struct itself to access table config value. Table configurations are currently encapsulated within DeltaTableState.TableMetadata, which is all private fields.

Smurphy000:
# Description
Added the table level API to get the current_metadata table configuration

# Related Issue(s)

- closes #353


Smurphy000:
Let me know if we need any specific tests for this PR, and I would be happy to write some

houqp (approved):

houqp:
Thank you @Smurphy000!

PR #362, opened 2021-08-08
Modify get_files_by_partitions to use partition values

### LINKED ISSUE #358: get_files_by_partitions should filter partitions using partition_values or partition_values_parsed field
# Environment

**Delta-rs version**: all

**Binding**: rust

***
# Bug

**What happened**:

In `DeltaTable::get_files_by_partitions`, we are generating partition values from file path.

We should be applying the filter using `Add.partition_values` or `Add.partition_values_parsed`. The path field in Add action should be transparent to the table parser. Writer can commit Add action with whatever path it wants so there is no guarantee that the path will actually contain partition key value parts. There is also no spec on how null partition values should be serialized in the file path.

**How to reproduce it**:

Try filter on files using null partition values, with empty string.

**More details**:

From https://github.com/delta-io/delta-rs/pull/354/files#r684593378.

zijie0:
# Description

# Related Issue(s)

- closes #358 

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -765,26 +765,15 @@ impl DeltaTable {
         &self,
         filters: &[PartitionFilter<&str>],
     ) -> Result<Vec<String>, DeltaTableError> {
-        let partitions_number = match &self
-            .state
-            .current_metadata
-            .as_ref()
-            .ok_or(DeltaTableError::NoMetadata)?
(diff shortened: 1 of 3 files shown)


houqp (approved):

houqp:
Thanks @zijie0 , the code is cleaner and more robust with this change now :)

PR #367, opened 2021-08-09
Implement atomic put_obj.

### LINKED ISSUE #361: File storage backend's `put_obj` implementation is not atomic
# Environment

**Delta-rs version**: all

**Binding**: rust

***
# Bug

**What happened**:

`put_obj` method in our file storage backend could result in partial update being visible to readers.

Content written out by `put_obj` should either be fully visible to readers or not visible at all.

To properly fix this, we need to write the file to a temp location first, then use atomic_rename to move the file to the target location. 

**How to reproduce it**:

concurrently read object updated by `put_obj` method to observe partial content updates.

**More details**:

Related to https://github.com/delta-io/delta-rs/pull/360.

### LINKED ISSUE #368: File storage should switch to sync_all instead of flush
# Environment

**Delta-rs version**: all

**Binding**: rust

***
# Bug

**What happened**:

We are only calling flush in file backend's `put_obj` implementation, which only flushes data into kernel buffer.

sync_call should be called instead to persist data from kernel buffer into disk.

**How to reproduce it**:

**More details**:

From https://github.com/delta-io/delta-rs/pull/367/files#r685403738

zijie0:
# Description
Use write and rename file to implement atomic put_obj in fs backend.

# Related Issue(s)

- closes #361 
- closes #368 

# Documentation

### DIFF
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -111,15 +111,15 @@ impl StorageBackend for FileStorageBackend {
             .await?;
 
         f.write_all(obj_bytes).await?;
-        f.flush().await?;
+        f.sync_all().await?;
         drop(f);
 
-        match fs::rename(tmp_path, path).await {
+        match self.rename_obj(tmp_path, path).await {
             Ok(_) => Ok(()),
             Err(e) => {
                 // If rename failed, clean up the temp file.
-                let _ = fs::remove_file(tmp_path).await?;
-                Err(StorageError::from(e))
+                self.delete_obj(tmp_path).await?;
+                Err(e)
             }
         }
     }
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -12,6 +12,7 @@ use tokio::io::AsyncWriteExt;
 use tokio_stream::wrappers::ReadDirStream;
 
 use super::{ObjectMeta, StorageBackend, StorageError};
+use uuid::Uuid;
 
 mod rename;
 
@@ -101,17 +102,26 @@ impl StorageBackend for FileStorageBackend {
         if let Some(parent) = Path::new(path).parent() {
             fs::create_dir_all(parent).await?;
         }
+        let tmp_path = &format!("{}_{}", path, Uuid::new_v4().to_string());
         let mut f = fs::OpenOptions::new()
             .create(true)
             .truncate(true)
             .write(true)
-            .open(path)
+            .open(tmp_path)
             .await?;
 
         f.write_all(obj_bytes).await?;
         f.flush().await?;
+        drop(f);
 
-        Ok(())
+        match fs::rename(tmp_path, path).await {
+            Ok(_) => Ok(()),
+            Err(e) => {
+                // If rename failed, clean up the temp file.
+                let _ = fs::remove_file(tmp_path).await?;
+                Err(StorageError::from(e))
+            }
+        }
     }
 
     async fn rename_obj(&self, src: &str, dst: &str) -> Result<(), StorageError> {


rust/src/storage/file/mod.rs
```diff
@@ -101,17 +102,26 @@ impl StorageBackend for FileStorageBackend {
...
             .write(true)
-            .open(path)
+            .open(tmp_path)
             .await?;
 
         f.write_all(obj_bytes).await?;
         f.flush().await?;
-
-        Ok(())
+        drop(f);
```
mosyp, line 115: @houqp Does the `f.flush()` that you've added makes sense now? Also, is it actually required to flush a handle since it's being closed immediately

mosyp (commented):
LGTM, although I'd like to resolve the `flush()` call question

rust/src/storage/file/mod.rs
houqp: yes, this is still needed, the content could still be partially cached in kernel without an explicit flush/sync.

after I took a second look at the doc, I think we should be using sync_all instead of flush.

rust/src/storage/file/mod.rs
```diff
@@ -101,17 +102,26 @@ impl StorageBackend for FileStorageBackend {
...
+
+        match fs::rename(tmp_path, path).await {
```
houqp, line 117: @zijie0 could you change this to use `rename::atomic_rename(src, dst)`?

rust/src/storage/file/mod.rs
houqp: Filed https://github.com/delta-io/delta-rs/issues/368 as follow up.

rust/src/storage/file/mod.rs
houqp: I just noticed our atomic rename is not async, which could block the tokio runtime and lead to latency issues in production. filed https://github.com/delta-io/delta-rs/issues/369 as a follow up for that.

zijie0: Done.

rust/src/storage/file/mod.rs
zijie0: I've also changed flush to sync_all.

houqp (approved):

houqp:
Thank you @zijie0 for the fix

rust/src/storage/file/mod.rs
houqp: Oh god, I made a very bad suggestion here :( I think the `fs:rename` you used here is the correct one. It does exactly what we needed, i.e. overwrite the destination if already exists. Our atomic rename should be named something else like `rename_if_not_exists` so people won't confuse it with `rename`.

houqp: doc to the builtin rename function: https://doc.rust-lang.org/std/fs/fn.rename.html

zijie0: @houqp It seems that rename on Unix does have some atomicity issue: 

> If newpath already exists, it will be atomically replaced, so
       that there is no point at which another process attempting to
       access newpath will find it missing.  However, there will
       probably be a window in which both oldpath and newpath refer to
       the file being renamed.

But other projects like `tempfile` use it anyway.

houqp: yes, the main goal we are trying to achieve here is to avoid reading partially written data from `newpath`/`targetpath`. It's ok for both path to exist and refer to the same file content as long as the file content itself is atomically presented to the readers.

zijie0: If @mosyp and you all agree, I may submit a new PR to use `tokio::fs::rename` in `put_obj` method.

mosyp: @zijie0 hey, I've the notification, sure, works for me!

PR #370, opened 2021-08-10
Make Format.options to be required field

mosyp:
# Description
Format.options is required field, see https://github.com/delta-io/delta/blob/eab445a15f478ed2454cf152edb795e9fece91f6/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L358-L360

### DIFF
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -362,12 +362,13 @@ pub struct Format {
     /// Name of the encoding for files in this table.
     provider: String,
     /// A map containing configuration options for the format.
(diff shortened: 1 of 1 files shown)


houqp (approved):

houqp:
Good catch @mosyp , following this, I found couple other fields that are incorrectly marked as optional including metadata.name and metadata.description, we should do a full audit of all the fields in action.rs to make sure we are not marking things as optional incorrectly.

houqp:
Filed #372 to track the audit

PR #374, opened 2021-08-11
Wrap DeltaTransactionError with DeltaTableError.

### LINKED ISSUE #346: Wrap DeltaTransactionError with DeltaTableError
# Description

**Use Case**

Today, `DeltaTableError` is wrapped as an enum variant in `DeltaTransactionError`, which forced some of the DeltaTable methods to return `DeltaTransactionError` as  the error type. I think things would be cleaner if we change the abstraction to have `DeltaTableError` wrapping `DeltaTransactionError` instead.

**Related Issue(s)**

From discussion in https://github.com/delta-io/delta-rs/pull/342#discussion_r680555797

zijie0:
# Description
Wrap DeltaTransactionError with DeltaTableError.

# Related Issue(s)

- closes #346 

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -184,6 +184,15 @@ pub enum DeltaTableError {
     /// Generic Delta Table error
     #[error("Generic DeltaTable error: {0}")]
     Generic(String),
+
+    /// Error that wraps an underlying DeltaTransaction error.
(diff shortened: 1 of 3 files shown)


houqp (approved):

houqp:
Thank you @zijie0 for the clean up!

PR #375, opened 2021-08-11
Bump arrow to 6.0.0-SNAPSHOT and bring map support to schema

mosyp:
# Description
Let's start merging feature specific branches one by one. This one with map support. I've also bumped the arrow-rs refs to the latest to pick other fixes in upstream.


mosyp:
Although we have an issue with #367. That PR changed semantics for `put_obj` because not it does not overwrite existing path which is required for `_last_checkpoint`. And overwrite is default behaviour for S3 also

mosyp:
Blocked by #376

mosyp:
Also blocked by datafusion deps, either version conflict or we have to updated our source code

houqp:
@mosyp does the fixes you need to pull in exist in arrow's active release branch (https://github.com/apache/arrow-rs/tree/active_release)? If so, pointing to commits in that branch should help fix the datafusion version conflict.

houqp:
FYI, upstream just started the release vote for arrow 5.2, so if everything goes well, we should be able to pin to 5.2 in 3 days.

mosyp:
@houqp Do you have a power to do later the same for datafusion? Their deps on arrow is what failing the compilation

houqp:
yeah, I have a PR for datafusion ready to merge once arrow 5.2 is released: https://github.com/apache/arrow-datafusion/pull/869.

Just to double check, we are not blocked on this PR for kdi right? If we need to merge this PR to release KDI, I am cool with that. Otherwise, it's better to keep reference this branch in KDI until 5.2 is out.

mosyp:
@houqp yes, we're not blocked and using this branch in kdi, lo yea, let's keep it as is for now

houqp:
@mosyp given that arrow 6.x release will still take some time, how about we merge this branch as is into main? I think it's fine to pin arrow by git reference.

.github/workflows/build.yml
```diff
@@ -88,7 +89,8 @@ jobs:
         toolchain: stable
         override: true
     - name: Run tests
-      run: cargo test --verbose --features datafusion-ext,azure
+      #run: cargo test --verbose --features datafusion-ext,azure TODO Revert back!
```
mosyp, line 92: @houqp fyi, we're turned off datafusion from build since it has deps on arrow that are not matching the hash that we have. Basically because we're on arrow-6.0.0-SNAPSHOT and datafusion is on 5.3.0

mosyp: I've tried to clone datafusion and reference the arrow rs deps to the main branch but has no success because of several tests failing

houqp: yeah, that's fine. datafusion is not ready for 6.x yet, see https://github.com/apache/arrow-datafusion/pull/984

mosyp: So what's yout take on that? Should we nerge it as is with datafusion commented out?

houqp: yes, i am fine with temporarily disable datafusion integration for now.

xianwill (approved):

rust/src/delta_arrow.rs
```diff
@@ -95,7 +114,7 @@ impl TryFrom<&schema::SchemaDataType> for ArrowDataType {
                     }
                     "timestamp" => {
                         // Issue: https://github.com/delta-io/delta/issues/643
-                        Ok(ArrowDataType::Timestamp(TimeUnit::Nanosecond, None))
+                        Ok(ArrowDataType::Timestamp(TimeUnit::Microsecond, None))
```
houqp, line 117: @fvaleye FYI, it looks like this PR reverted your nanosecond fix, which means your latest released python binding might be broken for tables created with default spark timestamp config :(

also cc @dispanser since he raised a similar issue.

houqp: Let's move our discussion to https://github.com/delta-io/delta-rs/issues/441 to figure out a better fix on our end. I have a feeling that this needs to be handled separately within pyarrow and datafusion.

PR #376, opened 2021-08-11
Allow filesystem backend put_obj to overwrite existing

mosyp:
A follow up of https://github.com/delta-io/delta-rs/pull/367. `put_obj` has to follow the semantic of replace existing which was violated with recent changes. There's atomic swap supported by additional flags of `rename2` but it only works if both files are present. As a workaround I'm adding a files_exists() call to determine which rename to call.

### DIFF
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -114,18 +114,21 @@ impl StorageBackend for FileStorageBackend {
         f.sync_all().await?;
         drop(f);
 
-        match self.rename_obj(tmp_path, path).await {
+        // atomic rename with swap=true only possible if both paths exists.
+        let swap = Path::new(path).exists();
+
+        match rename::atomic_rename(tmp_path, path, swap) {
             Ok(_) => Ok(()),
             Err(e) => {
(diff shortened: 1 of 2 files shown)


xianwill (approved):

rust/src/storage/file/mod.rs
```diff
@@ -114,18 +114,21 @@ impl StorageBackend for FileStorageBackend {
...
-        match self.rename_obj(tmp_path, path).await {
+        // atomic rename with swap=true only possible if both paths exists.
+        let swap = Path::new(path).exists();
+
+        match rename::atomic_rename(tmp_path, path, swap) {
             Ok(_) => Ok(()),
             Err(e) => {
                 // If rename failed, clean up the temp file.
-                self.delete_obj(tmp_path).await?;
+                //self.delete_obj(tmp_path).await?;
```
mosyp, line 124: 😞

houqp:
@mosyp I think we have a race condition in this change. The file could be deleted or created between the path exists check and the atomic rename call. Could you check my comment at https://github.com/delta-io/delta-rs/pull/367#discussion_r688288794 to see if what I said there makes sense?

PR #378, opened 2021-08-11
Refactoring of black, isort, mypy tools usages into pyproject.toml

fvaleye:
# Description
- Refactoring of Python black, isort and mypy tools inside `pyproject.toml`
- Add stubs for mypy
- Add pytest-clarify to improve the output of the tests suite results in Python


python/stubs/deltalake/deltalake.pyi
```diff
@@ -0,0 +1,4 @@
+from typing import Any
+
+RawDeltaTable: Any
+rust_core_version: Any
```
houqp, line 4: we could annotate this as a callable that returns a str

houqp (approved):
Nice cleanup!

PR #380, opened 2021-08-12
Audit action field optionality

### LINKED ISSUE #372: Audit action field optionality
# Environment

**Delta-rs version**: all

**Binding**: rust

***
# Bug

**What happened**:

Some of the fields are incorrectly marked as optional in action.rs

Make sure optionality annotation matches with what's in delta spec and reference implementation: https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala

**How to reproduce it**:

**More details**:

Follow up from https://github.com/delta-io/delta-rs/pull/370

fvaleye:
# Description
Audit Action attributes using https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala file

# Related Issue(s)
closes #372

### DIFF
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -406,7 +406,7 @@ pub struct MetaData {
     /// An array containing the names of columns by which the data should be partitioned
     pub partition_columns: Vec<String>,
     /// The time when this metadata action is created, in milliseconds since the Unix epoch
-    pub created_time: DeltaDataTypeTimestamp,
+    pub created_time: Option<DeltaDataTypeTimestamp>,
     /// A map containing configuration options for the table
     pub configuration: HashMap<String, Option<String>>,
 }
@@ -459,9 +459,10 @@ impl MetaData {
                         .clone();
                 }
                 "createdTime" => {
-                    re.created_time = record
-                        .get_long(i)
-                        .map_err(|_| gen_action_type_error("metaData", "createdTime", "long"))?;
+                    re.created_time =
+                        Some(record.get_long(i).map_err(|_| {
+                            gen_action_type_error("metaData", "createdTime", "long")
+                        })?);
                 }
                 "configuration" => {
                     let configuration_map = record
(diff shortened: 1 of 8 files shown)


fvaleye:
@houqp based on my investigation, tell me what you think about:
- Keeping optional fields in Actions and Metadata when there are null by default in delta (examples from Metadata: [names and description](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L372-L373)). 
`Option` will be used to avoid serialization/deserialization errors when reading the delta transaction log.
I encountered several issues when reading older delta table versions <1.0 (these fields were not in the transaction logs).

- Settting `created_time` in `Metadata` as Optional 
- Settting `deletion_timestamp` in `Remove` as Optional
- Setting the values of `data_change` and `extended_file_metadata` with their default values when reading delta table transaction log for Remove actions

houqp:
Sorry for the late response @fvaleye , catch up on notifications after getting back from vacation :) 

What you proposed makes total sense to me :+1:

rust/src/action.rs
```diff
@@ -537,7 +538,7 @@ pub struct Remove {
     /// The path of the file that is removed from the table.
     pub path: String,
     /// The timestamp when the remove was added to table state.
-    pub deletion_timestamp: DeltaDataTypeTimestamp,
+    pub deletion_timestamp: Option<DeltaDataTypeTimestamp>,
```
houqp, line 541: I think this is incorrectly annotated in the official delta spec, we should also send a PR to the main delta repo to correct it :)

houqp (approved):

fvaleye:
> Sorry for the late response @fvaleye , catch up on notifications after getting back from vacation :)
> 
> What you proposed makes total sense to me 👍

@houqp Oh, no worries at all! I hope you enjoyed it 🏖️

rust/src/action.rs
fvaleye: Good idea, will do!

fvaleye: https://github.com/delta-io/delta/pull/775

PR #381, opened 2021-08-12
Merge Cargo.toml into pyproject.toml

fvaleye:
# Description
Following the latest guidelines from `maturin`

# Documentation
https://github.com/PyO3/maturin/blob/main/Changelog.md#0110---2021-07-04


houqp (approved):

houqp:
nice, much cleaner than the old cargo definition :)

python/pyproject.toml
```diff
@@ -2,6 +2,39 @@
 requires = ["maturin==0.11.2"]
 build-backend = "maturin"
 
+[project]
+name = "deltalake"
+requires-python = ">=3.6"
+classifier = [
```
messense, line 8: This should be `classifiers`.

houqp: good catch @messense , do you want to submit a PR?

PR #382, opened 2021-08-12
Update datafusion and ballista links in README

ei-grad:
Their code has been moved out of arrow to separate repository.


houqp (approved):
Thanks @ei-grad

PR #384, opened 2021-08-13
Reuse table/storage instances in checkpoints

mosyp:
When creating checkpoints from table uri there's been at least 2 instances of backend. Also when calling with table_state, once has to have a table instance to call it, so we'll get additional instances of tables and backends.

I've also removed checkpoint struct as usually they created just for one call and hence unneeded. 

Also fixed a PR builds due to localstack image version bump

### DIFF
--- a/aws/delta-checkpoint/tests/lambda_checkpoint.rs
+++ b/aws/delta-checkpoint/tests/lambda_checkpoint.rs
@@ -48,7 +48,7 @@ async fn lambda_checkpoint_smoke_test() {
     };
     let result = client.get_object(request).await;
 
-    assert!(result.is_ok());
+    let _ = result.unwrap();
 
     // verify the _last_checkpoint file was created
     let request = GetObjectRequest {
--- a/rust/src/checkpoints.rs
+++ b/rust/src/checkpoints.rs
@@ -79,7 +79,13 @@ pub async fn create_checkpoint_from_table_uri(
     version: DeltaDataTypeVersion,
 ) -> Result<(), CheckpointError> {
     let table = open_table_with_version(table_uri, version).await?;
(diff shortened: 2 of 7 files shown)


aws/delta-checkpoint/tests/lambda_checkpoint.rs
```diff
@@ -48,7 +48,7 @@ async fn lambda_checkpoint_smoke_test() {
     };
     let result = client.get_object(request).await;
 
-    assert!(result.is_ok());
+    let _ = result.unwrap();
```
xianwill, line 51: This won't tell us whether the call succeeded and the checkpoint exists.

xianwill: nvm - i get it now. unwrap will fail if !is_ok

rust/src/checkpoints.rs
```diff
@@ -67,187 +67,170 @@ pub enum CheckPointWriterError {
...
+    create_checkpoint(
+        version,
+        table.get_state(),
+        table.storage.as_ref(),
+        &table.table_uri,
+    )
+    .await?;
+    table.update_incremental().await?;
+    Ok(())
+}
```
mosyp, line 111: Had a conversation with @xianwill on the behaviour of this call, some users might don't want to update to the latest version. But the `table.update_incremental().await?;` has been added in case the `table.load_version(version).await?;` was a time travel. So in this case the returned table will end up always with `>= pervious_version`

mosyp: Opened an issue https://github.com/delta-io/delta-rs/issues/385

cc @houqp

xianwill (approved):

PR #387, opened 2021-08-14
added initial commit info on create method for a DeltaTable

### LINKED ISSUE #350: Add commitInfo action to low level create table API
# Description
The commitInfo for Delta to store additional provenance data should be added to the create table API

# Additional Information
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#commit-provenance-information
https://github.com/delta-io/delta-rs/blob/main/rust/src/action.rs#L755

Smurphy000:
# Description
Adding commitInfo data for the create method for DeltaTable

# Related Issue(s)
closes #350

Looking for input on the field names for json provided by delta-rs
"delta-rs": version of delta-rs
"timestamp": unix time in milliseconds

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1741,6 +1741,20 @@ mod tests {
                 Action::metaData(action) => {
                     assert_eq!(DeltaTableMetaData::try_from(action).unwrap(), delta_md);
                 }
+                Action::commitInfo(mut action) => {
+                    let modified_action = if action.is_object() {
+                        let temp = action.as_object_mut().unwrap();
+                        temp["timestamp"] = Value::String("".to_string());
+                        Some(Value::Object(temp.clone()))
+                    } else {
+                        None
+                    };
+
+                    assert_eq!(
+                        modified_action.unwrap(),
+                        json!({"operation": "CREATE TABLE", "userName": "test user", "delta-rs": crate_version().to_string(), "timestamp": "".to_string()})
+                    )
+                }
                 _ => (),
             }
         }
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1081,16 +1081,36 @@ impl DeltaTable {
         })
     }
 
-    /// Create a DeltaTable with version 0 given the provided MetaData and Protocol
+    /// Create a DeltaTable with version 0 given the provided MetaData, Protocol, and CommitInfo
     pub async fn create(
         &mut self,
         metadata: DeltaTableMetaData,
         protocol: action::Protocol,
+        commit_info: &mut Value,
     ) -> Result<(), DeltaTableError> {
         let meta = action::MetaData::try_from(metadata)?;
 
-        // TODO add commit info action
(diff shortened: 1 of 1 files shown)


rust/src/delta.rs
```diff
@@ -1081,16 +1081,36 @@ impl DeltaTable {
         })
     }
 
-    /// Create a DeltaTable with version 0 given the provided MetaData and Protocol
+    /// Create a DeltaTable with version 0 given the provided MetaData, Protocol, and CommitInfo
     pub async fn create(
         &mut self,
         metadata: DeltaTableMetaData,
         protocol: action::Protocol,
+        commit_info: &mut Value,
```
houqp, line 1089: better to let the method take the ownership here just like the other arguments like metadata and protocol.
```suggestion
        commit_info: Value,
```

rust/src/delta.rs
```diff
@@ -1081,16 +1081,36 @@ impl DeltaTable {
...
+            Some(Value::Object(temp.clone()))
+        } else {
+            None
+        };
+
+        let actions = vec![
+            Action::commitInfo(altered_commit_info.unwrap()),
+            Action::protocol(protocol),
+            Action::metaData(meta),
+        ];
```
houqp, line 1113: Here is a cleaner way to implement this logic without unwraps that could lead to hard crash at runtime:

```suggestion
        let mut enriched_commit_info = match commit_info {
            Value::Object(map) => Ok(map),
            _ => Err(<Return error here>),
        }?;
        enriched_commit_info.insert(
            "delta-rs".to_string(),
            Value::String(crate_version().to_string()),
        );
        enriched_commit_info.insert(
            "timestamp".to_string(),
            Value::String(Utc::now().timestamp_millis().to_string()),
        );

        let actions = vec![
            Action::commitInfo(Value::Object(enriched_commit_info)),
            Action::protocol(protocol),
            Action::metaData(meta),
        ];
```

houqp (changes requested):

rust/src/delta.rs
```diff
@@ -1081,16 +1081,36 @@ impl DeltaTable {
...
+        // delta-rs commit info will include the delta-rs version and timestamp as of now
+        let altered_commit_info = if commit_info.is_object() {
+            let temp = commit_info.as_object_mut().unwrap();
+            temp.insert(
+                "delta-rs".to_string(),
+                Value::String(crate_version().to_string()),
+            );
+            temp.insert(
+                "timestamp".to_string(),
+                Value::String(Utc::now().timestamp_millis().to_string()),
```
houqp, line 1102: I think the timestamp here should be stored as number instead of string.

rust/src/delta.rs
Smurphy000: Would you also think that commit info should be optional since it isnt necessary? and in that case just create commit info that delta-rs supplies?

houqp: That's a good idea!

houqp (approved):

houqp:
Thank you @Smurphy000 !

Issue #268, opened 2021-05-25
Recover from S3 rename/commit pauses


mosyp:
# Description

When S3 pauses during a commit to the extent that the lock is expired and acquired by other worker then the commit is returned as failed. But the write will actually be completed in S3. As a result, caller should now retry with a newer version, which will cause duplicated writes.

There's a suggestion to recover from this failures as in https://github.com/delta-io/delta-rs/pull/233/files/3010f8290a8c3c42ea147791d75ff72111011840#r636692792.

The initial implementation was started with https://github.com/delta-io/delta-rs/pull/233/commits/f40aff8aa1979f9e0f85b7f4c1562da8c7b263b9, but was rejected due to the possibility of delta log inconsistency described in https://github.com/delta-io/delta-rs/pull/233#issuecomment-847847483. New implementation should address the issue descriobed in there.

houqp:
When a lock is repaired by another worker and the current worker resumes, we should still consider the commit successful right? What difference does it make if a specific rename is completed first by another worker, then repeated by the current worker after resume?

mosyp:
That is true, I made a false assumption in that comment due to initial implementation was failing in unit testing. But nevertheless I'm glad this has grown into separate issue/pr so it'll get more focus for implementation and testing.

PR #391, opened 2021-08-16
WIP: do not return error on successful rename with long pause

houqp:
# Description

Allow users of delta-rs to build distributed applications that can recover from long pauses. Applications that prefer to hard crash on pauses can still do so by match on the returned `RenameStatus` enum and execute crash code path if it is `RenameStatus::Repaired`.

# Related Issue(s)

close https://github.com/delta-io/delta-rs/issues/268


xianwill (approved):
LGTM

PR #395, opened 2021-08-18
Implement delete_objs in fs and s3 storage backends.

### LINKED ISSUE #394: Leverage batch delete in vacuum
# Description

**Use Case**

From https://github.com/delta-io/delta-rs/issues/393#issuecomment-900624545. Some of the backends like S3 support batch delete operation. We should leverage this in vacuum for better performance and lower cost (API call).

We could add a new `delete_objs` to the `StorageBackend` trait as an abstraction for batch deletes.

**Related Issue(s)**

#393

zijie0:
# Description

Implement delete_objs in fs and s3 storage backends. Add corresponding test cases.

# Related Issue(s)

- closes #394 

# Documentation

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -26,6 +26,7 @@ regex = "1"
 chrono = "0"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 lazy_static = "1"
+itertools = "0"
 
 # Azure
 reqwest = { version = "0", optional = true }
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -26,7 +26,6 @@ regex = "1"
 chrono = "0"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 lazy_static = "1"
-itertools = "0"
 
 # Azure
 reqwest = { version = "0", optional = true }
--- a/rust/src/storage/azure.rs
+++ b/rust/src/storage/azure.rs
@@ -201,7 +201,7 @@ impl StorageBackend for AdlsGen2Backend {
         unimplemented!("delete_obj not implemented for azure");
     }
 
-    async fn delete_objs(&self, _paths: &[&str]) -> Result<(), StorageError> {
+    async fn delete_objs(&self, _paths: &[String]) -> Result<(), StorageError> {
         unimplemented!("delete_objs not implemented for azure");
     }
 }
--- a/rust/src/storage/azure.rs
+++ b/rust/src/storage/azure.rs
@@ -200,6 +200,10 @@ impl StorageBackend for AdlsGen2Backend {
     async fn delete_obj(&self, _path: &str) -> Result<(), StorageError> {
         unimplemented!("delete_obj not implemented for azure");
     }
+
+    async fn delete_objs(&self, _paths: &[&str]) -> Result<(), StorageError> {
+        unimplemented!("delete_objs not implemented for azure");
+    }
 }
 
 #[cfg(test)]
--- a/rust/src/storage/azure.rs
+++ b/rust/src/storage/azure.rs
@@ -200,10 +200,6 @@ impl StorageBackend for AdlsGen2Backend {
     async fn delete_obj(&self, _path: &str) -> Result<(), StorageError> {
         unimplemented!("delete_obj not implemented for azure");
     }
-
-    async fn delete_objs(&self, _paths: &[String]) -> Result<(), StorageError> {
-        unimplemented!("delete_objs not implemented for azure");
-    }
 }
 
 #[cfg(test)]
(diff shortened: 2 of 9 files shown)


Dandandan:
Cool @zijie0 i see you were ahead of me.

https://github.com/delta-io/delta-rs/pull/396

Maybe there is some inspiration to take from the other PR

rust/src/storage/azure.rs
```diff
@@ -200,6 +200,10 @@ impl StorageBackend for AdlsGen2Backend {
     async fn delete_obj(&self, _path: &str) -> Result<(), StorageError> {
         unimplemented!("delete_obj not implemented for azure");
     }
+
+    async fn delete_objs(&self, _paths: &[String]) -> Result<(), StorageError> {
+        unimplemented!("delete_objs not implemented for azure");
```
Dandandan, line 205: I think this is not needed, as there is a default implementation

rust/src/storage/file/mod.rs
```diff
@@ -134,6 +134,17 @@ impl StorageBackend for FileStorageBackend {
     async fn delete_obj(&self, path: &str) -> Result<(), StorageError> {
         fs::remove_file(path).await.map_err(StorageError::from)
     }
+
+    async fn delete_objs(&self, paths: &[String]) -> Result<(), StorageError> {
```
Dandandan, line 138: I think `paths: &[&str]` would be a slight better signature.

rust/src/storage/mod.rs
```diff
@@ -430,6 +438,9 @@ pub trait StorageBackend: Send + Sync + Debug {
 
     /// Deletes object by `path`.
     async fn delete_obj(&self, path: &str) -> Result<(), StorageError>;
+
+    /// Deletes object by `path`.
+    async fn delete_objs(&self, paths: &[String]) -> Result<(), StorageError>;
```
Dandandan, line 443: We can add a default implementation here, based on `delete_obj`

rust/src/storage/s3/mod.rs
```diff
@@ -416,13 +417,47 @@ impl StorageBackend for S3StorageBackend {
...
+
+        let uris = paths
+            .iter()
+            .map(|path| {
+                let s3_obj = parse_uri(path).unwrap().into_s3object().unwrap();
+                (s3_obj.bucket, s3_obj)
+            })
+            .into_group_map();
+
+        for (bucket, objs) in uris {
```
Dandandan, line 442: This still does only one at a time. 
We should chunk them with maximum 1000 at once
 https://github.com/delta-io/delta-rs/pull/396/files#diff-87e43fff59e072c5dbd99295f15fbf86d5a0e53f6586689191b2c61e07ace91aR439
 
 and also checking whether the buckets are the same for all paths makes sense I think.

rust/Cargo.toml
```diff
@@ -26,6 +26,7 @@ regex = "1"
 chrono = "0"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 lazy_static = "1"
+itertools = "0"
```
Dandandan, line 29: Is this used anywhere?

rust/src/storage/azure.rs
zijie0: Oh yes, let me change this with default implementation.

Dandandan:
I added some comments @zijie0

rust/src/storage/s3/mod.rs
zijie0: Yes, I think it should be the same bucket in vacuum operation.

rust/Cargo.toml
zijie0: It is used for `into_group_map` op. I think your implementation is better and I will remove this later.

zijie0:
@Dandandan Thank you for your suggestion, learned a lot from your PR :)

rust/src/storage/file/mod.rs
zijie0: I originally used `&[&str]` but later met some problem when calling this method from `vacuum`... It said that: temporary value created here. So I think maybe we should use `String` instead of `&str` here?

houqp: Ideally, I think the method should be defined as a generic function taking `impl Iterator<AsRef<str>>` as input. But this trait is being used as trait object, so we can't use generic here, which makes things really annoying. I also think `&[&str]` would have been a cleaner interface, but it will force us to allocate both `Vec<String>` and `Vec<&str>` in vacuum method, which has performance overheads :( I also don't really have a good suggestion on how to workaround this issue here.

zijie0: @Dandandan Do you have any suggestion on this? Or we could go ahead with `&[String]` for now?

Dandandan: I think it's fine 👍

Dandandan:
I think this is a great improvement.

I am wondering a bit how errors in the S3 implementation are handled, as the `notfound` error is not explicitly raised.

But I don't think that should be covered in this PR.

houqp (approved):

houqp:
Thank you both for this new optimization @zijie0 @Dandandan !

PR #124, opened 2021-03-14
Fix update logic.

dispanser:
This prevents duplicate loading of already loaded checkpoints, and
replaying json despite a newer checkpoint being available.

closes #123 

The behavior has been tested using the approach discussed in #123.

1. start: 5 commits `createCommits(5, "<some delta path>")`

```
reading delta table: "<some path>"
[apply_logs_after_current_version] applied 0
[apply_logs_after_current_version] applied 1
[apply_logs_after_current_version] applied 2
[apply_logs_after_current_version] applied 3
[apply_logs_after_current_version] applied 4
[apply_logs_after_current_version] end of log : 5, rollback
initial table loaded, version 4
waiting on keypress for update, current version 4
```

2. 4 more commits: no checkpoints yet

```
updating ...
[update] process started
[update] no previous checkpoint, trying version = 5 ff
[apply_logs_after_current_version] applied 5
[apply_logs_after_current_version] applied 6
[apply_logs_after_current_version] applied 7
[apply_logs_after_current_version] applied 8
[apply_logs_after_current_version] end of log : 9, rollback
waiting on keypress for update, current version 8
```

3. 4 more commits: new checkpoint at v10

```
updating ...
[update] process started
[update] retrieved previous checkpoint CheckPoint { version: 10, size: 13, parts: None }, previous = None
[update] determined to load a checkpoint: CheckPoint { version: 10, size: 13, parts: None }
[restore_checkpoint]: CheckPoint { version: 10, size: 13, parts: None }
[apply_logs_after_current_version] applied 11
[apply_logs_after_current_version] applied 12
[apply_logs_after_current_version] end of log : 13, rollback
waiting on keypress for update, current version 12
```

4. 4 more commits: no new checkpoint, still at v10

```
updating ...
[update] process started
[update] retrieved previous checkpoint CheckPoint { version: 10, size: 13, parts: None }, previous = Some(CheckPoint { version: 10, size: 13, parts: None })
[apply_logs_after_current_version] applied 12
[apply_logs_after_current_version] applied 13
[apply_logs_after_current_version] applied 14
[apply_logs_after_current_version] applied 15
[apply_logs_after_current_version] applied 16
[apply_logs_after_current_version] end of log : 17, rollback
waiting on keypress for update, current version 16
```

OK: no checkpoint loaded, replaying deltas only.

```
updating ...
[update] process started
[...]

Issue #123, opened 2021-03-14
Unexpected behavior in `update()`


dispanser:
I believe there's a subtle bug in the `update()` method, and how it decides on
either loading from a checkpoint or applying the log from json files.

Here's the important bits:

```rust
Ok(last_check_point) => {
	if self.last_check_point.is_none()
		|| self.last_check_point == Some(last_check_point)
	{
		self.last_check_point = Some(last_check_point);
		self.restore_checkpoint(last_check_point).await?;
		self.version = last_check_point.version + 1;
	}
}
```

What this is doing is 
- load from checkpoint if new checkpoint is the same as previously used checkpoint
- if the new checkpoint is different, don't use it but instead apply the json deltas.

I believe what we want instead is use the json deltas if the new checkpoint is the
same as the previous one, and use the checkpoint if it is newer than what we previously
had loaded.

Note that the final result - an up to date delta table - is still achieved, but it
unnecessarily loads checkpoints or json deltas in either scenario, so it could be
more efficient.

I wasn't able to write a self-contained test inside rust, as I'm not sure about
the status of write support (and checkpointing in particular), but I validated
my assumptions by running a spark shell session to generate commits and a rust
session (with some `println!` sparkles) side by side, based on the following
"data generator" in scala:

```scala
def createCommits(numCommits: Int, deltaUrl: String): Unit =
     (0 until numCommits).foreach { c =>
		 Seq(c).toDF("version").write.format("delta").mode("append").save(deltaUrl)
	 }
```

1. start: 5 commits `createCommits(5, "<some delta path>")`
```
reading delta table: "<some delta path>"
[apply_logs_after_current_version] applied 0
[apply_logs_after_current_version] applied 1
[apply_logs_after_current_version] applied 2
[apply_logs_after_current_version] applied 3
[apply_logs_after_current_version] applied 4
[apply_logs_after_current_version] end of log : 5, rollback
initial table loaded, version 4
waiting on keypress for update, current version 4
```

Init: loading the table, all good.

2. 4 more commits: no checkpoints yet
```
updating ...
[update] process started
[update] no previous checkpoint, trying version = 5 ff
[apply_logs_after_current_version] applied 5
[apply_logs_after_current_version] applied 6
[apply_logs_after_current_version] applied 7
[apply_logs_after_current_version] applied 8
[apply_logs_after_current_version] end of log : 9, rollback
waiting on keypress for update, current version 8
```

Expected behavior: start with our current version and apply new comits in sequence.

[...]

PR #398, opened 2021-08-19
Fix duplicates on update call

mosyp:
# Description
Every time the `update` is called, the `Add` action from the latest checkpoints are being applied once more, leading into duplicates. 
As it turned out, it's a regression from https://github.com/delta-io/delta-rs/pull/124 changes.

First commit shows the repro steps, second revers mention changes and resolves the issue

### DIFF
--- a/rust/tests/read_delta_test.rs
+++ b/rust/tests/read_delta_test.rs
@@ -40,6 +40,9 @@ async fn read_delta_table_with_update() {
     let path = "./tests/data/simple_table_with_checkpoint/";
     let table_newest_version = deltalake::open_table(path).await.unwrap();
     let mut table_to_update = deltalake::open_table_with_version(path, 0).await.unwrap();
+    // calling update several times should not produce any duplicates
+    table_to_update.update().await.unwrap();
+    table_to_update.update().await.unwrap();
     table_to_update.update().await.unwrap();
 
     assert_eq!(
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -633,24 +633,10 @@ impl DeltaTable {
 
     /// Load DeltaTable with data from latest checkpoint
     pub async fn load(&mut self) -> Result<(), DeltaTableError> {
-        match self.get_last_checkpoint().await {
-            Ok(last_check_point) => {
-                self.last_check_point = Some(last_check_point);
-                self.restore_checkpoint(last_check_point).await?;
-                self.version = last_check_point.version + 1;
-            }
-            Err(LoadCheckpointError::NotFound) => {
-                // no checkpoint, start with version 0
-                self.version = 0;
-            }
-            Err(e) => {
-                return Err(DeltaTableError::LoadCheckpoint { source: e });
-            }
-        }
-
-        self.apply_logs_from_current_version().await?;
-
-        Ok(())
+        self.last_check_point = None;
+        self.version = -1;
+        self.state = DeltaTableState::default();
+        self.update().await
     }
 
     /// Updates the DeltaTable to the most recent state committed to the transaction log by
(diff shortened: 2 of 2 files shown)


mosyp:
So this is regression from https://github.com/delta-io/delta-rs/issues/123. 
#123 changes are not covered by tests, so before merging we have to cover changes from #123 with tests

xianwill (dismissed):

rtyler:
@dispanser If you're available, I think it would be helpful for you to chime in on this thread :smile:

houqp:
I think the optimization that @dispanser introduced is valid, the bug is caused by incorrect use of `apply_logs_from_current_version` :)

dispanser:
@rtyler : just coming back from vacation today, I'll take a closer look tomorrow.

dispanser:
Indeed, calling `update`with the fixes from #124 produces a duplicate file. Nice catch @mosyp !

However, I believe the fix proposed in this PR only solves this as a side effect of not doing the proper thing in the first place: `restore_checkpoint` clears the state, which has the same cost (and code path) as loading a table from scratch, which is IMHO undesired for a table without any updates..

What the condition basically says is "if the previously loaded latest checkpoint is the same as the currently seen checkpoint, reload from the checkpoint", which leads to the checkpoints being only used if it was current anyway, so doesn't contain anything not already loaded. If the version has advanced past the next checkpoint, we don't use it, which is not what we want because these checkpoints are supposed to allow you skipping all those jsons. Hence the rationale for #124.

As @hougp already mentioned, the root cause for the duplication is calling `apply_logs_from_current_version` with the _current version_ instead of the _next version_. The "restoring from checkpoint" code path works around that by increasing the version number `self.apply_logs_from_current_version`, which we could also do in an `else` branch:

```rust
            Ok(last_check_point) => {
                if self.last_check_point != Some(last_check_point) {
                    self.last_check_point = Some(last_check_point);
                    self.restore_checkpoint(last_check_point).await?;
                    self.version = last_check_point.version + 1;
                } else {
                    self.version += 1;
                }
            }
```

Of course, now we basically do a `+1` in all non-error cases which is a bit odd so it's probably a good idea to move that out. Also, there's `update_incremental`, maybe we just call that?
```

**Comment by houqp** [comment]
Yep, i think it would be cleaner to use `update_incremental` wherever applicable, then remove all the `self.version += 1` code. `apply_logs_from_current_version` is very easy to be missed use based on its name, so we should probably only use it in `update_incremental` as an internal method.

File: rust/tests/read_delta_test.rs [hunk 1]
```diff
@@ -41,6 +41,7 @@ async fn read_delta_table_with_update() {
     let table_newest_version = deltalake::open_table(path).await.unwrap();
     let mut table_to_update = deltalake::open_table_with_version(path, 0).await.unwrap();
     table_to_update.update().await.unwrap();
+    table_to_update.update().await.unwrap();
```
houqp, line 44: could we have a dedicated test case for checking data duplication on repeated updates? that way, future developers won't accidentally remove this extra update call here thinking it's just a typo.

mosyp: done

mosyp: oh, meaning I've added a comment on top of it. But I'd like to also have another test

mosyp:
@dispanser @houqp I got rid of `apply_logs_from_current_version` (it was basically an incomplete version of `update_incremental`) and applied your suggestion. 
So the `update` is pretty straightforward one. 
- If the latest checkpoint is the same - then just call `update_incremental`
- otherwise - clear state and apply latest checkpoint then call `update_incremental`.

It also lead to rework the `load` function. But I've done a little trick there. Before calling the `update_incremental`, I clear the state and set version to `-1` and then call `update_incremental`. This was we have the same behavior as before but with much less code. (the -1 is required for `update_incremental` to start from applying the version 0)

houqp (approved):
nice clean up @mosyp !

PR #399, opened 2021-08-19
Remove version param from create_checkpoint_from_table

### LINKED ISSUE #385: Explore behaviour of creating checkpoints with passed table reference
# Description
A follow up issue from https://github.com/delta-io/delta-rs/pull/384/files#r688514360. The create checkpoint call mutates the given table state, thus we have to make it expected to the caller. There's a few possibilities, like keeping an old version, the one that matches given version or bumping to the latest one (what we have now)

**Use Case**
When creating checkpoints with reference to the DeltaTable

mosyp:
# Description

Fixes #385

And takes into account https://github.com/delta-io/delta-rs/issues/385#issuecomment-900032974 suggestion

### DIFF
--- a/rust/src/checkpoints.rs
+++ b/rust/src/checkpoints.rs
@@ -89,24 +89,15 @@ pub async fn create_checkpoint_from_table_uri(
     Ok(())
 }
 
-/// Loads given `table` at given `version` and creates checkpoints for it.
-/// If given table's version does not match given `version`, then `table` is re-loaded with appropriate one.
-/// Once checkpoint is created, then the `table` is updated to the latest version.
-pub async fn create_checkpoint_from_table(
-    table: &mut DeltaTable,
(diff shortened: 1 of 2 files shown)


xianwill (approved):

rust/src/checkpoints.rs
```diff
@@ -172,6 +163,11 @@ fn parquet_bytes_from_state(state: &DeltaTableState) -> Result<Vec<u8>, Checkpoi
         })
         .collect();
 
+    println!("CREATING CHECKPOINT");
+    for a in state.files().iter() {
+        println!("{}", a.path);
+    }
+
```
houqp, line 170: @mosyp @xianwill looks like these debugging print should have been removed?

mosyp: yes, u're right, will remove then

PR #400, opened 2021-08-19
Add examples for reading delta table with Rust API.

zijie0:
# Description

Add examples for reading delta table with Rust API. Might be helpful for newcomers.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -22,6 +22,17 @@ objection store communication. Please file Github issue to request for critical
 openssl upgrade.
 
 
+Examples
+------------
+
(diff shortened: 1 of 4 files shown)


python/README.md
```diff
@@ -22,6 +22,17 @@ objection store communication. Please file Github issue to request for critical
 openssl upgrade.
 
 
+Examples
+------------
+
+The examples folder shows how to use Rust API to manipulate Delta tables.
```
rtyler, line 28: This README file change I think should be in `rust/README.md` rather than the `python/` one.

rtyler (changes requested):

houqp (approved):

PR #403, opened 2021-08-21
Use `tokio::fs::rename` in `put_obj`.

zijie0:
# Description

As a follow up of #367 , use syscall rename in `put_obj` to avoid race conditions.

# Related Issue(s)

- related to #361

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1044,7 +1044,7 @@ impl DeltaTable {
         // move temporary commit file to delta log directory
         // rely on storage to fail if the file already exists -
         self.storage
-            .rename_obj(&commit.uri, &self.commit_uri_from_version(version))
+            .rename_obj_noreplace(&commit.uri, &self.commit_uri_from_version(version))
             .await
             .map_err(|e| DeltaTableError::from(DeltaTransactionError::from(e)))?;
 
--- a/rust/src/storage/azure.rs
+++ b/rust/src/storage/azure.rs
@@ -193,8 +193,8 @@ impl StorageBackend for AdlsGen2Backend {
         unimplemented!("put_obj not implemented for azure");
     }
 
-    async fn rename_obj(&self, _src: &str, _dst: &str) -> Result<(), StorageError> {
-        unimplemented!("rename_obj not implemented for azure");
+    async fn rename_obj_noreplace(&self, _src: &str, _dst: &str) -> Result<(), StorageError> {
+        unimplemented!("rename_obj_noreplace not implemented for azure");
     }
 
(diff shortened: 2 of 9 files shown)


houqp (commented):
I think the swap argument we added to the atomic_rename method is not needed anymore right?

mosyp (dismissed):

mosyp:
> I think the swap argument we added to the atomic_rename method is not needed anymore right?

Feels like it, that was the only usage

zijie0:
@houqp @mosyp Do you think that we can also remove the swap argument in this PR? If so, I can help with the change.

houqp:
@zijie0 yes, let's do that :)

zijie0:
I also refactor the method name to `rename_if_not_exists`. If there's any problems, please let me know.

mosyp:
Tbh, not a fan of `rename_if_not_exists` naming, as it differs to the API we're calling, and it's basically a wrapping call. But that's just imho. @houqp wdyt?

mosyp (approved):
other than that, lgtm

zijie0:
Maybe `rename_noreplace` could be another option? :)

houqp:
`rename_noreplace` works for me :) I don't have a strong opinion on the naming as long as it differentiates from just `rename`, which is ambigous and could mean replace or no replace.

@mosyp when you said `as it differs to the API we're calling`, are you referring to `StorageBackend::rename`? If so, I think we should change the trait and rename that method to a more explicit name as well.

mosyp (changes requested):
Okay, having another thoughts on that, this is going to be a breaking change. No replace behavior will not work with `_last_checkpoint` file. As it's rewritten on each new checkpoint. 
Moreover, s3 backend is always replace. I discourage having different behavior for the same API function. 

There's no test failing as main is not having the latest arrow changes, that are required for checkpoints to fully work. But this branch does https://github.com/delta-io/delta-rs/tree/writer-map-support. There's test case which rewrites `_last_checkpoint` that should fails with changes in this PR

zijie0:
@mosyp I think `tokio::fs::rename` which is used in `put_obj` should support overwrite by default?

mosyp (approved):
@zijie0 Ah you're right, got confused on my coffee-less morning. Looking good then, I've just found that we have a test for this behavior https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/file/mod.rs#L145

mosyp:
@zijie0 thank you!

houqp:
Thank you @zijie0 and thank you @mosyp for the detailed review!

PR #405, opened 2021-08-21
Update pyproject definition in pyproject.toml

fvaleye:
# Description
- Following these PR discussions: https://github.com/delta-io/delta-rs/pull/381 and https://github.com/delta-io/delta-rs/pull/378

### DIFF
--- a/python/stubs/deltalake/deltalake.pyi
+++ b/python/stubs/deltalake/deltalake.pyi
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, Callable
 
 RawDeltaTable: Any
-rust_core_version: Any
+rust_core_version: Callable[[], str]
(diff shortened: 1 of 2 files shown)


python/stubs/deltalake/deltalake.pyi
```diff
@@ -1,4 +1,4 @@
 from typing import Any
 
 RawDeltaTable: Any
-rust_core_version: Any
\ No newline at end of file
+rust_core_version: str
```
houqp, line 4: I think this should annotated as a callable that returns a str?

houqp (commented):
:+1: The rest looks good to me. I wish there is a linter for pyproject.toml :)

houqp (approved):

Issue #404, opened 2021-08-21
Add additional documentation to setting up python development


Smurphy000:
# Description
Currently trying to setup the dev environment for python using maturin, however, based on the current documentation in the readme and looking for other sources, I can't seem to find a way to easily install the optional dependencies listed in the pyproject.toml file. 

This began after I successfully ran `make develop` but then tried running `make unit-test` where I release pytest was under optional dependencies and no optional dependencies were install with the `make develop` command

**Use Case**
Quickly get the dev environment setup including testing, formatting, and other optional dependencies

Smurphy000:
https://github.com/PyO3/maturin/issues/604

Smurphy000:
From linked issue, I overcame the current concern using `make build` followed by `pip install <path-to-wheel>/example.whl[devel]` or other forms could work. If that is the preferred way for now I could add that to the readme

houqp:
Yes, this is how I setup my local env as well. Thanks for helping with documentation.

fvaleye:
> From linked issue, I overcame the current concern using `make build` followed by `pip install <path-to-wheel>/example.whl[devel]` or other forms could work. If that is the preferred way for now I could add that to the readme

Yes, good idea! The `README.md` needs to be updated to include the optional dependencies for the local environment. You could take a look at the Python build Github workflow file: https://github.com/delta-io/delta-rs/blob/main/.github/workflows/python_build.yml#L58. Thanks for helping!

Smurphy000:
@fvaleye are we good to close this?

fvaleye:
> @fvaleye are we good to close this?

Yes, let me do this for you!

PR #406, opened 2021-08-22
Update python readme

Smurphy000:
# Description
This change was made to make it easier for new developers to quickly setup a dev environment for working on the python api portion of delta-rs

# Related Issue(s)
#404 

The command `make setup-dev` may not work if multiple versions of the whl file are created from `maturin build` in the `/target/wheels/` directory

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -2,7 +2,8 @@
 
 VENV := venv
 MATURIN_VERSION := $(shell awk -F '[ ="]+' '$$1 == "requires" { print $$4 }' pyproject.toml)
-TARGET_WHL := $(wildcard ../target/wheels/*.whl)
+PACKAGE_VERSION := $(shell cargo pkgid | cut -d\# -f2 | cut -d: -f2)
+TARGET_WHEEL := $(shell ls ../target/wheels/deltalake-${PACKAGE_VERSION}-*.whl)
 
 .PHONY: setup-venv
 setup-venv: ## Setup the virtualenv
@@ -16,13 +17,9 @@ setup: ## Setup the requirements
 
 .PHONY: develop
 develop: setup ## Install Python binding of delta-rs
-	$(info --- Maturin develop ---)
-	maturin develop
-
-.PHONY: setup-dev
-setup-dev: setup
+	$(info --- Maturin build ---)
 	maturin build
-	python -m pip install $(TARGET_WHL)[devel,pandas]
+	python -m pip install $(TARGET_WHEEL)[devel,pandas]
 
 .PHONY: format
 format: ## Format the code
--- a/python/Makefile
+++ b/python/Makefile
@@ -19,10 +19,6 @@ develop: setup ## Install Python binding of delta-rs
 	$(info --- Maturin develop ---)
 	maturin develop
 
-.PHONY: find-whl
-find-whl:
-	$(wildcard ../target/wheels/*.whl)
-
 .PHONY: setup-dev
 setup-dev: setup
 	maturin build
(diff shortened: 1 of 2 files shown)


python/Makefile
```diff
@@ -2,6 +2,7 @@
 
 VENV := venv
 MATURIN_VERSION := $(shell awk -F '[ ="]+' '$$1 == "requires" { print $$4 }' pyproject.toml)
+TARGET_WHL := $(wildcard ../target/wheels/*.whl)
```
houqp, line 5: i wonder if this will result in any issue if there are multiple wheel files with different versions?

houqp (approved):
Thanks for helping with documentation, I will leave it to @fvaleye to merge it.

python/Makefile
Smurphy000: I had the same thought, if there are any suggestions for picking the latest version or its on the dev to only have one version in the target/wheels directory.

fvaleye, line 5: We could retrieve the version of the `Cargo.toml` file using the command`cargo pkgid`.

 ```suggestion
PACKAGE_VERSION := $(shell cargo pkgid | cut -d\# -f2 | cut -d: -f2)
TARGET_WHEEL := $(shell ls ../target/wheels/deltalake-${PACKAGE_VERSION}-*.whl)
```

File: python/Makefile [hunk 2]
```diff
@@ -18,6 +19,11 @@ develop: setup ## Install Python binding of delta-rs
 	$(info --- Maturin develop ---)
 	maturin develop
 
+.PHONY: setup-dev
+setup-dev: setup
+	maturin build
+	python -m pip install $(TARGET_WHL)[devel,pandas]
+
```
fvaleye, line 26: Since the `maturin develop` command does not work because the optional dependencies are required (optional dependencies are required... 😄) to set up our local development environment, I would suggest replacing the [develop](https://github.com/delta-io/delta-rs/blob/83aeabbaed788848c2e8091de6cc4182518b91ec/python/Makefile#L17) which is useless by something like this:

```
.PHONY: develop
develop: setup ## Install Python binding of delta-rs
	$(info --- Maturin build ---)
	maturin build
	pip install $(TARGET_WHEEL)'[devel, pandas]'

```

python/README.md
```diff
@@ -45,6 +45,12 @@ To install development version of the package into your current Python environme
 $ make develop
 ```
 
+To setup development and include optional dependencies
+
+```bash
+$ make setup-dev
+```
+
```
fvaleye, line 53: Depending on my previous comment, you could remove this in the `README.md`

fvaleye:
@Smurphy000 Thx for the PR 👍 I added some comments!

Following this, you could also use the Makefile for the Github Actions [here](https://github.com/delta-io/delta-rs/blob/main/.github/workflows/python_build.yml#L60-L64).

Smurphy000:
@fvaleye what you referenced in the yml file does have a setting `maturin build --manylinux off` which differs from the make command currently. I'm not sure what the impact would be just replacing that run script with the modified make command and I am unsure how I could test its results are the same

python/Makefile
Smurphy000: I was worried about replacing an existing command, but it makes sense to me 😄

python/Makefile
```diff
@@ -15,8 +17,9 @@ setup: ## Setup the requirements
 
 .PHONY: develop
 develop: setup ## Install Python binding of delta-rs
-	$(info --- Maturin develop ---)
-	maturin develop
+	$(info --- Maturin build ---)
+	maturin build
+	python -m pip install $(TARGET_WHEEL)[devel,pandas]
```
fvaleye, line 22: You could also use pip directly.

```suggestion
    pip install $(TARGET_WHEEL)[devel,pandas]
```

fvaleye (approved):
LGTM! Thx @Smurphy000

PR #414, opened 2021-08-24
Add filesystem argument for reading DeltaTable in Python binding

fvaleye:
# Description
- Expose `filesystem` argument for reading DeltaTable with Pyarrow in the Python binding

# Related Issue(s)
#392

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -191,7 +191,7 @@ class DeltaTable:
         return pyarrow_schema_from_json(self._table.arrow_schema_json())
 
     def to_pyarrow_dataset(
-        self, partitions: Optional[List[Tuple[str, str, Any]]] = None, filesystem: FileSystem = None
+        self, partitions: Optional[List[Tuple[str, str, Any]]] = None, filesystem: Optional[FileSystem] = None
     ) -> pyarrow.dataset.Dataset:
         """
         Build a PyArrow Dataset using data from the DeltaTable.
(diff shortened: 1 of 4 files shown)


houqp (approved):
looks like there is a code conflict

python/deltalake/table.py
```diff
@@ -190,15 +191,16 @@ def pyarrow_schema(self) -> pyarrow.Schema:
         return pyarrow_schema_from_json(self._table.arrow_schema_json())
 
     def to_pyarrow_dataset(
-        self, partitions: Optional[List[Tuple[str, str, Any]]] = None
+        self, partitions: Optional[List[Tuple[str, str, Any]]] = None, filesystem: FileSystem = None
```
houqp, line 194: The type annotation here should be `Optional[FileSsytem]`

houqp (approved):

mattc-eostar:
I am not sure that this addresses the specific issue with how ADLS needs a container prepended to the file names for use within the file system object as described the issue related to this (#392).

prasadvaze:
@mattc-eostar  is correct. We still need to prepend container-name to the file names. So the issue is still open.  @fvaleye 

In the meantime tried using blobfuse instead of adlfs to connect and mount ADLS on ubuntu. The performance is much better (2 sec versus 6 sec to query ~50k rows delta table.)  And not required to use to_pyarrow_dataset( )

PR #417, opened 2021-08-30
Fix reading nullable extended_file_metadata in Remove action

mosyp:
# Description
Although extended_file_metadata is nullable, it's not in parquet parser


rust/src/action.rs
```diff
@@ -542,6 +542,9 @@ pub struct Remove {
     /// example, since it adds and removes files by combining many files into one.
     pub data_change: bool,
     /// When true the fields partitionValues, size, and tags are present
+    ///
+    /// NOTE: Although it's defined as required in scala delta implementation, but some writes
+    /// it's still nullable so we keep it as Option<> for compatibly.
     pub extended_file_metadata: Option<bool>,
```
mosyp, line 548: In scala it's defined as required field https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L321. But some spark writers still write it's as nullable. (We also have some examples in out test tabl, e.g. https://github.com/delta-io/delta-rs/blob/main/rust/tests/data/simple_table/_delta_log/00000000000000000001.json#L3 )

xianwill (approved):
LGTM

rust/src/action.rs
houqp: looks like we should send a PR to update https://github.com/delta-io/delta/blob/master/PROTOCOL.md  too

PR #418, opened 2021-08-30
Add S3StorageOptions to allow configuring S3 backend explicitly

xianwill:
# Description

Add S3StorageOptions to allow configuring S3 backend explicitly (rather than through env).

This enables different backend configuration for two different delta tables (which is a need in kafka-delta-ingest for dead lettering to delta). With this change, we can configure two different storage backends with different parameters (specifically DynamoDb lock name difference is needed for our use case) and use the `DeltaTable::new` factory (https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L1038).

# Related Issue(s)

Does not reference an existing delta-rs issue, but required for dead lettering to delta in kafka-delta-ingest - https://github.com/delta-io/kafka-delta-ingest/pull/69

### DIFF
--- a/rust/src/storage/s3/dynamodb_lock.rs
+++ b/rust/src/storage/s3/dynamodb_lock.rs
@@ -10,36 +10,32 @@ use rusoto_core::RusotoError;
 use rusoto_dynamodb::*;
 use uuid::Uuid;
 
-mod options {
-    /// Environment variable for `partition_key_value` option.
-    pub const PARTITION_KEY_VALUE: &str = "DYNAMO_LOCK_PARTITION_KEY_VALUE";
-    /// Environment variable for `table_name` option.
-    pub const TABLE_NAME: &str = "DYNAMO_LOCK_TABLE_NAME";
-    /// Environment variable for `owner_name` option.
-    pub const OWNER_NAME: &str = "DYNAMO_LOCK_OWNER_NAME";
-    /// Environment variable for `lease_duration` option.
-    pub const LEASE_DURATION: &str = "DYNAMO_LOCK_LEASE_DURATION";
-    /// Environment variable for `refresh_period` option.
-    pub const REFRESH_PERIOD_MILLIS: &str = "DYNAMO_LOCK_REFRESH_PERIOD_MILLIS";
-    /// Environment variable for `additional_time_to_wait_for_lock` option.
-    pub const ADDITIONAL_TIME_TO_WAIT_MILLIS: &str = "DYNAMO_LOCK_ADDITIONAL_TIME_TO_WAIT_MILLIS";
+/// DynamoDb option keys to use when creating DynamoDbOptions.
+/// The same key should be used whether passing a key in the hashmap or setting it as an environment variable.
+pub mod dynamo_lock_options {
+    /// Used as the partition key for DynamoDb writes.
+    /// This should be the same for all writers writing against the same S3 table.
+    pub const DYNAMO_LOCK_PARTITION_KEY_VALUE: &str = "DYNAMO_LOCK_PARTITION_KEY_VALUE";
+    /// The DynamoDb table where locks are stored. Must be the same between clients that require the same lock.
+    pub const DYNAMO_LOCK_TABLE_NAME: &str = "DYNAMO_LOCK_TABLE_NAME";
+    /// Name of the task that owns the DynamoDb lock. If not provided, defaults to a UUID that represents the process performing the write.
+    pub const DYNAMO_LOCK_OWNER_NAME: &str = "DYNAMO_LOCK_OWNER_NAME";
+    /// Amount of time to lease a lock. If not provided, defaults to 20 seconds.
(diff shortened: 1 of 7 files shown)


rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -38,34 +51,61 @@ pub struct Options {
...
-        fn str_env(key: &str, default: String) -> String {
-            std::env::var(key).unwrap_or(default)
+        let empty_map = HashMap::new();
+        Self::from_map(&empty_map)
+    }
+}
+
+impl DynamoDbOptions {
+    /// Creates a new DynamoDb options from the given map.
+    /// Keys not present in the environment are taken from the environment variable.
```
xianwill, line 69:
```suggestion
    /// Keys not present in the map are taken from the environment variable.
```

rust/src/storage/s3/mod.rs
```diff
@@ -22,10 +23,107 @@ use uuid::Uuid;
 
 pub mod dynamodb_lock;
 
-const AWS_S3_ASSUME_ROLE_ARN: &str = "AWS_S3_ASSUME_ROLE_ARN";
-const AWS_S3_ROLE_SESSION_NAME: &str = "AWS_S3_ROLE_SESSION_NAME";
+const AWS_ENDPOINT_URL: &str = "AWS_ENDPOINT_URL";
 const AWS_WEB_IDENTITY_TOKEN_FILE: &str = "AWS_WEB_IDENTITY_TOKEN_FILE";
```
rtyler, line 27: Why are these also not considered options for use with the S3StorageOptions? :confused:  (Particularly endpoint url since that's got some behavioral meaning later on)

rust/src/storage/s3/mod.rs
```diff
@@ -22,10 +23,107 @@ use uuid::Uuid;
...
+impl S3StorageOptions {
+    /// Creates an instance of S3StorageOptions from environment variables.
+    pub fn from_env() -> S3StorageOptions {
+        let empty_opts = HashMap::new();
+        Self::from_map(&empty_opts)
+    }
+
+    /// Creates an instance of S3StorageOptions from the given HashMap
+    pub fn from_map(options: &HashMap<String, String>) -> S3StorageOptions {
+        fn str_or_default(map: &HashMap<String, String>, key: &str, default: String) -> String {
```
rtyler, line 85: what's with the functions in your functions? As opposed to non public functions within the `impl`?

rust/src/storage/s3/mod.rs
```diff
@@ -22,10 +23,107 @@ use uuid::Uuid;
...
+    assume_role_arn: Option<String>,
+    role_session_name: Option<String>,
+    use_web_identity: bool,
+    locking_provider: Option<String>,
+    extra_opts: HashMap<String, String>,
+}
+
+impl S3StorageOptions {
+    /// Creates an instance of S3StorageOptions from environment variables.
+    pub fn from_env() -> S3StorageOptions {
```
rtyler, line 78: I might recommending moving this to a `Default` trait implementation.

rust/src/storage/s3/mod.rs
xianwill: i actually started w endpoint url as an option but that var is only set when testing against localstack and only used to create the region. given that, i felt like it was awkward to include that as part of the public api. we already set the env var in ci so i feel like omitting it from the options struct is a cleaner public api for consumers. i dont think anybody consuming delta-rs should need to use it

rust/src/storage/s3/mod.rs
xianwill: :xzibit:

hoisted these up.

rust/src/storage/s3/mod.rs
xianwill: :heavy_check_mark:

rust/src/storage/s3/mod.rs
xianwill: RE: `AWS_ENDPOINT_URL` - I moved this to S3StorageOptions in the latest commit in case there are other use cases for custom S3 endpoints. The URL is only used for creating the `region`, but I added an unused `_endpoint_url` field for information/debug needs.

RE: `AWS_WEB_IDENTITY_TOKEN_FILE` - these must be set in the env for the rusoto web identity provider to work the way it is called today (`from_k8s_env`). I went ahead added these to the `s3_storage_options` keys along w/ some code to set them in env if they are passed in options map.

rtyler: `AWS_ENDPOINT_URL` will be required for anybody working with an S3 compatible API but not using S3 directly, which is why I think it should be included. Yes we're using it for localstack, but excluding it from the options limits the potential for users to use non-S3 storage providers with this code :shrug:

xianwill: Cool. Its in the options now so I think we are good.

rtyler (approved):
This looks fine to me but a committer with some smarts should actually look at it before merging :)

mosyp (approved):
LGTM, love it!

fvaleye (approved):
Nice! 👍

PR #420, opened 2021-09-01
Introduce DeltaConfig and tombstones retention policy

mosyp:
There's a lot of delta configs which are ignored by delta-rs. This PR introduces the mechanism of working with these configs with the similar manner as in spark codebase https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L325.

When working with kafka-delta-ingest in production we noticed that removes logs are never cleared, however they are in spark writer, due to the config mention above. 

Note that the `vaccum` function has not been modified in this PR (however it has to use the retention interval value from the configs as default instead of const value). Because when comparing it with spark `vaccum`, we found that spark deletes **every** file which is in the table but not referenced in delta logs. 
Both delta logs entries and actual files time to live is controlled by `deletedFileRetentionDuration` which means, with the current version of vaccum we might end up with orphan files in the store. E.g. the vaccum called after the `remove` action is cleared from the log.

Also, there's bunch of other configs that are turned on by default in spark, such as `enableExpiredLogCleanup`.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -210,16 +210,17 @@ class DeltaTable:
         """
         return self._metadata
 
-    def vacuum(self, retention_hours: int, dry_run: bool = True) -> List[str]:
+    def vacuum(self, retention_hours: Optional[int] = None, dry_run: bool = True) -> List[str]:
         """
         Run the Vacuum command on the Delta Table: list and delete files no longer referenced by the Delta table and are older than the retention threshold.
 
-        :param retention_hours: the retention threshold in hours
+        :param retention_hours: the retention threshold in hours, if none then the value from `configuration.deletedFileRetentionDuration` is used or default of 1 week otherwise.
         :param dry_run: when activated, list only the files, delete otherwise
         :return: the list of files no longer referenced by the Delta Table and are older than the retention threshold.
         """
-        if retention_hours < 0:
-            raise ValueError("The retention periods should be positive.")
+        if retention_hours:
+            if retention_hours < 0:
+                raise ValueError("The retention periods should be positive.")
 
         return self._table.vacuum(dry_run, retention_hours)
 
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -20,7 +20,7 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fmt;
 use std::io::{BufRead, BufReader, Cursor};
-use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::time::{SystemTime, UNIX_EPOCH};
 use std::{cmp::Ordering, collections::HashSet};
 use uuid::Uuid;
 
@@ -28,10 +28,12 @@ use crate::action::Stats;
 
 use super::action;
 use super::action::{Action, DeltaOperation};
+use super::delta_config;
 use super::partitions::{DeltaTablePartition, PartitionFilter};
 use super::schema::*;
 use super::storage;
 use super::storage::{parse_uri, StorageBackend, StorageError, UriError};
+use crate::delta_config::DeltaConfigError;
 
 /// Metadata for a checkpoint file
 #[derive(Serialize, Deserialize, Debug, Default, Clone, Copy)]
@@ -311,6 +313,13 @@ pub enum ApplyLogError {
         /// Storage error details returned while reading the log content.
         source: StorageError,
     },
+    /// Error returned when reading delta config failed.
+    #[error("Failed to read delta config: {}", .source)]
+    Config {
+        /// Delta config error returned when reading delta config failed.
+        #[from]
+        source: DeltaConfigError,
+    },
     /// Error returned when a line from log record is invalid.
     #[error("Failed to read line from log record")]
     Io {
@@ -371,12 +380,23 @@ pub struct DeltaTableState {
     min_reader_version: i32,
     min_writer_version: i32,
     current_metadata: Option<DeltaTableMetaData>,
+    tombstone_retention_millis: DeltaDataTypeLong,
 }
 
 impl DeltaTableState {
     /// Full list of tombstones (remove actions) representing files removed from table state).
-    pub fn tombstones(&self) -> &Vec<action::Remove> {
+    pub fn all_tombstones(&self) -> &Vec<action::Remove> {
         self.tombstones.as_ref()
+    }
+
+    /// List of unexpired tombstones (remove actions) representing files removed from table state.
+    /// The retention period is set by `deletedFileRetentionDuration` with default value of 1 week.
+    pub fn unexpired_tombstones(&self) -> Vec<&action::Remove> {
+        let retention_timestamp = Utc::now().timestamp_millis() - self.tombstone_retention_millis;
+        self.tombstones
+            .iter()
+            .filter(|t| t.deletion_timestamp.unwrap_or(0) > retention_timestamp)
+            .collect()
     }
 
     /// Full list of add actions representing all parquet files that are part of the current
@@ -867,9 +887,9 @@ impl DeltaTable {
             .ok_or(DeltaTableError::NoMetadata)
     }
 
-    /// Returns a vector of tombstones (i.e. `Remove` actions present in the current delta log.
-    pub fn get_tombstones(&self) -> &Vec<action::Remove> {
-        &self.state.tombstones
+    /// Returns a vector of tombstones (i.e. `Remove` actions present in the current delta log).
+    pub fn get_tombstones(&self) -> Vec<&action::Remove> {
+        self.state.unexpired_tombstones()
     }
 
     /// Returns the current version of the DeltaTable based on the loaded metadata.
@@ -890,21 +910,29 @@ impl DeltaTable {
     }
 
     /// List files no longer referenced by a Delta table and are older than the retention threshold.
-    fn get_stale_files(&self, retention_hours: u64) -> Result<HashSet<&str>, DeltaTableError> {
-        if retention_hours < 168 {
+    fn get_stale_files(
+        &self,
+        retention_hours: Option<u64>,
+    ) -> Result<HashSet<&str>, DeltaTableError> {
+        let retention_millis = retention_hours
+            .map(|hours| 3600000 * hours as i64)
+            .unwrap_or(self.state.tombstone_retention_millis);
+
+        if retention_millis < self.state.tombstone_retention_millis {
             return Err(DeltaTableError::InvalidVacuumRetentionPeriod);
         }
-        let before_duration = (SystemTime::now() - Duration::from_secs(3600 * retention_hours))
-            .duration_since(UNIX_EPOCH);
-        let delete_before_timestamp = match before_duration {
-            Ok(duration) => duration.as_millis() as i64,
-            Err(_) => return Err(DeltaTableError::InvalidVacuumRetentionPeriod),
-        };
+
+        let tombstone_retention_timestamp = Utc::now().timestamp_millis() - retention_millis;
 
         Ok(self
-            .get_tombstones()
+            .state
+            .all_tombstones()
             .iter()
-            .filter(|tombstone| tombstone.deletion_timestamp.unwrap_or(0) < delete_before_timestamp)
+            .filter(|tombstone| {
+                // if the file has a creation time before the `tombstone_retention_timestamp`
+                // then it's considered as a stale file
+                tombstone.deletion_timestamp.unwrap_or(0) < tombstone_retention_timestamp
+            })
             .map(|tombstone| tombstone.path.as_str())
             .collect::<HashSet<_>>())
     }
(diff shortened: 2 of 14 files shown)


mosyp:
Leaving it as draft since the checkpoint tests have to be done against `writer-map-support` branch, since `arrow-rs` has not yet release map support.

rust/src/delta.rs
```diff
@@ -660,7 +680,10 @@ impl DeltaTable {
 
     /// Updates the DeltaTable to the latest version by incrementally applying newer versions.
     /// It assumes that the table is already updated to the current version `self.version`.
-    pub async fn update_incremental(&mut self) -> Result<(), DeltaTableError> {
+    ///
+    /// It's private since it's expected for users to call `update` as it takes into account
+    /// checkpoints which handle expired tombstones.
+    async fn update_incremental(&mut self) -> Result<(), DeltaTableError> {
```
mosyp, line 686: It's unsafe to leave `update_incremental` as public, because it'll ignore the checkpoints which might have removed expired tombstones. Basically, since the `update` is fixed and optimized, it is safe to call it instead, e.g. it'll call `update_incremental` if no new checkpoint is found anyway

rust/src/delta.rs
```diff
@@ -371,14 +380,25 @@ pub struct DeltaTableState {
...
 
+    /// List of non-expired tombstones (remove actions) representing files removed from table state.
+    /// The retention period is set by `deletedFileRetentionDuration` with default value of 1 week.
+    pub fn tombstones(&self) -> Vec<action::Remove> {
+        self.tombstones
+            .iter()
+            .filter(|t| t.deletion_timestamp > self.tombstone_retention_timestamp)
+            .map(|t| t.clone())
+            .collect()
+    }
```
mosyp, line 400: Need help on naming things. 

In order to not break things (for current vaccum func), I'm leaving the function to retrieve all tombstones, but renaming it to  `all_tombstones`, whereas `tombstones` will return only active tombstones. This is the default behavior I found in spark delta.

wdyt? Or we should just leave `tombstones` instead and create new `active_tombstones` / `unexpired_tombstones`?

mosyp:
@fvaleye I'd love especially your thoughts on that since you're implemented vaccum command

rust/src/delta.rs
xianwill: imho for naming - i like exactly two functions with specific names -  `all_tombstones` (aka the entire set) and `unexpired_tombstones` (aka the subset of unexpired).

fvaleye:
> @fvaleye I'd love especially your thoughts on that since you're implemented vaccum command

@mosyp I would be interested to see what are the files that remained after applying the `VACUUM` command 😃 

To provide more context on the execution:
When deleting files using the `VACUUM` command, we are listing all files using the DeltaTable URI
https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L941, we exclude files and folders used by Databricks (for optimization: bloom filters, cdc...) https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L916, we are deleting only the files `expired_tombstones` with a hard-coded 1-week limit https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L894, or the `valid_files`still in use.

It was implemented using the `0.8` version of delta, there are new configuration files introduced in the `1.0` version that we should include.

mosyp:
@fvaleye Oh I see, you're absolutely right, sorry for the confusion. 

Then what's left to do is to reuse the retention period form delta_config. 

One more question tho, wdyt of changing the API of vacuum for the parameter `retention_hours: u64` to `retention_hours: Option<u64>`, so, if that's an option then we use the default retention period.+


Also regarding the `if retention_hours < 168` , so in spark I found they use `val retentionSafe = retentionMs.forall(_ >= configuredRetention)` meaning that we might change it to `if retention_hours < self.tombstone_renetion` as well. E.g. to compare against configured in table instead of hardcoded 1 week.

fvaleye:
> @fvaleye Oh I see, you're absolutely right, sorry for the confusion.

@mosyp no worries 👍 
It's a good opportunity for me to see the difference between the `0.8` and `1.0` versions. We have to keep in mind to not remove useful files when applying the `VACUUM` command 😅. 

> 
> Then what's left to do is to reuse the retention period form delta_config.
Exactly!

> 
> One more question tho, wdyt of changing the API of vacuum for the parameter `retention_hours: u64` to `retention_hours: Option<u64>`, so, if that's an option then we use the default retention period.+
>
Good idea, this is the right thing to do to better use the new configuration parameter.
 
> Also regarding the `if retention_hours < 168` , so in spark I found they use `val retentionSafe = retentionMs.forall(_ >= configuredRetention)` meaning that we might change it to `if retention_hours < self.tombstone_renetion` as well. E.g. to compare against configured in table instead of hardcoded 1 week.
Agreed!

python/deltalake/table.py
```diff
@@ -210,17 +210,14 @@ def metadata(self) -> Metadata:
...
         """
         Run the Vacuum command on the Delta Table: list and delete files no longer referenced by the Delta table and are older than the retention threshold.
 
         :param retention_hours: the retention threshold in hours
         :param dry_run: when activated, list only the files, delete otherwise
         :return: the list of files no longer referenced by the Delta Table and are older than the retention threshold.
         """
-        if retention_hours < 0:
-            raise ValueError("The retention periods should be positive.")
-
```
fvaleye, line 223: I included this check in the Python binding to match https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala#L58, but it would be better to move it inside the `VACUUM` rust method.

rust/src/delta.rs
```diff
@@ -890,19 +913,21 @@ impl DeltaTable {
...
 
     /// List files no longer referenced by a Delta table and are older than the retention threshold.
-    fn get_stale_files(&self, retention_hours: u64) -> Result<HashSet<&str>, DeltaTableError> {
-        if retention_hours < 168 {
+    fn get_stale_files(
+        &self,
+        retention_hours: Option<u64>,
+    ) -> Result<HashSet<&str>, DeltaTableError> {
+        let delete_before_timestamp = retention_hours
+            .map(|h| Utc::now().timestamp_millis() - 3600000 * h as i64)
```
fvaleye, line 921: I would suggest to extract `Utc::now().timestamp_millis() - 3600000 * h as i64` into a variable (with an appropriate name) to make reading easier.

python/deltalake/table.py
mosyp: Oh ok, the reason why I removed this check is that 1) I could not figure out how check on optional in python and 2) the call would fail anyway, because if you add the lesser value that default it will fail anyways, even if 0. 

But yeah, I'll re-add this check on rust side then

rust/src/delta.rs
mosyp: ok

mosyp: hmm, well, `h` is depended on the option. Do you suggest to move it to the helper function instead?

python/deltalake/table.py
```diff
@@ -210,16 +210,17 @@ def metadata(self) -> Metadata:
...
 
         :param retention_hours: the retention threshold in hours
         :param dry_run: when activated, list only the files, delete otherwise
         :return: the list of files no longer referenced by the Delta Table and are older than the retention threshold.
         """
-        if retention_hours < 0:
-            raise ValueError("The retention periods should be positive.")
+        if retention_hours:
+            if retention_hours < 0:
+                raise ValueError("The retention periods should be positive.")
```
mosyp, line 223: @fvaleye I had to get this back to python as in rust we have u64, which omits negatives.

btw, does this check looks good?

fvaleye: Yes! 
Could you add a description in the docstring of the parameter `retention_hours` that explains what happens when it is None?

mosyp: sure, done

mosyp:
@xianwill @rtyler @houqp ready for review

rust/src/delta.rs
dispanser: Clarification question: does that mean we _have_ to use snapshots even though a sequence of commits may be substantially smaller, because only checkpoints capture the removal of tombstones?

My understanding is that for a typical query, the tombstone entries are not really applicable or relevant, as only the "live" files are required.

houqp: I have the same question, this seems a bit off to me. `update` still has more overhead compared to `update_incremental` due to extra object store call to get the latest checkpoint and loading the checkpoint itself if it's one checkpoint behind.

Why do we care about whether an expired tombstone has been removed from checkpoint or not? If we are keeping tracking of expired tombstone correctly in memory, we shouldn't be getting any expired tombstone from `unexpired_tombstones` call. I think it's fine to include expired tombestone in `all_tombstones` call returns even if those tombstones are not included in the latest checkpoint.

rust/src/delta.rs
```diff
@@ -869,7 +892,7 @@ impl DeltaTable {
 
     /// Returns a vector of tombstones (i.e. `Remove` actions present in the current delta log.
     pub fn get_tombstones(&self) -> &Vec<action::Remove> {
-        &self.state.tombstones
+        self.state.all_tombstones()
```
houqp, line 895: could we mark this method as deprecated in favor of the new `all_tombstones` method?

houqp (commented):
Left a question around tombstone expiration tracking. The rest looks good to me. The new config interface is really nice and has been a major missing feature :+1:

rust/src/delta.rs
mosyp: These changes was the easiest way to achieve the cleaning tombstones from state. Otherwise if we keep it as is and not address it, it'll overflow the memory. 
^ That's exactly how we found this in production, however from the writer perspective. 

But yeah, if additional call to get the latest checkpoints is overhead we could find other ways. Such as rely on the `CHECKPOINT_INTERVAL` for example to read the checkpoint or clear the state on update. I'll also try to find the spark reader behavior on this as I was focused primarily on a writer.

rust/src/delta.rs
```diff
@@ -371,14 +380,25 @@ pub struct DeltaTableState {
     min_reader_version: i32,
     min_writer_version: i32,
     current_metadata: Option<DeltaTableMetaData>,
+    tombstone_retention_timestamp: DeltaDataTypeTimestamp,
```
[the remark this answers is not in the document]
mosyp: oh you're right, nice catch

rust/src/delta.rs
mosyp: The `get_tombstones`? I wonder which one we should expose to the table level? I lean toward 
table level -> non stale tombstones
state level -> both all and non stale

rust/src/delta.rs
mosyp: I feel like spark does more than that, on each update call, they read every new file from latest checkpoint and if new checkpoint found then it's applied

mosyp: tbh, I'm open to any suggestion on this

rust/src/delta.rs
houqp: oh, my bad, i thought `all_tombstones` is a method of DetltaTable, not the state. I agree with you that we can keep this method and change it to return active tombstones.

rust/src/delta.rs
dispanser: Maybe a read-only application could specify somehow that it's not interested in tombstones at all? IIUC, from a reader perspective the only use case for tombstones would be the time travel feature.

Depending on that config, the `update` method would be able decide if loading a checkpoint is a good idea.

houqp: Ha, the tombstone GC problem is a good call. To me, it seems like the optimal setup would be to delete older in memory expired tombstones after an incremental update. I think there is value to have an incremental update code path for performance sensitive workloads and let users choose which one they want to use depending on the requirements.

houqp: Letting read-only application specifying whether tombstones should be kept at all sounds like a neat trick to reduce memory usage :) I think it's better handled as a separate self-contained PR though because it's an optimization specific to a subset of readers, while the problem we are dealing where applies to both readers and writers.

houqp: If needed, we could also fix tombstone gc for incremental updates in a separate PR since we already have `update` as a good workaround.

houqp: I am having a second thought on @dispanser 's optimization idea, looks like it's more applicable than just a subset of readers? I don't think tombstones are required for time travels since during time travel call, we just start with a clean state and reload the table up to specified versions. I think it's also not required for writers that are not writing out checkpoints, which is the case for how we run KDI. So it is only needed for checkpoint and vacuum?

In order words, we could just drop all tombstones in KDI to further reduce memory usage if it's configured to use the external lambda checkpoint writer?

xianwill: @houqp - fyi: we are using internal checkpoints at this point in our kdi deployment.

at the same time, an in-memory removal of tombstones after a checkpoint doesn't seem very far out of reach. even so, if we leave `update_incremental` public in this PR to not force clients to use `update`, this would allow us to address our short term issue in kdi quickly and follow up with optimization in separate PR.

thoughts?

mosyp: Okay, processing all of the above, I'll leave update_incremental public as it was and we might address performance improvements within an another scope

dispanser: Created #425 to track the no-tombstones idea.

rust/src/delta.rs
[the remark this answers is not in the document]
mosyp: done

rust/src/delta.rs
mosyp: done

mosyp:
@houqp @xianwill ready for the final review

xianwill (dismissed):

xianwill (approved):

houqp (dismissed):
LGTM!

houqp (approved):

PR #427, opened 2021-09-12
Fix checkpoint compatibility for remove fields

xianwill:
# Description

**NOTE: This PR targets the `writer-map-support` branch.** The affecting fields of the remove action are map types, so they are not present or testable on main.

We discovered a production bug in which delta-rs checkpoints that include remove actions added by DBR 7.x (OPTIMIZE), are not readable by DBR cluster versions >= 8.x. The log corruption is due to null extended file metadata fields in the checkpoint schema for remove actions. DBR 7.x does not write extended file metadata for removes. Since they are always included in the checkpoint schema written by delta-rs, we write them as nulls, and DBR 8.x cannot read them back.

This PR addresses the issue by conditionally selecting the remove schema to use. If **all** remove actions in the checkpoint include extended file metadata, we include extended file metadata in the remove schema. Otherwise, we do not.

Four fields are related to remove action extended file metadata. `extendedFileMetadata` specifies whether `partitionValues`, `size` and `tags` are included. Following [comments in the reference implementation](https://github.com/delta-io/delta/blob/fb0452c2fb142310211c6d3604eefb767bb4a134/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L311-L314), new writers should always write this field. The other three fields should be omitted from the schema if `extendedFileMetadata` is false.

# Documentation

* [Delta PROTOCOL add/remove fields section](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file)
* [Reference implementation comments on remove action](https://github.com/delta-io/delta/blob/fb0452c2fb142310211c6d3604eefb767bb4a134/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L311-L314)


houqp:
So am I correct that  DBR 8.x is still trying to read these extension fields even though `extendedFileMetadata` is set to false for a particular remove action? This seems to contradict with the comment in the reference code base.

mosyp:
@houqp So from the testing, it seems like at some of DBR 8.x, if `extendedFileMetadata` is `false` then there's should not be any of `size`, `partitionValues` and `tags` columns. Otherwise, even if they're null, spark fails. Although, I see that BDR 9.0 succeeds with it. But I feel like for the compability purposes we should omit them writing if  `extendedFileMetadata` is false.

mosyp:
@xianwill btw, wouldn't the easier hotfix will be a just write a `extended_file_metadata=false` ?

rust/src/delta_arrow.rs
```diff
@@ -369,12 +372,24 @@ pub(crate) fn delta_log_schema_for_table(
         ));
     }
 
+    // create remove fields with or without extendedFileMetadata
+    let mut remove_fields = REMOVE_FIELDS.clone();
+    if use_extended_remove_schema {
+        remove_fields.extend(REMOVE_EXTENDED_FILE_METADATA_FIELDS.clone());
+    }
+
```
mosyp, line 380: @xianwill btw, wouldn't the easier hotfix will be a just write a extended_file_metadata=false without any other columns

mosyp: however for the full compitability with delta 1.0 I agree that we must include them

rust/src/checkpoints.rs
```diff
@@ -170,6 +170,12 @@ fn parquet_bytes_from_state(state: &DeltaTableState) -> Result<Vec<u8>, Checkpoi
 
     let tombstones = state.unexpired_tombstones();
 
+    // if any, tombstones do not include extended file metadata, we must omit the extended metadata fields from the remove schema
+    // See https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file
+    let use_extended_remove_schema = tombstones
+        .iter()
+        .all(|r| r.extended_file_metadata == Some(true));
```
mosyp, line 177: That is to avoid extending parquet schema with null metadata? E.g. so it'll make DBR 8.x to fail I suppose

rust/src/delta_arrow.rs
xianwill: @mosyp - Just setting `extended_file_metadata` to false doesn't fix the issue. Same error still happens. If its false, the schema for the other three must be omitted entirely.

xianwill:
> So am I correct that DBR 8.x is still trying to read these extension fields even though `extendedFileMetadata` is set to false for a particular remove action? This seems to contradict with the comment in the reference code base.

@houqp - here is comment from reference code base with emphasized portion:

> Note that **for protocol compatibility reasons, the fields `partitionValues`, `size`, and `tags`
   are only present when the extendedFileMetadata flag is true**. New writers should generally be
   setting this flag, but old writers (and FSCK) won't, so readers must check this flag before
   attempting to consume those values.

So I am reading this to mean - if `extendedFileMetadata` is false, you must completely omit the fields. Setting them to null is not good enough. The tests bear that out as well.

rust/src/checkpoints.rs
xianwill: Exactly - this is to avoid writing out the additional fields in the schema and prevent the break in DBR 8.x.

xianwill:
> @houqp So from the testing, it seems like at some of DBR 8.x, if `extendedFileMetadata` is `false` then there's should not be any of `size`, `partitionValues` and `tags` columns. Otherwise, even if they're null, spark fails. Although, I see that BDR 9.0 succeeds with it. But I feel like for the compability purposes we should omit them writing if `extendedFileMetadata` is false.

@mosyp - i tested w/ 9.0 and that failed as well when fields are present.

mosyp (approved):

PR #428, opened 2021-09-13
Add history command in delta-rs

fvaleye:
# Description
- Add CommitInfo to provide history command
- Add history method in Rust / Python

# Documentation
- https://docs.databricks.com/delta/delta-utility.html#delta-history


rust/src/action.rs
```diff
@@ -856,6 +855,238 @@ pub enum OutputMode {
...
+    /// The notebook identifier
+    pub notebook_id: String,
+}
+
+/// CommitInfo holds provenance information about changes to the table. This action
+/// is not stored in the checkpoint and has reduced compatibility guarantees.
+/// Information stored in it is best effort.
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct CommitInfo {
```
houqp, line 887: Here is the quote from the spec:

> Implementations are free to store any valid JSON-formatted data

Looks like we should be modeling the content as `serde_json::Map` instead? The fields we have right now are databricks specific. An alternative implementation could write arbitrary fields, so if we model it as a static Rust struct, we won't be able to capture those.

python/deltalake/history.py
```diff
@@ -0,0 +1,183 @@
...
+
+@dataclass(init=False)
+class CommitInfo:
+    """Create a CommitInfo instance."""
+
+    def __init__(self, commit_info_raw: RawCommitInfoNotebookInfo) -> None:
+        self.commit_info_raw = commit_info_raw
+
+    @property
+    def version(self) -> Optional[int]:
```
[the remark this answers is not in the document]
houqp: On a similar note, if all we have here is just a read only proxy to the `RawCommitInfoNotebookInfo` object, perhaps we can just import `RawCommitInfoNotebookInfo` as `CommitInfoNotebookInfo` in python and avoid the wrapper entirely?

houqp (commented):
Great work @fvaleye! this looks really useful. I left two questions, the rest looks good to me 👍

rust/src/action.rs
fvaleye: Do you think I need to replace the `CommitInfo`  structure with a  `serde_json::Map` which will contain the `CommitInfo` JSON attribute as it is inside the transaction log? This will make it easier to get it inside the Python binding using a simple dictionary. On the other hand, the content of the `CommitInfo` will not be checked.

mosyp (commented):
LGTM, cool and long awaiting features!

However I lean towards the dynamic schema within rust binding for `CommitInfo`

rust/src/action.rs
houqp: Yes, I think so. I think we shouldn't be checking anything in the action content because the spec says it's supposed to be a free formed JSON object. If users can put anything they want into the object, then there is nothing much we can check :)

fvaleye: Done!

python/deltalake/history.py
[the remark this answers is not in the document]
fvaleye: We keep the content of the `CommitInfo` as a dictionary in Python, iit requires less wrapping 😄

rust/src/delta.rs
```diff
@@ -771,6 +771,19 @@ impl DeltaTable {
         }
     }
 
+    /// Returns provenance information, including the operation, user, and so on, for each write to a table.
+    /// Table history is retained for 30 days.
```
houqp, line 775: I believe this 30 is not hardcoded and largely depends on `delta.logRetentionDuration` config?

rust/src/action.rs
```diff
@@ -793,9 +792,7 @@ impl Action {
             "remove" => Action::remove(Remove::from_parquet_record(col_data)?),
             "txn" => Action::txn(Txn::from_parquet_record(col_data)?),
             "protocol" => Action::protocol(Protocol::from_parquet_record(col_data)?),
-            "commitInfo" => {
-                unimplemented!("FIXME: support commitInfo");
-            }
+            "commitInfo" => Action::commitInfo(commit_info_from_parquet_record(col_data)?),
```
houqp, line 795: commit info are excluded from checkpoints, so we can ignore them on checkpoint restore: https://github.com/delta-io/delta/blob/fb0452c2fb142310211c6d3604eefb767bb4a134/core/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala#L251.

commit info maps 1 to 1 to log version, but in checkpoint, all versions are merged into one state, so the concept of commit info is not applicable there.

python/src/lib.rs
```diff
@@ -197,6 +204,20 @@ impl RawDeltaTable {
...
 
+    // Run the History command on the Delta Table: Returns provenance information, including the operation, user, and so on, for each write to a table.
+    pub fn history(&mut self, limit: Option<usize>) -> PyResult<Vec<RawDeltaTableCommitInfo>> {
+        let history = self
+            ._table
+            .history(limit)
+            .map_err(PyDeltaTableError::from_raw)?;
+        Ok(history
+            .iter()
+            .map(|c| RawDeltaTableCommitInfo {
```
houqp, line 215: I don't see a lot of value having a `RawDeltaTableCommitInfo` struct here just wrapping a JSON string. Perhaps we could just return a vector of serialized string here to simplify the code? In the long run, we can also update this method to return a vector of python dictionary to avoid the extra round of JSON parsing in python.

rust/src/delta.rs
```diff
@@ -1108,9 +1115,14 @@ impl DeltaTable {
             "timestamp".to_string(),
             Value::Number(serde_json::Number::from(Utc::now().timestamp_millis())),
         );
+        let mut commit_info = Map::<String, Value>::new();
+        commit_info.insert(
+            "commitInfo".to_string(),
```
houqp, line 1120: I think this would result in `{"commitInfo": "commitInfo": {...}}` when we serialize `Action::commitInfo(commit_info)` to JSON string?

python/deltalake/table.py
```diff
@@ -210,7 +211,22 @@ def metadata(self) -> Metadata:
...
+    def history(self, limit: Optional[int] = None) -> List[CommitInfo]:
+        """
+        Run the history command on the DeltaTable.
+        The operations are returned in reverse chronological order.
+
+        :param limit: the commit info limit to return
+        :return: list of the commit infos registered in the transaction log
+        """
+        return [
+            CommitInfo(commit_info_raw=commit_info_raw)
```
houqp, line 223: I think we can further simplify the code by returning a list of python dictionaries here instead of `CommitInfo` objects. Given commit info is a very dynamic json dictionary that contains arbitrary keys, we don't get much value in creating a specific `CommitInfo` type here.

houqp (commented):
Great simplification! Looks good to me overall. I left couple minor comments because I think there is room for more simplifications :)

python/src/lib.rs
fvaleye: 👍 I agree. 
For now, we can parse the JSON string on the Python side even if it adds overhead. Implementing a direct conversion for a reasonable subset of Python's built-in types to serde_json::Value in PyO3 is a non-trivial amount of work.

rust/src/delta.rs
fvaleye: Good catch, let me fix 😃

rust/src/delta.rs
fvaleye: Right, I will mention that the 30 days is by default!

rust/src/action.rs
fvaleye: Oh, right 👍. Let me fix this!

houqp (approved):
LGTM, great work @fvaleye !

PR #429, opened 2021-09-16
Add is_non_acquirable field to the dynamodb lock

mosyp:
# Description
If there's any outside of delta-rs writes to the table that are done with s3 backend and dynamodb lock, we want the mechanism to immediately stop delta-rs writes if required. 

The example could be the dropping of a table. For example you acquire a lock and then try drop (or drop/create etc., something that modifies the delta log completely) a table. The dangerous part here is that the concurrent delta workers will still perform the write whenever the lock is available, because we effectively lock the `rename` operation. 

To prevent that, this PR introduces a `is_non_acquirable` that will cause delta-rs writer to fail immediately (or to be precise return the specific error) when the remote lock has this field set. For this to work, the outside of delta-rs writer, which performs the dangerous operation, has to set this field by itself once the lock is acquired.

### DIFF
--- a/rust/src/storage/s3/dynamodb_lock.rs
+++ b/rust/src/storage/s3/dynamodb_lock.rs
@@ -133,7 +133,7 @@ impl LockItem {
 }
 
 /// Error returned by the [`DynamoDbLockClient`] API.
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, PartialEq)]
 pub enum DynamoError {
     /// Error caused by the DynamoDB table not being created.
     #[error("Dynamo table not found")]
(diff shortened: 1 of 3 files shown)


xianwill (dismissed):

rust/src/storage/s3/dynamodb_lock.rs
```diff
@@ -154,6 +154,12 @@ pub enum DynamoError {
     #[error("Could not acquire lock for {0} sec")]
     TimedOut(u64),
 
+    /// Error returned by [`DynamoDbLockClient::acquire_lock`] which indicates that the lock could
+    /// not be acquired because the `is_non_acquirable` is set to `true`.
+    /// Usually this is done intentionally outside of [`DynamoDbLockClient`].
```
houqp, line 159: I recommend adding the drop table example you had in the PR description to help readers better understand the use-case.

houqp (approved):

rust/src/storage/s3/dynamodb_lock.rs
mosyp: sure, done

xianwill (approved):

PR #431, opened 2021-09-16
Optimize remove action apply with early iteration exit #424

### LINKED ISSUE #424: Optimize remove action apply with early iteration exit
# Description

**Use Case**

Right now, we use `Iterator::retain` to remove files from active file list when applying remove actions from the transaction log: https://github.com/delta-io/delta-rs/blob/0a05cb45ace7a75ae87bbd4acb3b919653e09233/rust/src/delta.rs#L1484

This can be further optimized by performing an early exit of the iteration when we reach the file that needs to be removed. There is no point going through the remaining of the file list and compare paths since file paths should be unique.

**Related Issue(s)**

akshay26031996:
# Description
Right now, we use `Iterator::retain` to remove files from active file list when applying remove actions from the transaction log. This can be optimized by performing an early exit of the iteration when we reach the file that needs to be removed.

1. Find the position of the file.
2. Remove the file from the vector.

# Related Issue(s)
- closes #424

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1509,7 +1509,10 @@ fn process_action(state: &mut DeltaTableState, action: Action) -> Result<(), App
             state.files.push(v);
         }
         Action::remove(v) => {
-            state.files.retain(|a| *a.path != v.path);
+            let index = { state.files.iter().position(|a| *a.path == v.path) };
+            if let Some(index) = index {
+                state.files.remove(index);
+            }
             state.tombstones.push(v);
         }
         Action::protocol(v) => {
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1511,7 +1511,7 @@ fn process_action(state: &mut DeltaTableState, action: Action) -> Result<(), App
         Action::remove(v) => {
             let index = { state.files.iter().position(|a| *a.path == v.path) };
             if let Some(index) = index {
-                state.files.remove(index);
+                state.files.swap_remove(index);
             }
             state.tombstones.push(v);
         }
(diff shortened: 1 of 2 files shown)


rust/src/delta.rs
```diff
@@ -1509,7 +1509,10 @@ fn process_action(state: &mut DeltaTableState, action: Action) -> Result<(), App
             state.files.push(v);
         }
         Action::remove(v) => {
-            state.files.retain(|a| *a.path != v.path);
+            let index = { state.files.iter().position(|a| *a.path == v.path) };
+            if let Some(index) = index {
+                state.files.remove(index);
```
Dandandan, line 1514: As the order is not important (I believe) `swap_remove` could be used, which is `O(1)` instead of `O(n)`

akshay26031996: I also think that order does not matter (although I am new to this repo). I am pushing the suggested change and the required changes in test. In test, I am sorting the vectors before comparing.

houqp (approved):

rust/src/delta.rs
houqp: That's a nice optimization, good call @Dandandan :)

houqp:
Thanks @akshay26031996 for your contribution.

rust/tests/read_simple_table_test.rs
```diff
@@ -15,14 +15,15 @@ async fn read_simple_table() {
     assert_eq!(table.get_min_writer_version(), 2);
     assert_eq!(table.get_min_reader_version(), 1);
     assert_eq!(
-        table.get_files(),
+        table.get_files().sort(),
```
mosyp, line 18: sort returns `()`

houqp: good catch, @akshay26031996 could you send a PR to correct the tests?

akshay26031996: Noted, let me go through this.

PR #432, opened 2021-09-17
Clean up DeltaTransactionError

mosyp:
The goal of this PR is to remove the confusion when working with tx api. The `DeltaTransactionError` is a bloat (imo), most of it is obsolete and the only `VersionAlreadyExists` is used. These changes remove the `DeltaTransactionError` and moves `VersionAlreadyExists` to the `DeltaTableError`.

The main reason why I opened this is when I've started to work on `DynamoError::NonAcquirableLock` in KDI. 

To get this, I need to traverse the 
```
DeltaTableError -> DeltaTransactionError -> StorageError -> DynamoError
```
but there's also possible way is 
```
DeltaTableError -> StorageError -> DynamoError
```

So which one to use? This brings the confusion as users should traverse through the code to figure out. Also the tx api uses  `DeltaTransactionError` on some functions, but `DeltaTableError` on others.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -183,18 +183,12 @@ pub enum DeltaTableError {
         "Invalid retention period, retention for Vacuum must be greater than 1 week (168 hours)"
     )]
     InvalidVacuumRetentionPeriod,
+    /// Error returned when transaction is failed to be committed because given version already exists.
+    #[error("Delta transaction failed, version {0} already exists.")]
+    VersionAlreadyExists(DeltaDataTypeVersion),
     /// Generic Delta Table error
     #[error("Generic DeltaTable error: {0}")]
     Generic(String),
-
-    /// Error that wraps an underlying DeltaTransaction error.
-    /// The wrapped error describes the specific cause.
-    #[error("Delta transaction failed: {source}")]
-    TransactionError {
-        /// The wrapped DeltaTransaction error.
-        #[from]
-        source: DeltaTransactionError,
-    },
 }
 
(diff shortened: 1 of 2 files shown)


rust/src/delta.rs
```diff
@@ -1463,9 +1415,7 @@ impl<'a> DeltaTransaction<'a> {
                 }
                 Err(e) => {
                     match e {
-                        DeltaTableError::TransactionError {
-                            source: DeltaTransactionError::VersionAlreadyExists { .. },
-                        } => {
+                        DeltaTableError::VersionAlreadyExists(_) => {
```
mosyp, line 1418: Also I like this better. 
btw @houqp, do we have a reason why we use struct like errors instead of value based? I remember something regarding docs but cannot find what exactly

xianwill (approved):
Nice simplification

rust/src/delta.rs
houqp: What do you mean by value based errors?

mosyp: e.g. A vs B in an example:
```rust
enum Error {
  A {
    source: String
  },
  B(String),
}
```

houqp: Oh i see. it's just for thiserror crate so it can implement the from trait automatically for us. See discussion related to From trait in https://docs.rs/thiserror/1.0.29/thiserror/index.html. It's mostly only useful for wrapping error type values though. For types like String, I don't think there is much value in storing it in a source field.

PR #434, opened 2021-09-17
Decode path in Add and Remove actions.

### LINKED ISSUE #433: Open delta table failed with special partition paths.
# Environment

**Delta-rs version**: 0.4.1

**Binding**: Rust

**Environment**: 
- **Cloud provider**: N/A
- **OS**: Windows 10/Ubuntu 20.04
- **Other**: Spark 3.0.3 + delta 0.8.0

***
# Bug

**What happened**:

When reading delta tables with special partition values (such as "A/A", "B B"), the returned files list is incorrect. We will get `FileNotFoundError` error while reading this kind of tables in Python API.

Should be able to read this kind of tables.

**How to reproduce it**:

Write a delta table with special partition values:

```
import pyspark


def delta_params():
    params = {
        "spark.jars.packages": "io.delta:delta-core_2.12:0.8.0",
        "spark.sql.extensions": "io.delta.sql.DeltaSparkSessionExtension",
        "spark.sql.catalog.spark_catalog": "org.apache.spark.sql.delta.catalog.DeltaCatalog"
    }
    return params


def init_spark():
    builder = pyspark.sql.SparkSession.builder.appName("MyApp")
    params = delta_params()
    for k, v in params.items():
        builder.config(k, v)
    spark = builder.getOrCreate()
    return spark


spark = init_spark()
sdf = spark.createDataFrame([['A/A', 1], ['B B', 2]], schema=['x', 'y'])
sdf.write.format("delta").partitionBy("x").save("special_partition")
```

Read the table in Python API:

```
In [2]: from deltalake import DeltaTable

In [3]: dt = DeltaTable("special_partition")

In [4]: df = dt.to_pandas()
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-4-3bcc9db4daed> in <module>
----> 1 df = dt.to_pandas()
```

**More details**:

According to delta spec here: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file

> The path is a URI as specified by RFC 2396 URI Generic Syntax, which needs to be decoded to get the file path.

I think we need to decode the file path in Add & Remove actions.

zijie0:
# Description

Fix file path decode according to delta spec.

Might need another PR to do the path encoding while writing delta log and checkpoints.

# Related Issue(s)

- closes #433


rust/src/action.rs
```diff
@@ -54,6 +55,13 @@ fn gen_action_type_error(action: &str, field: &str, expected_type: &str) -> Acti
     ))
 }
 
+fn decode_path(raw_path: &str) -> String {
+    percent_decode(raw_path.as_bytes())
+        .decode_utf8()
+        .unwrap()
```
houqp, line 61: I think it would be better to return a result in this function instead of a force unwrap.

houqp (approved):
Good catch @zijie0! Could you resolve the conflict with the latest main branch?

houqp:
cc @xianwill @mosyp we need to encode the file path in version commit and checkpoint otherwise it would break the scalar reader.

rust/src/action.rs
zijie0: Sure, I wrapped it with a possible ActionError.

houqp (approved):
LGTM!

fvaleye (approved):
LGTM! Thanks @zijie0

houqp:
Thanks @zijie0 !

FYI @dispanser @mgill25, this could add overhead to table loading, but I feel like it's probably unavoidable.

PR #443, opened 2021-09-22
Fixing test to compare sorted vec

akshay26031996:
# Description
Fixing test to compare sorted vectors.

[Previous PR #431](https://github.com/delta-io/delta-rs/pull/431)


houqp (approved):
thanks @akshay26031996 for the quick turnaround

PR #444, opened 2021-09-22
Batch-apply remove actions in tombstone handling

### LINKED ISSUE #442: Non-linear performance for large size of delta log
In an attempt to create a delta table for #425, I attempted to load a relatively large delta log containing ~ 700k add actions and 9.5million remove actions.

Load performance is extremely slow, and it gets worse the farther inside the load we progress. I've logged the time it takes to handle each separate checkpoint file:

```
  47.85s: loading checkpoint data from f=<delta_root>/_delta_log/00000000000000002710.checkpoint.0000000001.0000000021.parquet
 226.31s: loading checkpoint data from f=<delta_root>/_delta_log/00000000000000002710.checkpoint.0000000002.0000000021.parquet
 541.34s: loading checkpoint data from f=<delta_root>/_delta_log/00000000000000002710.checkpoint.0000000003.0000000021.parquet
 840.58s: loading checkpoint data from f=<delta_root>/_delta_log/00000000000000002710.checkpoint.0000000004.0000000021.parquet
1123.62s: loading checkpoint data from f=<delta_root>/_delta_log/00000000000000002710.checkpoint.0000000005.0000000021.parquet
1402.43s: loading checkpoint data from f=<delta_root>/_delta_log/00000000000000002710.checkpoint.0000000006.0000000021.parquet
1688.98s: loading checkpoint data from f=<delta_root>/_delta_log/00000000000000002710.checkpoint.0000000007.0000000021.parquet
```

It's seems that the load becomes slower and slower the more actions are already loaded. As this is not the case with a table that only includes add actions (see #435: 2 minutes for 5million files, no tombstones), the culprit must be in the remove actions.

```rust
fn process_action(state: &mut DeltaTableState, action: Action) -> Result<(), ApplyLogError> {
    match action {
        Action::add(v) => {
            state.files.push(v);
        }
        Action::remove(v) => {
            state.files.retain(|a| *a.path != v.path);
            state.tombstones.push(v);
        }
	...
	}
	...
}
```

This brings non-linearity: for each add action we seem to make a linear scan through the entire set of adds. The more add actions we have, the longer that takes.

One possible solution would be to collect all actions, build a set of all files that must be removed, and do that in one pass at the very end.

Also, it seems counter-intuitive for a checkpoint to contain both an add and a remove operation for the same file; I suspect that the behavior can be entirely skipped when loading a checkpoint, and only needs to be applied to (relatively small) incremental update operations.

``

dispanser:
Improve tombstone handling by only applying the remove actions as a
batch at the end of a load or incremental update operation.

For checkpoint loading, we entirely skip interpretation of the remove
actions because according to the delta spec, a remove must be in a later
revision than the associated add, so they wouldn't be both in the same
checkpoint.

The PR does not add any new tests, as we are only changing implementation details. I think the proper handling of tombstones is demonstrated by `read_simple_table()` as it makes sure that tombstones are applied properly when loading a table with remove operations.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -428,8 +428,8 @@ impl DeltaTableState {
     }
 
     /// merges new state information into our state
-    pub fn merge(&mut self, new_state: &mut DeltaTableState) {
-        // build a lookup table of all remove actions
+    pub fn merge(&mut self, mut new_state: DeltaTableState) {
+        self.files.append(&mut new_state.files);
 
         if !new_state.tombstones.is_empty() {
             let new_removals: HashSet<&str> = new_state
@@ -438,12 +438,9 @@ impl DeltaTableState {
                 .map(|s| s.path.as_str())
                 .collect();
 
-            // previous state files must filtered for removals,
-            // new state files are appended without looking.
             self.files
                 .retain(|a| !new_removals.contains(a.path.as_str()));
         }
-        self.files.append(&mut new_state.files);
         self.tombstones.append(&mut new_state.tombstones);
         if new_state.min_reader_version > 0 {
             self.min_reader_version = new_state.min_reader_version;
@@ -614,7 +611,7 @@ impl DeltaTable {
             process_action(&mut new_state, action)?;
         }
 
-        self.state.merge(&mut new_state);
+        self.state.merge(new_state);
 
         Ok(())
     }
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -428,7 +428,7 @@ impl DeltaTableState {
     }
 
     /// merges new state information into our state
-    pub fn merge(&mut self, new_state: &mut DeltaTableState) -> () {
+    pub fn merge(&mut self, new_state: &mut DeltaTableState) {
         // build a lookup table of all remove actions
 
         if !new_state.tombstones.is_empty() {
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -426,6 +426,49 @@ impl DeltaTableState {
     pub fn current_metadata(&self) -> Option<&DeltaTableMetaData> {
         self.current_metadata.as_ref()
     }
+
+    /// merges new state information into our state
+    pub fn merge(&mut self, new_state: &mut DeltaTableState) -> () {
+        // build a lookup table of all remove actions
+
+        if !new_state.tombstones.is_empty() {
+            let new_removals: HashSet<&str> = new_state
+                .tombstones
+                .iter()
+                .map(|s| s.path.as_str())
+                .collect();
+
+            // previous state files must filtered for removals,
+            // new state files are appended without looking.
+            self.files
+                .retain(|a| !new_removals.contains(a.path.as_str()));
+        }
+        self.files.append(&mut new_state.files);
+        self.tombstones.append(&mut new_state.tombstones);
+        if new_state.min_reader_version > 0 {
+            self.min_reader_version = new_state.min_reader_version;
+            self.min_writer_version = new_state.min_writer_version;
+        }
+
+        if new_state.current_metadata.is_some() {
+            self.tombstone_retention_millis = new_state.tombstone_retention_millis;
+            self.current_metadata = new_state.current_metadata.take();
+        }
+
+        new_state
+            .app_transaction_version
+            .drain()
+            .for_each(|(app_id, version)| {
+                *self
+                    .app_transaction_version
+                    .entry(app_id)
+                    .or_insert(version) = version
+            });
+
+        if !new_state.commit_infos.is_empty() {
+            self.commit_infos.append(&mut new_state.commit_infos);
+        }
+    }
 }
 
 #[inline]
@@ -565,10 +608,13 @@ impl DeltaTable {
         &mut self,
         reader: BufReader<R>,
     ) -> Result<(), ApplyLogError> {
+        let mut new_state = DeltaTableState::default();
         for line in reader.lines() {
             let action: Action = serde_json::from_str(line?.as_str())?;
-            process_action(&mut self.state, action)?;
+            process_action(&mut new_state, action)?;
         }
+
+        self.state.merge(&mut new_state);
 
         Ok(())
     }
@@ -1474,10 +1520,6 @@ fn process_action(state: &mut DeltaTableState, action: Action) -> Result<(), App
         }
         Action::remove(v) => {
             let v = v.path_decoded()?;
-            let index = { state.files.iter().position(|a| *a.path == v.path) };
-            if let Some(index) = index {
-                state.files.swap_remove(index);
-            }
             state.tombstones.push(v);
         }
         Action::protocol(v) => {


dispanser:
My initial idea was to capture the changes from an update operation in a `struct DeltaChangeSet`, but then this struct had exactly the same fields as `DeltaTableState`, because all these fields can be modified by actions. I then dropped that class, which has the nice advantage that all code paths stay the same, with the exception of `apply_log_from_bufread()`, which happens to be the only place that requires the alignment of add and remove actions.

houqp (commented):
I actually like your initial idea of introducing a change capture struct better ;) But the modification I would make is to make that struct dedicated for batch operations. Today, we only need to batch file list remove, but maybe there will be others in the future.

So I am thinking of something like this:

```rust
struct DeltaBatchChangeSet {
    remove_set: HashSet<&str>,
}

fn process_action(state, batch_change, action) {
    match action {
        ...
        Action::remove(v) => {
            state.tombstones.push(v);
            batch_change.remove_set.add(v.path.as_ref());
        }
        ...
    }
}

fn apply_logs_in_batch(
    state: &mut DeltaTableState,
    actions: impl IntoIterator<Item = Result<Action, ApplyLogError>>,
) -> Result<(), ApplyLogError> {
   let mut batch_change = DeltaBatchChangeSet::new();
   for action in actions {
       process_action(&mut state, &mut batch_change, action?)?;
   } 
   if !batch_change.remove_set.is_empty() {
        state.files
                .retain(|a| !batch_change.remove_set.contains(a.path.as_str()));
   }
   Ok(())
}

fn apply_log_from_bufread(self, reader) {
    apply_logs_in_batch(&mut self.state, reader.lines().map(|line| {
        serde_json::from_str(line?.as_str())
    }))
}
```

This way, we won't need to check and reassign for other fields in the merge method. We also won't need to go through the tombstone for a 2nd time to build the hashset.

What do you think?

dispanser:
> So I am thinking of something like this:
> 
> ```rust
> struct DeltaBatchChangeSet {
>     remove_set: HashSet<&str>,
> }
> 
> fn process_action(state, batch_change, action) {
>     match action {
>         ...
>         Action::remove(v) => {
>             state.tombstones.push(v);
>             batch_change.remove_set.add(v.path.as_ref());
>         }
>         ...
>     }
> }
> 
> fn apply_logs_in_batch(
>     state: &mut DeltaTableState,
>     actions: impl IntoIterator<Item = Result<Action, ApplyLogError>>,
> ) -> Result<(), ApplyLogError> {
>    let mut batch_change = DeltaBatchChangeSet::new();
>    for action in actions.into() {
>        process_action(&mut state, &mut batch_change, action?)?;
>    } 
>    if !batch_change.remove_set.is_empty() {
>         state.files
>                 .retain(|a| !batch_change.remove_set.contains(a.path.as_str()));
>    }
>    Ok(())
> }

The only disadvantage I can see here is that `retain` is called on a larger set, because even the add actions from the current update are considered as remove candidates. On the other hand, this potentially handles moving forward multiple versions at once (calling `retain` ony once).

> This way, we won't need to check and reassign for other fields in the merge method. We also won't need to go through the tombstone for a 2nd time to build the hashset.

Thanks for the great idea! The one thing I didn't like about the approach I've taken is indeed that I'm forced to handle all fields, even though only one of them is actually relevant for the merge operation.

> 
> What do you think?

I'll give it a go this evening. Thanks for the suggestion.

dispanser:
In that idea, we pass `batch_change` into `process_action`, irrespective of it being necessary. I wonder if we should pass in an `Option<DeltaBatchChangeSet>` instead so we don't have to build a potentially large hash set for checkpoint loading.

houqp:
> The only disadvantage I can see here is that retain is called on a larger set, because even the add actions from the current update are considered as remove candidates. On the other hand, this potentially handles moving forward multiple versions at once (calling retain ony once).

This is a good point, I think this optimization is relevant for applying actions from checkpoints. For incremental updates, I think we do still need to consider add actions from the current batch.

Another observation is even though we have to go through the newly added files in the retain call, we avoided the `self.files.append(&mut new_state.files);` call, so the number of times we need to iterate through the newly added batch is the same?

dispanser:
I'm a bit torn here. While your propsal reduces the code size and removes unnecessary handling of unrelated fields in the  change set, making it very clear that (and why) we handle tombstones separately, the initial PR has the advantage that `process_action` just "collects some actions", and introduces the notion of merging. The latter allows for a trivial way to do the checkpoint ingestion concurrently (on a per-file granularity), which would have been one of the next things planned to raise an issue for ;).

houqp:
> The latter allows for a trivial way to do the checkpoint ingestion concurrently (on a per-file granularity), which would have been one of the next things planned to raise an issue for ;).

Yeah, this is a good point. We should definitely design for concurrent apply. So looks like state merge code is unavoidable. In this case, the BatchChangeSet design becomes an optimization details. In other words, we can use it together with state merge later if going through tombstone the 2nd time to build the hashset becomes a performance issue.

In short, :+1: for the current approach :)

rust/src/delta.rs
```diff
@@ -426,6 +426,49 @@ impl DeltaTableState {
...
+            let new_removals: HashSet<&str> = new_state
+                .tombstones
+                .iter()
+                .map(|s| s.path.as_str())
+                .collect();
+
+            // previous state files must filtered for removals,
+            // new state files are appended without looking.
+            self.files
+                .retain(|a| !new_removals.contains(a.path.as_str()));
```
houqp, line 444: @dispanser we need to apply retain for new file list in new_state too if it is constructed from incremental updates instead of checkpoints right?

rust/src/delta.rs
```diff
@@ -426,6 +426,49 @@ impl DeltaTableState {
...
+        }
+
+        if new_state.current_metadata.is_some() {
+            self.tombstone_retention_millis = new_state.tombstone_retention_millis;
+            self.current_metadata = new_state.current_metadata.take();
+        }
+
+        new_state
+            .app_transaction_version
+            .drain()
```
houqp, line 460: if this method drains and mutates new_state, this means new_state is not usable after the call right? if so, I think it's better to pass in the new_state as value instead of reference to take full ownership.

houqp (dismissed):
Left two minor questions, the rest looks good to me :+1: Really nice performance improvement.

rust/src/delta.rs
dispanser: Interesting question. In the current impleentation, `merge` is called by `apply_log_from_bufread`, which is in turn called from `apply_log` for each version separately. So if we assume that no add + remove for the same file happens in the same version, we're safe.

We may be better of doing that "merge" once for multiple commits, and in that case the curent implementation becomes wrong.

I think I'll change it as you suggest because the way it is currently implemented makesa bit of a hidden assumption about how it's called, and the generic `merge` operation should handle all scenarios.

rust/src/delta.rs
dispanser: that sounds reasonable, I'll have a look :+1:

rust/src/delta.rs
houqp: oh yeah, you are right, this is not a problem for the current PR then :+1: we can revisit this when we start doing parallel table load.

dispanser: I changed it to do the append first, which makes the code shorter because it allows me to remove the comments explaining the pecularities of the implementation :-).

One final thought wrt to this topic: we could argue that a `DeltaTableState` must be consistent / self-contained according to some rules, i.e. that "removals for its own files are already resolved". In that case we wouldn't need that.

houqp: yeah, I think that's a good idea. It also avoids having to go through the files field twice, once for append, once for retain.

houqp (approved):

houqp:
The beginning of exciting optimizations :) Thanks @dispanser .

PR #445, opened 2021-09-27
No tombstone loading

dispanser:
# Description

This allows opting out of tombstone-loading, for scenarios / use cases that don't require them; this applies to all applications that are not interested in writing checkpoints, most notably read-only consumers.

The implementation uses a builder pattern `DeltaTableBuilder`, which changes the public-facing API.


dispanser:
If the initial design is OK, I'd incorporate the version / timestamp into the builder, if that makes sense.

houqp:
The design looks great! Have you considered moving storage backend into the option field as well? The only minor nitpick I have is it's a bit odd to declare a load method on the option struct itself.

dispanser:
> Have you considered moving storage backend into the option field as well? 

Not yet, because for the high-level entry points (`open_table` et al) it is derived from the URL.
I assume we want to keep that convenience. I guess we can put it in an `Option<StorageBackend>` and derive-if-None or similar.

I also considered moving the `table_uri` itself into the builder (mandatory, `DeltaTableBuilder::from_path(table_uri)`). Would that make sense as well?

> The only minor nitpick I have is it's a bit odd to declare a load method on the option struct itself.

Now that you mention it, indeed :thinking: . I was coming from the fluent API approach, so my idea of the natural flow is to specify mandatory + optional parameters and then sorta finalize, and get the thing of interest (`DeltaTable` in this case). 

I could certainly rewrite it to have a `DeltaTableLoadOptions` as the final result of a `build` method, which is then fed into a (new) `open_table(load_options)` method. This would also imply I'll have to rename to `DeltaTableLoadOptionsBuilder` though, which has a touch of  java enterprise applications :).

houqp:
> I also considered moving the table_uri itself into the builder (mandatory, DeltaTableBuilder::from_path(table_uri)). Would that make sense as well?

Yes, I think this makes a lot of sense.

> Now that you mention it, indeed thinking . I was coming from the fluent API approach, so my idea of the natural flow is to specify mandatory + optional parameters and then sorta finalize, and get the thing of interest (DeltaTable in this case).

Yes, that's what I have in mind as well, but I think it would be more natural to call `load`/`open` on `DeltaTableBuilder`, instead of `DeltaTableLoadOptions`. `DeltaTableLoadOptions` reads like something I would pass to a load function as argument.

> which has a touch of java enterprise applications :).

Yeah, let's avoid java enterprise naming convention, I get PTSD looking at them. I think both of your suggested approaches would work pretty well and I don't have a strong opinion over one or the other:

1. `DeltaTableBuilder` all the way, i.e. `let table = DeltaTableBuilder::new().uri().without_tombstones().load()`.
2. `DeltaTableLoadOptions` combined with `open_table`, i.e. `let table = open_table(DeltaTableLoadOptions::new().uri().without_tombstones())`

dispanser:
> * `DeltaTableBuilder` all the way, i.e. `let table = DeltaTableBuilder::new().uri().without_tombstones().load()`.
> * `DeltaTableLoadOptions` combined with `open_table`, i.e. `let table = open_table(DeltaTableLoadOptions::new().uri().without_tombstones())`

Oh, I think that's what the API is doing already, [test](https://github.com/delta-io/delta-rs/pull/445/files#diff-89738679c465beaf3c59903e7a7ce964679abb63e276df891911f468de2d3ba9R58):
```rust
    let table = DeltaTableBuilder::new()
        .without_tombstones()
        .load("./tests/data/delta-0.8.0")
```

I think it's clear now how to proceed - might not happen before weekend though, it's sometimes difficult to find a slot for OSS :).

houqp:
yeah, I feel you, I am in the same boat :)

rust/src/delta.rs
```diff
@@ -489,6 +489,23 @@ fn extract_rel_path<'a, 'b>(
     }
 }
 
+/// possible version specifications for loading a delta table
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DeltaVersion {
```
houqp, line 494: :+1:

houqp (commented):
The new builder interface looks solid :)

houqp:
@dispanser is this PR ready to merge? or are you still working on it?

houqp:
oh no, there is a conflict in the test :(

I think we can skip the python binding for now. This feature can be exposed to downstream bindings as separate PRs.

Perhaps one simplification we could make is to move `require_tombstones` into a field inside the `DeltaTable` struct, then we can get rid of the `DeltaTableConfig` struct entirely and just pass in `DeltaTableLoadOptions` as an argument into `DeltaTable::new` method.

dispanser:
@houqp : I'd rather keep the separate config, as I foresee potentially more things ending up in there - at the moment indeed the single option looks a bit lonely in there :). An advantage I see is that we can provide `Default`, so users of the API don't have to read up on these settings if they don't want to.

But ultimately I don't feel experienced enough in Rust and Delta to make a call here, single user, single opinion ;).

houqp:
Let's keep it as is then. I don't have a strong opinion on this either and we can iterate on the UX as we get more data points from actually using it. This new builder api looks great and makes it really easy to add backwards compatible changes going forward. Great work! We should start using it everywhere :D

PR #447, opened 2021-09-30
Run all tests under s3 feature flag

mosyp:
Some of the s3 tests was missed out by CI. I figure it easier to run every test with s3 feature on, rather than selective, it's far more reliable (also as of now we have missed the whole file from ci )


houqp (approved):

houqp:
looks like quay.io is down :facepalm: we can probably just ignore the python test failure for now.

mosyp:
@houqp works for me :)

PR #451, opened 2021-10-08
Generate new session name on assume role credentials provider refresh

mosyp:
The default autorefresh rusoto provider uses cached session name between assume role requests. This could cause trouble if there's more than one active sessions. In order to prevent conflicts, we should generate the new session name for each assume role request


mosyp:
@xianwill Another possible fix could be to lower the rusoto/hyper idle timeout as in https://github.com/hyperium/hyper/issues/2136#issuecomment-589488526.

At first I thought that creating "new" client with fresh session name could help, but now it seems that simply `StsClient::new(self.region.clone())` is a reuse for `Client::shared()` so the client stays the same, and if the issue is really in idle connection (which could be the case, since credentials are executed once in 30min), then we might try this approach

PR #452, opened 2021-10-09
return lazy iterator in get tombstone methods

houqp:
# Description

It's better to not force clone and iterator materialization in the API. Instead, we can return an iterator of reference and let caller decide whether they need to clone and when to materialize into a collection.


houqp:
@xianwill the the logic below, we have:

```rust
 let use_extended_remove_schema = tombstones
        .iter()
        .all(|r| r.extended_file_metadata == Some(true) && r.size.is_some());
```

Do we really check for all the tombstones? I thought it should be enough to set `use_extended_remove_schema` to false if one of the tombstone doesn't match the filer?

mosyp:
@houqp I feel like that's exactly what `all` is doing, e.g. from the rustdoc

```rust
    /// `all()` is short-circuiting; in other words, it will stop processing
    /// as soon as it finds a `false`, given that no matter what else happens,
    /// the result will also be `false`.
```

Issue #435, opened 2021-09-18
Reduce memory footprint


dispanser:
# Description

I've been playing with a relatively large delta table recently, in the context of #425 (ignore tombstones for readers).

While the 5.3 million file paths themselves occupy about 600M, the overall memory consumption of a small rust program after loading the delta table adds up to 49.4G.

Here's the memory consumption for the various fields in `action::Add`, on that table with 5.3 million files, and 19 columns (each column has file statistics enabled).

| field                   | mem ps | 
| ----------------------- | ------ |
| stats parsed            | 37.1G  |
| stats                   | 6.4G   |
| partition values        | 2.1G   |
| parsed partition values | 1.9G   |
| path                    | 600M   |

I think we should be able to get that down considerably. The biggest offender is stats_parsed, which is a  Option<parquet::record::Row>. If I understand the concept correctly, it has minValues, maxValues, nullCounts and numRecords as top-level entries. Each of the first three then has a list / vec of tuples of (column name, value), so it effectively repeats each of my column names three times, for every individual file.

mgill25:
I did some memory inspections. Just in case anyone else is working from a Mac (@dispanser uses Arch):

I tried to reproduce this and was very confused because my `htop` did not show high memory usage. I was expecting the program with the same large dataset to blow up/or at least be OOM-killed. But turns out both Memory Compression and Swap played a role in displaying an incomplete picture (as far as `htop`) was concerned.

`_delta_log` size: ~15G

`htop` after Table Load:

![Screenshot 2021-09-18 at 14 06 45](https://user-images.githubusercontent.com/1311924/133888660-4cb40be6-e298-4b51-936b-f71ecacc4378.png)

`Activity Monitor` shows the rest of the picture:
![Screenshot 2021-09-18 at 14 06 29](https://user-images.githubusercontent.com/1311924/133888709-d27685a7-d5d1-47f2-8542-7cbe4e281e45.png)

Here we indeed see ~44G memory utilization (post table load). We also observe heavy Swap and Memory Compression by the OS:

![Screenshot 2021-09-18 at 14 07 36](https://user-images.githubusercontent.com/1311924/133888737-f54d5d4c-4acf-4c56-bac2-d0ca23e3687e.png)

The numbers come down when the program is killed:
![Screenshot 2021-09-18 at 14 08 51](https://user-images.githubusercontent.com/1311924/133888831-7eceeedb-1d2f-41d3-bbec-34990ea6b2ea.png)

[...]

PR #454, opened 2021-10-09
WIP: Record Batch support for delta log stats_parsed (#435)

mgill25:
- Initial attempt at creating record batches out of stats_parsed from the delta log for `Add` actions.
- A new representation for the stats via `StatRecordBatch`
- Currently only supports creation of batches for columns that have integer types. Hard-coded `i64`s for now. Of course, this needs to be extended to provide support for more types.
- Added an associated test case.

The basic idea is to simply provide the Arrow schema for the columns for which we want to create the batches. I've left the selection of columns as an argument - I can see cases where it might not make sense to keep statistics for _all columns_.

I only parse `minValues`, `maxValues` and `numRecords` for now. `nullCounts` need to be added.

Additionally, `StatRecordBatch` is _not_ yet embedded into the primary `Add` struct yet. I ran into some issues there and plan to figure that out in the coming days/weeks.

First PR. I'm sure there's plenty of things that need to be improved here. Feedback is very much appreciated! :)


houqp:
Thanks @mgill25 for taking on this initiative! I think to fully leverage the columnar design, we can't store the stats in the action struct itself. We will need to store them in a level of abstraction that's higher than Action. I am going to take another look at the design to write up some of my thoughts as a follow up comment.

mosyp:
@mgill25 Hi, great PR. I wonder whether you're planning on finish that. Also alternatively @houqp, we could merge it to the new feature branch instead and work on it collectively. 

The mem issue has become the concern for us so we thinking either we can benefit from this feature or not storing struct actions in memory at all (for kafka-delta-ingest we don't need them)

manishgill-tomtom:
Re: the design ideas, @houqp I'm not well versed in Arrow enough to make the best design decision, so perhaps `Vec<Option<T>>` is the short term solution we take for now? I will try to make an update to the PR this week.

dispanser:
> The mem issue has become the concern for us so we thinking either we can benefit from this feature or not storing struct actions in memory at all (for kafka-delta-ingest we don't need them)

If you don't need the stats for the use case, not storing them at all is may be the best (and cheapest option). An alternative "workaround" for memory consumption problems that I can see would be to have configurable whilelist of columns for which the stats are useful, and discard the rest. For our use case, most of these column stats are useless (e.g., UUIDs min / max stats tend to not have any filtering capability). 

There's a possibility in delta itself to control this to some extend already on the write, but it involves reordering columns (and assumes that the table consumer can convince the table producer to perform that change).

houqp:
Sorry for the late response, last 2 weeks have been crazy for me, so I am still catching up on all the great discussions :)

> Re: the design ideas, @houqp I'm not well versed in Arrow enough to make the best design decision, so perhaps Vec<Option<T>> is the short term solution we take for now? I will try to make an update to the PR this week.

This sounds good to me :+1: 

> If you don't need the stats for the use case, not storing them at all is may be the best (and cheapest option). An alternative "workaround" for memory consumption problems that I can see would be to have configurable whilelist of columns for which the stats are useful, and discard the rest. For our use case, most of these column stats are useless (e.g., UUIDs min / max stats tend to not have any filtering capability).

I think supporting projection based stats loading could be an interesting and useful feature to add :+1: 

These optimization probably wont' help with kafka-delta-ingest though since it doesn't know if downstream consumer of the table will need these stats or not.

mgill25:
Apologies for zero activity on this for the last 2 months. 

I think for now I don't have the bandwidth to work on this, so if anyone wants to pick this up, please feel free to.

roeap:
superseded by #2037

PR #456, opened 2021-10-11
Do action reconciliation

### LINKED ISSUE #455: Add back removed file doesn't update tombstone
# Environment

**Delta-rs version**: all

**Binding**: all

***
# Bug

**What happened**:

Consider the following commits in order:

1. add file1
2. remove file1
3. add file1

Based on https://github.com/delta-io/delta/blob/master/PROTOCOL.md#action-reconciliation, the 2nd add action should clear file1 from tombstone. But this is not the case in delta-rs today. file1 will be stored in both file list and tombstone.

tombstone should not contain file1

viirya:
# Description

Based on https://github.com/delta-io/delta/blob/master/PROTOCOL.md#action-reconciliation, we should do action reconciliation between `add` and `remove` action collections.

# Related Issue(s)

closes #455 

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -443,12 +443,13 @@ impl DeltaTableState {
             new_state.tombstones.into_iter().for_each(|r| {
                 self.tombstones.insert(r);
             });
-        }
 
-        if !new_state.files.is_empty() {
-            let new_adds: HashSet<&str> = new_state.files.iter().map(|s| s.path.as_str()).collect();
-            self.tombstones
-                .retain(|a| !new_adds.contains(a.path.as_str()));
+            if !new_state.files.is_empty() {
+                let new_adds: HashSet<&str> =
+                    new_state.files.iter().map(|s| s.path.as_str()).collect();
+                self.tombstones
+                    .retain(|a| !new_adds.contains(a.path.as_str()));
+            }
         }
 
         self.files.append(&mut new_state.files);
--- a/rust/tests/checkpoint_writer_test.rs
+++ b/rust/tests/checkpoint_writer_test.rs
@@ -104,15 +104,15 @@ mod checkpoints_with_tombstones {
 
     #[tokio::test]
     async fn test_expired_tombstones() {
-        let mut table = create_table("./tests/data/checkpoints_tombstones/expired", Some(hashmap! {
+        let mut table = fs_common::create_table("./tests/data/checkpoints_tombstones/expired", Some(hashmap! {
             delta_config::TOMBSTONE_RETENTION.key.clone() => Some("interval 1 minute".to_string())
         })).await;
 
-        let a1 = add(3 * 60 * 1000); // 3 mins ago,
-        let a2 = add(2 * 60 * 1000); // 2 mins ago,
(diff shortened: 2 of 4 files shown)


dispanser:
I'm having a closer look now. @viirya : nice find. add-remove-add cycles indeed lead to a violation of the delta protocol.

Looking at the unit test `test_checkpoint_with_action_reconciliation` and the resulting final checkpoint (0003) however, it contains:

```
+--------------------+--------+----+--------------------+--------------------+
|            metaData|protocol| txn|                 add|              remove|
+--------------------+--------+----+--------------------+--------------------+
|                null|  {1, 2}|null|                null|                null|
|{7941a1c6-89fa-49...|    null|null|                null|                null|
|                null|    null|null|                null|{a50201d9-0b29-4c...|
|                null|    null|null|{e897fedf-978d-40...|                null|
|                null|    null|null|{a50201d9-0b29-4c...|                null|
+--------------------+--------+----+--------------------+--------------------+
```

The delta protocol talks about the content of the checkpoint, whereas our test only captures the state after reading. The proposed patch fixes the "view" of the data (i.e., the final `DeltaTableState`), but the checkpoint itself still violates the protocol.

The reason for this is a bit subtle. There are three code paths that eventually hit `process_acion`:

1. `DeltaTable::create`, 
	- fixed set of actions
	- add / remove not included
	- can be ignored.
2. `DeltaTable::restore_checkpoint`: 
    - here, we start with a clean state, loading a checkpoint. 
    - Reconcilation for adds or removes is not necessary, because the checkpoint should be consistent according to the protocol
	- this is safe
3. `DeltaTable::apply_log_from_bufread`:
    - called for every version incrementally, from `apply_log`
	- **uses a fresh, clean** table state
	- because of that, `process_action` does not see the full table state, so reconcilation is **not possible**
	- full table state is only considered in the `merge` operation which happens for every commit independently

The proposed patch fixes the problem at 2. , but that's not solving the underlying problem. It works in so far that it loads an invalid checkpoint and makes it valid, but the bad checkpoint is already in the wild. I made a quick test with spark, and delta 1.0.0 happily accepts the meta data, so it's not that bad but of course since there's no real, backing parquet files we can't see if its interpretation is correct.

Another way to see that it's not fixing all cases: if we remove the checkpoint writing, and force delta-rs to load from json commits instead, the test starts to fail: it has the same file in adds and removes.

I would like to think about the reconcilation rules as a set of invariants for the state of the meta data, and those should be valid not on checkpoint writing, but always. A checkpoint write then produces correct data in all cases.

In my opinion the only place where all information is available is the `merge` operation, as it has access to both the full table state and the newly created actions from the latest commit.

viirya:
Yes, when I'm moving the logic to `merge`, I noticed that the fix put in `process_action` for `restore_checkpoint` only fixed the view of checkpoint. We should do it in `merge` so to keep valid state in every commit and writing checkpoint.

viirya:
Does current approach look good? @houqp

dispanser:
> Does current approach look good? @houqp

To me it indeed looks good. I made a hand-wavvy performance test with one particular big delta table that was used for #444, and I don't see any negative impact on performance. Great job!

PR #462, opened 2021-10-15
Support hash lookup by path string for Remove action

### LINKED ISSUE #460: Support hash lookup by path string for Remove action
# Description

Follow up on https://github.com/delta-io/delta-rs/pull/456/files#r727686841.

**Use Case**

We could avoid building a hashset of strs in `DeltaTableState::merge` method if we could check for hash collisions in `HashSet<Remove>` using `Add.path`.

Turns out Rust's builtin HashSet is already designed with this use-case in mind, we just have to implement an extra Borrow trait for `Remove`:

```rust
use std::hash::Hash;
use std::hash::Hasher;
use std::collections::HashSet;
use std::borrow::Borrow;

#[derive(Eq, PartialEq)]
struct Remove {
    pub path: String,
}

impl Hash for Remove {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.path.hash(state);
    }
}

impl Borrow<str> for Remove {
    fn borrow(&self) -> &str {
        self.path.as_ref()
    }
}

fn main() {
    let r1 = Remove{ path: "foo".to_string() };
    let mut set = HashSet::new();
    set.insert(r1);
    
    assert!(set.contains("foo"));
    assert!(!set.contains("bar"));
}
```

Rust play gournd link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cc1ddea6e9f03ebb8c38bbb3cb774985

Stack overflow discussion: https://stackoverflow.com/questions/45384928/is-there-any-way-to-look-up-in-hashset-by-only-the-value-the-type-is-hashed-on

viirya:
# Description

By implementing `Borrow` trait for `Remove`, we can support lookup by path string for `Remove` action.

# Related Issue(s)

closes #460

# Documentation

### DIFF
--- a/rust/src/table_state.rs
+++ b/rust/src/table_state.rs
@@ -136,14 +136,8 @@ impl DeltaTableState {
     /// merges new state information into our state
     pub fn merge(&mut self, mut new_state: DeltaTableState, require_tombstones: bool) {
         if !new_state.tombstones.is_empty() {
-            let new_removals: HashSet<&str> = new_state
-                .tombstones
-                .iter()
-                .map(|s| s.path.as_str())
-                .collect();
-
             self.files
-                .retain(|a| !new_removals.contains(a.path.as_str()));
+                .retain(|a| !new_state.tombstones.contains(a.path.as_str()));
         }
 
         if require_tombstones {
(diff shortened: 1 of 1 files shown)


houqp:
Sorry @viirya I thought I merged this PR, turns out I didn't :( Could you resolve the conflict so I can merge it?

viirya:
@houqp No problem! Resolved it.

houqp:
@viirya looks like the map optimization in merge function got lost during the merge.

houqp:
:tada: for more performance boost.

viirya:
Thanks @houqp @dispanser @Dandandan !

PR #463, opened 2021-10-16
materialize tables in pyhton via native storage backend

### LINKED ISSUE #324: [python] Cannot load dataset from ADLS
# Environment

**Delta-rs version**: 0.4.0

**Binding**:  python

**Environment**:
- **Cloud provider**: Azure
- **OS**: OSX
- **Other**:

***
# Bug

**What happened**:

When loading delta table from Azure storage, an error is thrown when the dataset is loaded through the pyarrow bindings. It seems all functions using the Azure integration directly from delta-rs work (e.g. `dt.version()`) however when materializing the dataset (`dt.to_pyarrow_dataset()`) and the arrow bindings are involved, an error is thrown. 

The dataset to load.

**How to reproduce it**:

1. Configure then environment for ADLS
2. Try to load and materialize DeltaTable

**More details**:

Locally, i got this to work by patching the `to_pyarrow_dataset` with some slight updates and passing an [fsspec filesystem](https://arrow.apache.org/docs/python/filesystems.html#using-fsspec-compatible-filesystems) to arrow. If an additional (extra) dependency is acceptable, I could give submitting this workaround a shot. Then again the cleaner solution would likely be to add this natively to arrow. Or is there a way the use the native reader used within the delta-rs code?

Thanks for this great project!

roeap:
# Description

This PR introduces a simple implementation of a `FilesystemHandler` for pyarrow `PyFileSystem` backed by the native StorageBackends from the delta-rs crate.

# Related Issue(s)

- closes #324

# Open questions

- the initial intent was to only use the native backend in case of azure storage, however this introduced some complexities in differentiating cases on the python side when creating the handler instance. 
- in case we always want to use the native backend, is the `filesystem` argument still needed, since we should always be able to read with the native backed filesystem.
- Some tests were locally executed against adls gen2, are there azure credentials available in ci/cd?
- The S3 tests did not succeed locally, it seems like the credentials are expired, but bot sure, since I haven't worked with S3 before.
- The FileSystemHandler implementation is not useful beyond usage inside the to_dataset function. Specifically the file_info methods don't work for directories
- conversion to RecordBatches could be done entirely on the rust side

### DIFF
--- a/python/docs/source/conf.py
+++ b/python/docs/source/conf.py
@@ -51,8 +51,11 @@ nitpick_ignore = [
     ("py:class", "pyarrow.lib.Table"),
     ("py:class", "pyarrow.lib.DataType"),
     ("py:class", "pyarrow.lib.Field"),
+    ("py:class", "pyarrow.lib.NativeFile"),
     ("py:class", "pyarrow._fs.FileSystem"),
     ("py:class", "pyarrow._fs.FileInfo"),
+    ("py:class", "pyarrow._fs.FileSelector"),
+    ("py:class", "pyarrow._fs.FileSystemHandler"),
     ("py:class", "RawDeltaTable"),
     ("py:class", "pandas.DataFrame"),
 ]


roeap:
@houqp, I created a first draft implementing a simple filesystem backed by the native readers. Doing so some questions came up. Would be great to get some feedback on the current state.

houqp:
Thanks @roeap , this refactor is really cool!

> in case we always want to use the native backend, is the filesystem argument still needed, since we should always be able to read with the native backed filesystem.

Yes, I think we should still keep them so users will still have the flexibility to swap out the underlying implementation with something that better fits their use-cases.

The main downside with our current native object store implementation is we are only performing single threaded s3 get. Ideally, we should be splitting the object into smaller chunks and download them concurrently. I don't know if the current arrow builtin s3 implementation is already doing this, if so, we will be seeing some performance regressions. But I think this is something we should fix inside delta-rs.

> Some tests were locally executed against adls gen2, are there azure credentials available in ci/cd?

Unfortunately, no. Maybe one of these days we can get azure to sponsor us :D

> The S3 tests did not succeed locally, it seems like the credentials are expired, but bot sure, since I haven't worked with S3 before.

Yes, we have since moved on to use local-stack for S3 integration test, see https://github.com/delta-io/delta-rs/tree/main/rust#development.

> The FileSystemHandler implementation is not useful beyond usage inside the to_dataset function. Specifically the file_info methods don't work for directories

Do we need it to be used outside of the `to_dataset` function? There are ways to fake a directory in object stores if needed.

> conversion to RecordBatches could be done entirely on the rust side

Yes, we could work on this as a follow up task if you are interested in this optimization.

PR #464, opened 2021-10-17
Add new module for DeltaTableState

### LINKED ISSUE #461: Safe guard DeltaTableState using Rust's type system
# Description

**Use Case**

Follow up for https://github.com/delta-io/delta-rs/pull/456#issuecomment-941926489.

When we call `process_action` on a particular DeltaTableState, we need to make sure the passed in actions do not cross commit boundaries. We can restructure our code to have the Rust compiler to guarantee this property for us at build time.

Here is what I have in mind:

* Move `process_action` and `DeltaTableState` into its own module
* Add two new methods: `DeltaTableState::from_commit` and `DeltaTableState::from_checkpoint`

Since `DeltaTableState` is moved into its own module, other part of the code base won't be able to mutate a table state with arbitrary actions. The `from_commit` and `from_checkpoint` methods make sure users won't be able to construct table states from actions across commit boundaries.

**Related Issue(s)**

viirya:
# Description

This change tries to add a new module for `DeltaTableState`. `process_action` which updates state from given action is hidden in this module.

Other codes can only construct a `DeltaTableState` from a commit by `from_commit`, or a checkpoint by `from_checkpoint`. 

This is to make sure users cannot construct table states from actions across commit boundaries that could cause unexpected and incorrect states.

# Related Issue(s)

closes #461 

# Documentation


houqp:
FYI @dispanser

houqp:
Thanks @viirya !

viirya:
Thanks @houqp @dispanser !

Issue #310, opened 2021-07-10
Consider adding Polars as possible data processing framework.


ritchie46:
I see you already support `DataFusion` and `Rust DataFrame` as well, which both use `apache arrow`. It would be nice if we could read/write from/to [Polars](https://github.com/pola-rs/polars) as well.

If you'd be willing to, I could initialize a PR that sets that up.

houqp:
Yes, please do, we will be more than happy to merge the PR. Make sure you update the readme to mention Polars as well :)

ritchie46:
Sorry for the delay, I will work on is. Is it ok to also add an implementation based on arrow2 (under a feature flag)?

houqp:
@ritchie46 yes, we want to move to arrow2 in the long run as well :) but it would be good if we can manage it with a feature flag because it's going to take some amount of time before we can migrate kafka-delta-ingest to arrow2.

houqp:
If arrow2 feature flag is not possible, we can also maintain an official arrow2 branch

ritchie46:
> @ritchie46 yes, we want to move to arrow2 in the long run as well :) 

Great! Polars already runs great on arrow2, and this way, I hope to improve third party support as well. :)

> If arrow2 feature flag is not possible, we can also maintain an official arrow2 branch

Yes, I don't know how deep embedded arrow is in delta-rs?

houqp:
> Yes, I don't know how deep embedded arrow is in delta-rs?

It's not deep, I think the migration should be pretty straight forward. The hard part is isolating it into a feature flag so we can support both arrow and arrow2 in the short term.

wseaton:
Would love to see this (specifically an arrow2 feature flag), could consolidate a lot of my custom code around parquet partitioning and writing to object storage + gain the benefits of delta lake on top of that.

ritchie46:
I got bounced on the arrow2 migration. It was not so trivial to feature gate indeed. So that would probably need a separate branch?

houqp:
@ritchie46 I actually got something mostly working 2 weeks ago, but got distracted with datafusion and roapi so wasn't able to complete the PoC. Let me try to get something pushed out this weekend so we can take a look at it together to see if that's the right path forward.

mohitreddy1996:
@ritchie46 and @houqp thanks a lot for pushing the PoC. I was wondering if this is feature complete and any example I could refer to get started?

roeap:
[...]

PR #465, opened 2021-10-17
parquet2 implementation backed by parquet2 feature gate

houqp:
decouple core from arrow

# Description

WIP parquet2 implementation. The goal of this PR is to implement full read support leveraging parquet2. Write support is out of the scope and should be added as follow up PRs. Arrow2 integration is also out of the scope and should be added through follow up PR.

Currently all read tests are passing:

```
cargo test --no-default-features --features=arrow2,parquet2 
```

A quick benchmark shows more than 50% performance boost on checkpoint deserialization. Only tested with a very tiny checkpoint from the golden dataset, I would expect the performance gap would be bigger for larger real world tables.

Todo:

- [x] clean up duplicated code
- [x] support parsing map type
- [x] support parsing list type
- [x] benchmark

# Related Issue(s)

blocks #310 

# Documentation


houqp:
Sounds good @ritchie46 , I will complete the parquet parsing support for map and list this weekend. But the branch I have here right now should be enough to unblock ploars integration.

ahmedriza:
There are currently some issues in `parquet2` with regard to nested types.  See https://github.com/jorgecarleitao/arrow2/issues/937 for example.

andrei-ionescu:
@houqp, any updates on this?

houqp:
@andrei-ionescu I have implemented all the data types other than map and nested list, so it's very close to be complete. However, my time is limited now, so progress will be slow. Anyone if welcome to collaborate on this branch to push this over the finish line :)

houqp:
alright, this branch is now feature complete ;) now it's time to catch up to latest delta-rs and arrow2/parquet2 releases.

rust/tests/dataframe_test.rs
```diff
@@ -1,89 +1,83 @@
-extern crate arrow;
-extern crate deltalake;
 #[cfg(feature = "rust-dataframe-ext")]
-extern crate rust_dataframe;
+mod rust_dataframe {
```
houqp, line 2: no change in test, just moved code into its own mod to make conditional compilation annotation cleaner.

houqp:
@wjones127 @roeap ready for review.

rust/src/action/parquet2_read/dictionary/binary.rs
```diff
@@ -0,0 +1,49 @@
+use parquet2::encoding::get_length;
+use parquet2::error::Error;
+
+// FIXME: replace vec with reference
```
roeap, line 4: should this be fixed in this PR?

rust/src/action/parquet2_read/map.rs
```diff
@@ -0,0 +1,113 @@
+// use std::collections::HashMap;
```
roeap, line 1: remove?

rust/src/action/parquet2_read/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+        let schema_path = &column_desc.path_in_schema;
+
+        let deserialize_column_page = match schema_path[0].as_ref() {
+            "txn" => deserialize_txn_column_page,
+            "add" => deserialize_add_column_page,
+            "remove" => deserialize_remove_column_page,
+            "metaData" => deserialize_metadata_column_page,
+            "protocol" => deserialize_protocol_column_page,
+            "commitInfo" => deserialize_commit_info_column_page,
+            "cdc" => deserialize_cdc_column_page,
```
roeap, line 665: ❤️

rust/tests/dataframe_test.rs
roeap: don't think it matters much but `#![cfg(feature = "rust-dataframe-ext")]` might also work?

As a side note I was wondering if this still needs to / should be supported? It seems the repository for rust_dataframe has been archived, or did just move somewhere else?

roeap (commented):
Only had some very minor comments. While not an expert on parquet reading, I felt the tests should cover that quite nicely!

The one thing I was wondering is the `impl`'s for the actions, or the action.rs file in general. It seems we have alot of feature flags there now, and maybe it would be cleaner to have the parquet specific stuff also moved into its own mod. Maybe adopt something ile the trait based approach in the new parquet2 implementation?  

As this also seems like a step towards supporting `polars`, we will face the question of having good integration points to support various backends / engines or however one wnats to call it :).

great work!

rust/src/action/parquet2_read/dictionary/binary.rs
houqp: good catch, totally forgot about this one.

houqp:
> maybe it would be cleaner to have the parquet specific stuff also moved into its own mod

Good idea, I have moved all those code into a `parquet_read` mod to keep action lean.

I chatted this with Andy and Jorge at the Data+AI summit a couple of months ago, the long term goal is to develop an Arrow trait that allows users to switch between arrow-rs and arrow2 in different projects including datafusion and delta-rs. This will also open up the possibility for a 3rd GPU based arrow implementation ;)

roeap (approved):
LGTM!

PR #484, opened 2021-10-31
Clean up expired delta table commit logs after checkpoint

### LINKED ISSUE #437: Checkpoint should also merge old versions
# Description

**Use Case**

In the reference implementation, `delta.logRetentionDuration` is to control how old delta table commit logs are being purged during checkpoint. We should do the same thing for our checkpoint implementation as well otherwise the log folder will grow indefinitely.

**Related Issue(s)**

viirya:
# Description

This patch adds the config `logRetentionDuration` which controls how long delta table commit logs should be in retention. After creating a checkpoint, expired delta table commit logs will be deleted.

# Related Issue(s)

closes #437 

# Documentation

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -27,6 +27,7 @@ chrono = "0"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 lazy_static = "1"
 percent-encoding = "2"
+time = "0.1"
 
 # HTTP Client
 reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "stream"], optional = true}
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -27,7 +27,6 @@ chrono = "0"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 lazy_static = "1"
 percent-encoding = "2"
-time = "0.1"
 
 # HTTP Client
 reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "stream"], optional = true}
--- a/rust/src/checkpoints.rs
+++ b/rust/src/checkpoints.rs
@@ -203,11 +203,11 @@ async fn flush_delete_files<T: Fn(&(DeltaDataTypeVersion, ObjectMeta)) -> bool>(
     }
 
     let deleted = files_to_delete
-        .into_iter()
+        .iter_mut()
         .map(|file| async move {
             match storage.delete_obj(&file.1.path).await {
                 Ok(_) => Ok(1),
-                Err(e) => return Err(DeltaTableError::from(e)),
+                Err(e) => Err(DeltaTableError::from(e)),
             }
         })
         .collect::<Vec<_>>();
@@ -242,7 +242,7 @@ async fn cleanup_expired_logs(
     let mut deleted_log_num = 0;
 
     // Get file objects from table.
-    let log_uri = storage.join_path(&table_uri, "_delta_log");
+    let log_uri = storage.join_path(table_uri, "_delta_log");
     let mut candidates: Vec<(DeltaDataTypeVersion, ObjectMeta)> = Vec::new();
     let mut stream = storage.list_objs(&log_uri).await?;
     while let Some(obj_meta) = stream.next().await {
@@ -294,7 +294,7 @@ async fn cleanup_expired_logs(
 
     let mut current_file: (DeltaDataTypeVersion, ObjectMeta);
     loop {
-        if candidates.len() == 0 {
+        if candidates.is_empty() {
             deleted_log_num += flush_delete_files(
                 storage,
                 &mut maybe_delete_files,
--- a/rust/src/checkpoints.rs
+++ b/rust/src/checkpoints.rs
@@ -97,11 +97,9 @@ pub async fn create_checkpoint_from_table_uri(
     )
     .await?;
 
-    let new_version = table.version + 1;
-
-    if new_version > 0 {
+    if table.version >= 0 {
         let deleted_log_num = cleanup_expired_logs(
-            new_version,
+            table.version + 1,
             table.storage.as_ref(),
             table.get_state(),
             table_uri,
@@ -123,11 +121,9 @@ pub async fn create_checkpoint_from_table(table: &DeltaTable) -> Result<(), Chec
     )
     .await?;
 
-    let new_version = table.version + 1;
-
-    if new_version > 0 {
+    if table.version >= 0 {
         let deleted_log_num = cleanup_expired_logs(
-            new_version,
+            table.version + 1,
             table.storage.as_ref(),
             table.get_state(),
             &table.table_uri,
--- a/rust/src/checkpoints.rs
+++ b/rust/src/checkpoints.rs
@@ -285,11 +285,8 @@ async fn cleanup_expired_logs(
     // Init
     if !candidates.is_empty() {
         let removed = candidates.remove(0);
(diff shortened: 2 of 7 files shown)


viirya:
Seems due to the behavior of `std::fs::rename`on Windows, `rename_noreplace` is a bit flaky.

houqp:
sorry, running out of my time for PR review today, will need to push it to next week :(

viirya:
No problem. Thanks @houqp !

rust/tests/checkpoint_writer_test.rs
```diff
@@ -99,6 +99,57 @@ mod simple_checkpoint {
...
+            Err(e) => assert!(false, "Cannot load version 1"),
+        }
+
+        checkpoints::create_checkpoint_from_table(&table)
+            .await
+            .unwrap();
+        table.update().await.unwrap(); // make table to read the checkpoint
+        assert_eq!(table.get_files(), vec![a1.path.as_str(), a2.path.as_str()]);
+
+        match table.load_version(0).await {
```
houqp, line 136: same here, you can use `expect_err` to simplify the code.

rust/src/checkpoints.rs
```diff
@@ -140,6 +192,153 @@ async fn create_checkpoint(
...
+        })
+        .collect::<Vec<_>>();
+
+    let mut deleted_num = 0;
+    for x in deleted {
+        match x.await {
+            Ok(_) => deleted_num += 1,
+            Err(e) => return Err(e),
+        }
+    }
```
houqp, line 221: perhaps we can avoid the collection here and update the deleted_num count within the first files_to_delete loop?

houqp (commented):
@xianwill @mosyp could you help review this as well? I think this will have an impact to our production service as well.

mosyp:
@viirya great work! but give me like a couple of days to fully review it @houqp, as I'm in some deep kafka-delta-ingest related work right now

viirya:
Thanks @mosyp !

rust/src/checkpoints.rs
viirya: Hmm, tried but got:


```rust
error[E0308]: mismatched types
   --> rust/src/checkpoints.rs:206:48
    |
206 |       files_to_delete.iter_mut().for_each(|file| async move {
    |  ________________________________________________^
207 | |         match storage.delete_obj(&file.1.path).await {
208 | |             Ok(_) => {
209 | |                 deleted_num += 1;
...   |
213 | |         }
214 | |     });
    | |_____^ expected `()`, found opaque type
    |
   ::: /Users/viirya/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/future/mod.rs:61:43
    |
61  |   pub const fn from_generator<T>(gen: T) -> impl Future<Output = T::Return>
    |                                             ------------------------------- the found opaque type
    |
    = note: expected unit type `()`
             found opaque type `impl futures::Future`
```

when I change `map` to `for_each`.

Looks like I cannot return the error from `for_each` for the `Future` from `delete_obj`.

rust/Cargo.toml
```diff
@@ -27,6 +27,7 @@ chrono = "0"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 lazy_static = "1"
 percent-encoding = "2"
+time = "0.1"
```
mosyp, line 30: why do we need that?

rust/src/checkpoints.rs
```diff
@@ -100,6 +122,36 @@ pub async fn create_checkpoint_from_table(table: &DeltaTable) -> Result<(), Chec
         &table.table_uri,
     )
     .await?;
+
+    let new_version = table.version + 1;
+
+    if new_version > 0 {
+        let deleted_log_num = cleanup_expired_logs(
+            new_version,
```
mosyp, line 130: do we need that variable? at first look it confuses, e.g. since you're not creating `new` version

rust/src/checkpoints.rs
```diff
@@ -100,6 +122,36 @@ pub async fn create_checkpoint_from_table(table: &DeltaTable) -> Result<(), Chec
...
+            table.storage.as_ref(),
+            table.get_state(),
+            &table.table_uri,
+        )
+        .await?;
+        debug!("Deleted {:?} log files.", deleted_log_num);
```
mosyp, line 136: Also, instead of always running cleanup, could we propagate the `enableExpiredLogCleanup` delta table config as in https://github.com/delta-io/delta/blob/08b17e0fae5222179a52ee277f93d8ebc0029621/core/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L306

rust/src/checkpoints.rs
```diff
@@ -140,6 +192,150 @@ async fn create_checkpoint(
...
+        ObjectMeta {
+            path: String::new(),
+            modified: MIN_DATETIME,
+        },
+    );
+    let file_needs_time_adjustment =
+        |current_file: &(i64, ObjectMeta), last_file: &(i64, ObjectMeta)| {
+            last_file.0 < current_file.0
+                && last_file.1.modified.timestamp() >= current_file.1.modified.timestamp()
+        };
```
mosyp, line 276: could you explain this one?

rust/src/checkpoints.rs
viirya: Yea, I think so. Would you like to add here or we can do it in another PR after this?

rust/src/checkpoints.rs
mosyp: ty, makes sense

rust/src/checkpoints.rs
mosyp: sure, we can do that in a separate PR

rust/src/checkpoints.rs
viirya: yea, the variable can be removed.

rust/Cargo.toml
viirya: This is for the time duration at `checkpoint.rs` L308.

```rust
let updated = (
  current_file.0,
  ObjectMeta {
    path: current_file.1.path.clone(),
    modified: last_file.1.modified.add(Duration::seconds(1)),
  },
);
```

mosyp: yeah, but I believe we can use `core::time::Duration`  or even `chrono`'s one

mosyp (commented):
LGTM overall, ty. But I'd revert the new create adding if possible, it seems like the std duration should be enough. 

Also small nitpick, I'd work on the `create_checkpoint_from_table_without_cleaning_logs` and maybe introduce the function which accepts `bool` as to run cleanup or not, and the reuse this function with a new one which takes that bool param from the delta table confs. But we can work on that later.

rust/Cargo.toml
viirya: Ah, let me try it.

viirya: `chrono::Duration` works but `core::time::Duration` doesn't. I replaced it with `chrono::Duration`. Thanks.

viirya:
Thanks for review @mosyp!

I've reverted the new crate and use `chrono::Duration` now. Please let me know if this is good to go.

I will work on the extra config controlling cleanup or not in next work. Thanks.

mosyp (approved):

mosyp:
@xianwill wanna take a look as sell?

viirya:
Thanks @houqp @mosyp ! I will work on the follow up bool config soon.

PR #485, opened 2021-11-01
Try to fix flaky rename under Windows

viirya:
# Description

Seems async rename under Windows is flaky due to no native rename if not exists support.
Use a shared lock to prevent threads renaming at the same time.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/storage/file/rename.rs
+++ b/rust/src/storage/file/rename.rs
@@ -3,6 +3,15 @@ use crate::StorageError;
 #[cfg(windows)]
 mod imp {
     use super::*;
+    use lazy_static::lazy_static;
+    use std::sync::Arc;
+    use std::sync::Mutex;
+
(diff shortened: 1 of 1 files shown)


houqp:
Thanks @viirya , if this doesn't work we can also just replace the windows rename code with `unimplemented` since the current rename is not atomic anyway, so shouldn't be used in any serious context.

viirya:
It passed the windows test. Let me trigger it more times to verify.

viirya:
It also passed windows test in 2nd and third time. Looks okay? The failure of building Sphinx document for python_build looks unrelated.

houqp (approved):

houqp:
Thanks @viirya ! yeah, the python build is something different.

viirya:
Thank you @houqp !

Issue #59, opened 2021-01-09
WebPKIError when running azure test code


rtyler:
As of my current tree (`40443d6390674092b8763cbeba1b1b03144f446c`) it seems that the azure tests cannot properly work locally, but they continue to work fine (seemingly) in CI. I'm wondering if there's something magic about the tests passing in CI since they're technically running within Azure via GitHub Actions already.

## Error
```
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running /home/tyler/source/github/delta-io/delta.rs/target/debug/deps/azure_test-a5a0db4320167091

running 1 test
test azure::test_azure_simple ... FAILED

failures:

---- azure::test_azure_simple stdout ----
thread 'azure::test_azure_simple' panicked at 'called `Result::unwrap()` on an `Err` value: LoadCheckpoint { source: Storage { source: Azure { source: HyperError(hyper::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: WebPKIError(UnknownIssuer) } })) } } }', rust/tests/azure_test.rs:18:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:483
   1: core::panicking::panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/panicking.rs:85
   2: core::option::expect_none_failed
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/option.rs:1234
   3: core::result::Result<T,E>::unwrap
             at /home/tyler/.rustup/toolchains/stable-aarch64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:973
   4: azure_test::azure::test_azure_simple::{{closure}}
             at ./tests/azure_test.rs:16
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /home/tyler/.rustup/toolchains/stable-aarch64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/mod.rs:80
   6: tokio::runtime::basic_scheduler::BasicScheduler<P>::block_on::{{closure}}::{{closure}}
[...]

PR #486, opened 2021-11-01
Add authorization options for azure storage backend

### LINKED ISSUE #149: Add support for more authentication schemas for Azure
Currently the only way to authenticate against Azure is by providing accountname and/or key/sas token. For ADLS Gen2 with ACL enabled data lakes, we need also to authenticate by supplying service principals (Azure AD).
A good way of working would be to set env variables as follow:
* AZURE_STORAGE_AUTH_TYPE
* AZURE_STORAGE_SPN_CLIENT_ID
* AZURE_STORAGE_SPN_TENANT_ID
* AZURE_STORAGE_SPN_CLIENT_SECRET

### LINKED ISSUE #255: Fall Back on Azure DefaultCredential when AZURE_STORAGE_SAS and AZURE_STORAGE_KEY aren't set
# Description
When connecting to an Azure storage account, if AZURE_STORAGE_SAS and AZURE_STORAGE_KEY are not set, use Azure SDK's `DefaultCredential` to get a bearer token to access a storage account. 

`DefaultCredential` supports getting auth tokens for from multiple different sources, as well as supports authenticating using a service principal or a managed identity in Azure Functions or within Azure virtual machines.

**Use Case**
Falling back to `DefaultCredential` (on both the rust and python sides) opens up configuration-less authentication for services like Azure Functions.

**Related Issue(s)**
This should resolve this issue as well: https://github.com/delta-io/delta-rs/issues/149

roeap:
# Description

This PR extends the available options to authorize the azure storage backend against a storage account.

- add `AZURE_STORAGE_CONNECTION_STRING`
- Fall back on `DefaultCredential` if no other option is available
- `AZURE_STORAGE_ACCOUNT` is still required for all bur connection string.
- use latest azure crates

The handling of the container client, specifically handling client updates on token expiry is heavily inspired by how the rusoto crate managed credentials internally. 

# Related Issue(s)

- closes #149
- closes #255

the somewhat awkward implementation in `list_objs` may be related to Azure/azure-sdk-for-rust#377, but I lack expereince with the lifetime handling in rust, so I may just have missed something trivial.

While technically not being worked on, if we decide to adopt the test setup using azurite, the currently ignored azure test could be removed, which may be considered a fix for #59 as well. 

# Documentation


roeap:
@houqp , @fvaleye I tried implementing more storage options which turned out to be a bit more involved then expected. Since this includes many thins fairly new to be - especially in the rust world - I'd be very happy to get some feedback if this is going in the right direction. Specifically when it comes to handling the refresh of the container client.

houqp:
@roeap I am not very familiar with azure sdk, so I will need bit more time to read through the code this weekend.

houqp:
cc @thovoll since he is the expert of Azure rust binding :)

thovoll:
@roeap please update PR description based on removing Azurite, thanks!

thovoll:
@roeap to what degree can we use the [current testing approach](https://github.com/delta-io/delta-rs/blob/222837cf8f5e09379472ce2ad1c5db0d5bc8a9b6/rust/tests/azure_test.rs#L16) for this new code?

build/setup_azurite.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
thovoll, line 1: Do we still need this file?

docker-compose.yml
```diff
@@ -28,3 +28,28 @@ services:
...
+      - azurite
+    entrypoint: "/bin/bash"
+    command:
+      - /setup_azurite.sh
+    volumes:
+      - "./build/setup_azurite.sh:/setup_azurite.sh"
+      - "./rust/tests/data/golden:/data/golden"
+      - "./rust/tests/data/simple_table:/data/simple_table"
+      - "./rust/tests/data/simple_commit:/data/simple_commit"
+      - "./rust/tests/data/concurrent_workers:/data/concurrent_workers"
```
thovoll, line 55: Remove Azurite stuff?

thovoll (changes requested):
Looks good overall, thanks for tackling this! I'd like to see the remaining Azurite code removed and the existing test modified.

build/setup_azurite.sh
roeap: done

docker-compose.yml
roeap: done

roeap:
I removed the azurite testing code and had a look into how tests are currently done. To me it seems that the account being used may actually not even exist any more. As is the test is ignored, and when trying to ruin it, I get dns not found. Maybe @rtyler can tell us if that account still exists, and if so, if it should still be used.

Beyond that there are some challenges left to actually test the core of this PR. The "static" authorization schemes are actually quite similar. However the most interesting part - authenticating with a service principal - would not be tested. What I can say is that it worked locally using a work account, but that does not help much ... @thovoll any ideas how to test this scenario as well?

thovoll:
Thanks @roeap. Ultimately, an approach like the mock testing framework that is being used by the Azure Rust SDK would make sense for integration tests. Until then we are stuck with e2e testing using real Azure resources. 

I found the same DNS issue when running the ignored test last night and pinged @rtyler as well, but instead of resurrecting this, I propose we use the same approach that is used in the Azure Rust SDK where the (e2e) tests have to be run manually by developers on their local machine, using their own Azure resources (requires setting env variables to make the various auth schemes work). In CI, the integration tests are just `cargo check`ed (the integration tests using the mock testing framework are actually executed in CI). PR submitters would have to attest to the fact that they ran the e2e tests locally, like you did above. This is not ideal, but a incremental improvement nonetheless.

thovoll:
BTW, similar issues are being discussed over in the Azure Rust SDK [here](https://github.com/Azure/azure-sdk-for-rust/issues/428#issuecomment-965881344).

thovoll:
@roeap please also update the PR description to remove the Azurite references. Thanks!

houqp:
The change looks good to me overall, agree with @thovoll on testing :)

roeap:
Thanks for the feedback. Just for me to understand, in terms of testing to idea would now be to mirgate some of the S3 tests, mark them with ignored, and maybe write a few sentences for devs to understand how to do e2e testing locally. In subsequent PRs we would then gradually improve testing to eventually have it run in CI - ideally using the pipelines architecture from the azure sdk repo? @thovoll @houqp

thovoll:
I think we would leave the S3 tests alone, since they can work with the localstack emulator.

The [Azure test](https://github.com/delta-io/delta-rs/blob/main/rust/tests/azure_test.rs) would be run locally using your own Azure Storage account, I think you have already done this correct?

If you're up for it as part of this PR, you could [modify this to not hard code those 3 values](https://github.com/delta-io/delta-rs/blob/c5571706e09b97499f5f3546cdea2a2c650b4d40/rust/tests/azure_test.rs#L12-L16) and document how to run this test locally (something light, just saying that "the following 3 env variables have to be set").

thovoll:
And yes, if we have consensus we can eventually add a mock testing framework for the Azure tests so we can run them in CI without needing an emulator or real Azure resources.

roeap:
@thovoll @houqp - i updated the azure_test file and included a description how to setup and execute tests locally.

thovoll:
Thanks @roeap, looks good! I think this is good to merge. @houqp

houqp:
awesome, @roeap do you mind resolving the conflicts in Cargo.lock so we can merge it?

roeap:
Thanks @houqp @thovoll for your support/reviews. The conflicts should be resolved now.

houqp (approved):
LGTM

houqp:
Thank you @roeap for the hard work and @thovoll for the review!

PR #498, opened 2021-11-18
Fix nullable partition values

mosyp:
If checkpoint contains the partition with null value, the `populate_hashmap_with_option_from_parquet_map` will fail to read it because it expects the value to be present, despite the map being `HashMap<String, Option<String>>`


houqp (approved):
@mosyp could you add a test to this? I want to make sure this is covered in https://github.com/delta-io/delta-rs/pull/465 as well.

mosyp:
k, working on it

mosyp:
@houqp tests hash blocked by https://github.com/delta-io/delta-rs/pull/503 by now ready

xianwill (approved):

PR #499, opened 2021-11-18
Implement missing Azure storage backend methods

### LINKED ISSUE #489: Support writes to Azure Storage
# Description
The delta-rs Azure Storage backend currently doesn't support write operations:
https://github.com/delta-io/delta-rs/blob/50a53cad084f2aa65ed0d6c0f1f5ad620a785530/rust/src/storage/azure.rs#L1
As a result, Delta Tables can not be written to Azure Storage.

Several methods need to be implemented:
https://github.com/delta-io/delta-rs/blob/50a53cad084f2aa65ed0d6c0f1f5ad620a785530/rust/src/storage/azure.rs#L192
https://github.com/delta-io/delta-rs/blob/50a53cad084f2aa65ed0d6c0f1f5ad620a785530/rust/src/storage/azure.rs#L196
https://github.com/delta-io/delta-rs/blob/50a53cad084f2aa65ed0d6c0f1f5ad620a785530/rust/src/storage/azure.rs#L200

The delta-rs Azure Storage backend currently uses the [Blob API](https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-rest-api) to access the storage account (`ContainerClient`), which does not support atomic rename (rename_obj_noreplace). We need to use the [ADLS Gen2 API](https://docs.microsoft.com/en-us/rest/api/storageservices/data-lake-storage-gen2), which does support atomic rename. 

[Both APIs can be used in parallel](https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-multi-protocol-access) with ADLS Gen2 storage accounts, but we may want to simply switch all operations over to the ADLS Gen2 API unless there is an advantage to using the Blob API for some operations.

The other operations required by the delta-rs Azure Storage backend are:
https://github.com/delta-io/delta-rs/blob/50a53cad084f2aa65ed0d6c0f1f5ad620a785530/rust/src/storage/azure.rs#L122
https://github.com/delta-io/delta-rs/blob/50a53cad084f2aa65ed0d6c0f1f5ad620a785530/rust/src/storage/azure.rs#L141
https://github.com/delta-io/delta-rs/blob/50a53cad084f2aa65ed0d6c0f1f5ad620a785530/rust/src/storage/azure.rs#L157

For atomic rename to work, the actual Azure Storage Account resource being used [must have hierarchical namespaces enabled](https://docs.microsoft.com/en-us/azure/storage/blobs/create-data-lake-storage-account#enable-the-hierarchical-namespace), which is what makes it an "ADLS Gen2" storage account.

Apologies for any remaining confusion, this is a bit tricky to explain.

Also, the Azure Rust SDK [doesn't yet support write operations via the ADLS Gen API](https://github.com/Azure/azure-sdk-for-rust/issues/356), but work is underway.

**Use Case**
Write Delta Tables to Azure Storage

**Related Issue(s)**
None

thovoll:
# Description

- Adopt latest Azure Rust SDK from Github.
- Implement missing Azure storage backend methods.
- Temporarily remove all auth support except shared key (future PR will add this back in a different way).
- Switch from the `abfs://` to `adls2://` URI scheme (since we don't use the ABFS Hadoop driver behind the scenes).
- Add tests.
- Related: #489 
- Closes: #545


rust/src/storage/azure/mod.rs
```diff
@@ -22,6 +22,7 @@
 
 use azure_core::HttpError as AzureError;
 use azure_storage::blob::prelude::*;
+// use azure_storage::data_lake::prelude::*;
```
rtyler, line 25: Rather than comment this out, can this just be removed?

rust/Cargo.toml
```diff
@@ -33,7 +33,7 @@ reqwest = { version = "0.11", default-features = false, features = ["rustls-tls"
 
 # Azure
 azure_core = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "2f2caff49b3e1ba08f430f82e16994670dddda95" }
-azure_storage = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "2f2caff49b3e1ba08f430f82e16994670dddda95", features = ["blob", "account"] }
+azure_storage = { git = "https://github.com/Azure/azure-sdk-for-rust", optional = true, rev = "2f2caff49b3e1ba08f430f82e16994670dddda95", features = ["account", "blob", "data_lake"] }
```
rtyler, line 36: it doesn't appear that the `data_lake` feature is being used so can it be removed from the pull request, or is there still more work upstream coming to support this

thovoll: Thanks @rtyler. Sorry, went on vacation :) The goal of this PR is to switch everything to the ADLS Gen2 API. Upstream is ready enough at this point, but there will be more upstream changes coming that we can adopt in delta-rs later on.

rust/src/storage/azure/mod.rs
thovoll: When I promote it from draft PR, `azure_storage::blob` will be gone and we'll only use `azure_storage::data_lake`.

thovoll:
@rtyler @houqp @roeap this PR is ready to be reviewed. I'm still working on a test for a transaction commit, but there's a test for creating a delta table that passes.

We are still using the Blob API in addition to the ADLS Gen2 API at this point, and probably will continue to do so because some operations are more efficient using the Blob API.

thovoll:
@rtyler @houqp @roeap I added a tx commit to the test.

roeap:
@thovoll - Reviewed your changes and they look good to me. A couple of things - most of them you mentioned - we should then address in a follow up PR.

- With the datalake crate now fully supporting the mock testing framework, we should enable it here, so we can run the tests in CI/CD
- We should now be able to avoid eagerly collecting all items in the list operation, since the client lifetime is no longer tied to the list return.
- further improvements adopting lates changes in the azure crates were mentioned by you in the comments :).

thovoll:
Thanks @roeap. Agree with the follow-up work.

houqp (approved):

houqp:
Thank you for all your work on the azure support @thovoll and @roeap :D Will give @rtyler couple more days to review.

thovoll:
Thanks @houqp!

houqp (approved):

Issue #488, opened 2021-11-02
Cargo build not work due to "field is never read: `root`"


viirya:
# Environment

**Delta-rs version**: current main branch

**Binding**: rs

**Environment**:
- **Cloud provider**:
- **OS**: 
- **Other**:

cargo 1.58.0-nightly (6c1bc24b8 2021-10-24)
rustc 1.58.0-nightly (db062de72 2021-11-01)

***
# Bug

**What happened**:

rust compiler begins to complain about this recently:

```rust
error: field is never read: `root`
  --> rust/src/storage/file/mod.rs:34:5
   |
34 |     root: String,
   |     ^^^^^^^^^^^^
   |
note: the lint level is defined here
  --> rust/src/lib.rs:45:9
   |
45 | #![deny(warnings)]
   |         ^^^^^^^^
   = note: `#[deny(dead_code)]` implied by `#[deny(warnings)]`

error: could not compile `deltalake` due to previous error
```

I don't see this in few days ago.

**What you expected to happen**:

`cargo build` should be able to build the project.

**How to reproduce it**:

Run `cargo build`.

**More details**:

houqp:
Rust compiler often adds new linting rules on version updates, I think this is a valid warning and it should be safe to remove the root field.

mgabs:
I'd take a stab at this.

mgabs:
rust member lib is compiling fine, ruby however is throwing the following warnings.
```
 --> ruby/src/lib.rs:5:1
  |
5 | #[macro_use]
  | ^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused variable: `rtself`
  --> ruby/src/lib.rs:53:5
   |
53 |     rtself,
   |     ^^^^^^ help: if this is intentional, prefix it with an underscore: `_rtself`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: `extern` fn uses type `AnyObject`, which is not FFI-safe
  --> ruby/src/lib.rs:54:46
   |
54 |     fn ruby_table_new(table_uri: RString) -> AnyObject {
   |                                              ^^^^^^^^^ not FFI-safe
   |
   = note: `#[warn(improper_ctypes_definitions)]` on by default
   = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
   = note: this struct has unspecified layout

warning: `extern` fn uses type `rutie::RString`, which is not FFI-safe
[...]

PR #500, opened 2021-11-18
Add bool config enableExpiredLogCleanup

viirya:
# Description

This is the follow up work of #484. This patch adds one bool config `enableExpiredLogCleanup` which is also documented in official delta lake implementation. The config controls whether to clean up expired checkpoints and logs.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -31,6 +31,7 @@ mod rename;
 /// * Support for other platforms are not implemented at the moment.
 #[derive(Default, Debug)]
 pub struct FileStorageBackend {
+    #[allow(dead_code)]
     root: String,
 }
 
--- a/rust/src/checkpoints.rs
+++ b/rust/src/checkpoints.rs
@@ -97,7 +97,7 @@ pub async fn create_checkpoint_from_table_uri(
     )
     .await?;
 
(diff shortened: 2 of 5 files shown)


rust/src/storage/file/mod.rs
```diff
@@ -31,6 +31,7 @@ mod rename;
 /// * Support for other platforms are not implemented at the moment.
 #[derive(Default, Debug)]
 pub struct FileStorageBackend {
+    #[allow(dead_code)]
     root: String,
```
houqp, line 35: I think we can just remove this, see the issue you filed ;) https://github.com/delta-io/delta-rs/issues/488

houqp (approved):

houqp:
@mosyp @xianwill could you also help review this?

rust/src/storage/file/mod.rs
viirya: Oh, I saw someone wants to take that. :)

mosyp (approved):

mosyp:
Thanks!

viirya:
Thank you @houqp @mosyp !

houqp:
Thank you for the new feature @viirya :D

djouallah:
is this exposed in the python API ?

PR #503, opened 2021-11-24
Fix cleanup_expired_logs timestamp

mosyp:
1)  fix cleanup_expired_logs which used `timestamp()` which are seconds but not millis
2) fix delete_expired_delta_log_in_checkpoint by adding missing check on version `2` and using `modified` on the actual file (which is used in comparison rather then log action json)


mosyp:
cc @viirya

xianwill (approved):

viirya (commented):
lgtm

PR #510, opened 2021-11-29
Retry S3 get request on 500 Internal Server Error

mosyp:
# Description
We're getting 500 errors in kafka-delta-ingest on get requests (especially with several workers), which is expected and such requests have to be retried
```
<Error><Code>InternalError</Code><Message>We encountered an internal error. Please try again.</Message>
```
 
We've already tested this in high load environments and this fix helped us avoid unnecessary restarts https://github.com/delta-io/delta-rs/tree/kdi-partition-values-quick-fix.

I've tried to find some universal approach to retry every s3 requests, but couldn't figure out one due to rusoto restrictions

### DIFF
--- a/rust/src/storage/s3/mod.rs
+++ b/rust/src/storage/s3/mod.rs
@@ -11,8 +11,9 @@ use log::debug;
 use rusoto_core::{HttpClient, HttpConfig, Region, RusotoError};
 use rusoto_credential::AutoRefreshingProvider;
 use rusoto_s3::{
-    CopyObjectRequest, Delete, DeleteObjectRequest, DeleteObjectsRequest, GetObjectRequest,
-    HeadObjectRequest, ListObjectsV2Request, ObjectIdentifier, PutObjectRequest, S3Client, S3,
+    CopyObjectRequest, Delete, DeleteObjectRequest, DeleteObjectsRequest, GetObjectError,
+    GetObjectOutput, GetObjectRequest, HeadObjectRequest, ListObjectsV2Request, ObjectIdentifier,
+    PutObjectRequest, S3Client, S3,
 };
 use rusoto_sts::{StsAssumeRoleSessionCredentialsProvider, StsClient, WebIdentityProvider};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


rust/src/storage/s3/mod.rs
```diff
@@ -856,6 +851,37 @@ impl dyn LockClient {
     }
 }
 
+const S3_GET_INTERNAL_SERVER_ERROR_RETRIES: usize = 10;
```
xianwill, line 854: I think this should be an option with default. `10` seems like a fine default value.

xianwill (approved):

rust/src/storage/s3/mod.rs
mosyp: ok, working on it

rust/src/storage/s3/mod.rs
```diff
@@ -856,6 +851,37 @@ impl dyn LockClient {
...
+
+async fn get_object_with_retries(
+    client: &S3Client,
+    bucket: &str,
+    key: &str,
+) -> Result<GetObjectOutput, RusotoError<GetObjectError>> {
+    let mut tries = 0;
+    loop {
```
fvaleye, line 862: WDYT of having a small time

rust/src/storage/s3/mod.rs
```diff
@@ -856,6 +851,37 @@ impl dyn LockClient {
...
+                ..Default::default()
+            })
+            .await;
+        match result {
+            Err(RusotoError::Unknown(e))
+                if e.status.is_server_error() && tries < S3_GET_INTERNAL_SERVER_ERROR_RETRIES =>
+            {
+                log::warn!("Got {:?}, retrying", e);
+                tries += 1;
+                continue;
```
fvaleye, line 876: WDYT of having a tiny sleep before retrying right after? Retry with the exponential backoff would be perfect.

mosyp: The error is very rare, like 5-6 times per day for 2-3 topics with 5-10k msg/s rate. And it usually was resolved with single retry, maybe I saw once two. So I figured that the time in between would be redundant. But again, that's just for our use case

fvaleye:
@mosyp I was really surprised that the S3 retry was not managed by rusoto for these internal errors: https://github.com/rusoto/rusoto/issues/234

If you want another point of view for this workaround: https://github.com/abetterinternet/prio-server/pull/597/files

rust/src/storage/s3/mod.rs
mosyp: Oh cool ty, will take a look at it tomorrow

mosyp:
@fvaleye
Oh cool ty, will take a look at it tomorrow

fvaleye (dismissed):
LGTM

mosyp:
@fvaleye the solution in the links you've shared seems huge to implement within this PR and current needs, so I've opened https://github.com/delta-io/delta-rs/issues/514. 

Also, since rusoto is not maintained anymore, we can give some hope for https://github.com/awslabs/aws-sdk-rust to do the job in a future

xianwill (approved):

PR #512, opened 2021-11-29
Fix null value handling in parquet map.

### LINKED ISSUE #511: Action value for HashMap in parquet could be null
# Environment

**Delta-rs version**: 0.4.1

**Binding**: Rust

**Environment**:
- **Cloud provider**: N/A
- **OS**: Windows/Ubuntu

***
# Bug

**What happened**:

When reading delta table with null partition values from checkpoint file would fail with error: `InvalidField("Invalid partitionValues for add action: value for HashMap in parquet has to be a string")`.

We should be able to read this kind of tables correctly.

**How to reproduce it**:

Generate a delta table with null partition value, and overwrite it 10 times to trigger checkpoint.

```
sdf = spark.createDataFrame([['a', 1], [None, 2]], schema=['x', 'y'])
sdf.write.format("delta").partitionBy(["x"]).save("null_partition")
for _ in range(10):
    sdf.write.format("delta").mode("overwrite").save("null_partition")
```

Read the table:

```
let table = deltalake::open_table("./null_partition").await.unwrap();
```

It will throw an error: `InvalidAction { source: InvalidField("Invalid partitionValues for add action: value for HashMap in parquet has to be a string") }`.

Other types of action value field could also be null which might cause this problem happen.

zijie0:
# Description

Catch ParquetError with "Cannot access Null as Str" message and return None as the HashMap value.

# Related Issue(s)

- closes #511

### DIFF
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -2,6 +2,7 @@
 
 #![allow(non_snake_case, non_camel_case_types)]
 
+use parquet::errors::ParquetError;
 use parquet::record::{ListAccessor, MapAccessor, RowAccessor};
 use percent_encoding::percent_decode;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


houqp:
thanks @zijie0 , i think this is already covered by https://github.com/delta-io/delta-rs/commit/8c6bb95e9cf487578520c14605cb7d6a68c7c7c2?

zijie0:
@houqp Yes, I didn't see this PR yesterday. Should be able to close this now.

mosyp:
@zijie0 yeap, sorry that my PR took a while, we hotfixed the bug in a separate branch for our prod case and I was distracted with other issue

zijie0:
@mosyp No problem :)

houqp:
great minds think alike :100:

PR #517, opened 2021-12-02
Add support for creating Array and Map schema types

### LINKED ISSUE #513: Cannot create a new delta table with nested structures like lists and maps in the schema
# Environment

**Delta-rs version**:
Latest from git as dependency is defined as a git url. Current git commit is: https://github.com/delta-io/delta-rs/commit/52d20984dd6c59c6559dbec010c2d9aa49a78877.

**Binding**: 

**Environment**:
- **Cloud provider**: N/A
- **OS**: macOS Monterey
- **Other**: -

***
# Bug

**What happened**:

Unable to create a new delta table with nested structures like lists and maps in the schema. For example I cannot create a list type field or a map type field because all type's fields are private.

For example, this is one of the compile errors returned:
```
error[E0451]: field `r#type` of struct `SchemaTypeArray` is private

r#type: String::from("array"),
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ private field
```

To be able to create schemas with lists and maps.

**How to reproduce it**:

Try creating a `SchemaTypeArray`:

```
let array = SchemaTypeArray {
    r#type: String::from("array"),
    elementType: Box::new(...),
    containsNull: false,
};
```

All three `r#type`, `elementType`, `containsNull` properties will fail the compile process due to their visibility. 

**More details**:

The issue is because neither of `SchemaTypeArray` nor `SchemaTypeMap` have a `new` method similar to  `SchemaTypeStruct`.

andrei-ionescu:
# Description
This PR adds support for creating Array and Map schema types.

# Related Issue(s)
- closes #513 

# Documentation

### DIFF
--- a/rust/src/schema.rs
+++ b/rust/src/schema.rs
@@ -17,6 +17,9 @@ pub type DeltaDataTypeTimestamp = DeltaDataTypeLong;
 pub type DeltaDataTypeInt = i32;
 
 static STRUCT_TAG: &str = "struct";
+static ARRAY_TAG: &str = "array";
+static MAP_TAG: &str = "map";
+
 /// Represents a struct field defined in the Delta table schema.
 // https://github.com/delta-io/delta/blob/master/PROTOCOL.md#Schema-Serialization-Format
 #[derive(Serialize, Deserialize, PartialEq, Debug, Default, Clone)]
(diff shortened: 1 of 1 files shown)


houqp (approved):

andrei-ionescu:
Thank you @houqp!

houqp:
Thank you @andrei-ionescu for fixing this bug :)

PR #516, opened 2021-12-01
Update datafusion version to 6

matthewmturner:
# Description
The description of the main changes of your pull request

# Related Issue(s)

# Documentation

PR #518, opened 2021-12-03
Fix S3 list_objs and cleanup_metadata

mosyp:
The `list_objs` for s3 returned a list of `ObjectMeta` with a `path` as `key` instead of `s3_path`. E.g.
`ObjectMeta { path: 'table/_delta_log' }` instead of `ObjectMeta { path: 's3://bucket/table/_delta_log' }`. However, the `head_obj` for example returns expected s3 uri.

This was found when running `cleanup_metadata` on S3, since it uses `list_objs` to list objects. So primary testing are done within a scope of this feature. I've also replicated tests with fs to be consistent. Also note that there's no workaround to pauses as S3 doesnt allow you to change object modified metadata.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,9 +78,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "arrow"
-version = "6.1.0"
+version = "6.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a900a6164aa0abf26d7a6dfea727116a3619456decfbf573e60c7615bf49e84c"
+checksum = "46cb449d481a6b13ea4e82e9fc662ea3e29cb5dc0a310fc42e27592cef3cb611"
 dependencies = [
  "bitflags",
  "chrono",
@@ -397,9 +397,9 @@ dependencies = [
 
 [[package]]
 name = "comfy-table"
-version = "4.1.1"
+version = "5.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "11e95a3e867422fd8d04049041f5671f94d53c32a9dcd82e2be268714942f3f3"
+checksum = "c42350b81f044f576ff88ac750419f914abb46a03831bb1747134344ee7a4e64"
 dependencies = [
  "strum",
  "strum_macros",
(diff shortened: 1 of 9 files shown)


rust/src/checkpoints.rs
```diff
@@ -224,20 +211,17 @@ async fn flush_delete_files<T: Fn(&(DeltaDataTypeVersion, ObjectMeta)) -> bool>(
...
-/// the `logRetentionDuration` property of the Delta Table, 30 days by default.
-async fn cleanup_expired_logs(
-    version: DeltaDataTypeVersion,
+async fn cleanup_expired_logs_for(
+    until_version: DeltaDataTypeVersion,
     storage: &dyn StorageBackend,
-    state: &DeltaTableState,
+    log_retention_timestamp: i64,
     table_uri: &str,
 ) -> Result<i32, DeltaTableError> {
```
mosyp, line 219: private API slightly changes so it could be tested

xianwill (approved):

rust/src/storage/s3/mod.rs
```diff
@@ -955,6 +954,9 @@ mod tests {
             s3_storage_options::AWS_WEB_IDENTITY_TOKEN_FILE,
             "token_file",
         );
+        std::env::remove_var(s3_storage_options::AWS_S3_POOL_IDLE_TIMEOUT_SECONDS);
+        std::env::remove_var(s3_storage_options::AWS_STS_POOL_IDLE_TIMEOUT_SECONDS);
+        std::env::remove_var(s3_storage_options::AWS_S3_GET_INTERNAL_SERVER_ERROR_RETRIES);
```
mosyp, line 960: I was getting random errors, I believe it was caused by unpredictable order of execution since it uses the mutual state, envs. This fixes that

Cargo.lock
```diff
@@ -584,7 +584,8 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "6.0.0"
-source = "git+https://github.com/apache/arrow-datafusion?rev=e62cb455955e120cfe9d8935aa3ebc8726369d15#e62cb455955e120cfe9d8935aa3ebc8726369d15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "79e4a8a1f1ee057b2c27a01f050b9dffe56e8d43605d0201234b353a3cc1eb2f"
```
mosyp, line 588: lock was not updated in https://github.com/delta-io/delta-rs/pull/516 which caused build failures

rust/src/delta_arrow.rs
```diff
@@ -125,7 +125,7 @@ impl TryFrom<&schema::SchemaDataType> for ArrowDataType {
             schema::SchemaDataType::r#struct(s) => Ok(ArrowDataType::Struct(
                 s.get_fields()
                     .iter()
-                    .map(|f| <ArrowField as TryFrom<&schema::SchemaField>>::try_from(f))
+                    .map(<ArrowField as TryFrom<&schema::SchemaField>>::try_from)
```
mosyp, line 128: clippy got updated and required these changes

xianwill (approved):

PR #519, opened 2021-12-03
Add datafusion examples to docs

### LINKED ISSUE #515: Add documentation for using Delta Table with datafusion
# Description
As a user of delta-rs I would like to see examples of how to use the datafusion extension to query delta tables in the rust docs.

This would include examples for querying both local and s3 delta tables.

**Use Case**

**Related Issue(s)**

matthewmturner:
# Description

# Related Issue(s)

closes #515
# Documentation

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -45,7 +45,7 @@
 //! # Querying Delta Tables with Datafusion
 //!
 //! Querying from local filesystem:
-//! ```rust,no_run
+//! ```no_run
 //! use std::sync::Arc;
 //! use datafusion::execution::context::ExecutionContext;
 //!
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -45,7 +45,7 @@
 //! # Querying Delta Tables with Datafusion
 //!
 //! Querying from local filesystem:
-//! ```no_run
+//! ```ignore
 //! use std::sync::Arc;
 //! use datafusion::execution::context::ExecutionContext;
 //!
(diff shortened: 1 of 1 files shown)


matthewmturner:
Started putting together the docs, based on test in `delta_datafusion.rs`.  I havent been able to actually get querying delta tables to work locally yet, but wanted to publicize the work in case any comments.

matthewmturner:
@houqp can delta tables be queried with the datafusion python bindings?  i can add something for that as well.

matthewmturner:
and this might be a silly question, but why does the datafusion feature have the `ext` suffix but `s3` and `azure` dont?

matthewmturner:
i looked into this on the python side and it seems that the `ExecutionContext` is missing the `register_table` method that is used in rust to query `DeltaTables`.  I'll create an issue there to add that functionality as it would be nice to query from python.

now im going to try querying s3 `DeltaTable` in rust and adding docs for that then i think this should be good.

houqp:
> @houqp can delta tables be queried with the datafusion python bindings? i can add something for that as well.

Probably not yet because I don't think we support taking a list of file paths yet in the current table provider implementation :(

> and this might be a silly question, but why does the datafusion feature have the ext suffix but s3 and azure dont?

Because the `datafusion` feature conflicts with the `datafusion` crate name. But I learned later that we could use optional dependencies as features directly. I haven't tested it myself, if this works, it would be better to use `datafusion` as the feature name going forward in our docs.

matthewmturner:
> Probably not yet because I don't think we support taking a list of file paths yet in the current table provider implementation :(
> 
@qphou can you expand on that? I thought since you can query with datafusion in Rust it was just a matter of exposing that via Python with something like a `register_table` on context.

matthewmturner:
@houqp ready for review

houqp (approved):

houqp:
Thank you @matthewmturner !

Issue #509, opened 2021-11-28
High-level writer API


roeap:
# Description

Recently I have been trying to understand the current state of writer implementations for `delta-rs` and what could be the desired state for a high level writer API. All of it in the hopes to contribute to this effort. Looking into this it seems to me there can be a wide range of complexity for a potential implementation - from complex to very complex :D.

This may be already a too simplistic/naive view, but it seems to me on a lower level all operations would boil down to a number of add/remove actions to update the log and associated data that may contain new data or represent existing data with some records removed or updated. The files to be removed would depend on the specifics of a higher level operation. Based on what I understand right now, the files to be added would "only" depend on the partitioning of the table and the desired max file size for files written to the table.

If I understood the current implementations from this crate and `kafka-delta-ingest` correctly, they are mostly based on json messages to be written, which are distributed to several buckets based on partitioning of the table. On the arrow record batch level we already seem to assume individual batches contain data belonging to exactly one partition. Trying to find a way to generically implement partitioning for record batches it already seemed to me it may be worthwhile to consider using a higher level crate for this rather then re-implementing this using the arrow compute kernels.

Making the question a bit more specific, I see some options to move forward with this.

1) Minimal only

The writer API would expect pre-partitioned data (and panic otherwise) along with remove actions, write the record batches and wrap everything in a transaction. Current implementations for json data could be retained.

2) Automatic partitioning

The writer API can take generic record batches and remove actions. Partitioning would be implemented using arrow kernels. This approach could (I think) already support the `write` operations (add / overwrite) as they are implemented in the delta reference implementation. 

3) Full featured using external query engine

The writer API could support all higher level operations like `update`, `merge`, `delete` etc, but depend on an external crate like datafusion or polars to implement scanning individual files and passed data.

I assume each level could build on lower levels, but might look different based on how far we want to take this within this crate. Also hoping my thinking with regards to this topic did not go off in an entirely wrong direction and I did not miss somethting obvious. 

Any feedback on this is greatly appreciated.

houqp:
I think you are spot on with your analysis @roeap :) I don't have much to add. In step 3, we can also include `optimize`. I also think it would be a good idea to use arrow record batches as the input serialization format.

houqp:
> on how far we want to take this within this crate.

I want to take it very far, to the point where we can use it to process delta tables in a distributed setup using systems like ballista ;)

roeap:
I think this Issue has served its purpose :).

PR #523, opened 2021-12-13
[WIP] Towards high level writer

roeap:
# Description

This PR tries to consolidate existing efforts and lay the groundwork for advancing the high-level writer APIs. When preparing this PR, I learned a lot about rust and the internals of this crate. Thus I assume there is lots to improve.  The main part of low level internals for generating stats etc is taken from this crate or `kafka-delta-ingest`. 

There is still a lot to do before everything works and is tested as required, but I aimed for getting things to work end to end to collect some feedback on the general direction.

# Related Issue(s)

- towards #509

# Documentation

Roughly there are three levels. the writer api handles buffering data, writing the physical files and generating actions with the respective statistics. The transaction api wraps writing the low level api and handles the actions internally. The command api handles complete operations around transactions. Below are some simple code snipplets how one might perform writing data at the different levels.

## Writer API

- low level API handling record batches largely based on existing code
- the most critical (I think) update is partitioning of data based on record batches.
- generate Add actions for files written to storage

```rs
let dt: DeltaTable = ...
let batches: Vec<RecordBatch> = ...
let mut writer = DeltaWriter::for_table(dt, options);

for batch in batches {
    // batches are partitioned according to partition columns and
    // partition data written to PartitionBuffers
    writer.write(&batch).await?;
}

// files are persisted to disk and file statistics generated for add actions
// this should also handle segmenting files according to desired file size
let adds: Vec<Add> = writer.flush().await?;
```

## Transaction API

- write record batches directly and have transaction mechanics mostly hidden.

```rs
let dt: DeltaTable = ...
let batches: Vec<RecordBatch> = ...
let operation: DeltaOperation = ...

let mut txn = table.create_transaction(None);
// ... maybe add some actions to transaction ...

// files are directly written to disk, and add actions tracked internally
txn.write_files(batches).await?;

let _ = txn.commit(Some(operation)).await?;
```

## Command API

- build on top of abstractions from the `DataFusion` crate
- individual commands are implemented as datafusion `ExecutionPlan`s
- can hopefully serve as basis to support spark-like SQL semantics for delta

```rs
let table_uri = ...
let batches: Vec<RecordBatch> = ...

let commands = DeltaCommands::try_from_uri(table_uri).await.unwrap();
commands.write(batches).await.unwrap();
```

## Questions

- first and foremost, is this going in the right direction
- what to look out for to have as little data copies as possible
- should we do automatic schema updates on the table?
- should there be defaults / base entries for commit infos in operations
- how to best handle `DeltaTable` instances between different structs and specifically when we distribute via datafusion.


roeap:
@houqp, this is probably the most involved thing o ever did in rust, so any feedback is greatly appreciated :).

houqp (commented):
Thank you @roeap , this is an epic set of changes! Really cool to see it being hooked up with datafusion execution plan :)

I only quickly skimmed through the PR. I probably won't be able to go through it in details until this weekend due to the size of the change. But on a high level, it looks pretty good to me.

> first and foremost, is this going in the right direction

I think so, the abstractions look very clear to me.

> what to look out for to have as little data copies as possible

Could you elaborate what you mean by data copies? Did you mean arrow record batches?

> should we do automatic schema updates on the table?

Eventually yes we should support schema evaluation as part of the transaction :) But it's a lot of work so I think it's better to handle it as follow ups.

> should there be defaults / base entries for commit infos in operations

I think so, it will be useful metadata to auditing purpose.

> how to best handle DeltaTable instances between different structs and specifically when we distribute via datafusion.

I don't have a good answer to this without going through the implementation in details. I think we might need to introduce changes in datafusion as well to better support this.

rust/src/commands/transaction.rs
```diff
@@ -0,0 +1,118 @@
...
+            let serialized_actions = arrow::array::as_string_array(batch.column(0));
+            let mut new_actions = (0..serialized_actions.len())
+                .map(|idx| serde_json::from_str::<Action>(serialized_actions.value(idx)))
+                .into_iter()
+                .collect::<Result<Vec<_>, _>>()
+                .map_err(to_datafusion_err)?;
+            actions.append(&mut new_actions);
+        }
+
+        txn.add_actions(actions);
```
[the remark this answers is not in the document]
roeap: The main idea was that the child processes would be responsible for for generating any type of action that we simply add to the transaction here i.e. these could be add or remove. If these need to be consolidated, right now the input plan would have to take care of that. Initially I added multiple input plans, but that turned out to be more involved making sure they execute concurrently. In a follow up. I wanted to give delete a try to see how this works with mixed actions. But since this is still file by file I guess it should still work.. The real test would be optimize :)

rust/src/commands/transaction.rs
```diff
@@ -0,0 +1,118 @@
...
+        let mut table = open_table(&self.table_uri)
+            .await
+            .map_err(to_datafusion_err)?;
+        let table_exists = check_table_exists(&table)
+            .await
+            .map_err(to_datafusion_err)?;
+
+        if !table_exists {
+            todo!()
+        }
```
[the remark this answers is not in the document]
roeap: absolutely! I must have changed that sometime, since I don't think we can assume the table to already be initialized for every transaction we may have to create a "raw" table here instead of using `open_table`. Then again creating a table always requires the meta-data to be passed around, which is inconvenient.

rust/src/write/writer.rs
```diff
@@ -0,0 +1,580 @@
...
+            stringified_partition_value(&c.values.slice(range.start, range.end - range.start))
+                .unwrap()
+        });
+
+        let mut partition_values = HashMap::new();
+        for (key, value) in partition_columns.clone().iter().zip(partition_key_iter) {
+            partition_values.insert(key.clone(), value);
+        }
+
+        let batch_data = arrow_schema
```
roeap, line 408: I'm not entirely sure how to interpret this line form the Delta protocol. 

> Values for all partition columns present in the schema MUST be present for all files in the table.

Does this mean, the partition values should not be omitted in the data files? We are writing them in the log, so they are somehow available, but right now not written in the data files. The previous implementation did write the data.

houqp: Interesting, I was under the impression that we won't need need to duplicate partition column values in the data files. Let me double check this with the reference implementation tomorrow.

rust/src/commands/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+                min_reader_version: 1,
+                min_writer_version: 2,
+            },
+        };
+        let plan = Arc::new(CreateCommand::try_new(
+            &self.table.table_uri,
+            operation.clone(),
+        )?);
+
+        self.execute(operation, plan).await
```
houqp, line 144: perhaps wrapping create command in a transaction is an overkill here and we can just call `table.create()` here?

rust/src/commands/transaction.rs
```diff
@@ -0,0 +1,171 @@
...
+        Arc::new(OPERATION_SCHEMA.clone()),
+        vec![Arc::new(serialized)],
+    )?)
+}
+
+/// Write command
+#[derive(Debug)]
+pub struct DeltaTransactionPlan {
+    table_uri: String,
+    input: Arc<dyn ExecutionPlan>,
```
houqp, line 49: One problem I noticed with `DeltaTransactionPlan` is transaction actions need to be prepared for a specific version of a table so we can detect and resolve transaction conflicts. However, within its execute method, we always load the latest version of the table, which might be newer than the version we used to generate the actions.

For example, consider an overwrite table write actions generated for version 1, then another writer wrote a bunch of new files into version 2 before we commit the overwrite write. When we try to commit the overwrite transaction as version 3, we will miss the new files that got added in version 2 resulting in incorrect table state.

One way to solve this problem is to add a table version attribute to `DeltaTransactionPlan`. Alternative, it might be better to get rid of the `DeltaTransactionPlan` entirely and let individual command execution plans manage the transaction struct internally. A table transaction is just a simple json file write, so there is not much value in managing this write as an execution plan. Execution plan is especially useful for operations needs to be executed in a distributed fashion, for example, writing out partitioned parquet files for each add action.

rust/src/write/writer.rs
houqp: I double checked and can confirm that partition values should not be written out into the data file. This sentence in the spec is a little bit mislead to me at least. Worth filing an upstream ticket to further clarify it.

rust/src/delta.rs
```diff
@@ -1731,29 +1811,29 @@ mod tests {
...
-                        "delta-rs".to_string(),
-                        serde_json::Value::String(crate_version().to_string()),
-                    );
-                    expected.insert(
-                        "timestamp".to_string(),
-                        serde_json::Value::Number(timestamp),
-                    );
-                    assert_eq!(modified_action, expected)
-                }
+                // Action::commitInfo(action) => {
```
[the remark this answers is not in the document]
roeap: Currently that test fails since in this PR we now create basic commit info from the `DeltaOperation` enum. If we stick with that, the test needs to be refactored and maybe moved.

rust/src/commands/transaction.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn with_new_children(
+        &self,
+        _children: Vec<Arc<dyn ExecutionPlan>>,
+    ) -> DataFusionResult<Arc<dyn ExecutionPlan>> {
+        todo!()
+    }
+
+    async fn execute(&self, _partition: usize) -> DataFusionResult<SendableRecordBatchStream> {
+        let mut table =
+            get_table_from_uri_without_update(self.table_uri.clone()).map_err(to_datafusion_err)?;
```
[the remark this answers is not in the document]
roeap: When implementing this, I also often thought, that decoupling `DeltaTransaction` from the actual table instance would be beneficial. Haven't really looked at it more closely, but is this something we should try in this PR?

rust/src/commands/transaction.rs
```diff
@@ -0,0 +1,171 @@
...
+                    .try_commit_transaction(&prepared_commit, 0)
+                    .await
+                    .map_err(to_datafusion_err)?;
+                committed_version
+            }
+            _ => {
+                let mut txn = table.create_transaction(None);
+                txn.add_actions(actions);
+                let committed_version = txn
+                    .commit_with_info(Some(self.operation.clone()), self.app_metadata.clone())
```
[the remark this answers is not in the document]
roeap: The `commit_with_info` method was actually not meant to live very long. Main reason it exists ritght now, was to propose a way how we do basic commit info using the operation but provide a way for callers to pass in custom application data to be included in the commit. When I saw that `kafka-delta-ingest` uses the commit function I wanted to not break the curretn API for now, but the hope was to change the function signature of the current `commit` method. The current `create` method on delta table also does lower level commit logic. If we decide to use the operations enum as a basis for creating commit infos, create should - I think - use that route as well. And if the call sites in kafka ingest can be adjusted, my haope was to move to a single commit function on the transaction again.

rust/src/commands/transaction.rs
roeap: Writing this I was also a bit conflicted about how to best handle wrapping things in a transaction. Making sure we commit against a fixed version seems reasonable. For my understanding though, in the example above we would have to update the log to the latest version before retrying, right? 

The reason this found its way into an execution plan is, twofold. For one i tried to mimic what spark does as best I could. There they seem to have something that conceptually looks a bit like a contextmanager in python.

```scala
override def run(sparkSession: SparkSession): Seq[Row] = {
    deltaLog.withNewTransaction { txn =>

      val actions = write(txn, sparkSession)
      val operation = DeltaOperations.Write(mode, Option(partitionColumns),
        options.replaceWhere, options.userMetadata)
      txn.commit(actions, operation)
    }
    Seq.empty
  }
```
While I may have misunderstood whats going on there, I tried to come up with an implementation that could be used in  a similar fashion, i.e. collect all actions from child plans and commit them based on the operation. Of course we can achieve this without all the overhead of an execution plan. However if we wanted to support the Delta Specific SQL syntax from spark in the future based on datafusion / ballista, I thought wrapping things in an execution plan would be the mst straight forward way to go. But again I may have missed something :)

rust/src/commands/mod.rs
roeap: On its own definitely. Main reason it exists, is that it made it much easier to use of a table needs to be created from within other commands, and to have a consistent way individual commands are implemented. Also I was unsure in cases where we create the table as part of e.g. a write command. calling `table.create()` would generate a version and then the write would create another version. in the current implementation a single version / commit is created if the table is created as part of a write command. Not sure how this is handled in the reference implementation, but since table metadata is derived from the written data, i felt it more natural to be a single commit. In that I thought moving the actions around consistently was worth the overhead of having a execution plan. I guess it all depends on how we want to handle wrapping things inside a transaction in general.

rust/src/commands/transaction.rs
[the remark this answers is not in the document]
houqp: Yeah, I think it's something we can handle as follow ups since the scope of this PR is already quite large. We should leave a comment and link to a tracking issue though :)

rust/src/commands/transaction.rs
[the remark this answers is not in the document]
houqp: Sounds good, we can perform the operation check within `commit_with_info`. My understanding for the current `create` method not needing the operation enum is because there is no commit loop within it, i.e. it only tries to perform a commit for version 0.

rust/src/delta.rs
```diff
@@ -1454,14 +1502,57 @@ impl<'a> DeltaTransaction<'a> {
...
 
         Ok(version)
     }
 
+    /// Commit transaction with additional app metadata
+    pub async fn commit_with_info(
+        &mut self,
+        operation: Option<DeltaOperation>,
+        app_metadata: Option<Map<String, Value>>,
+    ) -> Result<DeltaDataTypeVersion, DeltaTableError> {
```
houqp, line 1518: Could you add a TODO here to remind checking operation and decide whether the commit is safe to retry in a loop?

rust/src/commands/mod.rs
houqp: I agree with you and think we should keep the `CreateCommand` plan. For writes to an non-existence table, I would prefer us to combine the create table actions and add actions into a single commit too.

For this particular call site, I was thinking we can avoid all the datafusion overhead and create the table directly since we are not writing extra data.

rust/src/commands/transaction.rs
houqp: > For my understanding though, in the example above we would have to update the log to the latest version before retrying, right?

The core of optimistic concurrency control is we always assume we already have the latest version or latest view of the world when performing a commit. Optimistic concurrency control is optimized for low concurrent write transaction rate use-cases, so the chance of us not on the latest version is low. A lot of the time, there is only a single writer to delta table, in which case, the writer will always have the latest view of the table without needing to do explicit state refreshes. On top of this, refreshing the state here doesn't really help us get the latest view of the table, it will only get us a fairly recent view since other writes could still happen right after the refresh and before we commit the write. Therefore it's not buying us much here.

But the more important issue this leads to is correctness compromise. For some of the writes like upsert and delete, the prepared actions are only valid for a specific version of the table. For example, let's say you have a transaction to delete file A which only exists in version 1. Then subsequently another writer deletes the same file in version 2. When we try to commit the same delete action on version 3, we should abort and return a conflict error to the users instead of letting the commit go through. If we record the targeted version with the commit, the writer will know to commit on version 2 and error out with object already exists error. If we always refresh before the commit, it will fetch the latest version (2), then increment to 3 and commit without a way to perform conflict resolution.

I think leveraging execution plan is the right move here, we just need to make sure there is way to provide the targeted commit version as well.

roeap:
@houqp - sorry for taking a while to respond, work got in the way :).

I finally wanted to get back to this PR and get some of it merged. From your perspective, does it make more sense to try and get this merged as a whole, or split it up in some smaller pieces? The datafusion part is currently likely no longer compatible with some of the changes in the latest datafusion versions. One way could be to remove the query plans from this PR and submit it in a later one. THis will likely the part where we need to iterate on most to get to a good solution.

Alternatively I should of course merge current master, and update the query plans as well.

DO you have a preference how to proceed?

houqp:
Sorry @roeap for the late reply, still catching on all my github notifications :P 

I don't have a preference, so I recommend you pick the route that requires the least amount of work for you. My guess is splitting it up into smaller pieces will be less work for you because it reduces chances of merge conflicts.

roeap:
Closing this, since it has been addressed in #584 and #573

PR #525, opened 2021-12-18
[Python] Create PyArrow dataset fragments from delta log

wjones127:
# Description

As currently written, `DeltaTable.to_pyarrow_dataset()` will start the dataset discovery process, requiring additional IO to discover partitions and read schemas. This PR changes the implementation to populate the dataset fragments based on the delta log, which will avoid some discovery IO and also allow the datasets API to benefit from the file statistics in the log.

In detail:

 * Changed `get_file_uris()` and `get_stats()` public methods to return iterators instead of materialized vectors.
 * Added new `get_partition_values()` method that returns iterator over the partition values of each file.
 * Created new `dataset_partitions()` method in Rust `RawDeltaTable` impl, which returns the list of files and an associated partition expression (one that is guaranteed to be true for all rows in the file). Constructing this in Rust is a little more verbose in than in Python, but allows us to avoid creating intermediate objects for the Rust-Python interface. 

# Related Issue(s)

I started prototyping this based on discussion in https://issues.apache.org/jira/browse/ARROW-14730. If this is successful, I'll create issues on this repo to gradually improve the dataset functionality.

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -239,14 +239,14 @@ impl RawDeltaTable {
             .map(|res| {
                 let result = match res {
                     Ok((path, Some(stats))) => FileStats {
-                        file_path: path.to_string(),
+                        path: path.to_string(),
                         num_records: stats.num_records,
                         min_values: HashMap::new(),
                         max_values: HashMap::new(),
                         null_counts: HashMap::new(),
                     },
                     Ok((path, None)) => FileStats {
-                        file_path: path.to_string(),
+                        path: path.to_string(),
                         num_records: 0,
                         min_values: HashMap::new(),
                         max_values: HashMap::new(),
@@ -263,7 +263,7 @@ impl RawDeltaTable {
 #[pyclass]
 pub struct FileStats {
     #[pyo3(get)]
-    file_path: String,
+    path: String,
     #[pyo3(get)]
     num_records: i64,
     #[pyo3(get)]
@@ -324,7 +324,7 @@ fn deltalake(py: Python, m: &PyModule) -> PyResult<()> {
     m.add_class::<RawDeltaTable>()?;
     m.add_class::<RawDeltaTableMetaData>()?;
     m.add_class::<DeltaStorageFsBackend>()?;
-    m.add_class::<FileStats>()?;
+    // m.add_class::<FileStats>()?;
     m.add("PyDeltaTableError", py.get_type::<PyDeltaTableError>())?;
     Ok(())
 }
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -238,15 +238,17 @@ impl RawDeltaTable {
             .get_stats_iter()
             .map(|res| {
                 let result = match res {
-                    Ok((path, Some(stats))) => FileStats {
-                        path: path,
+                    Ok((path, partition_values, Some(stats))) => FileStats {
+                        path,
+                        partition_values: partition_values.clone(),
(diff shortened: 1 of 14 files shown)


houqp:
This is very cool @wjones127, I wasn't aware one can do this with the dataset api :D

rust/src/delta.rs
```diff
@@ -950,25 +950,30 @@ impl DeltaTable {
         note = "Please use the get_file_uris function instead"
     )]
     pub fn get_file_paths(&self) -> Vec<String> {
-        self.get_file_uris()
+        self.get_file_uris().collect()
     }
 
     /// Returns a URIs for all active files present in the current table version.
-    pub fn get_file_uris(&self) -> Vec<String> {
+    pub fn get_file_uris(&self) -> impl Iterator<Item = String> + '_ {
```
wjones127, line 957: I'd prefer for these to be iterator and saw similar changes in #452, so thought this would be alright. Happy to change it back if you don't want to change the public API in this way.

python/src/lib.rs
```diff
@@ -232,6 +232,116 @@ impl RawDeltaTable {
...
+            .map(|((path, partition_values), stats)| {
+                let stats = stats.map_err(PyDeltaTableError::from_raw)?;
+                let expression = filestats_to_expression(py, partition_values, stats)?;
+                Ok((path, expression))
+            })
+            .collect()
+    }
+}
+
+fn json_value_to_py(value: &serde_json::Value, py: Python) -> PyObject {
```
rtyler, line 265: I'm struggling to understand why this function needs to handle only some of the branches of the `serde_json::Value` enum rather than everything or nothing. As far as I can tell `ColumnValueStat` in this case is always going to be `Value(DeltaDataTypeLong)` (i.e. `i64`).

I don't really have feedback here, just trying to understand

wjones127: `ColumnCountStat::Value` (which is used for null counts) is always going to be `Value(DeltaDataTypeLong)`. But `ColumnValueStat::Value` could be any JSON primitive value, depending on the type of column. Minimum values of int columns will be ints, strings will be strings. See an example here:

https://github.com/delta-io/delta-rs/blob/57133e161e03a34d8902abd210dca80731ce00d1/rust/tests/data/COVID-19_NYT/_delta_log/00000000000000000000.json#L4

Which turns into the following partition expression:

```python
<pyarrow.dataset.Expression ((((((((((((((((date >= "2020-01-21") and (fips >= 1001)) and (state >= "Alabama")) and (county >= "Abbeville")) and (deaths >= 0)) and (cases >= 0)) and (state <= "Wyoming")) and (date <= "2020-05-19")) and (county <= "Ziebach")) and (fips <= 78030)) and (cases <= 198114)) and (deaths <= 20298)) and is_valid(date)) and is_valid(state)) and is_valid(county)) and is_valid(cases))>
```
AFAIK I shouldn't need to handle array or object JSON values here. Maybe I should panic on them instead?

rust/src/delta.rs
houqp: this is a good change, we should propagate this practice through out the code base :)

python/src/lib.rs
houqp: I think it would be good to throw an exception in case of unexpected json value types.

houqp (approved):

houqp:
Left two minor comments that can be addressed as follow up PRs. @wjones127 thank you for your work on this :+1:

PR #508, opened 2021-11-27
Move dynamodb lock into its own crate

viirya:
# Description

Move dynamodb lock into its own crate.

# Related Issue(s)

closes #186 

# Documentation

PR #535, opened 2022-01-06
Remove coupling of DynamoDbLockClient from S3 storage

ahmedriza:
# Description
As part of the refactoring done by https://github.com/delta-io/delta-rs/pull/508, the `LockClient` used by the S3 storage implementation got coupled to a specific implementation of the `LockClient`, i.e. `DynamoDbLockClient`.  That doesn't look to be necessary and this PR addresses that by removing that coupling and let the S3 storage support any `LockClient`. 

This is useful in cases where we may not be using S3 in AWS.  For example, one could be using MinIO, Hitachi HCP etc which implement the S3 protocol and can be used just like S3.  In these cases, a DynamoDb based lock is not required and some other implementation can be used.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/.gitignore
+++ b/rust/.gitignore
@@ -1,1 +1,8 @@
 target/
+tests/data/action_reconciliation/
+tests/data/checkpoints_tombstones/expired/
+tests/data/checkpoints_tombstones/metadata_broken/
+tests/data/checkpoints_tombstones/metadata_false/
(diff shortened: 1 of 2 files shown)


rtyler (commented):
Nice little refactor! I like it, but will wait for @mosyp the original contributor of this code to approve before merge

mosyp (approved):
Nice one, thank you @ahmedriza

PR #538, opened 2022-01-08
Add golden files and create tests #179

### LINKED ISSUE #179: Link to golden dataset instead of copying it
# Description

Currently we copy only single table from [golden dataset](https://github.com/delta-io/connectors/tree/master/golden-tables/src/test/resources/golden) and use it in tests. Eventually it'd make sense to use every table from golden dataset and either use a git submodule or separate repo to link into it

Please see a discussion in https://github.com/delta-io/delta-rs/pull/175#discussion_r605072676

wjones127:
# Description

Working to link to golden test files in delta-io/connectors. My hope is to create tests for most of these datasets and create new issues for each failing one.

Based on my initial tests, 31 of 51 can be loaded without error. I'm not checking for correctness, so some of those may be wrong. And some of those errors may be trivial (e.g. I'm pointing to the wrong directory).

# Related Issue(s)

- closes #179

# Documentation


houqp:
Last time I checked (a year ago), all of the failures are caused by the same problem that's not critical so we never prioritized the fixes :) It would be good to have a clean pass on the golden test data indeed.

wjones127:
> Last time I checked (a year ago), all of the failures are caused by the same problem that's not critical so we never prioritized the fixes :) It would be good to have a clean pass on the golden test data indeed.

Good to know! I'll not spend too long on it then if I find there's not anything new.

wjones127:
Okay I see the issue now. Many of those delta logs are missing `schemaString` in their metadata entries, [even though it's required in the log protocol](https://github.com/delta-io/delta/pull/610), but we throw an error if it's missing. The requirement is newly clarified in the spec, so maybe we should make a PR to add the schema to the golden files.

houqp:
oh yeah, that was it! I agree fixing it upstream would be the right path forward :+1:

Issue #522, opened 2021-12-09
Improve DataCatalog documentation


matthewmturner:
# Description
I would like better documentation / user guides on how to use `DataCatalog`, in particular from AWS / Glue, in both Rust and Python

**Use Case**
I am already using AWS Glue / Athena and i am starting to explore if datafusion / deltalake can be used instead.

**Related Issue(s)**

fvaleye:
Hello @matthewmturner,

Thanks for submitting your issue!

Did you already see this [documentation page](https://delta-io.github.io/delta-rs/python/usage.html#deltatable) for the Python binding? (Create a DeltaTable using a Data Catalog)

How could we improve this based on your usage?

matthewmturner:
@fvaleye thanks for quick response!  yes, i saw that.

being new to deltalake, and wanting to set it up (i.e. not using an existing deltalake), for me the part that was missing was general context of the relation between the DataCatalog and deltalake and how they can be used together (i.e. is anything in particular needed in the db / table glue metadata).

i had naively tried using my existing glue db / table and of course it didnt work as it wasnt a deltatable yet.  so i was hoping for directions how to convert existing glue db / tables to deltatables.

thinking on it now maybe thats broader than adding to docs here though.  what do you think?

matthewmturner:
Im going to look for documentation elsewhere on AWS Glue / Data Catalog and how it works with Delta Lake.  But if you think its relevant im happy to add some details here as well.

fvaleye:
Let me try to find an example of how we can load a `DeltaTable` using delta-rs with the AWS Glue Data Catalog using [OSS Delta Lake](https://github.com/delta-io/delta).

**First:**
Before writing a `DeltaTable` as a table in the Hive Metastore with Spark, you need to specify AWS Glue Data Catalog as the Metastore for your cluster:
[OSS Delta Lake documentation](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-hive-metastore-glue.html) 
[Databricks Delta Lake documentation](https://docs.databricks.com/data/metastores/aws-glue-metastore.html)

**Then:**
You could use this method in Spark for creating the Delta Table in AWS S3 (or a SQL one):

> data_frame.format("delta")
>         .option("delta.compatibility.symlinkFormatManifest.enabled", "true")
>         .option("path", path)
>         .saveAsTable(f"{database_name}.{table_name}")
> 
[...]

PR #539, opened 2022-01-09
Refresh Python usage documentation

wjones127:
# Description

I rewrote most of the Python usage documentation.

 * Reflowed the usage documentation to start with loading, then look at log introspection, and finally querying tables. At the end I created a placeholder for writing, noting that it's not yet supported.
 * Added info about supported backends and data catalogs.
 * Added more examples and guidance on how to query Delta tables. Note: I'd like to include DataFusion here, but from what I've seen that's not yet possible, right?

By rewriting this I broke any links to sections *within* the page, but not to the page itself.

# Related Issue(s)

None

# Documentation

### DIFF
--- a/python/docs/source/usage.rst
+++ b/python/docs/source/usage.rst
@@ -1,17 +1,6 @@
 Usage
 ====================================
 
-.. TODO:
-.. [ ] Add links to classes
-.. [ ] How to query data (being considerate of memory size) (Pandas vs Dask vs DuckDB vs Datafusion)
-.. [ ] Clarify history and checkpoints
-.. [ ] What's a data catalog, which ones are supported?
-.. [ ] How to load file S3 and other services
-
-
-DeltaTable
-----------
-
 .. py:currentmodule:: deltalake.table
 
 A :class:`DeltaTable` represents the state of a delta table at a particular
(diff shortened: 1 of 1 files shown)


matthewmturner:
hi @wjones127 i was interested in this from a datafusion angle as well and added example to the rust documentation for querying here #519.

i had some issues with this though and had to point to specific git commits to get it to work.  i havent tried since then though so not sure if the issue will still persist.  im less familiar with the python bindings, so take this with a grain of salt, but i believe the general structures are in place to replicate how we query in rust with the datafusion python bindings.

on a separate but somewhat related note, i am working on adding s3 support to datafusion (https://github.com/datafusion-contrib/datafusion-objectstore-s3).  my loose understanding of delta lake is that its often cloud based, so getting s3 support added to datafusion should make querying it easier.  i had actually previously tried querying deltalake on s3 and that was how i found out that datafusion didnt support it which started me on that path.

hope this helps!

houqp (dismissed):
LGTM, thanks for the detailed write up @wjones127 !

houqp:
FYI @fvaleye @zijie0

fvaleye (dismissed):
Thanks @wjones127 

LGTM! There is a minor error in the CI before merging the PR.

python/docs/source/usage.rst
```diff
@@ -12,99 +14,310 @@ Resolve partitions for current version of the DeltaTable
...
+To load the current version, use the constructor:
+
+.. code-block:: python
+
+    >>> dt = DeltaTable("../rust/tests/data/delta-0.2.0")
+
+Alternatively, if you have a data catalog you can load it by reference to a 
+database and table name. Currently only AWS Glue is supported.
+
+.. TODO: auth to data catalog?
```
fvaleye, line 34: FYI #522 explains the requirements of the Data Catalog integration.

wjones127: Thanks!

fvaleye (approved):

PR #540, opened 2022-01-10
formally verify S3 atomic rename

houqp:
# Description

Our TLA+ proof is already out of date, so I took a stab at using stateright to formally remodel our S3 atomic rename design.

I think the current proof already discovered a bug that could lead to duplicated commits in our implementation. Here is what it found:

- writer A acquires lock for version 0
- writer A paused
- writer B acquired expired lock
- writer B repairs writer A's rename, we now have version 0 saved in S3
- writer A resumes execution
- writer A sees version 0 already exists in S3
- writer A's rename call returned version already exists error, it increments the version count and tries the commit again.

You can replicate the trace by running the following command in the proofs folder:

```
cargo run --release explore --worker-count 3
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "aws/delta-checkpoint",
     "dynamodb_lock",
 ]
+exclude = ["proofs"]
 
 [profile.dev]
 split-debuginfo = "unpacked"
--- a/proofs/Cargo.toml
+++ b/proofs/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "delta-proofs"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+stateright = "0"
+num_cpus = "1"
+env_logger = "0"
+clap = "3"
--- a/proofs/README.md
+++ b/proofs/README.md
@@ -1,0 +1,10 @@
+This folder contains formal specs for various parts of delta-rs implemented using [stateright](https://github.com/stateright/stateright).
+
+Usage
+-----
+
+To model check current specs:
+
+```rust
+cargo run --release explore --worker-count 3
+```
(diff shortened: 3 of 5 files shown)


houqp:
cc @mosyp @xianwill

mosyp:
@houqp So the duplication comes from the place after worker A resumes and tries to commit the same file? I feel like this is either human error or API misslead. E.g. if yours commit is failed, you most likely want to reset yours writers. That's why I believe worker A should exit with error.
E.g. after worker A is resumed and fails to commit version 0 because it is repaired by worker B, then worker A is in invalid state. For kafka-delta-ingest we reset worker A, meaning it clear the state and reads kafka partitions from the latest offsets from delta.

mosyp:
It might be the case that this approach is not generally applicable, but it seems to work with kafka-delta-ingest and avoid duplicates on worker staling. If **and only if** you're reseting yours stale worker if the commit is failed (due to lock is acquired by the repairer)

mosyp:
@houqp Oh I see what you mean https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/s3/mod.rs#L80-L106

In the code we first call rename, then lock.release. And after that we handle the errors first from rename and later from release. But then if release if failed (due to expiration) its error will be ignored and then AlreadyExists from rename operation returned, hence worker A will increment the counter and keep trying to commit the same file. 

It seems like the possible solution for this will be the proper handling of errors from rename/release calls

rtyler (commented):
This looks exciting! Would you mind adding a README to the `proofs/` directory sharing how to run the proofs.  I'm excited for more people to play with this

houqp:
@mosyp based on the idea in https://github.com/delta-io/delta-rs/pull/540#issuecomment-1011667336,  I came up with a design to fix the bug. You can see how conflicts are now being validated in https://github.com/delta-io/delta-rs/pull/540/commits/535fa42fdf3dc583fbd8fc87452383b0a0e51b8a.

The new design now passes stateright's formal verification under 3 concurrent workers.

houqp:
@rtyler @mosyp this PR is now ready for review. I will send the actual implementation fix in a separate PR .

Issue #413, opened 2021-08-24
Ability to get table updates in a streaming manner


blogle:
# Description
I would like to read a stream of updates on a delta table, where the table is append only meaning no inserts/updates or deletes.

**Use Case**
Today we have both a custom built messaging infrastructure built on-top of zmq and Google pub/sub acting as a message broker. For each topic there is a single producer that does some processing and submits the message, 1 or more consumers subscribe to the stream of records, do some transformation (invoke an ML model, create a transaction, etc) and might subsequently themselves publish their output as a stream for downstream consumers. To support offline use cases like testing, ml training, or analytics we have to maintain separate tooling to ETL this data into various db's and warehouses. I would like to evaluate 
- can delta lake be leveraged as a durable message broker obviating our need of pub/sub and related etl jobs?
- can latency of reads be tuned such that delta lake is feasible for serving ml features in "soft" real-time, obviating our zmq tooling and related etl?

For additional context see the slack thread: https://delta-users.slack.com/archives/C013LCAEB98/p1629763325074200
**Related Issue(s)**

PR #550, opened 2022-01-26
Implement polling for table updates. (#413)

Blajda:
Implement various ways for clients to consume and update the Delta table's state. 
Clients have a choice of viewing actions related to a commit and then applying those actions to obtain the next version of the table. 
I used the general outline provided in #413 and I'm looking for feedback

Thanks

### DIFF
--- a/rust/src/table_state.rs
+++ b/rust/src/table_state.rs
@@ -15,7 +15,7 @@ use super::{
     ApplyLogError, CheckPoint, DeltaDataTypeLong, DeltaDataTypeVersion, DeltaTable,
     DeltaTableError, DeltaTableMetaData,
 };
-use crate::action;
+use crate::action::{self, Action};
 use crate::delta_config;
 
 /// State snapshot currently held by the Delta Table instance.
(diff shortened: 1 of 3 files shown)


rust/src/table_state.rs
```diff
@@ -54,6 +54,15 @@ impl DeltaTableState {
         Ok(new_state)
     }
 
+    /// Construct a delta table state object from a list of actions
+    pub fn from_actions(actions: Vec<Action>) -> Result<Self, ApplyLogError> {
```
houqp, line 58: We could inline this method and reuse it in the `from_commit` method.

```suggestion
    #[inline]
    pub fn from_actions(actions: Vec<Action>) -> Result<Self, ApplyLogError> {
```

houqp:
Very cool, thanks @Blajda ! There are some minor build formatting errors, otherwise looks good to me :+1:

houqp (approved):

Blajda:
Thanks @houqp for reviewing and the general outline.

houqp:
Thank you @Blajda for a clean implementation :)

PR #551, opened 2022-01-26
Fix VACUUM by using the URI of the DeltaTable when filtering the files to delete

### LINKED ISSUE #549: VACUUM operation failed with a S3 path
# Environment

**Delta-rs version**: `0.5.5`

**Binding**: Python/Rust

**Environment**:
- **Cloud provider**: AWS
- **OS**: MacOS11

***
# Bug

**What happened**:
```
>>> DeltaTable("s3://bucket/database/table").vacuum(dry_run=True)
  File "/usr/local/lib/python3.9/site-packages/deltalake/table.py", line 249, in vacuum
    return self._table.vacuum(dry_run, retention_hours)
deltalake.PyDeltaTableError: Generic DeltaTable error: Parent path `database/table` is not a prefix of path `s3://bucket/database/table/_delta_log/00000000000000010988.json`
```

**What you expected to happen**: a `VACUUM` dry_run command

**More details**:
The root cause is located [here](https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L390) when we check if the [object meta path is valid ](https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L1144) with the current DeltaTable path.

fvaleye:
# Description
- The `obj.meta` is the absolute path of the object on AWS S3 so we could use directly the `table_uri` for comparing the listed files with the valid files of the DeltaTable.

# Related Issue(s)
- closes #549 

# Tested:
- AWS S3
- Local

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -26,7 +26,7 @@ use super::action::{Action, DeltaOperation};
 use super::partitions::{DeltaTablePartition, PartitionFilter};
 use super::schema::*;
 use super::storage;
-use super::storage::{parse_uri, StorageBackend, StorageError, UriError};
+use super::storage::{StorageBackend, StorageError, UriError};
 use super::table_state::DeltaTableState;
 use crate::delta_config::DeltaConfigError;
 
@@ -1132,16 +1132,9 @@ impl DeltaTable {
         let mut files_to_delete = vec![];
         let mut all_files = self.storage.list_objs(&self.table_uri).await?;
 
-        // TODO: table_path is currently only used in vacuum, consider precalcualte it during table
-        // struct initialization if it ends up being used in other hot paths
-        let table_path = parse_uri(&self.table_uri)?.path();
-
         while let Some(obj_meta) = all_files.next().await {
             let obj_meta = obj_meta?;
-            // We can't use self.table_uri as the prefix to extract relative path because
-            // obj_meta.path is not a URI. For example, for S3 objects, obj_meta.path is just the
-            // object key without `s3://` and bucket name.
-            let rel_path = extract_rel_path(&table_path, &obj_meta.path)?;
+            let rel_path = extract_rel_path(&self.table_uri, &obj_meta.path)?;
 
             if valid_files.contains(rel_path) // file is still being tracked in table
                 || !expired_tombstones.contains(rel_path) // file is not an expired tombstone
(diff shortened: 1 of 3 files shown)


houqp (approved):
LGTM, I think this is likely going to break for filesystem backend since it supports two different schemes: `table/path` and `file://table/path`. I think we should either remove the `file://` scheme so we have one to one mapping between backends and schemes. Or we should change the behavior of list_objs to return path without schemes. The list_obj behavior was changed in https://github.com/delta-io/delta-rs/pull/518, perhaps a better way to fix that problem is change head_obj call to return a path without scheme.

houqp:
also cc @mosyp

fvaleye:
Thanks for your review @houqp 🙏 

I will not merge the PR and wait for @mosyp review. I don't want to break anything with the different backend storage implementations.

rust/src/delta.rs
```diff
@@ -1132,16 +1132,9 @@ impl DeltaTable {
...
-        // struct initialization if it ends up being used in other hot paths
-        let table_path = parse_uri(&self.table_uri)?.path();
-
         while let Some(obj_meta) = all_files.next().await {
             let obj_meta = obj_meta?;
-            // We can't use self.table_uri as the prefix to extract relative path because
-            // obj_meta.path is not a URI. For example, for S3 objects, obj_meta.path is just the
-            // object key without `s3://` and bucket name.
-            let rel_path = extract_rel_path(&table_path, &obj_meta.path)?;
+            let rel_path = extract_rel_path(&self.table_uri, &obj_meta.path)?;
```
mosyp, line 1137: Ah, I got it, this comment is obsolete now with #518 changes

mosyp (approved):
LGTM, I don't expect this to impact kafka-delta-ingest as we don't use it there, maybe that's why it wasn't caught with #518

fvaleye:
Thanks again @houqp and @mosyp 

I merge it as it's ok and keep an eye out if any issues come up.

houqp:
If we go with adding scheme everywhere, then we should definitely remove this line: https://github.com/delta-io/delta-rs/blob/2fa81c611e74235c13e5d27813d2e9d815d10ad9/rust/src/storage/mod.rs#L233

Otherwise vacuum will not be working for table uris like `file://...`

PR #557, opened 2022-02-03
move delta-inspect to its own crate

### LINKED ISSUE #402: Move delta-inspect to its own crate
# Description

**Use Case**

Currently delta-rs Rust core depends on cli related dependencies like `clap`, which is unnecessary for downstream consumers. We should move delta-inspect code into its own crate within the same workspace like what we did for `aws/delta-checkpoint` so we can move all cli related dependencies to the cli crate and keep delta-rs dependency tree clean.

**Related Issue(s)**

https://github.com/delta-io/delta-rs/issues/401

houqp:
# Description

Remove CLI related deps from the rust core. Also change vacuum method to take a non-mut reference.

# Related Issue(s)

fix #402

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ members = [
     "aws/delta-checkpoint",
     "dynamodb_lock",
 ]
-exclude = ["proofs"]
+exclude = ["proofs", "delta-inspect"]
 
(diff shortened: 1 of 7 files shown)


fvaleye (approved):
LGTM!

PR #560, opened 2022-02-16
Update ADLSGen2-HOWTO.md

dgcaron:
# Description
I had a hard time figuring out how to connect to a delta table that is stored in ADLS Gen2 and only found a way by digging into the source code. I would like to save other people the same trouble by adding this to the docs.

# Related Issue(s)

# Documentation

### DIFF
--- a/docs/ADLSGen2-HOWTO.md
+++ b/docs/ADLSGen2-HOWTO.md
@@ -30,3 +30,18 @@ Docs: [Create a storage account for Azure Data Lake Storage Gen2 | Microsoft Doc
 * [Blob service REST API - Azure Storage | Microsoft Docs](https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-rest-api)
 * [Azure Data Lake Storage Gen2 REST API reference - Azure Storage | Microsoft Docs](https://docs.microsoft.com/en-us/rest/api/storageservices/data-lake-storage-gen2)
 
+## Example
+
+When you want to connect to a deltatable stored in ADLS Gen2, you need to set two environment variables. 
+* AZURE_STORAGE_ACCOUNT_NAME which holds the name of the storage account 
+* AZURE_STORAGE_ACCOUNT_KEY which contains the root key for the storage account. 
+
+The url for the table should follow a specific format adls2://{accountname}/{filesystem}/{path to table}. 
+
+example:
+```python
+  from deltalake import DeltaTable
+
+  delta = DeltaTable("adls2://<accountname>/<filesystem>/<path to table>")
+  dataFrames = delta.to_pyarrow_table().to_pandas()
+```
--- a/python/docs/source/usage.rst
+++ b/python/docs/source/usage.rst
@@ -56,7 +56,7 @@ Besides local filesystems, the following backends are supported:
 
 * AWS S3, detected by the prefix ``s3://``. AWS credentials can be specified using
   environment variables in the same way as the CLI.
-* Azure Data Lake Storage Gen 2, detected by the prefix ``abffs://``. Note that
+* Azure Data Lake Storage Gen 2, detected by the prefix ``adls2://``. Note that
   `specific instructions`_ must be followed to setup the Azure Storage Account.
 * Google Cloud Storage, detected by the prefix ``gs://``.
 


docs/ADLSGen2-HOWTO.md
```diff
@@ -30,3 +30,18 @@ Docs: [Create a storage account for Azure Data Lake Storage Gen2 | Microsoft Doc
...
+
+The url for the table should follow a specific format adls2://{accountname}/{filesystem}/{path to table}. 
+
+example:
+```python
+  from deltalake import DeltaTable
+
+  delta = DeltaTable("adls2://<accountname>/<filesystem>/<path to table>")
+  dataFrames = delta.to_pyarrow_table().to_pandas()
+```
```
wjones127, line 47: Thanks! Would you like to also add this to the [usage docs](https://delta-io.github.io/delta-rs/python/usage.html#loading-a-delta-table) as well?

It looks like I might have gotten the wrong prefix when I add this :facepalm::

https://github.com/delta-io/delta-rs/blame/main/python/docs/source/usage.rst#L59-L60

thovoll (dismissed):

docs/ADLSGen2-HOWTO.md
zeevm: Why use this non-standard uri schema for ADLS?
Should either use abfss if the underlying storage layer "knows" adls or use https with the appropriate blob storage endpoint that underlies every adls account.

wjones127: Actually, googling a bit, it seems like `abfss` is somewhat standard: https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri
Thanks @zeevm 

@dgcaron Does that schema work for you?

thovoll: `abfs[s]` refers to the Hadoop File System driver (see the docs linked by @wjones127), which is not being used here.

Using `https` would be inconsistent with the fact that delta-rs accepts a URI with a cloud specific scheme ("s3://" and "gs://") for the Amazon and Google backends.

thovoll: > Why use this non-standard uri schema for ADLS? Should either use abfss if the underlying storage layer "knows" adls or use https with the appropriate blob storage endpoint that underlies every adls account.

delta-rs relies on the atomic rename functionality of ADLS Gen2 and cannot work with "plain" Blob Storage.

dgcaron:
Abfss was what I tried initially, unsuspecting users that will not read the manual will most likely do the same.  After failing and not finding the right scheme I went into the code to see what it expects. To save anyone else the search I thought it was a good thing to add it to the docs.

houqp (approved):

houqp:
@dgcaron do you want to include the python usage docs update in this PR as well?

dgcaron:
Yes sure, I'll do that today

zeevm:
I'd think fixing the docs isn't the issue, the implementation should be fixed to use either abfss if delta-rs knows how to talk to the adls endpoint or https if it talks to the blob storage endpoint, 'adls2' is never the right scheme to use

thovoll:
> I'd think fixing the docs isn't the issue, the implementation should be fixed to use either abfss if delta-rs knows how to talk to the adls endpoint or https if it talks to the blob storage endpoint, 'adls2' is never the right scheme to use

Hi @zeevm, I addressed all these points above, stated again here with some more elaboration:

1. `abfs[s]://` refers to the Hadoop File System driver, which is not being used by delta-rs. From [the Azure docs](https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri): "The Hadoop Filesystem driver that is compatible with Azure Data Lake Storage Gen2 _is known by its scheme identifier abfs_ (Azure Blob File System). Consistent with other Hadoop Filesystem drivers, the ABFS driver employs a URI format to address files and directories within a Data Lake Storage Gen2 capable account."

1. Using `https` URLs would be inconsistent with the fact that delta-rs accepts URIs with a cloud specific scheme (`s3://` and `gs://`) for the Amazon and Google backends. 

Based on those two points, it makes sense to use a cloud specific URI scheme in delta-rs, but not `abfs[s]://` - hence: `adls2://`. 

Also: the delta-rs Azure backend relies on the atomic rename functionality of Azure Data Lake Storage (ADLS) Gen2 and doesn't work with Azure Blob Storage.

The `adls2://` URI scheme is an invention of delta-rs, but there are other libraries that have done something similar for Azure storage. The root cause of this need is that ADLS Gen2 doesn't define a URI scheme other than in the context of Hadoop. It's worth noting that the [S3 Hadoop File System driver](https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/index.html#Introducing_the_Hadoop_S3A_client.) doesn't use `s3://` but `s3a://` - the reasons are complicated but this supports introducing `adls2://` to distinguish from `abfs[s]://`.

wjones127 (approved):

dgcaron:
did i break the build here?

thovoll:
> did i break the build here?

Clippy errors can show up because of toolchain updates. Take a look at the error, it has a suggestion to try and fix the clippy error.

dgcaron:
yeah, i'll wait for your merge as it is a linter error for files I didn't touch at all.

thovoll:
> yeah, i'll wait for your merge as it is a linter error for files I didn't touch at all.

Actually, @houqp already fixed these yesterday. Just fetch upstream on your branch.

dgcaron:
do i need to do something additional to get this merged?

thovoll:
> do i need to do something additional to get this merged?

No, someone with merge privileges has to review, approve, and merge.

houqp:
Sorry, there were some intermittent CI errors, I was waiting for the rerun of the CI before doing a merge. Thanks again @dgcaron for the fix! And than kyou @thovoll and @wjones127 for the review :)

PR #572, opened 2022-03-18
[Python] Test in minimal and latest Python environments

wjones127:
# Description

In #566 we ran into an issue where changes meant docs couldn't build with Python older than 3.8. We should simply build them in a newer version of Python.

In addition, we should be testing the Python package with both the minimal PyArrow version and the latest PyArrow version, since they are several major versions apart. (PyArrow right now just releases new major versions quarterly, so breaking API changes are common in experimental parts of the package.)

# Related Issue(s)

None

# Documentation


houqp (approved):

houqp:
Thank you @wjones127 !

PR #573, opened 2022-03-22
Record Batch Writer

roeap:
# Description

This PR is part of the larger PR #523. Specifically an implementation of a record batch writer. The implementation is a composite of the json implementations from this crate and `kafka-delta-ingest`. The main addition is some logic to split a record batch into partitions according to the table partitioning.

# Related Issue(s)
 towards: #509

# Documentation


rust/src/write/writer.rs
```diff
@@ -0,0 +1,606 @@
...
+                    let _ = self.arrow_writers.insert(partition_key, writer);
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Writes the existing parquet bytes to storage and resets internal state to handle another file.
+    pub async fn flush(&mut self) -> Result<Vec<Add>, DeltaWriterError> {
```
houqp, line 125: the semantic of flush here is different from the one in the existing json writer, is there any particular reason why you to keep it different?

houqp (commented):
Thanks @roeap for splitting up your PR into smaller ones :)

Looks good to me overall, one question I have is we already have a json writer in the `deltalake::writer` module, have you thought of how to better organize these two writer implementations in a more consistent namespace? I can see users getting confused between `deltalake::writer` and `deltalake::write::writer` modules.

rust/src/write/writer.rs
roeap: Not 100% sure which part you are referring to, but there are some things i considered :). The implementation with the json writer already creates a commit. Since we are hoping to eventually use this writer in a distributed context i though we should avoid doing so here. When its referring to the error type, these should probably be harmonized. In general it may be worth exploring to have a shared writer trait that different writers could implement?

rust/src/write/writer.rs
```diff
@@ -0,0 +1,606 @@
...
+    cursor.write_all(bytes)?;
+    Ok(cursor)
+}
+
+fn new_underlying_writer(
+    cursor: InMemoryWriteableCursor,
+    arrow_schema: Arc<ArrowSchema>,
+    writer_properties: WriterProperties,
+) -> Result<ArrowWriter<InMemoryWriteableCursor>, ParquetError> {
+    ArrowWriter::try_new(cursor, arrow_schema, Some(writer_properties))
```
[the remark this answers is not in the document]
roeap: true - will get rid of it.

roeap:
@houqp - I did spend some time thinking on how to consolidate these namespaces, since having two namespaces with essentially the same name seems like a bad idea. Not only the namespace, but there is also a bunch of logic shared between the two writer implementations. In order to consolidate this we could maybe move the json writer and record batch writer into the same namespace and expose the same API via trait / generic. My guess would be, that the most prominent change would be to use the partitioning logic from the RB writer also in the json case. I was a bit hesitant to do this directly for being unsure about the performance implications. For larger data sizes, the sorting operations used in the RB writer could be fairly expensive, while the partitioning in the json writer seems quite clean and straight froward to me.

In any case, I am more then happy to try and consolidate these within this PR based on your suggestions.

rust/src/write/writer.rs
houqp: Yes, I was referring to the flush method from the json writer: https://github.com/delta-io/delta-rs/blob/346f51a277c23be09eca373cd6d775b745d59c05/rust/src/writer.rs#L94.

> Since we are hoping to eventually use this writer in a distributed context i though we should avoid doing so here.

I think this makes sense, if so, then it's better to use a different name other than flush to avoid confusions.

> In general it may be worth exploring to have a shared writer trait that different writers could implement?

I haven't put much thought into this yet, maybe think about what are the common interfaces we want to share between writers? From a quick glance, the writer method might not be a good fit because it takes different type of input depending on the input format that needs to be supported.

houqp:
>  Not only the namespace, but there is also a bunch of logic shared between the two writer implementations. In order to consolidate this we could maybe move the json writer and record batch writer into the same namespace and expose the same API via trait / generic. 

I agree this would be the ideal approach :+1: 

> My guess would be, that the most prominent change would be to use the partitioning logic from the RB writer also in the json case. I was a bit hesitant to do this directly for being unsure about the performance implications.

I recommend we leave the duplication and optimization work as follow ups so we can focus on coming up with the right abstraction and interface in this PR. Sometimes a little bit of code duplication is not a bad thing :)

> Maybe one way to have both functionalities is to add an additional method flush_and_commit to both writers?

That's a good idea to keep both writers consistent too. And we can leave the `flush` method to only write the parquet files and return the actions. Curious if @mosyp and @xianwill have any option on this.

roeap:
@houqp - When trying to consolidate the writers a bit, I stumbled across another I believe significant difference in how the writers are designed. The json writer uses the `add_file` method on the `DeltaTransaction` to write files. The json writer from `kafka-delta-ingest` as well the `RecordBatchWriter` make use of the storage backend directly. The `RecordBatchWriter` is generally modelled more against the the json implementation from the kafka crate. To make the writer APIs more harmonized i think the writers doing the writing the the only feasible alternative, since in the envisioned usage within datafusion having the transaction available in all writers seems quite cumbersome, if not prohibitive.

As such I am unsure if we would want to keep the `add_file` method on the transaction implementation - to me this might get a bit confusing. Additionally the `add_file` method does not seem to compute file statistics, I pulled that logic as well from the kafka crate... Just wanted to make sure before going ahead and removing it.

houqp:
@roeap I think what you proposed makes total sense :+1:  The add_file abstraction was probably added when we don't have a good writer abstraction. It's a good idea to fully decouple data file IO and table log metadata IO.

roeap:
After looking a bit into harmonizing the implementations, it turned out that the required adjustments ran much deeper than hoped. So I went for an easier path and migrated the Json implementation from kafka-delta-ingest. Since the record batch implementation was also modelled against that implementation they were already somewhat similar. Still some cleaning up to do, but hope this is going the right way.

houqp (approved):
LGTM, thanks @roeap !

houqp:
leaving it open for the weekend in case others want to chime in :)

rust/src/writer/record_batch.rs
```diff
@@ -0,0 +1,505 @@
+//! Main writer API to write record batches to delta table
```
wjones127, line 1: Would be nice to have a little more description and at least one simple usage example. Here's a suggestion:

```suggestion
//! Main writer API to write record batches to Delta Table
//!
//! Writes Arrow record batches to a Delta Table, handling partitioning and file statistics.
//! Each Parquet file is buffered in-memory and only written once `flush()` is called on
//! the writer. Once written, add actions are returned by the writer. It's the users responsibility
//! to create the transaction using those actions.
//!
//! # Examples
//! 
//! Write to an existing Delta Lake table:
//! ```rust
//! let table = DeltaTable::try_from_uri("../path/to/table")
//! let batch: RecordBatch = ...
//! let mut writer = RecordBatchWriter::for_table(table, /*storage_options=*/ HashMap::new())
//! writer.write(batch)?;
//! let actions: Vec<action::Action> = writer.flush()?.iter()
//!     .map(|add| Action::add(add.into()))
//!     .collect();
//! let mut transaction = table.create_transaction(Some(DeltaTransactionOptions::new(/*max_retry_attempts=*/3)));
//! transaction.add_actions(actions);
//! async {
//!     transaction.commit(Some(DeltaOperation::Write {
//!         SaveMode::Append,
//!         partitionBy: Some(table.get_metadata().partition_columns),
//!         predicate: None,
//!     }))
//! }
//! ```
```

rust/src/writer/stats.rs
```diff
@@ -0,0 +1,733 @@
...
+
+    for arr in byte_arrays.iter() {
+        buffer.extend_from_slice(arr);
+    }
+
+    let builder = ArrayData::builder(data_type)
+        .len(byte_arrays.len())
+        .add_buffer(buffer.into());
+
+    // TODO remove panic
```
wjones127, line 397: Can we address this?

wjones127 (commented):
This looks like great progress towards a rust-based writer!

I noticed a few TODOs that look like something we might want to address before merging. I think it might also be worth adding some documentation to the writer as is.

rust/src/writer/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+};
+use async_trait::async_trait;
+pub use json::JsonWriter;
+use parquet::{basic::LogicalType, errors::ParquetError};
+pub use record_batch::RecordBatchWriter;
+use serde_json::Value;
+use std::convert::TryFrom;
+use std::sync::Arc;
+
+impl TryFrom<Arc<ArrowSchema>> for Schema {
```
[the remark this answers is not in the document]
roeap: makes sense. Moved to the arrow file, and simplified it using existing impl.

rust/src/writer/record_batch.rs
```diff
@@ -0,0 +1,505 @@
...
+            writer_properties,
+            partition_columns: partition_columns.unwrap_or_default(),
+            arrow_writers: HashMap::new(),
+        })
+    }
+
+    /// Creates a DeltaWriter to write to the given table
+    pub fn for_table(
+        table: &DeltaTable,
+        options: HashMap<String, String>,
```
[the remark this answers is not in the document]
roeap: updates based on your suggestion.

rust/src/writer/record_batch.rs
```diff
@@ -0,0 +1,505 @@
...
+    }
+}
+
+/// Partition a RecordBatch along partition columns
+pub fn divide_by_partition_values(
+    arrow_schema: ArrowSchemaRef,
+    partition_columns: Vec<String>,
+    values: &RecordBatch,
+) -> Result<Vec<PartitionResult>, DeltaWriterError> {
+    // TODO remove panics within closures
```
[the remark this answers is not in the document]
roeap: got rid of all unwraps..

rust/src/writer/record_batch.rs
```diff
@@ -0,0 +1,505 @@
...
+                let new_cursor = cursor_from_bytes(current_cursor_bytes.as_slice())?;
+                let _ = std::mem::replace(&mut self.cursor, new_cursor.clone());
+                let arrow_writer = ArrowWriter::try_new(
+                    new_cursor,
+                    self.arrow_schema.clone(),
+                    Some(self.writer_properties.clone()),
+                )?;
+                let _ = std::mem::replace(&mut self.arrow_writer, arrow_writer);
+                // TODO we used to clear partition values here, but since we pre-partition the
+                // record batches, we should never try with mismatching partition values.
```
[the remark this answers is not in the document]
roeap: no - removed it.

rust/src/writer/stats.rs
roeap: yes, and did :)

wjones127 (approved):

rust/Cargo.toml
```diff
@@ -71,6 +71,7 @@ async-trait = "0.1"
 # rust-dataframe = {version = "0.*", optional = true }
 
 [features]
+# default = ["datafusion-ext"]
```
houqp, line 74: one minor ask, i think this should be removed?

houqp (approved):
LGTM with one minor nitpick

rust/Cargo.toml
roeap: done

houqp (approved):

houqp:
Thanks @roeap ! This is one big milestone :D

PR #581, opened 2022-03-27
Expose read and write options in public API

### LINKED ISSUE #578: Large timestamps (9999-12-31) not read correctly
# Environment

**Delta-rs version**: 0.5.6

**Binding**: Python (deltalake 0.5.6)

**Environment**:
- **Cloud provider**: Azure
- **OS**: Ubuntu 20.04
- **Other**: Python 3.9 (poetry venv)

***
# Bug

**What happened**:

The following code fails when trying to read a delta table that has TIMESTAMP column with `9999-12-31 00:00:00` values instead of nulls:
```
    delta_table = DeltaTable(path.to_delta_rs_path(), version=version)

    if batch_size:
        batches: Iterator[RecordBatch] = delta_table.to_pyarrow_dataset().to_batches(filter=row_filter, columns=columns, batch_size=batch_size)

        return map(lambda batch: batch.to_pandas(), batches)

    pyarrow_table: Table = delta_table.to_pyarrow_dataset().to_table(filter=row_filter, columns=columns)

    result = pyarrow_table.to_pandas()
```

Trace:

```
  File "/mnt/c/users/gzu/source/repos/github/proteus/proteus/storage/delta_lake/_functions.py", line 83, in load
    pyarrow_table: Table = delta_table.to_pyarrow_dataset().to_table(
  File "pyarrow/_dataset.pyx", line 323, in pyarrow._dataset.Dataset.to_table
  File "pyarrow/_dataset.pyx", line 2311, in pyarrow._dataset.Scanner.to_table
  File "pyarrow/error.pxi", line 143, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 99, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Casting from timestamp[ns] to timestamp[us] would lose data: -4852202631933722624
```

If `columns` doesn't contain a column that has `9999-12-31 00:00:00` timestamp values, everything works fine.

**What you expected to happen**: `result` contains pandas DataFrame and doesn't throw.

**How to reproduce it**:

Create a delta table with a timestamp column that contains `9999-12-31 00:00:00` values (from Spark, so it is stored as INT96).

**More details**:
I was able to fix this issue by changing `to_pyarrow_dataset` method:

```
def _to_pyarrow_dataset(
        table: RawDeltaTable,
        schema: pyarrow.Schema,
) -> pyarrow.dataset.Dataset:
    """
    Build a PyArrow Dataset using data from the DeltaTable.

    :param partitions: A list of partition filters, see help(DeltaTable.files_by_partitions) for filter syntax
    :param filesystem: A concrete implementation of the Pyarrow FileSystem or a fsspec-compatible interface. If None, the first file path will be used to determine the right FileSystem
    :return: the PyArrow dataset in PyArrow
    """
    if not filesystem:
        filesystem = pa_fs.PyFileSystem(
            DeltaStorageHandler(table.table_uri())
        )

    format = ParquetFileFormat(read_options=ParquetReadOptions(coerce_int96_timestamp_unit='ms'))

    fragments = [
        format.make_fragment(
            file,
            filesystem=filesystem,
            partition_expression=part_expression,
        )
        for file, part_expression in table.dataset_partitions(partitions)
    ]

    return FileSystemDataset(fragments, schema, format, filesystem)
```

Specifically, doing `format = ParquetFileFormat(read_options=ParquetReadOptions(coerce_int96_timestamp_unit='ms'))` instead of source code having `format = ParquetFileFormat()`

Then the pyarrow conversion will go through, but pandas conversion will fail. It is fixed via `pyarrow_table.to_pandas(timestamp_as_object=True)` instead of `pyarrow_table.to_pandas()`

## TLDR

Parquet read options are not exposed in public API. Would be nice to have them there, so issues like these can be resolved automatically :)

george-zubrienko:
# Description
Exposes read options for Parquet in `to_pyarrow_dataset` method, exposes file format-specific write options in `write_deltalake` method, and adds options for controlling file sizes produced when writing.

# Related Issue(s)
Closes #578 

# Documentation
- [Write options](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html) for PyArrow dataset
- [ParquetFileFormat](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.ParquetFileFormat.html#pyarrow.dataset.ParquetFileFormat.make_write_options)

# NB

Unit tests pending.


python/deltalake/writer.py
```diff
@@ -44,6 +44,10 @@ def write_deltalake(
     partition_by: Optional[List[str]] = None,
     filesystem: Optional[pa_fs.FileSystem] = None,
     mode: Literal["error", "append", "overwrite", "ignore"] = "error",
+    file_options: Optional[ds.FileWriteOptions] = None,
```
wjones127, line 47: I think we can make this parquet-specific, since we only support parquet:

```suggestion
    file_options: Optional[ds.ParquetFileWriteOptions] = None,
```

wjones127:
Thanks @george-zubrienko , your approach makes sense! Go ahead and finish up with the unit tests and I can review again.

george-zubrienko:
I've added unit tests for **large** timestamp case, and for writer options as well. Sadly I can't find anything in pyarrow docs re what kind of kwargs can be provided to `make_writer_options`, so just made a default call there. 

Re reads, this doesn't work for `0001-01-01` timestamps. Delta table I've added comes from Databricks 9.1 LTS workspace (Spark 3.1.2), which can write/read this fine, but when reading with delta-rs, pyarrow fails with overflow error. I've decided to leave this for a separate PR, if there is a solution for this problem.

wjones127:
> Sadly I can't find anything in pyarrow docs re what kind of kwargs can be provided to make_writer_options, so just made a default call there.

I reported that upstream in [ARROW-16130](https://issues.apache.org/jira/browse/ARROW-16130). FWIW the options available [are documented](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html#pyarrow.parquet.write_table) as the arguments of `pyarrow.parquet.write_table`. Maybe a good option to test is setting `flavor='spark'` and checking that you can roundtrip these timestamps?

george-zubrienko:
> > Sadly I can't find anything in pyarrow docs re what kind of kwargs can be provided to make_writer_options, so just made a default call there.
> 
> I reported that upstream in [ARROW-16130](https://issues.apache.org/jira/browse/ARROW-16130). FWIW the options available [are documented](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html#pyarrow.parquet.write_table) as the arguments of `pyarrow.parquet.write_table`. Maybe a good option to test is setting `flavor='spark'` and checking that you can roundtrip these timestamps?

Nice, will look into using this on weekend, thank you :)

I see there are issues with Py3.10/Pyarrow latest, will also look into options of dealing with that issue.

george-zubrienko:
Hi again, sorry got distracted a bit by the Easter :)

Re write options, it seems these are not expected:

```
>       write_deltalake(path, data, file_options=ParquetFileFormat().make_write_options(flavor='spark'))

tests/test_writer.py:340:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pyarrow/_dataset_parquet.pyx:181: in pyarrow._dataset_parquet.ParquetFileFormat.make_write_options
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: unexpected parquet write option: flavor
```

This approach gives the same error:
```
    opts = ParquetFileFormat().make_write_options().update(
        flavor='spark'
    )
```
 Apparently these are the default options:

```
use_dictionary
compression
version
write_statistics
data_page_size
compression_level
use_byte_stream_split
column_encoding
use_deprecated_int96_timestamps
coerce_timestamp
allow_truncated_timestamps
use_compliant_nested_type
```
From [here](https://github.com/apache/arrow/blob/master/python/pyarrow/_dataset_parquet.pyx#L492-L533)

I've added a test case with `compression` and `coerce_timestamp` to check that we can write-read back data correctly. Seems to be working fine. Will push a commit soon once I figure out Py3.10 issue

george-zubrienko:
Also a bit of additional information regarding edge case timestamps on the other side of a spectrum, like `0001-01-01`. Apparently python has issues with `timestamp` and with the related one `fromtimestamp` functions when working with small dates. Spark has no issues with those.
```
Python 3.8.10 (default, Mar 15 2022, 12:22:08)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> datetime.datetime(1,1,1)
NameError: name 'datetime' is not defined
>>> import datetime
>>> datetime.datetime(1,1,1)
datetime.datetime(1, 1, 1, 0, 0)
>>> datetime.datetime(1,1,1).timestamp()
ValueError: year 0 is out of range
```
Maybe we should add some sort of a note to docs until this is resolved one way or another in our code?

python/tests/test_writer.py
```diff
@@ -310,3 +322,51 @@ def test_writer_fails_on_protocol(existing_table: DeltaTable, sample_data: pa.Ta
...
+    assert sum([stat_entry['numRecords'] for stat_entry in stats]) == 1000
+
+
+def test_writer_with_options(tmp_path: pathlib.Path):
+    expected_column_values = [int(datetime(year_, 1, 1, 0, 0, 0).timestamp() * 1e6) for year_ in range(9000, 9010)]
+    data = pa.table(
+        {
+            "colA": pa.array(expected_column_values, pa.timestamp('us'))
+        }
+    )
```
wjones127, line 359: You should be able to pass in the raw datetime values into the array constructor:
```suggestion
    expected_column_values = pa.array(
        [datetime(year_, 1, 1, 0, 0, 0) for year_ in range(9000, 9010)], pa.timestamp('us'))
    expected = pa.table({"colA": expected_column_values})
```

python/tests/test_writer.py
```diff
@@ -310,3 +322,51 @@ def test_writer_fails_on_protocol(existing_table: DeltaTable, sample_data: pa.Ta
...
+        coerce_timestamps='us'
+    )
+    write_deltalake(path, data, file_options=opts)
+
+    table = DeltaTable(path).to_pyarrow_dataset(parquet_read_options=ParquetReadOptions(coerce_int96_timestamp_unit='us')).to_table().to_pydict()['colA']
+
+    # since either session (Spark) or system timezone is used when reading timestamps back into python
+    # a conversion is required for this synthetic case. Should not be an issue with a real table, as long as a system/Spark session timestamp matches the one used when data was written
+
+    assert [int(d.timestamp() * 1e6 - time.timezone * 1e6) for d in table] == expected_column_values
```
wjones127, line 372: Above, if you just let table be a `pa.Table`, you could simply make this assertion `assert table == expected` (or `data`).

wjones127 (changes requested):
Sorry I didn't get a chance to look at earlier. I have a few more suggestions before we are finished.

Also, if you aren't using them yet, the `make format` and `make check-python` commands are helpful to avoid linting errors.

python/deltalake/writer.py
```diff
@@ -144,6 +160,10 @@ def visitor(written_file: Any) -> None:
         schema=schema if not isinstance(data, RecordBatchReader) else None,
         file_visitor=visitor,
         existing_data_behavior="overwrite_or_ignore",
+        file_options=file_options,
+        max_rows_per_file=max_rows_per_file or 0,
+        min_rows_per_group=min_rows_per_group or 0,
+        max_rows_per_group=max_rows_per_group or 0
```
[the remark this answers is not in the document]
george-zubrienko: Added this option, should we test this or meh?

python/deltalake/writer.py
```diff
@@ -64,6 +68,18 @@ def write_deltalake(
         already exists. If 'append', will add new data. If 'overwrite', will
         replace table with new data. If 'ignore', will not write anything if
         table already exists.
+    :param file_options: Optional write options for the target data format.
+        Can be provided via make_write_options method for the target file format
```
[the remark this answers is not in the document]
george-zubrienko: Updated docstring

python/tests/test_writer.py
```diff
@@ -310,3 +322,51 @@ def test_writer_fails_on_protocol(existing_table: DeltaTable, sample_data: pa.Ta
...
+            "colB": pa.array([i * random.random() for i in range(0, 1000)], pa.float64())
+        }
+    )
+    path = str(tmp_path)
+    write_deltalake(path, data, file_options=ParquetFileFormat().make_write_options(), max_rows_per_file=100, max_rows_per_group=100)
+
+    table = DeltaTable(path)
+    stats = get_multifile_stats(table)
+
+    assert sum([stat_entry['numRecords'] for stat_entry in stats]) == 1000
```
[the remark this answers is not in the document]
george-zubrienko: Added `and` to this assertion as you suggested, and also parametized this test

python/tests/test_writer.py
george-zubrienko: applied, works nicely :)

python/tests/test_writer.py
george-zubrienko: applied, mourn all the time i spent figuring out timezone issue :)

george-zubrienko:
Linter seemed happy this time!
```
$ /mnt/c/users/gzu/source/repos/github/delta-rs-snd/python$ make format
--- Rust format ---
--- Python format ---
cargo fmt
black .
All done! ✨ 🍰 ✨
2 files left unchanged.
isort .
Skipped 1 files
```

george-zubrienko:
I see lint and documentation build failed again - tbh I don't understand why it don't get any reformats when running `make check-python` locally.
The documentation failure I don't get it either - it complains about import which clearly is valid?

`pyarrow._dataset_parquet.ParquetFileWriteOptions`

python/deltalake/writer.py
```diff
@@ -152,6 +176,11 @@ def visitor(written_file: Any) -> None:
         schema=schema if not isinstance(data, RecordBatchReader) else None,
         file_visitor=visitor,
         existing_data_behavior="overwrite_or_ignore",
+        file_options=file_options,
+        max_open_files=max_open_files or 1024,
+        max_rows_per_file=max_rows_per_file or 0,
+        min_rows_per_group=min_rows_per_group or 0,
+        max_rows_per_group=max_rows_per_group or 1024 * 1024,  # pyarrow default
```
houqp, line 183: curious why not set the default values through method argument defaults?

python/deltalake/writer.py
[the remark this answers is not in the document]
wjones127: I don't think it's necessary to test the behavior of those parameters; since it's well-tested upstream.

python/tests/test_writer.py
```diff
@@ -310,3 +321,73 @@ def test_writer_fails_on_protocol(existing_table: DeltaTable, sample_data: pa.Ta
...
+    )
+
+    table = DeltaTable(path)
+    stats = get_multifile_stats(table)
+    files_written = [f for f in os.listdir(path) if f != "_delta_log"]
+
+    assert (
+        sum([stat_entry["numRecords"] for stat_entry in stats]) == row_count
+        and len(files_written) == expected_files
+    )
```
wjones127, line 371: nit: I think these should be separate assert statements

```suggestion
    assert sum([stat_entry["numRecords"] for stat_entry in stats]) == row_count
    assert len(files_written) == expected_files
```

wjones127 (commented):
To fix the docs issue, you'll need to add this in the `nitpick_ignore` variable in `python/docs/source/conf.py`:
```python
    ("py:class", "pyarrow._dataset_parquet.ParquetFileWriteOptions"),
```

python/deltalake/writer.py
george-zubrienko: yeah good point, will update

wjones127:
Thank you @george-zubrienko !

Issue #528, opened 2021-12-26
Support file index


houqp:
# Description

See reference implementation at https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala.

tasks:

- [ ] read file index
- [ ] write file index

roeap:
With kernel the whole replay architecture changed considerably and different kinds of optimisations are being discussed. @houqp - I hope it is fine I close this for now :).

PR #584, opened 2022-04-04
High level Delta Operations with Datafusion

roeap:
# Description

This PR contains the next batch of changes from #523. Main purpose is to see how we can integrate datafusion `ExecutionPlan`s with high level operations against delta tables. So far only two operations are included - create and write - where write only supports append and overwrite mode. So nothing really interesting happening yet in terms of needing an SQL engine. However I was hoping we can use this PR to figure out how to handle transactions. Right now this is handled via a `DeltaTransactionPlan`, that wraps the actual operations plans. The operation plans would communicate their `Actions` to the transaction plant, which would do the ultimate commit to the Delta table.

There are some notable changes to the `DeltaOperations` as well. Specifically adding a create operation and generating commit info from the operation variants directly. On the transaction I added the possibility to include app specific metadata to be added to a commit. (@wjones127 do I remember correctly that you had that need as well?)

@houqp @wjones127 - any feedback is greatly appreciated.

# Related Issue(s)

towards:  #509

# Documentation


houqp:
Thank you @roeap , I will go through this PR in details this weekend. I also need to reread some of the discussions in #523 to refresh my thinking on how transaction handling logic :)

rust/src/operations/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub async fn create(
+        &mut self,
+        metadata: DeltaTableMetaData,
+        mode: Option<SaveMode>,
+    ) -> DeltaCommandResult<()> {
+        let operation = DeltaOperation::Create {
+            mode: mode.clone().unwrap_or(SaveMode::Ignore),
+            metadata: metadata.clone(),
+            location: self.table.table_uri.clone(),
+            // TODO get the protocol from somewhere central
```
wjones127, line 140: Yeah if I understand correctly, the protocol should be determined based on the existing protocol version (if any) and the latest metadata (which should contain all the information to determine the required protocol for the requested table schema, config, etc.).

rust/src/operations/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+        if data.is_empty() {
+            return Ok(());
+        }
+        let schema = data[0].schema();
+        let current_part = match self.table.update().await {
+            Ok(_) => {
+                let metadata = self.table.get_metadata()?;
+                if let Some(cols) = partition_columns {
+                    if cols != metadata.partition_columns {
+                        todo!("Schema updates not yet implemented")
```
wjones127, line 170: What do you think of returning a `DeltaCommandError::UnsupportedCommand` here instead?

rust/src/operations/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub async fn write(
+        &mut self,
+        data: Vec<RecordBatch>,
+        mode: Option<SaveMode>,
+        partition_columns: Option<Vec<String>>,
+    ) -> DeltaCommandResult<()> {
+        if data.is_empty() {
+            return Ok(());
+        }
+        let schema = data[0].schema();
```
wjones127, line 164: Should we also validate the schema matches the existing schema?

rust/src/operations/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+                    }
+                }
+            }
+            partitions.into_values().collect::<Vec<_>>()
+        } else {
+            vec![data]
+        };
+
+        let operation = DeltaOperation::Write {
+            mode: mode.clone().unwrap_or(SaveMode::Append),
```
wjones127, line 202: nit: Is there a reason we make mode optional? I feel like `None` isn't much more convenient than `SaveMode::Append`, but having to write `Some(SaveMode::Overwrite)` versus just `SaveMode::Overwrite` would be a bit annoying.

wjones127 (commented):
I've gone about halfway through so far. These are impressive changes and I like what I've seen so far. 

I've left several suggestions for now. I'll be back to finish reviewing soon.

rust/src/action.rs
```diff
@@ -845,32 +844,87 @@ impl Action {
...
+
     /// Represents a Delta `Write` operation.
     /// Write operations will typically only include `Add` actions.
+    #[serde(rename_all = "camelCase")]
     Write {
         /// The save mode used during the write.
         mode: SaveMode,
         /// The columns the write is partitioned by.
-        partitionBy: Option<Vec<String>>,
+        partition_by: Option<Vec<String>>,
```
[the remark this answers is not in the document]
roeap: We have that field here in case the table does not exist and we create it during the write operation. When it comes to changing the table schema and or partitioning I am still a bit fuzzy to be honest. I seem to remember reading something that changing partitioning in schema evolution is not allowed, but could not find a reference to it in the delta protocol definitions.

rust/src/action.rs
```diff
@@ -845,32 +844,87 @@ impl Action {
...
+                    "operation".to_string(),
+                    serde_json::Value::String("delta-rs.Write".to_string()),
+                );
+            }
+            DeltaOperation::StreamingUpdate { .. } => {
+                commit_info.insert(
+                    "operation".to_string(),
+                    serde_json::Value::String("delta-rs.StreamingUpdate".to_string()),
+                );
+            }
```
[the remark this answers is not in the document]
roeap: much better!

rust/src/delta.rs
```diff
@@ -1526,6 +1540,36 @@ impl<'a> DeltaTransaction<'a> {
         Ok(PreparedCommit { uri })
     }
 
+    /// Prepare a commit with commit info added
+    pub async fn prepare_commit_with_info(
```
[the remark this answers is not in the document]
roeap: Not really :) Main intent was to avoid breaking existing users, but probably in the end nicer to have just one commit method (same for prepare_commit). Will update the PR.

rust/src/operations/mod.rs
roeap: yes :)

rust/src/operations/mod.rs
roeap: Agreed, will update.

rust/src/action.rs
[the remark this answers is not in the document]
wjones127: > I seem to remember reading something that changing partitioning in schema evolution is not allowed
That sounds right to me.

I guess my point is I don't see the value in keeping it since it's recorded in metadata and shouldn't vary from that.

rust/src/operations/write.rs
```diff
@@ -0,0 +1,458 @@
...
+                    HashMap::new(),
+                );
+                let op = DeltaOperation::Create {
+                    location: self.table_uri.clone(),
+                    metadata: metadata.clone(),
+                    mode: SaveMode::ErrorIfExists,
+                    // TODO get the protocol from somewhere central
+                    protocol: Protocol {
+                        min_reader_version: 1,
+                        min_writer_version: 2,
```
wjones127, line 166: Technically, since you aren't enforcing `appendOnly` or Column invariants, this should be 1 for now.

In addition, I think you should be checking the protocol version of an existing table and erroring if it's > 1.

wjones127 (changes requested):
Looked through the remainder. I think the most important change needed is adding protocol enforcement. We need to make sure we aren't writing to tables with `min_writer_protocol` greater than what we support here. And I think based on what I've seen we are supporting version 1 not 2 ([see docs here](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-version-requirements)).

rust/src/delta.rs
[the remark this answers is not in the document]
roeap: Changed it to updating existing commit functions rather than creating new ones.

rust/src/delta.rs
```diff
@@ -1499,6 +1499,20 @@ impl<'a> DeltaTransaction<'a> {
         Ok(version)
     }
 
+    /// Commit transaction with additional app metadata
+    pub async fn commit_with_info(
+        &mut self,
+        operation: Option<DeltaOperation>,
+        app_metadata: Option<Map<String, Value>>,
+    ) -> Result<DeltaDataTypeVersion, DeltaTableError> {
+        // TODO check operation and decide whether the commit is safe to retry in a loop?
```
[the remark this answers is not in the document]
roeap: The comment in the original function gives us a much better indication what to do :). This is removed now..

rust/src/action.rs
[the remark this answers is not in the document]
roeap: We require that info when creating a new table during write. The original implementation also menations checking the coluumns against an exisiting table and failing otherwise. We do use it to crate a table, the check is not done right now.

rust/src/operations/transaction.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+fn deserialize_actions(data: &RecordBatch) -> DataFusionResult<Vec<Action>> {
+    let serialized_actions = arrow::array::as_string_array(data.column(0));
+    (0..serialized_actions.len())
+        .map(|idx| serde_json::from_str::<Action>(serialized_actions.value(idx)))
+        .into_iter()
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(to_datafusion_err)
```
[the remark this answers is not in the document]
roeap: indeed it does - using it now.

rust/src/operations/write.rs
roeap: absolutely. updated min version and added chack for exisiting table.

rust/src/operations/write.rs
```diff
@@ -0,0 +1,458 @@
...
+    }
+}
+
+#[derive(Debug)]
+/// Writes the partitioned input data into separate batches
+/// and forwards the add actions as record batches
+struct WritePartitionCommand {
+    table_uri: String,
+    // The save mode used in operation
+    // mode: SaveMode,
```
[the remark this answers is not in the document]
roeap: This should not be there at all. We do all the write mode checking in the higher level plan. - removed it.

rust/src/operations/write.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// Column names for table partitioning
+    partition_columns: Option<Vec<String>>,
+    // When using `Overwrite` mode, replace data that matches a predicate
```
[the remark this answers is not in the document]
roeap: yes.

rust/src/operations/write.rs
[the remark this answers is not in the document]
roeap: Thinking about it, we may need it when using predicates, but since we don't right now, keep it removed.

rust/src/operations/write.rs
```diff
@@ -0,0 +1,458 @@
...
+impl ExecutionPlan for WritePartitionCommand {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn schema(&self) -> ArrowSchemaRef {
+        Arc::new(ArrowSchema::new(vec![ArrowField::new(
+            "serialized",
+            DataType::Utf8,
+            true,
```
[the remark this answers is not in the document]
roeap: Makes sense, also be consistent we now use a single lazy static reference to the schema expected from all operations.

roeap:
@houqp, the lambda checkpoint smoke test is failing and to me its not directly obvious why. As I am not too familiar with AWS, could you give me a hint where to look? Or is it maybe even an unrelated cause? In any case I failed to find code that would be affected, but that does not mean much :).

houqp:
@roeap I think can ignore that smoke test for now. Perhaps @xianwill can help fix it in https://github.com/delta-io/delta-rs/pull/586.

rust/src/action.rs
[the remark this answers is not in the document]
houqp: would repartition a valid use for this argument?

rust/src/operations/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+        &mut self,
+        data: Vec<RecordBatch>,
+        mode: SaveMode,
+        partition_columns: Option<Vec<String>>,
+    ) -> DeltaCommandResult<()> {
+        if data.is_empty() {
+            return Ok(());
+        }
+        let schema = data[0].schema();
+        let current_part = match self.table.update().await {
```
houqp, line 169: I think the update call here is not needed because delta writes relies on optimistic concurrency control. I.e. we can always assume the current in memory state is already the latest version. During the table commit write, we will get a file already exists error if we don't have the latest version. When a conflict happens, we can then retry the write after a table update.

Most delta tables only have a single writer, so the table update before write is usually a no op.

rust/src/operations/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+        let operation = DeltaOperation::Write {
+            mode,
+            partition_by: current_part.clone(),
+            predicate: None,
+        };
+        let data_plan = Arc::new(MemoryExec::try_new(&data, schema, None)?);
+        let plan = Arc::new(WriteCommand::try_new(
+            &self.table.table_uri,
+            operation.clone(),
+            data_plan,
```
houqp, line 214: the data is prepared for a particular version of the delta table, so we need to pass in the table version here too for optimistic concurrency control conflict detection during the commit.

rust/src/operations/transaction.rs
```diff
@@ -0,0 +1,175 @@
...
+            .iter()
+            .flat_map(|batch| deserialize_actions(batch).unwrap())
+            .collect::<Vec<_>>();
+
+        if actions.is_empty() {
+            let empty_plan = EmptyExec::new(false, self.schema());
+            return empty_plan.execute(0, context).await;
+        }
+
+        let _new_version = match table.update().await {
```
houqp, line 133: again, I think the update here is not needed due to the optimistic concurrency control design.

rust/src/operations/transaction.rs
```diff
@@ -0,0 +1,175 @@
...
+                    .try_commit_transaction(&prepared_commit, 0)
+                    .await
+                    .map_err(to_datafusion_err)?;
+                committed_version
+            }
+            _ => {
+                let mut txn = table.create_transaction(None);
+                txn.add_actions(actions);
+                let committed_version = txn
+                    .commit(Some(self.operation.clone()), self.app_metadata.clone())
```
houqp, line 151: we need to commit with a specific version that the data is prepared for here, so we can trigger commit conflicts here and let the retry to resolve the conflict.

rust/src/operations/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+        };
+
+        let operation = DeltaOperation::Write {
+            mode,
+            partition_by: current_part.clone(),
+            predicate: None,
+        };
+        let data_plan = Arc::new(MemoryExec::try_new(&data, schema, None)?);
+        let plan = Arc::new(WriteCommand::try_new(
+            &self.table.table_uri,
```
houqp, line 212: This is something we can optimize as followups, but I think in the long run, it would be good to pass the table state here with the version so we won't need to perform a table load when executing the WriteCommand plan.

rust/src/operations/mod.rs
wjones127: Was this added anywhere?

rust/src/operations/transaction.rs
```diff
@@ -0,0 +1,190 @@
...
+        coalesce_partitions::CoalescePartitionsExec, common::compute_record_batch_statistics,
+        empty::EmptyExec, expressions::PhysicalSortExpr, Distribution, ExecutionPlan, Partitioning,
+        SendableRecordBatchStream, Statistics,
+    },
+};
+use lazy_static::lazy_static;
+use std::sync::Arc;
+
+lazy_static! {
+    /// Schema expected form plans wrapped by transaction
```
wjones127, line 27:
```suggestion
    /// Schema expected for plans wrapped by transaction
```

wjones127:
Hey @roeap. Sorry I haven't been able to review in a bit. I looked over and saw most of my comments resolved, but there are still a few that seem unaddressed.

Also looks like there is still some work to do to get the table version check working, but you have started it.

Thanks for your patience, and feel free to post in here or on Slack if you want any other guidance :)

rust/src/operations/mod.rs
roeap: now it is :)

rust/src/operations/transaction.rs
```diff
@@ -0,0 +1,179 @@
...
+        context: Arc<TaskContext>,
+    ) -> DataFusionResult<SendableRecordBatchStream> {
+        let mut table =
+            get_table_from_uri_without_update(self.table_uri.clone()).map_err(to_datafusion_err)?;
+
+        let data = collect(self.input.clone(), context.clone()).await?;
+        // TODO we assume that all children send a single column record batch with serialized actions
+        let actions = data
+            .iter()
+            .flat_map(|batch| deserialize_actions(batch).unwrap())
```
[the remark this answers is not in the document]
roeap: absolutely. to get it to work with `flat_map`, I had to explicitly match the result. A little more verbose then I was hoping for, but I think it's Ok.

rust/src/operations/transaction.rs
```diff
@@ -0,0 +1,179 @@
...
+                txn.add_actions(actions);
+                let committed_version = txn
+                    .commit_with_info(Some(self.operation.clone()), self.app_metadata.clone())
+                    .await
+                    .map_err(to_datafusion_err)?;
+                committed_version
+            }
+        };
+
+        // TODO report some helpful data - at least current version
```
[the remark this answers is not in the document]
roeap: Was hoping to leave updating the result data for another PR. My hope is to be able to maybe even include some statistics abut the new table or data written. TO do this right I think having a more complex command would help thinking about what that should look like.

rust/src/operations/write.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    async fn execute(
+        &self,
+        partition: usize,
+        context: Arc<TaskContext>,
+    ) -> DataFusionResult<SendableRecordBatchStream> {
+        let data = collect_batch(self.input.execute(partition, context).await?).await?;
+        if data.is_empty() {
+            return Err(DataFusionError::Plan(
+                DeltaCommandError::EmptyPartition("no data".to_string()).to_string(),
```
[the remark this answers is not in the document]
roeap: I think datafusion makes no assumptions abut this. Once we return some statistics the users would be informed, so changed it to no-op as suggested.

roeap:
@houqp @wjones127 - addressed all comments (i think). The topic around versioning probably needs some more thought, but maybe in a follow up PR? Specifically removing the dependency on a table entirely and working only with metadata (and version). I think in spark they talk of a snapshot. Elsewhere I have been experiment with a file index (#528), which could (should?) be part of this. This is obviously a larger piece. As a more immediate follow-up I wanted to look into supporting writer V2, as there we could already start to introduce some SQL via the column invariants and start to more explicitly leverage datafusion.

rust/src/operations/transaction.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        let data = collect(self.input.clone(), context.clone()).await?;
+        // TODO we assume that all children send a single column record batch with serialized actions
+        let actions = data
+            .iter()
+            .flat_map(|batch| match deserialize_actions(batch) {
+                Ok(vec) => vec.into_iter().map(Ok).collect(),
+                Err(er) => vec![Err(er)],
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
wjones127, line 133: I think if you separate `map` and `flatten`, it will populate the error nicely:

```suggestion
            .map(deserialize_actions)
            .flatten()
            .collect::<Result<Vec<_>, _>>()?;
```

rust/src/operations/write.rs
```diff
@@ -0,0 +1,466 @@
...
+//! New Table Semantics
+//!  - The schema of the [RecordBatch] is used to initialize the table.
+//!  - The partition columns will be used to partition the table.
+//!
+//! Existing Table Semantics
+//!  - The save mode will control how existing data is handled (i.e. overwrite, append, etc)
+//!  - The schema of the RecordBatch will be checked and if there are new columns present
+//!    they will be added to the tables schema. Conflicting columns (i.e. a INT, and a STRING)
+//!    will result in an exception
+//!  - The partition columns, if present are validated against the existing metadata. If not
```
wjones127, line 12: Adding some missing punctuation:

```suggestion
//!    will result in an exception.
//!  - The partition columns, if present, are validated against the existing metadata. If not
```

wjones127 (dismissed):
It looks good to me. I've made a couple optional style suggestions. Thanks for all your effort on this!

It seems like the concurrent commit is handled, but would love for that to have an explicit test at some point later.

houqp (approved):
I am cool with us handling the versioning as a follow up PR since this PR is getting big :D But I think we should explicitly mention in the doc that the current transaction commit implementation is incorrect and could lead to data duplication/loss.

houqp (approved):

rust/src/operations/transaction.rs
roeap: turns out this would need to chained `flattens`, definately a matter of taste, but in that case i personally like the explicit matching better :).

rust/src/action.rs
```diff
@@ -845,32 +844,75 @@ impl Action {
...
+        };
+        commit_info.insert(
+            "operation".to_string(),
+            serde_json::Value::String(operation.into()),
+        );
+
+        if let Ok(serde_json::Value::Object(map)) = serde_json::to_value(self) {
+            commit_info.insert(
+                "operationParameters".to_string(),
+                map.values().next().unwrap().clone(),
```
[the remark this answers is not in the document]
roeap: just checked again, turns out we are adding all the parameter keys and values. The map form `serde_json::to_value(self)` looks something like `{"create": {"key": "value", ...}}` so I think we can assume that we always have exactly one relevant key and are adding the full "inner" map.

roeap:
@houqp - just making sure i understand things correctly. By now we pass a single version to the transaction as a basis for our commit, and basically have three operations - create, write append and write overwrite. within the transaction we always try the commit loop. for the first two operations this seems fine to me, in the case of overwrite we might end up with more data then expected, since someone else might have written some data to the table while we were preparing our data which would then not be included as a remove action in the commit. Is that correct?

houqp:
>  for the first two operations this seems fine to me, in the case of overwrite we might end up with more data then expected,

My understanding is that only append is safe for automatic retry with a newer version. For create, if newer versions have been created, that means we should fail the create action, no?

For overwrite, the example you mentioned is a valid correctness bug. We could end up with data corruptions too. For example, if a newer version deleted the same file we are trying to delete in the prepared commit, we will end up with two remove action in the transaction log pointing to the same file, which is invalid. Here is another example that would result in unexpected data loss:

roeap:
This seems like something we want to get right ASAP :). Thinking about next steps, this should probably be handled in the try / commit loop directly?

roeap:
Had another quick read-through of the delta protocol file. The main thing there relevant for commits / retry I found was [this](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#consistency-between-table-metadata-and-data-files) rather slim section, which outlines the consistency between log and data files. I guess then it up to us to think about when its safe to retry and when not ...?

roeap:
@houqp @wjones127 - update the arrow and datafusion dependencies to the latest versions, which includes a change to the `ExecutionPlan` APIs. For now I did the minimal change to make it work, applying a pattern adopted form the datafusion internal plans.

houqp:
> Thinking about next steps, this should probably be handled in the try / commit loop directly?

Commit retry loop with automatic version increment is only safe to do for append only writes.

>  I guess then it up to us to think about when its safe to retry and when not ...?

The retry semantic needs to be handled at a layer that's higher than our transaction commit layer because it requires context of updates at row level as well as data dependency. The official write conflicts documentation can be used as a good reference for this: https://docs.databricks.com/delta/concurrency-control.html#write-conflicts. Any commits created from high level operations that's noted as "can conflict" cannot be retried automatically by delta-rs on version conflicts.

Basically the idea is to propagate the conflict error all the way up to the application layer (e.g. caller of delta-rs like datafusion). Then let the application layer decide how to resolve the conflict and retry. Sometimes it might be ok to just retry the same write without reading the new data, but most of the times, the application will need to catch up to the latest table state, recalculate the new set of rows to write, then retry the commit.

rust/src/action.rs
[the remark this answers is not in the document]
houqp: Oh i see, we are trying to get rid of the enum variant layer, thanks for the clarification :+1:  Could you a comment in the code to mention what you said in the comment? It was obvious to me when I first read it.

Ideally, it's better to manage each enum option as its own struct, then generate the option json object from within the first `match &self {` block. But since this is not a hot code path, I think it's fine to leave this as is for now :)

houqp (approved):
LGTM

PR #585, opened 2022-04-05
Replace table location prefix from s3a to s3

novakov-alexey:
Replace **s3a** table location prefix to **s3** when table location is taken from Glue catalog .

# Description
Table location from Glue Catalog is modified to let the Rust code to use native S3 protocol, instead of s3a, which is used when table is ingested by Spark. This patch is fixing "path not found" error when using Python API to load table from Data Catalog: 
- Loading Delta table in Python from Data Catalog: https://delta-io.github.io/delta-rs/python/usage.html#loading-a-delta-table

# Documentation

- Hadoop S3A used by Spark: https://hadoop.apache.org/docs/current2/hadoop-aws/tools/hadoop-aws/index.html#S3A
- Spark Cloud Integration: https://spark.apache.org/docs/latest/cloud-integration.html#recommended-settings-for-writing-to-object-stores

### DIFF
--- a/rust/src/data_catalog/glue/mod.rs
+++ b/rust/src/data_catalog/glue/mod.rs
@@ -91,6 +91,7 @@ impl DataCatalog for GlueDataCatalog {
                 metadata: "Storage Descriptor".to_string(),
             })?
             .location
+            .map(|l| l.replace("s3a", "s3"))
             .ok_or(DataCatalogError::MissingMetadata {
                 metadata: "Location".to_string(),
             });


houqp (approved):
That's a clever workaround :)

wjones127:
I haven't used AWS Glue before, but it sounds like this location field contains more than the protocol? From the [AWS docs](https://docs.aws.amazon.com/glue/latest/webapi/API_StorageDescriptor.html): 

> The physical location of the table. By default, this takes the form of the warehouse location, followed by the database location in the warehouse, followed by the table name.

Is it not possible to have a table name with the string "s3a"? I haven't been able to find docs that provide example of what this looks like, so I'm not sure where "s3a" shows up, but it might be better to have a more targeted replacement.

fvaleye (approved):
Thanks @novakov-alexey 

LGTM! 

Did you test test this patch with Glue on your side?

novakov-alexey:
@fvaleye Answering to your second question: if use official Delta Lake with Spark and run DataFrame save method with "delta" as format, it saves my table in Glue Catalog with s3a prefix as table location. I think it is standard behaviour of Apache Spark. Please take a look at my links to this PR related to Spark. 
If I change Table Location in Glue Table manually (using DDL or Glue Web-Console), then Python delta works fine.

novakov-alexey:
Python error I get when s3a is in table location: 
`PyDeltaTableError: Failed to read delta log object: Invalid object URI`

novakov-alexey:
Just tested my patch with AWS Glue Catalog successfully. 

```
In [10]: dt = DeltaTable.from_data_catalog(data_catalog=data_catalog, database_name=database_name, table_name=table_name)

In [16]: dt.version()
Out[16]: 8
```

houqp:
yeah, the s3a scheme is the one that's used in all hadoop/spark systems due to historical reasons :(

houqp (approved):
I will let @fvaleye merge the PR, thanks again @novakov-alexey

fvaleye:
Thanks again @novakov-alexey 👍

PR #586, opened 2022-04-15
Change private time_utils module to public.

xianwill:
# Description

In its custom writer implementation, [kafka-delta-ingest](https://github.com/delta-io/kafka-delta-ingest) uses the `timestamp_to_delta_stats_string` function which was recently moved into its own private `time_utils` module. This PR makes `time_utils` public so we continue using the same code when upgrading to the newest version of `delta-rs`.

Drive by: Temporarily disable lambda checkpoint test until we get the checkpoint memory pressure figured out.


houqp (dismissed):

houqp:
looks like the lambda smoke test is a little bit flaky :(

houqp (approved):

PR #587, opened 2022-04-16
Allow metadata for write_deltalake

### LINKED ISSUE #576: Python: allow metadata for write_deltalake
# Description

Currently, `write_deltalake()` doesn't support passing in a table name or description, or additional custom config values. 

**Related Issue(s)**
Umbrella issue: #542

PadenZach:
# Description
+ Add Name, Description, and Configuration arguments to write_new_datalake in python and python-rust bindings
+ Add round trip metadata writing test
+ Suppress write clippy "too many arguments" (Default max seems to be 7, function has 8). 

# Related Issue(s)
Should resolve #576 

# Documentation

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -56,6 +56,8 @@ def write_deltalake(
     to write to an existing table with a higher min_writer_version, this
     function will throw DeltaTableProtocolError.
 
+    Note that this function does NOT register this table in a data catalog.
+
     :param table_or_uri: URI of a table or a DeltaTable object.
     :param data: Data to write. If passing iterable, the schema must also be given.
     :param schema: Optional schema to write.
@@ -67,6 +69,9 @@ def write_deltalake(
         already exists. If 'append', will add new data. If 'overwrite', will
         replace table with new data. If 'ignore', will not write anything if
         table already exists.
+    :param name: User-provided identifier for this table.
+    :param description: User-provided description for this table.
+    :param configuration: A map containing configuration options for the metadata action.
     """
     if isinstance(data, pd.DataFrame):
         data = pa.Table.from_pandas(data)
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -3,6 +3,7 @@ import uuid
 from dataclasses import dataclass
 from datetime import date, datetime
 from decimal import Decimal
+from optparse import Option
 from typing import Any, Dict, Iterable, Iterator, List, Mapping, Optional, Union
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


houqp (dismissed):
LGTM, will leave it to @fvaleye and @wjones127 to do the final review and merge.

houqp:
looks like python linter is also complaining about too many arguments.

PadenZach:
Not sure why the lambda_checkpoint_build failed, doesn't seem to be related to changes made here. LMK if I need to change anything else here, but doesn't seem to be required so I'll be leaving it as-is for now.

python/deltalake/writer.py
```diff
@@ -44,6 +44,9 @@ def write_deltalake(
     partition_by: Optional[List[str]] = None,
     filesystem: Optional[pa_fs.FileSystem] = None,
     mode: Literal["error", "append", "overwrite", "ignore"] = "error",
+    name: Optional[str] = None,
+    description: Optional[str] = None,
+    configuration: Optional[Mapping[str, Optional[str]]] = None,
```
wjones127, line 49: Could you add entries to the docstring below for each of these parameters?

python/deltalake/writer.py
```diff
@@ -44,6 +44,9 @@ def write_deltalake(
     partition_by: Optional[List[str]] = None,
     filesystem: Optional[pa_fs.FileSystem] = None,
     mode: Literal["error", "append", "overwrite", "ignore"] = "error",
+    name: Optional[str] = None,
```
wjones127, line 47: Could we also add a note in the docstring that this function does *not* register this table in your data catalog? Users will have to either use the filesystem path to the table or manually register this table in the catalog. I'll create a follow up issue for us to handle that though.

wjones127 (changes requested):
Thanks! Code changes look pretty clean, just needs a few updates to the documentation. 👍 

Also this made me realize we'll also probably want AWS Glue and Hive metastore integration; otherwise giving the table a name isn't that useful 😕 . But I think it's worth doing that in a different PR.

wjones127:
And don't worry about the checkpoint build, we've seen it be flaky recently and it doesn't seem like it could be related to these changes.

python/deltalake/writer.py
PadenZach: Added a brief note. Let me know if more explanation re: manual add is needed.

python/deltalake/writer.py
PadenZach: Done 👍

fvaleye (approved):
Thank you @PadenZach 

LGTM 👍

wjones127 (approved):

wjones127:
Thank you @PadenZach!

Issue #575, opened 2022-03-23
Python: Support writer protocol V2 in write_deltalake


wjones127:
# Description

Most delta tables are now V2 by default, so writer isn't yet compatible with most systems.

Two features needed are:

 * [x] Add support for enforcing `delta.appendOnly`. Just need to check the delta config, check the `mode` parameter, and throw an error if needed. (https://github.com/delta-io/delta-rs/pull/590)
 * [ ] Add support for enforcing [column invariants](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-invariants). This part is more complex. Need to parse the SQL expression into a PyArrow expression, evaluate that expression on the data, and error as soon an invalid data is found. (https://github.com/delta-io/delta-rs/issues/592)
 

**Related Issue(s)**
Umbrella issue:#542

PadenZach:
@GraemeCliffe-inspirato Are you still interested in taking the append only part of this issue? If not, I may be able to look into it. If so I'll hold off and see if there's other good first issues for me on this :)

WarSame:
@PadenZach I'm still interested in the appendOnly section! Thank you for checking! I have just now got the project building and tests running, so I'm just starting on the issue

wjones127:
My initial guess on what needs to be done on invariants is something like this:

```python
# ...inside of write_deltalake
def iter_batches(data, invariants) -> Iterator[RecordBatch]:
    for batch in data:
        for sql_clause in invariants:
           res = execute(sql_clause, batch)
           if res != True:
               raise ValueError("Invariant violated: ...")
           yield batch

invariants = configuration["delta.invariants"]
data = convertToRecordBatchReader(data)

batch_iter = iter_batches(data, invariants)
# ... pass batch_iter to write_dataset
```

The hard part is we don't have a SQL parser in the deltalake package, so not sure how that `execute()` function would work.  One option is to turn on the datafusion delta-rs option in Python (which I suspect we might do eventually anyways) and then implement `execute()` in Rust using datafusion. Moving the record batch temporarily into Rust with zero-copy should be possible (that's used by python-datafusion), but might take a little bit of glue code.

It's probably worth researching what typical invariants are used and allowed by existing engines. The spec is very vague, but it's likely the Spark implementation has a limited set of column types and operations we need to care about.

PR #590, opened 2022-04-22
Python API - delta.appendOnly enforcement

WarSame:
# Description
Adding enforcement of `delta.appendOnly` to the Python API. If `delta.appendOnly` is passed in as a key to the configuration and the writing `mode` is not `append` then the write will fail with a ValueError.

# Related Issue(s)
This deals with the first bullet point in #575 . The second bullet point about column invariants is not handled here, so this PR does not close the issue on its own.

# Documentation

This configuration field is already covered in the `usage.rst` -> "Examining a Table" -> "Metadata" section.


wjones127 (commented):
Thanks for making this PR!

It looks like you've tested for the case where the configuration is passed while creating the table, but what about if the table already exists? That configuration should be saved with the table. To accomplish this, we should be able to read the existing table configuration from the metadata property: https://github.com/delta-io/delta-rs/blob/4aad6aa53df89513fb25d53a5805c55ea0b07693/python/deltalake/table.py#L54 (And sidenote: I just noticed the typehint on that configuration property is wrong; it should be `Dict[str, str]`, not `List[str]`. If you could correct that in this PR, it would be appreciated 👍 .)

python/tests/test_writer.py
```diff
@@ -172,6 +172,24 @@ def test_write_modes(tmp_path: pathlib.Path, sample_data: pa.Table):
     assert DeltaTable(path).to_pyarrow_table() == sample_data
 
 
+def test_append_only(tmp_path: pathlib.Path, sample_data: pa.Table):
```
fvaleye, line 175: **nitpick:** I would suggest renaming this method to `test_append_only_should_append_only_with_the_overwrite_mode` to improve the test readibility.

fvaleye:
Thank you @WarSame 👍 

LGTM, ready to merge after fixing the [formatting issue](https://github.com/delta-io/delta-rs/runs/6123515763?check_suite_focus=true) in the CI.

WarSame:
Sorry for the delay. I've had Covid for the last week so this has taken the backburner.

I believe I've fixed the formatting issues, though it appears the workflows need approval to run. I did add code to account for an existing table's metadata, and updated the tests to match that. Finally, it was getting to be a few lines of code to check these conditions so I pulled it out into its own functions. Please let me know if you prefer it inline.

python/deltalake/writer.py
```diff
@@ -202,6 +206,22 @@ def visitor(written_file: Any) -> None:
         )
 
 
+def __throw_error_if_delta_append_only_and_mode_not_append(
```
wjones127, line 209: nit: I think this function name could be a bit shorter. What do you think of `__enforce_append_only()`? I think the existing name would work well as a doc string.

python/tests/test_writer.py
```diff
@@ -176,6 +176,32 @@ def test_write_modes(tmp_path: pathlib.Path, sample_data: pa.Table):
...
+
+    table = DeltaTable(path)
+    write_deltalake(table, sample_data, mode="append")
+
+    fail_modes = ["overwrite", "ignore", "error"]
+
+    for mode in fail_modes:
+        with pytest.raises(
+            ValueError,
+            match="If configuration has delta.appendOnly = 'true', mode must be 'append'",
```
wjones127, line 196: I like that it checks the error message 👍

python/tests/test_writer.py
```diff
@@ -176,6 +176,32 @@ def test_write_modes(tmp_path: pathlib.Path, sample_data: pa.Table):
...
+        ):
+            write_deltalake(path, sample_data, mode=mode)
+            write_deltalake(table, sample_data, mode=mode)
```
wjones127, line 199: If it errors on the first of these, we aren't testing the second, right? Or if it doesn't error on the first line, the test could still pass but we wouldn't like that behavior!

So maybe better to make those separate iterations:

```python
# Test both passing in path string and DeltaTable
possible_args = [path, table]
for arg, mode in itertools.product(possible_args, fail_modes):
    with pytest.raises(...):
        write_deltalake(arg, sample_data, mode=mode)

```

python/tests/test_writer.py
```diff
@@ -176,6 +176,32 @@ def test_write_modes(tmp_path: pathlib.Path, sample_data: pa.Table):
...
+
+    expected = pa.concat_tables([sample_data, sample_data])
+    assert table.to_pyarrow_table() == expected
```
wjones127, line 202: To be extra safe, you could also test that the `DeltaTable.version()` hasn't changed since the last expected append. (It should equal 1, I think.)

wjones127:
> Sorry for the delay. I've had Covid for the last week so this has taken the backburner.

Oh no! Well I hope you are feeling better now.

> Finally, it was getting to be a few lines of code to check these conditions so I pulled it out into its own functions. Please let me know if you prefer it inline.

Yeah that looks good to me.

wjones127 (changes requested):
I have a few suggestions for the test.

python/deltalake/writer.py
```diff
@@ -202,6 +206,22 @@ def visitor(written_file: Any) -> None:
...
+    table: Optional[DeltaTable],
+    configuration: Optional[Mapping[str, Optional[str]]],
+    mode: str,
+) -> None:
```
fvaleye, line 213: Could you please add a docstring to this method (you could also mention that it will throw an exception)?

python/deltalake/writer.py
```diff
@@ -202,6 +206,22 @@ def visitor(written_file: Any) -> None:
...
+    if table:
+        configuration = table.metadata().configuration
+    config_delta_append_only = (
+        configuration and configuration.get("delta.appendOnly", "false") == "true"
+    )
+    if config_delta_append_only and mode != "append":
+        raise ValueError(
+            "If configuration has delta.appendOnly = 'true', mode must be 'append'"
```
fvaleye, line 221: **nitpick:** add the mode in the `ValueError` to get the current value.

python/tests/test_writer.py
WarSame: Thank you! I was not aware of this feature!

If it had errored on the first we wouldn't have tested the second, but if we hadn't errored on the first we should have tested the second. Either way, I had intended to test both, and your recommended method is great for that.

python/deltalake/writer.py
WarSame: Please let me know if you want more in the docstring.

python/deltalake/writer.py
WarSame: Thanks for the callout! This is a great point to help people who might cross this error.

wjones127:
@WarSame Looks like you just need to sort the imports, and then this PR is good-to-go. 😄

WarSame:
@wjones127 I've done so and it should be good to go!

wjones127 (approved):

PR #594, opened 2022-05-03
Introduce require_files for tracking the add files in table state

mosyp:
# Description
Since #454 dev is stopped. We aim to gradually address that issue.
Firstly, this PR. The changes introduces  `require_files` which similar to `require_tombstones`, filters out any files if the flag is set to `false`. Hence the table state will end up with metadata only. This is a perfect behavior for append only apps (like kafka delta ingest for example). 

Secondly, the catch is the how to create a checkpoint, because we need both adds & removes for that. One possible approach is a create checkpoint in another process, which is applicable. However we can omit that and do in process if we apply the work from #454. E.g. by operating on arrow/record batch objects to reduce memory usage / decentralization just enough to read/create checkpoint.

Thirdly, if 2nd step is successful, we can leverage that experience to finish/enhance the idea from #454

# Related Issue(s)
A follow up of https://github.com/delta-io/delta-rs/pull/445. But here we introduce ignoring of add files.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -429,21 +429,30 @@ impl Default for DeltaVersion {
     }
 }
 
-/// configuration options for delta table
+/// Configuration options for delta table
 #[derive(Debug)]
 pub struct DeltaTableConfig {
-    /// indicates whether our use case requires tracking tombstones.
-    /// read-only applications never require tombstones. Tombstones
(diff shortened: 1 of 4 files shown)


xianwill (approved):

houqp (approved):

houqp:
nice optimization for append only workload!

PR #595, opened 2022-05-04
Avoid building pandas and numpy from source

wjones127:
# Description

I noticed the CI job " Python Build (Python 3.7 PyArrow 4.0.0) " was taking a while, and it looks like it was building pandas from source. 

Before (16m45s): https://github.com/delta-io/delta-rs/runs/6282672393?check_suite_focus=true
After (8m57s): https://github.com/delta-io/delta-rs/runs/6283588059?check_suite_focus=true

Also added caching to the latest Python build, since it was easy. Can't add it to minimal due to issues with old glibc.

# Related Issue(s)

# Documentation

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -44,6 +44,16 @@ jobs:
     # docker based action.
     - uses: actions/checkout@v1
 
+    - uses: actions/cache@v2
+      with:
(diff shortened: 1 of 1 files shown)


houqp (approved):
nice, that's a big win :D

PR #597, opened 2022-05-05
Make sure pandas is optional

### LINKED ISSUE #596: The experimental writer need pandas dependency
# Environment

**Delta-rs version**: `python-v0.5.7`

**Binding**: Python

***
# Bug

**What happened**:
When installing the `deltalake` python package without [pandas](https://github.com/delta-io/delta-rs/blob/main/python/pyproject.toml#L24) as optional dependency.

The import fails in the `writer.py` file since it required the `pandas` dependency:
```
site-packages/deltalake/writer.py:8: in <module>
    import pandas as pd
E   ModuleNotFoundError: No module named 'pandas'
```

Have everything ready to work without requiring to install pandas if we are not using the writer capability.

**Possible solutions**:
- Create a dedicated `write_deltalake_from_pandas` method to avoid checking the instance of the data as dataframe
- Add only pandas for types checking like [this](https://github.com/delta-io/delta-rs/blob/main/python/deltalake/table.py#L10)
- Check if pandas is installed with a bool when importing and add this check before [this check](https://github.com/delta-io/delta-rs/blob/main/python/deltalake/writer.py#L100)

wjones127:
# Description

Pandas is supposed to be an optional dependency. This PR adjust CI to enforce that, fixes the issues in the writer sub-module, and also make pandas optional for unit tests.

Thanks @fvaleye for reporting this issue!

# Related Issue(s)

- closes #596

# Documentation

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -124,7 +124,7 @@ jobs:
     - name: Test without pandas
       run: |
         source venv/bin/activate
-        pip uninstall pandas
+        pip uninstall --yes pandas
         python -m pytest -m "not pandas and not integration"
 
     - name: Build Sphinx documentation
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -120,6 +120,11 @@ jobs:
       run: |
         source venv/bin/activate
         make unit-test
+    
+    - name: Test without pandas
+      run: |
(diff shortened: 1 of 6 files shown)


houqp (approved):

fvaleye (approved):
Perfect! 

Thank you @wjones127 👍

Issue #555, opened 2022-02-02
Enable configuring Azure backend explicitly


thovoll:
# Description
Enable configuring Azure backend explicitly (rather than through env).

This enables different backend configuration for two different delta tables (which is a need in kafka-delta-ingest for dead lettering to delta). 

**Use Case**
https://github.com/delta-io/kafka-delta-ingest/pull/69

**Related Issue(s)**
#418

PR #598, opened 2022-05-07
Configure Azure storage using a map (#555)

Blajda:
# Description
Allows a user to create a DeltaTable backed by ADLS storage by simply passing uri and a map of options.

This helps promote the usage of different storage backends by only having to change some configuration rather than explicitly creating a storage backend instance in the user's code. 

# Related Issue(s)
Related to #555 since it allows an alternative from using environment variables for creation.
# Documentation

### DIFF
--- a/rust/src/storage/azure/mod.rs
+++ b/rust/src/storage/azure/mod.rs
@@ -14,11 +14,17 @@ use azure_storage_blobs::prelude::*;
 use azure_storage_datalake::prelude::*;
 use futures::stream::Stream;
 use log::debug;
+use std::collections::HashMap;
 use std::env;
 use std::error::Error;
 use std::fmt::Debug;
 use std::sync::Arc;
 use std::{fmt, pin::Pin};
+
+///The ADLS Gen2 Access Key
+pub const AZURE_STORAGE_ACCOUNT_KEY: &str = "AZURE_STORAGE_ACCOUNT_KEY";
+///The name of storage account
+pub const AZURE_STORAGE_ACCOUNT_NAME: &str = "AZURE_STORAGE_ACCOUNT_NAME";
 
 /// An object on an Azure Data Lake Storage Gen2 account.
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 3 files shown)


roeap:
Thanks @Blajda! So far you PR looks good, and this change has been long overdue :).

Making initialization of the Azure backend more flexible and independent environment variables is a great improvement. I am curious as to your requirements. As one of the next tasks I wanted use the Credential structs from that Azure crates to pass configuration to the backend, which would also allow for using service principals and managed identities to meet the needs of production deployments.

Maybe we could start that journey, by modelling the storage options after how it is done with S3 
https://github.com/delta-io/delta-rs/blob/fae92787a8140c3b84f8cad4949776bdd70d2bae/rust/src/storage/s3/mod.rs#L212-L226

What do you think?

houqp:
Thanks @Blajda for taking a stab at this. Instead of using a generic map, why not create a strongly typed option struct? That way, we can perform type checks on option key(field) name and value type at build time.

Blajda:
@roeap Thanks for the feedback. 

This is a background of my requirements and the intended goal of this PR. When developing a new application that uses a Delta table I like to easily swap between the ADLS storage and the local filesystem without having to make any code changes. I'm able to do so using code snippet below.

I agree and I can try to start modeling support for alternative authentication methods. I took a quick look at the Azure crate and saw that the `DataLakeClient` now supports `TokenCredential` so we make some updates on our end add those authentication methods.

I'll try some stuff and update the PR accordingly.

```rust
pub fn create_config_from_env<S: Into<String>>(prefix: S) -> HashMap<String, String> {
    let mut map = HashMap::new();
    let pattern = prefix.into() + "_";

    for (key, value) in env::vars() {
        if let Some(k) = key.strip_prefix(&pattern) {
            map.insert(k.to_string(), value);
        }
    }

    map
}

pub async fn run() -> Result<(), Box<dyn Error>> {
    let config = create_config_from_env("MYTABLE");
    let uri = config.get("URI").unwrap().to_owned();
    let backend =
        get_backend_for_uri_with_options(&uri, config)?;
    let delta_builder = DeltaTableBuilder::from_uri(&uri)?
        .with_storage_backend(backend);
    let table = delta_builder.load().await?;
}
```

roeap:
@Blajda - on my end a fairly pressing need has come up to implement some of these changes as well as some follow up? Thus my question, are you working on this right now, and if not would you be OK if I take over?

Blajda:
Hi @roeap, I pushed my current changes. Currently all the backend tests pass.

I encountered some issues with updating everything to use the file system file in particular `list_objs`. Unfortunately the `Pageable` trait from Azure is `!Send` so I had to pull in a single threaded tokio executor to resolve it. I'm not 100% happy with the current solution but I if you have any guidance I'd appreciate it.

roeap:
@Blajda - that was quick :), thanks! Will have a look later today. The issue with `Pageable` in the azure SDK is know, there is a ticket already open in the repo over there - for now we have to live with it :(

rust/src/storage/azure/mod.rs
```diff
@@ -7,18 +7,28 @@
 /// `AZURE_STORAGE_ACCOUNT_NAME` is required to be set in the environment.
```
roeap, line 7: nit: maybe remove the comment that the env is required, we have that also on top of the new method. Ideally metion the new alternative init approaches.

rust/src/storage/azure/mod.rs
```diff
@@ -62,46 +72,92 @@ impl AdlsGen2Backend {
...
-                storage_account_name.to_owned(),
-                storage_account_key.to_owned(),
-            ),
+        map.insert(AZURE_STORAGE_ACCOUNT_NAME.to_string(), storage_account_name);
+        map.insert(AZURE_STORAGE_ACCOUNT_KEY.to_string(), storage_account_key);
+
+        Self::from_map(file_system_name, map)
+    }
+
+    ///TODO
```
roeap, line 91: do todo :)

rust/src/storage/azure/mod.rs
```diff
@@ -126,25 +182,75 @@ fn to_storage_err(err: Box<dyn Error + Sync + std::marker::Send>) -> StorageErro
 }
 
 fn to_storage_err2(err: azure_storage::core::Error) -> StorageError {
+    if let azure_storage::core::Error::CoreError(azure_core::Error::Other(ref other_err)) = err {
```
roeap, line 185: this does not look pretty, but unfortunately is required right now. Luckily there is ongoing work in the azure crates to "flatten" the errors...

rust/src/storage/azure/mod.rs
```diff
@@ -126,25 +182,75 @@ fn to_storage_err(err: Box<dyn Error + Sync + std::marker::Send>) -> StorageErro
...
+                }
+                None => {}
+            },
+            None => {}
+        }
+    }
     StorageError::AzureStorage { source: err }
 }
 
+async fn list_obj_future(
```
roeap, line 201: nit: maybe add a quick todo comment here, that we want to move away from this once, Pageable is sent in the azure sdk.

roeap:
@Blajda - great work, thanks! The pageable thing is unfortunate, but you handled it nicely. Very few very minor comments

Was thinking about having a typed options struct like in S3, but we can leave that for a follow up PR.

rust/src/storage/azure/mod.rs
Blajda: done!

rust/src/storage/azure/mod.rs
Blajda: Yeah it definitely a pain point when making this change.  There are multiple definitions of HttpError which was confusing at first.

rust/src/storage/azure/mod.rs
Blajda: Done!

rust/src/storage/azure/mod.rs
Blajda: Removed the top level comment since it was duplicate of the new comments. Updated it to mention alternative auth methods

Blajda:
Thanks @roeap I updated the todos and the comments

I'm not sure on the right approach for the typed option struct but at least having support for token credentials is a good start.

roeap (approved):
LGTM - thanks again @Blajda

PR #603, opened 2022-05-10
default to root for empty path in azure store

### LINKED ISSUE #602: ADLS2 backend fails to load table from container root
see discussion in #600

roeap:
# Description

Defaults to a root path if there is no path segment in an azure storage path.

i..e this does work right now `adls2://account/container/`, while this does not work `adls2://devargusadlsv2/simple`. THis PR makes the second option also work.

# Related Issue(s)

closes #602

# Documentation


thovoll:
Thanks @roeap! Are the tests passing from your local machine?

roeap:
yes, they are :) However, I am hoping to migrate the last function in the azrue sdk to finally activate tests in cicd...

thovoll (approved):
Thanks!

PR #606, opened 2022-05-15
Azure options

### LINKED ISSUE #527: Programmatic authentication for Azure
# Description
It would be nice to provide API which allows to specify the Azure credentials in the code. Currently the Azure storage implementation only relies on the environment variables, which makes it difficult to access two independent storages in the same program. In case of AWS S3 the problem does not exists because the API allows to use a custom rusoto S3 client with a static credentials provider.

**Use Case**
A program needs to access two delta tables stored in Azure using two independent access credentials.  

**Related Issue(s)**

### LINKED ISSUE #554: Use released Azure SDK crates
https://crates.io/crates/azure_core
https://crates.io/crates/azure_identity
https://crates.io/crates/azure_storage_blobs
https://crates.io/crates/azure_storage_datalake

### LINKED ISSUE #555: Enable configuring Azure backend explicitly

roeap:
# Description

The PR improves flexibility and ease of use for creating `Adls2StorageBackaends`.

* Introduce `AzureStorageOptions` a typed struct to make azure config options explicit.
* Harmonizes option handling between S3 and Azure.
* explicitly support service principal authorization ontop of previously introduced TokenCredential

The azure backend tests were successfully executed locally.

# Related Issue(s)

closes #555
closes #527
closes #554 
#554 was actually done before, but should be closed anyhow

# Documentation

### DIFF
--- a/rust/src/storage/mod.rs
+++ b/rust/src/storage/mod.rs
@@ -5,14 +5,12 @@ use std::pin::Pin;
 
 use chrono::{DateTime, Utc};
 use futures::Stream;
+use std::collections::HashMap;
 
 #[cfg(feature = "azure")]
 use azure_core::{Error as AzureError, HttpError as AzureHttpError};
 #[cfg(feature = "azure")]
 use std::error::Error;
-
-#[cfg(any(feature = "s3", feature = "s3-rustls"))]
-use self::s3::S3StorageOptions;
 
 #[cfg(feature = "azure")]
 pub mod azure;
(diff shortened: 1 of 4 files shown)


rust/src/storage/mod.rs
```diff
@@ -592,26 +590,47 @@ pub fn get_backend_for_uri(uri: &str) -> Result<Box<dyn StorageBackend>, Storage
...
+}
+
+#[allow(unused)]
+pub(crate) fn str_option(map: &HashMap<String, String>, key: &str) -> Option<String> {
+    map.get(key)
+        .map_or_else(|| std::env::var(key).ok(), |v| Some(v.to_owned()))
+}
+
+#[allow(unused)]
+pub(crate) fn u64_or_default(map: &HashMap<String, String>, key: &str, default: u64) -> u64 {
```
houqp, line 632: my personal preference is only move these helper functions into parent scope when they are actually being shared by more than one storage module. alternatively, if you want to keep them here, it might be better to set a conditional compilation attribute like `#[cfg(any(feature = "s3", feature = "s3-rustls"))]` instead of `#[allow(unused)]`

houqp (approved):
very nice clean up :+1:

rust/src/storage/mod.rs
roeap: i was going back and forth on this - i..e have all similar functions in one place vs. move only whats needed to top level. Your comment tipped the scales :). Also hiding the shared stuff behind feature gate for azure and s3* features...

houqp (approved):

Issue #98, opened 2021-03-01
Implement optimize command


xianwill:
Databricks delta lake provides an [optimize command](https://docs.databricks.com/spark/latest/spark-sql/language-manual/delta-optimize.html) that is incredibly useful for compacting small files (especially for files written by streaming jobs which are inevitably small). delta-rs should provide a similar optimize command.

I think a first pass could ignore the bin-pack and zorder features provided by Databricks and simply combine small files into large files while also setting `dataChange: false` in the delta log entry. An MVP might look like:

* Query the delta log for file stats relevant to the current optimize run (based on predicate)
* Group files that may be compacted based on common partition membership and less than optimal size (1 GB)
* Combine files by looping through each input file and writing its record batches into the new output file.
* Commit a new delta log entry with a single add action and a separate remove action for each compacted file. All committed add and remove actions must set the dataChange flag as false to prevent re-processing by stream consumers.

 Files no longer relavant to the log may be cleaned up later by vacuum (see https://github.com/delta-io/delta-rs/issues/97)

MironAtHome:
While vacuum is a big step forward, it does not obviate need for optimize. These two commands have different purpose, and ( looking further out ) optimize will eventually become part of either batch or background worker, continuously operating with lower priority, as is common to all relational engines.

MrPowers:
@wjones127 - looks like #607 was merged!!  Will it be relatively easy to expose this functionality via the Python bindings?

wjones127:
@MrPowers Yeah shouldn't be too difficult. I created #622 to track that.

wjones127:
Resolved by #607.

PR #607, opened 2022-05-16
Bin packing optimization

Blajda:
# Description
An optimization implementation provided by Databricks is bin-packing which coalesces small files into a larger file. This reduces the number of calls to the underlying storage and results in faster table reads.

This is a high level description of the process. A user can provide a filter for which partitions they want to optimize. Active add actions for those partitions are obtain and then placed into bins labeled by the partition. Once actions are in their respective bin, start building additional bins which consists of a list files to be merged together. Files with a size larger than `delta.targetFileSize` or bins with only a single file are not optimized.

Each bin is then processed and the smaller files are written to a larger one. Corresponding Add and Remove actions are created. Metrics on how many files were considered and total file size are also captured.

# Related Issue(s)
- MVP for #98

# Issues

1. Currently the writer's provided by delta-rs write partition information to the parquet file. This differs from the Databricks implementation which does not. This causes a schema mismatch to occur when packing files from different writers. This is currently handled by dropping partition columns when rewriting.

2. Similar to (1.), schema evolution allows the addition of new columns, converting NullTypes -> any other type, and upcasts. Currently this is not handled and will cause the a Schema mismatch error. Looking for some guidance on how these scenarios should be handled.

# Todo
- [x] Add file statistics using new writers
- [x] Implement partition column drop on partition write
- [ ] Determine criteria for when optimize should fail 
- [x] idempotent tests


Blajda:
Hi @houqp, let me know if there's anything additional you would like to see in the PR and would appreciate any feedback with the issues items. I'll work on finishing this up by adding statistics on writes.

roeap:
Thanks @Blajda - this is a great update to delta-rs. Personally I think it would be good to re-use the existing writers and fix/extend them where they fall short for the optimize scenario. At least the `RecordBatchWriter` _should_ handle dropping columns that are represented as partition folders within the `divide_by_partitions` function. There are some tests for that, but if they don't that's a bug we should definitely fix. They also handle the file naming and computing file statistics. That way we don't duplicate that logic.

Right now do not handle muti-part files or desired file sizes, which might be use case here. I.e. we could extend the writer to either eagerly flush files when the size of the in-memory writer reaches the desired files, or create a new underlying arrow writer and flush only when flush is called. In case of multiple files, these could be the multi-part files.

A downside of the current implementation is, that we always try to partition the data written into the writer, which might be a fairly expensive operation when its not needed - not sure however how it will behave, if we don't have the partition columns present, since there is nothing to partition by :).

We could try and use the `PartitionWriter` directly, but the logic for add actions / statistics is currently part of the higher level writer - i think.

houqp:
very cool @Blajda ! I agree with  @roeap the code reuse suggestion.

>  Currently the writer's provided by delta-rs write partition information to the parquet file. This differs from the Databricks implementation which does not. This causes a schema mismatch to occur when packing files from different writers. This is currently handled by dropping partition columns when rewriting.

Ideally, we should also not write out partition columns into the actual parquet file. I think this is an oversight from our end. User of delta-rs should be responsible for populating the partition columns. We could track it as a follow up clean up task.

> Similar to (1.), schema evolution allows the addition of new columns, converting NullTypes -> any other type, and upcasts. Currently this is not handled and will cause the a Schema mismatch error. Looking for some guidance on how these scenarios should be handled.

As far as I know, there is no easy way around this other than we need to manually compare the table schema with the schema we got after reading from the parquet file and backfill new columns in the arrow record batches before writing them out to the new parquet file.

roeap:
Just had a quick look into the RecordbatchWriter. To me it seems we do make sure the partition columns are not part of the data written out to storage.

https://github.com/delta-io/delta-rs/blob/5b2a09c59e1fa18f38c275475e41c2148ae02b5e/rust/src/writer/record_batch.rs#L184-L194

We use that partitioned schema to initialize the PartitonWriter and validate the Schema of data written into the PartitionWriter. Then again, I might have missed something, or there could be a bug ...

Blajda:
Hi @roeap,
I originally wrote this at the start of April with the original JSON writer and I haven't tested that scenario with the new writers. Glad to hear it was resolved and sorry about the confusion. The protocol link was helpful, I was trying determine if `default` values are supported when adding columns and this confirms that `null` for back-filling is acceptable. 

I want to avoid the usage of `divide_by_partition_values` since the sort is not required in this case but I do agree there are utilities such as `next_data_path` which can be reused.

Next steps for me is factoring out the drop partition columns functionality and creating some utility for back-filling.

roeap:
@Blajda - what do you think about adding a `write_partition` method to the Writer trait and essentially carving out the inner loop over partitions in the current write method? It would then be great to have the writer aware of the desired file size as well, since right not it is not. In the case of optimize you do have more prior knowledge about how to bin the files, that the optimize command can already leverage - but the general logic would be a great addition to the writer.

That way you get the create add (foremost stats) for free. More importantly though, as we move to support higher writer versions we will need to handle things like column alias / renames, column invariants, identity- and calculated columns etc. which I suspect will be non trivial logic. By leveraging a single writer struct I feel our lives might be a lot easier if we have a single code path that writes out data.

Blajda:
> what do you think about adding a `write_partition` method to the Writer trait and essentially carving out the inner loop over partitions in the current write method?

Yes that sounds like the right approach to ensure the writer's are unified. I assume the signature would look something like this.
```rust
async fn write_partition(&mut self, values: RecordBatch, partition: &?) -> Result<(), DeltaWriterError>
```
Now it's mostly a decision on how to represent the full partition path. We can reuse `get_partition_key` and have `?` be a String. My only concern is that it might be bit confusing for new users. Maybe we can make a new type called `PartitionPath` which can be created in the same way as `get_partition_key` and simply encapsulates a String.

I think that abstraction would be very helpful and would also help cleanup the bin packing implementation.

Let me know what you think

rust/src/optimize.rs
```diff
@@ -0,0 +1,464 @@
...
+            total_files: 0,
+            total_size: 0,
+        }
+    }
+}
+
+///TODO: Optimize
+#[derive(Default)]
+pub struct Optimize<'a> {
+    filters: &'a [PartitionFilter<'a, &'a str>],
```
wjones127, line 77: Should we also have `max_file_size` here? In case the user wishes to pass the value in directly?

rust/tests/optimize_test.rs
```diff
@@ -0,0 +1,91 @@
+#[cfg(feature = "datafusion-ext")]
+mod optiize {
```
wjones127, line 2: I think you may want a few more tests. Here are some properties we likely want to test for:

 1. Optimize is idempotent. If you run twice, the second time it won't write any new files.
 2. Optimize [bin packs](https://en.wikipedia.org/wiki/Bin_packing_problem). For example, with a max file size of 100MB, files of size 70MB, 100MB, 30MB will turn into two files of 100MB, regardless of the order they show up in the log.
 3. Optimize fails if a concurrent writer overwrites the tables. It might be able to succeed if a concurrent writer appends to the table, at the very least in the case where the append happens in a different partition.

wjones127 (commented):
It would be nice if `write_partition()` took a stream of record batches instead of a single one. Then optimize command would bin-pack to collect a set of record batch streams, and pass those streams on to `write_partition()` (one at a time, or maybe even in parallel). The function would then be responsible for merging the schemas. Also need to think about how the optimize metrics are passed back; maybe the `write_partition()` function returns metrics?

Left some other general comments as well.

Blajda:
I'm ready for another review of this PR.

For the scope of this work I did not implement backfill so it expected to fail when called on tables that had schema evolution. I got blocked when handling the evolution of Arrow Lists and had trouble determining the best way forward. 

This is expected to be used with tables that only perform appends. Support for different writing patterns would only require a check for remove actions on the impacted partitions so I think we are safe to ignore add actions to a partition. Once we have pattern for performing those check it should be small change.

@roeap I first tried adding a write_partition to the Writer trait but I wasn't completely sure what to do for the JSON writer case. For now I've only added it to the record batch writer and ensured that partitioned related columns are dropped when called. Let me know what you think and if anything else should be changed.

rust/src/optimize.rs
Blajda: Done. A user provided target size can now be passed in this builder. Otherwise the table config or default is used.

rust/tests/optimize_test.rs
Blajda: I changed the implementation to sort files to be optimized to ensure that it is idempotent. Added a couple of tests to validate that behavior too. For item 3 I think it will have to wait until we have a generalized pattern for non-append writers.

rust/src/optimize.rs
```diff
@@ -0,0 +1,356 @@
...
+                    self.metrics.files_added.total_files += 1;
+                    self.metrics.files_added.total_size += size;
+                    self.metrics.files_added.max =
+                        std::cmp::max(self.metrics.files_added.max, size);
+                    self.metrics.files_added.min =
+                        std::cmp::min(self.metrics.files_added.min, size);
+                    actions.push(action::Action::add(add));
+                }
+            }
+            //Currently a table without any partitions has a count of one partition. Check if that is acceptable
```
wjones127, line 234: That seems logical to me 👍

rust/src/optimize.rs
```diff
@@ -0,0 +1,356 @@
...
+                    self.metrics.files_removed.total_size += size;
+                    self.metrics.files_removed.max =
+                        std::cmp::max(self.metrics.files_removed.max, size);
+                    self.metrics.files_removed.min =
+                        std::cmp::min(self.metrics.files_removed.min, size);
+                }
+
+                //Save the file to storage and create corresponding add and delete actions. Do not commit yet.
+                let add_actions = writer.flush().await?;
+                for mut add in add_actions {
```
wjones127, line 220: In theory, we should be only getting one add action here, correct? Wondering if we've checked that.

rust/src/optimize.rs
```diff
@@ -0,0 +1,356 @@
...
+                        std::cmp::min(self.metrics.files_added.min, size);
+                    actions.push(action::Action::add(add));
+                }
+            }
+            //Currently a table without any partitions has a count of one partition. Check if that is acceptable
+            self.metrics.partitions_optimized += 1;
+        }
+
+        //try to commit actions to the delta log.
+        //Check for remove actions on the optimized partitions
```
wjones127, line 239: Looks like this needs to be marked TODO, right?

rust/src/writer/utils.rs
```diff
@@ -165,3 +193,36 @@ pub(crate) fn stringified_partition_value(
...
+            .fields()
+            .iter()
+            .filter(|f| !partition_columns.contains(f.name()))
+            .map(|f| f.to_owned())
+            .collect::<Vec<_>>(),
+    ))
+}
+
+/// Remove any partition related columns from the record batch
+pub(crate) fn record_batch_without_partitions(
```
wjones127, line 213: Something to consider: It might be simpler to implement this using `RecordBatch.project()` ([docs](https://docs.rs/arrow/latest/arrow/record_batch/struct.RecordBatch.html#method.project)).

rust/tests/optimize_test.rs
```diff
@@ -0,0 +1,344 @@
...
+    use serde_json::{Map, Value};
+    use std::{collections::HashMap, error::Error, sync::Arc};
+    use tempdir::TempDir;
+    struct Context {
+        pub tmp_dir: TempDir,
+        pub table: DeltaTable,
+    }
+
+    async fn setup_test(partitioned: bool) -> Result<Context, Box<dyn Error>> {
+        let schema = Schema::new(vec![
```
wjones127, line 29: Maybe worth saving this schema as a constant and then calling `try_into()` on it below to get the equivalent Arrow schema?

rust/tests/optimize_test.rs
```diff
@@ -0,0 +1,344 @@
...
+
+        let optimize = Optimize::default().filter(&filter).target_size(10_000_000);
+        let metrics = optimize.execute(&mut dt).await?;
+        assert_eq!(metrics.num_files_added, 1);
+        assert_eq!(metrics.num_files_removed, 2);
+
+        let metrics = optimize.execute(&mut dt).await?;
+
+        assert_eq!(metrics.num_files_added, 0);
+        assert_eq!(metrics.num_files_removed, 0);
```
wjones127, line 301: Maybe also assert that `dt.version` hasn't changed?

rust/tests/optimize_test.rs
```diff
@@ -0,0 +1,344 @@
...
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    ///Validate Metrics when no files are optimzied
```
wjones127, line 307:
```suggestion
    /// Validate Metrics when no files are optimized
```

rust/src/optimize.rs
```diff
@@ -0,0 +1,356 @@
...
+        self
+    }
+
+    ///Set the target file size
+    pub fn target_size(mut self, target: DeltaDataTypeLong) -> Self {
+        self.target_size = Some(target);
+        self
+    }
+
+    ///Perform the optimization. On completion, a summary of how many files were added and deleted is returned
```
wjones127, line 105: nit: It's probably clearer to say "removed" rather than "deleted", to make clear the underlying files aren't deleted.
```suggestion
    /// Perform the optimization. On completion, a summary of how many files were added and removed is returned
```

rust/src/optimize.rs
```diff
@@ -0,0 +1,356 @@
+//! Optimize a Delta Table
+//!
+//! Bin-packing is performed which merges smaller files into a larger file. This
+//! reduces the number of api calls required to read data and allows for the
+//! optimized files to cleaned up by vacuum.
```
wjones127, line 5: Two additional things worth mentioning in the docs, IMO:

 1. This increments the table version and removes files, but it does NOT delete old files. To delete old files use vacuum.
 2. It is safe to run concurrently with append-only workloads, but not yet other ones.

rust/src/optimize.rs
```diff
@@ -0,0 +1,356 @@
...
+    ///Set the target file size
+    pub fn target_size(mut self, target: DeltaDataTypeLong) -> Self {
+        self.target_size = Some(target);
+        self
+    }
+
+    ///Perform the optimization. On completion, a summary of how many files were added and deleted is returned
+    pub async fn execute(&self, table: &mut DeltaTable) -> Result<Metrics, DeltaWriterError> {
+        let plan = create_merge_plan(table, self.filters, (&self.target_size).to_owned())?;
+        let metrics = plan.execute(table).await?;
```
wjones127, line 108: I like this split of planning and then executing. If you are up to it, you should be able to use this to test concurrent writes (just need to make `create_merge_plan` `pub(crate)`.) Or maybe at least leave a TODO for that.

```rust
let plan = create_merge_plan(table, self.filters, (&self.target_size).to_owned())?;
// do a write to table
let metrics_res = plan.execute(table).await;
// assert success or failure
```

wjones127 (changes requested):
Great progress! I appreciate the additional tests ❤️ 

I left a few suggestions on documentation and testing. Given a few minor changes, I think this should be ready to merge.

rust/tests/optimize_test.rs
```diff
@@ -0,0 +1,344 @@
...
+        let backend = get_backend_for_uri_with_options(&p, HashMap::new())?;
+        let mut dt = DeltaTable::new(&p, backend, DeltaTableConfig::default())?;
+        let mut commit_info = Map::<String, Value>::new();
+
+        let protocol = action::Protocol {
+            min_reader_version: 1,
+            min_writer_version: 2,
+        };
+
+        commit_info.insert(
```
roeap, line 77: Seems like we are not using this commit info?

roeap:
Looking good!

Left some minor questions. Reading this I realized that we should probably treat other operations - like vacuum - the same as here. I.e. having it as a separtae struct that can be executed ... But this is something for a follow up.

rust/src/optimize.rs
```diff
@@ -0,0 +1,356 @@
...
+            }
+            //Currently a table without any partitions has a count of one partition. Check if that is acceptable
+            self.metrics.partitions_optimized += 1;
+        }
+
+        //try to commit actions to the delta log.
+        //Check for remove actions on the optimized partitions
+        let mut dtx = table.create_transaction(None);
+        dtx.add_actions(actions);
+        dtx.commit(None, None).await?;
```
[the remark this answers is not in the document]
Blajda: Added a new struct and associated tests for this.

rust/tests/optimize_test.rs
Blajda: Fixed!

rust/src/optimize.rs
Blajda: `pub(crate)` didn't work so I made them public. I don't have an issue with people calling that function directly.

Added two tests. The first checks concurrent writes and the second which is currently marked as ignored checks for failure on concurrent removes.

rust/src/lib.rs
```diff
@@ -93,8 +93,7 @@ pub mod data_catalog;
 mod delta;
 pub mod delta_arrow;
 pub mod delta_config;
-#[cfg(feature = "datafusion-ext")]
-pub mod operations;
```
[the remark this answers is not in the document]
Blajda: My Bad. Had an issue with using git cherry pick.

rust/tests/optimize_test.rs
Blajda: I think this is okay for now. We can always pull it out if we want to reuse it in the future.

Blajda:
Hi @roeap @wjones127 
I've implemented the suggested changes and add additional tests. Let me know what you think! :smiley:

rust/tests/optimize_test.rs
```diff
@@ -0,0 +1,506 @@
...
+        let commit_metrics =
+            serde_json::from_value::<Metrics>(last_commit["operationMetrics"].clone())?;
+
+        assert_eq!(commit_metrics, metrics);
+        assert_eq!(last_commit["readVersion"], json!(version));
+        assert_eq!(
+            last_commit["operationParameters"]["targetSize"],
+            json!(2_000_000)
+        );
+        assert_eq!(last_commit["operationParameters"]["predicate"], Value::Null);
```
wjones127, line 502: Should predicate equal the filter used above?

wjones127 (changes requested):
Thanks for adding those concurrency tests!

I just have one question on that commit info test.

rust/src/writer/utils.rs
```diff
@@ -165,3 +193,33 @@ pub(crate) fn stringified_partition_value(
 
     Ok(Some(s))
 }
+
+/// Remove any partition related fields from the schema
+pub(crate) fn schema_without_partitions(
```
wjones127, line 198: Also, do you still need this function?

Blajda: Nope. Reverted and restored the original function from where it was sourced.

rust/tests/optimize_test.rs
Blajda: Yes it should. I added an additional TODO here. There's isn't a function that obtains the String representation of PartitionFilters. Should be fairly simple to implement.

wjones127: Okay. That seems sufficient for now.

wjones127 (approved):
Looks good to me. I'll wait a few days before merging to give others a chance to provide any final feedback.

Great job on this!

rust/src/writer/utils.rs
```diff
@@ -12,31 +12,59 @@ use arrow::{
...
+    ) -> Result<Self, DeltaWriterError> {
+        let mut path_parts = vec![];
+        for k in partition_columns.iter() {
+            let partition_value = partition_values
+                .get(k)
+                .ok_or_else(|| DeltaWriterError::MissingPartitionColumn(k.to_string()))?;
+
+            let partition_value = partition_value
+                .as_deref()
+                .unwrap_or(NULL_PARTITION_VALUE_DATA_PATH);
```
roeap, line 40: Just realized that this crate traditionally uses the constant, but the protocol specifies a different behaviour when it comes to nulls. I opened a ticket to track this question #619 - since it was also not introduced here...

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization

roeap (approved):
LGTM - great work @Blajda!

Blajda:
Thanks @wjones127 and @roeap  for helping carry this to completion.

houqp:
Amazing work on this @Blajda !

PR #618, opened 2022-06-02
fix: Use relative paths for add paths

### LINKED ISSUE #617: Cannot read Delta Lake created with write_deltalake (Python bindings)
# Environment

**Delta-rs version**: 0.5.7

**Binding**: Python

**Environment**:
- **Cloud provider**: N/A (using localhost)
- **OS**: MacOS

***
# Bug

**What happened**: I created a Delta Lake using delta-rs with `write_deltalake` and was unable to read the Delta Lake using delta-rs

**What you expected to happen**: I expected the read operation to work

**How to reproduce it**: 
 
```python
import pandas as pd

# create Delta Lake
df = pd.DataFrame({"x": [1, 2, 3]})
write_deltalake("./tmp/delta-table", df)

# read Delta Lake
dt = DeltaTable("./tmp/delta-table")
dt.to_pandas()
```

Here's the error message: 

```
File ~/opt/miniconda3/envs/mr-delta-rs/lib/python3.9/site-packages/deltalake/fs.py:149, in DeltaStorageHandler.open_input_file(self, path)
    142 def open_input_file(self, path: str) -> pa.NativeFile:
    143     """
    144     Open an input file for random access reading.
    145 
    146     :param source: The source to open for reading.
    147     :return:  NativeFile
    148     """
--> 149     raw = self._storage.get_obj(path)
    150     return pa.BufferReader(pa.py_buffer(raw))

PyDeltaTableError: Object not found
```

**More details**: You can easily reproduce this by creating [this conda environment](https://github.com/MrPowers/delta-examples/blob/master/envs/mr-delta-rs.yml) and running [this Jupyter notebook](https://github.com/MrPowers/delta-examples/blob/master/notebooks/delta-rs/01_overview.ipynb).  

Other operations like `dt.files()` and `dt.version()` work as expected.  I am able to read Delta Tables created by [delta-io/delta](https://github.com/delta-io/delta) without any issue.

wjones127:
# Description

We weren't trimming the paths before creating add actions. When we used an absolute `table_uri`, we used absolute paths. When we used a relative `table_uri`, we would duplicate the root relative path in the add actions, making the table unreadable.

# Related Issue(s)

Fixes #617.

# Documentation

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -12,6 +12,7 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Tuple,
     Union,
 )
 
@@ -173,12 +174,12 @@ def write_deltalake(
     add_actions: List[AddAction] = []
 
     def visitor(written_file: Any) -> None:
-        partition_values = get_partitions_from_path(table_uri, written_file.path)
+        path, partition_values = get_partitions_from_path(table_uri, written_file.path)
         stats = get_file_stats_from_metadata(written_file.metadata)
 
         add_actions.append(
             AddAction(
-                written_file.path,
+                path,
                 written_file.metadata.serialized_size,
                 partition_values,
                 int(datetime.now().timestamp()),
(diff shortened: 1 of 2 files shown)


houqp (approved):

MrPowers:
Thanks for the fix @wjones127!  I will double check on my end when the next version is published to pypi.

Issue #620, opened 2022-06-03
write_deltalake fails with error: thread 'tokio-runtime-worker' panicked at 'not implemented'


jkausti:
# Environment

**Delta-rs version**: 0.5.7

**Binding**: python

**Environment**:
- **OS**: Ubuntu 20.04 running in WSL2 in Windows 10
- **Other**: Python 3.8

***
# Bug

**What happened**:
Error message when writing a new deltatable to local filesystem.

I have the following code:
```
import pandas as pd
import deltalake as dl

data = [("a", 1), ("b", 2)]
df = pd.DataFrame(data)
dl.write_deltalake(data=df, table_or_uri="tables/my_table")
```
When run, it generates the following error message:

```
thread 'tokio-runtime-worker' panicked at 'not implemented', rust/src/storage/file/rename.rs:104:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(...)', rust/src/storage/file/rename.rs:72:14
Traceback (most recent call last):
  File "main.py", line 204, in <module>
    main()
  File "main.py", line 20, in main
    dl.write_deltalake(data=df, table_or_uri="tables/my_table")
  File "/home/john/projects/deltalake-test/.venv/lib/python3.8/site-packages/deltalake/writer.py", line 158, in write_deltalake
    _write_new_deltalake(  # type: ignore[call-arg]
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: JoinError::Panic(...)
```

However, the delta table gets created even though the error message appears.

**What you expected to happen**:
No error message appears.

**How to reproduce it**:
Run the above code.

**More details**:
I checked the code in [rename.rs](https://github.com/delta-io/delta-rs/blob/8a0475ce14841296b3d68bb962a8cbe5e0fcf2ec/rust/src/storage/file/rename.rs) that the error message refers to and it seems as though the 'not implemented' error happens because of old glibc version (as per the comment on line 103 in rename.rs). However, when running "ldd --version" on my machine, it says my glibc version is 2.31.

wjones127:
Haven't confirmed, but I suspect this is because our Linux wheels have been built in an environment with old glibc for compatibility reasons. I think we'll need to upgrade to manylinux2014, or change that function (I think there might actually be a better approach).

As a workaround, this *should* work if you build from source (annoying, I know):

```bash
[...]

Issue #610, opened 2022-05-20
Adopt influxdata/object_store_rs?


wjones127:
# Description

This crate maintains its own notion of `StorageBackend`, but it might be more than we need to maintain for Delta Lake support.

[influxdata/object_store_rs](https://github.com/influxdata/object_store_rs/) implements an `ObjectStore` trait that covers the basics we need and provides a point of collaboration with the rest of the Rust data ecosystem. It looks likely that DataFusion will also adopt this crate eventually instead of it's internal `ObjectStore` trait, which is less complete (see apache/arrow-datafusion#2489).

If ObjectStore exposed the underlying client, it seems like we could simply make StorageBackend depend on that trait and only have to maintain the `rename_obj_noreplace()` method:

```rust
#[async_trait::async_trait]
pub trait StorageBackend: Send + Sync + Debug + ObjectStore {
    async fn rename_obj_noreplace(&self, src: &str, dst: &str) -> Result<(), StorageError>;
}
```

What do we think of that?

roeap:
I think this is a great idea. As you said in a first step it would reduce complexity in this crate. When adopted by datafusion. it would also greatly simplify the datafusion interop in the TableProvider and the operations.

One thing to mention is that right now it does not support ADLS Gen2 which we need for `rename_obj_noreplace` unless we also want to build a lock client for Azure. But I'd be happy to port the Gen2 support over there.

houqp:
+1 for me to push the rust data eng community towards a single feature rich and highly optimized object store crate!

wjones127:
We still need to add the dynamodb lock integration to object_store_rs before it's ready for us to start integrating. I suspect we will have to write some adapters to the object stores to instantiate them with the various configuration options we have. Anything else we need to do?

One important difference between our implementation and object_store_rs is that there is that it implements `rename_if_not_exists` with `copy_if_not_exists` and `delete` for all stores. The only ones we implement as a single call are local file system and Azure. They seem to be reluctant to add the local filesystem calls because they are so complex and OS-dependent; but if a strong case could be made they make a material difference on performance then we can likely add them.

In the Azure case, I think if we keep as two calls we can drop the requirement for ADLS2 and just support Blob store generally, right? Would we see that as an improvement? cc @roeap @thovoll

roeap:
> but if a strong case could be made they make a material difference on performance then we can likely add them.

[...]

PR #621, opened 2022-06-04
feat: impl rename_noreplace with std::fs::hard_link by default

wjones127:
# Description

I've been working in influxdata/object_store_rs, with the goal of switching to that object store implementation in delta-rs (see #610). It was suggested to [simply use hard_link](https://github.com/influxdata/object_store_rs/pull/14#discussion_r888154559) to implement `rename_no_replace()` easily cross platform.

Since this is causing issues in the Python writer right now (see #620), I'm thinking maybe I should patch this change immediately.

This PR provides a default implementation of `rename_noreplace()` using `std::fs::hard_link`, which should be supported on all platforms. It also enables Python writer tests for all platforms.

# Related Issue(s)

 * #610 
 * #620 
 * influxdata/object_store_rs#14

# Documentation


wjones127:
All tests are passing, including Python on old glibc (💯).

I'm throwing out a fair amount of code in this though, so I'd welcome any critical feedback.

rust/src/storage/file/rename.rs
```diff
@@ -1,123 +1,27 @@
...
+                err.into()
+            }
+        })?;
+
+        std::fs::remove_file(from_path)?;
+
+        Ok(())
+    })
+    .await
+    .unwrap()
```
houqp, line 24: should we propagate the io error back to the caller instead?

houqp (commented):
Great idea @wjones127! My current thinking on this is we could use hardlink as the fallback when native atomic rename is not available. Here is my reasoning for keeping renameat2:

1. Hardlink followed by remove is not atomic, so in very rare edge cases, we could leave the old path behind
2. It is less efficient compared to renameat2 because we need to issue 2 system calls instead of just one.

Both of the above shortcomings probably won't matter much in most of the production workloads because deltalake is designed for low frequency commits, but I think the tradeoff is worth it in this case because the added complexity is manageable (about 200+ LoC). Happy to hear your thought on this as well.

wjones127:
> 1. Hardlink followed by remove is not atomic, so in very rare edge cases, we could leave the old path behind

This is already true for the S3 and GCS, right? So only Azure and local filesystems support this operation with that kind of atomicity. I think object_store_rs is aiming to support operations that are common across object stores, with similar guarantees across them. Why enforce that atomicity for some but not others?

And Windows is not atomic in the way we want right now; it does a check for a file and then runs replace, which could overwrite if there is a concurrent writer. So I think `hard_link` is an improvement in that case, at least.

> 2. It is less efficient compared to renameat2 because we need to issue 2 system calls instead of just one.

True, but we only run this operation once per a commit, so it's doubtful there's a performance hit in the big picture. 

In object_store_rs, I implemented `copy_if_not_exists` instead for these reasons. At the time, [I thought that](https://github.com/influxdata/object_store_rs/pull/14#discussion_r888475372) no (non-local) object store actually implements replace, but seeing that Azure *does* makes me rethink that a bit. I don't think it's critical that delta-rs has this; but there are marginal advantages to reliability and performance you've mentioned that could be implemented for local and Azure backends.

So maybe for time being, I'll do this just for Windows and old glibc Linux, and we'll think more about whether we really need this for other platforms.

houqp:
> This is already true for the S3 and GCS, right? So only Azure and local filesystems support this operation with that kind of atomicity. I think object_store_rs is aiming to support operations that are common across object stores,  Why enforce that atomicity for some but not others?

This is what I have in mind as well. Just to clarify, I wasn't suggesting we add this to object_store_rs. I was suggesting we keep it within delta-rs assuming the maintenance burden is low. If that's not the case, I am cool with us switching to hardlink everywhere.

Like I mentioned in my previous comment, I don't think the overhead and correctness issue would matter in production. I am only suggested keeping it because I would like delta-rs to adhere to Rust and C++'s zero cost abstraction philosophy as much as we could.

I am also not the one who is writing the code, so I won't be blocking the merge for minor nitpicks like this :D

> And Windows is not atomic in the way we want right now; it does a check for a file and then runs replace, which could overwrite if there is a concurrent writer. So I think hard_link is an improvement in that case, at least.

100% agree, I was trying to suggest that we use renameat2 if it's available, otherwise fallback to hardlink+remove. The current windows implementation is not multi-writer safe as mentioned in the fs module's documentation.

> In object_store_rs, I implemented copy_if_not_exists instead for these reasons. At the time, https://github.com/influxdata/object_store_rs/pull/14#discussion_r888475372 no (non-local) object store actually implements replace, but seeing that Azure does makes me rethink that a bit. I don't think it's critical that delta-rs has this; but there are marginal advantages to reliability and performance you've mentioned that could be implemented for local and Azure backends.

The performance trade offs between PUT if absent and atomic rename is quite nuance. My current understanding of this is that for non-append only and single writer append only commits, PUT if absent is more performant and correct because it only requires a single API call. For highly concurrent append only commits, atomic rename is more performant because it avoids repetitive object writes on subsequent retries. Because delta lake is not designed for highly current write use-cases, I think we should go with PUT if absent whenever possible. This means even though both Azure and HDFS support atomic rename, we probably shouldn't be using it.

Therefore, I think we should probably rewrite the commit loop in delta-rs to use put if absent instead.

wjones127:
> Therefore, I think we should probably rewrite the commit loop in delta-rs to use put if absent instead.

Good points there; I haven't looked to far into that yet, but I might try that when writing the PR to switch to object_store_rs.

houqp (approved):
LGTM!

Issue #619, opened 2022-06-02
Is delta-rs properly handling partition folder naming?


roeap:
# Description
The protocols specifies to serialize empty partition values as 'null', however internally we use a constant value `__HIVE_DEFAULT_PARTITION__`  

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization

Are we doing the right thing when writing partition folders? 

@houqp @wjones127

wjones127:
Thanks for pointing this out, @roeap! I found I wasn't handling this correctly while I was writing integration tests with PySpark ([PR](https://github.com/delta-io/delta-rs/pull/625)). It seems that we should be writing directories as `part_col=__HIVE_DEFAULT_PARTITION__`, but within the add actions we need to specify `{"partitionValues": {"part_col": null } }`. (In the PyArrow writer, I was filling in the add action with the string `"__HIVE_DEFAULT_PARTITION__"` 🤦 .)

There's an example table that demonstrates this in the repo at `rust/tests/data/delta-0.8.0-null-partition`.

houqp:
@roeap @wjones127 i think we can close this issue now?

roeap:
Fixed in #625

Issue #579, opened 2022-03-23
Python: Create PySpark integration tests for reading/writing


wjones127:
# Description

The particular types used by Spark's parquet differ from PyArrow, so we should have integration tests to make sure we are compatible with them.

**Related Issue(s)**
Inspired by:
 * #578

rtyler:
We actually have these now sorta! wheee

PR #44, opened 2020-12-17
Bump the arrow dependencies to master for the ongoing development that's landing there

rtyler:
This has the nice side-effect of making delta.rs no longer require the nightly
compiler

"But Tyler" you say "the master branch is always changing!" To which I grumpily
reply "Cargo dot lock"

(hat tip to @andygrove for the mention that arrow now compiles with stable)

PR #625, opened 2022-06-06
fix: Add correct size and null paritition values to add actions

wjones127:
# Description

Fixes two bugs:

 1. The `size` field of add actions was always being set to zero. This PR corrrect that to always use the correct file size.
 2. Null partition values were being recorded in add actions with the string `"__HIVE_DEFAULT_PARTITION__"`. They are now correctly recorded as JSON `null` values. 

Other changes in service of fixing those bugs:

 1. Added pyspark integration tests that run in CI. This is what allowed me to catch the bugs. Right now it just tests that the output of our writer is readable in Spark. Our existing tests show that we can read their tables. We can't write to each others tables yet, because we don't support the same writer versions. (But maybe I should double check we are enforcing that 🤔 .)
 2. Added `size` field to `ObjectMeta`, so we can use `head_object()` to get file sizes to solve bug (1). In the future, I hope PyArrow will provide the file sizes in the visitor so we don't have to make these extra API calls. (object_store_rs already has this field in it's [ObjectMeta struct](https://github.com/influxdata/object_store_rs/blob/b98d6b5015cebc292f6174d06cfb5a424999bcc9/src/lib.rs#L123-L130).)

# Related Issue(s)

 * Related to #619.
 * This starts work on #579.

# Documentation

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -88,15 +88,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - uses: actions/cache@v2
-      with:
-        path: |
-          ~/.cargo/bin/
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: python-test-${{ hashFiles('**/Cargo.lock', '**/Cargo.toml') }}
+    - uses: Swatinem/rust-cache@v1
 
     - name: Install latest nightly
       uses: actions-rs/toolchain@v1
@@ -139,15 +131,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - uses: actions/cache@v2
-      with:
-        path: |
-          ~/.cargo/bin/
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: python-pyspark-test-${{ hashFiles('**/Cargo.lock', '**/Cargo.toml') }}
+    - uses: Swatinem/rust-cache@v1
 
     - name: Install latest nightly
       uses: actions-rs/toolchain@v1
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -132,3 +132,46 @@ jobs:
       run: |
         source venv/bin/activate
         make build-documentation
+
+  test-pyspark:
+    name: PySpark Integration Tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: python-pyspark-test-${{ hashFiles('**/Cargo.lock', '**/Cargo.toml') }}
+
+    - name: Install latest nightly
+      uses: actions-rs/toolchain@v1
+      with:
+          toolchain: stable
+          override: true
+          components: rustfmt, clippy
+
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.10' 
+    
+    - uses: actions/setup-java@v1
+      with:
+        java-version: '11'
+
+    - name: Build and install deltalake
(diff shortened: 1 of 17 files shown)


wjones127:
I have created an upstream issue to track the file sizes in PyArrow. But for now, I'll have to just use the object store to query the file sizes after the fact.

rust/src/storage/azure/mod.rs
```diff
@@ -390,22 +391,9 @@ async fn list_obj_future(
...
-            .map_err(to_storage_err2)?;
-
-        let modified = properties.last_modified;
-        Ok(ObjectMeta {
-            path: path.to_string(),
-            modified,
-        })
+        // Only list endpoint currently provides size
+        let mut objs = self.list_objs(path).await?;
+        objs.next().await.unwrap_or(Err(StorageError::NotFound))
```
wjones127, line 396: @roeap I don't have an Azure account yet, so haven't tested. But what do you think? I couldn't see any way to get the size from the properties.

roeap: Unfortunately we have also another bug in the azure sdk regarding the list operation. 
I have a PR open for that and the issue you raised Azure/azure-sdk-for-rust#798. Luckily just in time to not be included in the latest release :).

We have #644 open based on my branch of the azure SDKs which fixes there things. I also asked them to do a patch release once its merged.

Probably easiest is to get #44 merged which would give you the size from the property call.

python/tests/test_writer.py
```diff
@@ -92,6 +54,13 @@ def test_roundtrip_basic(tmp_path: pathlib.Path, sample_data: pa.Table):
         # Paths should be relative, and with no partitioning have no directories
         assert "/" not in add_path
 
+    for action in get_add_actions(delta_table):
+        path = os.path.join(tmp_path, action["path"])
+        actual_size = os.path.getsize(path)
+        assert action["size"] > 0
+        # TODO: fix this
```
roeap, line 61: is this todo still relevant?

rust/src/storage/gcs/mod.rs
```diff
@@ -41,6 +41,7 @@ impl From<tame_gcs::objects::Metadata> for ObjectMeta {
         ObjectMeta {
             path: metadata.name.unwrap(),
             modified: metadata.updated.unwrap(),
+            size: metadata.size.map(|s| s.try_into().unwrap()),
         }
```
roeap, line 45: this should probably be fallible, but since this was not introduced here, and we are hoping to have object_store_rs soon, this is likely a topic for another day :).

roeap (approved):
Great work! Having spark integration tests established will help us a lot with some of the upcoming work!

Left a few comments that you may or may not address :)

One related thing not in this PR I am not sure about is when we crate add actions in the RecordBatchWriter, we pass the size of the in memory buffer as file site to the add action. Do you you know if this will correspond to the size on disk, or should be do a head request there as well to get consistent sizes?

rust/src/storage/gcs/mod.rs
wjones127: Yeah I agree. I will make a note to look at that in object_store_rs

.github/workflows/python_build.yml
```diff
@@ -132,3 +132,46 @@ jobs:
...
         source venv/bin/activate
         make build-documentation
+
+  test-pyspark:
+    name: PySpark Integration Tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/cache@v2
```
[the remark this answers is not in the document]
wjones127: Yup that looks better.

python/tests/test_writer.py
wjones127: Nope, deleted now. Thanks!

roeap (approved):

wjones127:
> One related thing not in this PR I am not sure about is when we crate add actions in the RecordBatchWriter, we pass the size of the in memory buffer as file site to the add action. Do you you know if this will correspond to the size on disk, or should be do a head request there as well to get consistent sizes?

That should always be the correct size. The unit tests for `head_object` check that the size value is equal to the length of the buffer that was written.

houqp (approved):

houqp:
Thanks @wjones127 ! Really nice to have a pyspark integration test.

PR #632, opened 2022-06-09
feat: optimistic transaction protocol

### LINKED ISSUE #593: Protocol compliant commits for Delta operations
# Description

When committing transactions to the Delta log, make sure we never end up in an inconsistent or corrupt state.

**Use Case**

When performing mutating operations against a Delta table we have to make sure we never corrupt the table / data.

roeap:
# Description

This PR adds a `ConflictChecker` struct for conflict resolution in cases of concurrent commit failures. The implementation is heavily inspired by the [reference implementation](https://github.com/delta-io/delta/blob/fe36a53f3c70c5f9c9b5052c12cd1703f495da97/core/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala). So far we have most tests from spark that specifically target conflict resolution covered.

Working on this I thought a bit about what we may consider going forward, as we move through the protocol versions :).  In the end we could end up with three main structs that are involved in validating a commit.

* The existing `DataChecker`, which validates and potentially mutates data when writing data files to disk. (Currently supports invariants)
* The upcoming `ConflictChecker`, which checks if a commit can be re-tried in case of commit conflicts.
* A new `CommitChecker`, which does a-priory validation of the commit itself (e.g. append only and other rules covered by tests in  [spark](https://github.com/delta-io/delta/blob/master/core/src/test/scala/org/apache/spark/sql/delta/OptimisticTransactionSuite.scala))

My hope is to get this PR merged right after we release `0.8.0`, so there is some time to fill some holes and fully leverage the new feature for `0.9.0`. 

If folks agree, I would open some issues and start work on some follow-ups..

## Follow-ups
* Extend `ConflictChecker` support conflict resolution for streaming transactions
* Implement `CommitChecker`
* Deprecate old commit function.
* Extend `DataChecker`. 
* Consolidate record batch writer implementations.

# Related Issue(s)

part of #593 

# Documentation


houqp:
Thank you @roeap for picking up this work! I will take a closer look at it this weekend. :+1: from me for including datafusion-expr if needed.

houqp:
The structure looks good to me :+1: When we introduce a proper expression handling abstraction, we should upgrade what we have in partition handling with it as well: https://github.com/delta-io/delta-rs/blob/cc62c49c905a0818e5d1edcd1383deb512dc2ac9/rust/src/partitions.rs#L12

rust/src/conflict_checker.rs
```diff
@@ -0,0 +1,431 @@
...
+            IsolationLevel::Serializable | IsolationLevel::WriteSerializable => {
+                let mut files = self.winning_commit_summary.changed_data_added_files();
+                files.extend(self.winning_commit_summary.blind_append_added_files());
+                files
+            }
+            IsolationLevel::SnapshotIsolation => vec![],
+        };
+        // TODO here we need to check if the current transaction would have read the
+        // added files. for this we need to be able to evaluate predicates. Err on the safe side is
+        // to assume all files match
```
wjones127, line 339: What would you think of keeping this behavior if datafusion feature isn't included? That way we could still have the option of not including datafusion. I think there are some use cases where we don't want to require it as a dependency.

roeap: Absolutely - I was contemplating how to do this for a bit and parsing complex predicates would certainly benefit from some more advanced parsing. However right now we only have min max and nulls anyhow, which the current predicate parsing should be able to parse as well.

When we apply the filters my feeling is that we want to go the arrow route - i.e. put all files with bounds in  a record batch and use arrow kernels. I'll see how far I get using "just" arrow and as you said, falling back to the most conservative case is always an option.

rust/src/conflict_checker.rs
```diff
@@ -0,0 +1,459 @@
...
+    table_state::DeltaTableState, DeltaDataTypeTimestamp, DeltaDataTypeVersion, DeltaTable,
+    DeltaTableError, DeltaTableMetaData,
+};
+use serde::{Deserialize, Serialize};
+use serde_json::{Map, Value};
+use std::collections::{HashMap, HashSet};
+use std::io::{BufRead, BufReader, Cursor};
+
+/// The commitInfo is a fairly flexible action within the delta specification, where arbitrary data can be stored.
+/// However the reference implementation as well as delta-rs store useful information that may for instance
```
andrei-ionescu, line 13: Please add a link to the official documentation on the Delta Lake possible conflicts.

I did find this one — https://docs.databricks.com/optimizations/isolation-level.html#write-conflicts-on-databricks — but I'm not sure how complete it is.

rust/src/conflict_checker.rs
```diff
@@ -0,0 +1,459 @@
...
+    pub isolation_level: Option<IsolationLevel>,
+    pub is_blind_append: Option<bool>,
+}
+
+/// Exceptions raised during commit conflict resolution
+#[derive(thiserror::Error, Debug)]
+pub enum CommitConflictError {
+    /// This exception occurs when a concurrent operation adds files in the same partition
+    /// (or anywhere in an unpartitioned table) that your operation reads. The file additions
+    /// can be caused by INSERT, DELETE, UPDATE, or MERGE operations.
```
andrei-ionescu, line 34: I don't think `INSERT` / `INSERT` can result in this error. As far as I can understand from https://docs.databricks.com/optimizations/isolation-level.html#write-conflicts-on-databricks.

Maybe it will be useful to create that matrix from the given link and add it somewhere in this PR maybe on the enum explanation.

rust/src/conflict_checker.rs
```diff
@@ -0,0 +1,459 @@
...
+        let winning_commit_summary =
+            WinningCommitSummary::new(commit_actions, winning_commit_version);
+
+        let initial_current_transaction_info = CurrentTransactionInfo::try_new(table, operation)?;
+
+        Ok(Self {
+            initial_current_transaction_info,
+            winning_commit_summary,
+            winning_commit_version,
+            isolation_level: IsolationLevel::Serializable,
```
andrei-ionescu, line 289: Why forcing the `IsolationLevel::Serializable`? What about the other isolation levels?

rust/src/delta.rs
```diff
@@ -1319,6 +1326,7 @@ pub struct DeltaTransaction<'a> {
     delta_table: &'a mut DeltaTable,
     actions: Vec<Action>,
     options: DeltaTransactionOptions,
+    txn_id: String,
```
andrei-ionescu, line 1329: Should there be some explanation added regarding `txn_id` in the docs above?

rust/src/delta.rs
```diff
@@ -1349,32 +1358,51 @@ impl<'a> DeltaTransaction<'a> {
     /// This method will retry the transaction commit based on the value of `max_retry_commit_attempts` set in `DeltaTransactionOptions`.
     pub async fn commit(
         &mut self,
-        operation: Option<DeltaOperation>,
+        operation: DeltaOperation,
```
andrei-ionescu, line 1361: Changing this public method signature will result in backward compatibility issues. Any concerns? Is backward compatibility offered in the APIs?

rust/src/delta.rs
```diff
@@ -1349,32 +1358,51 @@ impl<'a> DeltaTransaction<'a> {
...
         app_metadata: Option<Map<String, Value>>,
     ) -> Result<DeltaDataTypeVersion, DeltaTableError> {
-        // TODO: stubbing `operation` parameter (which will be necessary for writing the CommitInfo action),
-        // but leaving it unused for now. `CommitInfo` is a fairly dynamic data structure so we should work
-        // out the data structure approach separately.
+        // TODO(roeap) in the reference implementation this logic is implemented, which seems somewhat strange,
```
andrei-ionescu, line 1364: In the master the default isolation level seems to be retrieved from table metadata: https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala#L1213

rust/src/delta.rs
```diff
@@ -1384,7 +1412,7 @@ impl<'a> DeltaTransaction<'a> {
     /// with `DeltaTable.try_commit_transaction`.
     pub async fn prepare_commit(
         &mut self,
-        operation: Option<DeltaOperation>,
+        operation: &DeltaOperation,
```
andrei-ionescu, line 1415: Ditto... This is a change that will break backward compatibility.

rust/src/delta.rs
```diff
@@ -1446,6 +1473,14 @@ impl<'a> DeltaTransaction<'a> {
                 Err(e) => {
                     match e {
                         DeltaTableError::VersionAlreadyExists(_) => {
+                            let checker = ConflictChecker::try_new(
+                                self.delta_table,
+                                version,
+                                operation.clone(),
+                            )
+                            .await?;
+                            let _result = checker.check_conflicts()?;
```
andrei-ionescu, line 1482: Is there a reason why the result is not used?

roeap:
@wjones127 @houqp - things are finally taking shape around here :). As you can see this PR is quite sizeable already and while it also contains changes from #1154 it will stay large.

Thus a quick question as for your preference in reviewing this. I could leave it as is (just rebase once obove PR is merged). I could rearrange the commits, such that its easier to review, or i could dissect into multiple PRs. Happy to hear your thoughts!

I am also preparing a small design document that I will share soon.

wjones127:
FYI I'm planning on reviewing this weekend. :)

rust/src/operations/filesystem_check.rs
```diff
@@ -154,10 +147,11 @@ impl FileSystemCheckPlan {
         }
 
         commit(
-            store,
-            version + 1,
+            self.store.as_ref(),
             &actions,
             DeltaOperation::FileSystemCheck {},
+            snapshot,
+            // TODO pass through metadata
```
roeap, line 154: I plan to do a pass through all our operations in the next Pr, where also I'll be focussing on testing the `commit` function, and addressing these kinds of TODOs.

rust/src/operations/transaction/state.rs
```diff
@@ -0,0 +1,335 @@
...
+            })?;
+
+        // TODO should we add the table name as qualifier when available?
+        let df_schema = DFSchema::try_from_qualified_schema("", self.arrow_schema()?.as_ref())?;
+        let context_provider = DummyContextProvider::default();
+        let sql_to_rel = SqlToRel::new(&context_provider);
+
+        Ok(sql_to_rel.sql_to_expr(sql, &df_schema, &mut Default::default())?)
+    }
+}
```
roeap, line 85: Parsing predicate expressions is actually kind of broken. Regardless of the schema we pass in, int fields are always recognized as i64. I hope, that the abstractions in the core project will eventually help us out. Until then I plan to either do a follow-up in re-writing these expressions, or, if this is a bug, addressing this upstream in datafusion.

rust/src/operations/transaction/test_utils.rs
roeap, line 1: Yet another set of test utilities, and yet another TODO for follow up PR, to consolidate our test helpers...

rust/src/operations/transaction/conflict_checker.rs
```diff
@@ -0,0 +1,862 @@
...
+            vec![file_added],
+            vec![],
+            false,
+        );
+        assert!(result.is_ok());
+
+        // TODO enable test once we have isolation level downcast
+        // add / read + no write
+        // transaction would have read file added by concurrent txn, but does not write data,
+        // so no real conflicting change even though data was added
```
wjones127, line 736: Does `read + no write` mean no data change? like an optimize?

rust/src/operations/transaction/conflict_checker.rs
```diff
@@ -0,0 +1,862 @@
...
+            None,
+            None,
+            vec![tu::create_metadata_action(None, None)],
+            vec![],
+            false,
+        );
+        assert!(matches!(result, Err(CommitConflictError::MetadataChanged)));
+
+        // upgrade / upgrade
+        // current and concurrent transactions chnage the protocol version
```
wjones127, line 812:
```suggestion
        // current and concurrent transactions change the protocol version
```

rust/src/operations/transaction/conflict_checker.rs
```diff
@@ -0,0 +1,862 @@
...
+    NoMetadata,
+}
+
+/// A struct representing different attributes of current transaction needed for conflict detection.
+#[allow(unused)]
+pub(crate) struct TransactionInfo<'a> {
+    txn_id: String,
+    /// partition predicates by which files have been queried by the transaction
+    #[cfg(not(feature = "datafusion"))]
+    read_predicates: Option<String>,
```
wjones127, line 91:
```suggestion
    /// partition predicates by which files have been queried by the transaction
    ///
    /// If any new data files or removed data files match this predicate, the 
    /// transaction should fail.
    #[cfg(not(feature = "datafusion"))]
    read_predicates: Option<String>,
```

rust/src/operations/transaction/conflict_checker.rs
```diff
@@ -0,0 +1,862 @@
...
+                        json_err: e,
+                        version: winning_commit_version,
+                        line,
+                    }
+                })?);
+            }
+            version_to_read += 1;
+        }
+
+        // TODO how to handle commit info for multiple read commits?
```
wjones127, line 229: Yeah, if there are multiple versions, should we just error for now? If there are multiple commits that beat our candidate, it seems like we should generate a WinningCommitSummary for each transaction after `read_version`, right? And then checking our candidate commit against each of them?

rust/src/operations/transaction/conflict_checker.rs
```diff
@@ -0,0 +1,862 @@
...
+#[cfg(feature = "datafusion")]
+use itertools::Either;
+
+/// Exceptions raised during commit conflict resolution
+#[derive(thiserror::Error, Debug)]
+pub enum CommitConflictError {
+    /// This exception occurs when a concurrent operation adds files in the same partition
+    /// (or anywhere in an un-partitioned table) that your operation reads. The file additions
+    /// can be caused by INSERT, DELETE, UPDATE, or MERGE operations.
+    #[error("Concurrent append failed.")]
```
wjones127, line 26: ConcurrentAppend is what went wrong, not what you are trying to do, right?

```suggestion
    #[error("Commit failed: a concurrent transactions added new files.\nHelp: This transaction's query must be rerun to include the new data. Also, if you don't care to require this check to pass in the future, the isolation level can be set to Snapshot Isolation.")]
```

rust/src/operations/transaction/conflict_checker.rs
```diff
@@ -0,0 +1,862 @@
...
+    ConcurrentAppend,
+
+    /// This exception occurs when a concurrent operation deleted a file that your operation read.
+    /// Common causes are a DELETE, UPDATE, or MERGE operation that rewrites files.
+    #[error("Concurrent delete-read failed.")]
```
wjones127, line 31: I think similar changes could be applied to all the errors to make them more user-friendly.

```suggestion
    #[error("Commit failed: a concurrent transaction deleted a file this operation read.\nHelp: This transaction's query must be rerun to exclude the removed data. Also, if you don't care to require this check to pass in the future, the isolation level can be set to Snapshot Isolation.")]
```

rust/src/action/mod.rs
```diff
@@ -641,9 +650,20 @@ impl DeltaOperation {
...
+
+    /// Denotes if the operation reads the entire table
+    pub fn read_whole_table(&self) -> bool {
+        match self {
+            // TODO just adding one operation example, as currently none of the
+            // implemented operations scan the entire table.
+            Self::Write { predicate, .. } if predicate.is_none() => false,
+            _ => false,
+        }
+    }
```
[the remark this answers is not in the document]
wjones127: Or could you provide examples of which operations will?

wjones127 (changes requested):
Overall looks like a great improvement. I think we want to be more conservative in cases where there are multiple concurrent transactions that beat the candidate one. Also the error messages could be improved.

rust/src/operations/transaction/conflict_checker.rs
roeap: Good point. Changed the logic, that out commit loop will always only check against the currently conflicting commit and as such the `WinningCommitSummary` will now summarize exactly one commit.

rust/src/action/mod.rs
```diff
@@ -641,9 +650,20 @@ impl DeltaOperation {
         match self {
             // TODO add more operations
             Self::Write { predicate, .. } => predicate.clone(),
+            Self::Delete { predicate, .. } => predicate.clone(),
             _ => None,
         }
     }
+
+    /// Denotes if the operation reads the entire table
+    pub fn read_whole_table(&self) -> bool {
```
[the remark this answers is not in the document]
roeap: Agreed! If OK with you, I would defer exploring that to a follow up PR. To really get a feel for how that API should look like, I would like to work a bit more with operations that actually need the conflict resolution :).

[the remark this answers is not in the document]
wjones127: That sounds good to me!

rust/src/action/mod.rs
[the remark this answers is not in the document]
roeap: Our operations should actually never use this directly right now. I went through the spark code base, and it seems this is invoked mostly in cases when a plan querying multiple tables with spark has a delta table set as a sink.. This usually uses generic plans and does not invoke the delta operations directly, or rather when we are not too sure what actually happened.

Did not get too deep though. Personally I am a bit conflicted on how to proceed with this. on one hand it does nothing useful right now. On the other hand, we can keep it as a reminder to mindful of this. While right now I think datafusion does not yet have the concept of a sink in their plans, I believe they might be added fairly soon. At this point we may encounter situations, where a plan wants to write to delta, but we have no way of knowing what was scanned.

That said, I do hope that we will be able to inspect the plan metrics and figure out what was read anyhow. Especially since we report the scanned files in out scan operator.

roeap:
@wjones127 - sorry for taking a long time to incorporate your feedback. I think it's mostly done now.

Mainly we now always check against the next version of commits. The questions around `read_whole_table` etc. I hope to address right after this.

As I plus I re-enabled an existing test for optimise, which we can now support again.

wjones127:
I am realizing we have another transaction implementation called `DeltaTransaction`, which is what we are currently using in Python. This improved implementation doesn't seem to have a low-level API where the user can provide the added and removed files, which is what we need for the PyArrow-based writer.

So I think we need to consolidate those into `operations/transaction/mod.rs`, make them public, and switch the Python implementation. That sounds right to you?

wjones127 (approved):
This is looking good now.

The PySpark integration tests will be fixed soon. I should take a look at the sporadic failures in the Azure tests.

roeap:
> So I think we need to consolidate those into `operations/transaction/mod.rs`, make them public, and switch the Python implementation. That sounds right to you?

Yes it does. So far I have deliberately kept the implementation of commits used in the operations module separate from the existing one on the table. Where I am not sure is if we need the `DeltaTransaction` to be a struct at all, I think this is somehow an artifact from the reference implementation, where the Traction is conceptually used kind of like a context in python (it's been a long time since I wrote scala :D). In the new trasaction module we just have the [`commit`](https://github.com/delta-io/delta-rs/blob/d9920aaf6c730fda2878c2271ed396a138243bb1/rust/src/operations/transaction.rs#L140-L156) function, which we should make public and which I think will give us what we need in python.

PR #639, opened 2022-06-14
Defer creation of storage backend in DeltaTableBuilder

### LINKED ISSUE #638: Unable to load multiple tables using DeltaTableBuilder
# Environment

**Delta-rs version**: main
***
# Bug

**What happened**:
Unable to create multiple ADLS backed Delta tables using `DeltaTableBuilder` since the builder creates a backend when calling `from_uri`. ADLS and other backends use environment variables by default to pull their configuration. Since the configuration is not setup in the environment but is stored in map it results in a error.

Either defer creation of the backend to when `build` is called or add a function to `DeltaTableBuilder` which accepts the uri and a map.

Blajda:
# Description
Defers the creation of the storage backend until `build` is called on the builder unless a backend is provided by the user.

An alternative approach to resolving this is to add a new method called `from_uri_with_map` which will use a map when creating a backend. If we want to use this approach I have no problem with changing the implementation.

# Related Issue(s)

- closes #638

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -463,7 +463,7 @@ pub struct DeltaTableLoadOptions {
     /// table root uri
     pub table_uri: String,
     /// backend to access storage system
-    pub storage_backend: Box<dyn StorageBackend>,
+    pub storage_backend: Option<Box<dyn StorageBackend>>,
     /// specify the version we are going to load: a time stamp, a version, or just the newest
     /// available version
     pub version: DeltaVersion,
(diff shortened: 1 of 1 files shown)


rust/src/delta.rs
```diff
@@ -542,22 +542,23 @@ impl DeltaTableBuilder {
 
     /// explicitely set a backend (override backend derived from `table_uri`)
```
roeap, line 543: nit: this comment is not quite correct any more 😆.

roeap (approved):
Looks good to me! I'll leave it open a bit for others to chime in.

Blajda:
Thanks @roeap . I appreciate it

roeap (approved):

PR #644, opened 2022-06-16
Refactor azure storage with crate updates

roeap:
# Description

With the latest version of the azure SDKs some annoyances got fixed. specifically, list now returns something `Send`able, which allows us to significantly simplify the `list_objs` method. Secondly the errors are no longer a deeply nested nightmare 😆.

Also added a test for validating that service principal access works (#609) - currently i am guessing the issues encountered there are more related to #643

# Related Issue(s)

validate the SP access works for #609

# Documentation

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,28 +31,17 @@ jobs:
         os:
           - ubuntu-latest
           - macOS-10.15
-          - windows-2019
+          - windows-latest
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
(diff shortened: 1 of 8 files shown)


Blajda:
Hi @roeap you can remove `tokio-util` from the dependency list since it was introduced for the !Send issue.

roeap:
Thanks @Blajda  - always happy to be able to slim down dependencies 😆

wjones127 (dismissed):
This is a great cleanup! LGTM.

roeap:
@wjones127 - we saw failing CI due to windows runners running out of disk space. Based on this comment https://github.com/pola-rs/polars/pull/3122#issuecomment-1096564533 I switched to using Swatinem/rust-cache, which seems to do some nice things around managing cache size and utilization.

Right now builds are passing again, but this could just be related to using a "fresh" cache, we'll have to keep an eye on it.

Could you have another look?

wjones127 (approved):
This looks like it helped Polars and they are still using it, so should be good. I'll keep on eye on that CI job.

PR #648, opened 2022-06-19
feat: add enforce_retention_duration param to vacuum method

houqp:
# Description

This maps to the reference implementation's
spark.databricks.delta.retentionDurationCheck.enabled config.

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -30,6 +30,8 @@ use super::storage::{StorageBackend, StorageError, UriError};
 use super::table_state::DeltaTableState;
 use crate::delta_config::DeltaConfigError;
 
+const MILLIS_IN_HOUR: i64 = 3600000;
+
 /// Metadata for a checkpoint file
 #[derive(Serialize, Deserialize, Debug, Default, Clone, Copy)]
 pub struct CheckPoint {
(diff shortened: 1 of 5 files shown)


rust/src/delta.rs
```diff
@@ -175,9 +175,14 @@ pub enum DeltaTableError {
     },
     /// Error returned when Vacuum retention period is below the safe threshold
     #[error(
-        "Invalid retention period, retention for Vacuum must be greater than 1 week (168 hours)"
+        "Invalid retention period, retention for Vacuum is configured to be greater than {} milliseconds, got {} milliseconds", .provided, .min
```
roeap, line 178: should the order of `.provided` and `.min` be inverted?

roeap (commented):
Looks good overall. Left a small comment for the error text.

rust/src/delta.rs
houqp: good catch!

roeap (approved):

rust/src/delta.rs
```diff
@@ -175,9 +175,14 @@ pub enum DeltaTableError {
     },
     /// Error returned when Vacuum retention period is below the safe threshold
     #[error(
-        "Invalid retention period, retention for Vacuum must be greater than 1 week (168 hours)"
+        "Invalid retention period, retention for Vacuum is configured to be greater than {} milliseconds, got {} milliseconds", .min, .provided
```
wjones127, line 178: Since the user provided hours, would it be possible to have the error message report back in hours instead of milliseconds?

houqp: updated :+1:

wjones127 (approved):
Thanks, this looks great!

PR #649, opened 2022-06-19
refactor: move version field to  `DeltaTableState`

### LINKED ISSUE #390: Move version field into `DeltaTableState` struct
# Description
`version` is currently a field in `DeltaTable`. It would make more sense to move this into `DeltaTableState` since version maps one to one to state.

roeap:
# Description

This PR moves the `version` field into `DeltaTableState`. The added consistency helps implementing #632 - split this in a separate PR to make review easier.

# Related Issue(s)

closes #390 

# Documentation

### DIFF
--- a/rust/src/table_state.rs
+++ b/rust/src/table_state.rs
@@ -165,11 +165,17 @@ impl DeltaTableState {
 
     /// The most recent metadata of the table.
     pub fn current_metadata(&self) -> Option<&DeltaTableMetaData> {
-        println!("current metadata");
         self.current_metadata.as_ref()
     }
 
     /// Merges new state information into our state
+    ///
+    /// The DeltaTableState also carries the version information for the given state,
+    /// as there is a one-to-one match between a table state and a version. In merge/update
+    /// scenarios we cannot infer the intended / correct version number. By default this
+    /// function will update the tracked version if the version on `new_state` is larger then the
+    /// currently set version however it is up to the caller to update the `version` field according
+    /// to the version the merged state represents.
     pub fn merge(
         &mut self,
         mut new_state: DeltaTableState,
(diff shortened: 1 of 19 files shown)


rust/src/table_state.rs
```diff
@@ -211,6 +222,11 @@ impl DeltaTableState {
         if !new_state.commit_infos.is_empty() {
             self.commit_infos.append(&mut new_state.commit_infos);
         }
+
+        // TODO should we make sure versions are consistent (i.e. +1) or just update to highest?
+        if self.version < new_state.version {
+            self.version = new_state.version
+        }
```
roeap, line 229: Here i was unsure how to best proceed. updating the version on merge seems a bit iffy to me. At best we get the state for whatever state has the highest version. However we have no way of knowing at this point if this represents an existing state, or some future state, or ... So I did not want to put any logic here.

wjones127: I'm not sure I understand the `merge` command. If you have two `DeltaTableState`s, isn't the correct one simply the newest one? Or is it possible for `DeltaTableState` to represent just incremental changes in table?

roeap: yes, `DetaTableState` can be created from a collection of actions - so a single commit or the entire reconciled table state. so the merge function is mainly used to increment the state after updates etc... previously it had no notion of versions and is e.g. also used to reconcile all actions for a checkpoint...

wjones127: I think this behavior should be fine as long as we document it in the method documentation.

roeap: added some more verbose doc string explaining how merge treats version numbers.

wjones127 (approved):

PR #650, opened 2022-06-19
Read remove action num records field

### LINKED ISSUE #614: Warning spam: Unexpected field name `numRecords` for remove action
# Environment

**Delta-rs version**: 0.5.7

**Binding**: Python

**Environment**: 
- **Cloud provider**: Azure
- **OS**: Ubuntu 18.04
- **Other**: Python 3.8

***
# Bug

**What happened**:
We get an enormous spam of warning logs 
```
[2022-05-30T18:54:37Z WARN  deltalake::action] Unexpected field name `numRecords` for remove action: Row { fields: [("path", Str("part-00031-24527240-103f-4850-b36a-96523433d62d-c000.snappy.parquet")), ("deletionTimestamp", Long(1653935751443)), ("dataChange", Bool(false)), ("extendedFileMetadata", Bool(true)), ("partitionValues", MapInternal(Map { entries: [] })), ("size", Long(1444)), ("tags", Null), ("numRecords", Null)] }
```
It looks like I get one warning log for every remove action in the transaction log of my table.

Opens the table without logging warnings which I'm pretty sure are un-important.

**How to reproduce it**:
1. Use delta-spark 1.2.1
2. Create a test delta table and apply some random modifications until you've made enough transactions to create a checkpoint. 
3. Try to create a `deltalake.DeltaTable` object from that table.
4. You should see a bunch of warnings like the one I included above. 

Python script I created to reproduce:
[reproduce_deltalake_warnings.zip](https://github.com/delta-io/delta-rs/files/8800129/reproduce_deltalake_warnings.zip)

**More details**:
I think a new version of delta-spark has introduced this new `numRecords` field so I guess delta-rs needs to be updated to handle this. It would be really nice to have a fix for this :slightly_smiling_face: . For large tables with long history retention these warnings are pretty unmanageable. 

I have put a hack in my code which sets the env var `RUST_LOG="error"` to work around this for now.

Tom-Newton:
# Description
When reading the field is parsed and set to null if the field is not present or `null`. When writing `numRecords` is set always set to null, the same as is currently the case when writing with spark based delta 1.2.1.

# Related Issue(s)
- closes #614 

# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -322,6 +322,7 @@ impl RawDeltaTable {
                         partition_values: Some(old_add.partition_values.clone()),
                         size: Some(old_add.size),
                         tags: old_add.tags.clone(),
+                        num_records: None,
                     });
                     actions.push(remove_action);
                 }
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -561,6 +561,8 @@ pub struct Remove {
     pub size: Option<DeltaDataTypeLong>,
     /// Map containing metadata about this file
     pub tags: Option<HashMap<String, Option<String>>>,
+    /// Number of records in the file associated with the log action.
+    pub num_records: Option<DeltaDataTypeLong>,
 }
 
 impl Hash for Remove {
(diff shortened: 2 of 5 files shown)


Tom-Newton:
Closing in favour of https://github.com/delta-io/delta-rs/pull/651

PR #640, opened 2022-06-15
Release of the Python binding version 0.5.8

fvaleye:
# Description
- Bump version of maturin to `0.12.19`
- Update the rust dependencies
- Prepare the release of the new version of the Python binding 0.5.8

PR #651, opened 2022-06-19
fix: Prevent warning spam when reading tables generated by delta 1.2.1

### LINKED ISSUE #614: Warning spam: Unexpected field name `numRecords` for remove action

Tom-Newton:
# Description
https://github.com/delta-io/delta/commit/6f396301f74bea16ad27404a7c79e9aa698d6e2c added a new numRecords field on remove actions. To prevent a large spam of warnings when reading tables with this new field delta-rs needs to be updated to accept this field.

I've just gone for the simplest possible fix which is to just ignore `numRecords` when reading. I had a go at [making it read the field](https://github.com/delta-io/delta-rs/pull/650) but being a novice I didn't really know what the impact of those changes would be. 

# Related Issue(s)
- closes #614

### DIFF
--- a/rust/src/action.rs
+++ b/rust/src/action.rs
@@ -657,6 +657,7 @@ impl Remove {
                 "size" => {
                     re.size = record.get_long(i).map(Some).unwrap_or(None);
                 }
+                "numRecords" => {}
                 _ => {
                     log::warn!(
                         "Unexpected field name `{}` for remove action: {:?}",


wjones127:
> I've just gone for the simplest possible fix which is to just ignore numRecords when reading. 

That's sounds like acceptable behavior to me for now. Until we have a use for it, no need to read that field.

Do we know what the motivation was for adding it? I don't see any pull request or discussion associated with that commit 😞

Tom-Newton:
> Do we know what the motivation was for adding it? I don't see any pull request or discussion associated with that commit disappointed

I was also wondering this... possibly @scottsand-db can enlighten us?

wjones127 (approved):

Tom-Newton:
Thanks for reviewing @wjones127. Do you think we could get this into the new Python release I'm seeing https://github.com/delta-io/delta-rs/pull/640 ?

scottsand-db:
I've made a PR to fix this here: https://github.com/delta-io/delta/pull/1230

Issue #577, opened 2022-03-23
Python: Support date32 and decimal statistics in PyArrow writer


wjones127:
# Description

In PyArrow, date32 and decimal statistics from parquet metadata aren't translated into appropriate type, but are instead raw bytes. We can either decode them ourselves, or wait for PyArrow to implement that conversion. Upstream issue: https://issues.apache.org/jira/browse/ARROW-7350

**Use Case**

**Related Issue(s)**
 * #542

wjones127:
The upstream issue is solved and will be released soon in PyArrow 8.0.0.

PR #659, opened 2022-06-27
Support date32 and decimal stats in write_deltalake

### LINKED ISSUE #655: dates statistics are not saved in the delta log
# Environment

**Delta-rs version**: 0.5.8

**Binding**: Python

**Environment**: 
- **Cloud provider**:  NA
- **OS**: Windows 
- **Other**:

***
# Bug

**What happened**: Date statistics are not saved in the log file

**What you expected to happen**: expect the statistics to be working

**How to reproduce it**:
```
import duckdb 
con = duckdb.connect()
df =con.execute('''
INSTALL 'httpfs';
LOAD 'httpfs';
SELECT countries_and_territories, date_rep,year(date_rep) as year,sum(deaths) as total FROM 
read_parquet('https://pandemicdatalake.blob.core.windows.net/public/curated/covid-19/ecdc_cases/latest/ecdc_cases.parquet')
group by all
order by total desc
''').fetch_arrow_table()

write_deltalake('C:/covid', df,mode='overwrite',partition_by=['year'])
```

**More details**:

this is literally the first time, I used Delta, I hope it is a supported feature

wjones127:
# Description

Since the updates in PyArrow 8.0, we can now support date32 and decimal stats. To check versions, I added `packaging` as a dependency.

# Related Issue(s)

- closes #655

# Documentation

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -18,6 +18,7 @@ dependencies = [
     "pyarrow>=4",
     'numpy<1.20.0;python_version<="3.6"',
     'dataclasses;python_version<="3.6"',
+    "packaging>=20",
 ]
 
 [project.optional-dependencies]
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -17,6 +17,8 @@ from typing import (
     Union,
 )
 
+from packaging.version import Version, parse
+
 from deltalake.fs import DeltaStorageHandler
 
 if TYPE_CHECKING:
@@ -38,6 +40,8 @@ except ModuleNotFoundError:
     _has_pandas = False
 else:
     _has_pandas = True
+
+PYARROW_VERSION = Version(pa.__version__)
 
 
 class DeltaTableProtocolError(PyDeltaTableError):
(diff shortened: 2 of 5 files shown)


python/pyproject.toml
```diff
@@ -18,6 +18,7 @@ dependencies = [
     "pyarrow>=4",
     'numpy<1.20.0;python_version<="3.6"',
     'dataclasses;python_version<="3.6"',
+    "packaging>=20",
```
houqp, line 21: Just my 2 cents, maybe not worth introducing an extra dependency if all we need is parsing pyarrow major version number. Would be good to keep the runtime dependency as lean as possible.

houqp (dismissed):

python/pyproject.toml
wjones127: Hmm yeah I guess I can do a string split to get the major version.

wjones127: It's a dependency of pytest, so we already have it implicitly in our dev environment FYI. So I'll leave it in the tests.

houqp (approved):
LGTM!

roeap (approved):

wjones127:
Also closes #577

Issue #647, opened 2022-06-19
Vacuum isuses for ADLS


Blajda:
# Environment

**Delta-rs version**: git 649fdce856

***
# Bug

**What happened**:
Calling `table.vacuum` with a ADLS backed table results in a panic. Working around the panic results in no files being picked up.
 
`thread 'vacuum::test_adls_vacuum' panicked at 'byte index 80 is out of bounds of adls2://mystorage@mystorageacct.dfs.core.windows.net/mystorage/mytable', library/core/src/str/mod.rs:107:9`

(Ignore the Index number. Modified the paths of my accounts)

Adding a  `/` to the end of the table URI does not work since some normalization is performed during table load but would not matter due to the below issues.

**More details**:

I removed the ` + 1` on this [line](https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L406) to check if it was the main issue but encounter other issues.

`list_obj` for ADLS may have two issues:
1. When `ObjectMeta` is created `path` from the callee of `list_obj` is used and not the actual paths from the API.
2. Currently recursive is set to false. Vacuum only calls `list_obj` for the table's URI hence only  `_delta_log` and partition directories are being returned. This might be an issue with how vacuum is implemented.

See: https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/azure/mod.rs#L369-L375

Blajda:
Hi @roeap, looking for you analysis on this. If (2.) should be handled from the vacuum side I can take that work on.

roeap:
@Blajda - thanks for the report! will take a look and see what needs to be done ...

Blajda:
Closing since #683 is merged and the vacuum tests passed.

PR #664, opened 2022-06-29
bugfix: Make sure vacuum works on relative paths

### LINKED ISSUE #663: vacuum isn't actually deleting files
# Environment

**Delta-rs version**: 0.5.8

**Binding**: Python

**Environment**: localhost

***
# Bug

**What happened**: Ran `dt.vacuum(retention_hours=0, enforce_retention_duration=False, dry_run=False)` and files didn't actually get deleted.

**What you expected to happen**: I thought the stale files would actually get deleted.

**How to reproduce it**: Here's [the notebook to reproduce this](https://github.com/MrPowers/delta-examples/blob/master/notebooks/delta-rs/intro-to-delta-rs.ipynb).

**More details**: `enforce_retention_duration` was added recently and that's successfully letting me bypass the retention duration error I was facing before.  Let me know if I'm missing something here and there's a way to get the files vacuumed with the existing code.

wjones127:
# Description

It seems that when we pass a relative path to `DeltaTable()`, vacuum doesn't actually delete files. I've added tests in Python which is more convenient now that we have a writer.

# Related Issue(s)

* Fixes #663

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1277,12 +1277,7 @@ impl DeltaTable {
         if dry_run {
             return Ok(files_to_delete);
         }
-
-        let paths = &files_to_delete
-            .iter()
-            .map(|rel_path| self.storage.join_path(&self.table_uri, rel_path))
-            .collect::<Vec<_>>();
-        match self.storage.delete_objs(paths).await {
+        match self.storage.delete_objs(&files_to_delete).await {
             Ok(_) => Ok(files_to_delete),
             Err(err) => Err(DeltaTableError::StorageError { source: err }),
         }
--- a/python/tests/test_table_read.py
+++ b/python/tests/test_table_read.py
@@ -194,33 +194,6 @@ def test_read_table_with_stats():
         assert data.num_rows == 0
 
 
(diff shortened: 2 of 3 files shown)


rust/src/delta.rs
```diff
@@ -1277,12 +1277,7 @@ impl DeltaTable {
         if dry_run {
             return Ok(files_to_delete);
         }
-
-        let paths = &files_to_delete
-            .iter()
-            .map(|rel_path| self.storage.join_path(&self.table_uri, rel_path))
-            .collect::<Vec<_>>();
-        match self.storage.delete_objs(paths).await {
+        match self.storage.delete_objs(&files_to_delete).await {
```
wjones127, line 1280: I'm not sure how this was working in object stores, but in local fs `files_to_delete` was already absolute. Local file system used `Path::join()`, which had the fortunate behavior of just using the second path if it was already absolute.

roeap: well... maybe its not :) - will have a closer look once i get to #647.

roeap (approved):
LGTM

Issue #667, opened 2022-06-30
Test vacuum deletes correct files on object_stores


wjones127:
# Description

I don't believe we actually test this; we only test with `dry_run = true`.

**Use Case**

**Related Issue(s)**

mrk-its:
vacuum definitely needs more tests (even with dry_run=true only). After short play with it I found number of issues not covered by tests:

* files not referenced by delta log at all (so not included on `DeltaTableState.files()` and `DeltaTableState.all_tombstones()` lists) are completely ignored and not vacuumed (this can easily happen if vacuum is not running regular way and old log entries are cleaned up by delta itself)

* while testing I noticed that for local `file` storage vacuum doesn't work for files nested in subdirectories. It looks like `FileStorageBackend.list_objs` returns only top-level files / dirs

I have a patch for first issue from above list, are you interested in PR?

mrk-its:
The good thing is `delta-rs` vacuum implementation is blazingly fast ;) comparing to official java VACUUM impl (at least for S3 backend). I did tests on real, production S3 dataset (in dry-run mode) and single-threaded delta-rs vacuum completed in ~7minutes while official java VACUUM running on multi-node EMR cluster didn't complete in 24h (https://github.com/delta-io/delta/issues/1248). 

The reason of such big difference is numer of S3 LIST requests issued by each implementation needed to fetch list of all files in dataset (java impl executes separate S3 LIST for each directory ending with over 1mln of requests for test dataset, delta-rs does that optimal way and needs ~1000 requests to list all the files).

wjones127:
> I have a patch for first issue from above list, are you interested in PR?
A PR would be very welcome!

And good to know our performance is good. I think once we get good tests in place we should release a blog post with some performance comparisons.

houqp:
Very cool to hear about that as well! Like you mentioned, we haven't even started any optimization work yet :D First obvious step is to parallelizing it with multiple threads. After that, we could leverage datafusion and ballista, to further spread the load on top multiple nodes.

wjones127:
Resolved by #669

PR #669, opened 2022-07-01
:WIP: Refactor vacuum to commit to log and add builder

Blajda:
# Description
Pull the vacuum operation out of the Delta table and commit to the log various metrics.

The structure of this code is similar to Optimize where the operation uses a struct which has a plan and execute component.
It also introduces a builder with reasonable defaults and helps ensure that future enhancements do not break the call API.

The current call API requires the user to specify the retention interval as hours. I have changed this to use Chrono's Duration since I find it more ergonomic then implement time conversions our selves. The Databricks implementation allows the user to pass a float value for periods smaller than a day which brings us in line.

# Related Issue(s)
I want to close these related issues. Let me know if there's anything else you would like to see
#97
#667 

## TODO

- [ ] Update python bindings
- [ ] Test that an actual run deletes stuff
- [ ] Validate log update scenario and data


mrk-its:
Can you implement this `commit to the log` feature as optional (configurable by builer)? 
Asking because I'm planning to use delta-rs as high-performant vacuum replacement for official VACUUM - but currently delta-rs is incompatible with S3 Multi Cluster setup (with DynamoDB, as described [here](https://docs.delta.io/2.0.0rc1/delta-storage.html#multi-cluster-setup))  Correct me if I'm wrong, but it seems vacuum-only without writing to delta log should be safe.

mrk-its:
@Blajda it was fast, thanks!

Blajda:
Hi @mrk-its,
I've made the commits during vacuum configurable with the builder as suggested. You are correct that vacuums can be performed without appending to the log.

You might want to hold off on using this implementation of vacuum since the new tests show multiple issues with deleting nested partitions. It may also have integration issues with aws since we have some for Azure.

mrk-its:
Sure, I'm going to be super careful, for now I'm simply running it in dry-run mode.

> You might want to hold off on using this implementation of vacuum since the new tests show multiple issues with deleting nested partitions. It may also have integration issues with aws since we have some for Azure.

Can you point me to the test build with these issues for nested partitions? I probably can help with tests on AWS

Blajda:
This PR contains new  vacuum tests for various scenarios in `tests/vacuum_error.rs`. Currently the tests that fail with local storage are marked as ignore.

I might split this PR since it also contains new test code to easily test with different backends without having to worry about the clean up and setup. If you look into `tests/common` you can now specify `DELTA_RS_TEST_BACKEND` with either `LOCALFS` or `AZURE_GEN2` and then setup the corresponding environment variables for the backend. If the main maintainers are okay with this approach it would be great to add some S3 support.

Blajda:
Closing this to break it down into multiple PRs.

PR #670, opened 2022-07-03
Fix linting / build on main

mrk-its:
# Description
Currently builds on main fail because of clippy linting, this PR fixes that


fvaleye (approved):
Thanks @mrk-its 

LGTM!

Issue #661, opened 2022-06-27
[discussion] move delta log handling to new struct `DeltaLog` and harmonize operations


roeap:
# Description

As we add more and more high level operations (vacuum, optimize, writes, ....) our `DeltaTable` struct is continuously growing in complexity, and maybe it's time for us to have another look at our APIs.

The main proposal would be

1. Move delta log handling into new `DeltaLog` struct. This would likely also make things like #528 easier.
2. Implement all operations as separate structs, much like current optimize or the high level write operations potentially leveraging the upcoming [`into_future`](https://github.com/rust-lang/rust/issues/67644#issuecomment-986257951) feature. Additionally using a builder pattern to configure the specific operation.
3. The `DeltaTable` APIs could then look somewhat like the current [pyspark apis](https://docs.delta.io/latest/api/python/index.html#delta.tables.DeltaTable)

what do you think? @houqp, @rtyler @xianwill @wjones127 @mosyp @fvaleye

Blajda:
Hi @roeap, I'm interested in performing the work of factoring out the vacuum code into a struct related to line item 2. It would also be good opportunity to add commit information since it is now supported.

I like a `DeltaLog` proposal since it would allow us to create different `DeltaTable` variants.  One variant can be created to support append only write operations without having to track the log in memory which be an issue for larger tables. Going further we can also use the trait system to validate if a particular operation requires a `DeltaLog` at compile time.

roeap:
@Blajda - obviously any contribuiton is highly welcome :).

Currently I am working on #632, where I am thinking a bit about how to structure our codebase to more efficiently handle out table state / log... One thing that keeps on coming up is that it might be beneficial to move the object store and stare around, rather then the whole table. The proposed delta log would probably be more or less immutable (except for a cache maybe)...

I guess what I am saying is, I'd be very keen on seeing an implementation for vacuum, but am not sure yet which way we end up agree an pursuing :).

wjones127:
I think a re-oganization is a good idea. I'd like to see some examples of how usage would change, though. Since many operations depend on having an up-to-date `DeltaLog` state, it's unclear how you can separate that from the top-level `DeltaTable` struct.

roeap:
[...]

PR #672, opened 2022-07-04
Factor vacuum and implement a builder

Blajda:
# Description
Factor out vacuum errors and operation into it's own file and use a builder with reasonable defaults. No changes in functionality except the user can provide a `Duration` when specifying the retention period. This brings us in line with the reference implementation which allows for time periods less than a hour.

This is a breakdown of  #669 to only include the builder functionality.


rust/src/delta.rs
```diff
@@ -1188,99 +1177,22 @@ impl DeltaTable {
...
-            {
-                continue;
-            }
-
-            files_to_delete.push(obj_meta.path);
+        enforce_retention_duration: bool,
+    ) -> Result<Vec<String>, VacuumError> {
+        let mut plan = Vacuum::default()
+            .dry_run(dry_run)
+            .enforce_retention_duration(enforce_retention_duration);
```
wjones127, line 1189: I really like that API! 😄

rust/src/vacuum.rs
```diff
@@ -0,0 +1,217 @@
+//! Vacuum a Delta table
+//!
```
wjones127, line 2: Hmm we should probably turn `wrap_comments = true` (in a separate PR).

rust/src/vacuum.rs
```diff
@@ -0,0 +1,217 @@
...
+//! Run the Vacuum command on the Delta Table: delete files no longer referenced by a Delta table and are older than the retention threshold.
+//! We do not recommend that you set a retention interval shorter than 7 days, because old snapshots
+//! and uncommitted files can still be in use by concurrent readers or writers to the table.
+//! If vacuum cleans up active files, concurrent readers can fail or, worse, tables can be
+//! corrupted when vacuum deletes files that have not yet been committed.
+//! If `retention_hours` is not set then the `configuration.deletedFileRetentionDuration` of
+//! delta table is used or if that's missing too, then the default value of 7 days otherwise.
```
wjones127, line 9: One other warning that may be worth calling out (even though it's implicit here): vacuuming for before a certain date means time travel may not work for commits before that date.

rust/src/vacuum.rs
```diff
@@ -0,0 +1,217 @@
...
+use futures::StreamExt;
+use std::collections::HashSet;
+use std::fmt::Debug;
+
+#[derive(Debug)]
+/// Vacuum a Delta table with the given options
+/// See this module's documentation for more information
+pub struct Vacuum {
+    /// Period of stale files allowed.
+    pub retention_period: Option<Duration>,
```
wjones127, line 23: Thank you for making this a duration 🙌

wjones127 (dismissed):
I only had one minor suggestion. Otherwise this looks great!

rust/src/vacuum.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    /// Perform the vacuum. Returns metrics on which files were deleted
+    pub async fn execute(self, table: &mut DeltaTable) -> Result<VacuumMetrics, VacuumError> {
+        let dry_run = self.dry_run;
+        let plan = create_vacuum_plan(table, self).await?;
+        if dry_run {
+            return Ok(VacuumMetrics {
+                files_deleted: plan.files_to_delete,
+                dry_run: true,
+            });
```
roeap, line 212: excellent - returning metrics like this is exactly the right thing to do. In  #673 this is what's needed to do proper conflict resolution. There we need to know all files read by a transaction. There is a `CurrentTransactionInfo` struct in that PR that holds all we want to know about the current transaction. I see now that this does not care about files removed from the table. but while the complete transaction info object is maybe not suitable here, maybe we want a more generic Metrics struct, that can also report on other metrics?

rust/src/delta.rs
```diff
@@ -1188,99 +1177,22 @@ impl DeltaTable {
...
         }
 
-        if dry_run {
-            return Ok(files_to_delete);
-        }
-        match self.storage.delete_objs(&files_to_delete).await {
-            Ok(_) => Ok(files_to_delete),
-            Err(err) => Err(DeltaTableError::StorageError { source: err }),
-        }
+        let res = plan.execute(self).await?;
```
roeap, line 1194: One of my hopes in proposing #661 was that we could eventually return the builder and arrive at an API like.

```rs
let dt = open_table(...);
let result = dt.vacuum().enforce_retention_duration(true).into_future().await;
```

once the [`IntoFuture`](https://doc.rust-lang.org/std/future/trait.IntoFuture.html) trait lands, this would simplify to.

```rs
let dt = open_table(...);
let result = dt.vacuum().enforce_retention_duration(true).await;
```

This would also mean we don't have to have vacuum itself be async anymore, but requires that the `VacuumPlan` has everything it needs to execute. I left some thoughts in to top level comment.

roeap (commented):
@Blajda - excellent work! I left some comments for discussion, but am not sure how deep those changes would run.

When proposing #661 I was hoping that we can move away from passing the entire table around to operations, but rather a snapshot of the table (`DeltaTableState`) and the object store, which may eventually be encapsulated in the mentioned dedicated `DeltaLog`. I believe that this would open us up to a bunch more clean-up and optimizations down the road, and would also simplify the implementation of #632. This would also much more mirror the designs found the in reference implementation.

My main concern with using just the state is, that we really don't want to clone that as it can be very expensive (maybe wrap in `Arc`?)

I haven't had the time to think about the implications this properly but maybe this PR is an opportunity to look into some of this?

Of course very curious what other think about this - @houqp @wjones127?

Blajda:
> When proposing #661 I was hoping that we can move away from passing the entire table around to operations, but rather a snapshot of the table (`DeltaTableState`) and the object store, which may eventually be encapsulated in the mentioned dedicated `DeltaLog`

Yes I fully agree with this. For this operation only the delta log at a particular version is required and since it's a snapshot isolate we don't have to worry about conflicts. 

When it comes to avoiding clones, I don't think just having the state in Arc would solve it since you would need to clone the state to put it into an Arc which then prevent writes. Naively you could use a linked list where each node points to the previous version and then pass around the 'head' of the list. Then operations can write to the log without impacting the view of others and reconcile/fail when they read the ground truth. The primary challenge here would be with Rust's ownership model but I feel like there's a functional data struct that would solve and it would definitely require some reference counting. I might be completely off here >.<

rust/src/vacuum.rs
Blajda: I'm not sure if a generic Metrics struct would really work in Rust but I think it would be a good place for traits. For example, tracking the time an operation took is common so a `TimeCostMetric` trait could be used to obtain that. Ditto for conflict resolution. A trait that specifies all the information required can be defined and then each operation is required to implement it.

rust/src/delta.rs
Blajda: Yeah if `DeltaLog` was implemented the `into_future()` change would be trivial.  I'll play on a nightly branch and see how hard it would be without the log.

rust/src/vacuum.rs
```diff
@@ -0,0 +1,217 @@
...
+pub struct VacuumMetrics {
+    /// Was this a dry run
+    pub dry_run: bool,
+    /// Files deleted successfully
+    pub files_deleted: Vec<String>,
+}
+
+/// Errors that can occur during vacuum
+#[derive(thiserror::Error, Debug)]
+pub enum VacuumError {
```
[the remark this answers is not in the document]
Blajda: Yeah creating appropriate error types is harder than the borrow checker at times :laughing:

Blajda:
Hi @wjones127 and @roeap 
Appreciate your comments and insights that you shared in this PR. I have updated the module description with a bit on time travel, added an example, and resolved the merge conflict.

wjones127 (approved):

roeap (approved):

PR #673, opened 2022-07-05
fix: inconsistent path in azure list

### LINKED ISSUE #671: 'list_objs` behaviour is not uniform among backends
# Bug
The file system and Azure  `list_objs` implementations  return only objects within the directory specified by `path` which differs with the S3 implementation which returns all objects that have a prefix of `path`. These issues were uncovered when implementing additional tests for vacuum in #669.

Looking at the doc string for the storage backend it states:  `/// Return a list of objects by path prefix in an async stream.` so it seems like the correct behavior is implemented by S3. 

Looking for confirmation on what is expected so we can resolve the outliers.

roeap:
# Description

This PR fixes the list method in the azure storage backend, that was introduced in the latest refactor 

# Related Issue(s)

closes #671

# Documentation


rust/src/storage/azure/mod.rs
```diff
@@ -368,7 +368,7 @@ impl StorageBackend for AdlsGen2Backend {
             .flat_map(|it| match it {
                 Ok(paths) => Either::Left(stream::iter(paths.into_iter().map(|p| {
                     Ok(ObjectMeta {
-                        path: path.to_string(),
+                        path: format!("adls2://{}/{}", self.file_system_name, path.to_owned()),
```
wjones127, line 371: It's unclear to me how this fixes the bug reported:

> The file system and Azure list_objs implementations return only objects within the directory specified by path which differs with the S3 implementation which returns all objects that have a prefix of path. 

It seems like this change just changes the format of the file paths in `list_dir`? I can clearly see `.recursive(false)` above, so maybe that's the change being asked for? Am I missing something?

wjones127:
Also what would you think of integration testing with the Azurite docker container? We could add it to the `docker_compose.yml` file at the top level.

rust/src/storage/azure/mod.rs
roeap: you are eight, i was too hasty there. .. but this yet another bug....

roeap:
Integration tests are definately the way to go - unfortunately azurite does not support the alds gen2 apis we use here. I once thought about migrating the current state in object_store to also support blob apis and do integration tests.

There is a way to something integration-like with the ADLS2 apis - a record / replay pattern. While this creates a lof of files in the repo at an un-configurable location, it dies do testing :).

On other days I prefer getting object_store up to state that we can use it here.

rust/src/storage/azure/mod.rs
roeap: btw. I do remember now where this comes from - likely the same for the local store. SO far (aside from vacuum) we never list the file directory, bur always have the delta log when accessing files. Vacuum assumes that there might be orphaned files, not discoverable via the delta log, so it does a list on the files. This is where folders first come into play ... in an ideal world this would not even be necessary, as we should only remove tombstones if we remove files - right?

wjones127: I think the reason we list is there might be some files left over from aborted transactions, which wouldn't show up in the logs.

roeap: Since we need to fix local as well, I was thinking to move the walkdir based implementation over from object_store_rs ... also really looking forward to having this migration :). I think we need some more more on the auth mechanism over there before we can switch, but getting there ...

wjones127: Yes, I'll be focusing on that soon.

PR #681, opened 2022-07-10
fix: traverse directories in local `list_objs`

### LINKED ISSUE #671: 'list_objs` behaviour is not uniform among backends

roeap:
# Description

So far we were not traversing into directories when listing objects with the local storage backend - this is required for the vacuum command to work properly (i.e discovering orphaned files not in the log).
I more or less ported the implementation from object_store_rs, as I believe the code over there is of very high quality.

As a "drive-by" I updated the return type definitions on all `list_objs` using the `BoxStream` shorthand.

@Blajda  - would the vacuum tests validate this fix?

# Related Issue(s)
towards #671

# Documentation

### DIFF
--- a/rust/src/storage/file/mod.rs
+++ b/rust/src/storage/file/mod.rs
@@ -2,18 +2,16 @@
 //!
 //! The local file storage backend is multi-writer safe.
 
+use super::{ObjectMeta, StorageBackend, StorageError};
+use chrono::DateTime;
+use futures::{stream::BoxStream, StreamExt};
+use std::collections::VecDeque;
 use std::io;
 use std::path::{Path, PathBuf};
-use std::pin::Pin;
-
-use chrono::DateTime;
-use futures::{Stream, TryStreamExt};
 use tokio::fs;
 use tokio::io::AsyncWriteExt;
-use tokio_stream::wrappers::ReadDirStream;
-
-use super::{ObjectMeta, StorageBackend, StorageError};
 use uuid::Uuid;
+use walkdir::WalkDir;
 
 mod rename;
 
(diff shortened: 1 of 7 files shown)


Blajda:
Hi @roeap I merged this branch with #682 and the only failing test is `test_non_managed_files` which brings it in line with S3. Enhancements need to be made to vacuum algorithm for that test to pass. No changes required from the backend.

rust/src/storage/file/mod.rs
```diff
@@ -87,33 +84,70 @@ impl StorageBackend for FileStorageBackend {
...
-                            StorageError::Generic("cannot convert to i64".to_string())
-                        })?),
-                    })
+    ) -> Result<BoxStream<'a, Result<ObjectMeta, StorageError>>, StorageError> {
+        let walkdir = WalkDir::new(path)
+            // Don't include the root directory itself
+            .min_depth(1);
+
+        let s =
+            walkdir.into_iter().flat_map(move |result_dir_entry| {
```
wjones127, line 93: Same thing over here: I think `filter_map` makes more sense for this case:

```suggestion
            walkdir.into_iter().filter_map(move |result_dir_entry| {
```

rust/src/storage/file/mod.rs
```diff
@@ -87,33 +84,70 @@ impl StorageBackend for FileStorageBackend {
...
+                match convert_walkdir_result(result_dir_entry) {
+                    Err(e) => Some(Err(e)),
+                    Ok(None) => None,
+                    Ok(entry @ Some(_)) => entry
+                        .filter(|dir_entry| dir_entry.file_type().is_file())
+                        .map(|entry| {
+                            let file_path =
+                                String::from(entry.path().to_str().ok_or_else(|| {
```
wjones127, line 101: This might be bike shedding a bit, but might be nice to have this as a match guard and save a level of indentation. (Note the suggestion requires additional edits to compile.)

```suggestion
                    Ok(Some(entry)) if entry.file_type().is_file() => {
                        let file_path =
                            String::from(entry.path().to_str().ok_or_else(|| {
```

rust/src/storage/file/mod.rs
```diff
@@ -87,33 +84,70 @@ impl StorageBackend for FileStorageBackend {
...
-                        size: Some(meta.len().try_into().map_err(|_| {
-                            StorageError::Generic("cannot convert to i64".to_string())
-                        })?),
-                    })
+    ) -> Result<BoxStream<'a, Result<ObjectMeta, StorageError>>, StorageError> {
+        let walkdir = WalkDir::new(path)
+            // Don't include the root directory itself
+            .min_depth(1);
+
+        let s =
```
wjones127, line 92: nit: one letter variable name?

wjones127 (commented):
Thanks for doing this. I think functionally it looks good, but had a few style comments.

rust/src/storage/file/mod.rs
roeap: I tried to make this flatter, but as is turns out, when doing so, we have to wrap the result returned in this branch in an option. This in turn lead to having to internally deal with all the fallible operations, since we no longer can implicitly convert as we are returning an `Option<Result<..>>`. Tried several permutations, but always felt, the resulting inner branching / nesting complexity was not worth the flatter code...

If you feel very strongly about this, I can try and dig deeper, but would prefer to just leave it as is.

rust/src/storage/file/mod.rs
roeap: very true :)

rust/src/storage/file/mod.rs
roeap: done.

wjones127 (approved):

PR #683, opened 2022-07-10
fix: omit common prefixes in azure `list_objs`

roeap:
# Description
Adls gen2 with hierarchical namespaces enabled has a notion of "directories" in an object store, this is in contrast to other object store implementations. This PR brings the behaviour in line with other storage backend implementations in this crate.

leaving this as a draft to see what else needs to be done for azure to be en par with local and s3 in #682.

# Related Issue(s)
towards #647

# Documentation


Blajda:
Hi @roeap,  The vacuum tests panic with the following error message:
```
`thread 'test_non_partitioned_table' panicked at 'called `Result::unwrap()` on an `Err` value: DeltaTable(Generic("Parent path `adls2://deltarstests/delta-rs-test-1657486937-37104` is not a prefix of path `adls2://delta-rs-test-1657486937-37104/adls2://deltarstests/delta-rs-test-1657486937-37104`"))', rust/tests/vacuum_test.rs:123:35`
```

roeap:
@Blajda - how about now? :)

Blajda:
Still fails due to the prefix but it's closer. Just missing the account name
```
thread 'test_ignored_files' panicked at 'called `Result::unwrap()` on an `Err` value: DeltaTable(Generic("Parent path `adls2://deltarstests/delta-rs-test-1657490502-18515` is not a prefix of path `adls2://delta-rs-test-1657490502-18515/.dotfile`"))', rust/tests/vacuum_test.rs:219:35
```

roeap:
One more iteration ...

Blajda:
Yeah that works. Everything passes now expect for `test_non_managed_files` all the backends are aligned now :partying_face:

rust/src/storage/azure/mod.rs
```diff
@@ -362,12 +362,21 @@ impl StorageBackend for AdlsGen2Backend {
             .directory(obj.path)
             .into_stream()
             .flat_map(|it| match it {
-                Ok(paths) => Either::Left(stream::iter(paths.into_iter().map(|p| {
-                    Ok(ObjectMeta {
-                        path: format!("adls2://{}/{}", self.file_system_name, path.to_owned()),
-                        modified: p.last_modified,
-                        size: Some(p.content_length),
-                    })
+                Ok(paths) => Either::Left(stream::iter(paths.into_iter().flat_map(|p| {
```
wjones127, line 365: Huh I didn't know `flat_map` worked on `Option`. Would have guessed to use `filter_map` usually, and it might be a bit clearer?

```suggestion
                Ok(paths) => Either::Left(stream::iter(paths.into_iter().filter_map(|p| {
```

wjones127 (changes requested):
Looks good, but had a readability suggestion that I think is worth including.

wjones127 (approved):

Issue #592, opened 2022-04-29
Write `enforce_invariant()` function


wjones127:
# Description

For both the datafusion and pyarrow-based writers to support writer protocol v2, we'll need to support enforcing invariants. It seems like the following signature could be reused by both implementations:

```rust
fn enforce_invariant(batch: RecordBatch, invariants: &Vec<(i32, &str)>) -> Result<(), DatafusionError> {
    // rough implementation:
    for (column_index, sql_invariant) in invariants {
        // ... (run data fusion query)
        // ... If failure, return error indicating which invariant failed.
    }
    Ok(())
}
```

Then this function could be applied to each record batch that comes in during a write.

We might also need to check whether the column is nullable and make sure we are enforcing that too, either as part of this or part of the schema enforcement. Should add a test for that.

What do you think @roeap?

**Related Issue(s)**

* #575 

## Related docs

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-invariants
https://books.japila.pl/delta-lake-internals/constraints/Invariants/

wjones127:
BTW here are the invariant test suites: https://github.com/delta-io/delta/blob/master/core/src/test/scala/org/apache/spark/sql/delta/schema/InvariantEnforcementSuite.scala

Also note "invariants" are distinct from "constraints", so there's a separate [test suite](https://github.com/delta-io/delta/blob/master/core/src/test/scala/org/apache/spark/sql/delta/schema/CheckConstraintsSuite.scala) for that. Constraint enforcement is part of writer version 3, so don't need to worry about it yet; But it seems like it has a very similar implementation as invariants, so maybe write this function with constraints in mind.

If I understand correctly, the main difference between invariants and constraints is that the former is a property of a single column, whereas a constraint is a property of the table and thus can enforce relationships *between* columns ([example constraint](https://github.com/delta-io/delta/blob/6905ce757f67935960a9a13ecb6854d53c117d31/core/src/test/scala/org/apache/spark/sql/delta/schema/CheckConstraintsSuite.scala#L197)).

roeap:
At first sight this sounds very reasonable :). With regards to testing, maybe it makes sense to combine that effort with the pyspark integration tests, along the lines of do we get the same results (error or not) when writing here or with pyspark?

wjones127:
Two notes:

 * We are currently waiting on clarification to the protocol: delta-io/delta#1241
[...]

PR #684, opened 2022-07-11
Implement Python Schema in Rust

### LINKED ISSUE #95: Create and bind Delta Schema structures with pyo3 for Python.
Currently, we use a `JSON` string for binding Rust and Python for reading the Delta table schema. The improvement is to create and use pyo3 structure in Rust for binding the Schema.

wjones127:
# Description

These changes move the schema code down into Rust, removing (most of) the JSON serialization and making the Python bindings closer to the Rust implementation. This will make it easier to add methods onto the Rust DeltaSchema and expose them in Python.

It also adds `to_pyarrow()` and `from_pyarrow()` methods to each of the classes, since this was easy to implement.

It does mean some API changes, such as removing the `DataType` base class, because we can't yet properly do inheritance in PyO3. I had to rewrite the unit tests for schemas, but I left the unit tests for all other modules unchanged (minus moving away from a deprecated function).

Also, I moved the main typestubs into the module, which I think means we will ship them now with the code. Haven't confirmed that yet though.

# Related Issue(s)

 - Closes #95

Will also help with #592

# Documentation


houqp:
This is really nice improvement, thanks @wjones127 for picking it up!

rust/src/delta_arrow.rs
```diff
@@ -42,7 +57,7 @@ impl TryFrom<&schema::SchemaTypeArray> for ArrowField {
 
     fn try_from(a: &schema::SchemaTypeArray) -> Result<Self, ArrowError> {
         Ok(ArrowField::new(
-            "element",
+            "item",
```
wjones127, line 60: This and the below changes are necessary to make them inline with the default field names in Rust and C++ Arrow, allowing easy round tripping between Arrow types and Delta Lake types.

python/src/lib.rs
```diff
@@ -1,7 +1,9 @@
-#![deny(warnings)]
+// #![deny(warnings)]
```
roeap, line 1: does this need to be disabled?

python/tests/test_schema.py
```diff
@@ -55,270 +60,194 @@ def test_table_schema_pyarrow_020():
...
+
+
+def test_delta_field():
+    args = [
+        ("x", PrimitiveType("string"), True, {}),
+        ("y", "float", False, {"x": {"y": 3}}),
+        ("z", ArrayType(StructType([Field("x", "integer", True)]), True), True, None),
+    ]
+
+    # TODO: are there field names we should reject?
```
roeap, line 211: the only thing i can think of is when we use the field names for partitioning. We have some test for special characters. Not sure though if we could handle (or should be able to) if an "=" were to appear in the field name.

python/tests/test_schema.py
```diff
@@ -55,270 +60,194 @@ def test_table_schema_pyarrow_020():
...
+#             "name": field_name,
+#             "nullable": False,
+#             "metadata": metadata,
+#             "type": {"name": "floatingpoint", "precision": "HALF"},
+#         }
+#     )
+#     assert pyarrow_field.name == field_name
+#     assert pyarrow_field.type == pyarrow.float16()
+#     assert dict(pyarrow_field.metadata) == metadata
+#     assert pyarrow_field.nullable is False
```
roeap, line 94: is this test still needed?

roeap (approved):
looks great overall! left some minor comments.

python/tests/test_schema.py
wjones127: I think I may leave this for later. There may be complications with column mapping.

python/src/lib.rs
wjones127: Ah thanks. I had disabled earlier while refactoring and forgot to re-enable.

python/tests/test_schema.py
wjones127: Whoops! I had deleted but hadn't added those changes somehow. It's now deleted.

roeap (dismissed):

houqp (approved):
This is really high quality work, thanks @wjones127 !

rust/src/delta_arrow.rs
Tom-Newton: I think this change might cause issues when reading complex nested types. For example a map of arrays gives  
```
pyarrow.lib.ArrowNotImplementedError: Unsupported cast to map<string, list<item: string>> from map<string, list<element: string> ('map_arrays')>
```
It seems PyArrow cannot cast `element` -> `item` for some complex types.

Tom-Newton: I created a test to reproduce this issue https://github.com/Tom-Newton/delta-rs/pull/13

roeap: @wjones127 @Tom-Newton - is this something blocking us from mergi8n, or would this conflict somehow with #714?

Tom-Newton: It's unrelated to #714. But it relates to making map types work correctly.

roeap: great, this mean we can merge 714, right :)?

wjones127: Thanks for the info @Tom-Newton. I'll add that test case to this PR and make sure it is handled.

wjones127: I have reported an upstream issue: https://issues.apache.org/jira/browse/ARROW-17349

For now, I guess I'll see if I can align things with using `element`, since Spark seems to not allow customizing that field name.

wjones127: Tried two approaches and neither worked very well:

 1. First, I tried to change back to always convert Delta Array types to ListTypes with "element" as field name. With that change, we cannot roundtrip from `Arrow -> Delta -> Arrow` without having to re-map the field names, which I think will be annoying beyond just our test cases.
 2. Second, I tried changing so that whenever we read and write, we always use the "element" field name. But PyArrow errors when trying to write a table with a schema not matching the output schema, including field names. And if we try to cast we will run into exactly the same error you are getting now when reading.

Since this case isn't something we already support, I'm thinking the best course of action is to fix the issue upstream. Does that seem acceptable @Tom-Newton ?

Tom-Newton: From my perspective this PR is a big improvement and does not introduce any regression, so I certainly think this is good to merge. However if this particular change was not needed it would have been even more helpful when it comes to making map types work.

I would like to understand why always using `element` doesn't work though.

wjones127: > I would like to understand why always using `element` doesn't work though.

First, there is the annoyance that is breaks roundtripping:

```python
pa_type = pa.list_(pa.int32()) # uses `item` as field name
delta_type = deltalake.schema.ArrayType.from_pyarrow(pa_type) # drops the field name
pa_type_back = delta_type.to_pyarrow() # adds `element` as field name
pa_type == pa_type_back # is False
```

But more importantly, it means that when *writing* to a Delta table, we need to make sure we use the field name "element" for list types. Otherwise when reading, we'll get the same error you were getting, but now complaining it can't cast from "item" to "element" instead of the other way around. That means when we write data, we have to cast any other field name to "element". But casting the field names is what's broken in the first place! So instead the writer would be broken in this edge case.

I guess ultimately it's a tradeoff: if we use "item", we break the reader in a certain edge case. If we use "element", we break the writer in that edge case *and* we break pyarrow type roundtripping. The latter seems slightly worse to me. Either way, we should have this fixed by the PyArrow 10.0.0, which should be released in October.

houqp (approved):

rust/src/delta_arrow.rs
Tom-Newton: Thanks for explaining 🙂 . So if I understand correctly the issue is because `write_deltalake` supports arbitrary pyarrow schemas so depending on whether the user's data uses `item` or `element` a cast may be needed. 

For the round tripping I guess you could make the opposite argument that `pa_type = pa.list_(pa.field("element", pa.int32()))` cannot be round tripped with the current implementation.

Issue #690, opened 2022-07-17
Allow access to underlying storage backend


wjones127:
# Description

We currently don't allow external access to the underlying storage of a `DeltaTable`. But for the Python bindings we will need access to be able to pass the storage backend to other functions that need to do IO (see #689).

The type is currently `Box<dyn StorageBackend>`. We could instead making it `Arc<dyn StorageBackend>`, and then can allow users to retrieve the backend by cloning the Arc pointer. `StorageBackend` is already `Send + Sync`, so I don't think there should be any problem allowing shared ownership.

There are plenty of places where we create a new storage backend from a URI and storage options, when we really should just be taking an `Arc<StorageBackend>`. Here's a good example, IMO:

https://github.com/delta-io/delta-rs/blob/58b26299e09ab344755928001c7e594589cd7a0c/rust/src/writer/json.rs#L182-L190

**Use Case**

In Python, we need to extract the storage backend so we can pass the filesystem to PyArrow functions such as the dataset scanner.

**Related Issue(s)**

* This blocks fixing #689

Issue #689, opened 2022-07-15
Support get_backend_for_uri_with_options in Read Path


joshuarobinson:
# Description

Goal: I am trying to read a Delta Table from a non-AWS S3 object store (like minio). I seem to be able to read the table metadata but then when reading data I get an error that seems to indicate that my endpoint_override is being ignored and requests are sent to aws instead.

My code looks like this:
```
from deltalake import DeltaTable
storage_options = {"AWS_ACCESS_KEY_ID": ACCESS_KEY, "AWS_SECRET_ACCESS_KEY":SECRET_KEY, "AWS_ENDPOINT_URL": ENDPOINT_URL, "AWS_REGION": 'us-east-1'}
dt = DeltaTable("s3://warehouse/tpcds_sf1000_dlt/item/", storage_options=storage_options)
print(dt.dt.pyarrow_schema()) # SUCCESS
print(dt.to_pandas().head()) # FAILS
```

In looking through the code, I wonder if the line here needs to use the "with_options" variant:
https://github.com/delta-io/delta-rs/blob/d10b428f15906dd74f5d33cf592288cdf7dd1e66/python/src/lib.rs#L485
in order to pass through my configs?

**Use Case**

Delta on S3 object stores like Minio, Swift, FlashBlade, Vast, ECS, etc.

**Related Issue(s)**

wjones127:
Thanks for reporting this! Our current API is problematic in that we can't retrieve the storage backend once it's been created in the DeltaTable. We'll need to refactor a bit, probably changing the storage backend to use an `Arc`, instead of an owned `Box` pointer.

For now, the workaround is to explicitly pass a configured PyArrow filesystem:

```python
from deltalake import DeltaTable
from pyarrow.fs import S3FileSystem

storage_options = {
    "AWS_ACCESS_KEY_ID": ACCESS_KEY,
    "AWS_SECRET_ACCESS_KEY":SECRET_KEY,
    "AWS_ENDPOINT_URL": ENDPOINT_URL,
    "AWS_REGION": 'us-east-1'
}
dt = DeltaTable("s3://warehouse/tpcds_sf1000_dlt/item/", storage_options=storage_options)

pa_fs = S3FileSystem(
    access_key = storage_options["AWS_ACCESS_KEY_ID"],
    secret_key = storage_options["AWS_SECRET_ACCESS_KEY"],
    endpoint_override = storage_options["AWS_ENDPOINT_URL"],
[...]

Issue #257, opened 2021-05-22
Refactor DeltaTable to generic over storage backend trait


houqp:
# Description

**Use Case**

Right now DetalTable references storage backend using dynamic dispatch, this could turn into a potential performance bottleneck because our access to storage backend is now tied to the number of files we process from the delta table.

On top of this, using storage backend as trait object also prevents us from adding generic methods to the storage backend trait.

Instead, we should refactor DeltaTable struct to generic over storage backend trait, then return the table struct as a trait object in the high level `open_table` API.

viirya:
I tried this a little bit for curiosity. I saw a few issues when returning the table struct as a trait object. For example, not all fields are public in DeltaTable struct. We rely on the struct to hide internal stuffs. If we have it as a trait object, we will expose these internals.

houqp:
Yeah, very good point @viirya . I now think it might be better to not use DeltaTable as trait object, at least not for the first version, otherwise it will also restrict us from defining generic methods as well.

rtyler:
No longer needed as far as I can tell

PR #697, opened 2022-07-17
feat: sharable reference to storage backend

roeap:
# Description

While working on #632 i saw that a shareable storage backend is also required for #690 - as such i decided to move this to a separate PR, to get things stared for #689.

# Related Issue(s)

# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -24,6 +24,7 @@ use pyo3::types::{PyBytes, PyTuple, PyType};
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::convert::TryFrom;
+use std::sync::Arc;
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
 
@@ -474,7 +475,7 @@ fn filestats_to_expression<'py>(
 
 #[pyclass]
 pub struct DeltaStorageFsBackend {
-    _storage: Box<dyn StorageBackend>,
+    _storage: Arc<dyn StorageBackend>,
 }
 
 #[pymethods]
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -3,6 +3,7 @@
 // Reference: https://github.com/delta-io/delta/blob/master/PROTOCOL.md
 //
 
+use crate::action::{Add, Stats};
 use arrow::error::ArrowError;
 use chrono::{DateTime, Duration, FixedOffset, Utc};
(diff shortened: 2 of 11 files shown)


wjones127 (approved):
Thank you for pulling this out into a separate PR. 🙌 

This looks great to me!

houqp (approved):

houqp:
In the long run, I wonder if it would be better to add a Clone trait constraint to the StorageBackend trait instead.

wjones127:
> In the long run, I wonder if it would be better to add a Clone trait constraint to the StorageBackend trait instead.

Is the expectation there that internally they might use an Arc so that cloning is cheap? Not sure which way is more idiomatic in Rust.

houqp:
> Is the expectation there that internally they might use an Arc so that cloning is cheap? Not sure which way is more idiomatic in Rust.

I think using Arc internally would be the last resort if the upstream sdk client doesn't implement Clone because it incurs extra runtime overheads. All of our current storage backends should be able to derive Clone without using Arcs from a quick glance. Most of the derived clone implementation should be very cheap and if it is not it shouldn't be a problem assuming storage backend clone is a very infrequent operation. On the other hand, accessing the backend itself for read/write is a frequent runtime operation, so it's better to not force the atomic reference counter overhead everywhere only because we need infrequent cheap clones.

If clone is a frequent runtime operation, then wrapping all backend with an Arc is likely the right design choice.

We even have a very old task that talked about getting rid of the trait object constraint from storage backend entirely in https://github.com/delta-io/delta-rs/issues/257 :)

wjones127:
> On the other hand, accessing the backend itself for read/write is a frequent runtime operation, so it's better to not force the atomic reference counter overhead everywhere only because we need infrequent cheap clones.

I was under the impression the increment / decrement overhead just happened in `Arc::clone()` and `Arc::drop()`, and thus we shouldn't have any overhead within objects, just a little bit when we are creating new structs where we need to pass a reference to the filesystem. Is it called also when calling methods of `T` on an `Arc<T>`?

houqp:
> I was under the impression the increment / decrement overhead just happened in Arc::clone() and Arc::drop(), and thus we shouldn't have any overhead within objects,

I was wrong and you are right in this case @wjones127  :) Accessing the heap object should have no overhead, Arc's overhead only comes from clone and drop. So in this case, it won't make a difference for us because clone is a very infrequent operation.

Issue #654, opened 2022-06-21
Python: use file_visitor to get size in write_deltalake()


wjones127:
# Description

In PyArrow 9.0.0 (to be released in a few months) we will have the ability to get file size directly from the visitor (https://github.com/apache/arrow/pull/13338). Currently we make an extra object store call to check the file size after it's written. (Introduced in https://github.com/delta-io/delta-rs/pull/625.)

**Use Case**

Will improve performance of `write_deltalake()`, especially when there are many files written.

**Related Issue(s)**

PR #704, opened 2022-07-19
Get file size from Pyarrow directly (>= 9.0.0)

Bernolt:
# Description

Following up on #654, Get size from written_file if PyArrow >= 9.0.0


Bernolt:
I tested it locally with a local build of arrow 9.0.0. It didn't break the unit-tests.

wjones127 (approved):
Looks good. Thank you!

PR #712, opened 2022-07-26
Fix map type support

Tom-Newton:
# Description
Currently `DeltaTable.to_pyarrow_dataset()` fails for any table containing map types. There are 2 reasons for this:
1. The [python code to parse the schema json](https://github.com/delta-io/delta-rs/blob/347aa472ffa80fcc9a811d837b922431be11681c/python/deltalake/schema.py#L177-L272) returned by the rust code does not support map types correctly.
2. The version of rust arrow used (15) does not support map types. After fixing point 1 we get `ArrowException: C Data interface error: The datatype ""+m"" is still not supported in Rust implementation` from [this line](https://github.com/Tom-Newton/delta-rs/blob/b35048d5b827192e85c01281b5daccb303c88253/python/deltalake/table.py#L299). 

# Related Issue(s)

#713

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,7 +87,33 @@ dependencies = [
  "comfy-table",
  "csv",
  "flatbuffers",
- "half",
+ "half 1.8.2",
+ "hex",
+ "indexmap",
+ "lazy_static",
+ "lexical-core",
+ "multiversion",
+ "num",
+ "rand 0.8.5",
+ "regex",
+ "serde",
+ "serde_derive",
+ "serde_json",
+]
+
+[[package]]
+name = "arrow"
+version = "19.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


wjones127:
Note: I am almost done rewriting the schema stuff here #684. I can add the tests here to make sure it fixes this issue.

roeap:
Hi @Tom-Newton - Thanks for the work on this! Within #703 I started the migration of arrow, but more importantly datafusion. Unfortunately we have to keep those versions aligned. However the simple task of updating datafusion got a bit bigger as they fundamentally changed the internal path handling, and I am right now working on migrating the delta internal path stuff to the new patterns. THis also constitutes an important step towards integrating with the new object store abstractions.

Tom-Newton:
It sounds like I just need to be patient then :slightly_smiling_face: . It seems like everything that is needed is in progress :rocket:.

Tom-Newton:
Closing since there are 2 in progress PRs that I believe will solve this. https://github.com/delta-io/delta-rs/pull/703 https://github.com/delta-io/delta-rs/pull/684

PR #724, opened 2022-08-06
Prune scanned files on column stats

roeap:
# Description

This PR deepens the integration with datafusion by leveraging the column statistics from the delta log to prune the files that need to be scanned when additional constraints are supplied. Luckily datafusion provides some excellent utilities to implement this. Specifically we implement `PruningStatistics` for `DeltaTable`,  use that with `PruningPredicate`, and the rest just kind of happens 😆. 

I still need to do cleanup and more testing as well as have a look how much effort and dependency growth it would be to adopt datafusion expressions for our partition / stats handling. However if there is feedback as to if this is the right way to go, I'd be happy to hear it.

_cc_ @houqp @wjones127 - maybe even @tustvold has some feedback? :) 

# Related Issue(s)

Especially the statistics recorded during scans should get us a lot closer to finishing #632

# Documentation

### DIFF
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -351,7 +351,7 @@ fn to_scalar_value(stat_val: &serde_json::Value) -> Option<datafusion::scalar::S
             }
         }
         serde_json::Value::String(s) => Some(ScalarValue::from(s.as_str())),
-        // TODO is it permissible to encode arrays / objects as partition values
+        // TODO is it permissible to encode arrays / objects as partition values?
         serde_json::Value::Array(_) => None,
         serde_json::Value::Object(_) => None,
         serde_json::Value::Null => None,
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -359,9 +359,11 @@ impl TableProvider for delta::DeltaTable {
                             Some(partitioned_file_from_action(action, &schema))
                         }))
                 })
-                .ok_or(DataFusionError::Execution(
-                    "Failed to evaluate table pruning predicates.".to_string(),
-                ))??
+                .ok_or_else(|| {
+                    DataFusionError::Execution(
+                        "Failed to evaluate table pruning predicates.".to_string(),
+                    )
+                })??
                 .for_each(|f| {
                     file_groups
                         .entry(f.partition_values.clone())
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -26,6 +26,7 @@ use std::sync::Arc;
 
 use arrow::datatypes::{DataType as ArrowDataType, Schema as ArrowSchema, TimeUnit};
 use async_trait::async_trait;
+use chrono::{DateTime, NaiveDateTime, Utc};
 use datafusion::datasource::file_format::parquet::ParquetFormat;
 use datafusion::datasource::file_format::FileFormat;
 use datafusion::datasource::listing::PartitionedFile;
@@ -37,6 +38,7 @@ use datafusion::physical_plan::file_format::FileScanConfig;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::{ColumnStatistics, Statistics};
 use datafusion::scalar::ScalarValue;
+use object_store::{path::Path, ObjectMeta};
 use url::Url;
 
 use crate::action;
@@ -262,6 +264,7 @@ impl TableProvider for delta::DeltaTable {
             url.host_str().unwrap_or_default(),
             self.object_store(),
         );
+        let table_partition_cols = self.get_metadata().unwrap().partition_columns.clone();
 
         // TODO prune files based on file statistics and filter expressions
         let partitions = self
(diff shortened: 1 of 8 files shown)


roeap:
@wjones127 - the python 3.7 builds seem to have started failing here and in other PRs. It seems it tries to build pyarrow from source again and fails to find Arrow C++. While we could install it, my understand is this _should_ not be the case -also everything worked until very recently, any ideas?

rust/src/delta_datafusion.rs
```diff
@@ -310,6 +448,7 @@ fn to_scalar_value(stat_val: &serde_json::Value) -> Option<datafusion::scalar::S
             }
         }
         serde_json::Value::String(s) => Some(ScalarValue::from(s.as_str())),
+        // TODO is it permissible to encode arrays / objects as partition values?
```
houqp, line 451: not for deltalake: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization

houqp:
For the py37 build error, it's because pyarrow 9 stopped releasing manylinux2010 wheels. Compare https://pypi.org/project/pyarrow/9.0.0/#files with https://pypi.org/project/pyarrow/8.0.0/#files. We might need to bump our manylinux support to 2014 too :(

@wjones127 is the manylinux2010 support removal in arrow 9 release expected?

rust/src/delta_datafusion.rs
```diff
@@ -263,29 +345,85 @@ impl TableProvider for delta::DeltaTable {
...
-            .map(|action| {
-                Ok(vec![PartitionedFile::new(
-                    action.path.clone(),
-                    action.size as u64,
-                )])
+            .zip(files_to_prune.into_iter())
+            .filter_map(|(action, prune_file)| {
+                if prune_file {
+                    return None;
+                }
```
houqp, line 369: might be worth specializing this so we are not paying the penalty of iterating and checking prune file array when there is no file to prune. For example, the code below can be abstracted into a function, then we have two types of iterator loops that calls it. One iterator loop zips with files_to_prune, the other one simply just iterates through `get_state().files().`

roeap: Makes sense! Updated it and to the opportunity to batch files by partition values. Thats not going to be optimal for many cases, but by default hopefully better than each file in its own (datafusion-)partition.

rust/tests/datafusion_test.rs
```diff
@@ -4,12 +4,95 @@ mod s3_common;
...
+            &mut self,
+            plan: &dyn ExecutionPlan,
+        ) -> std::result::Result<bool, Self::Error> {
+            if let Some(exec) = plan.as_any().downcast_ref::<ParquetExec>() {
+                let files = get_scanned_files(exec);
+                self.scanned_files.extend(files);
+            }
+            Ok(true)
+        }
+    }
```
roeap, line 64: Just did this for testing right now, but I would like to use something like this to collect the statistics we need for proper conflict resolution.

rust/src/delta_datafusion.rs
```diff
@@ -193,39 +221,86 @@ impl delta::DeltaTable {
...
-//                     })
-//                     .collect(),
-//             ),
-//             is_exact: true,
-//         }
-//     } else {
-//         Statistics::default()
-//     }
-// }
+impl PruningStatistics for delta::DeltaTable {
```
roeap, line 224: Implementing this made me think of how to best store stats, which is an ongoing topic... Maybe `PruningStatistics`s view on the work helps?

Somewhere along the lines of

```rs
pub struct Stats {
    files: HashMap<Path, (usize, PartitionedFile)>,
    max_values: RecordBatch,
    min_values: RecordBatch,
}
```

... and then implement some convenience accessors to get data per column / per file.

rust/src/delta_datafusion.rs
```diff
@@ -263,29 +338,70 @@ impl TableProvider for delta::DeltaTable {
...
-            .collect::<DataFusionResult<_>>()?;
+        // TODO we group files together by their partition values. If the table is partitioned
+        // and partitions are somewhat evenly distributed, probably not the worst choice ...
+        // However we may want to do some additional balancing in case we are far off from the above.
+        let mut file_groups: HashMap<Vec<ScalarValue>, Vec<PartitionedFile>> = HashMap::new();
+        if !filters.is_empty() {
+            combine_filters(filters)
+                .map::<DataFusionResult<_>, _>(|predicate| {
+                    let pruning_predicate = PruningPredicate::try_new(predicate, schema.clone())?;
+                    let files_to_prune = pruning_predicate.prune(self)?;
```
roeap, line 349: I was a bit concerned about interpreting the filters properly, but luckily the hard part was done 😆.

rust/src/delta_datafusion.rs
houqp: yep, switching to columnar format will help in many places :)

rust/src/delta_datafusion.rs
```diff
@@ -263,29 +338,70 @@ impl TableProvider for delta::DeltaTable {
...
+                            }
+                            Some(partitioned_file_from_action(action, &schema))
+                        }))
+                })
+                .ok_or_else(|| {
+                    DataFusionError::Execution(
+                        "Failed to evaluate table pruning predicates.".to_string(),
+                    )
+                })??
+                .for_each(|f| {
```
houqp, line 367: This iterates through the vector again no? I think we should be able to perform the hashmap insertion within filter_map callback above?

houqp:
Thanks @roeap , this is a great demonstration of datafusion's extensibility :) The rest looks good to me, left a very minor comment.

wjones127:
> @wjones127 is the manylinux2010 support removal in arrow 9 release expected?

Yes. Sorry I saw that in the release notes and didn't connect the dots. 

Odd though, I'm unsure why it's being installed in the job that's supposed to only install PyArrow 4.X For some reason `maturin develop` is doing some `pip --force-reinstall` and that's causing an upgrade. I need to look more into this.

rust/src/delta_datafusion.rs
roeap: absolutely! we are avoid that extra iteration now.

houqp (approved):
LGTM!

PR #733, opened 2022-08-09
Python: fix: fix minimal test and bump minimum pyarrow version

wjones127:
# Description

I thought we were testing with `pyarrow==4`, but turns out `maturin develop` silently upgrades all dependencies :facepalm:.

It seems that our actual minimum version is pyarrow 7.0.0 (released in February 2022), which was mostly pushed up by my work on the writer. This PR pushes up the minimum version. If that feels to aggressive, I could refactor so that only the writer module requires this more recent version.

# Related Issue(s)

# Documentation


houqp (approved):
I think the version bump is fine. We can revisit this if we start to get complains from our users :) At the current stage of the project, I think it's better to optimize for speed of iteration, which ultimately will benefit our users more.

PR #742, opened 2022-08-16
fix: Address clippy lint warnings.

### LINKED ISSUE #741: Lint Errors with latest Rust tools
# Environment

**Delta-rs version**:
main

**Binding**: 
python -> make rust_check

**Environment**:
- **Cloud provider**: Vmware VM
- **OS**: ubuntu 20.04
- **Other**: Docker build using `curl https://sh.rustup.rs -sSf` to install rust. I end up with rust 1.63.0 which was release 2022-08-11

***
# Bug

**What happened**: The `make check-rust` target in the python directory failed with lint errors

**What you expected to happen**: I expect the build to pass on the main branch

**How to reproduce it**: Run `make check-rust` from any python environment. It appears to be related to a new release of the rust toolchain.

**More details**:

tsh56:
# Description
This addresses clippy lint warnings that appeared with the 1.63.0 release of rust. Most of the changes are reasonable and straight forward. I'm unsure about the best way to resolve python/src/lib.rs. I whitelisted the check, as py03 wants pymethods to have parameters with known sizes as compile time.

# Related Issue(s)
- closes #741 

# Documentation
https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq
https://rust-lang.github.io/rust-clippy/master/index.html#borrow_deref_ref

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -1,4 +1,5 @@
 #![deny(warnings)]
+#![allow(clippy::borrow_deref_ref)]
 
 extern crate pyo3;
 
--- a/rust/src/vacuum.rs
+++ b/rust/src/vacuum.rs
@@ -236,7 +236,7 @@ impl Vacuum {
             });
         }
 
-        return plan.execute(table).await;
+        plan.execute(table).await
     }
 }
 
--- a/dynamodb_lock/src/lib.rs
+++ b/dynamodb_lock/src/lib.rs
@@ -81,7 +81,7 @@ pub mod dynamo_lock_options {
 /// Configuration options for [`DynamoDbLockClient`].
 ///
 /// Available options are described in [dynamo_lock_options].
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct DynamoDbOptions {
     /// Partition key value of DynamoDB table,
     /// Should be the same among the clients which work with the lock.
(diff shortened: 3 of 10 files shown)


rust/src/vacuum.rs
```diff
@@ -236,7 +236,8 @@ impl Vacuum {
             });
         }
 
-        return plan.execute(table).await;
+        //return plan.execute(table).await;
```
houqp, line 239: looks like this should be deleted

python/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(warnings)]
+#![allow(clippy::borrow_deref_ref)]
```
houqp, line 2: curious which line is violating this error? we should probably fix that instead.

tsh56: The issue is directly related to pyo3.  There's already a [ticket](https://github.com/PyO3/pyo3/issues/2555) filed, and it appears to be fixed in main. Once the fixed is release this clippy exception shouldn't be necessary.

fvaleye: As a follow-up before merging, could you please add an issue to delete this clippy exception later?

fvaleye:
Let me do it for you!

python/src/lib.rs
houqp: Second to what @fvaleye said, would be good to add a follow up issue on github. Also a good practice to add comment in the code for temp workarounds that we plan to remove in the future.

PR #762, opened 2022-08-22
Bump version of the Python binding to 0.6.0

fvaleye:
# Description
- Prepare the version [0.6.0](https://github.com/delta-io/delta-rs/releases/tag/untagged-3b7e7d8aa27e35a0a550) of the Python binding
- Bump maturin to version `0.13.2`

PR #770, opened 2022-08-26
Remove Python version 3.6 support And run multiple python versions

### LINKED ISSUE #768: Missing required package typing_extensions under Python 3.10
# Environment

**Delta-rs version**:

**Binding**: python 3.10

**Environment**:
- **Cloud provider**: -
- **OS**: Ubuntu 22.04
- **Other**: - 

***
# Bug

**What happened**: 
After installing the python package under python 3.10.6 in a new venv, the package cannot be imported:

```{python
import deltalake
> ...deltalake\writer.py line 29, in <module>
>   from typing_extensions import Literal
> ModuleNotFoundError: No module named 'typing_extensions'
```

Import should work also under Python 3.10

**How to reproduce it**:
1. create a new venv using python 3.10
2. pip install deltalake
3. python -c 'import deltalake'

**More details**:
After installing typing-extensions, the package deltalake can be imported without errors.

fvaleye:
# Description
- Remove the support of Python version `3.6`
- Add a Github Action to ensure the running of deltalake with multiple versions of Python 
- Change usage of `typing_extensions` by `typing` 

# Related Issue(s)
- closes #768
- #762

# Documentation
- https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module


roeap:
I think we need to update all occurrences of `typing_extensions` in our code. like here:

https://github.com/delta-io/delta-rs/blob/dbc2994c5fddfd39fc31a8f9202df74788f59a01/python/deltalake/deltalake.pyi#L4

To use the imports from the `typings` module.

wjones127 (approved):
Thanks for doing this!

PR #772, opened 2022-08-29
turn table state version into a private field

houqp:
# Description

Semantically, table version is specific to a particular table state, so it doesn't make much sense to let delta-rs users change it from outside. In general, if fields should be declared as private when there are invariants that need to be maintained between them. In this case, version is dependent on the rest of the table state, so mutations need to go through methods to make sure are they always in sync.


fvaleye (approved):
Good catch!

PR #774, opened 2022-08-30
chore: cleanup

roeap:
# Description

This PR removes some no longer needed files, as well as the now outdated docs for ADLS Gen2 (we now support blob as well).

Additionally I removed the rust_dataframe support, since the repo has been archived, and tests were removed in the parquet2 PR.

# Related Issue(s)

# Documentation


rust/src/lib.rs
```diff
@@ -97,8 +97,6 @@ pub mod vacuum;
 pub mod checkpoints;
 #[cfg(all(feature = "arrow", feature = "parquet"))]
 pub mod delta_arrow;
-#[cfg(feature = "rust-dataframe-ext")]
```
wjones127, line 100: Could you also update the doc comment at the top of the module, which still mentions ADLS2?

wjones127 (commented):
BTW have you double checked that the docs can be built with all storage backend features enabled? I'm not sure [this reference](https://github.com/delta-io/delta-rs/blob/8d6e4354d3626b5f43b1162666057310f3777e40/rust/src/builder.rs#L173) still exists for example.

wjones127 (approved):

PR #784, opened 2022-09-02
cleanup errors and make `DeltaWriterError` internal

### LINKED ISSUE #118: Add high level read API to return file content as stream of bytes
This makes it easier for downstream client to load table content because they won't need to write specific code for each storage backend.

roeap:
# Description

## Update

I have been working again on the high level writer APIs. Specifically how we use data fusion and trying to move towards a structure that can more easily integrate with other query backends. These changes are fairly well contained in the `operations` module, But they do include the changes here. While mostly cosmetic so far, I was hoping to merge this PR as is, as to make review in the follow ups easier.. Right now the main focus is to end up somewhere we we can finally finish #632. 

<hr/>

This in preparation to make a pass through the `RecordBatchWriter` and some convenience around loading partitioned tables,
Publishing this already to get some feedback on the general error handling approach.  I took the same approach in the updated file and s3 object stores. Essentially the idea would be that we aim to make all module specific error (in this case the `DeltaWriterError`) private to that crate and eventually expose a single error type that contains all error we expect users to take direct action on, and a Generic error type that transports all non-surfaces mod errors. The main motivation is to simplify error handling for consumers, as we have by now some quite nested scenarios ... 

I guess there are several other alternatives to avoid some of the nesting in errors, but I personally have grown to like that approach working on object_store, and seems to me that from all alternatives that come to my mind, this is the ones that we can most easily transition to from where we are.

cc @houqp @wjones127 

# Related Issue(s)

a step towaards #118

# Documentation

### DIFF
--- a/rust/src/action/parquet2_read/mod.rs
+++ b/rust/src/action/parquet2_read/mod.rs
@@ -38,7 +38,7 @@ pub enum ParseError {
     InvalidAction(String),
     /// Error returned when parsing checkpoint parquet using parquet2 crate.
     #[error("Failed to parse parquet: {}", .source)]
-    ParquetError {
+    Parquet {
         /// Parquet error details returned when parsing the checkpoint parquet
         #[from]
         source: parquet2::error::Error,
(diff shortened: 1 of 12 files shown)


roeap:
@houqp, we are hitting the macos-10,15 brownout again. With the object store PR, I added macos-11, but the 10.15 ones are still marked as required, and I think I cannot remove them, as long as that is the case?

houqp:
@roeap I have removed 10.5 from required checks

houqp (approved):

Issue #786, opened 2022-09-05
BlobNotFound on to_pandas() or to_pyarrow_table() using Azure (ADLS gen 2)


dyrx:
# Environment

**Delta-rs version**: 0.6.0

**Binding**: Python

**Environment**:
- **OS**: Windows

***
# Bug

When using ADLS Gen 2 storage, the `to_pandas()` and `to_pyarrow_table()` methods produce a BlobNotFound error.
This is true for all suggested address formats in the [usage docs](https://github.com/delta-io/delta-rs/blob/main/python/docs/source/usage.rst):

> az://\<container\>/\<path\>
> adl://\<container\>/\<path\>
> abfs://\<container\>/\<path\>

```python
from deltalake import DeltaTable

storage_options = {}
storage_options["AZURE_STORAGE_ACCOUNT_NAME "] = "XXXXXXXX"
storage_options["AZURE_STORAGE_ACCOUNT_KEY "] = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
container = "my_container"
path_to_delta = "path/to/my/delta/table"

dt = DeltaTable(f"az://{container}/{path_to_delta}", storage_options=storage_options) # Succeeds
dt.schema() # Succeeds
dt.history() # Succeeds
df = dt.to_pandas() # Fails
```

**What happened**:

The .to_pandas() method seems to produce a malformed address:

> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "C:\Users\<MY USER>\AppData\Local\pypoetry\Cache\virtualenvs\sandbox-oS5I-KK--py3.10\lib\site-packages\deltalake\table.py", line 347, in to_pandas
>     return self.to_pyarrow_table(
>   File "C:\Users\DYEREX\AppData\Local\pypoetry\Cache\virtualenvs\sandbox-oS5I-KK--py3.10\lib\site-packages\deltalake\table.py", line 329, in to_pyarrow_table
>     return self.to_pyarrow_dataset(
>   File "pyarrow\_dataset.pyx", line 331, in pyarrow._dataset.Dataset.to_table
>   File "pyarrow\_dataset.pyx", line 2577, in pyarrow._dataset.Scanner.to_table
>   File "pyarrow\error.pxi", line 144, in pyarrow.lib.pyarrow_internal_check_status
>   File "pyarrow\_fs.pyx", line 1544, in pyarrow._fs._cb_open_input_file
[...]

Issue #778, opened 2022-09-01
read delta table from GCP regression in 0.6.0


djouallah:
# Environment

**Delta-rs version**:0.6.0

**Binding**: Python 

**Environment**:
- **Cloud provider**: Colab
- **OS**:Linux
- **Other**:

***
# Bug

```
from deltalake import DeltaTable
dt = DeltaTable("gs://test_delta1/delta").to_pyarrow_dataset()
print ('number of files ' + str(len(dt.files)))
dt.head(1)
```
now generate this erros
PyDeltaTableError: Object at location gs:/test_delta1/delta/Date=2022-01-08/0-745c11c3-5354-4b89-b476-dba8dbdadf86-0.parquet not found: response error "No such object: test_delta1/gs:/test_delta1/delta/Date=2022-01-08/0-745c11c3-5354-4b89-b476-dba8dbdadf86-0.parquet", after 0 retries: HTTP status client error (404 Not Found) for url

wjones127:
We just swapped out our GCS implementation, and need to setup automated integration testing for it. cc @roeap

roeap:
I'll look into enabling integration tests for gcs.

roeap:
Quick update. Testing our gcs backed object store was almost straight forward. Turns out there seems to be a bug upstream with the *_if_not_exists methods. Hopefully I can resolve that though since we need an upstream PR anyhow.

However this does indicate, that the error "might" be somewhere on the python side.

@djouallah could you see what happens if you do not convert the table to a pyarrow dataset and just print the version?

djouallah:
@roeap 
![image](https://user-images.githubusercontent.com/12554469/187882667-a652fd68-027f-48c0-9406-098975305704.png)

roeap:
We have some initial integration tests running (#779), So far this does not include the tests where we rely on if_not_exists behaviour. This should however already cover both the read and write path on the rust side.

So far my best guess is that somehow the paths on the python side got messed up when we did the switch to `object_store`.

wjones127:
Once you have integration tests for Rust setup, I can work on having some Python-specific ones and should be able to fix this.

roeap:
WE could just merge the PR currently in draft. In parallel I could see upstream to get the if_not_exists stuff fixed. But some integration tests are running and passing :)

roeap:
@djouallah - we just released `0.6.1`, could you check if this gets resolved?

djouallah:
[...]

PR #787, opened 2022-09-05
hotfix: python object store paths

roeap:
# Description

After the migration to `object_store`, the pyarrow integration broke when using remote storages. The root cause is that paths are handled a bit differently by object_store, and these differences were not accounted for in all scenarios.

Specifically, when using the `DeltaStorageFsBackend`, we are passing in the full path including cloud prefixes - e.g. `s3://<bucket>/<path>`. Object store now expects paths relative the the object store root. For remote stores, the root is always the bucket level, for local it can be anywhere, but we make sure its always the file system root. (note the `DeltaObjectStore` wrapping the backend is always rooted in the table root).

Since recently we have a new struct `StorageUrl`, which encapsulates all the logic of url parsing specific to our object store.

While there is an effort to properly integrate storages, so we have seamless interop between delta-rs and the pyarrow world, spearheaded by @wjones127, I think we should deploy an hotfix, to remedy the immediate issues users are experiencing.

The following changes are included.

- We pass any string passed to `head_obj` and `get_obj` through the `StorageUrl` to extract the proper path for the storage backend. THe term prefix is admittedly misleading here :)
- When possible we create the Storage backend from the table, to have consistency in terms of storage options
- There is a (mostly redundant) Path to initialize the storage backend with options now, this is mainly useful for testing / experimentation.

P.S. for me the import of `configure_spark_with_delta_pip`, did not work, so updated it.

# Related Issue(s)

should close after feedback: #786,  #600, #778

# Documentation

### DIFF
--- a/python/deltalake/fs.py
+++ b/python/deltalake/fs.py
@@ -11,8 +11,13 @@ class DeltaStorageHandler(FileSystemHandler):
     DeltaStorageHander is a concrete implementations of a PyArrow FileSystemHandler.
     """
 
-    def __init__(self, table_uri: str) -> None:
-        self._storage = DeltaStorageFsBackend(table_uri)
+    def __init__(
+        self,
+        table_uri: str,
+        options: Optional[Dict[str, str]] = None,
+        backend: Any = None,
+    ) -> None:
+        self._storage = backend or DeltaStorageFsBackend(table_uri, options)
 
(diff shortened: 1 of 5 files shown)


houqp (approved):

djouallah:
i can't test it without a python binary

PR #799, opened 2022-09-11
feat: integrate `object_store` for read/write with pyarrow

### LINKED ISSUE #574: Python: Support passing filesystem in write_deltalake
# Description

`write_deltalake` has an argument for filesystem, but it's not used yet.  

**Related Issue(s)**
Part of umbrella issue: #542 
Blocked by: #570

### LINKED ISSUE #689: Support get_backend_for_uri_with_options in Read Path

### LINKED ISSUE #696: We pass scheme prefixes (s3://) to PyArrrow
# Description

PyArrow Filesystems don't expect paths to be provided as URIs with a scheme prefix (e.g. `s3://`). The only reason that `DeltaTable.to_pyarrow_dataset()` works (if you don't pass a filesystem) is that our `DeltaStorageHandler` accepts URIs.

We need to change so that:

 1. `DeltaStorageHandler` rejects scheme prefixes
 2. We always pass file paths to PyArrow functions *without* scheme prefixes (strip them out)

# Environment

**Delta-rs version**:

**Binding**:  Python

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

**What happened**:

See: https://github.com/delta-io/delta-rs/issues/689#issuecomment-1186576157

**How to reproduce it**:

**More details**:

roeap:
# Description

This PR embraces the object store crate also on the python side and at least for the current test base supports reading and writing using the dataset and other pyarrow APIs. Thanks to @wjones127 's mutipart upload in object_store, implementing the write functionality was actually quite straigt forward. We now implement `ObjectInputFile` and `ObjectOutputStream`, which - if wrapped in `pyarrow.PythonFile` will work with the arrow ecosystem (so far ;))

There was on bigger and breaking deign decision, but I hope people agree :).

Essatially I just accepted, that working exclusively with the relative delta paths makes life much more convenient.. As such rather then adding and removing paths prefixes all the time, I thought it would be reasonable to as users to wrap their own filesystems in a `pyarrow.fs.SubTreeFileSystems`, which points at the table root..

```py
import pyarrow.fs as fs

path = "<path/to/table>"
filesystem = fs.SubTreeFileSystem(path, fs.LocalFileSystem())

dt = DeltaTable(path)
ds = dt.to_pyarrow_dataset(filesystem=filesystem)
```

Of course the hope is to eventually be at comparable (or higher :)) performance then the c++ file systems. Then there would be little reason (I guess) to still provide a "custom" file system at all.

# Related Issue(s)

closes (at least to some degree) #570 
closes #574
closes #696
closes  #689
towards #542

# Documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -690,6 +690,7 @@ dependencies = [
  "chrono",
  "deltalake",
  "env_logger",
+ "futures",
  "lazy_static",
  "pyo3",
  "regex",
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -14,15 +14,17 @@ name = "deltalake"
 crate-type = ["cdylib"]
 
 [dependencies]
-tokio = { version = "1", features = ["rt-multi-thread"] }
+chrono = "0"
 env_logger = "0"
+futures = "0.3"
+lazy_static = "1"
+regex = "1"
+serde_json = "1"
+tokio = { version = "1", features = ["rt-multi-thread"] }
+
 # reqwest is pulled in by azure sdk, but not used by python binding itself
 # for binary wheel best practice, statically link openssl
 reqwest = { version = "*", features = ["native-tls-vendored"] }
-serde_json = "1"
-chrono = "0"
-regex = "1"
-lazy_static = "1"
 
 [dependencies.pyo3]
 version = "0.16"
--- a/python/deltalake/deltalake.pyi
+++ b/python/deltalake/deltalake.pyi
@@ -7,12 +7,12 @@ else:
     from typing_extensions import Literal
 
 import pyarrow as pa
+import pyarrow.fs as fs
 
 from deltalake.writer import AddAction
 
 RawDeltaTable: Any
 rust_core_version: Callable[[], str]
-DeltaStorageFsBackend: Any
 
 class PyDeltaTableError(BaseException): ...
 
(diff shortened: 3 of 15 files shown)


houqp (approved):
LGTM, great work!

PR #807, opened 2022-09-13
Updated the object_store version to 0.5

### LINKED ISSUE #781: Remove cargo.patch for `object_store` crate
# Description

Currently we patch the version of the `object_store` crate to take advantage of the removed S3 and azure SDKs, once the new version is published (apache/arrow-rs#2620), we need to remember to remove the patch from Cargo.,toml

**Use Case**

**Related Issue(s)**

kesavkolla:
# Description
Updated the obect_store crate to latest version 0.5

# Related Issue(s)

# Documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,33 +21,33 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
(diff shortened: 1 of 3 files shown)


roeap:
@kesavkolla, thanks for the PR!. There is a few more adjustments needed though. We also need to update ci/cd as in the new version the gcs client no longer uses an unsecured connection, but rather fake-gcs needs to run with http. The services in github action do not provide that config. The object_store integration tests are a good reference on how to do this. 

Also, the builder for gcs slightly changed ...

kesavkolla:
@roeap I am so sorry; I am not familiar with github workflows. The object_store at least has some docker commands to run the fake-gcs server. In this repo there is no docker run command.

roeap:
exactly, right now we use the github action integration for running services right now. However this does not have an option to pass in alternative args. So we have to switch to running the services "manually" as is done in object_store right now. Here we would probably adjust the docker-compose file and just start that before running integration tests.

roeap:
superseded  by #814

Issue #813, opened 2022-09-15
Use `ParquetRecordBatchStreamBuilder` for reading parquet files


roeap:
# Description

The `SerializedFileReader` and associated stucts have been deprecated in `arrow-rs`. We need to adopt the new builders and should leverage the async functionality that comes with it.

**Use Case**

**Related Issue(s)**

roeap:
this has been implemented.

PR #814, opened 2022-09-15
chore: bump arrow and friends

### LINKED ISSUE #757: Remove PyO3 clippy workaround once 0.17 is release
# Environment

**Delta-rs version**:
main branch

**Binding**: 
Python

**Environment**:
- **Cloud provider**: NA
- **OS**: linux
- **Other**: NA

***
# Bug

**What happened**: A clippy inner attribute was added to work around a lint error that comes from the PyO3 project in conjunction with the Rust 1.63.0 release. 

**What you expected to happen**: The project should build without lint checks disabled.

**How to reproduce it**: N/A 

**More details**:  The [PyO3 ticket](https://github.com/PyO3/pyo3/issues/2555) is expected to be resolved Tuesday 2022-08-23 with the [0.17 Release](https://github.com/PyO3/pyo3/pull/2575).

**Fix**: The following [line](https://github.com/delta-io/delta-rs/blob/main/python/src/lib.rs#L2) should be removed, and the build is expected to succeed.

### LINKED ISSUE #781: Remove cargo.patch for `object_store` crate

roeap:
# Description

This PR bumps the arrow-family dependencies, and makes the necessary updates. Most notably. we can no longer use the github services, but need to return to using the compose file for runnig emulators. Reason being we cannot pass arguments to fake-gcs to enable http otherwise.

The `SerializedFileReader` has been deprecated, I created #813 to track the updates to the readers. I omitted that in this PR to keep things small. 

superceds #807

## Update

turns out, the updates let the compiler to also complain a naming ambiguity with the native python module. I thus renamed it to `_internal`. 

# Related Issue(s)

closes #781
closes #757

# Documentation


python/Cargo.toml
```diff
@@ -13,6 +13,9 @@ keywords = ["deltalake", "delta", "datalake", "pandas", "arrow"]
 name = "deltalake"
 crate-type = ["cdylib"]
 
+[package.metadata.maturin]
+name = "deltalake._internal"
```
wjones127, line 17: Thanks! I was thinking we should rename this as well to indicate it's not part of the public API.

wjones127:
I see some warnings in the CI job such as

```
warning: function `setup_azure_gen2_context` is never used
  --> rust/tests/common/adls.rs:20:14
   |
20 | pub async fn setup_azure_gen2_context() -> TestContext {
   |              ^^^^^^^^^^^^^^^^^^^^^^^^
```

Should we clean that up here too? Or is that unrelated?

wjones127 (approved):
These changes looks good, but consider also fixing the warnings I mentioned.

roeap:
The warning is unrelated to the changes in this PR. That being said, they definitely should be taken care of, but this requires some more cleanup, specifically in the test fixtures we use for integration tests. My plan is to take care of this when updating the writer module.

PR #822, opened 2022-09-17
feat: improve storage configuration

### LINKED ISSUE #766: Azure SAS support
# Description

Currently, authentication to Azure ADLS Gen2 is supported only via the **storage account key**. As the key is very powerful, in many cases it would be more secure to use SAS tokens for more restricted access.

**Use Case**

Alternative, more secure way of authenticating to Azure ADLS Gen2 service.

**Related Issue(s)**
- #767

roeap:
# Description

This PR mainly improves the handling of configuration - hopefully. In essence we get closed to the options usually used in fsspec or the azure sdk, while retaining the current key names and the option to configure via the environment.

Previously I also believe that in cases where an applicable environment variable was defined, it may have taken precedence over alternative configuration passed to the Delta table.

For testing purposed the in-memory store can now also be used as a backend to delta tables. (I am working on updating the operations, and bringing in changes piece-by-piece to keep PR size reasonable).

For now I left out migrating the S3 config, as it has a bit more logic in it. Also since the currently documented way is still supported, I have not yet updated the documentation. I plant to do both in a follow up PR,

# Related Issue(s)

closes #766

# Documentation

### DIFF
--- a/rust/src/builder/azure.rs
+++ b/rust/src/builder/azure.rs
@@ -209,14 +209,8 @@ impl AzureConfig {
 
 fn parse_boolean(term: &str) -> Option<bool> {
     match term.to_lowercase().as_str() {
-        "true" => Some(true),
-        "1" => Some(true),
-        "yes" => Some(true),
-        "y" => Some(true),
-        "false" => Some(false),
-        "0" => Some(false),
-        "no" => Some(false),
-        "n" => Some(false),
+        "true" | "1" | "yes" | "y" => Some(true),
+        "false" | "0" | "no" | "n" => Some(false),
         _ => None,
     }
 }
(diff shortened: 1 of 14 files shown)


rust/src/builder/azure.rs
```diff
@@ -0,0 +1,245 @@
...
+        Err(BuilderError::MissingCredential.into())
+    }
+}
+
+fn parse_boolean(term: &str) -> Option<bool> {
+    match term.to_lowercase().as_str() {
+        "true" => Some(true),
+        "1" => Some(true),
+        "yes" => Some(true),
+        "y" => Some(true),
```
houqp, line 215: I think you can use `|` here: `"true" | "1" | "yes" | "y"`.

houqp (approved):

PR #823, opened 2022-09-18
Datafusion-imports

roeap:
# Description

Going forward we may introduce a dependency on `datafusion_expr`, or other sub crates, to use for handling predicates etc. AFAIK, the datafusion crate is also going to deprecate re-exports of some of these sub-crates. To be prepared for this, we refactor the imports in `delta-rs` to import from the sub-crates.

As a drive-by I also consolidated some helper functions in `storage::utils`. 

# Related Issue(s)

# Documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -722,6 +722,8 @@ dependencies = [
  "chrono",
  "criterion",
  "datafusion",
+ "datafusion-common",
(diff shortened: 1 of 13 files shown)


wjones127 (approved):
These changes looks good. Thanks for doing this!

PR #834, opened 2022-09-22
Add invariant enforcement support

### LINKED ISSUE #575: Python: Support writer protocol V2 in write_deltalake

### LINKED ISSUE #592: Write `enforce_invariant()` function

wjones127:
# Description

Adds support to retrieve invariants from the Delta schema and also a struct `DeltaDataChecker` to use DataFusion to check them and report useful errors.

This also hooks it up to the Python bindings, allowing `write_deltalake()` to support Writer Protocol V2.

I looked briefly at the Rust writer, but then realized we don't want to introduce a dependency on DataFusion. We should discuss how we want to design that API. I suspect we'll turn DeltaDataChecker into a trait, so we can have a DataFusion one available but also allow other engines to implement it themselves if they don't wish to use DataFusion.

# Related Issue(s)

- closes #592
- closes #575

# Documentation

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-invariants

### DIFF
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -603,6 +603,18 @@ fn left_larger_than_right(left: ScalarValue, right: ScalarValue) -> Option<bool>
     }
 }
 
+fn enforce_invariants(
+    record_batch: RecordBatch,
+    invariants: &Vec<(String, String)>,
+) -> Result<(), DeltaTableError> {
+    let ctx = SessionContext::new();
+    // TODO: How does one query a record batch in data fusion??
+    for invariant in invariants.iter() {
+        let sql = format!("SELECT {} FROM data WHERE {} LIMIT 1", name, invariant);
+    }
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -47,6 +47,12 @@ use datafusion_expr::{combine_filters, Expr};
 use object_store::{path::Path, ObjectMeta};
 use url::Url;
 
+use crate::action;
+use crate::delta;
+use crate::schema;
+use crate::DeltaTableError;
+use crate::Invariant;
+
 impl From<DeltaTableError> for DataFusionError {
     fn from(err: DeltaTableError) -> Self {
         match err {
@@ -607,7 +613,7 @@ fn left_larger_than_right(left: ScalarValue, right: ScalarValue) -> Option<bool>
 /// Checks that the record batch adheres to the given invariants.
 pub fn enforce_invariants(
     record_batch: &RecordBatch,
-    invariants: &Vec<(String, String)>,
+    invariants: &Vec<Invariant>,
 ) -> Result<(), DeltaTableError> {
     // Invariants are deprecated, so let's not pay the overhead for any of this
     // if we can avoid it.
@@ -622,8 +628,8 @@ pub fn enforce_invariants(
 
     let mut violations: Vec<String> = Vec::new();
 
-    for (column_name, invariant) in invariants.iter() {
-        if column_name.contains(".") {
+    for invariant in invariants.iter() {
+        if invariant.field_name.contains('.') {
             return Err(DeltaTableError::Generic(
                 "Support for column invariants on nested columns is not supported.".to_string(),
             ));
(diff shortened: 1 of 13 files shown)


rust/src/delta_datafusion.rs
```diff
@@ -603,11 +606,84 @@ fn left_larger_than_right(left: ScalarValue, right: ScalarValue) -> Option<bool>
...
+            if invariant.field_name.contains('.') {
+                return Err(DeltaTableError::Generic(
+                    "Support for column invariants on nested columns is not supported.".to_string(),
+                ));
+            }
+
+            let sql = format!(
+                "SELECT {} FROM data WHERE not ({}) LIMIT 1",
+                invariant.field_name, invariant.invariant_sql
+            );
```
houqp, line 657: how about using the dataframe api directly to skip the sql parsing overhead ;)

wjones127: I had started down that path, but I couldn't figure out how to pass a SQL string into the `.filter()` method. Any ideas on that?

houqp: ha i see, you would need to call `sql_to_rex` to parse the expression, it's not worth it then in this case :D

houqp:
> I suspect we'll turn DeltaDataChecker into a trait, so we can have a DataFusion one available but also allow other engines to implement it themselves if they don't wish to use DataFusion.

I think this is a good idea :+1:

wjones127:
BTW I just measured, and adding the `datafusion-ext` feature into the Python package increases the (release) wheel size from 11MB to 18MB. I think that's fine, but we might keep an eye on the size and look out for opportunities to slim in down.

rust/src/delta_datafusion.rs
```diff
@@ -603,11 +606,84 @@ fn left_larger_than_right(left: ScalarValue, right: ScalarValue) -> Option<bool>
...
+        if self.invariants.is_empty() {
+            return Ok(());
+        }
+
+        let table = MemTable::try_new(record_batch.schema(), vec![vec![record_batch.clone()]])?;
+        self.ctx.register_table("data", Arc::new(table))?;
+
+        let mut violations: Vec<String> = Vec::new();
+
+        for invariant in self.invariants.iter() {
```
houqp, line 647: Did you test whether it would be faster to concatenate all invariants into a single filter clause using `OR`?

houqp (dismissed):
Great work!

houqp:
> BTW I just measured, and adding the datafusion-ext feature into the Python package increases the (release) wheel size from 11MB to 18MB

If there are particular fat in datafusion that we don't need, we could send upstream PRs to gate those code by features.

Also now that datafusion has become a hard dependency for the python binding, we can start to expose fancy query capabilities in the python libraries. For example, we can now add a `sql` method to the python delta table class to allow users query delta tables directly by passing a sql query!

roeap (dismissed):
Really excited to to see this feature land! Great job!

Left a comment about creating the tokio runtime, but we can figure that out once we integrate this into the rust write path.

houqp:
Yes, agree with Robert that making these methods async and manage the tokio
runtime outside of our rust core is the right thing to do here.

roeap:
> I looked briefly at the Rust writer, but then realized we don't want to introduce a dependency on DataFusion. We should discuss how we want to design that API. I suspect we'll turn DeltaDataChecker into a trait, so we can have a DataFusion one available but also allow other engines to implement it themselves if they don't wish to use DataFusion.

I have been struggling with this for some time, and am currently experimenting with an updated writer implementation that can honor some of the delta settings. We do have a writer trait already, but right now that is simply used to have support for different types of data being written (json, record_batch, ..) Maybe that writer trait should eventually expose functions for checking invariants and constraints as well a `max_supported_version`. Then we could hopefully keep datafusion optional and enable higher writer versions with a "reasonable" amount of features gates on the implementation.

rust/src/delta_datafusion.rs
wjones127: Oh sorry I missed this. I did it separately so that we can provide the user the exact value that violated an invariant. But if it become a performance problem we can re-evaluate. Probably will care more for Constraits than invariants.

houqp (approved):

PR #842, opened 2022-09-25
use rustls for delta checkpoint lambda

houqp:
also test rustls feature in CI


roeap (dismissed):

wjones127 (dismissed):

houqp:
Had to disable integration tests when rustls is enabled because it doesn't support http :(

roeap (approved):

PR #852, opened 2022-09-27
feat: rewrite operations

roeap:
# Description

This PR incorporates some of the learnings with regards to how datafusion "should" be used (or what I think I understood so far how it should be used) and how this applies to our operations module. It also embraces the `IntoFuture` trait stabilized in rust `1.64`. 

More specifically:

- Model operations as builders that implement `IntoFuture`
- While we figure out how to deal with state sharing - just consume the table as it needs to be updated anyhow (right?)
- push datafusion dependencies into command implementations, so we can more easily extend to commands that do not require datafuison.
- have command-spcecific errors that map to `DeltaTableError` (the top level error variants likely have to be refined, for now, many command errors map to a `GenericError`, which at least shows meaningful error messages.)
- an updated `PartitionWriter` implementation that allows more fine graunlar control on how data is written

Some of choices or trials here mainly (hopefully) make sense when viewed as preparation for what's to come. 

- For the conflict resolution (#632), we want to leverage some information on especially the scanned files during an operation. The plan is to leverage the execution metrics from the datafusion plans for this. We already do this in some of the datafusion tests, to make sure we properly use the file metrics in scans.
- The new partition writer implementation is kept separate from the existing one for now, to be able to more easily iterate.
- I replicated a minimal implementation of the transactions, but the main update for this will follow with conflict resolution.

To keep it somewhat simpler to review I tried to keep the major changes contained in the operations module. If we adopt this the idea is to migrate all existing operations (create, vacuum, optimize) to the builder pattern and into the operations module, and have the methods on `DeltaTable` just return a pre-populated builder. I think this is where `IntoFuture` shines, as we can await them just like before.

Implementing this, one of the things I found most convenient, is that we now have the `memory://` store available. Coupled with the `sync_stores` helper, it makes setting up and validating test cases that mutate an existing table very convenient - yay `object_store` 😆. 

There are some things I want to clean up, but it would be great to get some feedback on if this is where we want to go - @houqp @wjones127 @fvaleye.

# Related Issue(s)

# Documentation


rust/src/operations/create.rs
```diff
@@ -1,251 +1,367 @@
...
+                StorageUrl::parse(self.location.ok_or(CreateError::MissingLocation)?)?;
+            (
+                DeltaTableBuilder::from_uri(&storage_url)
+                    .with_storage_options(self.storage_options.unwrap_or_default())
+                    .build()?,
+                storage_url,
+            )
+        };
+
+        // TODO configure more permissive versions based on configuration. Also how should this ideally be handled?
```
wjones127, line 202: We should write a function that looks at a list of actions and detects which features are used, then determine the protocol versions. It will be much easier if that logic is the responsibility of a single function than spread out.

I think for updates we only need to run that function on new actions and merge the result with the existing protocol versions on the table. There's probably also some concurrency resolution that needs to happen too, but I haven't yet thought about that part much..

wjones127: I might even say we should let clients set the protocol; we should consider that the responsibility of the library. What do you think? Are there use cases where they should set it?

rust/src/operations/create.rs
```diff
@@ -1,251 +1,367 @@
...
+    fn into_future(self) -> Self::IntoFuture {
+        let this = self;
+
+        Box::pin(async move {
+            let mode = this.mode.clone();
+            let (mut table, actions, operation) = this.into_table_and_actions()?;
+            if table.object_store().is_delta_table_location().await? {
+                match mode {
+                    SaveMode::ErrorIfExists => return Err(CreateError::TableAlreadyExists.into()),
+                    SaveMode::Append => return Err(CreateError::AppendNotAllowed.into()),
```
wjones127, line 260: Why is append not allowed?

rust/src/operations/create.rs
```diff
@@ -1,251 +1,367 @@
...
+                    SaveMode::Ignore => {
+                        table.load().await?;
+                        return Ok(table);
+                    }
+                    SaveMode::Overwrite => {
+                        let curr_files =
+                            flatten_list_stream(table.object_store().as_ref(), None).await?;
+                        table.object_store().delete_batch(&curr_files).await?;
```
wjones127, line 268: Why delete the current files?

wjones127 (commented):
I really like these new Builder APIs so far.

I do still like the idea of a having a three-tier API: (action-based, engine-agnostic, DataFusion-based). I think for clarity, it would be best to have those in separate modules of the crate. So, for example, I don't think the Create command and the Load/Write commands belong in the same module, since the first you pass in actions (low-level) while the others are more high-level and deal with data. I'll think about this some more though, since I'm not 100% sure this makes sense.

rust/src/operations/create.rs
roeap: Well .. in this case I was not sure, but conceptually I felt we are not overwriting the data in a table and creating a new version, but creating an entirely new table at version 0. If we were to support also updating table metadata, schema etc via this route, I guess there is more work to be done here validating all changes.

But this one I was definitely unsure about.

roeap: Had a look at the spark implementation - seems they agree with you and updating the metadata and evoling the table is the way to go

https://github.com/delta-io/delta/blob/1f6ab824e14794c17202b5e4e5df6a95357a799c/core/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala#L196-L208

wjones127:
> So, for example, I don't think the Create command and the Load/Write commands belong in the same module, since the first you pass in actions (low-level) while the others are more high-level and deal with data. I'll think about this some more though, since I'm not 100% sure this makes sense.

Okay I think I mis-read the purpose of Create: it's not a low-level API, it's just to create a table, so there's isn't any data writing interaction involved. So we should just think of `operations` module as the DataFusion-based high-level API. 👍

rust/src/operations/load.rs
```diff
@@ -0,0 +1,141 @@
...
+//     Task,
+// }
+//
+// impl From<LoadError> for DeltaTableError {
+//     fn from(err: LoadError) -> Self {
+//         DeltaTableError::GenericError {
+//             source: Box::new(err),
+//         }
+//     }
+// }
```
wjones127, line 24: should we delete this?

```suggestion
```

rust/src/operations/mod.rs
```diff
@@ -1,329 +1,121 @@
...
-        let session_ctx = SessionContext::new();
-        let task_ctx = session_ctx.task_ctx();
-        let _ = collect(transaction, task_ctx).await?;
-        self.table.update().await?;
-
-        Ok(())
+    /// Create a new [`DeltaOps`] instance, backed by an un-initialized in memory table
+    ///
+    /// Using this will not persist any changes beyond the lifetime of the table object.
+    /// THe main purpose of in-memory tables is for use in testing.
```
wjones127, line 57:
```suggestion
    /// The main purpose of in-memory tables is for use in testing.
```

rust/src/operations/mod.rs
```diff
@@ -1,329 +1,121 @@
...
+    ///
+    /// ```
+    /// use deltalake::DeltaOps;
+    ///
+    /// let ops = DeltaOps::new_in_memory();
+    /// ```
+    #[must_use]
+    pub fn new_in_memory() -> Self {
```
wjones127, line 65: This is very cool!

rust/src/operations/mod.rs
```diff
@@ -1,329 +1,121 @@
...
+    ///
+    /// Using this will not persist any changes beyond the lifetime of the table object.
+    /// THe main purpose of in-memory tables is for use in testing.
+    ///
+    /// ```
+    /// use deltalake::DeltaOps;
+    ///
+    /// let ops = DeltaOps::new_in_memory();
+    /// ```
+    #[must_use]
```
wjones127, line 64: This is new to me. Why `#[must_use]`?

rust/src/operations/transaction.rs
```diff
@@ -1,194 +1,203 @@
...
+use serde_json::{Map, Value};
+
+const DELTA_LOG_FOLDER: &str = "_delta_log";
+
+#[derive(thiserror::Error, Debug)]
+enum TransactionError {
+    #[error("Tried committing existing table version: {0}")]
+    VersionAlreadyExists(DeltaDataTypeVersion),
+
+    /// Error returned when reading the delta log object failed.
```
wjones127, line 18: Is this description accurate? It's duplicated below, and `reading` and serializing seem like opposites.

wjones127 (changes requested):
I'm only halfway through and will try to wrap up reviewing tomorrow. Have some initial comments / suggestions.

rust/src/operations/writer.rs
```diff
@@ -0,0 +1,512 @@
...
+            self.config.file_schema(),
+            self.config.partition_columns.clone(),
+            values,
+        )
+        .map_err(|err| WriteError::Partitioning(err.to_string()))?)
+    }
+
+    /// Write a batch to the partition induced by the partition_values. The record batch is expected
+    /// to be pre-partitioned and only contain rows that belong into the same partition.
+    pub async fn write_partition(
```
wjones127, line 182:
```suggestion
    /// to be pre-partitioned and only contain rows that belong into the same partition.
    /// However, it should still contain the partition columns.
    pub async fn write_partition(
```

rust/src/operations/writer.rs
```diff
@@ -0,0 +1,512 @@
...
+                expected_schema: self.config.file_schema.clone(),
+            }
+            .into());
+        }
+
+        let max_offset = batch.num_rows();
+        for offset in (0..max_offset).step_by(self.config.write_batch_size) {
+            let length = usize::min(self.config.write_batch_size, max_offset - offset);
+            self.write_batch(&batch.slice(offset, length))?;
+            // flush currently buffered data to disc once we meet ot exceed the tart file size.
```
wjones127, line 417:
```suggestion
            // flush currently buffered data to disk once we meet or exceed the target file size.
```

rust/src/storage/mod.rs
```diff
@@ -157,10 +157,16 @@ impl DeltaObjectStore {
 
     /// Check if the location is a delta table location
     pub async fn is_delta_table_location(&self) -> ObjectStoreResult<bool> {
-        match self.head(self.log_path()).await {
-            Ok(_) => Ok(true),
-            Err(ObjectStoreError::NotFound { .. }) => Ok(false),
-            Err(err) => Err(err),
+        // TODO We should really be using HEAD here, but this fails in windows tests
```
wjones127, line 160: Have you created a ticket for this in object-store? I can look into this.

wjones127 (changes requested):
Finished looking through, and few more comments mostly around clean up and follow up issues.

rust/src/operations/create.rs
```diff
@@ -1,251 +1,369 @@
...
-            )
-            .try_flatten(),
-        )))
+    /// Set options used to initialize storage backend
+    ///
+    /// Options may be passed in the HashMap or set as environment variables.
+    ///
+    /// [crate::builder::s3_storage_options] describes the available options for the AWS or S3-compliant backend.
+    /// [dynamodb_lock::DynamoDbLockClient] describes additional options for the AWS atomic rename client.
+    pub fn with_storage_options(mut self, storage_options: HashMap<String, String>) -> Self {
```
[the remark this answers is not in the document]
roeap: absolutely - there is even a larger topic I realize now I "forgot". By now we are consuming the table when creating operations, and the cases where we take the entire table is not covered yet. especially we should leverage already loaded state. but that likely better in a follow up? WIll add some docs around prio for now.

rust/src/operations/mod.rs
roeap: I also only fairly recently learned about must use. The idea is, if you do not consume the result of this call, clippy (or even the compiler) will complain. which to me makes sense, since if not using the return it literally does nothing and should be removed.

I think futures for instance may also be must_use since if not consumed they also so nothing ...

rust/src/operations/transaction.rs
```diff
@@ -1,194 +1,203 @@
...
     }
+    Ok(jsons.join("\n"))
+}
 
-    fn output_ordering(&self) -> Option<&[PhysicalSortExpr]> {
-        None
+/// Low-level transaction API. Creates a temporary commit file. Once created,
+/// the transaction object could be dropped and the actual commit could be executed
+/// with `DeltaTable.try_commit_transaction`.
+async fn prepare_commit(
```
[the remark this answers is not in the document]
roeap: this is correct. I wanted to iterate a bit more on commits (as well as the writer) before adopting it in the main code paths. Next I wanted to finally look into the conflict resolution again, where I expect more changes to the commit behavior.

rust/src/operations/write.rs
```diff
@@ -1,517 +1,508 @@
...
+                let config = WriterConfig::new(
+                    inner_plan.schema(),
+                    partition_columns.clone(),
+                    None,
+                    this.target_file_size,
+                    this.write_batch_size,
+                );
+                let mut writer = DeltaWriter::new(object_store.clone(), config);
+
+                let mut stream = inner_plan.execute(i, task_ctx)?;
```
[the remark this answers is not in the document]
roeap: good point, I think its completely pull based, so we should be fine. But let's make sure.

rust/src/operations/write.rs
```diff
@@ -1,517 +1,508 @@
...
-        }
+impl Default for WriteBuilder {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
-impl ExecutionPlan for WriteCommand {
-    fn as_any(&self) -> &dyn Any {
+impl WriteBuilder {
```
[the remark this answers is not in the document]
roeap: I went back and forth a bit on this. If we eventually want to have support for both parquet implementations, maybe we should not expose the specific option structs directly, on the other hand there is more options to consider...

rust/src/operations/write.rs
```diff
@@ -1,517 +1,508 @@
...
+                }
+            } else {
+                Ok(this.partition_columns.unwrap_or_default())
+            }?;
+
+            let plan = if let Some(plan) = this.input {
+                Ok(plan)
+            } else if let Some(batches) = this.batches {
+                if batches.is_empty() {
+                    Err(WriteError::MissingData)
```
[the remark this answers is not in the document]
roeap: I guess ... :). if we do we need to either make it a no-op or add some logic to handle the case when we do not have an explicitly defined table schema, in case we need to create the table.

rust/src/operations/write.rs
```diff
@@ -1,517 +1,508 @@
...
-    let schema = input.schema().clone();
-    let data = collect_batch(input).await?;
-    if data.is_empty() {
-        let stream = EmptyRecordBatchStream::new(Arc::new(OPERATION_SCHEMA.clone()));
-        return Ok(Box::pin(stream));
-    }
-    let mut writer =
-        RecordBatchWriter::try_new(table_uri.clone(), schema, partition_columns.clone(), None)
-            .map_err(to_datafusion_err)?;
+            // TODO should we build checkpoints based on config?
```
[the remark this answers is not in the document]
roeap: #913

rust/src/operations/writer.rs
```diff
@@ -0,0 +1,512 @@
...
+    pub fn new(object_store: ObjectStoreRef, config: WriterConfig) -> Self {
+        Self {
+            object_store,
+            config,
+            partition_writers: HashMap::new(),
+        }
+    }
+
+    /// Creates a [`DeltaWriter`] to write data to provided Delta Table
+    // pub fn for_table(
```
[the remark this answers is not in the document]
roeap: gone :)

Cargo.toml
```diff
@@ -8,6 +8,3 @@ members = [
     "dynamodb_lock",
 ]
 exclude = ["proofs", "delta-inspect"]
-
-[profile.dev]
-split-debuginfo = "unpacked"
```
roeap, line 13: Not entirely sure why why had this option. however it seems it is only stable on macOS, and was causing build issues for Windows after 1,65 was released, so I removed it.

rust/src/operations/write.rs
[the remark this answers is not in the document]
roeap: fairly certain its pull based by now :)

rust/src/operations/create.rs
```diff
@@ -1,251 +1,369 @@
...
-        let table_dir = tempfile::tempdir().unwrap();
-        let table_path = table_dir.path();
-        let table_uri = table_path.to_str().unwrap().to_string();
+        let table = DeltaOps::new_in_memory()
+            .create()
+            .with_columns(table_schema.get_fields().clone())
+            .with_save_mode(SaveMode::Ignore)
+            .await
+            .unwrap();
+        assert_eq!(table.version(), 0)
```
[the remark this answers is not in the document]
roeap: done.

rust/src/operations/create.rs
roeap: I think you are right - having the user choose the protocol and defaulting to "what spark uses" seems like the way to go. WE may in that case have to check option compatibility though?

rust/src/operations/create.rs
roeap: it's what spark does :)

rust/src/operations/create.rs
roeap: Since the metadata update is a larger operation, I opted for raising not implemented for now in this PR. I opened #917 to track this.

wjones127 (approved):

roeap:
@wjones127 - sorry for letting this sit for so long. Did some minor tweaks, mainly related to so deprecations in latest chrono and resolved some conflicts with main.

could you re-approve? :)

wjones127 (approved):
No worries. Thanks for coming back to it. I'm excited to get this merged :)

PR #866, opened 2022-10-07
Added without_files flag to DeltaTable constructor

MykhailoHevak:
# Description
Added _without_files_ flag to DeltaTable constructor. If the flag is set it will be called method _without_files_ in the **DeltaTableBuilder**. And this allows users to control memory usage. Some append-only applications might have no need of tracking any files.

# Documentation
Indicates whether DeltaTable should not track files. This defaults to "false" (files are tracked). Some append-only applications might have no need of tracking any files. So, DeltaTable will be loaded with significant memory reduction.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -77,7 +77,7 @@ class DeltaTable:
         table_uri: str,
         version: Optional[int] = None,
         storage_options: Optional[Dict[str, str]] = None,
-        without_files: Optional[bool] = None
+        without_files: bool = False
     ):
         """
         Create the Delta Table from a path with an optional version.
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -77,6 +77,7 @@ class DeltaTable:
         table_uri: str,
         version: Optional[int] = None,
         storage_options: Optional[Dict[str, str]] = None,
+        without_files: Optional[bool] = None
     ):
         """
         Create the Delta Table from a path with an optional version.
@@ -89,7 +90,7 @@ class DeltaTable:
         """
         self._storage_options = storage_options
         self._table = RawDeltaTable(
-            table_uri, version=version, storage_options=storage_options
+            table_uri, version=version, storage_options=storage_options, without_files=without_files
         )
         self._metadata = Metadata(self._table)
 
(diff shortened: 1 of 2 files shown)


python/deltalake/table.py
```diff
@@ -89,7 +90,7 @@ def __init__(
         """
```
wjones127, line 90: Will want to document this parameter:

```suggestion
        :param without_files: If True, will load table without tracking files. Some 
          append-only applications might have no need of tracking any files. So, the
          DeltaTable will be loaded with a significant memory reduction.
        """
```

python/deltalake/table.py
```diff
@@ -77,6 +77,7 @@ def __init__(
         table_uri: str,
         version: Optional[int] = None,
         storage_options: Optional[Dict[str, str]] = None,
+        without_files: Optional[bool] = None
```
wjones127, line 80: Why not make it clear what the default is in the signature?

```suggestion
        without_files: bool = False
```

wjones127 (changes requested):
What's the behavior of `DeltaTable.file_uris()` when `without_files=True`? Does it return empty? Or raise an error? Or load the files? We might want to add a unit test for that in Python.

wjones127:
> What's the behavior of DeltaTable.file_uris() when without_files=True? Does it return empty? Or raise an error? Or load the files? We might want to add a unit test for that in Python.

Just checked and it returns an empty list. We should document that for now, and consider changing that behavior in the future (#867).

wjones127:
@MykhailoHevak looks like the merge you pushed overwrote the changes you accepted in the GitHub UI. Could you add those back?

MykhailoHevak:
> @MykhailoHevak looks like the merge you pushed overwrote the changes you accepted in the GitHub UI. Could you add those back?

Oooops... I did it accidentally. Is it OK now?

MykhailoHevak:
Do I have to do something else?

python/src/lib.rs
```diff
@@ -106,6 +106,7 @@ impl RawDeltaTable {
         table_uri: &str,
         version: Option<deltalake::DeltaDataTypeLong>,
         storage_options: Option<HashMap<String, String>>,
+        without_files: bool
```
wjones127, line 109:
```suggestion
        without_files: bool,
```

wjones127 (approved):

wjones127:
Thanks @MykhailoHevak!

dennyglee:
Hey @MykhailoHevak I believe this is your first PR to the delta repo.  Can you ping me (dennyglee) via the [Delta Users slack](https://go.delta.io/slack) so we can send you some swag?  Thanks!

PR #892, opened 2022-10-17
Add TableProviderFactory and test for SQL to register tables dynamically at runtime

### LINKED ISSUE #891: Support `CREATE EXTERNAL TABLE dt STORED AS DELTATABLE` syntax with datafusion
# Description

In datafusion, SQL users can use the `create external table xyz stored as csv` option to register CSV, JSON, and parquet data. It would be nice (essential in my company's case) to allow users to register deltalake tables in the same manner.

avantgardnerio:
# Description

Use (hopefully soon-to-be-merged) Datafusion functionality to allow dynamic (i.e. SQL-based) registration of deltalake tables at runtime.

# Related Issue(s)
- closes #891

# Documentation

- TBD

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -46,9 +46,10 @@ rusoto_dynamodb = { version = "0.48", default-features = false, optional = true 
 rusoto_glue = { version = "0.48", default-features = false, optional = true }
 
 # Datafusion
-datafusion = { version = "14", optional = true }
-datafusion-expr = { version = "14", optional = true }
-datafusion-common = { version = "14", optional = true }
+datafusion = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
+datafusion-expr = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
+datafusion-common = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
+datafusion-proto = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
 
 # NOTE dependencies only for integration tests
 fs_extra = { version = "1.2.0", optional = true }
@@ -72,7 +73,7 @@ tempfile = "3"
 utime = "0.3"
 
 [build-dependencies]
-glibc_version = "0"
+glibc_version = { path = "../glibc_version" }
 
 [features]
 default = ["arrow", "parquet"]
@@ -80,6 +81,7 @@ datafusion-ext = [
     "datafusion",
     "datafusion-expr",
     "datafusion-common",
+    "datafusion-proto",
     "arrow",
     "parquet",
 ]
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -46,10 +46,10 @@ rusoto_dynamodb = { version = "0.48", default-features = false, optional = true 
 rusoto_glue = { version = "0.48", default-features = false, optional = true }
 
 # Datafusion
-datafusion = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
-datafusion-expr = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
-datafusion-common = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
-datafusion-proto = { git = "https://github.com/apache/arrow-datafusion.git", rev = "4a67d0d2613d8c4b323e0dd5a6afe1eb1115c7ba", optional = true }
+datafusion = { version = "14", optional = true }
+datafusion-expr = { version = "14", optional = true }
+datafusion-common = { version = "14", optional = true }
+datafusion-proto = { version = "14", optional = true }
 
 # NOTE dependencies only for integration tests
 fs_extra = { version = "1.2.0", optional = true }
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -23,28 +23,37 @@
 use std::any::Any;
 use std::collections::HashMap;
 use std::convert::TryFrom;
+use std::fmt::Debug;
 use std::sync::Arc;
 
-use crate::action;
-use crate::schema;
+use crate::{action, open_table};
+use crate::{schema, DeltaTableBuilder};
 use crate::{DeltaTable, DeltaTableError};
 
 use arrow::array::ArrayRef;
 use arrow::compute::{cast_with_options, CastOptions};
-use arrow::datatypes::{DataType as ArrowDataType, Schema as ArrowSchema, TimeUnit};
+use arrow::datatypes::{DataType as ArrowDataType, Schema as ArrowSchema, SchemaRef, TimeUnit};
 use arrow::record_batch::RecordBatch;
 use async_trait::async_trait;
 use chrono::{DateTime, NaiveDateTime, Utc};
+use datafusion::datasource::datasource::TableProviderFactory;
 use datafusion::datasource::file_format::{parquet::ParquetFormat, FileFormat};
 use datafusion::datasource::{listing::PartitionedFile, MemTable, TableProvider, TableType};
-use datafusion::execution::context::{SessionContext, SessionState};
+use datafusion::execution::context::{SessionContext, SessionState, TaskContext};
+use datafusion::execution::runtime_env::RuntimeEnv;
+use datafusion::execution::FunctionRegistry;
+use datafusion::optimizer::utils::conjunction;
+use datafusion::physical_expr::PhysicalSortExpr;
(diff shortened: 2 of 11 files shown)


rust/tests/datafusion_test.rs
```diff
@@ -79,6 +84,37 @@ async fn prepare_table(
...
+    table_factories.insert("DELTATABLE".to_string(), Arc::new(DeltaTableFactory {}));
+    let cfg = RuntimeConfig::new()
+        .with_table_factories(table_factories);
+    let env = RuntimeEnv::new(cfg).unwrap();
+    let ses = SessionConfig::new();
+    let ctx = SessionContext::with_config_rt(ses, Arc::new(env));
+
+    let mut d = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    d.push("tests/data/delta-0.8.0-partitioned");
+    let sql = format!("CREATE EXTERNAL TABLE demo STORED AS DELTATABLE LOCATION '{}'", d.to_str().unwrap());
```
avantgardnerio, line 99: This is the goal of the PR

rust/Cargo.toml
```diff
@@ -9,7 +9,7 @@ description = "Native Delta Lake implementation in Rust"
 edition = "2021"
 
 [dependencies]
-arrow = { version = "22", optional = true }
+arrow = { version = "24", optional = true }
```
avantgardnerio, line 12: I had to upgrade arrow to match Datafusion master

rust/Cargo.toml
```diff
@@ -45,9 +44,9 @@ rusoto_dynamodb = { version = "0.48", default-features = false, optional = true
 rusoto_glue = { version = "0.48", default-features = false, optional = true }
 
 # Datafusion
-datafusion = { version = "12", optional = true }
-datafusion-expr = { version = "12", optional = true }
-datafusion-common = { version = "12", optional = true }
+datafusion = { git = "https://github.com/spaceandtimelabs/arrow-datafusion.git", rev = "cb9241982556745939175b0597194ad819bab613", optional = true }
```
avantgardnerio, line 47: This will point back at `apache/arrow-datafusion` once that PR (#3867) is merged. Then probably `v14.0.0` once it is released.

avantgardnerio:
This is working in Ballista from DataGrip through FlightSQL! :partying_face: 

Screenshot from 2022-10-23 12-44-02

@alamb that's the end goal for the related datafusion PR

avantgardnerio:
@houqp I was finally able to get Ballista working running deltalake queries. I'd appreciate a review when you get the chance (and a CI kick would be appreciated as well).

avantgardnerio:
I fixed the CI checks I could, and fixed datafusion to not require `delta-rs` to mess with the build. If someone could fire off CI, I'd appreciate it :raised_hands:

rust/src/delta_datafusion.rs
```diff
@@ -318,21 +340,15 @@ impl TableProvider for DeltaTable {
...
-        );
+        register_store(&self, session.runtime_env.clone());
 
         // TODO we group files together by their partition values. If the table is partitioned
         // and partitions are somewhat evenly distributed, probably not the worst choice ...
         // However we may want to do some additional balancing in case we are far off from the above.
         let mut file_groups: HashMap<Vec<ScalarValue>, Vec<PartitionedFile>> = HashMap::new();
-        if let Some(Some(predicate)) = (!filters.is_empty()).then_some(combine_filters(filters)) {
+        if let Some(Some(predicate)) =
+            (!filters.is_empty()).then_some(conjunction(filters.to_vec()))
```
wjones127, line 350: Perhaps:

```suggestion
        if let Some(predicate) = conjunction(filters)
```

([conjunction() returns None if input is empty](https://github.com/apache/arrow-datafusion/blob/7cba7582a67c9353dfd851243be048149c849c2c/datafusion/optimizer/src/utils.rs#L127))

rust/src/delta_datafusion.rs
```diff
@@ -367,27 +383,86 @@ impl TableProvider for DeltaTable {
...
+
+        Ok(Arc::new(delta_scan))
     }
 
     fn as_any(&self) -> &dyn Any {
         self
     }
 }
 
+/// A wrapper for parquet scans that installs the required ObjectStore
```
wjones127, line 414: Note: this will likely also need to perform [column mapping](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-mapping) later when we support reader protocol v2.

rust/src/delta_datafusion.rs
```diff
@@ -294,6 +304,18 @@ impl PruningStatistics for DeltaTable {
...
+// each delta table must register a specific object store, since paths are internally
+// handled relative to the table root.
+fn register_store(table: &DeltaTable, env: Arc<RuntimeEnv>) {
+    let object_store_url = table.storage.object_store_url();
+    let url: &Url = object_store_url.as_ref();
+    env.register_object_store(
+        url.scheme(),
+        url.host_str().unwrap_or_default(),
+        table.object_store(),
+    );
```
wjones127, line 316: IIUC, any URI passed with the same scheme and bucket name will match to this object store? And the object stores options will be preserved? I'm worried about the cases where [storage options](https://github.com/delta-io/delta-rs/blob/a2af20bd963ae7c7923d3e4030e946a6a8199779/rust/src/builder/mod.rs#L199) were directly passed into the `DeltaTable` instance.

rust/src/delta_datafusion.rs
```diff
@@ -673,6 +748,118 @@ impl DeltaDataChecker {
...
+}
+
+/// Responsible for creating deltatables
+pub struct DeltaTableFactory {}
+
+#[async_trait]
+impl TableProviderFactory for DeltaTableFactory {
+    async fn create(&self, url: &str) -> datafusion::error::Result<Arc<dyn TableProvider>> {
+        let provider = open_table(url).await.unwrap();
+        Ok(Arc::new(provider))
```
wjones127, line 859: This is perhaps a place where `storage_options` will get lost, if we care about that. If we only want to support passing credentials via env variables for systems that use this (just distributed, right?), that might be sensible.

wjones127 (commented):
This is very cool!

I have an option suggestion and a question on how we are handling storage options.

roeap:
I just marked #852 as ready for review. That should fix the python builds which are breaking due to the arrow version bump as well as handle the parquet2 build that needed some changes as parquet-format moved into the parquet crate.

rust/src/delta_datafusion.rs
roeap: the object store url we get from `table.storage.object_store_url` is just for internal purposes and should not collide with common urls. It uses `delta-rs://` as a scheme and does a little bit of pruning required for widows paths.

rust/src/delta_datafusion.rs
roeap: should we add a TODO comment for this?

rust/src/delta_datafusion.rs
```diff
@@ -367,27 +383,86 @@ impl TableProvider for DeltaTable {
...
+    fn children(&self) -> Vec<Arc<dyn ExecutionPlan>> {
+        vec![self.parquet_scan.clone()]
+    }
+
+    fn with_new_children(
+        self: Arc<Self>,
+        children: Vec<Arc<dyn ExecutionPlan>>,
+    ) -> DataFusionResult<Arc<dyn ExecutionPlan>> {
+        ExecutionPlan::with_new_children(self.parquet_scan.clone(), children)
+    }
```
roeap, line 449: I may be completely off, but can a scan operation even have children? `ParquetScan` itself just returns `Ok(self)` .

roeap (commented):
great work! really excited to see this land.

rust/src/delta_datafusion.rs
avantgardnerio: > should we add a TODO comment for this?

Done!

rust/src/delta_datafusion.rs
avantgardnerio: > we only want to support passing credentials via env variables for systems that use this (just distributed, right?), that might be sensible.

For now, that is all I am attempting to do with this PR. I can possibly see someone wanting to file another PR to extend it in the future, but if they are passing `storage_options`, I think that means they are registering tables in Rust (not SQL), and therefor already are writing a fork of Ballista, in which case they can write their own serde?

It's a worthwhile thought experiment, but I wouldn't recommend coding for it quite yet...

rust/src/delta_datafusion.rs
avantgardnerio: > can a scan operation even have children?

I would think not, but the goal of this implementation was merely to proxy every call to the wrapped `ParquetScan` with the one additional behavior of registering the `ObjectStore` prior to proxying the `execute()` request... so I just followed that pattern here. I think it's probably best to leave it this way just in case there is some future change where a `ParquetScan` might have children for some reason - possibly because someone wants to register their own proxy in the same way I just did :)

avantgardnerio:
I did some fixes, so there's a reasonable chance CI should pass now.

avantgardnerio:
It looks like the remaining errors are in src/python, so I assume they will be resolved by https://github.com/delta-io/delta-rs/pull/852 .

rust/src/delta_datafusion.rs
wjones127: That's fair. I just wanted to make sure you were aware of that limitation.

wjones127 (dismissed):

avantgardnerio:
I see #852 was merged. I think we can merge this now. If someone wouldn't mind kicking off CI, I'd appreciate it!

avantgardnerio:
I don't understand the python test failure... it looks like a timeout, is it an intermittent test? Or should I try to dig into it more?

wjones127 (approved):
Looks good! Thanks!

avantgardnerio:
Thanks everyone!

roeap:
Thank you!

Issue #883, opened 2022-10-13
Python write_deltalake to S3 fails to write due to "invalid json"


joshuarobinson:
# Environment

**Delta-rs version**: 0.6.2

**Binding**: Python

**Environment**: Ubuntu 22.04, Python 3.10, deltalake==0.6.2, Running against non-AWS S3 (Swift)

***
# Bug

**What happened**:
DeltaLake write fails.

My test code to write 'df' (a pandas dataframe) to an S3 location:
```
storage_options = {"AWS_ACCESS_KEY_ID": ACCESS_KEY, "AWS_SECRET_ACCESS_KEY":SECRET_KEY, "AWS_ENDPOINT_URL": ENDPOINT_URL, "AWS_REGION": 'us-east-1'}
write_deltalake('s3://joshuarobinson/test_deltalake/', df, storage_options=storage_options)
```
fails with the following error:
```
Traceback (most recent call last):
  File "/delta_write.py", line 19, in <module>
    write_deltalake('s3://joshuarobinson/test_deltalake/', df, storage_options=storage_options)
  File "/usr/local/lib/python3.10/site-packages/deltalake/writer.py", line 156, in write_deltalake
    table = try_get_deltatable(table_or_uri)
  File "/usr/local/lib/python3.10/site-packages/deltalake/writer.py", line 332, in try_get_deltatable
    return DeltaTable(table_uri)
  File "/usr/local/lib/python3.10/site-packages/deltalake/table.py", line 91, in __init__
    self._table = RawDeltaTable(
deltalake.PyDeltaTableError: Failed to load checkpoint: Invalid JSON in checkpoint: expected value at line 1 column 1
```

Note that the destination path is empty, i.e., I'm writing a brand new table
```
$ s5cmd ls s3://joshuarobinson/test_deltalake/
ERROR "ls s3://joshuarobinson/test_deltalake/": no object found
```

Also tried: 
 1) I have tested with all four values of "mode" and had the same result.
 2) also tried to manually build a pyarrow filesystem and pass that but did not work.

Thelin90:
@joshuarobinson I have the same issue.

Looking at the code, it currently expects the table to already exist:

`write_deltalake` performs:

```python
    if isinstance(table_or_uri, str):
        if "://" in table_or_uri:
            table_uri = table_or_uri
        else:
            # Non-existant local paths are only accepted as fully-qualified URIs
            table_uri = "file://" + str(Path(table_or_uri).absolute())
[...]

PR #893, opened 2022-10-18
fix: pass storage options down when getting delta table

wjones127:
# Description

We forgot to pass down storage options, so storage options were being ignored if you were passing a URI and not a DeltaTable.

This also adds a new S3 option: `AWS_S3_ALLOW_UNSAFE_RENAME` which allows using the default S3 backend without support for concurrent writers. This is an opt-in option and there is a helpful error message to provide guidance to users.

To test, I implemented integration tests that run against localstack.

# Related Issue(s)

- might be a fix for #883

# Documentation

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -196,17 +196,6 @@ def write_deltalake(
     else:  # creating a new table
         current_version = -1
 
-        # TODO: Don't allow writing to non-empty directory
-        # Blocked on: Finish filesystem implementation in fs.py
-    assert (
-        len(
-            filesystem.get_file_info(
-                pa_fs.FileSelector(table_uri, allow_not_found=True)
-            )
-        )
-        == 0
-    )
-
     if partition_by:
         partition_schema = pa.schema([schema.field(name) for name in partition_by])
         partitioning = ds.partitioning(partition_schema, flavor="hive")
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -147,6 +147,9 @@ def write_deltalake(
         else:
             schema = data.schema
 
+    if filesystem is not None:
+        raise NotImplementedError("Filesystem support is not yet implemented.  #570")
+
     if isinstance(table_or_uri, str):
         if "://" in table_or_uri:
             table_uri = table_or_uri
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -162,8 +162,8 @@ def write_deltalake(
 
     if filesystem is None:
         if table is not None:
-            storage_options = (table._storage_options or {}).update(
-                storage_options or {}
+            storage_options = dict(
+                **(table._storage_options or {}), **(storage_options or {})
             )
 
         filesystem = pa_fs.PyFileSystem(DeltaStorageHandler(table_uri, storage_options))
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -153,7 +153,7 @@ def write_deltalake(
         else:
             # Non-existant local paths are only accepted as fully-qualified URIs
             table_uri = "file://" + str(Path(table_or_uri).absolute())
-        table = try_get_deltatable(table_or_uri)
+        table = try_get_deltatable(table_or_uri, storage_options)
     else:
         table = table_or_uri
         table_uri = table._table.table_uri()
@@ -162,17 +162,11 @@ def write_deltalake(
 
     if filesystem is None:
         if table is not None:
-            filesystem = pa_fs.PyFileSystem(
-                DeltaStorageHandler(
-                    table._table.table_uri(),
-                    table._storage_options,
-                    table._table.get_py_storage_backend(),
-                )
+            storage_options = (table._storage_options or {}).update(
+                storage_options or {}
             )
-        else:
-            filesystem = pa_fs.PyFileSystem(
-                DeltaStorageHandler(table_uri, storage_options)
-            )
+
+        filesystem = pa_fs.PyFileSystem(DeltaStorageHandler(table_uri, storage_options))
 
     if table:  # already exists
         if schema != table.schema().to_pyarrow() and not (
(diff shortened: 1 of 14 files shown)


python/deltalake/writer.py
```diff
@@ -162,18 +162,12 @@ def write_deltalake(
...
-                    table._storage_options,
-                    table._table.get_py_storage_backend(),
-                )
-            )
-        else:
-            filesystem = pa_fs.PyFileSystem(
-                DeltaStorageHandler(table_uri, storage_options)
+            storage_options = (table._storage_options or {}).update(
+                storage_options or {}
             )
```
wjones127, line 168: @roeap Does this seems like an acceptable simplification? This would mean the storage options provided will override any ones that were originally specified in the table.

roeap: I think this should work fine. Right now we have an additional issue though, when is comes to storage options. Not entirely sure, but in some cases environment variables set may take precedence over configuration ... I remember fixing this in some places, but fore S§ I am not 100% sure..

roeap (dismissed):
LGTM!

python/deltalake/writer.py
houqp: isn't environment variable overriding code configuration expected behavior?

roeap: Hmm, not sure. My personal expectation would be that my explicit config takes precedence over the environment. Looking at e.g. multiple targets it seems to be easier to manage. IIRC the python file systems behave like that.

python/deltalake/writer.py
```diff
@@ -153,7 +153,7 @@ def write_deltalake(
         else:
             # Non-existant local paths are only accepted as fully-qualified URIs
             table_uri = "file://" + str(Path(table_or_uri).absolute())
-        table = try_get_deltatable(table_or_uri)
+        table = try_get_deltatable(table_or_uri, storage_options)
```
Thelin90, line 156: I tried this myself by also adding:

```python
def try_get_deltatable(table_uri: str, storage_options: Dict[str, str], version: Optional[int] = None) -> Optional[DeltaTable]:
    try:
        return DeltaTable(table_uri=table_uri, storage_options=storage_options, version=version)
    except PyDeltaTableError as err:
        # TODO: There has got to be a better way...
        if "Not a Delta table" in str(err):
            return None
        elif "cannot find" in str(err):
            return None
        elif "No such file or directory" in str(err):
            return None
        else:
            raise
```

This will return `None` when the table does not exist, instead of failing for creds.

python/deltalake/writer.py
```diff
@@ -162,18 +162,12 @@ def write_deltalake(
...
-                DeltaStorageHandler(
-                    table._table.table_uri(),
-                    table._storage_options,
-                    table._table.get_py_storage_backend(),
-                )
-            )
-        else:
-            filesystem = pa_fs.PyFileSystem(
-                DeltaStorageHandler(table_uri, storage_options)
+            storage_options = (table._storage_options or {}).update(
```
Thelin90, line 165: This is the same as I tried:

```python
    if filesystem is None:
        if table is not None:
            storage_options = (table._storage_options or {}).update(
                storage_options or {}
            )

        filesystem = pa_fs.PyFileSystem(DeltaStorageHandler(table_uri, storage_options))
```

python/deltalake/writer.py
Thelin90: I currently get:

```python
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Generic { store: "S3", source: CreateMultipartRequest { source: Error { retries: 0, message: "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided
```

With these changes.

Thelin90 (commented):
Some insights from me, if valuable at all.

python/deltalake/writer.py
wjones127: > This will return None when the table does not exist, instead of failing for creds, but potentially also taking the version into account later...

I'm not sure what you are saying here. It's supposed to return None when the table does not exist. Are you saying that if it just fails to authenticate to the Object store it might also return None?

Thelin90: Sorry, to be clear.

I'm saying the logic seem correct and yes it should return None in this case.

python/tests/test_fs.py
```diff
@@ -29,3 +32,75 @@ def test_read_simple_table_from_remote(s3_localstack):
     table_path = "s3://deltars/simple"
     dt = DeltaTable(table_path)
     assert dt.to_pyarrow_table().equals(pa.table({"id": [5, 7, 9]}))
+
+
+@pytest.mark.s3
```
Thelin90, line 37: @wjones127 is this you adding an integration test for S3? Based on our other conversations, since it seems like the unit-test were just testing on your local filesystem writing new delta tables.

wjones127: Yes

python/deltalake/writer.py
```diff
@@ -147,13 +147,16 @@ def write_deltalake(
         else:
             schema = data.schema
 
+    if filesystem is not None:
+        raise NotImplementedError("Filesystem support is not yet implemented.  #570")
```
wjones127, line 151: I made a note of what we have to do here: https://github.com/delta-io/delta-rs/issues/570#issuecomment-1288249167

wjones127:
cc @houqp @roeap either you have a moment to re-review? I added integration tests and made a few more meaningful changes.

fvaleye (approved):
Thanks @wjones127 👍 

It's great to include integration tests; we will avoid regressions in our future changes.

LGTM, it's OK to include this [PR](https://github.com/delta-io/delta-rs/pull/924) in the `v0.6.3` version of the Python binding!

Thelin90:
Does this now work based on previous issues? Is it released in the latest version?

Issue #653, opened 2022-06-20
Can't utilise struct delta stats with delta_table.to_pyarrow_dataset()


Tom-Newton:
# Environment

**Delta-rs version**: 0.5.7

**Binding**: Python 

**Environment**:
- **Cloud provider**: Azure (bug is reproduced locally)
- **OS**: Ubuntu 18.04
- **Other**: Python 3.8

***
# Bug

**What happened**:
I want to utilise the cool new data skipping feature implemented by @wjones127 in https://github.com/delta-io/delta-rs/pull/525 and https://github.com/delta-io/delta-rs/pull/565. 
However when using `delta_table.to_pyarrow_dataset()`, for a table which only has struct stats, it opens every parquet file unnecessarily.

**What you expected to happen**:
The delta stats provide sufficient information to narrow it down to just one parquet file so it should only need to open the one file.

**How to reproduce it**:
Script to reproduce: [reproduce_struct_stats_issue.zip](https://github.com/delta-io/delta-rs/files/8940152/reproduce_struct_stats_issue.zip)

Steps this script does
1. Create a delta table with properties: `"delta.checkpoint.writeStatsAsJson": "false"` and `"delta.checkpoint.writeStatsAsStruct": "true"`.
2. Append rows to the table until it creates a checkpoint. 
3. Open the table using `DeltaTable.to_pyarrow_dataset()`
4. Get the parquet fragments when applying a filter. Check the number of fragments is equal to the number of files that could match the filter expression when you have knowledge of the delta stats.

I have also created a branch with a unittest to catch this https://github.com/Tom-Newton/delta-rs/pull/1.

**More details**:
Logging the file and part_expressions returned:
```
/tmp/reproduce_stats_issue/part-00000-431a48bb-dc24-438f-8329-60e9eae4ca22-c000.snappy.parquet None
/tmp/reproduce_stats_issue/part-00000-1cbe22db-1188-46b9-b52d-e1c4a3ab6b61-c000.snappy.parquet None
/tmp/reproduce_stats_issue/part-00000-c84ba96d-a42a-441f-b440-a22c1a48d344-c000.snappy.parquet None
/tmp/reproduce_stats_issue/part-00000-beb74db9-8db2-431d-9be0-a437a01abe47-c000.snappy.parquet None
/tmp/reproduce_stats_issue/part-00000-3c21a7a2-0405-4d33-8be5-b77286af386a-c000.snappy.parquet None
[...]

PR #901, opened 2022-10-24
Fix parsing struct stats after schema evolution

Tom-Newton:
# Description
When a delta table's schema is evolved the struct stat schemas in checkpoints are also evolved. Since the struct stats are stored in a columnar way adding a single file with the new columns will cause nulls to appear in the struct stats for all other files. This is  a significant difference compared to the json stats.

Unfortunately I overlooked this in https://github.com/delta-io/delta-rs/pull/656 for both nullCounts and min/max values. This caused parsed struct stats to have extra columns full of nulls. I don't know if this was actually an issue at all but it should be fixed even if just for the sake of the warnings spam. 
```
[2022-10-24T22:13:22Z WARN  deltalake::action::parquet_read] Expect type of nullCount field to be struct or int64, got: null
[2022-10-24T22:13:22Z WARN  deltalake::action::parquet_read] Expect type of nullCount field to be struct or int64, got: null
[2022-10-24T22:13:22Z WARN  deltalake::action::parquet_read] Expect type of nullCount field to be struct or int64, got: null
[2022-10-24T22:13:22Z WARN  deltalake::action::parquet_read] Expect type of nullCount field to be struct or int64, got: null
[2022-10-24T22:13:22Z WARN  deltalake::action::parquet_read] Expect type of nullCount field to be struct or int64, got: null
``` 

# Related Issue(s)
- Relates to https://github.com/delta-io/delta-rs/issues/653 but for the most part its an already solved issue.

# Changes:
- Replace the test data with similar test data that includes a schema evolution.
- Add error handling for min/max values to ensure warnings will be logged for other unexpected types (there probably shouldn't be any). As a total rust noob I originally filled with nulls but I think that was a mistake. 
- Ignore nulls for min/max stats and null count stats since these are expected after schema evolution and should be ignored without logging a warning. 

Usual disclaimer on a PR from me: I don't know what I'm doing writing rust code. (thanks to wjones for tidying up my dodgy  rust code 🙂)


rust/src/action/parquet_read/mod.rs
```diff
@@ -241,29 +285,27 @@ impl TryFrom<&Field> for ColumnCountStat {
...
+        Field::Float(value) => Ok(json!(value)),
+        Field::Double(value) => Ok(json!(value)),
+        Field::Str(value) => Ok(json!(value)),
         Field::Decimal(decimal) => match BigInt::from_signed_bytes_be(decimal.data()).to_f64() {
-            Some(int) => json!(int / (10_i64.pow((decimal.scale()).try_into().unwrap()) as f64)),
-            _ => serde_json::Value::Null,
+            Some(int) => Ok(json!(
+                int / (10_i64.pow((decimal.scale()).try_into().unwrap()) as f64)
+            )),
+            _ => Err("Invalid type for nullCounts"),
```
wjones127, line 302:
```suggestion
            _ => Err("Invalid type for min/max values."),
```

rust/src/action/parquet_read/mod.rs
```diff
@@ -241,29 +285,27 @@ impl TryFrom<&Field> for ColumnCountStat {
...
-        Field::Date(date) => serde_json::Value::String(convert_date_to_string(*date)),
-        _ => {
-            log::warn!("Unexpected field type {:?}", field,);
-            serde_json::Value::Null
-        }
+        Field::TimestampMillis(timestamp) => Ok(serde_json::Value::String(
+            convert_timestamp_millis_to_string(*timestamp),
+        )),
+        Field::Date(date) => Ok(serde_json::Value::String(convert_date_to_string(*date))),
+        _ => Err("Invalid type for nullCounts"),
```
wjones127, line 308:
```suggestion
        _ => Err("Invalid type for min/max values."),
```

rust/src/action/parquet_read/mod.rs
```diff
@@ -202,15 +231,30 @@ impl Add {
...
+                                field
+                            );
+                            None
+                        }
+                    }),
+            ))),
+            _ => match primitive_parquet_field_to_json_value(field) {
+                Ok(value) => Ok(ColumnValueStat::Value(value)),
+                _ => Err("Invalid type for min/max values."),
+            },
```
wjones127, line 257:
```suggestion
            _ => primitive_parquet_field_to_json_value(field).map(ColumnValueStat::Value),
```

rust/src/action/parquet_read/mod.rs
```diff
@@ -173,14 +197,19 @@ impl Add {
...
-                                        Ok(count) => {
-                                            stats.null_count.insert(name.clone(), count);
-                                        },
+                                    match field {
+                                        Field::Null => {}
                                         _ => {
-                                            log::warn!("Expect type of nullCount field to be struct or int64, got: {}", field);
-                                        },
-                                    };
+                                            match field.try_into() {
```
wjones127, line 203: huh I'm unsure how the compiler knows what the destination types is here, but I guess it's ColumnCountStat?

rust/src/action/parquet_read/mod.rs
```diff
@@ -153,7 +153,19 @@ impl Add {
...
+                                    match field {
+                                        Field::Null => {},
+                                        _ => {
+                                            match field.try_into() {
+                                                Ok(values) => {
+                                                    stats.min_values.insert(name.clone(), values);
+                                                }
+                                                _ => {
+                                                    log::warn!("Unexpected type of minValues field, got: {}", field);
+                                                }
```
wjones127, line 165: It's unfortunate that we are duplicating this warning here; it would be nice if we could somehow consolidate this with the one below.

wjones127 (commented):
This is still a draft, but thought I might leave some early feedback.

rust/src/action/parquet_read/mod.rs
Tom-Newton: I'm afraid I might need a clearer pointer than that 😅. The 2 warnings are coming from different `Err`s so as I understand both need to be handled. I was hoping to find a way to share more of the code between minValues, maxValues and nullCounts but I think that would require understanding Generics in rust.

wjones127: It is quite tricky, having looked at it more. The warning was being generated separately for top-level fields and their children. I created a PR with suggested changes that will simplify some of the logic and avoid the duplication of this warning. https://github.com/Tom-Newton/delta-rs/pull/26

Tom-Newton: Thanks @wjones127. I like to have a go at these things, despite my lacking rust knowledge, but I'm grateful that you are always so helpful.

wjones127 (approved):

wjones127:
Thanks @Tom-Newton

PR #902, opened 2022-10-30
changes for bump to datafusion13/arrow24

timvw:
# Description
Update code such that it uses datafusion 13 (arrow 24)


wjones127:
@roeap is already making these upgrades in #852, and we are halfway done with review. So perhaps we should finish that instead? Or do we want to get this one through first?

timvw:
I was not aware this was already happening.

PR #932, opened 2022-11-13
fix(python): enforce reader protocol version

### LINKED ISSUE #929: Support protocol version checks
# Description
Add protocol version checks and report the supported Delta protocol version.  If 

**Use Case**
User is reading a Delta table with delta-rs but if at the delta-rs does not support deletion vectors, it should report back that this is currently unsupported by delta-rs. 

**Related Issue(s)**
#928, #930

wjones127:
# Description

We weren't enforcing the reader protocol version, causing issues when trying to read tables with column mapping. 

# Related Issue(s)

- closes #929

# Documentation


fvaleye (approved):
Thank you @wjones127 

LGTM 👍

PR #941, opened 2022-11-16
Allow for reading columns as dictionaries using to_pyarrow_dataset

### LINKED ISSUE #938: to_pyarrow_dataset() ignores dictionary_columns in read_options
# Environment

**Delta-rs version**: 0.6.3

**Binding**: python

**Environment**:
- **Cloud provider**: Azure
- **OS**: Windows/linux
- **Other**:

***
# Bug
When passing `parquet_read_options` to `to_pyarrow_dataset`, it is possible to use `dictionary_columns` to control which columns should be dictionary encoded as they are read.

**What happened**:
Such columns are not dictionary encoded when they are read, rather they are of type string, as they would be had `dictionary_columns` been empty.

I expect the particular columns to be dictionary encoded.

**How to reproduce it**:

```python
import pyarrow.dataset as ds

dt = DeltaTable(path)
read_options = ds.ParquetReadOptions(dictionary_columns=["test"])
data = dt.to_pyarrow_dataset(parquet_read_options=read_options)
table = data.to_table()
print(table.schema)
```

I expect the column "test" to be dictionary encoded, rather than just of type string.

**More details**:
I believe that the problem is in [table.py at line 338](https://github.com/delta-io/delta-rs/blob/15c4bd522b4271d7d68578b9dd891d9a6a5f3f42/python/deltalake/table.py#L338), the schema of `self.schema.to_pyarrow()` has the wrong type for the columns which are to be dictionary encoded.

Maybe it is possible to use the `physical_schema` property of the fragments defined just above to get the right schema, or otherwise parse the read options  to modify the schema?

Kuhlwein:
# Description
When passing `parquet_read_options` to `to_pyarrow_dataset` it is now possible to use `dictionary_columns` to control which columns should be dictionary encoded as they are read.

# Related Issue(s)
- closes #938 

# Documentation

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -334,9 +334,12 @@ class DeltaTable:
             )
         ]
 
-        return FileSystemDataset(
-            fragments, self.schema().to_pyarrow(), format, filesystem
-        )
+        if parquet_read_options is not None and parquet_read_options.dictionary_columns:
+            schema = pyarrow.unify_schemas([f.physical_schema for f in fragments])
+        else:
+            schema = self.schema().to_pyarrow()
+
+        return FileSystemDataset(fragments, schema, format, filesystem)
 
     def to_pyarrow_table(
         self,
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -334,10 +334,16 @@ class DeltaTable:
             )
         ]
 
-        if parquet_read_options is not None and parquet_read_options.dictionary_columns:
-            schema = pyarrow.unify_schemas([f.physical_schema for f in fragments])
-        else:
-            schema = self.schema().to_pyarrow()
+        schema = self.schema().to_pyarrow()
+
+        dictionary_columns = format.read_options.dictionary_columns or set()
+        if dictionary_columns:
+            for index, field in enumerate(schema):
+                if field.name in dictionary_columns:
+                    dict_field = field.with_type(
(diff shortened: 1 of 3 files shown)


python/tests/test_table_read.py
```diff
@@ -160,6 +160,21 @@ def test_read_table_with_column_subset():
     )
 
 
+def test_read_table_as_category():
```
wjones127, line 163: We should make sure this new code path works if there are partitioning columns. I think as it is written it likely won't.

python/deltalake/table.py
```diff
@@ -334,9 +334,12 @@ def to_pyarrow_dataset(
             )
         ]
 
-        return FileSystemDataset(
-            fragments, self.schema().to_pyarrow(), format, filesystem
-        )
+        if parquet_read_options is not None and parquet_read_options.dictionary_columns:
+            schema = pyarrow.unify_schemas([f.physical_schema for f in fragments])
```
wjones127, line 338: The physical schema doesn't contain the partition columns, so I'm not confident this will work. Perhaps instead we can modify the schema from `schema = self.schema().to_pyarrow()`?

wjones127 (changes requested):

python/deltalake/table.py
Kuhlwein: You're right, it should be fixed now

wjones127 (approved):
This looks good. Thank you for the contribution!

PR #953, opened 2022-11-21
docs: add simple operations example

roeap:
# Description

Adding a simple example how to use operations APIs to create and work with delta tables and some minor documentation tweaks.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/examples/basic_operations.rs
+++ b/rust/examples/basic_operations.rs
@@ -38,13 +38,14 @@ fn get_table_batches() -> RecordBatch {
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<(), deltalake::DeltaTableError> {
-    // create a delta operations client pointing at an un-initialized im memory location
-    // in a productive environment this would be created with "try_new" and point at
+    // Create a delta operations client pointing at an un-initialized in-memory location.
+    // In a production environment this would be created with "try_new" and point at
     // a real storage location.
     let ops = DeltaOps::new_in_memory();
 
-    // the operations module uses a builder pattern that allows specifying several options
-    // on how the command behaves.
+    // The operations module uses a builder pattern that allows specifying several options
+    // on how the command behaves. The builders implement `Into<Future>`, so once
+    // options are set you can run the command using `.await`.
     let table = ops
         .create()
         .with_columns(get_table_columns())
@@ -59,7 +60,7 @@ async fn main() -> Result<(), deltalake::DeltaTableError> {
 
     assert_eq!(table.version(), 1);
 
-    // setting options allows us to e.g. overwrite the table rather then appending to it.
+    // To overwrite instead of append (which is the default), use `.with_save_mode`:
     let table = DeltaOps(table)
         .write(vec![batch.clone()])
         .with_save_mode(SaveMode::Overwrite)
@@ -68,7 +69,7 @@ async fn main() -> Result<(), deltalake::DeltaTableError> {
     assert_eq!(table.version(), 2);
 
     let (_table, stream) = DeltaOps(table).load().await?;
-    let data = collect_sendable_stream(stream).await?;
+    let data: Vec<RecordBatch> = collect_sendable_stream(stream).await?;
 
     println!("{:?}", data);
 
--- a/rust/examples/basic_operations.rs
+++ b/rust/examples/basic_operations.rs
@@ -1,0 +1,73 @@
+use arrow::{
+    array::{Int32Array, StringArray},
+    datatypes::{DataType, Field, Schema as ArrowSchema},
+    record_batch::RecordBatch,
+};
+use deltalake::operations::collect_sendable_stream;
+use deltalake::{action::SaveMode, DeltaOps, SchemaDataType, SchemaField};
+use std::sync::Arc;
+
+fn get_table_columns() -> Vec<SchemaField> {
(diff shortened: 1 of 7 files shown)


rust/examples/basic_operations.rs
```diff
@@ -0,0 +1,76 @@
...
+    let int_values = Int32Array::from(vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]);
+    let str_values = StringArray::from(vec!["A", "B", "A", "B", "A", "A", "A", "B", "B", "A", "A"]);
+
+    RecordBatch::try_new(schema, vec![Arc::new(int_values), Arc::new(str_values)]).unwrap()
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<(), deltalake::DeltaTableError> {
+    // create a delta operations client pointing at an un-initialized im memory location
+    // in a productive environment this would be created with "try_new" and point at
```
wjones127, line 42:
```suggestion
    // Create a delta operations client pointing at an un-initialized in-memory location.
    // In a production environment this would be created with "try_new" and point at
```

rust/examples/basic_operations.rs
```diff
@@ -0,0 +1,76 @@
...
+    // a real storage location.
+    let ops = DeltaOps::new_in_memory();
+
+    // the operations module uses a builder pattern that allows specifying several options
+    // on how the command behaves.
```
wjones127, line 47:
```suggestion
    // The operations module uses a builder pattern that allows specifying several options
    // on how the command behaves. The builders implement `Into<Future>`, so once
    // options are set you can run the command using `.await`.
```

rust/examples/basic_operations.rs
```diff
@@ -0,0 +1,76 @@
...
+        .await?;
+
+    assert_eq!(table.version(), 0);
+
+    let batch = get_table_batches();
+    let table = DeltaOps(table).write(vec![batch.clone()]).await?;
+
+    assert_eq!(table.version(), 1);
+
+    // setting options allows us to e.g. overwrite the table rather then appending to it.
```
wjones127, line 62:
```suggestion
    // To overwrite instead of append (which is the default), use `.with_save_mode`:
```

rust/examples/basic_operations.rs
```diff
@@ -0,0 +1,76 @@
...
+    let table = DeltaOps(table)
+        .write(vec![batch.clone()])
+        .with_save_mode(SaveMode::Overwrite)
+        .await?;
+
+    assert_eq!(table.version(), 2);
+
+    let (_table, stream) = DeltaOps(table).load().await?;
+    let data = collect_sendable_stream(stream).await?;
```
wjones127, line 71: For the sake of the example, it would be nice to let the user know what type this returns:

```suggestion
    let data: Vec<RecordBatch> = collect_sendable_stream(stream).await?;
```

rust/src/operations/mod.rs
```diff
@@ -95,7 +102,7 @@ impl DeltaOps {
     /// Write data to Delta table
     #[cfg(feature = "datafusion-ext")]
     #[must_use]
-    pub fn write(self, batches: Vec<RecordBatch>) -> WriteBuilder {
+    pub fn write(self, batches: impl IntoIterator<Item = RecordBatch>) -> WriteBuilder {
```
wjones127, line 105: 😄

rust/src/table_properties.rs
```diff
@@ -67,3 +67,5 @@ pub const SET_TRANSACTION_RETENTION_DURATION: &str = "delta.setTransactionRetent
 pub const TARGET_FILE_SIZE: &str = "delta.targetFileSize";
 /// The target file size in bytes or higher units for file tuning. For example, 104857600 (bytes) or 100mb.
 pub const TUNE_FILE_SIZES_FOR_REWRITES: &str = "delta.tuneFileSizesForRewrites";
+/// Invariants defined on the table
+pub const INVARIANTS: &str = "delta.invariants";
```
wjones127, line 71: I don't think this belongs here. INVARIANTS isn't a table-level property, it's at the field level in the schema. See an example here:

https://github.com/delta-io/delta-rs/blob/f3176e0b9fa350cefd462297e4abe861f280612a/python/tests/pyspark_integration/test_write_to_pyspark.py#L57-L68

wjones127 (changes requested):
Thank you for adding this!

rust/src/operations/create.rs
```diff
@@ -157,7 +153,33 @@ impl CreateBuilder {
...
+        mut self,
+        key: impl Into<String>,
+        value: Option<impl Into<String>>,
+    ) -> Self {
+        self.configuration
+            .insert(key.into(), value.map(|v| v.into()));
+        self
+    }
+
+    /// Append custom (application specific) metadata to created table
```
wjones127, line 173: So this metadata doesn't go to the table, but IIRC it goes into the commit info:

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#commit-provenance-information

```suggestion
    /// Append custom (application-specific) metadata to the commit.
    ///
    /// This might include provenance information such as an id of the
    /// user that made the commit or the program that created it.
```

houqp (approved):

wjones127 (approved):

Issue #686, opened 2022-07-11
Python: Automatically convert Pandas types to valid Delta Lake types in write_deltalake()


wjones127:
# Description

Many Pandas types aren't automatically converted into valid Delta Lake types when converted into Arrow tables. For example, Pandas Timestamps are converted into timestamps with nanosecond precision by default, but Delta Lake only supports microsecond precision. This makes `write_deltalake()` difficult to use for Pandas users.

We should write a test that validates all Pandas types can be written with `write_deltalake()` without manual conversion.

I'm not sure yet how to configure the conversion here:

https://github.com/delta-io/delta-rs/blob/431d0ea47700bb3b388a5679ce8d906fa84bb1c6/python/deltalake/writer.py#L128-L129

It's possible that we can pass in an adjusted schema to the schema parameter of `pyarrow.Table.from_pandas()` and that will make the correct conversion. 

**Use Case**

**Related Issue(s)**

Based on #685

blaze225:
Would appreciate if this can be prioritized. Right now this is forcing us to use spark over delta-rs.

ion-elgreco:
This also happens when you write delta from Polars with columns with nano precision datetime. However it's slightly more easy to circumvent you just have to do the casting first to micro precision.

ion-elgreco:
> Would appreciate if this can be prioritized. Right now this is forcing us to use spark over delta-rs.

You can also switch to polars, which casts the dtypes correctly to a delta compatible schema: https://github.com/pola-rs/polars/pull/10165/files#diff-843e4fa7334b1cfcdf4ebe039377c0d724d0abb51bcde68c9aaae1b93868e20b

thehappycheese:
I made this as a stopgap solution. Its a dumb solution but it helped me actually get it to write and test out the library.

```python
import deltalake as dl
from deltalake import DeltaTable
from typing import Union

def strip_categorical(df:pd.DataFrame):
    """convert categorical columns back into integer types,
    and return a dataframe of the categories
    
    Example:

    ```python
    (original_df, categories) = strip_categorical(df)
    ```"""
    categories = {}
    df=df.copy()
    for col in df.columns:
        if pd.api.types.is_categorical_dtype(df[col]):
            print(f"Converting categorical column to integer: '{col}' - {dict(enumerate(df[col].cat.categories))}")
            categories[col] = df[col].cat.categories
            df[col] = df[col].cat.codes
    return df, pd.DataFrame(categories)

[...]

Issue #685, opened 2022-07-11
Cannot write delta table from pandas DataFrame with Timestamp column


ghost:
# Environment

**Delta-rs version**: 0.5.8

**Binding**:  Python

**Environment**:
- **OS**: OSX
- **Other**: Arrow 8.0.0

***
# Bug

**What happened**: The python delta lake writer cannot handle pandas DataFrames with Timestamp columns natively because of the nanosecond precision:

```python
from deltalake.writer import write_deltalake

df = pd.DataFrame({"timestamp": [pd.Timestamp('2022-01-01')]})
write_deltalake("test_with_timestamp", df)
```

yields the following traceback:

```
---------------------------------------------------------------------------
ArrowException                            Traceback (most recent call last)
Input In [403], in <module>
      1 from deltalake.writer import write_deltalake
      3 df = pd.DataFrame({"timestamp": [pd.Timestamp('2022-01-01')]})
----> 4 write_deltalake('test_with_timestamp', df)

File ~/Library/Caches/pypoetry/virtualenvs/test-GfuZs_x0-py3.8/lib/python3.8/site-packages/deltalake/writer.py:226, in write_deltalake(table_or_uri, data, schema, partition_by, filesystem, mode, file_options, max_open_files, max_rows_per_file, min_rows_per_group, max_rows_per_group, name, description, configuration, overwrite_schema)
    208 ds.write_dataset(
    209     data,
    210     base_dir=table_uri,
   (...)
    222     max_rows_per_group=max_rows_per_group,
    223 )
    225 if table is None:
--> 226     _write_new_deltalake(  # type: ignore[call-arg]
    227         table_uri,
    228         schema,
    229         add_actions,
    230         mode,
    231         partition_by or [],
    232         name,
    233         description,
    234         configuration,
    235     )
    236 else:
    237     table._table.create_write_transaction(
    238         add_actions,
    239         mode,
    240         partition_by or [],
    241         schema,
    242     )

ArrowException: Schema error: Invalid data type for Delta Lake: Timestamp(Nanosecond, None)
```

The same issue happens if I do `write_deltalake("test_with_timestamp", pyarrow.Table.from_pandas(df))`

**What you expected to happen**:

I would have expected this to work out of the box with pandas, without having a priori knowledge of which columns contain timestamps and having to create a new arrow table with downcasted values. Note that the call to `pyarrow.dataset.write_dataset()` is not the issue.

wjones127:
[...]

PR #958, opened 2022-11-26
Handle pandas timestamps

hayesgb:
# Description
As described in #686 some pandas datatypes are not converted to a format that is compatible with delta lake.  This handles the instance of timestamps, which are stored with `ns` resolution in Pandas.  Here, if is a schema is not provided, we specify converting the timestamps to `us` resolution.

We also update `python/tests/test_writer.py::test_write_pandas` to reflect this change.

# Related Issue(s)
#685

### DIFF
--- a/python/deltalake/schema.py
+++ b/python/deltalake/schema.py
@@ -1,7 +1,40 @@
-from typing import Union
+from typing import Tuple, Union
+
+import pandas as pd
+import pyarrow as pa
 
 from ._internal import ArrayType, Field, MapType, PrimitiveType, Schema, StructType
 
 # Can't implement inheritance (see note in src/schema.rs), so this is next
 # best thing.
 DataType = Union["PrimitiveType", "MapType", "StructType", "ArrayType"]
+
+
+def delta_arrow_schema_from_pandas(
+    data: pd.DataFrame,
+) -> Tuple[pd.DataFrame, pa.Schema]:
+    """ "
+    Infers the schema for the delta table from the Pandas DataFrame.
+    Necessary because of issues such as:  https://github.com/delta-io/delta-rs/issues/686
+
+    :param data: Data to write.
+    :returns A Pyarrow Table and the inferred schema for the Delta Table
+    """
+
+    table = pa.Table.from_pandas(data)
+    _schema = table.schema
+    schema_out = []
+    for _field in _schema:
+        if isinstance(_field.type, pa.TimestampType):
+            f = pa.field(
+                name=_field.name,
+                type=pa.timestamp("us"),
+                nullable=_field.nullable,
+                metadata=_field.metadata,
+            )
+            schema_out.append(f)
+        else:
+            schema_out.append(_field)
+    schema = pa.schema(schema_out, metadata=_schema.metadata)
+    data = pa.Table.from_pandas(data, schema=schema)
+    return data, schema
--- a/python/deltalake/schema.py
+++ b/python/deltalake/schema.py
@@ -12,7 +12,7 @@ DataType = Union["PrimitiveType", "MapType", "StructType", "ArrayType"]
 
 def delta_arrow_schema_from_pandas(
     data: pd.DataFrame,
-) -> Tuple[pd.DataFrame, pa.Schema]:
+) -> Tuple[pa.Table, pa.Schema]:
     """ "
     Infers the schema for the delta table from the Pandas DataFrame.
     Necessary because of issues such as:  https://github.com/delta-io/delta-rs/issues/686
--- a/python/deltalake/schema.py
+++ b/python/deltalake/schema.py
@@ -35,6 +35,6 @@ def delta_arrow_schema_from_pandas(
             schema_out.append(f)
         else:
             schema_out.append(field)
-    schema = pa.schema(schema_out, metadata=_schema.metadata)
+    schema = pa.schema(schema_out, metadata=schema.metadata)
     data = pa.Table.from_pandas(data, schema=schema)
     return data, schema
--- a/python/deltalake/schema.py
+++ b/python/deltalake/schema.py
@@ -1,7 +1,9 @@
-from typing import Tuple, Union
+from typing import Tuple, Union, TYPE_CHECKING
 
-import pandas as pd
 import pyarrow as pa
+
+if TYPE_CHECKING:
+    import pandas as pd
 
 from ._internal import ArrayType, Field, MapType, PrimitiveType, Schema, StructType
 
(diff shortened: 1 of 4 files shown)


python/deltalake/writer.py
```diff
@@ -132,8 +132,28 @@ def write_deltalake(
     :param overwrite_schema: If True, allows updating the schema of the table.
     :param storage_options: options passed to the native delta filesystem. Unused if 'filesystem' is defined.
     """
+
     if _has_pandas and isinstance(data, pd.DataFrame):
-        data = pa.Table.from_pandas(data)
+        if schema is not None:
+            data = pa.Table.from_panda(data, schema=schema)
```
wjones127, line 138: Is this code path not tested? We might want to add a test for that.

```suggestion
            data = pa.Table.from_pandas(data, schema=schema)
```

python/deltalake/writer.py
```diff
@@ -132,8 +132,28 @@ def write_deltalake(
...
+        else:
+            _data = pa.Table.from_pandas(data)
+            _schema = _data.schema
+            schema_out = []
+            for _field in _schema:
```
wjones127, line 143: Would you be willing to pull this out into a helper function (`_delta_arrow_schema_from_pandas`)?

wjones127 (changes requested):
Thanks for the contribution! Have two requests on it.

python/deltalake/writer.py
hayesgb: Thanks for the catch.  Added a test for this as well.

python/deltalake/writer.py
hayesgb: Really like this idea.  Factored this out into a helper function, but I'd prefer it to be public rather than private.  Would like to leverage this helper function elsewhere.  Thoughts?

hayesgb:
Not sure why we `mypy` is complaining.

houqp:
@hayesgb you might need to add timestamp to https://github.com/delta-io/delta-rs/blob/main/python/stubs/pandas/__init__.pyi in order to fix the mypy error

python/deltalake/writer.py
wjones127: Sure, but if you make it public, then I'd suggest putting it in `schema.py`.

hayesgb: Done.

python/deltalake/schema.py
```diff
@@ -1,7 +1,40 @@
...
 from ._internal import ArrayType, Field, MapType, PrimitiveType, Schema, StructType
 
 # Can't implement inheritance (see note in src/schema.rs), so this is next
 # best thing.
 DataType = Union["PrimitiveType", "MapType", "StructType", "ArrayType"]
+
+
+def delta_arrow_schema_from_pandas(
+    data: pd.DataFrame,
+) -> Tuple[pd.DataFrame, pa.Schema]:
```
wjones127, line 15:
```suggestion
) -> Tuple[pa.Table, pa.Schema]:
```

python/deltalake/schema.py
```diff
@@ -1,7 +1,40 @@
...
+    """ "
```
wjones127, line 16:
```suggestion
    """
```

python/deltalake/schema.py
```diff
@@ -1,7 +1,40 @@
...
+        if isinstance(_field.type, pa.TimestampType):
+            f = pa.field(
+                name=_field.name,
+                type=pa.timestamp("us"),
+                nullable=_field.nullable,
+                metadata=_field.metadata,
+            )
+            schema_out.append(f)
+        else:
+            schema_out.append(_field)
```
wjones127, line 37:
```suggestion
    schema = table.schema
    schema_out = []
    for field in schema:
        if isinstance(field.type, pa.TimestampType):
            f = pa.field(
                name=field.name,
                type=pa.timestamp("us"),
                nullable=field.nullable,
                metadata=field.metadata,
            )
            schema_out.append(f)
        else:
            schema_out.append(field)
```

wjones127 (changes requested):

python/deltalake/schema.py
```diff
@@ -1,7 +1,40 @@
...
+
+def delta_arrow_schema_from_pandas(
+    data: pd.DataFrame,
+) -> Tuple[pa.Table, pa.Schema]:
+    """
+    Infers the schema for the delta table from the Pandas DataFrame.
+    Necessary because of issues such as:  https://github.com/delta-io/delta-rs/issues/686
+
+    :param data: Data to write.
+    :returns A Pyarrow Table and the inferred schema for the Delta Table
```
wjones127, line 21:
```suggestion
    :return: A PyArrow Table and the inferred schema for the Delta Table
```

python/deltalake/schema.py
```diff
@@ -1,7 +1,40 @@
...
+            f = pa.field(
+                name=field.name,
+                type=pa.timestamp("us"),
+                nullable=field.nullable,
+                metadata=field.metadata,
+            )
+            schema_out.append(f)
+        else:
+            schema_out.append(field)
+    schema = pa.schema(schema_out, metadata=_schema.metadata)
```
wjones127, line 38:
```suggestion
    schema = pa.schema(schema_out, metadata=schema.metadata)
```

python/deltalake/schema.py
```diff
@@ -1,7 +1,40 @@
...
 
 from ._internal import ArrayType, Field, MapType, PrimitiveType, Schema, StructType
 
 # Can't implement inheritance (see note in src/schema.rs), so this is next
 # best thing.
 DataType = Union["PrimitiveType", "MapType", "StructType", "ArrayType"]
+
+
+def delta_arrow_schema_from_pandas(
+    data: pd.DataFrame,
```
wjones127, line 14:
```suggestion
    data: "pd.DataFrame",
```

python/deltalake/schema.py
```diff
@@ -1,7 +1,40 @@
-from typing import Union
+from typing import Tuple, Union
+
+import pandas as pd
+import pyarrow as pa
```
wjones127, line 4: I *think* this is the last thing: We'll want to keep Pandas an optional dependency:

```suggestion
from typing import Tuple, Union, TYPE_CHECKING

import pyarrow as pa

if TYPE_CHECKING:
    import pandas as pd
```

python/deltalake/schema.py
```diff
@@ -1,7 +1,42 @@
-from typing import Union
+from typing import Tuple, Union, TYPE_CHECKING
```
wjones127, line 1:
```suggestion
from typing import TYPE_CHECKING, Tuple, Union
```

wjones127 (approved):
The CI test is being flaky. I tested locally tested and all tests pass. 

Thank you @hayesgb!

Issue #604, opened 2022-05-15
Support CDC


houqp:
# Description

See design doc at: https://docs.google.com/document/d/1FQSl0S0Irckim331YvcAKyxxGBhdLiJcJH6v3fEyiC0/edit#. This will make non-append only table usable as a streaming source!

kietay:
I could take this on as would be useful in current project

roeap:
We welcome all contributions, large and small - my assumption this would fall well in the former category 🙂.

I only skimmed the design docs, but it seems we will have to work in the datafusion / query integration quite a bit for all to work...

houqp:
Yes, reading should be pretty straight forward to implement and will help unlock a lot of the streaming use-cases.

kietay:
@roeap I'm sorry I missed this notification - let me do some digging into this to understand the problem and then propose a change to see what you think

PR #959, opened 2022-11-26
feat: add the cdc actions to out log parsing

herry13:
# Description

This PR...
- [x] adds a new struct `AddCDCFile` as defined in [the specification of the CDC action](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-cdc-file)
- [ ] adds the CDC actions to out log parsing

# Related Issue(s)

[#604](https://github.com/delta-io/delta-rs/issues/604)

# Documentation

- [Delta Lake Change Data Feed](https://docs.google.com/document/d/1FQSl0S0Irckim331YvcAKyxxGBhdLiJcJH6v3fEyiC0/edit#)
- [Protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-cdc-file)

### DIFF
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -149,6 +149,22 @@ pub struct StatsParsed {
     pub null_count: HashMap<String, DeltaDataTypeLong>,
 }
 
+/// Delta AddCDCFile action that describes a parquet CDC data file.
+#[derive(Serialize, Deserialize, Clone, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct AddCDCFile {
+    /// A relative path, from the root of the table, to a CDC file
+    pub path: String,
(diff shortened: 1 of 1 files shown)


roeap:
Thanks for the contribution @herry13! Looks good so far, but we need to do a few more things to make this work :).

FOr one we need to add the new CDC struct as a variant to the `Action` enum in the same file. Then - i think, haven't read the specs - we need to add implementation for the cdc struct in the checkpoint parquet parser (ideally 1 & 2) - the logic should be quite similar to e.g. the add action, but hopefully a little less complex.

Last but not least a test would be nice as well :).

rtyler:
I am going to close this pull request, I believe this support already exists in `main` but we didn't update open issues and pull requests.

PR #976, opened 2022-11-29
Add a new release github action for Python binding: macos with universal2 wheel

### LINKED ISSUE #972: MacOS arm64 wheels are generated incorrectly
# Environment

**Binding**: Python

**Environment**:
- **OS**: MacOS 13 (M1 / arm64)
- Python 3.10.8
- Pip 22.3.1

***

# Bug

**What happened**:

On the PyPI releases page (https://pypi.org/project/deltalake/0.6.4/#files), the arm64 wheel file is published as `
deltalake-0.6.4-cp37-abi3-macosx_10_9_arm64.whl`. However, it should be `macosx_11_0_arm64.whl`. This causes pip to not install from source instead of using the pre-built wheel.

**What you expected to happen**: Pip should install from wheel.

**More details**:

Note that `pip -v debug` does not contain `macosx_10_9_arm64` but does contain `macosx_11_0_arm64`. As such, `macosx_10_9_arm64` is never a compatible ABI tag. I believe the actions release workflow is not generating the correct ABI tag.

Release log: https://github.com/delta-io/delta-rs/actions/runs/3567491762/jobs/5995258547#step:3:668

Possible fixes
1. build universal2 wheels on mac (not sure if this is actually possible for delta-rs)
2. increase the min macosx version to 11_0 for the arm64 wheel.

fvaleye:
# Description
- Add a new Python binding release Github Action for `universal2` wheel in PyPy

# Related Issue(s)
- closes #972 

# Documentation
- **universal2:**
This wheel contains both architectures, causing it to be up to twice the size (data files do not get doubled, only compiled code). It requires Pip 20.3 (Packaging 20.6+) to load on Intel, and Pip 21.0.1 (Packaging 20.9+) to load on Apple Silicon.


.github/workflows/python_release.yml
```diff
@@ -35,14 +35,30 @@ jobs:
...
+    - uses: actions/checkout@v3
+    
+    - name: Publish to pypi (without sdist)
+    uses: messense/maturin-action@v1
+    env:
+        MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+    with:
+        target: ${{ matrix.target }}
+    command: publish
+    args: -m python/Cargo.toml --no-sdist --universal2
```
wjones127, line 60: Is this indentation correct? are these all supposed to be part of the step? And is `command` and `args` supposed to be under `with`?

fvaleye: Oh, thanks; good catch!

Indeed, I need to reformat the file 👍

wjones127 (approved):
Thanks!

PR #980, opened 2022-12-01
feat: check invariants in write command

roeap:
# Description

This PR integrates the `DeltaDataChecker` in the write path of the operations API. I was unsure whether to integrate this in the (experimental) writer in the operations module, but opted for keeping the writer itself focussed on the lower level write operations. 

# Related Issue(s)

# Documentation


wjones127 (approved):
This looks great! Seems like the right place to put it to me.

PR #988, opened 2022-12-03
Adds handling of nested partitions

hayesgb:
# Description
The python writer includes a function `get_partitions_from_path` that currently handles a path like `part=A/filename.parquet` properly.  However, it fails to handle `first_folder/second_folder/part=A/filename.parquet`.  This PR extends the function to handle these second cases properly.

This PR also adds a test for these cases.

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -210,7 +210,6 @@ def write_deltalake(
 
     def visitor(written_file: Any) -> None:
         path, partition_values = get_partitions_from_path(written_file.path)
-        print(written_file.path)
         stats = get_file_stats_from_metadata(written_file.metadata)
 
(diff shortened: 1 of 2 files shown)


python/deltalake/writer.py
```diff
@@ -210,6 +210,7 @@ def write_deltalake(
 
     def visitor(written_file: Any) -> None:
         path, partition_values = get_partitions_from_path(written_file.path)
+        print(written_file.path)
         stats = get_file_stats_from_metadata(written_file.metadata)
```
fvaleye, line 214: I think you forgot this print statement!

wjones127:
@hayesgb do you have an example invocation of `write_deltalake()` that would trigger this? `write_deltalake()` always writes with Hive-style partitioning so not sure how these other folders would be added.

PR #1005, opened 2022-12-09
Loosen version requirement for maturin

### LINKED ISSUE #1004: Python package: Loosen version requirements for maturin
# Description

The current python package requires an exact maturin version (currently 0.14.2): 

https://github.com/delta-io/delta-rs/blob/c1f5a7e9497fceeea4d5b20c9eaf6ebe109f1b65/python/pyproject.toml#L1-L2

Would you consider loosening it, for example to what is suggested in maturin's Readme?
```toml
[build-system]
requires = ["maturin>=0.14,<0.15"]
```

**Use Case**

This would let us build the python package with the latest maturin version (currently 0.14.5) rather than requiring an older version.
In my case it is to package the deltalake python package for archlinux.

The current workaround is to build the package in a virtualenv, but it's generally discouraged when a "direct" build is possible.

Thanks!

gyscos:
# Description

This loosen the version requirement for maturin when building the python package.

# Related Issue(s)

- closes #1004

# Documentation

Using an exact version can be excessively restrictive, for example preventing the use of the latest (but still compatible) version of maturin. Especially when packaging for rolling-releases distributions (like Archlinux), being able to build with the latest version would be beneficial.

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := help
 
 VENV := venv
-MATURIN_VERSION := $(shell awk -F '[ ="]+' '$$1 == "requires" { print $$4 }' pyproject.toml)
+MATURIN_VERSION := $(shell grep 'requires =' pyproject.toml | cut -d= -f2- | tr -d '[ "]')
 PACKAGE_VERSION := $(shell grep version Cargo.toml | head -n 1 | awk '{print $$3}' | tr -d '"' )
 
 .PHONY: setup-venv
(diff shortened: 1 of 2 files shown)


gyscos:
Makefile updated to properly parse the maturin version.

(Note that the entire build step could also be done as simply as `python -m build`, and it will use virtualenv by default. Was there a reason for the more complex Makefile solution?)

wjones127:
> (Note that the entire build step could also be done as simply as python -m build, and it will use virtualenv and maturin as expected. Was there a reason for the more complex Makefile solution?)

That's an interesting point. I'll want to try that out and see how it works in the development workflow, but potentially could be a useful way to go.

fvaleye (approved):
Good catch, thanks @gyscos 👍

gyscos:
It looks like one test timeouted, not sure if it's related to this PR.

PR #1016, opened 2022-12-13
feat: make `DeltaStorageHandler` pickle serializable

### LINKED ISSUE #1015: DeltaStorageHandler is not serializable
# Environment

**Delta-rs version**: latest

**Binding**: 
Python

**Environment**:
- **Cloud provider**: all
- **OS**: mac ventura, m1
- **Other**:

***
# Bug

**What happened**:
I am trying to add delta connector to [Polars](https://github.com/pola-rs/polars), but which doing a lazy scan polars tries to serialize function and objects, see [here](https://github.com/pola-rs/polars/blob/master/py-polars/polars/internals/anonymous_scan.py#L81). This line breaks with the below exception as `DeltaFileSystemHandler` is not serializable I suppose,
```
TypeError: cannot pickle 'builtins.DeltaFileSystemHandler' object
```

It will be great if this could be serializable, otherwise on Polars side we will have to rely on `fsspec` or `pyarrow.fs`, bypassing the `DeltaStorageHandler` implementation altogether.

Exception should not occur.

**How to reproduce it**:
You can use the following minimal code to reproduce this.

```
import deltalake
from deltalake.fs import DeltaStorageHandler
dsh = DeltaStorageHandler("./delta-table")

import pickle
pickle.dumps(dsh)
TypeError: cannot pickle 'builtins.DeltaFileSystemHandler' object

```

**More details**:

roeap:
# Description

Integrating with polars requires the `DeltaStorageHandler` to be serializable with pickle. this PR implements the required dunder methods to make it so...

Unfortunately we lost the ability to instantiate the `DeltaStorageHandler` with an existing object store, however I do believe that this is not a critical loss.

cc @chitralverma @ritchie46

# Related Issue(s)

closes #1015

# Documentation


python/deltalake/table.py
```diff
@@ -317,7 +317,7 @@ def to_pyarrow_dataset(
                 DeltaStorageHandler(
                     self._table.table_uri(),
                     self._storage_options,
-                    self._table.get_py_storage_backend(),
+                    # self._table.get_py_storage_backend(),
```
wjones127, line 320: delete this?

python/src/filesystem.rs
```diff
@@ -242,6 +254,35 @@ impl DeltaFileSystemHandler {
...
+            .map_err(PyDeltaTableError::from_raw)?;
+
+        self.inner = storage;
+        self.rt = Arc::new(rt()?);
+        self.config = config;
+
+        Ok(())
+    }
+
+    pub fn __getnewargs__(&self) -> PyResult<(String, Option<HashMap<String, String>>)> {
```
wjones127, line 280: I wonder if just this method is sufficient. If we define this, do we still need `__setstate__` and `__getstate__`?

python/src/lib.rs
```diff
@@ -83,6 +84,8 @@ enum PartitionFilterValue<'a> {
 #[pyclass]
 struct RawDeltaTable {
     _table: deltalake::DeltaTable,
+    // storing the config additionally on the table helps us make pickling work.
+    _config: FsConfig,
```
wjones127, line 88: Why is that? I wouldn't expect we would need to modify this struct to make the storage handler pickle-able.

wjones127 (commented):
This is cool!

I'm new to pickling myself, so I might be wrong on my suggestions, but it does seem like we might be able to simplify this implementation.

python/src/filesystem.rs
roeap: Initially added this to fix a bug, but you are correct, this si all we need. Removing the rest makes the implementation for pickling quite minimal 😆.

python/src/lib.rs
roeap: I went this way to ceated the `DeltaFileSystemHandler` from the delta table instance, since we need the options to be avaialable even if we reuse the underlying object store. An alternative would have been to expose the option on the object store itself, but somehow I felt better not exposing these at the lower level.

A point of concern by the way was, that when pickling, we are writing potentially long lived secrets to disc in plain text. Should we add a comment to that somewhere?

wjones127 (approved):

python/src/lib.rs
wjones127: Yeah not sure where to leave a comment. But I think if people should implicitly understand that a storage handler contains secrets and they should be careful where they send the pickles.

Issue #971, opened 2022-11-28
Release delta-rs `0.6.0` (next release after `0.5.0`)


iajoiner:
Here is our first monthly release:
- Planned Release and Publish to crates.io: 2022-12-16

CI and release-related enhancements we want to add in this or the next release:
- [x] #1000 
- [ ] #996 
- [ ] #984
- [ ] #1026

Bugs we want to fix in this or the next release:
- [ ]  #677 / #925 / #985
- [ ] #973 / #1002
- [ ] #1025

Items:

- [x] PR to update version and CHANGELOG (#1023):
- [ ] Update version to 0.6.0 and add changelog
- [ ] Release candidate created:
- [ ] Release candidate approved:
- [ ] Release to crates.io:

See full list here:
https://github.com/delta-io/delta-rs/compare/rust-v0.5.0...main
cc @houqp @wjones127 @fvaleye @rtyler @roeap @xianwill @mosyp

iajoiner:
Please let me know whether the schedule is ok.

roeap:
Datafusion is in the process of publishing the next release - given the timeline and our (at least current) ties to datafusion for some of the operations, does it maybe make sense to release a bit earlier - lets say a week after datafusion - and then stick to the montly schedule from there on?

This might also make conflicts with holiday obligations easier for the current release cycle :)

iajoiner:
It is release time! :)

roeap:
We are still finishing up merging the datafusion update ... but are in the last steps ..

iajoiner:
@roeap Sure. I will generate the CHANGELOG now.

PR #1021, opened 2022-12-15
Support DataFusion 15

### LINKED ISSUE #1020: Support Apache Arrow DataFusion 15
# Description

**Use Case**

Apache Arrow DataFusion is at version 15. Make Delta-RS compatible with DataFusion 15.

**Related Issue(s)**

andrei-ionescu:
# Description

Add support for DataFusion 15. 

Upgrades dependencies:
- `DataFusion` from `14` to `15`
- `Arrow` from `26` to `28`

# Related Issue(s)

- closes #1020 

# Documentation


roeap:
Hi @andrei-ionescu thanks for the contribution, in this case  unfortunately there already is a PR open for this :) #1014

rtyler:
@roeap since this pull request is smaller and more isolated from other changes, is there a problem with working to bring this in? That would allow us to evaluate the non-string partition changes more independent of the datafusion changes

rust/src/delta_datafusion.rs
```diff
@@ -373,25 +375,44 @@ impl TableProvider for DeltaTable {
...
+            .partition_columns
+            .clone()
+            .into_iter()
+            .flat_map(|c| {
+                let f_types = schema
+                    .fields()
+                    .iter()
+                    .map(|e| (e.name(), e.data_type()))
+                    .filter(|&(col_name, _)| c.eq_ignore_ascii_case(col_name))
+                    .collect::<Vec<_>>();
```
roeap, line 389: maybe use `schema.field_with_name(col)?.data_type()` here to avoid the iteration?

rust/src/delta_datafusion.rs
```diff
@@ -373,25 +375,44 @@ impl TableProvider for DeltaTable {
...
+
+        let parquet_scan = ParquetFormat::new(session.config_options())
             .create_physical_plan(
                 FileScanConfig {
                     object_store_url: self.storage.object_store_url(),
                     file_schema,
                     file_groups: file_groups.into_values().collect(),
                     statistics: self.datafusion_table_statistics(),
-                    projection: projection.clone(),
+                    projection: projection.map(|o| o.to_owned()),
```
roeap, line 409: `projectction.cloned()` should do the trick as well.

roeap, line 409: My suggestion here would be to just clone the partition cols, then use them to filter the file schema fields. and afterwards use them to create the tuple array needed. Thus we avoid creating a unnecessary intermediate structure.

rust/src/delta_arrow.rs
```diff
@@ -205,12 +200,10 @@ impl TryFrom<&ArrowField> for schema::SchemaField {
             arrow_field.is_nullable(),
             arrow_field
                 .metadata()
-                .as_ref()
-                .map_or_else(HashMap::new, |m| {
-                    m.iter()
-                        .map(|(k, v)| (k.clone(), serde_json::Value::String(v.clone())))
-                        .collect()
-                }),
+                .to_owned()
```
roeap, line 203: i think we can avoid the `to_owned` call here since we are cloning the elements already inside the map.

rust/src/delta_arrow.rs
```diff
@@ -36,16 +34,13 @@ impl TryFrom<&schema::SchemaField> for ArrowField {
...
-                metadata
-                    .iter()
-                    .map(|(key, val)| Ok((key.clone(), serde_json::to_string(val)?)))
-                    .collect::<Result<_, serde_json::Error>>()
-                    .map_err(|err| ArrowError::JsonError(err.to_string()))
-            })
-            .transpose()?;
+        let metadata = f
+            .get_metadata()
+            .to_owned()
```
roeap, line 39: this `to_owned` is redundant I think.

roeap (changes requested):

roeap:
@rtyler - certainly, whatever gets us there quickest :).

@andrei-ionescu - In order to make the tests pass, we need to update the schema validation in the `operations/writer` file as well. you can look at #1014 to see one option to do this.

rust/src/delta_datafusion.rs
andrei-ionescu: The `projection` is of `Option<&Vec<usize>>` type. I wanted to deref the inner value. Switching to `projection.cloned()`.

andrei-ionescu:
@roeap, @rtyler, I implemented the suggested feedback. Thanks for giving the me hint about #1014 PR. 

All checks did pass except the **"Python 3.10 PyArrow latest"** step. It seems that it did timeout. Do you know the root cause of this? Is it transient?

roeap:
Thanks @andrei-ionescu, excellent work!. The python 3.10 test is in fact flaky and not caused by this PR.

One last ask, there is a test for the new type handling in datafusion 15 in the other PR. As we do the migration here, could you move the test as well? Otherwise looks great and we should be able to merge before #971 happens.

andrei-ionescu:
@roeap Thanks for the suggestion with the test. Added it into here.

roeap:
@andrei-ionescu thanks for adding the test - looks great! - It seems thought that a single dev artifact slipped into the code. There is a new parquet file in a new `data/` folder within in root of the repo, that i cannot place.

andrei-ionescu:
@roeap: You're right! I may have been running a cargo command on the root of the project that created that I did not noticed it in the PR. Sorry for that! I removed it.

roeap:
Looking good - waiting for CI to finish and we are ready to merge :)

PR #1007, opened 2022-12-11
fix truncating signature on SAS

damiondoesthings:
# Description
This fixes the truncation of Azure SAS signatures ending in "=", i.e. when using non-url-encoded SAS tokens.

# Related Issue(s)

closes #1003

PR #1036, opened 2022-12-21
fix: azure sas key url encoding

### LINKED ISSUE #910: Azure SAS authorization fails with `<AuthenticationErrorDetail>Signature fields not well formed.`
# Environment

**Delta-rs version**: 0.6.2
**Binding**: python

Using a sas token to access a deltalake on AzureStorage throws an authorization error. The sas token was verified out of band.

Interestingly, the thrown error provides a failed blob url. Comparing this with a working url, the sas fields in the failed url are url-quoted. For instance:
```
>>> from urllib.parse import quote_plus
>>> failed_url_sas_sig == quote_plus(working_url_sas_sig)
True
```

roeap:
# Description

The object store crate encode the sas key values internally. When passing in an already encoded sas key, they would get double encoded. Since sas keys acquired from the azure portal and azure storage explorer are encoded, and this is "ususally" how a user would get them we decode them before passign them to object store.

I was not completely sure what the desired bahvior should be, alternatively we could update the documentation, to show a sas keys from the portal should be handled.

```py
import deltalake
from urllib.parse import unquote

storage_options = dict(account_name=storage_account, sas_token=unquote(sas))
dt = deltalake.DeltaTable(azure_uri, storage_options=storage_options)
```

# Related Issue(s)

closes #910

# Documentation

### DIFF
--- a/rust/src/builder/mod.rs
+++ b/rust/src/builder/mod.rs
@@ -37,7 +37,7 @@ enum BuilderError {
     Required(String),
     #[error("Failed to find valid credential.")]
     MissingCredential,
-    #[error("Failed to decode sas key. {0}")]
+    #[error("Failed to decode SAS key: {0}\nSAS keys must be percent-encoded. They come encoded in the Azure portal and Azure Storage Explorer.")]
     Decode(String),
 }
 
--- a/rust/src/builder/mod.rs
+++ b/rust/src/builder/mod.rs
@@ -37,6 +37,8 @@ enum BuilderError {
     Required(String),
     #[error("Failed to find valid credential.")]
     MissingCredential,
+    #[error("Failed to decode sas key. {0}")]
+    Decode(String),
 }
 
 impl From<BuilderError> for DeltaTableError {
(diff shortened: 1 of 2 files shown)


dominik7680:
The proposed workaround
```
import deltalake
from urllib.parse import unquote

storage_options = dict(account_name=storage_account, sas_token=unquote(sas))
dt = deltalake.DeltaTable(azure_uri, storage_options=storage_options)
```
does not properly work in latest deltalake-0.6.4. 
SIG section in the SAS token ends with `%3D`, which is `=` URL encoded

In result, PyDeltaTable is trying to access the URL with removed `=` from the end.
Ex.:
Original URL Suffix:
image

URL Suffix from the `PyDeltaTableError: Signature fields not well formed.` exception
image

roeap:
That is correct. THis issue has been fixed in a previous PR that is on main, but not yet in the released python bindings. #1007

rust/src/builder/mod.rs
```diff
@@ -37,6 +37,8 @@ enum BuilderError {
     Required(String),
     #[error("Failed to find valid credential.")]
     MissingCredential,
+    #[error("Failed to decode sas key. {0}")]
```
wjones127, line 40: Perhaps it's worth adding more detail to the error message:

```suggestion
    #[error("Failed to decode SAS key: {0}\nSAS keys must be percent-encoded. They come encoded in the Azure portal and Azure Storage Explorer.")]
```

wjones127 (approved):

PR #1043, opened 2022-12-28
feat: enable passing storage options to Delta table builder via DataFusion's CREATE EXTERNAL TABLE

gruuya:
# Description
We've recently added Delta table support to [Seafowl](https://github.com/splitgraph/seafowl) using delta-rs, which utilizes the new `OPTIONS` clause in sqlparser/DataFusion. It allows propagating a set of key/values down to the `DeltaTableBuilder`, which in turn can use those to instantiate a corresponding object store client. This means someone can now define a delta table without relying on env vars as:
```sql
CREATE EXTERNAL TABLE my_delta
STORED AS DELTATABLE
OPTIONS ('AWS_ACCESS_KEY_ID' 'secret', 'AWS_SECRET_ACCESS_KEY' 'also_secret', 'AWS_REGION' 'eu-west-3') 
LOCATION 's3://my-bucket/my-delta-table/'
```

I've also changed the existing datafusion integration tests to use this approach to exercise it.

I'm not sure whether it makes sense to merge this PR upstream, but opening this PR just in case it does.

# Related Issue(s)
Didn't find any related issues.

# Documentation

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -1482,6 +1482,19 @@ pub async fn open_table(table_uri: impl AsRef<str>) -> Result<DeltaTable, DeltaT
     Ok(table)
 }
 
+/// Same as `open_table`, but also accepts storage options to aid in building the table for a deduced
+/// `StorageService`.
+pub async fn open_table_with_storage_options(
+    table_uri: impl AsRef<str>,
+    storage_options: HashMap<String, String>,
+) -> Result<DeltaTable, DeltaTableError> {
+    let table = DeltaTableBuilder::from_uri(table_uri)
+        .with_storage_options(storage_options)
+        .load()
+        .await?;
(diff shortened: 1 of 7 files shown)


houqp (approved):

houqp:
Thanks @gruuya , very cool feature. Looks like the test refactor broke some of the tests, could you fix that?

gruuya:
Thanks, sure thing! 

I realized that I tested the refactored `test_datafusion_sql_registration` with `integration_test` feature on, so I didn't catch the build problem. As a resolution I've now moved the new DF test utility function into `tests/common/datafusion`.

houqp (approved):
LGTM. @roeap do you want to take another look?

wjones127 (approved):
This looks good to me as well. Thanks for contributing this!

Issue #1024, opened 2022-12-17
Support for LargeUtf8 column type


chitralverma:
# Environment

**Delta-rs version**: latest

**Binding**: 
Python

**Environment**:
- **Cloud provider**: all
- **OS**: mac ventura, m1
- **Other**:

***
# Bug

**What happened**:
While writing a delta table to a location from a PyArrow table, the writer breaks with the following error,
```
deltalake.PyDeltaTableError: Schema error: Invalid data type for Delta Lake: LargeUtf8
```

Not sure if this is write only issue or also a read issue.

**What you expected to happen**:
PyArrow types like `LargeUtf8`, `LargeString` should be supported.

**How to reproduce it**:
You can use the following minimal code to reproduce this.

```
import pyarrow as pa
from deltalake.writer import write_deltalake

# Create a pyarrow table with LargeString column 'name'
pylist = [{'name': 'Joey', 'age': 14}, {'name': 'Ivan', 'age': 32}]
schema = pa.schema([pa.field('name', pa.large_string()), pa.field('age', pa.int64())])
t = pa.Table.from_pylist(pylist, schema=schema)

write_deltalake("/tmp/test-delta", data=t)
```

**More details**:

houqp:
Deltalake itself doesn't distinguish between large and regular string/byte bytes: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types. I think we can just represent them using the primitive `string` and `binary` types.

PR #1044, opened 2022-12-30
Support for LargeUtf8 and LargeBinary

chitralverma:
# Description
Added missing mapping from below mentioned arrow types to delta types,
- `LargeUtf8` (LargeString) -> `string`
- `LargeBinary` -> `binary`

# Related Issue(s)
closes https://github.com/delta-io/delta-rs/issues/1024

### DIFF
--- a/rust/src/delta_arrow.rs
+++ b/rust/src/delta_arrow.rs
@@ -211,6 +211,7 @@ impl TryFrom<&ArrowDataType> for schema::SchemaDataType {
     fn try_from(arrow_datatype: &ArrowDataType) -> Result<Self, ArrowError> {
         match arrow_datatype {
             ArrowDataType::Utf8 => Ok(schema::SchemaDataType::primitive("string".to_string())),
+            ArrowDataType::LargeUtf8 => Ok(schema::SchemaDataType::primitive("string".to_string())),
             ArrowDataType::Int64 => Ok(schema::SchemaDataType::primitive("long".to_string())), // undocumented type
             ArrowDataType::Int32 => Ok(schema::SchemaDataType::primitive("integer".to_string())),
             ArrowDataType::Int16 => Ok(schema::SchemaDataType::primitive("short".to_string())),
@@ -219,6 +220,7 @@ impl TryFrom<&ArrowDataType> for schema::SchemaDataType {
             ArrowDataType::Float64 => Ok(schema::SchemaDataType::primitive("double".to_string())),
             ArrowDataType::Boolean => Ok(schema::SchemaDataType::primitive("boolean".to_string())),
             ArrowDataType::Binary => Ok(schema::SchemaDataType::primitive("binary".to_string())),
+            ArrowDataType::LargeBinary => Ok(schema::SchemaDataType::primitive("binary".to_string())),
             ArrowDataType::Decimal128(p, s) => Ok(schema::SchemaDataType::primitive(format!(
                 "decimal({},{})",
                 p, s


chitralverma:
@houqp  this should be enough to close https://github.com/delta-io/delta-rs/issues/1024  right?

wjones127:
@chitralverma I would recommend writing a unit test for the Python writer to be sure. IIRC we validate the schema, so we might need a modified equality function that ignores the large string difference.

houqp:
+1 to what @wjones127 said :) Thanks @chitralverma for taking a stab at this.

chitralverma:
cool, I'll add some tests on python side

BMeyn:
@chitralverma Any updates about the python tests? I'm happy to support when you are to busy.

ritchie46:
@chitralverma any news? This would enable a delta-writer in polars as well. :+1:

chitralverma:
I will work on the tests that are pending for this and rebase the changes soon

alexwilcoxson-rel:
Should `FixedSizedBinary` from arrow be mapped to "binary" as well?

chitralverma:
> Should `FixedSizedBinary` from arrow be mapped to "binary" as well?

added

wjones127:
@chitralverma The schema isn't derived from the Parquet file but should be derived from the table schema. Delta Lake has [its own standard](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#schema-serialization-format) for describing table schemas and its doesn't differentiate these two types. If it helps, you could write a function for converting delta schemas to arrow schemas that takes an option to use large string and large binary.

This is the line where that converts back:

https://github.com/delta-io/delta-rs/blob/70a9d095bfa7bcffaecaff3dc9404282366848b2/rust/src/delta_arrow.rs#L91

chitralverma:
@wjones127 made the changes as per your suggestions.

Added an option of `as_large_types` which converts the fields of Delta Schema to pyarrow schema with large types.

chitralverma:
@wjones127 I have fixed the linting issues, and updated with the main branch. I also added support for uint* and float16 as well. Now only the following types remain, any ideas for these because I don't think we can map all of them?

            DataType::Null => {}
            DataType::Date64 => {}
            DataType::Time32(_) => {}
            DataType::Time64(_) => {}
            DataType::Duration(_) => {}
            DataType::Interval(_) => {}
            DataType::Union(_, _, _) => {}
            DataType::Dictionary(_, _) => {}
            DataType::RunEndEncoded(_, _) => {}

chitralverma:
> @wjones127 I have fixed the linting issues, and updated with the main branch. I also added support for uint* and float16 as well. Now only the following types remain, any ideas for these because I don't think we can map all of them?
> 
> ```
>         DataType::Null => {}
>         DataType::Time32(_) => {}
>         DataType::Time64(_) => {}
>         DataType::Duration(_) => {}
>         DataType::Interval(_) => {}
>         DataType::Union(_, _, _) => {}
>         DataType::Dictionary(_, _) => {}
>         DataType::RunEndEncoded(_, _) => {}
> ```

From the above remaining types, here is a proposed strategy,

- Time32 => Int32
- Time64 => Int64
- Duration => Int64
- Interval => Int64
- Union | Dictionary | RunEndEncoded => Error as Not Supported
- Null => String or Error as Not Supported

Any suggestions on this ?

stinodego:
Personally, I'd suggest "Error as not supported" for all of those. That requires the user to explicitly acknowledge that they are writing unsupported types, and cast to supported types themselves.

python/src/schema.rs
```diff
@@ -1008,14 +1008,69 @@ impl PySchema {
...
+                ArrowDataType::Utf8 => field.with_data_type(ArrowDataType::LargeUtf8),
+
+                ArrowDataType::Binary | ArrowDataType::FixedSizeBinary(_) => {
+                    field.with_data_type(ArrowDataType::LargeBinary)
+                }
+
+                ArrowDataType::List(f) | ArrowDataType::FixedSizeList(f, _) => {
+                    let sub_field = convert_to_large_type(*f.clone(), f.data_type().clone());
+                    field.with_data_type(ArrowDataType::LargeList(Box::from(sub_field)))
+                }
```
wjones127, line 1032: Why convert fixed size list to large list? It doesn't have any indices, so it doesn't have a small / large variant. Is this just because Polars doesn't support FSL?

python/src/schema.rs
```diff
@@ -1008,14 +1008,69 @@ impl PySchema {
     /// Return equivalent PyArrow schema
     ///
     /// :rtype: pyarrow.Schema
-    #[pyo3(text_signature = "($self)")]
-    fn to_pyarrow(self_: PyRef<'_, Self>) -> PyResult<PyArrowType<ArrowSchema>> {
+    #[pyo3(text_signature = "($self, as_large_types)")]
```
wjones127, line 1011: I think this should handle the default value:
```suggestion
    #[pyo3(signature = "(as_large_types=false)")]
```

python/src/schema.rs
```diff
@@ -1008,14 +1008,69 @@ impl PySchema {
     /// Return equivalent PyArrow schema
     ///
```
wjones127, line 1009:
```suggestion
    ///
    /// :param as_large_types: get schema with all variable size types (list,
    ///   binary, string) as large variants (with int64 indices). This is for
    ///   compatibility with systems like Polars that only support the large 
    ///   versions of Arrow types.
    ///
```

python/src/schema.rs
```diff
@@ -1008,14 +1008,69 @@ impl PySchema {
...
     /// :rtype: pyarrow.Schema
-    #[pyo3(text_signature = "($self)")]
-    fn to_pyarrow(self_: PyRef<'_, Self>) -> PyResult<PyArrowType<ArrowSchema>> {
+    #[pyo3(text_signature = "($self, as_large_types)")]
+    fn to_pyarrow(
+        self_: PyRef<'_, Self>,
+        as_large_types: Option<bool>,
```
wjones127, line 1014:
```suggestion
    fn to_pyarrow(
        self_: PyRef<'_, Self>,
        as_large_types: bool,
```

python/src/schema.rs
wjones127: It doesn't seem like this is necessary to make the PR work, so I'd propose we change this to keep FSB and FSL as the same type. We still should transform the inner type of FSL like we do other nested types.

wjones127 (changes requested):
Thanks for sticking with this @chitralverma 

I have a few changes suggested. Once addressed I think this is good to merge.

python/src/schema.rs
chitralverma: yes you are correct, this was mostly for compatibility, but we can also do the cast or error handling for unsupported types in Polars directly.

Made the requested changes.

python/src/schema.rs
chitralverma: Made the requested changes.

chitralverma:
> Thanks for sticking with this @chitralverma
> 
> I have a few changes suggested. Once addressed I think this is good to merge.

@wjones127 Thanks for the review comments, I have made the requested changes.

chitralverma:
some tests timeout

wjones127:
Failures are unrelated. GitHub is being rather weird today.

wjones127 (approved):

chitralverma:
Thanks @wjones127

Issue #688, opened 2022-07-13
Clean up CRC and index files in vacuum commands


roeap:
# Description

In the current implementation of vacuum, we do not clean up the associated `.crc` files as well as the possibly existing bloom index files.

**Use Case**

**Related Issue(s)**

rtyler:
We do support lite and full vacuum now, closing 😄

PR #1045, opened 2022-12-31
refactor: move vacuum command to operations module

roeap:
# Description

Moving the `vacuum` operation into the operations module and adopting `IntoFuture` for the command builder. This is breaking the APIs for the builder (now with consistent setter names) but we are able to keep the APIs for `DeltaTable` in rust and python. 

In a follow up I would like to move th optimize command as well, This however may require refactoring the `PartitionValue` since we can only deal with `static` lifetimes when using `IntoFuture`, A while back we talked about pulling in `ScalarValue` from datafusion to optimize that implementation and maybe that's a good opportunitiy to look into that as well.

# Related Issue(s)

# Documentation


houqp:
I do agree that the builder interface makes the style more consistent and the argument passing less error prone :+1: 

That said, I think having an extra thin wrapper API on the table struct like below will provide a better developer UX:

```rust
        let result = table.vacuum()
            .retention_period(Duration::hours(169))
            .dry_run(true)
            .await
            .unwrap();
```

roeap:
>  wrapper API on the table struct like below will provide a better developer UX

Absoluetely, even more so on hte python side of things. This unfortuantely lead me down a little bit of a rabbit hole :D. Essentially again the quest on how to most efficiently share the log for commands etc. Right now in the code we just clone that state, but to my understanding this is not the way to go ...

Currently I am exploring a few options without a clear winner yet. However I do want to avoid making the `DeltaTableState` mutable when being shared, since this to me defeats the point of a snapshot, which is what we need for conflict resolution etc. Also looking forward to @wjones127's work in #1033 getting merged, since this should give us a nice baseline to finaly move to an arrow based state :). I do hope to make some general progress on this in this PR though to prepare for adopting this.

rust/src/delta.rs
```diff
@@ -1123,6 +1059,26 @@ impl DeltaTable {
...
+        let state = std::mem::take(&mut self.state);
+        let mut plan = VacuumBuilder::new(self.object_store(), state)
+            .with_dry_run(dry_run)
+            .with_enforce_retention_duration(enforce_retention_duration);
+        if let Some(hours) = retention_hours {
+            plan = plan.with_retention_period(Duration::hours(hours as i64));
+        }
+
+        let (table, metrics) = plan.await?;
+        self.state = table.state;
```
roeap, line 1078: Here I am a little bit concernbed about the error handling. If we fail duing the execution, we are still left with a DeltaTable with an empty state. If users wanted to recover from a failure and keep using the table, they would probably see unexpexted results.

Once we manage the state in arrow, we can likely have near zero-cost clones of the state and mitigate the issue. For one a way to handle this could be to handle the error internally, and load the previous table state again. But not sure what users wold expect / want to see.

The same applies on the python side.

rust/src/table_state.rs
```diff
@@ -330,6 +345,39 @@ impl DeltaTableState {
 
         Ok(())
     }
+
+    /// Obtain Add actions for files that match the filter
+    pub fn get_active_add_actions_by_partitions<'a>(
```
roeap, line 350: Moving this function to the state was part of an experiment. I did not move it back, since it exclusiveliy works on the state and from what I can see we are going to need functionality as such as we are evolving operations capabilities.

rust/src/delta.rs
wjones127: That would make the mutation operations like `merge()` and `process_action()` less efficient.

rust/src/delta.rs
```diff
@@ -512,6 +513,21 @@ impl DeltaTable {
         }
     }
 
+    /// Create a new [`DeltaTable`] from a [`DeltaTableState`] without loading any
+    /// data from backing storage.
+    ///
+    /// NOTE: This is for advanced users. If you don't know why you need to use this method,
+    /// please call one of the `open_table` helper methods instead.
+    pub fn new_with_state(storage: ObjectStoreRef, state: DeltaTableState) -> Self {
```
wjones127, line 521: Could we make this `pub crate` too?
```suggestion
    pub crate fn new_with_state(storage: ObjectStoreRef, state: DeltaTableState) -> Self {
```

rust/src/delta.rs
wjones127: Can we clone the state for now, and as a follow up let's figure out how to do this more efficiently?

wjones127 (changes requested):
This is a nice improvement.

> However I do want to avoid making the DeltaTableState mutable when being shared

I'm thinking maybe the builders need to take `&mut DeltaTable` or `&mut DeltaTableState`. Then the borrow checker will enforce this. `DeltaTableState` is clone-able, so that means users can always clone themselves if they don't want to deal with the lifetimes.

rust/src/operations/vacuum.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    /// Check if the specified retention period is less than the table's minimum
+    pub fn with_enforce_retention_duration(mut self, enforce: bool) -> Self {
+        self.enforce_retention_duration = enforce;
+        self
+    }
+
+    /// add a time source for testing
+    pub fn with_clock(mut self, clock: Arc<dyn Clock>) -> Self {
```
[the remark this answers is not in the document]
roeap: I don't really see a need for users to also test with this, however not sure how to elgantly exclude this, since `pub(crate)` will not be available to tests in the `tests/` folder. WE could re-use the `integration-test` feature, which has a similar purpose.

rust/src/operations/vacuum.rs
```diff
@@ -0,0 +1,325 @@
...
+    let path_name = path.to_string();
+    Ok((path_name.starts_with('.') || path_name.starts_with('_'))
+        && !path_name.starts_with("_delta_index")
+        && !path_name.starts_with("_change_data")
+        && !snapshot
+            .current_metadata()
+            .ok_or(DeltaTableError::NoMetadata)?
+            .partition_columns
+            .iter()
+            .any(|partition_column| path_name.starts_with(partition_column)))
```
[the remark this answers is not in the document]
roeap: Just deleting the line unfortunately led to failing tests, but I did not dive deeper. My general feeling is we have to re-visit the logic anyhow, since we are not yet cleaning up "associated" files. I can look further into this here, but would prefer in a follow up :) - maybe when doing #688.

rust/src/operations/vacuum.rs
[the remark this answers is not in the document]
wjones127: Thats fair. Let's just hide from docs then:

```suggestion
    #[doc(hidden)]
    pub fn with_clock(mut self, clock: Arc<dyn Clock>) -> Self {
```

rust/src/operations/vacuum.rs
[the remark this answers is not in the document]
wjones127: Okay. Thanks for looking into it.

wjones127 (approved):
LGTM. Though if you want, hiding that one function would be good.

roeap:
@wjones127 - applied your suggestions! Could you re-approve? :)

wjones127 (approved):

Issue #1017, opened 2022-12-14
Transaction log JSON formatting issue when writing data via Python bindings


tobyych:
hi there, i am using deltalake 0.6.4 python bindings to try writing delta tables to a S3-compatible storage.

i noticed an issue that the transaction log generated is not entirely the same as the example in <https://github.com/delta-io/delta/blob/master/PROTOCOL.md> (refer to the image attached, left one is the log generated by deltalake package, right one is the example on GitHub)

![image](https://user-images.githubusercontent.com/44737479/207498757-b4b02836-481c-4dc4-b609-bb241e9d2090.png)

this caused the data query engine Trino (version: 397)
 not able to read the delta log properly.

the relevant part of the error message is:
Caused by: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.io.IOException: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.lang.String` from Array value (token `JsonToken.START_ARRAY`)
 at [Source: (String)"{"commitInfo":{"clientVersion":"delta-rs.0.5.0","operation":"delta-rs.Write","operationParameters":{"mode":"Append","partitionBy":[],"predicate":null},"timestamp":1670980236192}}"; line: 1, column: 131] (through reference chain: io.trino.plugin.deltalake.transactionlog.DeltaLakeTransactionLogEntry["commitInfo"]-&gt;io.trino.plugin.deltalake.transactionlog.CommitInfoEntry["operationParameters"]-&gt;java.util.LinkedHashMap["partitionBy"])

may i know if there is any method to work around the issue?

[Slack Message](https://delta-users.slack.com/archives/CJ70UCSHM/p1670982230660309?thread_ts=1670982230.660309&cid=CJ70UCSHM)

rtyler:
As I mentioned in Slack, I think it's not that big of a deal for us to add some quotes around the `partitionBy`. I think the Trinio plugin also has a bug here because [commitInfo is optional](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#commit-provenance-information) in the protocol and therefore I don't believe it should be considered to have a required and agreed upon schema

sfilimonov-exos:
`delta-spark` also fails to read such a table with unquoted `partitionBy`.

Code to reproduce:

```
import pandas as pd
from delta import DeltaTable
from delta import configure_spark_with_delta_pip
from deltalake.writer import write_deltalake
[...]

PR #1054, opened 2023-01-08
Save operational params in the same way with delta io

ismoshkov:
# Description
Currently writing "operationParameters" in commit info is misaligned with delta io connector. 

[Here](https://github.com/delta-io/delta/blob/36a7edb8cf507e713700ba827c5fb5ad32b9163e/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L695) the sample of structure which is used in delta io. 

So the goal of this PR is to align with delta io approach and the PR do two thins: convert all values to string and delete keys with null values. 

# Related Issue(s)
Closes [issue #1017](https://github.com/delta-io/delta-rs/issues/1017)

### DIFF
--- a/rust/tests/command_optimize.rs
+++ b/rust/tests/command_optimize.rs
@@ -493,7 +493,7 @@ async fn test_commit_info() -> Result<(), Box<dyn Error>> {
     assert_eq!(last_commit["readVersion"], json!(version));
     assert_eq!(
         last_commit["operationParameters"]["targetSize"],
-        json!(2_000_000)
+        json!("2000000")
     );
     // TODO: Requires a string representation for PartitionFilter
     assert_eq!(last_commit["operationParameters"]["predicate"], Value::Null);
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -510,9 +510,11 @@ impl DeltaOperation {
 
         if let Ok(serde_json::Value::Object(map)) = serde_json::to_value(self) {
             let all_operation_fields = map.values().next().unwrap().as_object().unwrap();
-            let converted_operation_fields: Map<String, Value> = all_operation_fields.iter()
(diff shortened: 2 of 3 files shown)


rtyler (dismissed):

rust/tests/command_optimize.rs
```diff
@@ -493,7 +493,7 @@ async fn test_commit_info() -> Result<(), Box<dyn Error>> {
     assert_eq!(last_commit["readVersion"], json!(version));
     assert_eq!(
         last_commit["operationParameters"]["targetSize"],
-        json!(2_000_000)
+        json!("2000000")
```
wjones127, line 496: I'd also like to see a test for `partitionValues`. Could you add that?

imoshkov-exos: Sorry, could you please clarify your request? I ask because "partitionValues" is not part of the commit info and operating params that have been changed in this PR. 
Perhaps you want to check that "partitionBy" as part of the operating params is built correctly?
 
If you want exactly "partitionValues", I can figure out how to check that. But it looks like it should be in a separate PR.

ismoshkov: Sorry, could you please clarify your request? I ask because "partitionValues" is not part of the commit info and operating params that have been changed in this PR.
Perhaps you want to check that "partitionBy" as part of the operating params is built correctly?

If you want exactly "partitionValues", I can figure out how to check that. But it looks like it should be in a separate PR.

wjones127: Yes, sorry. I meant "partitionBy".

ismoshkov: Added separated unit test, because the most of commit info sections doesn't pass operation info. So they doesn't contains operational parameters which I can use to check.

houqp (approved):

wjones127 (approved):
Thanks @ismoshkov!

PR #1058, opened 2023-01-09
Expose checkpoint creation for current table state in python

ismoshkov:
# Description
Current python wrapper hasn't any functionlity to create checkpoints.
This PR exposes rust functionality which is creates checkpoint at current table version.

# Documentation
Sample of usage:
```Python
    delta_table = DeltaTable(some_path)
    # apply actions...
    delta_table.create_checkpoint()
```

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -426,7 +426,7 @@ given filters.
         """
         self._table.update_incremental()
 
-    def create_checkpoint(self):
+    def create_checkpoint(self) -> None:
         self._table.create_checkpoint()
 
     def __stringify_partition_values(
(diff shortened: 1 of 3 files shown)


roeap:
Thanks @ismoshkov for the contribution! One more "make format" and we are good to go :).

python/deltalake/table.py
```diff
@@ -426,6 +426,9 @@ def update_incremental(self) -> None:
         """
         self._table.update_incremental()
 
+    def create_checkpoint(self):
```
wjones127, line 429:
```suggestion
    def create_checkpoint(self) -> None:
```

wjones127 (approved):
Thanks @ismoshkov!

ismoshkov:
@wjones127 Thanks for your help!

PR #1085, opened 2023-01-21
Add an example of writing to a delta table with a RecordBatch

rtyler:


johnbatty:
This looks awesome.

I'm new to delta-rs and have been frustrated by lack of examples demonstrating how to write to tables from Rust. This is exactly what I wanted! 🚀

rust/examples/recordbatch-writer.rs
```diff
@@ -0,0 +1,248 @@
...
+                    .expect("Could not convert table path to a str"),
+            )
+            .await?
+        }
+        false => {
+            /* The table directory has not been initialized as a Delta table */
+            info!("It doesn't look like our delta table has been created");
+            create_initialized_table(&table_path).await
+        }
+    };
```
wjones127, line 51: I feel like we ought to have a better idiom for checking if a table exists, that doesn't involve introspecting the directory structure. Could we instead suggest checking for a `DeltaTableError::NotATable`? If that doesn't work, we should create an issue to improve this UX.

```suggestion
    let table = deltalake::open_table(table_path).await?;
    let mut table = {
        Ok(table) => table,
        Err(DeltaTableError::NotAtable(_)) => {
            info!("It doesn't look like our delta table has been created");
            create_initialized_table(&table_path).await
        },
        Err(err) => {
            err.unwra
        }
    }
```

rust/examples/recordbatch-writer.rs
```diff
@@ -0,0 +1,248 @@
...
+use chrono::prelude::*;
+use deltalake::action::*;
+use deltalake::arrow::array::*;
+use deltalake::arrow::record_batch::RecordBatch;
+use deltalake::writer::{DeltaWriter, RecordBatchWriter};
+use deltalake::*;
+use log::*;
+
+use std::collections::HashMap;
+use std::path::Path;
```
wjones127, line 19: Object store paths may work better with our APIs. For example, they implement `AsRef<str>`, so we don't need to call `.to_str().expect()` like you do below.

```suggestion
use object_store::path::Path;
```

wjones127 (changes requested):
Thanks for adding this! It's good to see examples to understand the state of our APIs.

rust/examples/recordbatch-writer.rs
roeap: The internal object store has a is_delta_location method. Usually it should do something similar, but due to an issue with windows we actually list, which is nit the best. Nevertheless, we could expose that on the table and improve internally.

rtyler:
@wjones127 I've committed the suggestions, anything else?

wjones127:
> @wjones127 I've committed the suggestions, anything else?

I'm afraid my suggestions weren't complete in themselves, so some modifications are needed to get it to compile. I can provide more suggestions if needed.

rust/examples/recordbatch-writer.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    let table = deltalake::open_table(table_path).await?;
+    let mut table = {
+        Ok(table) => table,
+        Err(DeltaTableError::NotAtable(_)) => {
+            info!("It doesn't look like our delta table has been created");
+            create_initialized_table(&table_path).await
+        },
+        Err(err) => {
+            err.unwra
```
ismoshkov, line 43: typo?

wjones127 (approved):

roeap (approved):

wjones127:
Thanks @rtyler !

PR #1097, opened 2023-01-25
Make rustls default across all packages

### LINKED ISSUE #985: `cargo test` does not run due to tls conflict
# Environment

**Delta-rs version**:
Commit 4efc67c718121dd3b5512a772e4036b640cc62ff
**Binding**: 

**Environment**:
- **Cloud provider**:
- **OS**: Ubuntu 22.04
- **Other**:

***
# Bug

**What happened**:
```
error[E0252]: the name `tls` is defined multiple times
  --> /home/iajoiner/.cargo/registry/src/github.com-1ecc6299db9ec823/rusoto_core-0.48.0/src/lib.rs:25:5
   |
23 | use hyper_rustls as tls;
   |     ------------------- previous import of the module `tls` here
24 | #[cfg(feature = "native-tls")]
25 | use hyper_tls as tls;
   |     ^^^^^^^^^^^^^^^^ `tls` reimported here
   |
   = note: `tls` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
   |
25 | use hyper_tls as other_tls;
   |     ~~~~~~~~~~~~~~~~~~~~~~

error[E0599]: no function or associated item named `new` found for struct `hyper_rustls::HttpsConnector` in the current scope
   --> /home/iajoiner/.cargo/registry/src/github.com-1ecc6299db9ec823/rusoto_core-0.48.0/src/request.rs:224:41
    |
224 |         let connector = HttpsConnector::new();
    |                                         ^^^ function or associated item not found in `hyper_rustls::HttpsConnector<_>`

error[E0599]: no function or associated item named `new` found for struct `hyper_rustls::HttpsConnector` in the current scope
   --> /home/iajoiner/.cargo/registry/src/github.com-1ecc6299db9ec823/rusoto_core-0.48.0/src/request.rs:244:41
    |
244 |         let connector = HttpsConnector::new();
    |                                         ^^^ function or associated item not found in `hyper_rustls::HttpsConnector<_>`

```

**How to reproduce it**:
Run `cargo test` at the top level

**More details**:
The essence of the problem is that we are using an unofficial AWS SDK crate, rusoto, that has been in maintenance mode since 2020 when there is in fact an official AWS SDK crate, https://github.com/awslabs/aws-sdk-rust now. Since rusoto has two mutually exclusive features without adding a compiler error to detect them being used in combination we have this bug.

wjones127:
# Description

Clean up our tls options so that we can run `cargo test` at the top level again.

Also fixed clippy warnings that had accumulated.

# Related Issue(s)

- closes #985

# Documentation


aws/delta-checkpoint/Cargo.toml
```diff
@@ -23,12 +23,17 @@ tokio-stream = { version = "0", features = ["fs"] }
...
 
 [dev-dependencies]
-rusoto_core = { version = "0.48", default-features = false, features = ["rustls"] }
+rusoto_core = { version = "0.48", default-features = false }
 rusoto_credential = "0.48"
-rusoto_s3 = { version = "0.48", default-features = false, features = ["rustls"] }
+rusoto_s3 = { version = "0.48", default-features = false }
+
+[features]
+default = ["native-tls"]
```
houqp, line 33: for the checkpoint lambda, we should probably set rustls as the default feature because it's hard to get openssl working in the lambda runtime.

delta-inspect/Cargo.toml
```diff
@@ -17,4 +17,4 @@ env_logger = "0"
 [dependencies.deltalake]
 path = "../rust"
 version = "0"
-features = ["s3", "azure", "glue", "gcs"]
+features = ["azure", "glue", "gcs"]
```
houqp, line 20: what's the reason for removing s3 from the feature set?

houqp:
Thank you @wjones127 for taking a stab at this!

aws/delta-checkpoint/Cargo.toml
wjones127: If we want to be able to run `cargo test` at the root, we need all packages to either default to `native-tls` or `rustls`. I can add a note in the readme that we recommend rustls for this package.

delta-inspect/Cargo.toml
wjones127: Whoops, that was an accident.

aws/delta-checkpoint/Cargo.toml
houqp: Ha I see. What if we default all to rustls? I think this might be what's preferred for our python releases as well?

wjones127: Or maybe I'll just make `rustls` default.

rust/src/operations/create.rs
```diff
@@ -333,8 +333,12 @@ mod tests {
     #[tokio::test]
     async fn test_create_local_relative_path() {
         let table_schema = get_delta_schema();
-        let name = Alphanumeric.sample_string(&mut rand::thread_rng(), 16);
-        let table = DeltaOps::try_from_uri(format!("./{}", name))
+        let tmp_dir = TempDir::new_in(".", "tmp_").unwrap();
```
wjones127, line 336: These tests were creating directories without cleaning them up.

glibc_version/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 //!
 //! # Example
 //!
-//! ```
+//! ```no_run
```
wjones127, line 7: Can't run this on MacOS, and wasn't sure how to opt out of doc test based on platform.

wjones127:
Well I've figured out most of the issues, but the Windows build I haven't figured out 😢

houqp:
@wjones127 have you tried kicking off the windows build without loading the previous build cache?

delta-inspect/Cargo.toml
```diff
@@ -17,4 +17,10 @@ env_logger = "0"
 [dependencies.deltalake]
 path = "../rust"
 version = "0"
-features = ["s3", "azure", "glue", "gcs"]
+features = ["azure", "gcs"]
```
houqp, line 20: I think we should still keep the s3 and glue feature flag here?

python/Cargo.toml
```diff
@@ -38,4 +39,9 @@ features = ["extension-module", "abi3", "abi3-py37"]
 [dependencies.deltalake]
 path = "../rust"
 version = "0"
-features = ["s3", "azure", "glue", "gcs", "python", "datafusion"]
+features = ["azure", "gcs", "python", "datafusion"]
```
houqp, line 42: s3 and glue got removed here, expected?

delta-inspect/Cargo.toml
roeap: I stumbled across this as well, I think they are now enabled via the section right below this, but a bit more fine granular...

wjones127:
@houqp Thanks, clearing the cache worked!

rust/src/storage/config.rs
```diff
@@ -230,16 +216,10 @@ impl ObjectStoreKind {
             }),
             #[cfg(feature = "gcs")]
             ObjectStoreKind::Google => {
-                let store = GoogleCloudStorageBuilder::new()
+                let store = GoogleCloudStorageBuilder::from_env()
```
wjones127, line 219: The integration tests were failing for GCP without this change. In the tests, we configure GCP through the `GOOGLE_SERVICE_ACCOUNT` environment variable, but instead this code was creating a GCS storage with default configuration.

Is there a reason we shouldn't be picking up the environment variables first? I assume that the storage options override them. @roeap

roeap: That must have been a bug, or at least an inconsistency. I think for S3 and Azure we use `from_env` as well. That said., this is actually a point I struggled with a bit when writing this. Mainly for S3 and Azure where we have many different ways of authorising. One scenario in particular was troubling me. In case the environment contains enough information for the store to find a credential, and the storage options contain another complete set of config for a credential. In this case object_store will pick the credential based on the order they are checked in the builder, which does not necessarily give precedence to the one in the options, which I assume is what the user would want.

The initial approach was to try and create one only from the storage_options, and if we fail, load from environment and from options. TO be honest, I do not remember what the specific issue was why I abandoned this approach, but I do remember, that the special handling of `allow_http` is definitely related. In the end I guess, I though in scenarios where tables in multiple locations are managed people should be careful what to put in the env...

That said, anything in storage options will overwrite any info parsed from the environment.

rust/src/operations/create.rs
roeap: Thanks for cleaning that up!

roeap (approved):
Great work - will be nice running cargo from the root again 😆 .

PR #1101, opened 2023-01-25
Selectively overwrite data with python

ismoshkov:
# Description
Currently high-level python writer isn't support partial partition overwrite. 
This PR enable usage of partitions filtering for writing data

The functionlity is similar to:
https://docs.databricks.com/delta/selective-overwrite.html

The logic checks that data should contains only partitions that passing filtering.  

# Documentation
```python
    write_deltalake(
        delta_path,
        sample_data,
        mode="overwrite",
        partitions_filters=[("partition_a", ">", "1")],
    )
```

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -85,6 +85,7 @@ def write_deltalake(
     configuration: Optional[Mapping[str, Optional[str]]] = None,
     overwrite_schema: bool = False,
     storage_options: Optional[Dict[str, str]] = None,
+    partitions_filters: Optional[List[Tuple[str, str, Any]]] = None,
 ) -> None:
     """Write to a Delta Lake table (Experimental)
 
@@ -132,6 +133,7 @@ def write_deltalake(
     :param configuration: A map containing configuration options for the metadata action.
     :param overwrite_schema: If True, allows updating the schema of the table.
     :param storage_options: options passed to the native delta filesystem. Unused if 'filesystem' is defined.
+    :param partitions_filters: the partition filters that will be used for partition overwrite.
     """
     if _has_pandas and isinstance(data, pd.DataFrame):
         if schema is not None:
@@ -179,6 +181,8 @@ def write_deltalake(
 
         if partition_by:
             assert partition_by == table.metadata().partition_columns
+        else:
+            partition_by = table.metadata().partition_columns
 
         if table.protocol().min_writer_version > MAX_SUPPORTED_WRITER_VERSION:
             raise DeltaTableProtocolError(
@@ -224,8 +228,27 @@ def write_deltalake(
         invariants = table.schema().invariants
         checker = _DeltaDataChecker(invariants)
 
+        def check_data_is_aligned_with_partition_filtering(
+            batch: pa.RecordBatch,
+        ) -> None:
+            if table is None:
+                return
+            allowed_partitions = table._table.get_active_partitions(partitions_filters)
+            for column_index, column_name in enumerate(batch.schema.names):
+                if column_name in table.metadata().partition_columns:
+                    for value in batch.column(column_index).unique():
+                        if (column_name, str(value)) not in allowed_partitions:
+                            raise ValueError(
+                                f"Data should be aligned with partitioning. "
+                                f"Partition '{column_name}'='{value}' should be filtered out from data."
+                            )
+
         def validate_batch(batch: pa.RecordBatch) -> pa.RecordBatch:
             checker.check_batch(batch)
+
+            if mode == "overwrite" and partitions_filters:
+                check_data_is_aligned_with_partition_filtering(batch)
+
             return batch
 
         if isinstance(data, RecordBatchReader):
@@ -277,6 +300,7 @@ def write_deltalake(
             mode,
             partition_by or [],
             schema,
+            partitions_filters,
         )
 
 
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -240,15 +240,19 @@ def write_deltalake(
 
             if table is None:
                 return
+            existed_partitions = table._table.get_active_partitions()
             allowed_partitions = table._table.get_active_partitions(partitions_filters)
             for column_index, column_name in enumerate(batch.schema.names):
                 if column_name in table.metadata().partition_columns:
                     for value in batch.column(column_index).unique():
-                        print(type(value))
-                        if (
+                        partition = (
                             column_name,
                             arrow_value_to_partition_string(value),
-                        ) not in allowed_partitions:
+                        )
+                        if (
+                            partition not in allowed_partitions
+                            and partition in existed_partitions
+                        ):
                             raise ValueError(
                                 f"Data should be aligned with partitioning. "
                                 f"Partition '{column_name}'='{value}' should be filtered out from data."
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -32,7 +32,7 @@ else:
 import pyarrow as pa
 import pyarrow.dataset as ds
 import pyarrow.fs as pa_fs
-from pyarrow.lib import RecordBatchReader
+from pyarrow.lib import BooleanScalar, RecordBatchReader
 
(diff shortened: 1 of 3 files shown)


python/tests/test_writer.py
```diff
@@ -496,3 +496,90 @@ def test_writer_with_options(tmp_path: pathlib.Path):
     )
 
     assert table == data
+
+
+def test_partition_overwrite(tmp_path: pathlib.Path):
```
wjones127, line 501: Could we parameterize this test by the partition column datatype? At very least, we should test `date32`, `timestamp`, and `bool`. I just want to verify that the data check function you wrote can handle those different types.

wjones127 (changes requested):
Thank you for writing this! I'd like a few more data types tested, and then this should be good to go. 😄

python/tests/test_writer.py
wjones127: You are correct on timestamp. I forgot about that. Created #1111 to track that.

python/deltalake/writer.py
```diff
@@ -235,8 +239,38 @@ def visitor(written_file: Any) -> None:
...
+                if column_name in table.metadata().partition_columns:
+                    for value in batch.column(column_index).unique():
+                        print(type(value))
+                        if (
+                            column_name,
+                            arrow_value_to_partition_string(value),
+                        ) not in allowed_partitions:
+                            raise ValueError(
+                                f"Data should be aligned with partitioning. "
+                                f"Partition '{column_name}'='{value}' should be filtered out from data."
```
wjones127, line 265: Do we allow writing to a *new* partition with this method? It seems like the partition must already exist. We might want to either allow that or provide a more helpful error message if `allowed_partitions` comes back empty.

python/tests/test_writer.py
```diff
@@ -496,3 +496,218 @@ def test_writer_with_options(tmp_path: pathlib.Path):
...
+                        ],
+                        pa.date32(),
+                    ),
+                    "val": pa.array([2, 3, 1, 3], pa.int64()),
+                }
+            ),
+        },
+    ],
+)
+def test_partition_overwrite(tmp_path: pathlib.Path, cases_data: Dict[str, pa.Array]):
```
wjones127, line 647: The type for cases_data doesn't seem correct, since there is also tables and filters in there.

python/tests/test_writer.py
```diff
@@ -496,3 +496,218 @@ def test_writer_with_options(tmp_path: pathlib.Path):
...
+    delta_table.update_incremental()
+    assert (
+        delta_table.to_pyarrow_table().sort_by(
+            [("p1", "ascending"), ("p2", "ascending")]
+        )
+        == expected_data
+    )
+
+
+def test_partition_overwrite_unfiltered_data_fails(tmp_path: pathlib.Path):
```
wjones127, line 696: Could we add a test that validates we get an error if you pass a partition filter that contains columns not in the partition columns?

python/tests/test_writer.py
```diff
@@ -496,3 +496,218 @@ def test_writer_with_options(tmp_path: pathlib.Path):
     )
 
     assert table == data
+
+
+@pytest.mark.parametrize(
+    "cases_data",
+    [
```
wjones127, line 503: Sorry this ended up quite long!

What if instead of varying the data, you used the same table for all of them but varied which column in the table was the partition column. See, [this test as an example](https://github.com/delta-io/delta-rs/blob/d984ccfdb5ae3d8ddfc4b6611c6c9ad599541cc9/python/tests/test_writer.py#L153-L182).

Then maybe you can parametrize by 

```python
@pytest.mark.parametrize(
    "cases",
    [("int64", [("int64", "=", "1")], [("int64", ">", "1")]),
     ("bool", [("bool", "=", "true")], [("bool", "!=", "true")]),
     ...
    ]
)
```

If it ends up not being possible to make more succinct, that fine. But it seems worth a try 😄

python/tests/test_writer.py
wjones127: Also what happens if you pass them to a non-partitioned table?

wjones127 (changes requested):
Thank you for testing the additional types. I had a few ideas for some quick additional tests, if you could add those too. Then I think this should be ready to go.

python/deltalake/writer.py
ismoshkov: Now it allows create new partition and overwrites existing ones. Not sure if we want to have it configurable.

python/tests/test_writer.py
ismoshkov: Added test cases

python/tests/test_writer.py
ismoshkov: Updates

python/deltalake/writer.py
```diff
@@ -235,8 +239,38 @@ def visitor(written_file: Any) -> None:
...
+                return
+            allowed_partitions = table._table.get_active_partitions(partitions_filters)
+            for column_index, column_name in enumerate(batch.schema.names):
+                if column_name in table.metadata().partition_columns:
+                    for value in batch.column(column_index).unique():
+                        print(type(value))
+                        if (
+                            column_name,
+                            arrow_value_to_partition_string(value),
+                        ) not in allowed_partitions:
```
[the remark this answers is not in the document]
ismoshkov: Sorry, couldn't understand how it should affect current code. Could you clarify it more?

python/tests/test_writer.py
```diff
@@ -532,3 +532,197 @@ def test_try_get_table_and_table_uri(tmp_path: pathlib.Path):
...
+        (False, True, pa.bool_(), "false"),
+        (date(2022, 1, 1), date(2022, 1, 2), pa.date32(), "2022-01-01"),
+    ],
+)
+def test_partition_overwrite(
+    tmp_path: pathlib.Path,
+    value_1: Any,
+    value_2: Any,
+    value_type: Any,
+    filter_string: Any,
```
wjones127, line 550:
```suggestion
    value_type: pa.DataType,
    filter_string: str,
```

wjones127 (changes requested):
Thank you for refactoring the tests and adding more. They look good.

I think we can use the existing DeltaJSONEncoder and then fix some other small issues and we are good-to-go.

Could you make sure to rebase and re-run `make format`. We just changed our linter to `ruff` so there may be some new rules enabled :)

python/deltalake/writer.py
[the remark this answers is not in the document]
wjones127: Sorry that was unclear. I realized we had actually already implemented the encoder, so I've left concrete suggestions in my latest review.

python/tests/test_writer.py
```diff
@@ -532,3 +532,197 @@ def test_try_get_table_and_table_uri(tmp_path: pathlib.Path):
...
+    )
+    expected_data = pa.table(
+        {
+            "p1": pa.array(["1", "1", "2", "2"], pa.string()),
+            "p2": pa.array([value_1, value_2, value_1, value_2], value_type),
+            "val": pa.array([2, 2, 1, 1], pa.int64()),
+        }
+    )
+    write_deltalake(
+        str(tmp_path),
```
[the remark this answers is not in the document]
ismoshkov: done

python/deltalake/writer.py
```diff
@@ -226,8 +230,42 @@ def visitor(written_file: Any) -> None:
...
+            if table is None:
+                return
+            existed_partitions = table._table.get_active_partitions()
+            allowed_partitions = table._table.get_active_partitions(partitions_filters)
+            for column_index, column_name in enumerate(batch.schema.names):
+                if column_name in table.metadata().partition_columns:
+                    for value in batch.column(column_index).unique():
+                        partition = (
+                            column_name,
+                            arrow_value_to_partition_string(value),
```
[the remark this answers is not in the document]
ismoshkov: done

wjones127 (approved):
Excellent! I'm excited for when we release this :) 
Thanks @ismoshkov!

PR #1102, opened 2023-01-25
chore: remove unmaintained ruby bindings

### LINKED ISSUE #1099: Possibly remove the Ruby bindings
From what I understand, the Ruby bindings are unusable and aren't going to be supported anytime soon.

It's possible that marketing this project as "A native Rust library for Delta Lake, with bindings into Python and Ruby" is a bit misleading.

Should we just remove the Ruby bindings for now?

roeap:
# Description

removes ruby bindings

cc @rtyler 

# Related Issue(s)

closes #1099

# Documentation


wjones127 (approved):
I'm all for removing components that aren't maintained.

PR #1103, opened 2023-01-26
Implement filesystem check

### LINKED ISSUE #1092: Support FSCK REPAIR TABLE Operation
# Description
Databricks  provides an operation called [FSCK REPAIR TABLE](https://docs.databricks.com/sql/language-manual/delta-fsck.html) that removes active files that no longer can be found in the underlying file system.

**Use Case**
Due to a hardware issue some parquet files were corrupted when written. This data is non-critical and I simply would like to delete to from the underlying storage and then use this operation to reconcile the log.

This operation also supports a dry run which can be used to check if files are missing due to an external issue.

**Related Issue(s)**

Blajda:
# Description
Implementation of the filesystem check operation.

The implementation is fairly straight forward with a HEAD call being made for each active file to check if it exists.
A remove action is then made for each file that is orphaned.

An alternative solution is instead to maintain a hashset with all active files and then recursively list all files. If the file exists then remove from the set. All remaining files in the set are then considered orphaned.
 
Looking for feedback and if the second approach is preferred I can make the changes

# Related Issue(s)
- closes #1092

### DIFF
--- a/rust/src/operations/filesystem_check.rs
+++ b/rust/src/operations/filesystem_check.rs
@@ -87,7 +87,7 @@ impl FileSystemCheckBuilder {
                     &active.path
                 ))
             })?;
-            if url.scheme().len() == 0 {
+            if url.scheme().is_empty() {
                 files_relative.insert(active.path.as_str(), active);
             } else {
                 files_absolute.insert(active.path.as_str(), active);
--- a/rust/src/operations/filesystem_check.rs
+++ b/rust/src/operations/filesystem_check.rs
@@ -1,0 +1,167 @@
+//! Audit the Delta Table for active files that do not exist in the underlying filesystem and remove them
+//!
+//! # Example
+//! ```rust ignore
+//! let mut table = open_table("../path/to/table")?;
+//! let (table, metrics) = FileSystemCheckBuilder::new(table.object_store(). table.state).await?;
+//! ````
+use crate::action::{Action, Add, DeltaOperation, Remove};
+use crate::operations::transaction::commit;
+use crate::storage::DeltaObjectStore;
+use crate::table_state::DeltaTableState;
+use crate::DeltaDataTypeVersion;
+use crate::{DeltaDataTypeLong, DeltaResult, DeltaTable, DeltaTableError};
+use futures::future::BoxFuture;
+pub use object_store::path::Path;
+use object_store::Error as ObjectStoreError;
+use object_store::ObjectStore;
+use std::fmt::Debug;
+use std::sync::Arc;
+use std::time::SystemTime;
+use std::time::UNIX_EPOCH;
+
+/// Audit the Delta Table's active files with the underlying file system.
+/// See this module's documentaiton for more information
+#[derive(Debug)]
+pub struct FileSystemCheckBuilder {
+    /// A snapshot of the to-be-checked table's state
+    state: DeltaTableState,
+    /// Delta object store for handling data files
+    store: Arc<DeltaObjectStore>,
+    /// Don't remove actions to the table log. Just determine which files can be removed
+    dry_run: bool,
+}
+
+/// Details of the FSCK operation including which files were removed from the log
+#[derive(Debug)]
+pub struct FileSystemCheckMetrics {
+    /// Was this a dry run
+    pub dry_run: bool,
+    /// Files that wrere removed successfully
+    pub files_removed: Vec<String>,
+}
+
+struct FileSystemCheckPlan {
+    /// Version of the snapshot provided
+    version: DeltaDataTypeVersion,
+    /// Delta object store for handling data files
+    store: Arc<DeltaObjectStore>,
+    /// Files that no longer exists in undlying ObjectStore but have active add actions
+    pub files_to_remove: Vec<Add>,
+}
+
+impl FileSystemCheckBuilder {
+    /// Create a new [`FileSystemCheckBuilder`]
+    pub fn new(store: Arc<DeltaObjectStore>, state: DeltaTableState) -> Self {
+        FileSystemCheckBuilder {
+            state,
+            store,
+            dry_run: false,
+        }
+    }
+
+    /// Only determine which add actions should be removed
+    pub fn with_dry_run(mut self, dry_run: bool) -> Self {
+        self.dry_run = dry_run;
+        self
+    }
+
+    async fn create_fsck_plan(&self) -> DeltaResult<FileSystemCheckPlan> {
+        let mut files_to_remove = Vec::new();
+        let version = self.state.version();
+        let store = self.store.clone();
+
+        for active in self.state.files() {
+            let res = self.store.head(&Path::from(active.path.as_str())).await;
+            if let Err(ObjectStoreError::NotFound { path: _, source: _ }) = res {
+                files_to_remove.push(active.to_owned());
+            } else {
+                res.map_err(DeltaTableError::from)?;
+            }
+        }
+
+        Ok(FileSystemCheckPlan {
+            files_to_remove,
+            version,
+            store,
+        })
+    }
+}
+
+impl FileSystemCheckPlan {
+    pub async fn execute(self) -> DeltaResult<FileSystemCheckMetrics> {
+        if self.files_to_remove.is_empty() {
+            return Ok(FileSystemCheckMetrics {
+                dry_run: false,
+                files_removed: Vec::new(),
+            });
+        }
+
+        let mut actions = Vec::new();
+        let mut removed_file_paths = Vec::new();
+        let version = self.version;
+        let store = &self.store;
+
+        for file in self.files_to_remove {
+            let deletion_time = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+            let deletion_time = deletion_time.as_millis() as DeltaDataTypeLong;
+            removed_file_paths.push(file.path.clone());
+            actions.push(Action::remove(Remove {
+                path: file.path,
+                deletion_timestamp: Some(deletion_time),
+                data_change: true,
+                extended_file_metadata: None,
+                partition_values: Some(file.partition_values),
+                size: Some(file.size),
+                tags: file.tags,
+            }));
+        }
+
+        if !actions.is_empty() {
+            commit(
+                store,
+                version + 1,
+                actions,
+                DeltaOperation::FileSystemCheck {},
+                None,
+            )
+            .await?;
+        }
+
+        Ok(FileSystemCheckMetrics {
+            dry_run: false,
+            files_removed: removed_file_paths,
+        })
+    }
+}
+
+impl std::future::IntoFuture for FileSystemCheckBuilder {
+    type Output = DeltaResult<(DeltaTable, FileSystemCheckMetrics)>;
+    type IntoFuture = BoxFuture<'static, Self::Output>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        let this = self;
+
+        Box::pin(async move {
+            let plan = this.create_fsck_plan().await?;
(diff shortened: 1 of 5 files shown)


houqp:
Instead of heading each individual file, wouldn't it be more efficient to issue an prefixed list call to gather all current files in the table? This way, we can reduce API call count by about 1000x.

Blajda:
@houqp I modified the process to instead use a list call but I'm not quite sure on how I would take advantage of prefixes in this case. For partitioned tables it could perform a list per partition path but I think non-partitioned tables would still require a full list on the root.

For non-partitioned tables would it be possible to perform a list without including the files in `_delta_log`?

houqp (dismissed):

houqp:
@Blajda the prefix filter can only be applied with the table root unfortunately, so it will always include the _detal_log folder unfortunately. The main benefit it provides is we can restrict the listing to the table we are running the operation on, not all other tables stored in the same bucket.

houqp:
@roeap @wjones127 want to take a final look?

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,174 @@
+//! Audit the Delta Table for active files that do not exist in the underlying filesystem and remove them
+//!
```
wjones127, line 2: A few additional details might be helpful.

```suggestion
//! Audit the Delta Table for active files that do not exist in the underlying filesystem and remove them.
//!
//! Active files are ones that have an add action in the log, but no corresponding remove action.
//! This operation creates a new transaction containing a remove action for each of the missing files.
//!
//! This can be used to repair tables where a data file has been deleted accidentally or 
//! purposefully, if the file was corrupted.
```

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub async fn execute(self) -> DeltaResult<FileSystemCheckMetrics> {
+        if self.files_to_remove.is_empty() {
+            return Ok(FileSystemCheckMetrics {
+                dry_run: false,
+                files_removed: Vec::new(),
+            });
+        }
+
+        let mut actions = Vec::new();
+        let mut removed_file_paths = Vec::new();
```
wjones127, line 108:
```suggestion
        let mut actions = Vec::with_capacity(self.files_to_remove.len());
        let mut removed_file_paths = Vec::with_capacity(self.files_to_remove.len());
```

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,174 @@
...
+                deletion_timestamp: Some(deletion_time),
+                data_change: true,
+                extended_file_metadata: None,
+                partition_values: Some(file.partition_values),
+                size: Some(file.size),
+                tags: file.tags,
+            }));
+        }
+
+        if !actions.is_empty() {
```
wjones127, line 127: This should never be false, right? Since we already return early above if `self.files_to_remove` is empty, and `actions` should have the same length as that, right?

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,174 @@
...
+        Box::pin(async move {
+            let plan = this.create_fsck_plan().await?;
+            if this.dry_run {
+                return Ok((
+                    DeltaTable::new_with_state(this.store, this.state),
+                    FileSystemCheckMetrics {
+                        files_removed: plan
+                            .files_to_remove
+                            .iter()
+                            .map(|f| f.path.clone())
```
wjones127, line 161: I think you can consume the plan with `into_iter()` and just take ownership of the plan:
```suggestion
                            .into_iter()
                            .map(|f| f.path)
```

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,174 @@
...
+    version: DeltaDataTypeVersion,
+    /// Delta object store for handling data files
+    store: Arc<DeltaObjectStore>,
+    /// Files that no longer exists in undlying ObjectStore but have active add actions
+    pub files_to_remove: Vec<Add>,
+}
+
+impl FileSystemCheckBuilder {
+    /// Create a new [`FileSystemCheckBuilder`]
+    pub fn new(store: Arc<DeltaObjectStore>, state: DeltaTableState) -> Self {
```
wjones127, line 56: Should we validate that the provided state is the latest version of the table? Since I don't think this is a valid operation for earlier states? or maybe that is an overall issue with the operations module?

rust/tests/command_filesystem_check.rs
```diff
@@ -0,0 +1,112 @@
+#![cfg(all(feature = "integration_test"))]
```
wjones127, line 1: Thanks for adding integration tests!

rust/tests/command_filesystem_check.rs
```diff
@@ -0,0 +1,112 @@
...
+async fn test_filesystem_check_local() -> TestResult {
+    Ok(test_filesystem_check(StorageIntegration::Local).await?)
+}
+
+#[cfg(any(feature = "s3", feature = "s3-rustls"))]
+#[tokio::test]
+#[serial]
+async fn test_filesystem_check_aws() -> TestResult {
+    set_env_if_not_set("AWS_S3_ALLOW_UNSAFE_RENAME", "true");
+    set_env_if_not_set("AWS_S3_LOCKING_PROVIDER", "none");
```
wjones127, line 23: Could be wrong, but I *think* you only need the first environment variable. If that's not the case, we might want to file an issue for that.
```suggestion
```

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut files_to_remove = Vec::new();
+        let mut files_to_check: HashMap<String, &Add> = HashMap::new();
+        let version = self.state.version();
+        let store = self.store.clone();
+
+        self.state.files().iter().for_each(|active| {
+            files_to_check.insert(active.path.to_owned(), active);
+        });
+
+        let mut files = self.store.list(None).await?;
```
wjones127, line 80: Delta Tables are allowed to have data files outside the table root. Currently, if that is the case, this operation will remove them all from the log. I see two options:

1. Call HEAD on each file outside the root to check if they exist.
2. Error if we detect any files outside the root.

wjones127 (changes requested):
Thank you for adding this! I've been thinking having an operation like this would let me sleep better at night :)

I think we should address the case where data files are outside the root, either explicitly not supporting or using HEAD for those ones. Also noticed a few minor issues.

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    async fn create_fsck_plan(&self) -> DeltaResult<FileSystemCheckPlan> {
+        let mut files_to_remove = Vec::new();
+        let mut files_to_check: HashMap<String, &Add> = HashMap::new();
+        let version = self.state.version();
+        let store = self.store.clone();
+
+        self.state.files().iter().for_each(|active| {
+            files_to_check.insert(active.path.to_owned(), active);
+        });
```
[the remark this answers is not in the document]
houqp: yeah, have you tried `HashMap<&str, &Add>`?

rust/src/operations/filesystem_check.rs
Blajda: I think dry run of this operation would be helpful for diagnosing issues that arise from a combination of vacuum + time travel. If executed on a older version I do expect it to fail during the commit operation.
I can add a test to demonstrate that.

rust/tests/command_filesystem_check.rs
Blajda: Yes it's required. I had a commit  with just allow unsafe set and it failed since http is not supported for the lock client. Currently the lock client is always constructed and is not determined by `UNSAFE_RENAME`

See: https://github.com/delta-io/delta-rs/blob/main/rust/src/storage/s3.rs#L402

rust/src/operations/filesystem_check.rs
Blajda: Makes sense. Do any examples of tables that use absolute paths? 
I just checked the protocol and noticed absolute paths are encoded as defined in [rfc2396](https://www.ietf.org/rfc/rfc2396.txt)

It states the following
```
An absolute URI contains the name of the scheme being used (<scheme>)
   followed by a colon (":") and then a string (the <scheme-specific-
   part>) whose interpretation depends on the scheme.
```

So I can segment paths based of it the scheme is set on the path and then use the correct api call for each.

wjones127: Oh that's a good point! Checking for the scheme at the beginning makes sense.

We don't have any example tables right now. We don't generally support them in most operations, but we should try to keep them in mind. Hard to have example tables without setting up a public S3 bucket or something.

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /// Only determine which add actions should be removed
+    pub fn with_dry_run(mut self, dry_run: bool) -> Self {
+        self.dry_run = dry_run;
+        self
+    }
+
+    async fn create_fsck_plan(&self) -> DeltaResult<FileSystemCheckPlan> {
+        let mut files_to_remove = Vec::new();
```
wjones127, line 76:
```suggestion
```

rust/src/operations/filesystem_check.rs
Blajda: I'm not comfortable with supporting absolute paths without having some integration tests. Currently the operation will fail when an absolute path is encountered.

rust/src/operations/filesystem_check.rs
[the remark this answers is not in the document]
Blajda: Yes I works. Updated to avoid the clones on the path.

Blajda:
@wjones127 I updated the PR to fail when an absolute path exists in the Delta log. Also added an additional test to demonstrate the operation will only work on the latest table version.

rust/src/operations/filesystem_check.rs
```diff
@@ -0,0 +1,221 @@
...
+fn is_absolute_path(path: &str) -> DeltaResult<bool> {
+    match Url::parse(path) {
+        Ok(_) => Ok(true),
+        Err(ParseError::RelativeUrlWithoutBase) => Ok(false),
+        Err(_) => Err(DeltaTableError::Generic(format!(
+            "Unable to parse path: {}",
+            &path
+        ))),
+    }
+}
```
roeap, line 70: I think this will not work for local paths. One pattern we apply in `object_store` and here is to try and cannonicalize the path and if it fails proceed with url parsing. e.g.

https://github.com/delta-io/delta-rs/blob/a60129bb48a24f94e8312c12940f6ec96685af05/rust/src/builder.rs#L355-L358

roeap (commented):
Thanks @Blajda  - one small question along the lines of hoping to eventually support absolute paths in the delta log. Otherwise looks great!

rust/src/operations/filesystem_check.rs
Blajda: I don't follow on why it would not work on local paths. Paths in the Delta log must follow [rfc2396](https://www.ietf.org/rfc/rfc2396.txt) which states absolute paths must have a scheme.
Do you have a particular example that would cause the unit test for this function to fail?

roeap: well  .. in that case I am wrong! If that paths are required to have a scheme, then url parsing should always work for valid paths ... The windows case is also covered in your tests, so all is well :).

roeap (approved):

wjones127 (approved):

PR #1106, opened 2023-01-27
chore: clippy

roeap:
# Description

The latest rust release comes with new more opinionated clippy :). This PR fixes the new clippy errors and and runs `cargo clippy --fix` on all our crates.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/writer/mod.rs
+++ b/rust/src/writer/mod.rs
@@ -1,7 +1,7 @@
 #![cfg(all(feature = "arrow", feature = "parquet"))]
 //! Abstractions and implementations for writing data to delta tables
 
-use crate::action::{Action, Add, ColumnCountStat, Stats};
+use crate::action::{Action, Add, ColumnCountStat};
 use crate::{DeltaDataTypeVersion, DeltaTable, DeltaTableError};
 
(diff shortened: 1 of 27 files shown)


rust/src/writer/mod.rs
```diff
@@ -54,12 +54,12 @@ pub(crate) enum DeltaWriterError {
...
     },
 
-    // TODO: derive Debug for Stats in delta-rs
     /// Serialization of delta log statistics failed.
     #[error("Serialization of delta log statistics failed")]
     StatsSerializationFailed {
         /// The stats object that failed serialization.
-        stats: Stats,
+        #[from]
+        stats: serde_json::Error,
```
roeap, line 62: This is the only manual change. Clippy was complaining that the error variant was much larger then all the others. I felt forwarding the error rather then the actual data that caused the error is a reasonable change.

wjones127: That seems fair. The other option is to `Box` the Stats, but I think the `serde_json` error should be sufficient.

Could we update the doc comment above, though?

roeap: Of course! Going forward we way consider lifting the docs requirement for error variants. I feel the description test and variant name are usually expressive enough.

wjones127 (approved):

Issue #1093, opened 2023-01-23
deltalake.DeltaTable constructor hangs in Jupyter


xbrianh:
# Environment

**Delta-rs version**:0.6.0

**Binding**: python 3.10

**Environment**:
- **Cloud provider**: Azure
- **OS**: ubuntu, windws WSL
- **Other**:

***
# Bug

**What happened**: DeltaTable constructor hangs in Jupyter
```
dt = deltatable.DeltaTable(“az://container/…”, storage_options={‘account_name’: “foo”, “AZURE_STORAGE_ACCOUNT_KEY”: “bar”})
```

**What you expected to happen**: Not hang

**How to reproduce it**: Attempt to create an azure storage backed DeltaTable in Jupyter

**More details**: This was lightly discussed in #915

jabbera:
I am seeing this issue as well.

MrPowers:
@xbrianh - can you try delta-rs v0.6.4?  If I recall correctly, this was a 0.6.0 bug that was fixed.  Let me know if that helps!

xbrianh:
@MrPowers Same issue:
```
>pip show deltalake
Name: deltalake
Version: 0.6.4
Summary: Native Delta Lake Python binding based on delta-rs with Pandas integration
Home-page: https://github.com/delta-io/delta-rs
Author: Qingping Hou <dave2008713@gmail.com>
Author-email: Qingping Hou <dave2008713@gmail.com>
License: Apache-2.0
Location: /Users/xath/.virtualenvs/p310/lib/python3.11/site-packages
Requires: pyarrow
Required-by:
```

craustin:
This was never resolved for me in https://github.com/delta-io/delta-rs/issues/915. I can still repro in `deltalake==0.7.0`.

jabbera:
@MrPowers I compiled a develop version and ran a backtrace. Looks like it's hanging writing to stdio for some reason: https://github.com/delta-io/delta-rs/blob/aecccbaf8b024f4ebb07bcc6096e5ed3c3d9fdc2/rust/src/action/parquet_read/mod.rs#L468

```
__GI___libc_write (nbytes=792, buf=0x32a9f90, fd=2) at ../sysdeps/unix/sysv/linux/write.c:26
26      ../sysdeps/unix/sysv/linux/write.c: No such file or directory.
(gdb) bt
[...]

PR #1112, opened 2023-01-29
fix: change unexpected field logging level to debug

houqp:
# Description
logging this error per row is too verbose when reading newer unsupported versions of delta log

# Related Issue(s)
closes https://github.com/delta-io/delta-rs/issues/1093


roeap (approved):
Thanks!

PR #1115, opened 2023-02-01
fix: avoid some allocations in DeltaStorageHandler

### LINKED ISSUE #1109: Write Table to Azure does not work: TypeError: argument 'add_actions': failed to extract field PyAddAction.size
# Environment

**Delta-rs version**:
0.7.0

**Environment**:
- **Cloud provider**: Azure
- **OS**: Mac

***
# Bug

**What happened**:
Error message when trying to create a new table in Azure Datalake Storage:
TypeError: argument 'add_actions': failed to extract field PyAddAction.size
image

Delta Table should be created and data appended

**How to reproduce it**:
My Python script:

> from deltalake.writer import write_deltalake
> import pandas as pd
> df = pd.DataFrame({'x': [1, 2, 3]})
> 
> delta_table_path = f"adl://eventhub/delta_table_name"
> 
> storage_options = {
>     "account_name": "storage_account_name",
>     "azure_storage_tenant_id" : "280...3e8",
>     "AZURE_STORAGE_ACCOUNT_KEY": "abc123...==",
> }
> 
> write_deltalake(table_or_uri=delta_table_path, data=df, mode="overwrite", storage_options=storage_options)

**More details**:
The Azure container "eventhub" does exist, otherwise a more specific "container is missing" error is thrown
The azure container is empty (no delta table exists yet) before the script execution. Afterwards the parquet is there but the logs are missing:
image

Im getting the same error when trying to append to an existing delta table created with spark
Also when switching filesystem notations as mentioned in the [docs](https://delta-io.github.io/delta-rs/python/usage.html?highlight=adl#)

roeap:
# Description

Fixes an error in the `DeltaFileSystemHandler`, when reading file metadata from remote storages. Due to an inconsistency between the behaviour object stores when invoking list operations on a path that points to a file, we incorrectly returned an Directory type for files in case of object stores. The bug only surfaced when using pyarrow < 9, since we used the call only when getting the file size, which we avoid when using more recent pyarrow versions.

@tustvold - I seem to vaguely remember discussing this at some point, but am not sure anymore. Is this something we should look into in object-store? 

Update: validated locally, that the upstream fixes will fix the linked issue, so the main reason for this PR is resolved elsewhere. There are some changes included which safe us some allocation (admittedly very few), but hopefully an improvement anyhow.

# Related Issue(s)

closes #1109 

# Documentation


tustvold:
> I seem to vaguely remember discussing this at some point, but am not sure anymore. Is this something we should look into in object-store?

`object_store` should not return directories in listing responses, it follows the object-store model and not a filesystem model. If one of the implementations is returning directories that is a bug.

roeap:
From what I could tell happend. If you have a store with a single object `file` at the root. Using list (or list with prefix) like

```rs
path = Path::from("file")
store.list(Some(path))
```

I think the remote stores will return a list containing that single object, whereas `LocalFileSystem` will return an empty list.

roeap:
turning this back into a draft, since the upstream fixes are in the works ...

wjones127 (approved):

PR #1119, opened 2023-02-05
improve debuggability of json ser/de errors

houqp:
# Description

JSON serde errors are very hard to debug because they are usually displayed with the following vague error messages:

```
Error: Invalid JSON in log record: EOF while parsing a value at line 1 column 1                                                                                                              
                                                                                                                                                                                             
Caused by:                                                                                                                                                                                   
    EOF while parsing a value at line 1 column 1  
```

This PR separates JSON ser/de errors by different scenarios as well as adding more context to each error whenever applicable.

### DIFF
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -8,7 +8,7 @@ mod parquet_read;
 #[cfg(feature = "parquet2")]
 pub mod parquet2_read;
 
-use crate::{schema::*, DeltaTableMetaData};
+use crate::{schema::*, DeltaTableError, DeltaTableMetaData};
 use percent_encoding::percent_decode;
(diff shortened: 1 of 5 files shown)


rust/src/action/mod.rs
```diff
@@ -332,6 +332,25 @@ impl MetaData {
     }
 }
 
+impl TryFrom<DeltaTableMetaData> for MetaData {
```
houqp, line 335: drive by refactor, makes more sense to keep all `MetaData` impl in the same place.

houqp:
Looks like the python azure test has flaky timeout errors.

wjones127 (approved):
This is great! Thank you.

PR #1120, opened 2023-02-05
minor: optimize partition lookup for vacuum loop

houqp:
# Description

Avoid resolving table partition inside the file iteration loop


wjones127 (approved):
Nice! Thanks for addressing this.

PR #1122, opened 2023-02-05
add test for min_max_schema_for_fields

marijncv:
Signed-off-by: Marijn Valk <marijncv@hotmail.com>

# Description
Adds tests for `min_max_schema_for_fields`

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/delta_arrow.rs
+++ b/rust/src/delta_arrow.rs
@@ -734,61 +734,47 @@ mod tests {
     }
 
     #[test]
-    fn test_max_min_schema_for_fields_simple_type() {
-        let mut dest: Vec<ArrowField> = Vec::new();
-        let f = ArrowField::new("simple", ArrowDataType::Int32, true);
-        let expected = vec![ArrowField::new("simple", ArrowDataType::Int32, true)];
-        max_min_schema_for_fields(&mut dest, &f);
-        assert_eq!(dest, expected)
-    }
-
-    #[test]
-    fn test_max_min_schema_for_fields_struct_type() {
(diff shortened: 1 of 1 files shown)


rust/src/delta_arrow.rs
```diff
@@ -732,4 +732,63 @@ mod tests {
...
+        let f = ArrowField::new(
+            "struct",
+            ArrowDataType::Struct(vec![ArrowField::new("simple", ArrowDataType::Int32, true)]),
+            true,
+        );
+        let expected = vec![ArrowField::new(
+            "struct",
+            ArrowDataType::Struct(vec![ArrowField::new("simple", ArrowDataType::Int32, true)]),
+            true,
+        )];
```
wjones127, line 757: We don't have to repeat this, and it's probably clearer they are meant to be the same field if we just clone `f`. Same applies for test above.
```suggestion
        let f = ArrowField::new(
            "struct",
            ArrowDataType::Struct(vec![ArrowField::new("simple", ArrowDataType::Int32, true)]),
            true,
        );
        let expected = vec![f.clone()];
```

wjones127 (changes requested):
Since `max_min_schema_for_fields` is meant to be used multiple times on the same destination, perhaps it might be better to combine the tests into one where you call the function on each of the fields, and then assert the final `dest` result contains the right fields.

marijncv:
Thanks for the pointers @wjones127!

wjones127 (approved):
Thanks @marijncv!

PR #1138, opened 2023-02-12
add function & test for parsing table_or_uri

### LINKED ISSUE #1123: Python: mixing up parameters to write_deltalake gives confusing error
# Description

```python
import pandas as pd

example_df = pd.DataFrame({
    "part": ["a", "a", "b", "b"],
    "value": [1, 2, 3, 4]
})

write_deltalake(example_df, "example_table")
```

```
ValueError: You must provide schema if data is Iterable
```

We should instead validate the `path` argument to make sure it can be used as a path. (Also maybe we should support `pathlib.Path` type objects too? That would simplify our tests.)

marijncv:
# Description
Adds a more specific ValueError if a wrong `table_or_uri` is provided in the `write_delta_lake` function. It also adds support for a pathlib `Path` object.

# Related Issue(s)
- closes #1123 

# Documentation

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -317,23 +317,33 @@ def try_get_table_and_table_uri(
     table_or_uri: Union[str, Path, DeltaTable],
     storage_options: Optional[Dict[str, str]] = None,
 ) -> Tuple[Optional[DeltaTable], str]:
-    """Parses `table_or_uri` and returns `table` & `table_uri`.
+    """Parses the `table_or_uri`.
 
-    Raises a ValueError if `table_or_uri` is not of type `str`, `Path` or `DeltaTable`
+    Parameters
+    ----------
+    table_or_uri : str, Path or DeltaTable
+        URI of a table or a DeltaTable object.
+    storage_options : dict(str, str), optional
+        Options passed to the native delta filesystem.
+
+    Returns
+    -------
+    table: DeltaTable
+        DeltaTable object
+    table_uri: str
+        URI of the table
+
+    Raises
+    ------
+    ValueError
+        If `table_or_uri` is not of type str, Path or DeltaTable
     """
     if not isinstance(table_or_uri, (str, Path, DeltaTable)):
         raise ValueError("table_or_uri must be a str, Path or DeltaTable")
 
-    if isinstance(table_or_uri, str):
-        if "://" in table_or_uri:
-            table_uri = table_or_uri
-        else:
-            # Non-existant local paths are only accepted as fully-qualified URIs
-            table_uri = "file://" + str(Path(table_or_uri).absolute())
+    if isinstance(table_or_uri, (str, Path)):
         table = try_get_deltatable(table_or_uri, storage_options)
-    elif isinstance(table_or_uri, Path):
-        table_uri = "file://" + str(table_or_uri.absolute())
-        table = try_get_deltatable(table_uri, storage_options)
+        table_uri = str(table_or_uri)
     else:
         table = table_or_uri
         table_uri = table._table.table_uri()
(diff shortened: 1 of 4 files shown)


python/tests/test_writer.py
```diff
@@ -501,3 +501,40 @@ def test_writer_with_options(tmp_path: pathlib.Path):
     )
 
     assert table == data
+
+
+def test_try_get_table_and_table_uri(tmp_path: pathlib.Path):
+    data = pa.table({"vals": pa.array(["1", "2", "3"])})
+    table_or_uri = str(tmp_path / "delta_table")
+    write_deltalake(table_or_uri, data)
```
wjones127, line 509: Shouldn't this be unnecessary now?
```suggestion
    table_or_uri = tmp_path / "delta_table"
    write_deltalake(table_or_uri, data)
```

python/tests/test_writer.py
```diff
@@ -501,3 +501,40 @@ def test_writer_with_options(tmp_path: pathlib.Path):
...
+    delta_table = DeltaTable(table_or_uri)
```
wjones127, line 510: I see. We need `DeltaTable.__init__` to support `Path` too. Would you be willing to add that as well?

python/tests/test_writer.py
wjones127: Nevermind, see below.

wjones127 (commented):
Thank you for doing this. Would you be wiling to add `Path` support to `DeltaTable` too? Then we can clean up most of the unit tests that have to do the Path to str conversion.

python/deltalake/writer.py
```diff
@@ -322,6 +313,34 @@ def default(self, obj: Any) -> Any:
...
+    """
+    if not isinstance(table_or_uri, (str, Path, DeltaTable)):
+        raise ValueError("table_or_uri must be a str, Path or DeltaTable")
+
+    if isinstance(table_or_uri, str):
+        if "://" in table_or_uri:
+            table_uri = table_or_uri
+        else:
+            # Non-existant local paths are only accepted as fully-qualified URIs
+            table_uri = "file://" + str(Path(table_or_uri).absolute())
```
roeap, line 332: by now we actually extended the path handling on the rust side to also support relative paths and scheme-less paths, and also create a directory if it does not yet exists.

https://github.com/delta-io/delta-rs/blob/67b512f601c85fb89839735164f9a1c818d58f5f/rust/src/builder.rs#L353-L376

Is it maybe a good idea to rely on the same logic on the python side as well, so we are always consistent in the supported paths?

Of course supporting `Path` objects will always be python territory :).

python/deltalake/writer.py
```diff
@@ -322,6 +313,34 @@ def default(self, obj: Any) -> Any:
         return json.JSONEncoder.default(self, obj)
 
 
+def try_get_table_and_table_uri(
+    table_or_uri: Union[str, Path, DeltaTable],
+    storage_options: Optional[Dict[str, str]] = None,
+) -> Tuple[Optional[DeltaTable], str]:
+    """Parses `table_or_uri` and returns `table` & `table_uri`.
+
+    Raises a ValueError if `table_or_uri` is not of type `str`, `Path` or `DeltaTable`
```
roeap, line 322: Should we maybe write this in numpy docstring format? https://numpydoc.readthedocs.io/en/latest/format.html#raises

roeap: I am very sorry, but I said something wrong. We are not using the numpy format... we are using the reST format. not sure how raises are done there. @wjones127 - do you know? I could not find good docs for that.

python/tests/test_writer.py
marijncv: Added it there! And also in `try_get_deltatable`

python/deltalake/writer.py
marijncv: Thanks for the pointer! That makes the python code a bit simpler

python/deltalake/writer.py
marijncv: Ah yes, the format shown [here](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) looks very similar to what is used in other places in the codebase and it also specifies a way to describe a raise. I can rewrite it in that format if it's the correct one to use

wjones127: Yes that's the format. Could you rewrite that?

wjones127 (approved):
Thanks! This is a great improvement :)

PR #1147, opened 2023-02-13
use Path object in writer tests

marijncv:
Signed-off-by: Marijn Valk <marijncv@hotmail.com>

# Description
Now that #1138 is merged, we can use `pathlib.Path` instead of `str` in `write_deltalake()` and `DeltaTable.__init__()`


wjones127 (approved):
Thank you!

Issue #1136, opened 2023-02-11
Reorganize crate into fewer top-level modules


wjones127:
# Description

I starting to look at better documenting the Rust crate, but I realized it partly looks complex because the crate has so many top-level modules.

It seems like we could simplify with the following:

* [ ] Make `schema.rs` into a directory module (`schema/mod.rs`) and move `delta_arrow.rs` into that directory.
* [ ] Merge `delta_config.rs` and `table_properties.rs`.
* [ ] Move `optimize.rs` into `operations`?
* [ ] Rename `delta.rs` to `table/mod.rs`. Move `builder.rs`, `table_state.rs`, and `table_arrow_state.rs` into this module. And maybe `checkpoint.rs`?

**Use Case**

**Related Issue(s)**

roeap:
This has been on my mind for a while as well. I think your proposed structure makes a lot of sense. One thing to consider with the checkpoints is, that these are very closely tied to the parquet implementations / features. SO maybe it makes sense, to move `checkpoint.rs` into the `action` module.

Also - and this is just an assumption - as we move forward I see us iterating on how we process table state and an even stricter separation of the state related apis and the still low level, but higher level then state apis on the table might be desirable as well. Especially considering the upcoming delta-core discussions.

wjones127:
> One thing to consider with the checkpoints is, that these are very closely tied to the parquet implementations / features. SO maybe it makes sense, to move checkpoint.rs into the action module.

That's a good point. The actions module seems like a better place for that 👍

PR #1154, opened 2023-02-17
feat: move and update Optimize operation

roeap:
# Description

This PR moves the optimize operation into the operations module. As part of this we do a few updates to the operation as well

- adopt `IntoFuture` pattern
- use writer from operations module
- replace `SerializedFileReader` with `ParquetRecordBatchStream`

As part of this we also update datafusion and arrow, which in turn requires updating pyo3. This requires updating some deprecated features. i.e. how function signatures are annotated. It also leads to a breaking change in the python funcitons - specifically the order of arguments in `dataset_partitions`. 

There is one more thing is was considering while doing these updates. Essentially we may be getting some undesired behaviour, as the writer in the operations module already considers file size, so not all data in a bin might actually end up in the same file. there may also be a naïve yet much simpler way to do this now, by just passing all data in a partition through the partition writer. This should yield files of the desired size within the accuracy of the configured row-group size. It may of course lead to a rather small remainder file.

cc @Blajda

# Related Issue(s)

part of #1136

# Documentation

### DIFF
--- a/rust/src/writer/json.rs
+++ b/rust/src/writer/json.rs
@@ -339,16 +339,20 @@ impl DeltaWriter<Vec<Value>> for JsonWriter {
         }
 
         if !partial_writes.is_empty() {
-            let sample = partial_writes.first().map(|t| t.to_owned());
-            if let Some((_, e)) = sample {
-                return Err(DeltaWriterError::PartialParquetWrite {
-                    skipped_values: partial_writes,
-                    sample_error: e,
-                }
-                .into());
-            } else {
-                unreachable!()
+            return Err(DeltaWriterError::PartialParquetWrite {
+                sample_error: match &partial_writes[0].1 {
+                    ParquetError::General(msg) => ParquetError::General(msg.to_owned()),
+                    ParquetError::ArrowError(msg) => ParquetError::ArrowError(msg.to_owned()),
+                    ParquetError::EOF(msg) => ParquetError::EOF(msg.to_owned()),
+                    ParquetError::External(err) => ParquetError::General(err.to_string()),
+                    ParquetError::IndexOutOfBound(u, v) => {
+                        ParquetError::IndexOutOfBound(u.to_owned(), v.to_owned())
+                    }
+                    ParquetError::NYI(msg) => ParquetError::NYI(msg.to_owned()),
+                },
+                skipped_values: partial_writes,
             }
+            .into());
         }
 
         Ok(())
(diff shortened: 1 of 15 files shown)


rust/src/writer/json.rs
```diff
@@ -339,16 +339,20 @@ impl DeltaWriter<Vec<Value>> for JsonWriter {
...
+                    ParquetError::General(msg) => ParquetError::General(msg.to_owned()),
+                    ParquetError::ArrowError(msg) => ParquetError::ArrowError(msg.to_owned()),
+                    ParquetError::EOF(msg) => ParquetError::EOF(msg.to_owned()),
+                    ParquetError::External(err) => ParquetError::General(err.to_string()),
+                    ParquetError::IndexOutOfBound(u, v) => {
+                        ParquetError::IndexOutOfBound(u.to_owned(), v.to_owned())
+                    }
+                    ParquetError::NYI(msg) => ParquetError::NYI(msg.to_owned()),
+                },
+                skipped_values: partial_writes,
```
roeap, line 353: I was not too happy doing this, but the parquet errors are no longer cloneable, so `to_onwed` does not do much. My assumption was that we have to keep all the individual errors, since kafka-delta-ingest relies on them? @rtyler

rust/tests/command_optimize.rs
```diff
@@ -311,15 +315,16 @@ async fn test_conflict_for_remove_actions() -> Result<(), Box<dyn Error>> {
...
     transaction.commit(None, None).await?;
 
-    let maybe_metrics = plan.execute(&mut dt).await;
+    let maybe_metrics = plan.execute(dt.object_store()).await;
     assert!(maybe_metrics.is_err());
     assert_eq!(dt.version(), version + 1);
     Ok(())
 }
 
+#[ignore = "we do not yet re-try in operations commits."]
```
roeap, line 327: I'm finally working up to doing the commits conflict resolution. so this will be re-enabled soonishly ..

Blajda:
@roeap The optimize changes LGTM.

> There is one more thing is was considering while doing these updates. Essentially we may be getting some undesired behaviour, as the writer in the operations module already considers file size, so not all data in a bin might actually end up in the same file. there may also be a naïve yet much simpler way to do this now, by just passing all data in a partition through the partition writer. This should yield files of the desired size within the accuracy of the configured row-group size. It may of course lead to a rather small remainder file.

The bin packing optimization must be idempotent so I think just passing the entire partition would violate that. Sorting files by their size ensures that. It's good that you kept the check for only one add action so any undesired behaviour should be caught and reported.

wjones127 (approved):

PR #1158, opened 2023-02-18
first setup of ruff for python linting

### LINKED ISSUE #1149: Adopt ruff for python linting in CI/CD
# Description

As proud members of the Rust community, as well as advocates of rust backed python, I thought we could adopt [ruff](https://pypi.org/project/ruff/) as our main python linter in CI/CD.

Its not a huge pain right now, but the change is also quite minimal in terms of effort, and I personally made great experiences with it several projects.

As a bonus we could add delta-rs to the list of adopters alongside many other great projects like, pydantic, polars, pandas and many more...

**Use Case**

**Related Issue(s)**

marijncv:
# Description
This PR sets up ruff for python linting. It also enables isort from within ruff and removes the standalone isort. 

# Related Issue(s)
closes #1149

### DIFF
--- a/python/deltalake/schema.py
+++ b/python/deltalake/schema.py
@@ -5,7 +5,14 @@ import pyarrow as pa
 if TYPE_CHECKING:
     import pandas as pd
 
-from ._internal import ArrayType, Field, MapType, PrimitiveType, Schema, StructType
+from ._internal import (  # noqa: F401
+    ArrayType,
+    Field,
+    MapType,
+    PrimitiveType,
+    Schema,
+    StructType,
+)
 
 # Can't implement inheritance (see note in src/schema.rs), so this is next
 # best thing.
--- a/python/deltalake/schema.py
+++ b/python/deltalake/schema.py
@@ -5,14 +5,12 @@ import pyarrow as pa
 if TYPE_CHECKING:
     import pandas as pd
 
-from ._internal import (  # noqa: F401
-    ArrayType,
-    Field,
-    MapType,
-    PrimitiveType,
-    Schema,
-    StructType,
-)
+from ._internal import ArrayType as ArrayType
+from ._internal import Field as Field
+from ._internal import MapType as MapType
+from ._internal import PrimitiveType as PrimitiveType
+from ._internal import Schema as Schema
+from ._internal import StructType as StructType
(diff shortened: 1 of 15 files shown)


wjones127 (approved):
This looks great. Was easy to get working locally for me.

Before I can merge, could you fix the conflicts in the test file?

python/deltalake/schema.py
```diff
@@ -5,7 +5,14 @@
 if TYPE_CHECKING:
     import pandas as pd
 
-from ._internal import ArrayType, Field, MapType, PrimitiveType, Schema, StructType
+from ._internal import (  # noqa: F401
```
roeap, line 8: This is not so critical, however some static type checkers (I know about pyright, not sure if mypy has something similar), will consider this internal symbols. That why I usually add them to the `__all__` field if the intend is fo users to import from that location, which fixes F401 as well. 

More recently I cam across another alternative, that I think I like ...

```py
from ._internal import ArrayType as ArrayType
```

while this puts every "export" on a single line, I somehow find it more aesthetically pleasing then the __all__ way 😆.

The only scenario where making the exports explicit is - AFAIK - when someone uses "*" imports.

python/tests/conftest.py
```diff
@@ -13,11 +13,10 @@
 
 def wait_till_host_is_available(host: str, timeout_sec: int = 30):
     spacing = 2
-    attempts = timeout_sec / spacing
     while True:
         try:
             subprocess.run(["curl", host], timeout=500, check=True)
-        except:
+        except BaseException:
```
roeap, line 19: I think we should not be catching base exception here, but rather `Exception`.

learned the hard way once what it says in pep8 :).

> Direct inheritance from BaseException is reserved for exceptions where catching them is almost always the wrong thing to do.

The case that comes to mind is e.g. keyboard interrupts, which we would catch with the BaseException as well.

roeap (changes requested):
Thanks! I had one comment around catching BaseException vs. Exception, but overlall looks great!

marijncv:
Thanks for the pointers @roeap & @wjones127!

wjones127 (approved):

roeap (approved):

PR #1159, opened 2023-02-19
add test for extract_partition_values

marijncv:
# Description
Adds a test for `rust/src/writer/json.rs::extract_partition_values`


wjones127 (approved):
Thanks, this looks good @marijncv

As a related follow-up, would you be interested in working on #1170? That could also use more tests.

Issue #1146, opened 2023-02-13
Python write_deltalake fails if pyarrow table contains binary columns ending with 0x5c


sfilimonov-exos:
# Environment

**Delta-rs version**:  0.7.0

**Environment**:
- **OS**: OS X or Ubuntu 22.04

***
# Bug

**What happened**: write_deltalake fails when trying to add file to the transaction because it cannot encode some binary values.

```
Exception ignored in: 'pyarrow._dataset._filesystemdataset_write_visitor'
Traceback (most recent call last):
  File "/Users/abc/repos/study_python/venv/lib/python3.9/site-packages/deltalake/writer.py", line 229, in visitor
    json.dumps(stats, cls=DeltaJSONEncoder),
  File "/usr/local/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/local/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/abc/repos/study_python/venv/lib/python3.9/site-packages/deltalake/writer.py", line 315, in default
    return obj.decode("unicode_escape")
UnicodeDecodeError: 'unicodeescape' codec can't decode byte 0x5c in position 1: \ at end of string

```

**How to reproduce it**:

```
import pyarrow

from deltalake.writer import write_deltalake

value = b'\x00\\'
pa_table = pyarrow.Table.from_pydict({'field_one': [value]})

write_deltalake("some_table_1234", pa_table)
```

PR #1167, opened 2023-02-20
Python write_deltalake fails if pyarrow table contains binary columns

rbushri:
# Description
Python write_deltalake fails if pyarrow table contains binary columns ending with 0x5c

# Related Issue(s)
Python write_deltalake fails if pyarrow table contains binary columns ending with 0x5c #1146

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -333,7 +333,7 @@ def __enforce_append_only(
 class DeltaJSONEncoder(json.JSONEncoder):
     def default(self, obj: Any) -> Any:
         if isinstance(obj, bytes):
-            return obj.decode("unicode_escape", "ignore")
+            return obj.decode("unicode_escape", "backslashreplace")
         elif isinstance(obj, date):
             return obj.isoformat()
         elif isinstance(obj, datetime):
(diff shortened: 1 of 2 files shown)


python/deltalake/writer.py
```diff
@@ -302,7 +302,7 @@ def __enforce_append_only(
 class DeltaJSONEncoder(json.JSONEncoder):
     def default(self, obj: Any) -> Any:
         if isinstance(obj, bytes):
-            return obj.decode("unicode_escape")
+            return obj.decode("unicode_escape", "ignore")
```
wjones127, line 305: Based on [the protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization), I think we actually want to escape with `\u`:
```suggestion
            return obj.decode("unicode_escape", "backslashreplace")
```

wjones127 (changes requested):
Also, I tested the same data in PySpark, and they don't write stats. So we could also disable statistics collection for binary columns in the first place. I suppose not many people are using inequality filters on binary data. But this change should be sufficient for now.

wjones127:
Also, could you add a quick unit test based on your example in `python/tests/test_writer.py`?

wjones127:
@rbushri are you wiling to add that unit test?

wjones127 (approved):
Thanks for opening this @rbushri!

PR #1173, opened 2023-02-23
Make DeltaTable clonable

### LINKED ISSUE #1172: Make DeltaTable clonable
# Description

**Use Case**
I'm working on migrating the storage backend of [Seafowl](https://github.com/splitgraph/seafowl) to use delta-rs instead (thus replacing our [custom](https://www.splitgraph.com/blog/seafowl-time-travel) MVCC/Lakehouse protocol).

It would be really handy if `DeltaTable` was `Clone`, among other things to make use of the datafusion's [SchemaProvider::register_table](https://github.com/apache/arrow-datafusion/blob/a79f9c889c29a736e727dcb29ede1c2d25618a70/datafusion/core/src/catalog/schema.rs#L45-L53) method.

In particular, this method provides as input `Arc<dyn TableProvider>`, and if we store `DeltaTable`s in the map we need to downcast, which only provides a reference to the table behind the trait object. Consequently, to update the map we'd need to re-create the table using e.g. a builder, instead of simply calling `.clone()` on the ref.

**Related Issue(s)**

gruuya:
# Description
Simply add `Clone` attribute to `DeltaTable` as well as the contained `DeltaTableConfig` (see #1172 for motivation).

# Related Issue(s)
Closes #1172

# Documentation

### DIFF
--- a/rust/src/builder.rs
+++ b/rust/src/builder.rs
@@ -54,7 +54,7 @@ impl Default for DeltaVersion {
 }
 
 /// Configuration options for delta table
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct DeltaTableConfig {
     /// Indicates whether our use case requires tracking tombstones.
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -443,6 +443,7 @@ pub enum PeekCommit {
 }
 
 /// In memory representation of a Delta Table
+#[derive(Clone)]
 pub struct DeltaTable {
     /// The state of the table as of the most recent loaded Delta log entry.
     pub state: DeltaTableState,


roeap:
While in principle a minor change, I am not 100% certain the we can promise to maintain that going forward. Cloning the internal state is potentially very expensive, in some cases for e.g. streaming ingest prohibitively so. And we are currently looking into some potentially significant changes to our state handling.

 That said, we are eager to support anyone adopting delta.rs - especially on the Rust side. However I am not entirely sure I understood the issue. In our tests we use the register_table methods - at lest on the context object - with relative ease. In which scenario would we update the registered table? When the underlying table got updated?

Would it also be sufficient to make `DeltaTable::new_with_state` a public method, so you could clone the state directly and create a new table from that?

Then again, maybe I am overthinking this :)

gruuya:
> Cloning the internal state is potentially very expensive, in some cases for e.g. streaming ingest prohibitively so.

Oh I see. I was wondering what was the reason for the lack of `Clone`.

> In which scenario would we update the registered table?

One scenario is multi-node deployment; if a write gets executed on one node we want to `load` the table on the other node(s) to pick up the changes. For instance if the tables are stored as `tables: RwLock<HashMap<Arc<str>, Arc<DeltaTable>>>` it would be easiest to just call `load` on them, but that is not doable since `Arc` can't deref to a mutable variable (which `load` needs). (We define the `tables` map like so since it's handy to satisfy the entire `SchemaProvider` interface.)

The next simplest option is to clone the table to a new mutable variable and then call `load` on it, hence this issue/PR. 

Alternatively, I can rebuild and load the table from scratch when Datafusion requests it via [`SchemaProvider::table`](https://github.com/apache/arrow-datafusion/blob/main/datafusion/core/src/execution/context.rs#L1858) which is slightly more verbose but should work, so I'll try that instead.

So I think `DeltaTable::new_with_state` can stay private outside of the repo for now.

Thanks.

PR #1176, opened 2023-02-24
feat: delete operation

### LINKED ISSUE #832: Implement simple delete case
# Description

It's not uncommon to delete along partition boundaries. In those cases the delete should be very simple to implement; just commit some remove actions.

I think we should probably design the API so we can eventually support any where clause in deletes, rather than have a separate function for this simple case and the case where we have to rewrite files.

See the `write` operation implementation for inspiration: https://github.com/delta-io/delta-rs/blob/main/rust/src/operations/write.rs

**Use Case**

**Related Issue(s)**

Blajda:
# Description

This is a full implementation of the Delta Delete Command. Users can now delete records that match a predicate. This implementation is not limited to only partition columns and allows non-partition columns.

This also implements a `find_files` function  which can be used to implement the Update command.

# Related Issue(s)

- closes #832 

# Documentation

### DIFF
--- a/rust/src/operations/delete.rs
+++ b/rust/src/operations/delete.rs
@@ -28,6 +28,7 @@ use crate::operations::write::write_execution_plan;
 use crate::storage::DeltaObjectStore;
 use crate::storage::ObjectStoreRef;
 use crate::table_state::DeltaTableState;
+use crate::DeltaDataTypeVersion;
 use crate::DeltaTable;
 use crate::DeltaTableError;
 
@@ -47,6 +48,7 @@ use datafusion::physical_plan::file_format::{wrap_partition_type_in_dict, FileSc
 use datafusion::physical_plan::filter::FilterExec;
 use datafusion::physical_plan::limit::LocalLimitExec;
 use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::RecordBatchStream;
 use datafusion::prelude::Expr;
 use datafusion_common::scalar::ScalarValue;
 use datafusion_common::DFSchema;
@@ -57,8 +59,11 @@ use parquet::file::properties::WriterProperties;
 use serde_json::Map;
 use serde_json::Value;
 use std::collections::HashMap;
+use std::pin::Pin;
 use std::sync::Arc;
 use std::time::{Instant, SystemTime, UNIX_EPOCH};
+
+const PATH_COLUMN: &str = "__delta_rs_path";
 
 /// Delete Records from the Delta Table.
 /// See this module's documentaiton for more information
@@ -74,7 +79,7 @@ pub struct DeleteBuilder {
     /// Properties passed to underlying parquet writer for when files are rewritten
     writer_properties: Option<WriterProperties>,
     /// Additional metadata to be added to commit
-    app_metadata: Option<HashMap<String, serde_json::Value>>,
+    app_metadata: Option<Map<String, serde_json::Value>>,
 }
 
 #[derive(Default, Debug)]
@@ -108,7 +113,6 @@ async fn find_files<'a>(
 ) -> DeltaResult<Vec<&'a Add>> {
     let mut files = Vec::new();
     let mut candidate_map: HashMap<String, &'a Add> = HashMap::new();
-    const PATH_COLUMN: &str = "__delta_rs_path";
 
     let table_partition_cols = snapshot
         .current_metadata()
@@ -175,38 +179,63 @@ async fn find_files<'a>(
     let limit: Arc<dyn ExecutionPlan> = Arc::new(LocalLimitExec::new(filter, 1));
 
     let task_ctx = Arc::new(TaskContext::from(state));
-    for i in 0..limit.output_partitioning().partition_count() {
-        let mut stream = limit.execute(i, task_ctx.clone())?;
+    let partitions = limit.output_partitioning().partition_count();
+    let mut tasks = Vec::with_capacity(partitions);
 
-        while let Some(res) = stream.next().await {
-            let batch: RecordBatch = res?;
-            let array = batch
-                .column_by_name(PATH_COLUMN)
-                .unwrap()
-                .as_any()
-                .downcast_ref::<StringArray>()
-                .ok_or(DeltaTableError::Generic(format!(
-                    "Unable to downcast column {}",
-                    PATH_COLUMN
-                )))?;
+    for i in 0..partitions {
+        let stream = limit.execute(i, task_ctx.clone())?;
+        tasks.push(handle_stream(stream));
+    }
 
-            for path in array.into_iter() {
-                let path = path.ok_or(DeltaTableError::Generic(format!(
-                    "{} cannot be null",
-                    PATH_COLUMN
-                )))?;
-                match candidate_map.remove(path) {
-                    Some(action) => files.push(action),
-                    None => {
-                        return Err(DeltaTableError::Generic(
-                            "Unable to map __delta_rs_path to action.".to_owned(),
-                        ))
-                    }
+    for res in futures::future::join_all(tasks).await.into_iter() {
+        let path = res?;
+        if let Some(path) = path {
+            match candidate_map.remove(&path) {
+                Some(action) => files.push(action),
+                None => {
+                    return Err(DeltaTableError::Generic(
+                        "Unable to map __delta_rs_path to action.".to_owned(),
+                    ))
                 }
             }
         }
     }
+
     Ok(files)
+}
+
+async fn handle_stream(
+    mut stream: Pin<Box<dyn RecordBatchStream + Send>>,
+) -> Result<Option<String>, DeltaTableError> {
+    if let Some(maybe_batch) = stream.next().await {
+        let batch: RecordBatch = maybe_batch?;
+        if batch.num_rows() > 1 {
+            return Err(DeltaTableError::Generic(
+                "Find files returned multiple records for batch".to_owned(),
+            ));
+        }
+        let array = batch
+            .column_by_name(PATH_COLUMN)
+            .unwrap()
+            .as_any()
+            .downcast_ref::<StringArray>()
+            .ok_or(DeltaTableError::Generic(format!(
+                "Unable to downcast column {}",
+                PATH_COLUMN
+            )))?;
+
+        let path = array
+            .into_iter()
+            .next()
+            .unwrap()
+            .ok_or(DeltaTableError::Generic(format!(
+                "{} cannot be null",
+                PATH_COLUMN
+            )))?;
+        return Ok(Some(path.to_string()));
+    }
+
+    Ok(None)
 }
 
 struct ExprProperties {
@@ -415,7 +444,7 @@ impl DeleteBuilder {
         mut self,
         metadata: impl IntoIterator<Item = (String, serde_json::Value)>,
     ) -> Self {
-        self.app_metadata = Some(HashMap::from_iter(metadata));
+        self.app_metadata = Some(Map::from_iter(metadata));
         self
     }
 
@@ -548,7 +577,7 @@ async fn execute(
     state: SessionState,
     writer_properties: Option<WriterProperties>,
     app_metadata: Option<Map<String, Value>>,
-) -> DeltaResult<DeleteMetrics> {
+) -> DeltaResult<((Vec<Action>, DeltaDataTypeVersion), DeleteMetrics)> {
     // TODO:
     // Metrics for records written and deleted
     let mut metrics = DeleteMetrics::default();
@@ -591,6 +620,7 @@ async fn execute(
         .as_millis() as i64;
 
     let mut actions: Vec<Action> = add_actions.into_iter().map(Action::add).collect();
+    let mut version = snapshot.version();
     metrics.num_removed_files = to_delete.len();
     metrics.num_added_files = actions.len();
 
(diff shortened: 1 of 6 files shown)


Blajda:
I want to continue to push this forward to have a MVP for deletion. My biggest issue with the current implementation is  the `find_files` function since it has iterate through each file in series to determine if it contains a record.  
If Datafusion supported a method like `input_file_name` then the multiple scans can be performed in parallel and would speed up the operation.

Requires further changes when #1303 is merged.
Currently there is an optimization for when the columns in the expression are only partitions columns. I may disable it until all pruning issues are resolved.

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,812 @@
...
+    type IntoFuture = BoxFuture<'static, Self::Output>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        let this = self;
+
+        Box::pin(async move {
+            let writer_properties = this
+                .writer_properties
+                .unwrap_or_else(|| WriterProperties::builder().build());
+
```
roeap, line 578: We can just pass the `Option` into the `write_execution_plan` function, which will create the default.

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,812 @@
...
+            });
+
+            let metrics = execute(
+                this.predicate,
+                this.store.clone(),
+                &this.snapshot,
+                state,
+                writer_properties,
+                None,
+            )
```
roeap, line 595:
```suggestion
            let metrics = execute(
                this.predicate,
                this.store.clone(),
                &this.snapshot,
                state,
                writer_properties,
                this.app_metadata,
            )
```

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,812 @@
...
+            .await?;
+            let mut table = DeltaTable::new_with_state(this.store, this.snapshot);
+            table.update().await?;
```
roeap, line 598: Here we can probably avoid some io by returning the actions and version from `execute` and updating the new state directly. Essentially something like.

```rs
this.snapshot.merge(DeltaTableState::from_actions(actions, version)?, true, true);
```

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,812 @@
...
+    let add_actions = write_execution_plan(
+        snapshot,
+        state.clone(),
+        filter.clone(),
+        table_partition_cols.clone(),
+        object_store.clone(),
+        None,
+        None,
+        Some(writer_properties),
+    )
```
roeap, line 470:
```suggestion
    let add_actions = write_execution_plan(
        snapshot,
        state.clone(),
        filter.clone(),
        table_partition_cols.clone(),
        object_store.clone(),
        Some(snapshot.table_config().target_file_size() as usize),
        None,
        Some(writer_properties),
    )
```

I guess a better way would be to extract defaults from the table config within `write_execution_plan`.

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,812 @@
...
+        let execution_props = ExecutionProps::new();
+
+        let predicate_expr = create_physical_expr(
+            expression,
+            &logical_schema,
+            &physical_schema,
+            &execution_props,
+        )?;
+        let filter = Arc::new(FilterExec::try_new(predicate_expr, parquet_scan)?);
+        let limit: Arc<dyn ExecutionPlan> = Arc::new(GlobalLimitExec::new(filter, 0, Some(1)));
```
roeap, line 136: Haven't fully thought this through, but maybe one way for us to exploit parallelism may be to create the `ParquetScan` in a way that each file group contains exactly one file and use `LocalLimitExec`. Looked a little bit throught the df code, and I think the order would be preserved, so we can infer from the partition number the file that was read.

If that cannot work, and we have to add the file name as a column, that should be feasible by treating it as as "virtual partition coluimn". However this may required some more updates and special caeses how we create the schema / statistics in various places.. Including the file name as partition value in `PartitionedFile` is straight forward though.

roeap (commented):
This is great work @Blajda!

Left some comments mostly around some Option treatment. Also had a thought on how we can achieve the parallelisim you mentioned when scanning files, but not sure yet if that will actually work :)

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,812 @@
...
+                &mut metrics,
+                writer_properties,
+                expr_properties,
+            )
+            .await?
+        }
+        None => (Vec::<Add>::new(), snapshot.files().to_owned()),
+    };
+
+    let predicate = predicate.unwrap_or(Expr::Literal(ScalarValue::Boolean(Some(true))));
```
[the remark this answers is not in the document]
Blajda: I wanted the output of predicate to be similar to the Spark impl. For Spark, when no predicate is provided the commit entry looks like
```
{"commitInfo":{"timestamp":1682378206638,"operation":"DELETE","operationParameters":{"predicate":"[\"true\"]"},"readVersion":0,"isolationLevel":"Serializable","isBlindAppend":false,"operationMetrics":
```
My concern is if we leave if None then it would write `"predicate":null`

rust/src/operations/delete.rs
Blajda: Thanks. The primary insight that I missed was adding the file path as a partition column. I'll give this a try and report any issues :)

rust/src/operations/delete.rs
Blajda: I implemented this and I imagine we would want to use this pattern for other operations. Should we return the version and actions in an optional? Currently I return an empty list + the current version if there are no changes.

rust/src/operations/delete.rs
Blajda: I was able to implement the suggestions. Let me know if you have any suggestions for the new implementation.

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,944 @@
...
+    expr: &Expr,
+    partition_columns: &Vec<String>,
+    properties: &mut ExprProperties,
+) -> DeltaResult<()> {
+    // TODO: We can likely relax the volatility to STABLE. Would require further
+    // research to confirm the same value is generated during the scan and
+    // rewrite phases.
+    match expr {
+        Expr::ScalarVariable(_, _) | Expr::Literal(_) => (),
+        Expr::Alias(expr, _) => validate_expr(expr, partition_columns, properties)?,
```
wjones127, line 258: Perhaps we can avoid recursion using the `TreeNode` trait methods on `Expr`?
https://docs.rs/datafusion/23.0.0/datafusion/prelude/enum.Expr.html#impl-TreeNode-for-Expr

Would help avoid stack overflows if there is a very nested expression. I could imagine that happening if someone passes a filter like:

```sql
x = 1
OR x = 2
OR x = 3
...
```

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,944 @@
...
+        Expr::Exists { .. } => {
+            return Err(DeltaTableError::Generic(
+                "Exists expression is not allowed in delete expression".to_string(),
+            ))
+        }
+        Expr::InSubquery { .. } => {
+            return Err(DeltaTableError::Generic(
+                "Subquery expression is not allowed in delete expression".to_string(),
+            ))
+        }
```
wjones127, line 370: I think we could probably replace a lot of these
```suggestion
        _ => {
            return Err(DeltaTableError::Generic(format!(
                "Expression is not allowed in delete expression: {}", expr
            )))
        }
```

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,944 @@
...
+                .merge(DeltaTableState::from_actions(actions, version)?, true, true);
+            let table = DeltaTable::new_with_state(this.store, this.snapshot);
+
+            Ok((table, metrics))
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
wjones127, line 696: I like all of these test 😄

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,944 @@
...
+                infinite_source: false,
+                output_ordering: None,
+            },
+            None,
+        )
+        .await?;
+
+    let filter: Arc<dyn ExecutionPlan> =
+        Arc::new(FilterExec::try_new(predicate_expr, parquet_scan.clone())?);
+    let limit: Arc<dyn ExecutionPlan> = Arc::new(LocalLimitExec::new(filter, 1));
```
wjones127, line 179: Does this issue a scan to every file? And maybe skips partitions that are irrelevant?

I wonder if we can do this in memory just with the partition columns. This example from DataFusion might be good inspiration:

https://github.com/apache/arrow-datafusion/blob/a38480951f40abce7ee2d5919251a1d1607f1dee/datafusion/core/src/datasource/listing/helpers.rs#L233-L247

wjones127 (changes requested):
I'll take a closer look tomorrow, but provided some initial comments. Thanks for working on this!

rust/src/operations/delete.rs
Blajda: Pruning is performed based on the predicate but I do agree if only partition columns exist in the predicate then we just do it in memory and would better than relying on pruning being correct.

wjones127: Ah I see the issue. We'll need to make some improvements to the DataFusion expressions so we can handle this better.

rust/src/operations/delete.rs
Blajda: The Datafusion implementation still uses recursion but overflow issues are now on them. The visitor pattern also makes the code look cleaner too!

rust/src/operations/delete.rs
Blajda: I implemented the solution to create a memory table and load it with the partition values. I did not map all possible types but if the current approach looks good I'll finish it up.

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,1074 @@
...
+                partition_columns: current_metadata.partition_columns.clone(),
+                volaility: Volatility::Immutable,
+                result: Ok(()),
+            };
+
+            TreeNode::visit(expr, &mut expr_properties)?;
+            expr_properties.result?;
+
+            if expr_properties.partition_only {
+                let scan_start = Instant::now();
```
wjones127, line 528:
```suggestion
            if expr_properties.partition_only {
                // If the expression only refers to partition columns, we can perform
                // the deletion just by removing entire files, so there is no need to
                // do an scan.
                let scan_start = Instant::now();
```

rust/src/operations/delete.rs
wjones127: Thanks, this looks pretty good!

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,1074 @@
...
+        let (table, metrics) = DeltaOps(table).delete().await.unwrap();
+        assert_eq!(table.version(), 2);
+        assert_eq!(metrics.num_added_files, 0);
+        assert_eq!(metrics.num_removed_files, 0);
+        assert_eq!(metrics.num_deleted_rows, None);
+        assert_eq!(metrics.num_copied_rows, None);
+    }
+
+    #[tokio::test]
+    async fn test_delete_on_nonpartiton_column() {
```
wjones127, line 912:
```suggestion
    async fn test_delete_on_nonpartition_column() {
```

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,1074 @@
...
+        // Delete based on a nonpartition column
```
wjones127, line 913: How are null values handled? For example, if I have a column `x: [1, 2, null, 4]`, does `DELETE FROM table_name WHERE x > 2` delete just the last row, or also the third row? Seems like it's worth a unit test to verify.

rust/src/operations/delete.rs
```diff
@@ -0,0 +1,1074 @@
...
+        DataType::Dictionary(_, value) => Ok(value.as_ref().to_owned()),
+        _ => Err(DeltaTableError::Generic(format!(
+            "Unable to create builder for arrow type {}",
+            t
+        ))),
+    }
+}
+
+// Create a record batch that contains the partition columns plus the path of the file
+fn create_partition_record_batch(
```
wjones127, line 722: Maybe you can re-use this function for now:

https://github.com/delta-io/delta-rs/blob/e046a774b608ac7df5114c014a1c22dbb60da38f/rust/src/table_state_arrow.rs#L51

It handles the different data types for partition values. It has a little more overhead since it also parses out the statistics, but I think that's fine for now. Later on, I expect we'll replace this with expression simplification which will let us use statistics and remove redundant parts of the predicate:

https://github.com/apache/arrow-datafusion/issues/6171

wjones127 (changes requested):
Thanks for the updates. I have a few more suggestions.

rust/src/operations/delete.rs
wjones127: > The Datafusion implementation still uses recursion but overflow issues are now on them

🤣

rust/src/operations/delete.rs
Blajda: Yeah great call out.
The previous commit would delete the row. I've added a new test and changed the behavior to not delete the third record in this case. 

A record should only be deleted if the predicate evaluates to true other it is kept. null > 2 evaluates to UNKNOWN.
I've checked the spark implementation and the behavior aligns

Blajda:
Hi @wjones127 @roeap 
I've made the requested changes. Let me know if it is good to merge or any additional concerns.

roeap (dismissed):
From my end this look great - thanks for this excellent contribution @Blajda.

I'll leave it open though for @wjones127 to chine in.

wjones127 (approved):

PR #1179, opened 2023-02-26
Implement pruning on partition columns

### LINKED ISSUE #1175: File pruning does not occur on partition columns
# Environment

**Delta-rs version**: Main

**Binding**: rust

***
# Bug

**What happened**:
When using Datafusion, statistics from the delta log are used to prune which files are scanned.
Delta logs do not contain statistics for partition columns in the regular stats columns but store the partition values in a separate key.

Partition information should be used while pruning. 

**How to reproduce it**:

This test which currently fails demonstrates that the scan visits all files for a partition that does not exist in the table.
The implementation also needs to consider null partitions are handled correctly.

```rust
#[tokio::test]
async fn test_datafusion_prune_partitioned() -> Result<()> {
    use datafusion::prelude::*;
    // Validate that partition information in include in table statistics
    let table = deltalake::open_table("./test/data/delta-2.2.0-partitioned-types")
        .await
        .unwrap();
    let statistics = table.datafusion_table_statistics();


    assert_eq!(statistics.num_rows, Some(3),);
    assert_eq!(statistics.total_byte_size, Some(452 * 3));

    async fn get_scan_metrics(table: &DeltaTable, state: &SessionState, e: Expr) -> Result<ExecutionMetricsCollector> {
        let mut metrics = ExecutionMetricsCollector::default();
        let scan = table.scan(state, None, &[e], None).await?;
        if scan.output_partitioning().partition_count() > 0 {
            let plan = CoalescePartitionsExec::new(scan);
            let task_ctx = Arc::new(TaskContext::from(state));
            let _result = collect(plan.execute(0, task_ctx)?).await?;
            visit_execution_plan(&plan, &mut metrics).unwrap();
        }

        return Ok(metrics);
    }

    let state = ctx.state();

    let e = col("c1").eq(lit(1));
    let metrics = get_scan_metrics(&table, &state, e).await?;
    println!("{:?}", metrics.scanned_files);
    assert!(metrics.num_scanned_files() == 0);

    let e = col("c1").eq(lit(4));
    let metrics = get_scan_metrics(&table, &state, e).await?;
    println!("{:?}", metrics.scanned_files);
    assert!(metrics.num_scanned_files() == 1);

    let e = col("c3").eq(lit(4));
    let metrics = get_scan_metrics(&table, &state, e).await?;
    println!("{:?}", metrics.scanned_files);
    assert!(metrics.num_scanned_files() == 1);

    let e = col("c3").eq(lit(0));
    let metrics = get_scan_metrics(&table, &state, e).await?;
    println!("{:?}", metrics.scanned_files);
    assert!(metrics.num_scanned_files() == 0);


    Ok(())
} 
```

Blajda:
# Description
Exposes partition columns in Datafusion's `PruningStatistics` which will reduce the number of files scanned when the table is queried.

This also resolves another partition issues where involving `null` partitions. Previously  `ScalarValue::Null` was used which would cause an error when the actual datatype was obtained from the physical parquet files.

# Related Issue(s)
- closes #1175

### DIFF
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -575,11 +575,9 @@ fn get_null_of_arrow_type(t: &ArrowDataType) -> ScalarValue {
         | ArrowDataType::Time64(_)
         | ArrowDataType::Duration(_)
         | ArrowDataType::Interval(_)
+        | ArrowDataType::RunEndEncoded(_, _)
         | ArrowDataType::Map(_, _) => {
-            panic!(
-                "{}",
-                format!("Implement data type for Delta Lake {}", t.to_string())
-            );
+            panic!("{}", format!("Implement data type for Delta Lake {}", t));
         }
     }
 }
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -57,7 +57,7 @@ use object_store::{path::Path, ObjectMeta};
 use url::Url;
 
 use crate::builder::ensure_table_uri;
-use crate::{action, open_table, open_table_with_storage_options};
+use crate::{action, open_table, open_table_with_storage_options, SchemaDataType};
 use crate::{schema, DeltaTableBuilder};
 use crate::{DeltaResult, Invariant};
 use crate::{DeltaTable, DeltaTableError};
@@ -241,6 +241,13 @@ fn get_prune_stats(table: &DeltaTable, column: &Column, get_max: bool) -> Option
         .get_schema()
         .ok()
         .map(|s| s.get_field_with_name(&column.name).ok())??;
+
+    // See issue 1214. Binary type does not support natural order which is required for Datafusion to prune
+    if let SchemaDataType::primitive(t) = &field.get_type() {
+        if t == "binary" {
+            return None;
+        }
+    }
 
     let data_type = field.get_type().try_into().ok()?;
     let partition_columns = &table.get_metadata().ok()?.partition_columns;
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -541,7 +541,6 @@ fn get_null_of_arrow_type(t: &ArrowDataType) -> ScalarValue {
         ArrowDataType::UInt16 => ScalarValue::UInt16(None),
         ArrowDataType::UInt32 => ScalarValue::UInt32(None),
         ArrowDataType::UInt64 => ScalarValue::UInt64(None),
-        ArrowDataType::Float16 => panic!("Unsupported type"),
         ArrowDataType::Float32 => ScalarValue::Float32(None),
         ArrowDataType::Float64 => ScalarValue::Float64(None),
         ArrowDataType::Date32 => ScalarValue::Date32(None),
@@ -553,9 +552,19 @@ fn get_null_of_arrow_type(t: &ArrowDataType) -> ScalarValue {
         ArrowDataType::LargeUtf8 => ScalarValue::LargeUtf8(None),
         ArrowDataType::Decimal128(precision, scale) => {
             ScalarValue::Decimal128(None, precision.to_owned(), scale.to_owned())
-        }
+        },
+        ArrowDataType::Timestamp(unit, tz) => {
+            let tz = tz.to_owned();
+            match unit {
+                TimeUnit::Second => ScalarValue::TimestampSecond(None, tz),
+                TimeUnit::Millisecond => ScalarValue::TimestampMillisecond(None, tz),
+                TimeUnit::Microsecond => ScalarValue::TimestampMicrosecond(None, tz),
+                TimeUnit::Nanosecond => ScalarValue::TimestampNanosecond(None, tz),
+            }
+        },
         //Unsupported types...
-        ArrowDataType::Decimal256(_, _)
+        ArrowDataType::Float16 
+        | ArrowDataType::Decimal256(_, _)
         | ArrowDataType::Union(_, _, _)
         | ArrowDataType::Dictionary(_, _)
         | ArrowDataType::LargeList(_)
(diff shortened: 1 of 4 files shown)


Blajda:
Looks like the windows build is flaking on the main branch. Not sure of the cause. If possible would be nice to rerun just that step.

wjones127 (changes requested):
Thanks for writing this @Blajda.

Would you be willing to add some more tests? I think it's worth making sure this is very robust :)

rust/tests/datafusion_test.rs
```diff
@@ -249,30 +272,57 @@ async fn test_files_scanned() -> Result<()> {
...
+    let metrics = get_scan_metrics(&table, &state, &[e]).await?;
+    assert!(metrics.num_scanned_files() == 1);
+    */
+
+    // Check pruning for null partitions
+    let e = col("k").is_null();
+    let metrics = get_scan_metrics(&table, &state, &[e]).await?;
+    assert!(metrics.num_scanned_files() == 1);
+
+    // Check pruning for null partitions. Since there are no record count statistics pruning cannot be done
```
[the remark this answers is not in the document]
Blajda: In Datafusion, null indicates that statistics for a particular column do not exist. Also null  counts are only used for the functions `is_null` and `is_not_null`. Further work is required to accommodate the edge case where all records for a columns are actually null. I imagine this can be accomplished by checking the null count statistic with a record count statistic.
I'd take a further look and raise an issue on their end.

rust/tests/datafusion_test.rs
```diff
@@ -228,8 +228,31 @@ async fn test_datafusion_stats() -> Result<()> {
     Ok(())
 }
 
+async fn get_scan_metrics(
```
roeap, line 231: not strictly required in this PR, but we have a very basic implementation for a metrics collector, that I plan to promote to a first class delta-rs member soon. The aim would be to harmonise our approach to metrics collection across operations. Just in case you also want to use that here :)

https://github.com/delta-io/delta-rs/blob/8dcf46e614370b93a6d47402456621b713768674/rust/tests/datafusion_test.rs#L39-L59

Blajda: Yeah I'm already using this implementation with a minor change if there are zero records. I found it really useful!

Blajda:
@wjones127 
Tests were updated to check all primitive types

There seems to be incomplete stats for `date` and `binary` since they do not have min and max stats written.

I think `decimal` stats are written in correctly. Currently the underlying i128 value is written were I was expecting the actual decimal string representation to be written.

Currently the following types cannot be used as partition column

| Type  | Error |
| --- | --- |
| float64 | Unimplemented data type: Float64', rust/src/writer/utils.rs:148:13 |
| float32 |  Unimplemented data type: Float32', rust/src/writer/utils.rs:148:13|
| timestamp |Unimplemented data type: Timestamp(Microsecond, None) |
| decimal | Unimplemented data type: Decimal128(10, 2)', rust/src/writer/utils.rs:148:13 |
| date | Unimplemented data type: Date32', rust/src/writer/utils.rs:148:13 |
| binary |called `Result::unwrap()` on an `Err` value: Arrow { source: InvalidArgumentError("The data type type Binary has no natural order") }', rust/src/operations/write.rs:318:30  |

rust/src/delta_datafusion.rs
```diff
@@ -510,6 +529,59 @@ impl ExecutionPlan for DeltaScan {
...
+        | ArrowDataType::Struct(_)
+        | ArrowDataType::List(_)
+        | ArrowDataType::FixedSizeList(_, _)
+        | ArrowDataType::Time32(_)
+        | ArrowDataType::Time64(_)
+        | ArrowDataType::Duration(_)
+        | ArrowDataType::Interval(_)
+        | ArrowDataType::RunEndEncoded(_, _)
+        | ArrowDataType::Map(_, _) => {
+            panic!("{}", format!("Implement data type for Delta Lake {}", t));
```
roeap, line 580: We usually try to be a panic-free crate, would it be possible to surface this error?

roeap (commented):
Thanks for this great work, especially the tests. Happy to see that some of the bigger holes in our datafusion story are being fixed 😀.

rust/src/delta_datafusion.rs
Blajda: It's unfortunately mostly used at an interface boundary with Datafusion that doesn't return a result. I removed the panic and restored the original behavior of returning a null scalar when it cannot be handled.

Blajda:
If there are no additional fixes required I would like to wait for #1180 to be merged first so I can check if the partition tests are resolved.

rust/tests/datafusion_test.rs
```diff
@@ -284,51 +289,376 @@ async fn test_datafusion_stats() -> Result<()> {
...
+            lit(8 as i8),
+            lit(3 as i8),
+        ),
+        (
+            "float64",
+            lit(1 as f64),
+            lit(5 as f64),
+            lit(8 as f64),
+            lit(3 as f64),
+        ),
```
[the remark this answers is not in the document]
wjones127: Need to replace the lines below too. The GH UI doesn't make it easy to select all of them :)

wjones127 (changes requested):
Thanks for adding all these tests :) Created #1209 and #1208 as follow ups.

Just one change I'd like to see to make the test a little more readable.

Blajda:
Rebased on main to retest on #1180. There are still some issues with time type columns. See #1215.
I also omitted pruning for binary columns. Datafusion cannot prune this type yet since a natural order does not exist. See #1214

rust/tests/datafusion_test.rs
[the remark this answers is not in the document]
Blajda: @wjones127 I've update the cases based on your outline. I agree it makes it a lot easier to read. Thanks for the suggestion :)

wjones127 (approved):
This looks excellent. Thanks for helping identify all those cases we need to fix. 😄

PR #1187, opened 2023-02-27
Enable passing Datafusion session state to WriteBuilder

### LINKED ISSUE #1186: Writing from a Delta table scan using WriteBuilder fails due to missing object store
# Environment

**Delta-rs version**: 0.7.0 (latest main)

**Binding**: Rust

**Environment**:
- **Cloud provider**: N/A
- **OS**: MacOS
- **Other**: 

***
# Bug

**What happened**:

When [performing a write](https://github.com/splitgraph/seafowl/blob/main/tests/statements/ddl.rs#L46-L55) into a delta table from a different delta table using the `WriteBuilder::with_input_execution_plan`, the write fails since the builder generates a new blank [session/task context](https://github.com/delta-io/delta-rs/blob/main/rust/src/operations/write.rs#L349-L350) to go along with the execution plan.

This in turn leads to the loss of the object store [registered](https://github.com/delta-io/delta-rs/blob/main/rust/src/delta_datafusion.rs#L358) for the scan previously, and consequently a failed write (`No suitable object store found for delta-rs://...`).

A proper task context should be supplied when executing the input plan.

**How to reproduce it**:

See the test here for an example: https://github.com/delta-io/delta-rs/pull/1187

**More details**:

gruuya:
# Description
The session state can keep some context pertinent to the [input plan](https://github.com/delta-io/delta-rs/blob/main/rust/src/operations/write.rs#L153-L157), so passing it allows for building useful `TaskContext`s for executing the plan.

This is especially the case when the input plan references another Delta table, which has previously registered it's object store (that needs to be accessible during the execution of the physical plan).

# Related Issue(s)
Closes #1186 

Curiously, if in the test I use `delta-0.8.0-partitioned` the final `SELECT` query fails with:
```
External(Execution("Failed to map column projection for field year. Incompatible data types Dictionary(UInt16, Utf8) and Utf8"))
```
I'm not sure yet why this happens, but I think it has nothing to do with this change.

# Documentation

### DIFF
--- a/rust/src/operations/write.rs
+++ b/rust/src/operations/write.rs
@@ -35,7 +35,7 @@ use crate::writer::utils::PartitionPath;
 
 use arrow::datatypes::{DataType, SchemaRef as ArrowSchemaRef};
 use arrow::record_batch::RecordBatch;
-use datafusion::execution::context::{SessionContext, TaskContext};
+use datafusion::execution::context::{SessionContext, SessionState, TaskContext};
 use datafusion::physical_plan::{memory::MemoryExec, ExecutionPlan};
 use futures::future::BoxFuture;
 use futures::StreamExt;
@@ -78,6 +78,8 @@ impl From<WriteError> for DeltaTableError {
 pub struct WriteBuilder {
     /// The input plan
     input: Option<Arc<dyn ExecutionPlan>>,
+    /// Datafusion session state relevant for executing the input plan
+    state: Option<SessionState>,
     /// Location where the table is stored
     location: Option<String>,
     /// SaveMode defines how to treat data already written to table location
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This looks sensible to me 👍

wjones127:
> I'm not sure yet why this happens, but I think it has nothing to do with this change.

Yeah that seems to be another issue. I created #1194 so we can look into it further.

roeap:
This is in fact sensible, one word of caution though. The conflict resolution requires us to gain some insights into the query execution that we may to collect as metrics.

So while we will have to manage sessions and plans, there is a good chance we to think more about the APIs.

PR #1206, opened 2023-03-03
feat: extend configuration handling

roeap:
# Description

This PR refactors and extends the table configuration handling. The approach is analogous to how we and object_store handle configuration via storage properties. The idea is to provide a somewhat typed layer over the untyped configuration keys.

There was one surprising thing along the way. From what I can tell, we may have been omitting the `delta.` prefix on the config keys we parse. So this would definitely be breaking behaviour, since we no longer recognize keys we were parsing before. We can in principle handle aliases for keys quite easily, but I was not sure what the desired behaviour is.

cc @rtyler @xianwill - This change would probably affect `kafka-delta-ingest`, so especially interested in your opinions! 

# Related Issue(s)

part of #632

# Documentation


houqp:
I am leaning towards us making the breaking change now and move on. Then come back to add the compatibility layer if someone complains.

houqp (approved):

PR #1207, opened 2023-03-04
feat: typed commit info

roeap:
# Description

Another PR on the road to #632 - ~~keeping it a draft, as it is based on #1206~~

While the `commitInfo` action is defined as completely optional, spark and delta-rs write at the very least interesting, but often also quite helpful information into the commit info. To make it easier to work with and centralize some conventions, we introduce a `CommitInfo` struct, that exposes some of the fields at the top level. Additionally we harmonize a bit between spark and delta-rs conventions.

# Related Issue(s)

part of #632 

# Documentation

### DIFF
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -590,25 +590,26 @@ impl DeltaOperation {
     }
 
     /// Parameters configured for operation.
-    pub fn operation_parameters(&self) -> DeltaResult<impl Iterator<Item = (String, Value)>> {
-        // TODO remove unwrap
-        let serialized = serde_json::to_value(self)
-            .map_err(|err| ActionError::SerializeOperation { source: err })?;
-        if let serde_json::Value::Object(map) = serialized {
-            let all_operation_fields = map.values().next().unwrap().as_object().unwrap().clone();
-            Ok(all_operation_fields
-                .into_iter()
+    pub fn operation_parameters(&self) -> DeltaResult<HashMap<String, Value>> {
+        if let Some(Some(Some(map))) = serde_json::to_value(self)
+            .map_err(|err| ActionError::SerializeOperation { source: err })?
+            .as_object()
+            .map(|p| p.values().next().map(|q| q.as_object()))
+        {
+            Ok(map
+                .iter()
                 .filter(|item| !item.1.is_null())
                 .map(|(k, v)| {
                     (
-                        k,
+                        k.to_owned(),
                         serde_json::Value::String(if v.is_string() {
                             String::from(v.as_str().unwrap())
                         } else {
                             v.to_string()
                         }),
                     )
-                }))
+                })
+                .collect())
         } else {
             Err(ActionError::Generic(
                 "Operation parameters serialized into unexpected shape".into(),
@@ -633,7 +634,7 @@ impl DeltaOperation {
         // TODO infer additional info from operation parameters ...
         CommitInfo {
             operation: Some(self.name().into()),
-            operation_parameters: self.operation_parameters().ok().map(|iter| iter.collect()),
+            operation_parameters: self.operation_parameters().ok(),
             ..Default::default()
         }
     }
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -768,7 +768,6 @@ mod tests {
         let info = serde_json::from_str::<CommitInfo>(raw);
         assert!(info.is_ok());
 
-
         // assert that commit info has no required filelds
         let raw = "{}";
         let info = serde_json::from_str::<CommitInfo>(raw);
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -449,9 +449,6 @@ pub struct Protocol {
 #[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct CommitInfo {
-    /// Version number the commit corresponds to
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub version: Option<DeltaDataTypeVersion>,
     /// Timestamp in millis when the commit was created
     #[serde(skip_serializing_if = "Option::is_none")]
     pub timestamp: Option<DeltaDataTypeTimestamp>,
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -589,7 +589,7 @@ impl DeltaOperation {
         }
     }
 
-    /// Paraemters configured for operation.
+    /// Parameters configured for operation.
     pub fn operation_parameters(&self) -> DeltaResult<impl Iterator<Item = (String, Value)>> {
         // TODO remove unwrap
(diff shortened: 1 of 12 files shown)


rust/src/action/mod.rs
```diff
@@ -517,45 +574,71 @@ pub enum DeltaOperation {
...
                         } else {
                             v.to_string()
                         }),
                     )
-                })
-                .collect();
+                }))
+        } else {
+            Err(ActionError::Generic(
+                "operation parameetrs serialized into unexpected shape".into(),
```
wjones127, line 614:
```suggestion
                "Operation parameters serialized into unexpected shape".into(),
```

rust/src/action/mod.rs
```diff
@@ -517,45 +574,71 @@ pub enum DeltaOperation {
...
-            let all_operation_fields = map.values().next().unwrap().as_object().unwrap();
-            let converted_operation_fields: Map<String, Value> = all_operation_fields
-                .iter()
+    /// Paraemters configured for operation.
+    pub fn operation_parameters(&self) -> DeltaResult<impl Iterator<Item = (String, Value)>> {
+        // TODO remove unwrap
+        let serialized = serde_json::to_value(self)
+            .map_err(|err| ActionError::SerializeOperation { source: err })?;
+        if let serde_json::Value::Object(map) = serialized {
+            let all_operation_fields = map.values().next().unwrap().as_object().unwrap().clone();
```
wjones127, line 598: Looks like that TODO still needs to be addressed.

rust/src/action/mod.rs
```diff
@@ -517,45 +574,71 @@ pub enum DeltaOperation {
...
+            DeltaOperation::Optimize { .. } => "OPTIMIZE",
+            DeltaOperation::FileSystemCheck { .. } => "FSCK",
+        }
+    }
 
-        if let Ok(serde_json::Value::Object(map)) = serde_json::to_value(self) {
-            let all_operation_fields = map.values().next().unwrap().as_object().unwrap();
-            let converted_operation_fields: Map<String, Value> = all_operation_fields
-                .iter()
+    /// Paraemters configured for operation.
```
wjones127, line 592:
```suggestion
    /// Parameters configured for operation.
```

rust/src/action/mod.rs
```diff
@@ -517,45 +574,71 @@ pub enum DeltaOperation {
...
+    }
 
-            commit_info.insert(
-                "operationParameters".to_string(),
-                serde_json::Value::Object(converted_operation_fields),
-            );
-        };
+    /// Denotes if the operation changes the data contained in the table
+    pub fn changes_data(&self) -> bool {
+        !matches!(self, Self::Optimize { .. })
```
wjones127, line 622: I wonder if it's worth us being verbose here so we don't forget to update this when implementing new operations:

```suggestion
        match self {
            Self::Optimize { .. } => false,
            Self::Create { .. }
            | Self::FileSystemCheck {}
            | Self::StreamingUpdate { .. }
            | Self::Write { .. } => true,
        }
```

rust/src/action/mod.rs
```diff
@@ -517,45 +574,71 @@ pub enum DeltaOperation {
...
+        // TODO infer additional info from operation parameters ...
+        CommitInfo {
+            operation: Some(self.name().into()),
+            operation_parameters: self.operation_parameters().ok().map(|iter| iter.collect()),
+            ..Default::default()
+        }
+    }
 
-        commit_info
+    /// Get predicate expression applien when the operation reads data from the table.
```
wjones127, line 635:
```suggestion
    /// Get predicate expression applied when the operation reads data from the table.
```

rust/src/action/mod.rs
```diff
@@ -654,4 +737,65 @@ mod tests {
...
+                "numOutputBytes": "1356"
+            },
+            "engineInfo": "Apache-Spark/3.3.1 Delta-Lake/2.2.0",
+            "txnId": "046a258f-45e3-4657-b0bf-abfb0f76681c"
+        }"#;
+
+        let info = serde_json::from_str::<CommitInfo>(raw);
+        assert!(info.is_ok());
+
+        println!("{:?}", info);
```
wjones127, line 765: Can we drop this?
```suggestion
```

rust/src/operations/transaction.rs
```diff
@@ -62,34 +62,41 @@ fn log_entry_from_actions(actions: &[Action]) -> Result<String, TransactionError
...
+    Ok(bytes::Bytes::from(log_entry_from_actions(actions)?))
+}
+
+/// Low-level transaction API. Creates a temporary commit file. Once created,
+/// the transaction object could be dropped and the actual commit could be executed
+/// with `DeltaTable.try_commit_transaction`.
+pub(crate) async fn prepare_commit(
+    storage: &dyn ObjectStore,
+    operation: &DeltaOperation,
+    actions: &mut Vec<Action>,
```
wjones127, line 95: Why is this changed to a `&mut` instead of just an owned value?

wjones127 (changes requested):
This looks like a good improvement. It seems unfortunate the format of commitInfo isn't standardized in the protocol.

rust/src/action/mod.rs
```diff
@@ -435,7 +443,46 @@ pub struct Protocol {
...
-type CommitInfo = Map<String, Value>;
+/// The commitInfo is a fairly flexible action within the delta specification, where arbitrary data can be stored.
+/// However the reference implementation as well as delta-rs store useful information that may for instance
+/// allow us to be more permissive in commit conflict resolution.
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CommitInfo {
+    /// Version number the commit corresponds to
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub version: Option<DeltaDataTypeVersion>,
```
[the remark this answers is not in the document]
roeap: It is, and nobody actually writes it to the commit info in the log. Initially I was following the spark implementation for the optimistic commits quite closely, where this field is being used. Not sure if we still do. I'll remove it for now, and bring it back in case its needed in the follow up PR...

rust/src/operations/transaction.rs
roeap: I extracted this from the optimistic commit PR where we are using this in a loop and need access to the actions afterwards. That said, having this mutable in that scenario is likely not the best idea, and we need it only to add the commit info. Changed is so we just take a borrowed value now.

rust/src/action/mod.rs
```diff
@@ -435,7 +443,46 @@ pub struct Protocol {
...
+    /// Timestamp in millis when the commit was created
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub timestamp: Option<DeltaDataTypeTimestamp>,
```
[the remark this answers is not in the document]
roeap: Absolutely, we may also be able to implement custom serde, so it is parsed as a datetime.

wjones127 (approved):

Issue #1219, opened 2023-03-12
Python: overwrite partition doesn't seem to work with multiple conditions


wjones127:
# Environment

**Delta-rs version**: 0.8.0

**Binding**: Python

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

**What happened**:

```python
from datetime import date

import numpy as np
import numpy.random
import pandas as pd
from deltalake import DeltaTable, write_deltalake

nrows = 9
data = pd.DataFrame(
    {
        "year": np.repeat(2023, nrows),
        "month": np.repeat([1, 2, 3], nrows / 3),
        "values": numpy.random.normal(size=nrows),
    }
)
table_path = "tables/monthly"
write_deltalake(
    table_path,
    data,
    partition_by=["year", "month"],
    mode="overwrite",
)

nrows = 4
new_data = pd.DataFrame(
    {
        "year": np.repeat(2023, nrows),
        "month": np.repeat([4], nrows),
        "values": numpy.random.normal(size=nrows) + 10,
    }
)

write_deltalake(
    table_path,
    new_data,
    mode="overwrite",
    partition_filters=[("year", "=", "2023"), ("month", "=", "4")],
)
DeltaTable(table_path).to_pandas()
```

Produces the error

```
ValueError: Data should be aligned with partitioning. Partition 'year'='2023' should be filtered out from data.
```

**What you expected to happen**:

It should overwrite the partition with `year=2023` and `month=4`.

**How to reproduce it**:

**More details**:

PR #1221, opened 2023-03-12
improve err msg on use of non-partitioned column

### LINKED ISSUE #1218: Support partition filters on date columns
# Description

When I tried filtering with the partition filters on date columns, I got the error:

```
Invalid partition filter found: [PartitionFilter { key: "observation_date", value: Equal("2023-01-03") }].
```

We should make sure to support this. This is especially important now that we have overwrite partition in Python.

marijncv:
# Description
Updates the error message when trying to use a partition filter with a column that is not partitioned. 

Code to trigger the error (adapted from #1219):
```python
from datetime import date

import numpy as np
import numpy.random
import pandas as pd

nrows = 9
data = pd.DataFrame(
    {
        "year": np.repeat(2023, nrows),
        "month": np.repeat([1, 2, 3], nrows / 3),
        "values": numpy.random.normal(size=nrows),
    }
)
table_path = "tables/monthly"
write_deltalake(
    table_path,
    data,
    partition_by=["year", "month"],
    mode="overwrite",
)

nrows = 4
new_data = pd.DataFrame(
    {
        "year": np.repeat(2023, nrows),
        "month": np.repeat([4], nrows),
        "values": numpy.random.normal(size=nrows) + 10,
    }
)

write_deltalake(
    table_path,
    new_data,
    mode="overwrite",
    partition_filters=[("values", "=", "1")],
)
DeltaTable(table_path).to_pandas()
```

# Related Issue(s)
closes #1218 

# Documentation

### DIFF
--- a/rust/src/table_state.rs
+++ b/rust/src/table_state.rs
@@ -369,13 +369,16 @@ impl DeltaTableState {
     ) -> Result<impl Iterator<Item = &'a Add> + '_, DeltaTableError> {
         let current_metadata = self.current_metadata().ok_or(DeltaTableError::NoMetadata)?;
 
-        for f in filters {
-            if !current_metadata.partition_columns.contains(&f.key.into()) {
-                let column = f.key.to_string();
-                return Err(DeltaTableError::ColumnNotPartitioned {
-                    column: format!("{column:?}"),
-                });
-            }
+        let nonpartitioned_columns: Vec<String> = filters
+            .iter()
+            .filter(|f| !current_metadata.partition_columns.contains(&f.key.into()))
+            .map(|f| f.key.to_string())
+            .collect();
+
+        if !nonpartitioned_columns.is_empty() {
+            return Err(DeltaTableError::ColumnsNotPartitioned {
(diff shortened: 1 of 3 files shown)


rust/src/table_state.rs
```diff
@@ -368,13 +368,14 @@ impl DeltaTableState {
...
-            .all(|f| current_metadata.partition_columns.contains(&f.key.into()))
-        {
-            return Err(DeltaTableError::InvalidPartitionFilter {
-                partition_filter: format!("{filters:?}"),
-            });
+
+        for f in filters {
+            if !current_metadata.partition_columns.contains(&f.key.into()) {
+                let column = f.key.to_string();
+                return Err(DeltaTableError::ColumnNotPartitioned {
```
houqp, line 375: nitpick, the original implementation returns all columns that aren't partitioned, but this implementation does an early return, which results in a different behavior.

marijncv: Thanks @houqp, good point! I tried to restore to the original behavior in the latest commit

wjones127 (approved):
Sorry it took a bit to get to reading this. Looks good :)

PR #1226, opened 2023-03-16
Introduce a failing test for writing with map columns and partitions

rtyler:
# Description

This test will fail with the following error:

```
---- writer::record_batch::tests::test_divide_record_batch_with_map_single_partition stdout ----
thread 'writer::record_batch::tests::test_divide_record_batch_with_map_single_partition' panicked at 'not implemented: Take not supported for data type Map(Field { name: "key_value", data_type: Struct([Field { name: "key", data_type: Utf8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "value", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, false)', /home/tyler/.cargo/registry/src/github.com-1ecc6299db9ec823/arrow-select-33.0.0/src/take.rs:234:14
```

As far as I can tell this _should_ be supported behavior and a supported schema. Note that with no partition columns the write works just fine.

=# Related Issue(s)

https://github.com/apache/arrow-rs/issues/3875


rtyler:
The problem appears to be the `take` for `batch_data` in the `divide_by_partition_values` function. That kernel is not supported by Arrow, and I think the function can be fixed to avoid that for Map data but RecordBatches are sufficiently difficult to work with that I haven't yet landed on a more succinct test.

wjones127:
I created a PR to add the kernel upstream.

rtyler:
Now looking at these clippy errors :unamused:

rtyler:
This will be addressed by #1249 , so once that pull request is ready, this test should be merged into that branch

python/tests/test_schema.py
```diff
@@ -9,7 +11,7 @@ def test_table_schema():
     table_path = "../rust/tests/data/simple_table"
     dt = DeltaTable(table_path)
     schema = dt.schema()
-    assert schema.json() == {
+    assert json.loads(schema.to_json()) == {
```
wjones127, line 14: I'm confused. This seems like an API-breaking change? What happened?

rust/src/writer/utils.rs
```diff
@@ -114,12 +111,27 @@ pub(crate) fn next_data_path(
...
     message_buffer: &[Value],
-) -> Result<RecordBatch, DeltaTableError> {
-    let mut value_iter = message_buffer.iter().map(|j| Ok(j.to_owned()));
-    let options = DecoderOptions::new().with_batch_size(message_buffer.len());
-    let decoder = Decoder::new(arrow_schema, options);
+) -> DeltaResult<RecordBatch> {
+    let mut buf = vec![];
+    for message in message_buffer {
+        buf.write_all(
+            serde_json::to_string(&message)
```
wjones127, line 118: They are making you serialize the messages?! Ouch...

wjones127: Did you consider this API? https://docs.rs/arrow-json/39.0.0/arrow_json/reader/struct.Decoder.html#method.serialize

That's what Raphael built for the use case you were asking about, I think.

rust/src/checkpoints.rs
```diff
@@ -388,12 +390,28 @@ fn parquet_bytes_from_state(state: &DeltaTableState) -> Result<bytes::Bytes, Che
...
+    let mut decoder = ReaderBuilder::new(arrow_schema)
+        .with_batch_size(CHECKPOINT_RECORD_BATCH_SIZE)
+        .build_decoder()?;
+
+    let mut buf = vec![];
+    for res in jsons {
+        let json = res?;
+        buf.write_all(serde_json::to_string(&json)?.as_bytes())?;
+    }
+    let mut consumed = 0;
```
wjones127, line 402: Same thing here, right? we shouldn't need to serialize to json string if in the end we are trying to make a parquet file?

wjones127 (approved):

PR #1242, opened 2023-03-25
Add Max Partitions Arg to Write

ColeMurray:
# Description

Add optional max partitions arg to be passed to pyArrow dataset.write_dataset.

# Related Issue(s)
This resolves an issue where users are unable to write more than 1024 partitions and have no way of overriding the parameter

```
  File "/lib/python3.8/site-packages/deltalake/writer.py", line 293, in write_deltalake
    ds.write_dataset(
  File "/lib/python3.8/site-packages/pyarrow/dataset.py", line 900, in write_dataset
    _filesystemdataset_write(
  File "pyarrow/_dataset.pyx", line 2479, in pyarrow._dataset._filesystemdataset_write
  File "pyarrow/error.pxi", line 99, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Fragment would be written into 1246 partitions. This exceeds the maximum of 1024
```

# Documentation

https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -77,6 +77,7 @@ def write_deltalake(
     filesystem: Optional[pa_fs.FileSystem] = None,
     mode: Literal["error", "append", "overwrite", "ignore"] = "error",
     file_options: Optional[ds.ParquetFileWriteOptions] = None,
+    max_partitions: Optional[int] = None,
     max_open_files: int = 1024,
     max_rows_per_file: int = 10 * 1024 * 1024,
     min_rows_per_group: int = 64 * 1024,
@@ -87,7 +88,6 @@ def write_deltalake(
     overwrite_schema: bool = False,
     storage_options: Optional[Dict[str, str]] = None,
     partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
-    max_partitions: Optional[int] = None,
 ) -> None:
     """Write to a Delta Lake table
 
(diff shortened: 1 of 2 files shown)


python/deltalake/writer.py
```diff
@@ -87,6 +87,7 @@ def write_deltalake(
     overwrite_schema: bool = False,
     storage_options: Optional[Dict[str, str]] = None,
     partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
+    max_partitions: Optional[int] = None,
```
wjones127, line 90: Could you place this up near `max_open_files`, since it logically belongs with those options? Since we have the `*`, changing the order of the kwargs doesn't break backwards compatibility.

python/deltalake/writer.py
```diff
@@ -135,6 +136,7 @@ def write_deltalake(
     :param overwrite_schema: If True, allows updating the schema of the table.
     :param storage_options: options passed to the native delta filesystem. Unused if 'filesystem' is defined.
     :param partition_filters: the partition filters that will be used for partition overwrite.
+    :param max_partitions: the maximum number of partitions that will be used
```
wjones127, line 139: Again, also good to place near other similar options.

wjones127 (dismissed):
Thanks for this contribution @ColeMurray. Could you add a unit test. This could just set max_partitions=1 and verify the standard sample dataset fails.

python/deltalake/writer.py
ColeMurray: Yes, my concern was around the backwards compat, didn't see the `*`.
Updated

wjones127 (approved):
Looks great. Thanks @ColeMurray!

PR #1243, opened 2023-03-26
feat: add package version

### LINKED ISSUE #1227: Add version to deltalake python package
# Description
Small task of adding a `__version__` to the init.py of deltalake python package.

Example of how `pyarrow` does it is available [here](https://github.com/apache/arrow/blob/main/python/pyarrow/__init__.py#L43).
This is a low priority easy additional and I am happy to raise a PR for this.

**Use Case**

 This will allow users to,
```
import deltalake as dl
print(dl.__version__)
'0.8.1'
```

**Related Issue(s)**

wjones127:
# Description

Low maintenance way to expose package version. It will come from the `cargo.toml`.

Note that we have a separate function already called `rust_core_version` that provides the version of the `deltalake` Rust crate. That currently matches the Python library, but I don't think we provide any guarantee it will continue to.

# Related Issue(s)

For example:

- closes #1227 

# Documentation

### DIFF
--- a/python/deltalake/__init__.py
+++ b/python/deltalake/__init__.py
@@ -1,4 +1,4 @@
-from ._internal import PyDeltaTableError, RawDeltaTable, rust_core_version
+from ._internal import PyDeltaTableError, RawDeltaTable, __version__, rust_core_version
 from .data_catalog import DataCatalog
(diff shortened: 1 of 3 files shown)


roeap (approved):
Thanks!

PR #1244, opened 2023-03-26
fix: use native-tls for python deltalake releases

### LINKED ISSUE #1224: WARN  Could not load all certificates - BadDER
# Environment

**Delta-rs version**: 0.8.0

**Binding**:  Python (3.10.8)

**Environment**:
- **Cloud provider**: AWS
- **OS**: OSX
- **Other**:

***
# Bug

**What happened**:
After [upgrading](https://github.com/aws/aws-sdk-pandas/pull/2110) to 0.8.0, all operations display the following warning:
```
[2023-03-15T09:59:13Z WARN  hyper_rustls::connector] Could not load all certificates: Custom { kind: InvalidData, error: Custom { kind: InvalidData, error: BadDER } }
```

Looks like deltalake is not able to load certificates in the native OSX truststore. I do have a few unsigned/self-signed certs in my truststore so that might be it, but in that case I'd expect a message saying it's not able to validate it, not "bad" cert.

**How to reproduce it**:

**More details**:

wjones127:
# Description

We changed default to make our workspace harmonious, but didn't mean to change this for the Python release.

# Related Issue(s)

- closes #1224

# Documentation

### DIFF
--- a/.github/workflows/python_release.yml
+++ b/.github/workflows/python_release.yml
@@ -8,6 +8,11 @@ defaults:
   run:
     working-directory: ./python
 
+env:
+  # For ease of development, we make rustls default. But for release we should
+  # use native TLS.
+  FEATURES_FLAG: --no-default-features --features native-tls
+
 jobs:
   validate-release-tag:
     name: Validate git tag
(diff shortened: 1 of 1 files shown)


houqp (approved):

houqp:
Would be nice to add a regression test for this, am a little bit concerned that our dev environment is always different from the production environment.

PR #1245, opened 2023-03-26
WIP: failing test for spaces in table path

### LINKED ISSUE #1189: Fail to read delta table on mounted disk
# Environment

**Delta-rs version**: 
0.7.0 (latest main)

**Binding**: 
Python

**Environment**:
- **Cloud provider**: N/A (mounted server disk)
- **OS**: *NIX
- **Other**:

***
# Bug

**What happened**:
I'm trying to read a delta-table, that located on a local server.

If we run the following code:
```[python]
path = 'file:///mnt/path/to/table/'
DeltaTable(path).to_pyarrow_table()
```
We will get the following error:
<details>
<summary>Traceback</summary>

```shell
  File "/path/to/main.py", line 2, in <module>
    dt.to_pyarrow_table()
  File "/path/to/deltalake/table.py", line 401, in to_pyarrow_table
  File "/path/to/deltalake/fs.py", line 22, in open_input_file
deltalake.PyDeltaTableError: Object at location /mnt/network_expansion/path/to/table/part-00000-aaaaaa-1111-2222-bbbbb-333333333.c000.snappy.parquet not found: No such file or directory (os error 2)
```
The `*.parquet` file is present.

</details>

Succeful read

**How to reproduce it**:

**More details**:

Worth noting:
1. If we run similar code, but written in Rust, it will work just fine.
2. The full network path contains characters that get URL-encoded.
3. If we try to save schema (`file://`) in url-path, as suggested in [this](https://github.com/delta-io/delta-rs/issues/1157) issue, the code will do some more progress (it will find and read all `*.parquet` files), but will fail with a SIGABRT.

From (1) I'm taking a guess, that there's a broken logic either somewhere in deltalake python wrapper, because (looking at the traceback):
`File "pyarrow/_fs.pyx", line 1551, in pyarrow._fs._cb_open_input_file` is a call from PyArrow to a `SystemHandler`, which is `DeltaFileSystemHandler`

wjones127:
# Description

# Related Issue(s)

- closes #1189

# Documentation

### DIFF
--- a/rust/src/builder.rs
+++ b/rust/src/builder.rs
@@ -389,7 +389,10 @@ mod tests {
         assert_eq!("memory://", uri.as_str());
         let uri = ensure_table_uri("s3://tests/data/delta-0.8.0/").unwrap();
         assert_eq!("s3://tests/data/delta-0.8.0", uri.as_str());
-        let _uri = ensure_table_uri("s3://tests/data/delta-0.8.0//").unwrap();
-        assert_eq!("s3://tests/data/delta-0.8.0", uri.as_str())
+        let uri = ensure_table_uri("s3://tests/data/delta-0.8.0//").unwrap();
+        assert_eq!("s3://tests/data/delta-0.8.0", uri.as_str());
+
+        let uri = ensure_table_uri("file://test path/").unwrap();
+        assert_eq!("file://test path/", uri.as_str());
     }
 }
--- a/python/src/filesystem.rs
+++ b/python/src/filesystem.rs
@@ -46,8 +46,14 @@ impl DeltaFileSystemHandler {
         })
     }
 
-    fn __repr__<'py>(self_: PyRef<Self>, py: Python<'py>) -> PyResult<String> {
-        // let self_: PyRef<Self> = py_self.extract(py)?;
+    fn __repr__(self_: PyRef<Self>) -> PyResult<String> {
+        let pyself = self_.clone().into_py(self_.py());
+        let class_name: String = pyself
+            .as_ref(self_.py())
+            .getattr("__class__")?
+            .getattr("__name__")?
+            .extract()?;
+
         let maybe_options = if self_.config.options.is_empty() {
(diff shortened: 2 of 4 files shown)


rust/src/builder.rs
```diff
@@ -389,7 +389,10 @@ mod tests {
         assert_eq!("memory://", uri.as_str());
         let uri = ensure_table_uri("s3://tests/data/delta-0.8.0/").unwrap();
         assert_eq!("s3://tests/data/delta-0.8.0", uri.as_str());
-        let _uri = ensure_table_uri("s3://tests/data/delta-0.8.0//").unwrap();
-        assert_eq!("s3://tests/data/delta-0.8.0", uri.as_str())
+        let uri = ensure_table_uri("s3://tests/data/delta-0.8.0//").unwrap();
+        assert_eq!("s3://tests/data/delta-0.8.0", uri.as_str());
+
+        let uri = ensure_table_uri("file://test path/").unwrap();
+        assert_eq!("file://test path/", uri.as_str());
```
rtyler, line 396: @wjones127 These URIs are not valid file URLs to begin with and I believe that the test failure is a misnomer. According to [very official and always correct sources](https://en.wikipedia.org/wiki/File_URI_scheme#Unix) three slashes are required.

I checked this topic branch out, merged the latest `main` with it and ran the tests with three slashes.  The assertion does fail when I use `file:///tmp/test path`

```
thread 'builder::tests::test_ensure_table_uri' panicked at 'assertion failed: `(left == right)`
  left: `"file:///tmp/test path/"`,
 right: `"file:///tmp/test%20path"`', rust/src/builder.rs:450:9
```

_However_, the `/tmp/test path/` directory does exist correctly on the file system, having been created by the code which lives in `ensure_table_uri`.

I can push my merge of `main` if you'd like.

I think this test _may_ not be correct and we may have fixed the spaces-in-table-names issues since the rest of the assertions pass with the latest `HEAD` of `main`

PR #1249, opened 2023-03-27
Upgrade datafusion to 22 which brings arrow upgrades with it

rtyler:
Both @roeap and I have upgraded to datafusion 22 and arrow 36 in this pull request. We're not yet ready to adopt arrow 37 unfortunately

### DIFF
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -18,7 +18,7 @@ doc = false
 name = "deltalake._internal"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


rtyler:
The deprecation of `Decoder` is going to present some problems. I think I have found a bug in the new `RawReaderBuilder` and how it might handle the schema of checkpoint files upstream. Trying to give them a reproduction case :pleading_face:

wjones127 (approved):
LGTM!

PR #1253, opened 2023-03-28
Fixed Documentation for `get_add_actions` function

JHibbard:
Fixed Documentation for `get_add_actions` function

# Description
The code example uses `get_add_actions_df` as the method name instead of the correct `get_add_actions` method name.

# Related Issue(s)

None

# Documentation

The [[examples](https://delta-io.github.io/delta-rs/python/api_reference.html#deltalake.table.DeltaTable.get_add_actions)] section contains the incorrect method name.


wjones127 (approved):
Thanks for catching that @JHibbard

PR #1258, opened 2023-04-03
Fix names and nullability when creating RecordBatch from MapArray

### LINKED ISSUE #1257: Not matching column names when creating a RecordBatch from MapArray
# Environment

**Delta-rs version**: 0.8.0

**Binding**: Rust

**Environment**:
- **Cloud provider**: N/A
- **OS**: Linux
- **Other**: RustExplorer ( https://www.rustexplorer.com/b/o7bfm4 )

***
# Bug

**What happened**: When trying to create a RecordBatch with MapArray, I get the following error:
```
thread 'main' panicked at 'Failed to create RecordBatch: InvalidArgumentError("column types must match schema types, expected Map(Field { name: \"key_value\", data_type: Struct([Field { name: \"key\", data_type: Utf8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: \"value\", data_type: Binary, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }]), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, false) but found Map(Field { name: \"entries\", data_type: Struct([Field { name: \"keys\", data_type: Utf8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: \"values\", data_type: Binary, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }]), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, false) at column index 0")', src/main.rs:66:90
```

**What you expected to happen**: Successful creation of `RecordBatch` without any errors.

**How to reproduce it**: Create a `MapArray` and `RecordBatch` with it (See rust explorer link above).

**More details**:
I believe these lines are at fault. https://github.com/delta-io/delta-rs/blob/main/rust/src/delta_arrow.rs#L145-L161
Hopefully I will be able to test this hypothesis and make a PR in few hours.

balbok0:
# Description
When creating a RecordBatch with one of the columns being a MapArray, there are issues with naming of sub-columns, as well as nullability of the "elements/key_value" sub-column.

# Related Issue(s)

- closes #1257

# Documentation

See https://www.rustexplorer.com/b/o7bfm4 for a breaking example.
I initially thought that the issue is specifically for array with 0-length maps, but `nullable` mismatch happens even when all of the map elements have values.

### DIFF
--- a/rust/src/delta_arrow.rs
+++ b/rust/src/delta_arrow.rs
@@ -754,7 +754,7 @@ mod tests {
 
     #[test]
     fn test_delta_from_arrow_map_type() {
-        let timestamp_field = ArrowDataType::Map(
+        let arrow_map = ArrowDataType::Map(
             Box::new(ArrowField::new(
                 "key_value",
                 ArrowDataType::Struct(vec![
@@ -765,8 +765,10 @@ mod tests {
             )),
             false,
         );
+        let converted_map: crate::SchemaDataType = (&arrow_map).try_into().unwrap();
+
         assert_eq!(
-            <crate::SchemaDataType as TryFrom<&ArrowDataType>>::try_from(&timestamp_field).unwrap(),
+            converted_map,
             crate::SchemaDataType::map(crate::SchemaTypeMap::new(
                 Box::new(crate::SchemaDataType::primitive("byte".to_string())),
                 Box::new(crate::SchemaDataType::primitive("binary".to_string())),
--- a/rust/src/delta_arrow.rs
+++ b/rust/src/delta_arrow.rs
@@ -142,24 +142,24 @@ impl TryFrom<&schema::SchemaDataType> for ArrowDataType {
             }
             schema::SchemaDataType::map(m) => Ok(ArrowDataType::Map(
                 Box::new(ArrowField::new(
-                    "key_value",
+                    "entries",
                     ArrowDataType::Struct(vec![
                         ArrowField::new(
-                            "key",
+                            "keys",
                             <ArrowDataType as TryFrom<&schema::SchemaDataType>>::try_from(
                                 m.get_key_type(),
                             )?,
                             false,
                         ),
                         ArrowField::new(
-                            "value",
+                            "values",
                             <ArrowDataType as TryFrom<&schema::SchemaDataType>>::try_from(
                                 m.get_value_type(),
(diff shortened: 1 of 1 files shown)


rtyler:
Thanks for the pull request! @balbok0 would you be able to take your failing example and turn that into a unit or integration test so we don't have this regression in the future?

wjones127:
I’d like for us to fix this, but I also hope we can find a way to simultaneously support older PyArrow versions in the Python bindings.

balbok0:
@wjones127 are there some specific pyarrow versions you have in mind? I wasn't able to find a specific arrow-cpp/pyarrow version that changed the naming scheme.

balbok0:
Added failing example as a test.

rust/src/delta_arrow.rs
```diff
@@ -751,6 +752,71 @@ mod tests {
...
+                ArrowDataType::Struct(vec![
+                    ArrowField::new("key", ArrowDataType::Int8, false),
+                    ArrowField::new("value", ArrowDataType::Binary, true),
+                ]),
+                true,
+            )),
+            false,
+        );
+        assert_eq!(
+            <crate::SchemaDataType as TryFrom<&ArrowDataType>>::try_from(&timestamp_field).unwrap(),
```
wjones127, line 769:
```suggestion
        );
        let converted_map: crate::SchemaDataType = (&timestamp_field).try_into().unwrap();
        assert_eq!(
            converted_map,
```

wjones127 (changes requested):
This generally looks good. I was worried about PyArrow and other languages because the Arrow standard specifies field names as "key" not "keys" and "value" not "values" (arrow-rs deviates from this). But the PyArrow C Data Interface import actually ignores these field names on map types anyways, so it shouldn't matter that they are different. So we can ignore that worry.

I have a few small nits on the tests, but otherwise looks good to go.

wjones127:
Oh and could you rebase? We just fixed some of the deprecation warnings.

balbok0:
Thanks for comments! I applied changes and rebased. Lmk if there is anything else that needs to done before merge.

wjones127 (approved):

PR #1266, opened 2023-04-08
feat: read schema from parquet files in datafusion scans

### LINKED ISSUE #441: Datafusion table provider: issues with  timestamp types
In a recent attempt to write a where clause on a time stamp column in one of my favorite delta tables, I was not able to express that condition in delta-rs / datafusion. I believe this is due to a mismatch between the logical field type recorded in the delta log (and delivered to datafusion by delta-rs) and the expectation of datafusion to find the actual physical column type in the schema.

Concretely, given a delta table with a column that, according to (java) parquet tools, has the following type: 
```
optional int64 endOfDsegTime (TIMESTAMP_MILLIS);
```

and according to delta-rs, the type: 
```
SchemaField {
	name: "endOfDsegTime",
	type: primitive(
		"timestamp",
	),
	nullable: true,
	metadata: {},
},
```

When trying to use that field in a datafusion query `endOfDsegTime >= TO_TIMESTAMP_MILLIS(\'2021-03-19T00:00:00\')`, I see: `Error during planning: 'Timestamp(Nanosecond, None) >= Timestamp(Millisecond, None)' can't be evaluated because there isn't a common type to coerce the types to`, so datafusion seems to think that `endOfDsegTime` is in nanoseconds precision, which it isn't.

If I rewrite the where clause to `endOfDsegTime >= TO_TIMESTAMP(\'2021-03-19T00:00:00\')`, datafusion accepts the where clause but then fails somewhere deep in arrow compute territory:way to distinguish the different timestamp types from the metadata alone.

```
Cannot evaluate binary expression GtEq with types Timestamp(Millisecond, None) and Timestamp(Nanosecond, None)
```

I rewrote the table to int96 / nanoseconds; In that scenario, the second where clause works. I also noticed that according to delta-rs (and also, `_delta_log/0.json`), the schema is identical between these two versions of the table - delta does not distinguish based on the different physical parquet type, only in parquet the actual concrete type manifests.

Notes:
- not sure how spark does it, but there "everything just works" irrespective of the column type in parquet
- reading either of the two delta tables using `datafusion::datasource::parquet` also just works, because the schema derived from the datasource is based on the parquet file, not on the meta data in the delta log
- I'm not sure this can be solved solely on delta side - it seems that there's no way to distinguish the different timestamp types from the metadata alone.

roeap:
# Description

This PR updates table scans with datafusion to read the file schema from the parquet file within the latest add action of the table. This is to work around some issues, where the schema we derive from metadata does not match the data in the parquet files - e.g. nanosecond timestamps vs. micorsoecond.

We also update the `Load` command to handle column selections and make it more consistent with the other operations.

# Related Issue(s)

closes #441

# Documentation

### DIFF
--- a/rust/src/operations/transaction/state.rs
+++ b/rust/src/operations/transaction/state.rs
@@ -107,9 +107,8 @@ impl DeltaTableState {
             .schema()
             .clone();
 
-        let schema = self.arrow_schema()?;
         let table_schema = Arc::new(ArrowSchema::new(
-            schema
+            self.arrow_schema()?
                 .fields
                 .clone()
                 .into_iter()
--- a/rust/src/operations/transaction/state.rs
+++ b/rust/src/operations/transaction/state.rs
@@ -87,7 +87,7 @@ impl DeltaTableState {
 
     /// Get the pysical table schema.
     ///
-    /// This will construct a schema derived from the parqet schema of the latest data file,
+    /// This will construct a schema derived from the parquet schema of the latest data file,
     /// and fields for partition columns from the schema defined in table meta data.
     pub async fn physical_arrow_schema(
         &self,
--- a/rust/src/operations/transaction/state.rs
+++ b/rust/src/operations/transaction/state.rs
@@ -93,35 +93,33 @@ impl DeltaTableState {
         &self,
         object_store: Arc<dyn ObjectStore>,
     ) -> DeltaResult<ArrowSchemaRef> {
-        let file_meta = self
-            .files()
-            .iter()
-            .max_by_key(|obj| obj.modification_time)
-            .ok_or(DeltaTableError::Generic("No active file actions to get physical schema. Maybe the current state has not yet been loaded?".into()))?
-            .try_into()?;
+        if let Some(add) = self.files().iter().max_by_key(|obj| obj.modification_time) {
+            let file_meta = add.try_into()?;
+            let file_reader = ParquetObjectReader::new(object_store, file_meta);
+            let file_schema = ParquetRecordBatchStreamBuilder::new(file_reader)
+                .await?
+                .build()?
+                .schema()
+                .clone();
 
-        let file_reader = ParquetObjectReader::new(object_store, file_meta);
-        let file_schema = ParquetRecordBatchStreamBuilder::new(file_reader)
-            .await?
-            .build()?
-            .schema()
-            .clone();
+            let table_schema = Arc::new(ArrowSchema::new(
+                self.arrow_schema()?
+                    .fields
+                    .clone()
+                    .into_iter()
+                    .map(|field| {
+                        file_schema
+                            .field_with_name(field.name())
+                            .cloned()
+                            .unwrap_or(field)
+                    })
+                    .collect(),
+            ));
 
-        let table_schema = Arc::new(ArrowSchema::new(
-            self.arrow_schema()?
-                .fields
-                .clone()
-                .into_iter()
-                .map(|field| {
-                    file_schema
-                        .field_with_name(field.name())
-                        .cloned()
-                        .unwrap_or(field)
-                })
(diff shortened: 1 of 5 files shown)


rust/src/operations/transaction/state.rs
```diff
@@ -82,6 +84,45 @@ impl DeltaTableState {
 
         Ok(sql_to_rel.sql_to_expr(sql, &df_schema, &mut Default::default())?)
     }
+
+    /// Get the pysical table schema.
+    ///
+    /// This will construct a schema derived from the parqet schema of the latest data file,
+    /// and fields for partition columns from the schema defined in table meta data.
+    pub async fn physical_arrow_schema(
```
wjones127, line 92: I don't think we can guarantee this schema is consistent across all parquet files in the table; different writers may have written to the table with different physical types for timestamps. IMO this should be handled in the scan of each Parquet file. That is, we should cast the physical type to microsecond timestamps as needed.

wjones127: In PyArrow, we handle the int96 timestamp issue by passing an argument to the reader to coerce it to microsecond precision. Maybe we could implement something similar upstream?

https://github.com/delta-io/delta-rs/blob/34d43b6b336c065731dca4c43268ac18b0beef9c/python/tests/test_table_read.py#L32

roeap: There definitely are no guarantees as to the file schema being consistent. Datafusion however needs a consistent schema. Once we get into column mappings etc, things might get even more demanding and we may have to roll our own parquet scan, or rather start putting logic into our DeltaScan.

That said, I do believe using the schema from the latest file is an improvement over the current way, which at least for me fails for more or less every databricks written table where there are timestamps involved.

Not sure about the best way forward, but I'm happy to keep that logic on a private branch somewhere until we have a more general fix.

roeap: somewhat related, so I already published it as WIP - in #1267 I did some work on the write command. There my plan was to use the same schema to validates writes. But there it would be even more confusing, since we might end up on situation, where writing the "official" schema of the chart would not be permissible. But somehow it feels very strange to me to have potentially many schemas in the same table.

i guess spark must allow at least some flexibility in what schema it expects at write time, otherwise how would we end up in this discussion at all :D.

wjones127: Yeah we are definitely hitting the limits of DataFusion's scanner. I've created an issue upstream https://github.com/apache/arrow-datafusion/issues/5950

I'm fine with moving this forward; I mostly care that we have a more robust implementation in the future and have at least some momentum towards it.

rust/src/operations/transaction/state.rs
```diff
@@ -82,6 +84,45 @@ impl DeltaTableState {
...
+        &self,
+        object_store: Arc<dyn ObjectStore>,
+    ) -> DeltaResult<ArrowSchemaRef> {
+        let file_meta = self
+            .files()
+            .iter()
+            .max_by_key(|obj| obj.modification_time)
+            .ok_or(DeltaTableError::Generic("No active file actions to get physical schema. Maybe the current state has not yet been loaded?".into()))?
```
wjones127, line 100: Does this error propagate to the user? Does this mean trying to scan an empty tables leads to an error? I don't think it should.

wjones127 (changes requested):
Finished looking through. Just one other comment.

rust/src/operations/transaction/state.rs
roeap: It does. at the time I though we could fail on a scan if no files have been added yet, but you are right there are several valid scenarios where we have no files in a table and still should be able to do a scan.

Fixed it so we a re falling back to the schema from metadata.

wjones127 (approved):

PR #1267, opened 2023-04-09
feat: write command improvements

roeap:
# Description

This PR updates the write operation, specifically:

* makes handling state consistent with other operations
* avoids some IO when creating updated table state

One potentially more controversial change is that we now try to case all batches written to either the physical schema (derived from latest parquet file) or the schema derived from the table. I feel this logic may eventually need some updating or at least we need to make sure this is consistently handled across all write paths. In any case, using the file schema we dis already see some schema mis-matches in current tests, so that I included some casting for record batches here.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -13,12 +13,18 @@ readme = "README.md"
 edition = "2021"
 
 [dependencies]
-arrow = { version = "37.0.0", optional = true }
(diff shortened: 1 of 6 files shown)


rtyler (approved):
I'm looking forward to trying some of these changes out with my writer code. This pull request has been hanging out for a little bit so I'm going to land it

PR #1273, opened 2023-04-10
Support reading from HDFS via datafusion-objectstore-hdfs

yjshen:
# Description

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/test_utils.rs
+++ b/rust/src/test_utils.rs
@@ -52,6 +52,7 @@ impl IntegrationContext {
             StorageIntegration::Microsoft => format!("az://{}", &bucket),
             StorageIntegration::Google => format!("gs://{}", &bucket),
             StorageIntegration::Local => format!("file://{}", &bucket),
+            StorageIntegration::Hdfs => format!("hdfs://localhost:9000/{}", &bucket),
         };
         // the "storage_backend" will always point to the root ofg the object store.
         // TODO should we provide the store via object_Store builders?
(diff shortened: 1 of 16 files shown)


rtyler:
Interestingly this does not build on my machine which does not have Java installed:

```
error: failed to run custom build command for `fs-hdfs3 v0.1.11`

Caused by:
  process didn't exit successfully: `/usr/home/tyler/source/github/delta-io/delta-rs/target/debug/build/fs-hdfs3-60a8f4d683c18085/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'JAVA_HOME shell environment must be set: environment variable not found', /home/tyler/.cargo/registry/src/github.com-1ecc6299db9ec823/fs-hdfs3-0.1.11/build.rs:146:13
warning: build failed, waiting for other jobs to finish...
```

yjshen:
Hi @rtyler, do you mean you cannot compile delta-rs with `hdfs` enabled because of a lack of JVM env?

I think this is expected since fs-hdfs3 requires a JVM env setting to build its hdfs lib and ffi https://github.com/datafusion-contrib/fs-hdfs/blob/hadoop-3.1.4/build.rs#L24-L28

rust/src/test_utils.rs
```diff
@@ -447,3 +458,46 @@ pub mod gs_cli {
         set_env_if_not_set("GOOGLE_ENDPOINT_URL", "http://localhost:4443/storage/v1/b");
     }
 }
+
+/// small wrapper around hdfs cli
+pub mod hdfs_cli {
+    use std::env;
+    use std::path::PathBuf;
+    // use super::set_env_if_not_set;
```
wjones127, line 466: Can we drop this?
```suggestion
```

wjones127 (changes requested):
Thanks for submitting this PR. The changes look good, but before we merge could you enable the tests in CI? We'll need to enable the feature in:

https://github.com/delta-io/delta-rs/blob/c8371b38fdf22802f0f91b4ddc2a47da6be97c68/.github/workflows/build.yml#L108

And then also install the Java dependencies in that same job. Could you add that change to this PR?

iajoiner:
@yjshen consented to me taking over this PR..so it has been replaced with #1279.

PR #1275, opened 2023-04-11
feat: use new conflict checker in Python

### LINKED ISSUE #1254: Data is replicated when overwriting data concurrently
# Environment

**Delta-rs version**: 0.8.1

**Environment**:
- **Cloud provider**: s3 or locally
- **OS**: Ubuntu 20.04

***
# Bug

**What happened**: Data is replicated in the Delta Table when writing to the table with `write_deltalake(..., mode="overwrite")`

**What you expected to happen**: Data is not replicated, there are no extra duplications in the data.

**How to reproduce it**:
```
import threading
from shutil import rmtree

import pandas as pd

path = "demo"

rmtree("demo", ignore_errors=True)

df = pd.DataFrame(data=[('one', 1), ('two', 2)], columns=["name", "val"])

write_deltalake(path, data=df, mode="append")

assert len(DeltaTable(path).to_pandas()) == 2, len(DeltaTable(path).to_pandas())


def overwrite(delta_table: DeltaTable):
    table = delta_table.to_pyarrow_table()  # appears like redundant file is being read here
    write_deltalake(delta_table, table, mode="overwrite")


def comp():
    for _ in range(5):
        overwrite(DeltaTable(path))


for i in range(5):
    overwrite(DeltaTable(path))

assert len(DeltaTable(path).to_pandas()) == 2, len(DeltaTable(path).to_pandas())  # Sequential - ok

t1 = threading.Thread(target=comp)
t2 = threading.Thread(target=comp)
t1.start()
t2.start()
t1.join()
t2.join()

res = DeltaTable(path).to_pandas()

assert len(res) == 4, len(res)  # Concurrent - fails

```

**More details**:

The same happens when writing to S3 having enabled locking through dynamo (`AWS_S3_LOCKING_PROVIDER = "dynamodb"` in envs).

wjones127:
# Description

Switches Python implementation over to the new transaction API.

# Related Issue(s)

* Closes #1254

# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -16,12 +16,10 @@ use deltalake::builder::DeltaTableBuilder;
 use deltalake::checkpoints::create_checkpoint;
 use deltalake::datafusion::prelude::SessionContext;
 use deltalake::delta_datafusion::DeltaDataChecker;
+use deltalake::operations::transaction::commit;
 use deltalake::operations::vacuum::VacuumBuilder;
 use deltalake::partitions::PartitionFilter;
-use deltalake::{
-    DeltaDataTypeLong, DeltaDataTypeTimestamp, DeltaTableMetaData, DeltaTransactionOptions,
-    Invariant, Schema,
-};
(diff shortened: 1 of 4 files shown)


roeap (approved):
A small-ish change with a great effect :).

PR #1287, opened 2023-04-14
WIP: Handle timestamp conversions from nanosecond to microseconds on write

rtyler:
This is a work in progress


ion-elgreco:
@rtyler can we close this?

PR #1300, opened 2023-04-19
Enable the json feature for the parquet crate

rtyler:
This feature makes it much easier to read/write JSON to parquet. This is immediately useful in kafka-delta-ingest but I believe will be much more generally useful for all consumers of the package


houqp (approved):
It would be good to turn it into an optional feature so downstream consumers can opt out if they don't need to pull in all the json serde code.

PR #1293, opened 2023-04-16
issue-1292,issue-1291: test cases to reproduce issue

cmackenzie1:
# Description
This commit adds a unit test demonstrating the issues described in #1292 and #1291

# Related Issue(s)
- #1292 
- #1291 
- #1168

# Documentation

N/A

PR #1303, opened 2023-04-20
fix: scan partitioned tables with datafusion

### LINKED ISSUE #1168: DataFusion integration incorrectly handles partition columns defined "first" in schema
# Environment

**Delta-rs version**: 0.7.0

**Binding**: Rust

**Environment**:
- **Cloud provider**: n/a
- **OS**: macOS
- **Other**: M1

***
# Bug

**What happened**:
With Datafusion, when filtering a table on a non-partition column in a partitioned table leads to unexpected results or errors, when partition columns are defined "first" in schema.

Expected filter to work on non-partition column correctly.

**How to reproduce it**:
Wrote this test. This fails because it is treating value (not partitioned as the part column Dict(UInt16, Int32) instead of Int32.

```rust
#[tokio::test]
async fn temp_test() -> Result<()> {
    let table_loc = "./tests/data/temp-test";
        .with_location(table_loc)
        .with_columns(vec![
            SchemaField::new("part".to_owned(), deltalake::SchemaDataType::primitive("integer".to_owned()), false, HashMap::new()),
            SchemaField::new("value".to_owned(), deltalake::SchemaDataType::primitive("integer".to_owned()), false, HashMap::new()),
        ])
        .with_partition_columns(vec!["part"])
        .await?;

    let arrow_schema = table.schema().unwrap().try_into()?;
    let value_col = Int32Array::from(vec![10, 20, 30]);
    let part_col = Int32Array::from(vec![1, 2, 3]);
    let record_batch = RecordBatch::try_new(
        Arc::new(arrow_schema),
        vec![Arc::new(value_col), Arc::new(part_col)]
    )?;

    let table = WriteBuilder::new()
        .with_location(table_loc)
        .with_input_batches(vec![record_batch])
        .await?;

    println!("{:?}", table.schema());
    println!("{:?}", table.get_partition_values().collect::<Vec<_>>());

    ctx.register_table("demo", Arc::new(table))?;

    let actual = ctx
        .sql("SELECT part, value FROM demo WHERE value = 1")
        .await?
        .collect()
        .await?;

    let expected = vec![
        "+------+-------+",
        "| part | value |",
        "+------+-------+",
        "| 10   | 1     |",
        "+------+-------+",
    ];

    assert_batches_sorted_eq!(&expected, &actual);

    Ok(())
}
```

**More details**:
This seems to be happening because in the DataFusion scan, the FileScanConfig's FileSchema is built from non-partition cols and partition cols are provided as table_partition_cols. This is expected but the projection pass to FileScanConfig is not updated to account for this.

See the doc for projection on FileScanConfig:
```
    /// Columns on which to project the data. Indexes that are higher than the
    /// number of columns of `file_schema` refer to `table_partition_cols`.
```

As a quick fix I added this to the scan:

```rust
        let field_lookup = file_schema.fields().iter().map(|f| f.name()).chain(table_partition_cols.iter()).collect::<Vec<_>>();
        let updated_projection = if let Some(proj) = projection {
            Some(proj.iter().map(|p| field_lookup.iter().position(|f| *f == schema.field(*p).name()).unwrap()).collect::<Vec<usize>>())
        } else {
            None
        };
```
Basically update the projection to use the indexes into file_schema + table_partition_cols.
This assumes the projection indexes are from `schema.field()` which it looks like it is up the stack in datafusion.

Making this change does fix the example tests but breaks `test_datafusion_partitioned_types` because the column output ordering is different.

I can dig in further and submit a PR (if something like this is the correct fix) once I go through my company's open source contribution policy.

### LINKED ISSUE #1291: Unable to query partitioned tables
# Environment

**Delta-rs version**: `0.9.0`

**Binding**: 

**Environment**:
- **Cloud provider**: N/A
- **OS**: `macOS 13.3.1`
- **Other**:

***
# Bug

**What happened**:

Trying to query a partitioned table using `datafusion` is failing. 

Should return the results matching the partition filter. 

**How to reproduce it**:

1. Create a partitioned table (replicated using https://github.com/delta-io/delta-rs/tree/main/rust/tests/data/delta-0.8.0-partitioned)
2. Add data to table
3. Query table using `SELECT * FROM t WHERE year = 2020`

```
Internal("The type of Dictionary(UInt16, Utf8) Eq Utf8 of binary physical should be same")
```

**More details**:

Trying again where the partition column type is string yields no results and no error. I think the resulting parquet file is missing the partition column as data.

### LINKED ISSUE #1292: Datafusion: SQL projection returns wrong column for partitioned data
# Environment

**Delta-rs version**: `0.9.0`

**Binding**: `rust`

**Environment**:
- **Cloud provider**: N/A
- **OS**: `macOS 13.3.1`
- **Other**: 

***
# Bug

**What happened**:

Querying a partitioned table using a column projection results in the wrong column being returned. 

The correct column projection and data should be returned

**How to reproduce it**:

See #1293 

**More details**:

Looks like same issue described in #1168

roeap:
# Description

This PR builds on #1293 and tries to address some of the issues we have seen with scanning
partitioned tables with datafusion. And while all our tests (mostly - more on that later) pass,
the fix involves some behaviours that we may or may not want to adopt. Specifically, Datafusion
appends partition columns at the end of the schema fields, while we have been reporting them
as leading columns.

In recent datafusion versions also changed the default for dictionary encoding partition columns
to be opt in. My thinking was that for the vast majority of tables keeping dictionary encoding for
partition columns would be the desired behaviour. (@wjones127, do you have an opinion on that?).
This was also a root cause or at least related to the second failing test.

I did have to comment out some caeses within out file pruning tests where we create expression with
nulls, as I have thus far not been able to create an expression that datafusion is happy with. I'll
keep trying, but have some work on expression parsing for handling user inputs planned as well.
There already is a draft PR open (#1267), which does not contain that yet, but where I plan to
address this.

cc @cmackenzie1 

# Related Issue(s)

- closes #1291
- closes #1292
- closes #1168


rtyler (approved):
All things considered I think this is fine to land. Are there some issues filed with Data Fusion that can be dropped in these comments for the missing kernels?

wjones127:
> Specifically, Datafusion appends partition columns at the end of the schema fields, while we have been reporting them as leading columns.

Eventually, I think our desired behavior is going to be put the partition columns where they are in the Delta table schema.

> In recent datafusion versions also changed the default for dictionary encoding partition columns to be opt in. My thinking was that for the vast majority of tables keeping dictionary encoding for partition columns would be the desired behaviour.

Looking at the PR for that https://github.com/apache/arrow-datafusion/pull/5545#issue-1618827927, they make a good case that dictionary encoding partition columns is really only helpful for string columns. So perhaps we should only use them for string / binary columns (any maybe dates too)? And if we wanted to get fancy we could choose the dictionary index bit width based on the known number of unique values (that is, `i8` if 256 or fewer, `i16` if there are much more). [^1]

[^1]: Eventually, I think we'd like them to use the new REE (run-end encoded) arrays. But that's far in the future.

rust/src/operations/transaction/state.rs
```diff
@@ -21,16 +23,38 @@ use crate::action::Add;
...
+                meta.schema
+                    .get_fields()
+                    .iter()
+                    .filter(|f| meta.partition_columns.contains(&f.get_name().to_string()))
+                    .map(|f| {
+                        let field = ArrowField::try_from(f)?;
+                        let corrected = match field.data_type() {
+                            // TODO Datafusion seems to currently not implement comparsion kernes for booleans in
+                            // dict arrays, so we have to omit converting that type for now.
+                            DataType::Boolean => field.data_type().clone(),
```
wjones127, line 48: I think it's actually a good idea to not encode booleans.
```suggestion
                            // Dictionary encoding columns bools is a waste of space
                            DataType::Boolean => field.data_type().clone(),
```

rust/src/operations/transaction/state.rs
```diff
@@ -21,16 +23,38 @@ use crate::action::Add;
...
+                            _ => wrap_partition_type_in_dict(field.data_type().clone()),
+                        };
+                        // let dict_type = wrap_partition_type_in_dict(field.data_type().clone());
```
wjones127, line 51:
```suggestion
```

wjones127 (dismissed):
Some minor suggestions for comments but otherwise seems good.

rust/src/operations/transaction/state.rs
```diff
@@ -43,8 +43,14 @@ impl DeltaTableState {
...
                         let field = ArrowField::try_from(f)?;
                         let corrected = match field.data_type() {
-                            // Dictionary encoding columns bools is a waste of space
-                            DataType::Boolean => field.data_type().clone(),
+                            // Dictionary encoding for types around the size of dictionary values does not make much sense.
+                            DataType::Boolean
+                            | DataType::Int8
+                            | DataType::Int16
+                            | DataType::Int32
+                            | DataType::Utf8
```
wjones127, line 51: I think you mean `Uint8` here? but honestly I think it still is probably worth it for all the types except boolean. Reasoning explained here: https://github.com/apache/arrow-datafusion/pull/5545#issuecomment-1526917997

roeap: Indeed I did - glad you caught that! Removed the other values again, as per your comment - makes a lot of sense!

wjones127 (approved):

Issue #622, opened 2022-06-04
Expose optimize in Python bindings


wjones127:
# Description

#607 added optimize implementation using datafusion in rust. If we enable the `datafusion` option in the Python package, we can expose this in Python. We'll likely want this feature anyways to support writer protocol version 2 (see #592).

**Use Case**

**Related Issue(s)**

a-m-w:
Is there an approximate timeline when optimization will be supported via Python? Thanks in advance!

wjones127:
I don't believe anyone has volunteered to work on this yet, so no timeline for now. PRs for this are welcome though! 😄

loleek:
Maybe I can contribute this. I've already implement it today but without doc and unit-tests.This is my first PR, it maybe take sometime.

roeap:
@loleek - that would be awesome! 

We are always happy to help you along with reviews and guidance if needed.

ion-elgreco:
@wjones127 this can be closed now, right?

wjones127:
Yes, thanks for reminding me!

PR #1313, opened 2023-04-27
feat: add optimize command in python binding

loleek:
# Description
This is a implementation of the Optimize Command for python binding.

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/622


python/deltalake/table.py
```diff
@@ -313,6 +313,19 @@ def vacuum(
...
         return self._table.vacuum(dry_run, retention_hours, enforce_retention_duration)
 
+    def optimize(
+            self,
+            partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
+            target_size: Optional[int] = None,
+    ) -> Dict[str, Any]:
+        """
+        :param partition_filters: the partition filters that will be used for getting the matched files
+        :param target_size: desired file size after bin-packing files
```
wjones127, line 323:
```suggestion
        :param target_size: desired file size after bin-packing files, in bytes. If not
          provided, will attempt to read the table configuration value ``delta.targetFileSize``.
          If that value isn't set, will use default value of 256MB.
```

python/deltalake/table.py
```diff
@@ -313,6 +313,19 @@ def vacuum(
...
+        :return: the metrics from optimize
+        """
+        metrics = self._table.optimize(partition_filters, target_size)
+        return json.loads(metrics)
```
wjones127, line 327: It would be nice to have this be typed for the sake of autocompletion, but doing that in Rust right now involves a bit of boilerplate. One lightweight option is https://docs.python.org/3.8/library/typing.html#typing.TypedDict, but that's Python>=3.8 only. We still have a month or two until 3.7 is EOL.

python/docs/source/usage.rst
```diff
@@ -399,7 +399,23 @@ only list the files to be deleted. Pass ``dry_run=False`` to actually delete fil
...
 ~~~~~~~~~~~~~~~~~
 
-Optimizing tables is not currently supported.
+Optimizing a table will perform bin-packing on a Delta Table which merges small files
+into a large file. Bin-packing reduces the number of API calls required for read operations.
+Optimizing will increments the table's version and creates remove actions for optimized files.
+Optimize does not delete files from storage. To delete files that were removed, call :meth:`DeltaTable.vacuum`.
+
+Use :meth:`DeltaTable.optimize` to perform the optimize operation. Note that currently optimize only supports
+append-only workflows. Use with other workflows may corrupt your table state.
```
wjones127, line 408: I don't think this is true anymore ever since #632. We should just say that it will fail if there are concurrent transactions other than appends.

```suggestion
Use :meth:`DeltaTable.optimize` to perform the optimize operation. Note that this method will fail if a concurrent writer performs an operation that removes any files (such as an overwrite).
```

python/deltalake/table.py
wjones127: Perhaps something like:

```python
if TYPE_CHECKING and sys.version_info().minor >= 8:
    from typing import TypedDict

    class MetricDetails(TypeDict):
        min: int
        max: int
        avg: float
        total_files: int
        total_size: int
    
    class OptimizeMetrics(TypeDict):
        num_files_added: int
        num_files_removed: int
        files_added: MetricDetails
        files_removed: MetricDetails
        partitions_optimized: int
        num_batches: int
        total_considered_files: int
        total_files_skipped: int
        preserve_insertion_order: bool
```

Then you can modify the signature:


```python
def optimize(
            self,
            partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
            target_size: Optional[int] = None,
    ) -> "OptimizeMetrics":
```

wjones127 (changes requested):
This looks quite good. Added a few suggestions.

python/deltalake/table.py
wjones127: (I'm happy to leave this for a follow up later FYI)

djouallah:
can you guys please don't release the next version of python package before you merge this, this is a very important functionality

python/deltalake/table.py
loleek: ok. I'll leave this in current PR.

wjones127 (approved):

PR #1314, opened 2023-04-28
feat: upgrade to Arrow 37 and Datafusion 23

rtyler:
Lots of fun API changes, `Fields` is the biggest impact in terms of lines of code however.

BREAKING CHANGE:  new major versions for arrow and datafusion

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -321,18 +321,6 @@ impl RawDeltaTable {
             .collect())
     }
 
-    pub fn arrow_schema_json(&self) -> PyResult<String> {
-        let schema = self
-            ._table
-            .get_schema()
-            .map_err(PyDeltaTableError::from_raw)?;
-        serde_json::to_string(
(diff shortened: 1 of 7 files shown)


python/src/lib.rs
```diff
@@ -321,18 +321,6 @@ impl RawDeltaTable {
...
-        let schema = self
-            ._table
-            .get_schema()
-            .map_err(PyDeltaTableError::from_raw)?;
-        serde_json::to_string(
-            &<ArrowSchema as TryFrom<&deltalake::Schema>>::try_from(schema)
-                .map_err(PyDeltaTableError::from_arrow)?,
-        )
-        .map_err(|_| PyDeltaTableError::new_err("Got invalid table schema"))
-    }
```
wjones127, line 334: I don't think this was used in our public API, and I know the arrow-rs folks wanted to remove the unofficial JSON schema serialization from their public API as well. I think it should be fine to drop this.

wjones127 (approved):

PR #1315, opened 2023-04-28
feat: use asynchronous programming in the Python binding

fvaleye:
# Description
- Leveraging `asyncio` in Python in `DeltaTable` by providing asynchronous methods with `_async` suffix.
- Limiting only to read operations (only`update_incremental()` is supported).
- Add the capability to create the DeltaTable without loading the Delta Table.
- Provide `load()` method for synchronous and asynchronous loading of the Delta Table.
- Add fixtures and `pytest-asyncio` in the test for async operations

# Documentation
- https://pyo3.rs/
- https://docs.python.org/fr/3/library/asyncio.html

# Check-list
- [x] Decide on a better encapsulation mechanism of the `RawDeltaTable` from`RawDeltaTableAsync`. 
- [x] Add documentation and examples on how to use `DeltaTable` with asynchronous methods.
- [x] Decide to remove the support of Python 3.7


djouallah:
can you guys please don't release the next version of python package before you merge this, this is a very important functionality

wjones127:
## Dropping support for Python 3.7

I think we'll probably remove support for Python 3.7 in a month or two, once it reaches EOL. https://endoflife.date/python

## Use cases for Asyncio

IIUC, we want async functions for cases where we want to load multiple Delta tables concurrently, or load them without blocking (such as in an async server). That means we should probably add some function that does the same things as `DeltaTable.__init__`, but is async

```python
async def load_table(table_or_uri: str, ...) -> DeltaTable:
    ...
```

Then a user could load several tables at a time with:

```python
table1, table2 = await asyncio.gather(
    load_table("path/to/table1"),
    load_table("path/to/table2")
)
```

I'm not sure, but it doesn't seem strictly necessary to have a separate class for the async methods. Let me know if I am missing something. But perhaps we can have async methods post-fixed with `_async`? (or maybe there is something more standard we should use?)

```python
class DeltaTable:
    def update_incremental() -> None:
        ...
    async def update_incremental_async(self) -> None:
        ...
```

BTW, I was surprised to learn that Jupyter has an "autoawait" feature that makes async friendly in notebooks. https://ipython.readthedocs.io/en/stable/interactive/autoawait.html

fvaleye:
Hey @wjones127 👋,

> ## Dropping support for Python 3.7
> I think we'll probably remove support for Python 3.7 in a month or two, once it reaches EOL. https://endoflife.date/python
> 
Ok, I was asking the question for this [failing step](https://github.com/delta-io/delta-rs/actions/runs/4832444156/jobs/8611219118?pr=1315) in the CI, to not block this PR for it.

> ## Use cases for Asyncio
> IIUC, we want async functions for cases where we want to load multiple Delta tables concurrently, or load them without blocking (such as in an async server). That means we should probably add some function that does the same things as `DeltaTable.__init__`, but is async
> 
> ```python
> async def load_table(table_or_uri: str, ...) -> DeltaTable:
>     ...
> ```
> 
> Then a user could load several tables at a time with:
> 
> ```python
> table1, table2 = await asyncio.gather(
>     load_table("path/to/table1"),
>     load_table("path/to/table2")
> )
> ```
> 
> I'm not sure, but it doesn't seem strictly necessary to have a separate class for the async methods. Let me know if I am missing something. But perhaps we can have async methods post-fixed with `_async`? (or maybe there is something more standard we should use?)

I had initially come up with the idea to only create `_async` methods in Python and Rust. However, I faced a challenge when it came to including the asynchronous methods in Rust in order to allow `pyo3` and `pyo3_asyncio` to work together on the same `RawDeltaTable` structure. The need of two structures to enable async calls was only way to make it work. Ultimately, I decided to stick with what was working, and using the `DeltaTable` inheritance in `Python` proved to be the easiest solution.

If you find something on how to make it work in Rust, let's refactor this. I will ask the question on my side.

> 
> ```python
> class DeltaTable:
>     def update_incremental() -> None:
>         ...
>     async def update_incremental_async(self) -> None:
>         ...
> ```
> 
> BTW, I was surprised to learn that Jupyter has an "autoawait" feature that makes async friendly in notebooks. https://ipython.readthedocs.io/en/stable/interactive/autoawait.html

Nice 👍 I don't know it existed!

wjones127:
@fvaleye I'm in favor of dropping 3.7 support now. But I think I'd like the upgrade to be done in a separate branch. That CI job should be changed to use Python 3.7 + PyArrow 7. I can start looking at that this week.

fvaleye:
> @fvaleye I'm in favor of dropping 3.7 support now. But I think I'd like the upgrade to be done in a separate branch. That CI job should be changed to use Python 3.7 + PyArrow 7. I can start looking at that this week.

All right! I changed the CI job configuration.
Let's drop the Python 3.7 in another PR and merge this one!

python/src/lib.rs
```diff
@@ -94,9 +95,10 @@ enum PartitionFilterValue<'a> {
 
 #[pyclass]
 struct RawDeltaTable {
-    _table: deltalake::DeltaTable,
+    _table: Arc<Mutex<deltalake::DeltaTable>>,
     // storing the config additionally on the table helps us make pickling work.
     _config: FsConfig,
+    _rt: tokio::runtime::Runtime,
```
roeap, line 101: Since we are now tracking a runtime directly on the table, we should likely figure how how we can share that best with both the filesystem implementation and the data checker.

python/src/lib.rs
```diff
@@ -302,16 +361,18 @@ impl RawDeltaTable {
         retention_hours: Option<u64>,
         enforce_retention_duration: bool,
     ) -> PyResult<Vec<String>> {
-        let mut cmd = VacuumBuilder::new(self._table.object_store(), self._table.state.clone())
+        let mut table = self._rt.block_on(self._table.lock());
+        let mut cmd = VacuumBuilder::new(table.object_store(), table.state.clone())
```
roeap, line 365: is there a way for us to not clone the state? Since are are consuming it in the builder, and replacing the table object later we are actually not re-using the current state anywhere, right?

The same applies to other operations.

python/deltalake/table.py
wjones127, line 1: Since I started to think about #1361, it seems like nearly every method could involve some IO, so they will all have async versions. So now I think I might be in favor of a separate async API instead now:

```python
# Can use sync API:
# Or async API:
from deltlake.async import DeltaTable
```

I'm not sure the cleanest way to implement this, but ideally we should be able to do so without having too much duplication?

Maybe `deltalake.DeltaTable` just becomes a wrapper around `deltalake.async.DeltaTable`, implemented like:

```python
class DeltaTable:
   def load_with_datetime():
        asyncio.run(self.inner.load_with_datetime())
```

What do you think of that? Not sure what the implementations are of this. Ideally, the sync API can be used within multiple Python threads and still get concurrency (since we don't hold the GIL in most operations).

python/src/lib.rs
wjones127: Actually should we have a global runtime for the Python bindings? IiUC if each table has a separate runtime then each one has their own thread pool which probably isn’t a great idea.

roeap: Agreed - this would ideal. Had a quick read through the pyo3-asyncio docs in the hopes that they would somehow have a way to manage a runtime, but this does not seem to be the case. I guess this would bring us to a session of some sorts, or at least some globally managed state?

There are very likely other solutions, but off the top of my head we may be able to create some sort of python singleton, that houses the runtime, which we can pass around transparent to the user.

python/deltalake/table.py
roeap: To me that makes a lot of sense, since as you say we are essentially async on the rust side, and need to transform somewhere. We should however probably find a quick benchmark to compare alternatives. IIRC, we saw a big overhead in the file system when we were instantiating a tokio runtime each time, so I guess there may be implications when interacting with python loops as well.

python/Cargo.toml
```diff
@@ -34,7 +34,11 @@ reqwest = { version = "*", features = ["native-tls-vendored"] }
 
 [dependencies.pyo3]
 version = "0.18"
-features = ["extension-module", "abi3", "abi3-py37"]
+features = ["extension-module", "abi3", "abi3-py37", "auto-initialize"]
```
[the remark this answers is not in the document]
fvaleye: We will drop the support of Python 3.7 in another PR! So, I keep it for now.

python/src/lib.rs
fvaleye: Agreed! I will try to find a way to integrate a Python singleton and let you know!

python/deltalake/table.py
fvaleye: It's the best refactoring to do if we want to fully embrace async in the Python binding whitout having too much duplication. I will try to revamp the approach but I'm wondering if it's not better to wait for #1361, WDYT @wjones127?

python/docs/source/usage.rst
```diff
@@ -87,6 +87,27 @@ For Databricks Unity Catalog authentication, use environment variables:
 .. _`azure options`: https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html#variants
 .. _`gcs options`: https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html#variants
 
+To lazily load the table, use the :meth:`DeltaTable.load` method.
+.. code-block:: python
+
+    >>> from deltalake import DeltaTable
+    >>> dt = DeltaTable("../rust/tests/data/delta-0.2.0", init=false)
```
watfordkcf, line 94:
```suggestion
    >>> dt = DeltaTable("../rust/tests/data/delta-0.2.0", init=False)
```

python/src/lib.rs
```diff
@@ -118,12 +120,13 @@ struct RawDeltaTableMetaData {
 #[pymethods]
 impl RawDeltaTable {
     #[new]
-    #[pyo3(signature = (table_uri, version = None, storage_options = None, without_files = false))]
+    #[pyo3(signature = (table_uri, version = None, storage_options = None, without_files = false, init = true))]
```
watfordkcf, line 123: Forgive my naivete with pyo3 signatures, but wouldn't it be:
```suggestion
    #[pyo3(signature = (table_uri, version = None, storage_options = None, without_files = False, init = True))]
```

watfordkcf (commented):
I'm working on federating access to delta tables via GraphQL and would enjoy async resolvers in python, but I'm new to Rust. Happy to help get this over the finish line however I can.

ion-elgreco:
Any chance of reviving this PR?

wjones127:
> Any chance of reviving this PR?

I think we need a clearer case of why we need an async API and how we can do that without a huge maintenance burden (due to duplicated APIs).

IMO, if you are working in an async context, you might be just as well off running delta lake operations in the background using [loop.run_in_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor). We don't (or shouldn't) hold the GIL during operations, so we won't block the main thread. We can add some examples of this to the documentation, if you'd like.

watfordkcf:
> We don't (or shouldn't) hold the GIL during operations, so we won't block the main thread.

Is the GIL held during queries? Right now we'd like all of our graphql resolvers to be async. We've migrated most of the work to Rust at the moment, but would prefer Python when possible (our primary language).

rtyler:
I am going to close this. I personally think it's a cool change, but I don't know that the other Python maintainers support it enough to revive this pull request

PR #1323, opened 2023-05-01
chore: add deprecation notices for commit logic on `DeltaTable`

roeap:
# Description

Now that we have a decent test coverage to the low-level commit logic in the transactions module, we should deprecate the commit logic exposed directly on `DeltaTable`.

In some places I also updated internal call sites, however did not go through all of them.

# Related Issue(s)

# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -19,7 +19,9 @@ use deltalake::delta_datafusion::DeltaDataChecker;
 use deltalake::operations::transaction::commit;
 use deltalake::operations::vacuum::VacuumBuilder;
 use deltalake::partitions::PartitionFilter;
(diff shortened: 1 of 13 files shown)


python/src/lib.rs
```diff
@@ -740,33 +744,45 @@ fn write_new_deltalake(
...
+        .unwrap_or_default()
+        .into_iter()
+        .filter_map(|(key, value)| {
+            if let Ok(key) = DeltaConfigKey::from_str(&key) {
+                Some((key, value))
+            } else {
+                None
+            }
+        })
+        .collect();
```
roeap, line 766: Not quite sure how we want to handle table configuration vs. metadata. With this change only known configuration can be passed to delta table, but as of right now there is no way to pass in metadata. First question would be how we want to deal with invalid config. While the code just silently ignores it, I am starting to lean towards raising an error instead. For metadata we would need to introduce a new parameter.

PR #1324, opened 2023-05-01
fix: double url encode of partition key

### LINKED ISSUE #1228: Error reading partitioned table with slash in partition key
# Environment

**Delta-rs version**: 0.8.1

**Binding**: python

**Environment**:
- Azure and locally

***
# Bug

**What happened**:
I'm saving a partitioned table in Databricks (DBR 12.1) into Azure Data Lake Storage V2 which is partitioned on a string column containing slashes. When trying to read the table using deltalake from my local pc I get an error saying there is 'No body' in the parquet file.

```
  File "[..]/delta_connect.py", line 32, in <module>
    table = dt.to_pandas(
  File "[..]/python3.10/site-packages/deltalake/table.py", line 418, in to_pandas
    return self.to_pyarrow_table(
  File "[..]/python3.10/site-packages/deltalake/table.py", line 400, in to_pyarrow_table
  File "[..]/python3.10/site-packages/deltalake/fs.py", line 22, in open_input_file
deltalake.PyDeltaTableError: Object at location partition_table_test/value=you%25252Fthere/part-00007-adfd81b1-1bff-4431-83c2-948e2274a892.c000.snappy.parquet not found: response error "No Body", after 0 retries: HTTP status client error (404 Not Found) for url [..]
```
If I query for a key without a slash it returns the data without issue.

Succesfully read the data

**How to reproduce it**:
In Databricks saving to ADLS2:

```python
df = spark.createDataFrame([
    (0, "hej"),
    (1, "you/there"),
    (2, "the two of us"),
], schema="id int, value string")
df.write.partitionBy("value").saveAsTable("mytable")
```

Locally running in python (3.10) gives the above error:

```python

container_name = "my_container"
storage_account_name = "my_sa"
table_name = "my_table"
sas_token = "my_sas_token"

abfs_path = (
    f"abfs://{container_name}@{storage_account_name}.dfs.core.windows.net/{table_name}/"
)

dt = DeltaTable(
    abfs_path,
    storage_options={
        "sas_key": sas_token,
    },
)

table = dt.to_pandas(
    partitions=[("value", "=", "you/there")],
)
```

Using the other value "hej" works fine:
```python

table = dt.to_pandas(
    partitions=[("value", "=", "hej")],
)
```
returns:

```
    id value
0   0   hej
```
**More details**:

mrjoe7:
# Description
This PR is fixing #1228.

# Related Issue(s)

-  #1228

### DIFF
--- a/rust/src/storage/utils.rs
+++ b/rust/src/storage/utils.rs
@@ -91,10 +91,37 @@ impl TryFrom<&Add> for ObjectMeta {
             Utc,
         );
         Ok(Self {
-            // TODO this won't work for absoute paths, since Paths are always relative to store.
-            location: Path::from(value.path.as_str()),
+            // TODO this won't work for absolute paths, since Paths are always relative to store.
+            location: Path::parse(value.path.as_str())?,
             last_modified,
             size: value.size as usize,
         })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_object_meta_from_add_action() {
+        let add = Add {
+            path: "x=A%252FA/part-00007-b350e235-2832-45df-9918-6cab4f7578f7.c000.snappy.parquet"
+                .to_string(),
+            size: 123,
+            modification_time: 123456789,
+            ..Default::default()
+        };
+
+        let meta: ObjectMeta = (&add).try_into().unwrap();
+        assert_eq!(
+            meta.location,
+            Path::parse(
+                "x=A%252FA/part-00007-b350e235-2832-45df-9918-6cab4f7578f7.c000.snappy.parquet"
+            )
+            .unwrap()
+        );
+        assert_eq!(meta.size, 123);
+        assert_eq!(meta.last_modified.timestamp_millis(), 123456789);
     }
 }
--- a/rust/src/delta_datafusion.rs
+++ b/rust/src/delta_datafusion.rs
@@ -54,7 +54,7 @@ use datafusion_physical_expr::execution_props::ExecutionProps;
 use datafusion_physical_expr::{create_physical_expr, PhysicalExpr};
 use datafusion_proto::logical_plan::LogicalExtensionCodec;
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
-use object_store::{path::Path, ObjectMeta};
+use object_store::ObjectMeta;
 use url::Url;
 
 use crate::builder::ensure_table_uri;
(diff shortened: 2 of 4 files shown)


rust/src/storage/utils.rs
```diff
@@ -91,8 +92,12 @@ impl TryFrom<&Add> for ObjectMeta {
...
         );
         Ok(Self {
-            // TODO this won't work for absoute paths, since Paths are always relative to store.
-            location: Path::from(value.path.as_str()),
+            // TODO this won't work for absolute paths, since Paths are always relative to store.
+            location: Path::from_url_path(
+                utf8_percent_encode(value.path.as_str(), percent_encoding::NON_ALPHANUMERIC)
+                    .to_string()
+                    .as_str(),
+            )?,
```
wjones127, line 100: How about:

```suggestion
            location: Path::parse(value.path.as_str())?,
```

I've tested this works locally.

rust/src/storage/utils.rs
```diff
@@ -9,6 +9,7 @@ use chrono::{DateTime, NaiveDateTime, Utc};
 use futures::{StreamExt, TryStreamExt};
 use object_store::path::Path;
 use object_store::{DynObjectStore, ObjectMeta, Result as ObjectStoreResult};
+use percent_encoding::utf8_percent_encode;
```
wjones127, line 12: If you make the other change.

```suggestion
```

rust/src/storage/utils.rs
wjones127, line 1: Would be nice to have a more direct unit test:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_object_meta_from_add_action() {
        let add = Add {
            path: "x=A%252FA/part-00007-b350e235-2832-45df-9918-6cab4f7578f7.c000.snappy.parquet"
                .to_string(),
            size: 123,
            modification_time: 123456789,
            ..Default::default()
        };

        let meta: ObjectMeta = (&add).try_into().unwrap();
        assert_eq!(
            meta.location,
            Path::parse(
                "x=A%252FA/part-00007-b350e235-2832-45df-9918-6cab4f7578f7.c000.snappy.parquet"
            )
            .unwrap()
        );
        assert_eq!(meta.size, 123);
        assert_eq!(meta.last_modified.timestamp_millis(), 123456789);
    }
}
```

wjones127 (changes requested):
Thanks for identifying the cause of this issue. Added a few suggestions.

rust/src/storage/utils.rs
mrjoe7: Good idea. But, I would like to keep both test cases if you are ok with that.

wjones127: Yes, this should be an additional case rather than a replacement.

wjones127 (approved):

PR #1331, opened 2023-05-03
feat: added support for Databricks Unity Catalog

nohajc:
# Description
I'm currently working with Azure Databricks which uses Unity Catalog for storing Delta Lake table paths and I thought it would be nice if `delta-rs` provided basic support for it.

There isn't any Databricks SDK library for Rust as far as I know, so this change introduces direct dependency to `reqwest`, `reqwest-middleware` and `reqwest-retry` crates when the `unity` feature flag is enabled.

# Documentation
https://docs.databricks.com/api-explorer/workspace/tables/get

# Notes
I manually tested against Azure Databricks workspace but the implementation should work with any cloud provider offering Databricks.

This is my first contribution, so I welcome any feedback. Thanks.

### DIFF
--- a/rust/src/data_catalog/unity/mod.rs
+++ b/rust/src/data_catalog/unity/mod.rs
@@ -1,0 +1,125 @@
+//! Databricks Unity Catalog.
+//!
+//! This module is gated behind the "unity" feature.
+use super::{DataCatalog, DataCatalogError};
+use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
+use reqwest_retry::{RetryTransientMiddleware, policies::ExponentialBackoff};
+use reqwest::header;
+use serde::Deserialize;
+
+/// Databricks Unity Catalog - implementation of the `DataCatalog` trait
+pub struct UnityCatalog {
+    client_with_retry: ClientWithMiddleware,
+    workspace_url: String,
+}
+
+impl UnityCatalog {
+    /// Creates a new UnityCatalog
+    pub fn new() -> Result<Self, DataCatalogError> {
+        let token_var = "DATABRICKS_ACCESS_TOKEN";
+        let access_token = std::env::var(token_var).map_err(|_|
+            DataCatalogError::MissingEnvVar { var_name: token_var.into() })?;
+
+        let auth_header_val = header::HeaderValue::from_str(
+            &format!("Bearer {}", &access_token)
+        ).map_err(|_| DataCatalogError::InvalidAccessToken)?;
+
+        let headers = header::HeaderMap::from_iter([
+            (header::AUTHORIZATION, auth_header_val),
+        ]);
+        let client = reqwest::Client::builder()
+            .default_headers(headers).build()?;
+
+        let retry_policy = ExponentialBackoff::builder()
+            .build_with_max_retries(10);
+
+        let client_with_retry = ClientBuilder::new(client)
+            .with(RetryTransientMiddleware::new_with_policy(retry_policy))
+            .build();
+
+        let workspace_var = "DATABRICKS_WORKSPACE_URL";
+        let workspace_url = std::env::var(workspace_var).map_err(|_|
+            DataCatalogError::MissingEnvVar { var_name: workspace_var.into() })?;
+
+        Ok(Self{
+            client_with_retry,
+            workspace_url,
+        })
+    }
+}
+
+#[derive(Deserialize)]
+#[serde(untagged)]
+enum TableResponse {
+    Success { storage_location: String },
+    Error { error_code: String, message: String },
+}
+
+/// Possible errors from the unity-catalog/tables API call
+#[derive(thiserror::Error, Debug)]
+pub enum GetTableError {
+    #[error("GET request error: {source}")]
+    /// Error from reqwest library
+    RequestError {
+        /// The underlying reqwest_middleware::Error
+        #[from]
+        source: reqwest_middleware::Error,
+    },
+
+    /// Request returned error response
+    #[error("Invalid table error: {error_code}: {message}")]
+    InvalidTable {
+        /// Error code
+        error_code: String,
+        /// Error description
+        message: String,
+    }
+}
+
+impl From<reqwest_middleware::Error> for DataCatalogError {
+    fn from(value: reqwest_middleware::Error) -> Self {
+        value.into()
+    }
+}
+
+impl From<reqwest::Error> for DataCatalogError {
+    fn from(value: reqwest::Error) -> Self {
+        reqwest_middleware::Error::Reqwest(value).into()
+    }
+}
+
+#[async_trait::async_trait]
+impl DataCatalog for UnityCatalog {
+    /// Get the table storage location from the UnityCatalog
+    async fn get_table_storage_location(
+        &self,
+        catalog_id: Option<String>,
+        database_name: &str,
+        table_name: &str,
+    ) -> Result<String, DataCatalogError> {
+        let resp = self.client_with_retry.get(format!(
+            "{}/api/2.1/unity-catalog/tables/{}.{}.{}",
+            &self.workspace_url,
+            catalog_id.as_deref().unwrap_or("main"),
+            &database_name,
+            &table_name
+        )).send().await?;
+
+        let parsed_resp: TableResponse = resp.json().await?;
+        match parsed_resp {
+            TableResponse::Success { storage_location } =>
+                Ok(storage_location),
+            TableResponse::Error { error_code, message } =>
+                Err(GetTableError::InvalidTable {
+                    error_code,
+                    message
+                }.into()),
+        }
+    }
+}
+
+impl std::fmt::Debug for UnityCatalog {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(fmt, "UnityCatalog")
+    }
+}
--- a/rust/src/data_catalog/unity/mod.rs
+++ b/rust/src/data_catalog/unity/mod.rs
@@ -1,6 +1,6 @@
 //! Databricks Unity Catalog.
 //!
-//! This module is gated behind the "unity" feature.
+//! This module is gated behind the "unity-experimental" feature.
 use super::{DataCatalog, DataCatalogError};
 use reqwest::header;
 use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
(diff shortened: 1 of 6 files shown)


rust/src/data_catalog/unity/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Databricks Unity Catalog - implementation of the `DataCatalog` trait
+pub struct UnityCatalog {
+    client_with_retry: ClientWithMiddleware,
+    workspace_url: String,
+}
+
+impl UnityCatalog {
+    /// Creates a new UnityCatalog
+    pub fn new() -> Result<Self, DataCatalogError> {
```
roeap, line 18: For our clients etc, we usually adopt a builder pattern, that allows us to configure options on the client before building the client. We would then have some methos like `UnitiCatalogClientBuilder::from_env()` to support parsing the configuration from the environment.

That said, haven't worked with our catalogs for a long time, so not sure if we properly pass things through right now to make that work. IF not, this maybe something we want to address in a a follow-up.

rust/src/data_catalog/unity/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+impl From<reqwest_middleware::Error> for DataCatalogError {
+    fn from(value: reqwest_middleware::Error) -> Self {
+        value.into()
+    }
+}
+
+impl From<reqwest::Error> for DataCatalogError {
+    fn from(value: reqwest::Error) -> Self {
+        reqwest_middleware::Error::Reqwest(value).into()
```
roeap, line 88: This seems a bit convoluted. We are wrapping the actual error in another error to then convert it again. Not sure how many variants the middleware error has, but maybe its viable to unpack the middleware error?

rust/src/data_catalog/unity/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+#[derive(Deserialize)]
+#[serde(untagged)]
+enum TableResponse {
+    Success { storage_location: String },
+    Error { error_code: String, message: String },
+}
+
+/// Possible errors from the unity-catalog/tables API call
+#[derive(thiserror::Error, Debug)]
+pub enum GetTableError {
```
roeap, line 61: Are we using these error variants somewhere?

While still very much a long term things, we have been trying to reduce the number of error variants we expose to users in this crate and also make them more actionable. Maybe we can make some common errors explicit - like 403?

roeap (commented):
Thanks for this great contribution @nohajc - left some initial comments, but will give this a more thorough review tomorrow.

rust/src/data_catalog/unity/mod.rs
nohajc: There should be only two variants, unpacking sounds like a good idea. I was just trying to solve a type checking error, didn't give it much thought...

rust/src/data_catalog/unity/mod.rs
nohajc: I did it in a similar way to the existing AWS glue module but I understand it's not perfect.

Didn't want to start refactoring existing code either. :D At least not yet.

rust/src/data_catalog/unity/mod.rs
nohajc: I could also convert `reqwest::Error` to `anyhow::Error` but then I'm losing type information... Not sure it's even a good practice to use anyhow in library code.

nohajc: I suppose it would be nice to unify some of the `DataCatalogError` variants across implementations but I'd prefer to do it separately.

roeap:
Had a look through our current code and thought a bit about where we might want to go with our catalogues in general. I landed on taking a similar approach to what we did in the object_store crate. There we went through quite some effort to make everything SDK-free. Good news is the unitiy catalog implementation already follows that pattern to some extend :).

The reasoning is quite similar. Our needs towards what we want from the APIs is usually a very small subset of what the overall catalog exposes, so implementing a few rest calls should be quite maintainable. Thought about just exposing the client from object_store, but I guess this would violate our goal over there to keep the API contact with consumers minimal and not expose any internals.

So how to go from there? I would be happy to merge this PR rather soon, but maybe we can rename the feature flag to something like `unitiy-experimental`. So we can do a bit of refactoring afterwards. Essentially this might entail:

* Porting over the http client from object store, and adopting it for the needs of the catalog APIs. (i.e. getting rid of `rusoto` dependency in glue client.
* Establishing a builders / configuration system like object_store (which is in turn what we use here for configuring storage)
* and some more stuff I likely missed :).

We could then also leverage the mock server implementation from over there to test the clients.

cc @wjones127 @rtyler @houqp @MrPowers

nohajc:
> 
> * Porting over the http client from object store, and adopting it for the needs of the catalog APIs. (i.e. getting rid of `rusoto` dependency in glue client.
> 
By porting over you mean duplicating the code or separating it into its own crate?

When I was going through the dependencies to find out what kind of http clients are used in the project, I saw that `object_store` already depends on `reqwest`. So, how does the client built on top of it compare to `reqwest-middleware` with `reqwest-retry`?

Anyway, I'd be happy to contribute to this refactoring effort if I find the time.

roeap:
> By porting over you mean duplicating the code or separating it into its own crate?

For now I would say, duplicating and adjusting, a new crate somewhere between an http client and the sdks brings not much value to the community - I think.

>  So, how does the client built on top of it compare to reqwest-middleware with reqwest-retry?

They probably do many things quite similar, but since we already have a well tested implementation that does not require these dependencies, we may as well use them :). More importantly though, we implemented quite a few different authorization mechanisms for the various cloud providers that we can leverage to also integrate with other catalogs like Pruview in Azure, for glue to to remove rusoto, or whatever google has.

In my experience, providing credentials for the sometimes quite different and complex ways of acquiring credentials in the various environments a code can run in, is usually the hardest part of this. So the main aim is to leverage as much of this as possible.

@tustvold - I just discounted the idea, that we want to make the http client and some credentials from `object_store` public, but maybe you have some thoughts on that? Maybe we could also find synergies with `datafusion` when it comes to implementing catalog clients?

nohajc:
> 
> More importantly though, we implemented quite a few different authorization mechanisms for the various cloud providers that we can leverage to also integrate with other catalogs like Pruview in Azure, for glue to to remove rusoto, or whatever google has.
> 
> In my experience, providing credentials for the sometimes quite different and complex ways of acquiring credentials in the various environments a code can run in, is usually the hardest part of this. So the main aim is to leverage as much of this as possible.
> 
This makes a lot of sense but I'd say complex auth flows are another reason for code reuse rather than duplication. Even if it didn't have value in the community when the SDKs already exist, it still has value as a common building block of the ecosystem you're building.

Just my two cents.

nohajc:
I renamed the feature flag as suggested.

roeap (approved):
Thanks @nohajc, from my end we can merge this.

I should get around to add some tests for this soon, and hopefully migrate the glue catalog.

I'll leave this open for a little bit in case others want to chime in.

dennyglee:
Thanks for this PR @nohajc - once this gets merged, please ping me via [Delta Users Slack](https://go.delta.io/slack) - `dennyglee` so we can send you some swag for your first Delta Lake contribution, eh?!  Thanks!

djouallah:
Does this support reading Databricks Delta table with merge on read, deletion vector etc

dennyglee:
> Does this support reading Databricks Delta table with merge on read, deletion vector etc

Hey @djouallah - good to hear from you!  As noted in this PR, it's supporting basic unity catalog integration hence it's' not tackling MoR, deletion vectors.  Why not create issues related to them as we have discussed deletion vector (and then afterwards MoR)  support but I don't think we've had a chance to create an issue for this.  And PRs are always welcome :)

ion-elgreco:
@nohajc I am not able to connect to the unity catalog with my code, how did you get it working?:
```python

import os
from deltalake import DataCatalog, DeltaTable

os.environ['DATABRICKS_WORKSPACE_URL'] = URL
os.environ['DATABRICKS_ACCESS_TOKEN'] = SECRET

catalog_name = 'private_catalog'
schema_name = 'private_schema'
table_name = 'private_table'
data_catalog = DataCatalog.UNITY
dt = DeltaTable.from_data_catalog(data_catalog=data_catalog, data_catalog_id=catalog_name, database_name=schema_name, table_name=table_name)

```

It results in this: 

```
PyDeltaTableError: Failed to load checkpoint: Failed to read checkpoint content: Generic MicrosoftAzure error: Error authorizing request: Error performing token request: response error "request error", after 0 retries: error sending request for url (http://<ip.address>/metadata/identity/oauth2/token?api-version=2019-08-01&resource=https%3A%2F%2Fstorage.azure.com%2F.default): error trying to connect: tcp connect error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. (os error 10060)
```

nohajc:
> @nohajc @roeap I am not able to connect to the unity catalog with my code, how did you get it working?

I didn't have to do anything special. My access token just worked. This could be any number of things... maybe firewall issue?

roeap:
From the error message it seems that communication with unity catalog worked, but the credentials to access the storage are not available.

We are working on improving those apis. For now you could likely use the environment to configure storage credentials.

At least thats my best guess based on the error message.

nohajc:
That's right, I forgot to mention I also had the azure storage account credentials in the environment. This is not explicitly said in the documentation but it is a requirement.

ion-elgreco:
> > @nohajc @roeap I am not able to connect to the unity catalog with my code, how did you get it working?
> 
> I didn't have to do anything special. My access token just worked. This could be any number of things... maybe a firewall issue?

Could be, the databricks workspaces are managed by our platform teams.



> That's right, I forgot to mention I also had the azure storage account credentials in the environment. This is not explicitly said in the documentation but it is a requirement.

Ah, that would make sense, I don't have this, nor would I get the account credentials from our platform team. I am only able to access the table with my personal credentials through the catalog.

Would it be possible to expose unity catalog tables without storage credentials to delta-rs?

ion-elgreco:
> > Would it be possible to expose unity catalog tables without storage credentials to delta-rs?
> 
> Well, I think you have to authenticate against the storage one way or another. If this is Azure, it could be as simple as running `az login` in your shell session. Assuming your user has the correct permissions.

Just to update this worked. Now I have two ways to read delta tables, directly on filepath or through UC schema's.

Issue #1336, opened 2023-05-05
partition_filter doesn't work in optimize in Python


wjones127:
# Environment

**Delta-rs version**: 0.9.0 (pre-release)

**Binding**: Python

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

This seems to affect inequalities but not equalities.

```python
import pyarrow as pa
import pyarrow.compute as pc
from deltalake import DeltaTable, write_deltalake
from datetime import datetime, timedelta
import itertools

def record_observations(date: datetime) -> pa.Table:
    """Pulls data for a certain datetime"""
    nrows = 1000
    return pa.table({
        "date": pa.array([date.date()] * nrows),
        "timestamp": pa.array([date] * nrows),
        "value": pc.random(nrows),
    })

hours_iter = (datetime(2021, 1, 1) + timedelta(hours=i) for i in itertools.count())

# Write 24 hours worth of data
for timestamp in itertools.islice(hours_iter, 24):
    write_deltalake(
        "observation_data",
        record_observations(timestamp),
        partition_by=["date"],
        mode="append",
    )

dt.file_uris(partition_filters=[("date", ">=", "2021-01-01")])

dt.optimize(partition_filters=[("date", ">=", "2021-01-01")])
```

```
['/Users/willjones/Documents/delta-rs/observation_data/date=2021-01-02/part-00000-752ca50d-bcc2-4129-90f0-82b235dce004-c000.snappy.parquet',
 '/Users/willjones/Documents/delta-rs/observation_data/date=2021-01-04/part-00000-94c0d79c-3529-43be-9c91-b4dbb459ce45-c000.snappy.parquet',
 '/Users/willjones/Documents/delta-rs/observation_data/date=2021-01-03/part-00000-36aaeffb-68b8-4851-b94c-aeadb62602f7-c000.snappy.parquet',
 '/Users/willjones/Documents/delta-rs/observation_data/date=2021-01-01/part-00000-9af11e90-9ee5-4304-8e20-9aeb28888205-c000.snappy.parquet',
 '/Users/willjones/Documents/delta-rs/observation_data/date=2021-01-06/part-00000-c858dc5f-d267-42ff-9b98-bf1caa241a5d-c000.snappy.parquet',
 '/Users/willjones/Documents/delta-rs/observation_data/date=2021-01-05/part-00000-24ecf869-0687-48d4-b6d0-0ba0e4509fe9-c000.snappy.parquet']
```

```
[...]

PR #1337, opened 2023-05-06
feat: update incremental after operations

wjones127:
# Description

It can be really confusing when writing to a Delta Table that afterwards the `DeltaTable` still needs to have `update_incremental` to reflect the changes. I felt this myself in #1336.

So this adds `update_incremental` at the end of write and optimize.

I also add it to the beginning of write, since I discovered you could get commit errors if you are passing in an old table state.

Also, made it so the `partition_by` argument can handle just a single string. (I think PyArrow does the same, which must be why I have muscle memory for it.)

# Related Issue(s)
No issue.

# Documentation

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -334,6 +334,7 @@ given filters.
         :return: the metrics from optimize
         """
         metrics = self._table.optimize(partition_filters, target_size)
+        self.update_incremental()
         return json.loads(metrics)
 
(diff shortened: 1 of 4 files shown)


rtyler (approved):

fvaleye (approved):

GraemeBrownTL:
@wjones127 In your opinion, would it be useful to add this same functionality to DeltaTable.delete()? I have noticed the same issue there.

PR #1338, opened 2023-05-06
feat: integrate unity catalog with datafusion

roeap:
# Description

This is a follow-up from #1331, where we added support for the unity catalog and does a couple of things:
- Adds a port of the client implementation from object_store to communicate with unity catalog.
- refactors and extends the current client to accept configuration much in the way we do for the storage backends
- adds additional (so far only Azure) credentials to authorize requests
- adds an integration with datafusion's `SchemaProvider` and `CatalogProvider`
- add unit tests for various components.

At first glance porting the whole client implementation may seem excessive, but my hope is that with this we can eventually replace the rusoto dependency from the glue catalog, as well as implement the `LogStore` locking mechanisms suggested by @nkarpov while keeping everything SDK-free :).

@nohajc - unfortunately I do not have access to a unity catalog instance right now. Would it be possible for you to validate, that the requests are still working? 

# Related Issue(s)

No issues


nohajc:
> @nohajc - unfortunately I do not have access to a unity catalog instance right now. Would it be possible for you to validate, that the requests are still working?

Sure, I will gladly test it.

rust/src/data_catalog/unity/credential.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub const AZURE_CHINA: &str = "https://login.chinacloudapi.cn";
+    /// Germany-based Azure Authority Host
+    pub const AZURE_GERMANY: &str = "https://login.microsoftonline.de";
+    /// US Government Azure Authority Host
+    pub const AZURE_GOVERNMENT: &str = "https://login.microsoftonline.us";
+    /// Public Cloud Azure Authority Host
+    pub const AZURE_PUBLIC_CLOUD: &str = "https://login.microsoftonline.com";
+}
+
+/// Trait for providing authrization tokens for catalog requests
```
nohajc, line 30: Typo in `authrization`

rust/src/data_catalog/unity/mod.rs
```diff
@@ -75,17 +36,416 @@ pub enum GetTableError {
...
+        let token = self.get_credential().await?;
+        // https://docs.databricks.com/api-explorer/workspace/tables/get
+        let resp = self
+            .client
+            .get(format!(
+                "{}/tables/{}.{}.{}",
+                self.catalog_url(),
+                catalog_id
+                    .map(|c| c.as_ref().to_string())
+                    .unwrap_or("main".into()),
```
nohajc, line 440: You should be able to do this instead:
```suggestion
                    .as_ref()
                    .map(|c| c.as_ref())
                    .unwrap_or("main"),
```

rust/src/data_catalog/unity/models.rs
```diff
@@ -0,0 +1,367 @@
...
+
+/// Get table response
+#[derive(Deserialize)]
+#[serde(untagged)]
+pub enum GetTableResponse {
+    /// Successful response
+    Success(Table),
+    /// Error response
+    Error(ErrorResponse),
+}
```
nohajc, line 24: I tested it and unfortunately `Error(ErrorResponse)` is not equivalent to `Error { error_code: String, message: String }`, so right now it fails to parse (the same for `Success(Table)`.

So far I haven't found any other way other than defining the structs inline, like it was before.

I thought the following might work but it didn't.
```
#[derive(Deserialize)]
#[serde(untagged)]
pub enum GetTableResponse {
    /// Successful response
    Success {
        /// Table information
        #[serde(flatten)]
        table: Table,
    },
    /// Error response
    Error {
        /// Error information
        #[serde(flatten)]
        err: ErrorResponse,
    },
}
```

rust/src/data_catalog/unity/mod.rs
```diff
@@ -75,17 +36,416 @@ pub enum GetTableError {
...
+    type Err = DataCatalogError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "workspace_url" | "unity_workspace_url" | "databricks_workspace_url" => {
+                Ok(UnityCatalogConfigKey::WorkspaceUrl)
+            }
+            "access_token" | "unity_access_token" | "databricks_access_token" => {
+                Ok(UnityCatalogConfigKey::AccessToken)
+            }
```
nohajc, line 121: Seems like this should also match AZURE_TENANT_ID, AZURE_CLIENT_ID and AZURE_CLIENT_SECRET, right?
Btw, I wouldn't use the `unity_` alias for Azure tenant and client because it's not Databrick/Unity specific.

rust/src/data_catalog/unity/models.rs
roeap: hmm .. if we have to we'll do it, but this makes the types a bit annoying to work with.

The example responses from the docs parse fine in the tests though. would it be possible to provide a response body that doe snot parse? Admittedly, the error case I just inferred from your code :).

nohajc: Ok, now I'm even more confused. From the test, it looks like it should work. I don't know what serde is doing but it didn't parse the real response before I changed the definition back.

I'll try to write a test case to reproduce the problem.

rust/src/data_catalog/unity/models.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub created_by: String,
+
+    /// Name of table, relative to parent schema.
+    pub name: String,
+
+    /// Username of user who last modified the table.
+    pub updated_by: String,
+
+    /// List of schemes whose objects can be referenced without qualification.
+    pub sql_path: String,
```
nohajc, line 160: This is the culprit. My Unity Catalog returns response without any `sql_path`.
```suggestion
    #[serde(default)]
    pub sql_path: String,
```
I fixed it like this but looking at the documentation, it looks like most of the response fields could be optional (even though it doesn't make much sense in some cases).

rust/src/data_catalog/unity/models.rs
nohajc: So, the error cause was different than I thought (see the comment in `struct Table`). At least the definitions don't have to be inlined.

rust/src/data_catalog/unity/models.rs
```diff
@@ -0,0 +1,367 @@
...
+            "updated_at": 0,
+            "metastore_id": "string"
+        }"#;
+
+    pub(crate) const GET_TABLE_RESPONSE: &str = r#"
+    {
+        "created_by": "string",
+        "name": "table_name",
+        "updated_by": "string",
+        "sql_path": "string",
```
nohajc, line 240: We could add a test case with omitted `sql_path` but who knows which other fields can be missing from the response under different circumstances...

nohajc:
@roeap  So far I only verified it works with Databricks personal access token but if I figure out which AAD roles are needed, I could check the Oauth flows (+ azure cli) as well.

rust/src/data_catalog/unity/models.rs
roeap: I guess we can just make a bunch of these fields optional. the ones we actually need are quite limited and should always be available - like name etc..

rust/src/data_catalog/unity/mod.rs
roeap: I have been going back and forth on this and hope to experiment a bit here.

The main nuisance I encountered was that in scenarios where you are juggling many connections sometime we pick up things from the environment that aren't intended.

Generally speaking though yes, these are quite common and users would likely expect this to be picked up.

roeap:
> So far I only verified it works with Databricks personal access token but if I figure out which AAD roles are needed, I could check the OAuth flows (+ azure cli) as well.

That would be awesome. Haven't worked with this yet. but from the docs i gather you don't really need specific roles, but rather "Just" need the AAD object user to be available / assigned in the unity catalog.

rust/src/data_catalog/unity/models.rs
roeap: WE now just derive default for `Schema` and `Table`

nohajc: I rechecked with the latest changes and deriving Default for the struct does not change the deserializer behavior.
The `#[serde(default)]` macro still needs to be used either per field or on the whole struct making all the fields optional with empty default. An alternative would be to use `Option<String>` where it makes sense.

jordandakota:
> That would be awesome. Haven't worked with this yet. but from the docs i gather you don't really need specific roles, but rather "Just" need the AAD object user to be available / assigned in the unity catalog.

This is correct. If you have an AAD bearer token with scope set to the azure databricks application you can do everything a databricks personal access token can do. The scope being -> 2ff814a6-3304-4ab8-85cb-cd0e6f879c1d

rust/src/data_catalog/unity/mod.rs
```diff
@@ -75,17 +37,548 @@ pub enum GetTableError {
...
+                Default::default(),
+                Box::new(ClientSecretOAuthProvider::new(
+                    client_id,
+                    client_secret,
+                    authority_id,
+                    self.authority_host.as_ref(),
+                )),
+            ));
+        }
+
```
nohajc, line 398: Tried with Azure CLI auth after adding the following and it works fine.
```suggestion
+        if self.use_azure_cli {
+            return Some(CredentialProvider::TokenCredential(
+                Default::default(),
+                Box::new(AzureCliCredential::new()),
+            ));
+        }
+
```

rust/src/data_catalog/unity/mod.rs
```diff
@@ -75,17 +37,548 @@ pub enum GetTableError {
...
+            self.client_id.as_ref(),
+            self.client_secret.as_ref(),
+            self.authority_id.as_ref(),
+        ) {
+            return Some(CredentialProvider::TokenCredential(
+                Default::default(),
+                Box::new(ClientSecretOAuthProvider::new(
+                    client_id,
+                    client_secret,
+                    authority_id,
```
nohajc, line 393: Using Azure service principal account works (specifying Azure tenant ID as `authority_id` and account appId with password as `client_id` and `client_secret` respectively).

nohajc:
> That would be awesome. Haven't worked with this yet. but from the docs i gather you don't really need specific roles, but rather "Just" need the AAD object user to be available / assigned in the unity catalog.

Successfully tested service principal (client ID / client secret) and Azure CLI credentials. I don't have an opportunity to try with managed identity or workload identity right now...

roeap:
Thanks for testing this @nohajc!

> I don't have an opportunity to try with managed identity or workload identity right now...

I guess we know now, that we are using the right scopes, the rest is just how we get the AAD token, that we have validated elsewhere, so we should be good 🤞.

nohajc:
(I still need to use `#[serde(default)]` on the `sql_path` response field.)

roeap:
> (I still need to use #[serde(default)] on the sql_path response field.)

on it :) - applying some fixes and adding datafusion tests.

ion-elgreco:
@roeap Is this ready to get merged?

rtyler (approved):

PR #1342, opened 2023-05-07
fix: include stats for all columns (#1223)

### LINKED ISSUE #1223: .get_add_actions() returns wrong column statistics when dataSkippingNumIndexedCols property of the table was changed
# Environment

**Delta-rs version**:
0.7.0

**Binding**: 
Python

***
# Bug

**What happened**:
`.get_add_actions() `returns wrong column statistics when dataSkippingNumIndexedCols property of the table was changed.
It returns only the stats for columns from the latest file.

It should return all the stats for all columns of all the files.

**How to reproduce it**:

Had to do it in delta-spark, since I don't think there is a way of setting table properties in delta-rs.

```
path = f"./latestversion"

data = [
    (1, "a", None),
    (2, "b", "b"),
    (3, "c", "c"),
]
df = spark.createDataFrame(
    data,
    [
        "col1",
        "col2",
        "col3",
    ],
)
    
df.coalesce(1).write.mode("append").format("delta").save(path)

# .get_add_actions() will return min, max and null_count for col1, col2 and col3

spark.sql("ALTER TABLE delta.`FULL_PATH` SET TBLPROPERTIES(delta.dataSkippingNumIndexedCols = 1);")

df.coalesce(1).write.mode("append").format("delta").save(path)

# .get_add_actions() will only return min, max and null_count for col1
```

mrjoe7:
# Description
This is a proposal for how #1223 could be fixed.

# Related Issue(s)
- fixes #1223

# Documentation
The current implementation excludes all columns that lack statistical information. The proposed fix will generate information for all columns, with missing statistical values being replaced by 'null' values. However, it is unclear if this is the correct behavior since the `stats_as_batch` function lacks documentation.

### DIFF
--- a/rust/src/table_state_arrow.rs
+++ b/rust/src/table_state_arrow.rs
@@ -9,8 +9,8 @@ use crate::DeltaTableError;
 use crate::SchemaDataType;
 use crate::SchemaTypeStruct;
 use arrow::array::{
-    ArrayRef, BinaryArray, BooleanArray, Date32Array, Float64Array, Int64Array, StringArray,
-    StructArray, TimestampMicrosecondArray, TimestampMillisecondArray,
+    Array, ArrayRef, BinaryArray, BooleanArray, Date32Array, Float64Array, Int64Array, NullArray,
+    StringArray, StructArray, TimestampMicrosecondArray, TimestampMillisecondArray,
 };
 use arrow::compute::cast;
 use arrow::compute::kernels::cast_utils::Parser;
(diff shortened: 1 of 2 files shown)


wjones127:
Thanks for working on this.

I'm not sure, but I'm thinking for columns that don't have _any_ statistics, we should either:

1. exclude them from the output
2. use `NullArray` for their stats, to minimize memory overhead.

roeap:
My 2 cents would be to exclude them, since delta allows for a config how many columns to collect metrics for (which delta-rs does not yet honor :D), my expectation would be to only get metrics for these columns. This defaults to the first 32 columns.

https://learn.microsoft.com/en-us/azure/databricks/delta/table-properties (`delta.dataSkippingNumIndexedCols`)

mrjoe7:
Updated the PR to not include cols without any stats.

code to verify:
```python
import pyspark
from delta import *

builder = pyspark.sql.SparkSession.builder.appName("MyApp") \
    .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension") \
    .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog")

spark = configure_spark_with_delta_pip(builder).getOrCreate()

location = "/tmp/delta/people10m"

DeltaTable.createIfNotExists(spark).addColumn("firstName", "STRING").addColumn("lastName", "STRING").addColumn("gender", "STRING").property("description", "table with people data").location(location).execute()

# create stats for firstName, lastName
spark.sql(f"ALTER TABLE delta.`{location}` SET TBLPROPERTIES(delta.dataSkippingNumIndexedCols = 2);")

columns = ["firstName", "lastName", "gender"]

data = [("Maria", "Dimas", "female"), ("John", "Francis", "male"), ("Frank", "Sinatra", "male")]
rdd = spark.sparkContext.parallelize(data)
df = rdd.toDF(columns)
df.coalesce(1).write.format("delta").mode("append").save(location)

# This adds a new update to the table, but this time all three columns will have stats.
spark.sql(f"ALTER TABLE delta.`{location}` SET TBLPROPERTIES(delta.dataSkippingNumIndexedCols = 3);")
data = [("Martin", "Johnson", "male"), ("Victoria", "Neal", "female")]
rdd = spark.sparkContext.parallelize(data)
df = rdd.toDF(columns)
df.coalesce(1).write.format("delta").mode("append").save(location)
```

```rust
async fn main() -> Result<(), deltalake::DeltaTableError> {
    let table_path = "/tmp/delta/people10m";
    let table = deltalake::open_table(table_path).await?;
    println!("{table}");
    let acts = table.get_state().add_actions_table(true).unwrap();
    dbg!(acts);
    Ok(())
}
```

rust/src/table_state_arrow.rs
```diff
@@ -304,56 +304,81 @@ impl DeltaTableState {
...
                             .as_ref()
                             .map(|stat| resolve_column_count_stat(&stat.null_count, &path))
                     })
-                    .collect::<Option<Vec<DeltaDataTypeLong>>>()
-                    .map(arrow::array::Int64Array::from)
-                    .map(|arr| -> ArrayRef { Arc::new(arr) });
+                    .collect::<Vec<Option<DeltaDataTypeLong>>>();
+                let null_count: Option<ArrayRef> = empty_stats_into_null_array(Some(Arc::new(
+                    arrow::array::Int64Array::from(null_count),
+                )));
```
wjones127, line 341: We can save allocations if we detect the all-null case directly from `null_count`, rather than after we create the `Int64Array`. Same goes for the min and max below.

mrjoe7: Updated the PR

wjones127 (approved):
Looks good. Thanks!

PR #1349, opened 2023-05-09
feat(python): add filters argument to DeltaTable.to_pandas() for filter pushdown

### LINKED ISSUE #1316: feat(python): add filters argument to DeltaTable.to_pandas() for filter pushdown
Reading Delta tables into pandas DataFrames could be more intuitive.

Here's the current syntax:

```python
import pyarrow.dataset as ds

dt = DeltaTable(f"{pathlib.Path.home()}/data/delta/G1_1e8_1e2_0_0")
dataset = dt.to_pyarrow_dataset()
condition = ds.field("id1") == "id016"
df = dataset.to_table(filter=condition, columns=["id1", "id2", "v1"]).to_pandas()
```

Could we provide an interface that's more similar to `pandas.read_parquet`?  Something like this:

```
import deltalake

dt = DeltaTable(f"{pathlib.Path.home()}/data/delta/G1_1e8_1e2_0_0")
df = dt.filters([("id1", "==", "id016")]).to_pandas()
```

The list of tuples of how predicates are passed in `pandas.read_parquet`.

I'm open to other interfaces too.  I think something that hides `pyarrow.dataset`, is less code, and is familiar with existing pandas syntax would be good.

ognis1205:
# Description

This pull request adds support for filtering to the `to_pandas` method of `DeltaTable`. The new filters argument allows
users to specify filtering criteria in a format that is compatible with `pyarrow.compute.Expression`.

- Adding the `filters` argument to `DeltaTable.to_pandas`.
- Adding the `_filters_to_expression` function to `table` module, which is based on [this implementation](https://github.com/apache/arrow/blob/b9cc5df8a4f7c7fe09f40ba92a74981dee5e536a/python/pyarrow/parquet/core.py#LL155C5-L155C26), but with improved type consistency.
- ~~Based on [the existing conventional unit tests](https://github.com/delta-io/delta-rs/blob/b5230835bc1d1b01d59da3649033f1180232ddb7/python/tests/test_table_read.py#L392), I did not add any additional test cases for this feature. Instead, I tested the feature on my local development environment.~~
- Adding a unit test for the feature.

# Related Issue(s)

- closes #1316

# Documentation

[The DNF filter format](https://github.com/apache/arrow/blob/b9cc5df8a4f7c7fe09f40ba92a74981dee5e536a/python/pyarrow/parquet/core.py#LL155C5-L155C26).

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, Tuple, 
 import pyarrow
 import pyarrow.fs as pa_fs
 from pyarrow.dataset import FileSystemDataset, ParquetFileFormat, ParquetReadOptions
+from pyarrow.parquet.core import filters_to_expression
 
 if TYPE_CHECKING:
     import pandas
@@ -412,6 +413,9 @@ given filters.
         partitions: Optional[List[Tuple[str, str, Any]]] = None,
         columns: Optional[List[str]] = None,
         filesystem: Optional[Union[str, pa_fs.FileSystem]] = None,
+        filters: Optional[
+            Union[List[Tuple[str, str, Any]], List[List[Tuple[str, str, Any]]]]
+        ] = None,
     ) -> pyarrow.Table:
         """
         Build a PyArrow Table using data from the DeltaTable.
@@ -419,17 +423,23 @@ given filters.
         :param partitions: A list of partition filters, see help(DeltaTable.files_by_partitions) for filter syntax
         :param columns: The columns to project. This can be a list of column names to include (order and duplicates will be preserved)
         :param filesystem: A concrete implementation of the Pyarrow FileSystem or a fsspec-compatible interface. If None, the first file path will be used to determine the right FileSystem
+        :param filters: A disjunctive normal form (DNF) for filter pushdown. The filters are expressed in a format that is compatible with `pyarrow.compute.Expression`
         :return: the PyArrow table
         """
+        if filters is not None:
+            filters = filters_to_expression(filters)
         return self.to_pyarrow_dataset(
             partitions=partitions, filesystem=filesystem
-        ).to_table(columns=columns)
+        ).to_table(columns=columns, filter=filters)
 
     def to_pandas(
         self,
         partitions: Optional[List[Tuple[str, str, Any]]] = None,
         columns: Optional[List[str]] = None,
         filesystem: Optional[Union[str, pa_fs.FileSystem]] = None,
+        filters: Optional[
+            Union[List[Tuple[str, str, Any]], List[List[Tuple[str, str, Any]]]]
+        ] = None,
     ) -> "pandas.DataFrame":
         """
         Build a pandas dataframe using data from the DeltaTable.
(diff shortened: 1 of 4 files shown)


ognis1205:
I will check this error:

https://github.com/delta-io/delta-rs/actions/runs/4926020116/jobs/8800992032?pr=1349#step:8:18

ognis1205:
https://github.com/delta-io/delta-rs/actions/runs/4943095151/jobs/8837254340?pr=1349#step:5:545

python/deltalake/table.py
```diff
@@ -412,35 +521,48 @@ def to_pyarrow_table(
...
+            Union[List[Tuple[str, str, Any]], List[List[Tuple[str, str, Any]]]]
+        ] = None,
     ) -> pyarrow.Table:
         """
         Build a PyArrow Table using data from the DeltaTable.
 
         :param partitions: A list of partition filters, see help(DeltaTable.files_by_partitions) for filter syntax
         :param columns: The columns to project. This can be a list of column names to include (order and duplicates will be preserved)
         :param filesystem: A concrete implementation of the Pyarrow FileSystem or a fsspec-compatible interface. If None, the first file path will be used to determine the right FileSystem
+        :param filters: A disjunctive normal form (DNF) for filter pushdown. The filters are expressed in a format that is compatible with `pyarrow.compute.Expression`
```
wjones127, line 534: I don't think we need to mention PyArrow or pushdown here, as it's an implementation detail, right? But we should be clear that unlike `partitions` it filters rows.

```suggestion
        :param filters: A disjunctive normal form (DNF) predicate for filtering rows. If you pass a filter you do not need to pass ``partitions``.
```

python/deltalake/table.py
```diff
@@ -412,35 +521,48 @@ def to_pyarrow_table(
...
+        ).to_table(columns=columns, filter=filters)
 
     def to_pandas(
         self,
         partitions: Optional[List[Tuple[str, str, Any]]] = None,
         columns: Optional[List[str]] = None,
         filesystem: Optional[Union[str, pa_fs.FileSystem]] = None,
+        filters: Optional[
+            Union[List[Tuple[str, str, Any]], List[List[Tuple[str, str, Any]]]]
+        ] = None,
```
wjones127, line 550: Could we pull out the DNF filter type definition into a variable above? I feel like that would help with readability.

wjones127 (changes requested):
I'd still like you to add a test at least for `to_pyarrow_table()` exercising the new argument. Ideally it should include a case for a conjunctionm one with a disjunction, and one single predicate.

python/deltalake/table.py
ognis1205: > Ideally it should include a case for a conjunctionm one with a disjunction, and one single predicate.
Sure thing! I will add some unit tests.

> I don't think we need to mention PyArrow or pushdown here, as it's an implementation detail, right? But we should be clear that unlike partitions it filters rows.
Agreed. I will fix this.

python/deltalake/table.py
ognis1205: > I feel like that would help with readability.

Yes, we can pull out it into variable. I will fix this.

ognis1205:
@wjones127 

Sorry I had forgotten to mention you, but I believe I have now resolved all the issues. Would it be possible for you to review the PR again when you have a moment? I would really appreciate it. Thank you!

wjones127 (approved):
Thanks, those tests look great!

wjones127 (approved):
Thanks, those tests look great!

Issue #1358, opened 2023-05-12
Support HDFS in Python package


Kimahriman:
# Description

Add support for HDFS in Python now that it's supported in Rust.

**Use Case**
Using deltalake Python package to read Delta tables in HDFS

**Related Issue(s)**
https://github.com/delta-io/delta-rs/issues/300

roeap:
Unfortunately, HDFS driver requires JAVA, and one of the core interests is to not depend on that. so closing this as we right now see no way to provide this without java.

SchutteJan:
@Kimahriman 

I see you are working on a Rust implementation for hdfs (No JAVA!) in https://github.com/Kimahriman/hdfs-native and an object store implementation in https://github.com/datafusion-contrib/hdfs-native-object-store

Are you also working on integrating your work into delta-rs? I am experimenting with that myself (https://github.com/SchutteJan/delta-rs) and I am curious what your plans are.

Kimahriman:
> Are you also working on integrating your work into delta-rs? I am experimenting with that myself (https://github.com/SchutteJan/delta-rs) and I am curious what your plans are.

Yeah I do have a branch I've been working on, just been lazy on getting it cleaned up and making a PR. The biggest limitation right now is the libgssapi dependency, which adds a dynamic link and requires some changes to how the wheels would be built. I think I'll try to get the Rust side finished up and make a PR with an option to build a custom Python wheel. I'm currently working on an update to do all gssapi/kerberos things via runtime dynamic loading with `libloading`. If I can get that working it should simplify building wheels/cross compilation and theoretically could be added to the published wheels if it's accepted.

Kimahriman:
https://github.com/delta-io/delta-rs/pull/2612

PR #1359, opened 2023-05-12
feat: add hdfs support in python

Kimahriman:
# Description

Adds support for HDFS in the Python package by enabling the `hdfs` feature from the Rust library. Additionally adds a simple integration test, and docker-compose support to run the integration test locally.

Had to update some of the workflows to accommodate. The trickiest one was the Python 3.7 stage that uses an old OS, which doesn't have a new enough clang library to compile the HDFS object store I guess?

This might require Java to build and develop on any Python related thing now? Not totally sure what side effects this might induce, first time playing around with this library (avid Spark user).

# Related Issue(s)

Closes https://github.com/delta-io/delta-rs/issues/1358

# Documentation

### DIFF
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -39,7 +39,7 @@ features = ["extension-module", "abi3", "abi3-py37"]
 [dependencies.deltalake]
 path = "../rust"
 version = "0"
-features = ["azure", "gcs", "python", "datafusion", "unity-experimental"]
+features = ["azure", "gcs", "python", "datafusion", "unity-experimental", "hdfs"]
 
 [features]
 default = ["rustls"]
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -39,9 +39,10 @@ features = ["extension-module", "abi3", "abi3-py37"]
 [dependencies.deltalake]
 path = "../rust"
 version = "0"
-features = ["azure", "gcs", "python", "datafusion", "unity-experimental", "hdfs"]
+features = ["azure", "gcs", "python", "datafusion", "unity-experimental"]
 
 [features]
 default = ["rustls"]
+hdfs = ["deltalake/hdfs"]
 native-tls = ["deltalake/s3-native-tls", "deltalake/glue-native-tls"]
 rustls = ["deltalake/s3", "deltalake/glue"]
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -50,6 +50,11 @@ jobs:
           toolchain: stable
           override: true
           components: rustfmt, clippy
+      
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '17'
 
       - uses: Swatinem/rust-cache@v2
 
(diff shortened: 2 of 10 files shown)


roeap:
Thanks for doing this @Kimahriman. Could we add some documentation for users on what the prerequisites are for using this?

However one thing you mentioned made me a little concerned, but lets make sure I got this right. Would this mean that we always require JAVA to be available when building the wheels and or installing the library?

In that case I think we would have to find another way to make HDFS available on the python side. For me not requiring Java is one of the main driver to even get involved in the project 😆, and having that as a requirement would break a lot of use cases.

Kimahriman:
Yeah it's kinda unfortunate. The pyarrow approach would be interesting since that's how all the data files would be read at that point right? I'm curious if building pyarrow requires Java for however they do the libhdfs bindings? 

Definitely not my area of expertise, just a pyspark power user that thought this would be interesting to figure out.

python/Cargo.toml
```diff
@@ -39,7 +39,7 @@ features = ["extension-module", "abi3", "abi3-py37"]
 [dependencies.deltalake]
 path = "../rust"
 version = "0"
-features = ["azure", "gcs", "python", "datafusion", "unity-experimental"]
+features = ["azure", "gcs", "python", "datafusion", "unity-experimental", "hdfs"]
```
wjones127, line 42: Could we make this an optional feature of the Python library, since it has that extra java dependency?

You can add below:

```toml
hdfs = ["deltalake/hdfs"]
```

And then in Python release add it to the feature flags for release:

https://github.com/delta-io/delta-rs/blob/2077e6a6d0c47f74c83c7bc908517a02a6ecefee/.github/workflows/python_release.yml#L14

wjones127 (commented):
Could we make `hdfs` an optional (and non-default) feature for the Python package?

Seems like it adds Java as a build dependency. That's okay, if it's not default. But does it mean it's also a runtime dependency for users who aren't using HDFS? If not, I'm fine including in wheels by default. Otherwise, we might be better off finishing the fsspec wrapper.

Kimahriman:
Yeah I think that all would work! I made it an optional dependency that gets installed for the integration tests, didn't touch the release workflows.

Kimahriman:
Well nevermind, I think I was wrong. I used a fresh docker container, installed Java, built with HDFS support, then uninstalled Java and the unit tests fail with
```
ImportError while loading conftest '/data/python/tests/conftest.py'.
tests/conftest.py:11: in <module>
deltalake/__init__.py:1: in <module>
    from ._internal import PyDeltaTableError, RawDeltaTable, __version__, rust_core_version
E   ImportError: libjvm.so: cannot open shared object file: No such file or directory
```

so this might just not be viable, might only be able to do the pyarrow based approach.

wjones127:
That's unfortunate. Either datafusion-objectstore-hdfs needs to change to load the library when called (using libloading) or we'll need to instead support fsspec.

rtyler:
Given the state of this work, I am going to convert it to a draft since I do not believe it would be suitable to merge at this point.

Kimahriman:
Closing in favor of https://github.com/delta-io/delta-rs/pull/2612

PR #1360, opened 2023-05-12
feat: port docs to mdBook

MrPowers:
This PR ports the user guide to mdBooks.

This should make it easier to update the docs.  Markdown files are more familiar than RST (restructured text).

mdBook also generates default URLs that are more SEO friendly.

mdBook also generates docs that are nicer looking ;)

We still need to figure out how to continue generating the API docs and properly link to them with mdBook.  Let's collaborate on next steps!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ Cargo.lock
 !/delta-inspect/Cargo.lock
 !/proofs/Cargo.lock
 
+docs/book
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,21 @@
+# MrPowers Book
+
+This book is built with [mdBook](https://github.com/rust-lang/mdBook).
+
+This book is hosted here: https://mrpowers.github.io/
+
+## Deploy
+
+Build the latest files with `mdbook build`.
+
+Manually copy these files into the [mrpowers.github.io](https://github.com/MrPowers/mrpowers.github.io) repository.
+
+Yea, I know, it's not a great deploy process :|
+
+## Install
+
+Download the `mdbook` binary.
+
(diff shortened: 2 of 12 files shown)


wjones127:
I know this comment is a little late: but have we considered [mkdocs](https://www.mkdocs.org/) instead of mdBook? It's still markdown, but it has an autodoc plugin, so it can also handle the API reference section. That way we can still have cross references in the User Guide, which I think is important. (It also occurred to me I was mixing these up in my head earlier since the names are so similar :facepalm:)

mkdocs is used by [Ibis](https://ibis-project.org/) and [Polars](https://pola-rs.github.io/polars-book/user-guide/) (also Pydantic and FastAPI). Ibis has user guide and API reference, while Polars just has a user guide in it. I could help setup the API reference part of that.

rtyler:
@MrPowers I think this would be something good for us to tidy up and land before Data and AI Summit. Personally I would like to have a unified book with different sections for Rust and Python rather than simply Python in our root `docs/` directory. Are you open to me making that change? 

I don't really see much reason not to land this work in progress and continue to iterate however

MrPowers:
@rtyler - yea, I'd love getting this out too.

I'm cool with @wjones127 suggestion to use mkdocs.  I don't have the bandwidth to work on the mkdocs implementation, but would fully support someone else building it (and copying the markdown files in this PR if that'd be useful).

The only point that I am really passionate about is choosing a technology that properly structures URLs / HTML and organizing the content in a SEO friendly manner.

MrPowers:
Closing in favor of a mkdocs PR: https://github.com/delta-io/delta-rs/pull/1548

rtyler:
I believe this was supposed to be closed previously

PR #1373, opened 2023-05-16
fix: add conversion for string for `Field::TimestampMicros` (#1372)

### LINKED ISSUE #1372: Unable to read timestamp fields from column statistics
# Environment

**Delta-rs version**: 0.10.0

**Binding**: rust

**Environment**:
N/A

***
# Bug

**What happened**:

With a field defined as 
```rs
SchemaField::new(
  "EdgeEndTimestamp".to_string(),
  SchemaDataType::primitive("timestamp".to_string()),
  true,
  HashMap::new(),
),
```
 
I get the following error when querying using datafusion.

```json
{
  "timestamp": "2023-05-16T18:58:40.206275Z",
  "level": "WARN",
  "fields": {
    "message": "Unexpected type when parsing min/max values for EdgeEndTimestamp. Found 2023-05-16 18:56:10 +00:00",
    "log.target": "deltalake::action::parquet_read",
    "log.module_path": "deltalake::action::parquet_read",
    "log.file": "/root/.cargo/registry/src/github.com-1ecc6299db9ec823/deltalake-0.10.0/src/action/parquet_read/mod.rs",
    "log.line": 216
  },
  "target": "deltalake::action::parquet_read"
}
```

Query to succeed, statistics read correctly.

**How to reproduce it**:

1. define a timestamp field
2. write data
3. query back data

**More details**:

It looks like the format might have changed for timestamp fields? It no longer looks like RFC3339. I have some data written using this (maybe incorrect) format. Would be great if the fix also supports this format too

cmackenzie1:
# Description

Add a match statement to handle the case when timestamp is encoded as `TimestampMicros` in the parquet file.

# Related Issue(s)

- resolves #1372.

# Documentation
N/A

### DIFF
--- a/rust/src/action/parquet_read/mod.rs
+++ b/rust/src/action/parquet_read/mod.rs
@@ -260,6 +260,9 @@ fn primitive_parquet_field_to_json_value(field: &Field) -> Result<serde_json::Va
             )),
             _ => Err("Invalid type for min/max values."),
         },
+        Field::TimestampMicros(timestamp) => Ok(serde_json::Value::String(
+            convert_timestamp_micros_to_string(*timestamp)?,
+        )),
         Field::TimestampMillis(timestamp) => Ok(serde_json::Value::String(
             convert_timestamp_millis_to_string(*timestamp)?,
         )),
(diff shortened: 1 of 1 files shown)


rust/src/action/parquet_read/mod.rs
```diff
@@ -269,13 +272,27 @@ fn primitive_parquet_field_to_json_value(field: &Field) -> Result<serde_json::Va
...
 
+fn convert_timestamp_micros_to_string(value: i64) -> Result<String, &'static str> {
+    let seconds = value / 1_000_000;
+    let microseconds = (value % 1_000_000) as u32;
+
+    let dt: chrono::DateTime<Utc> = Utc
+        .timestamp_opt(seconds, microseconds * 1_000)
+        .single()
+        .ok_or("Value out of bounds")?;
+    Ok(dt.to_rfc3339_opts(SecondsFormat::Micros, true))
```
cmackenzie1, line 293: Do we want `SecondsFormat::Micros` here or match with Millisecond format as above?

wjones127: Seems like micros is fine. Spec specifically allows that one. https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization

wjones127 (approved):

roeap (approved):

PR #1377, opened 2023-05-18
docs: updated schema.rs to use the right signature for decimal data type in documentation

rahulj51:
# Description
docs: Minor change in the docs. The documentation says Decimal(\<scale\>, \<precision\>) but it should be Decimal(\<precision\>, \<scale\>)

# Related Issue(s)

# Documentation


roeap (approved):
Thanks for fixing this @rahulj51!

PR #1381, opened 2023-05-20
feat: add datafusion storage catalog

roeap:
# Description
Adds a datafusion `SchemaProvider` implementation based on discovering Delta tables in a storage location via `_delta_log` folders.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -21,7 +21,6 @@ async-trait = "0.1"
 bytes = "1"
 chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
 cfg-if = "1"
-dashmap = "5"
 datafusion-objectstore-hdfs = { version = "0.1.3", default-features = false, features = [
     "hdfs3",
     "try_spawn_blocking",
@@ -69,6 +68,7 @@ reqwest-middleware = { version = "0.2.1", optional = true }
 reqwest-retry = { version = "0.2.2", optional = true }
 
 # Datafusion
+dashmap = { version = "5", optional = true }
 datafusion = { version = "24", optional = true }
 datafusion-expr = { version = "24", optional = true }
 datafusion-common = { version = "24", optional = true }
@@ -118,6 +118,7 @@ datafusion = [
     "sqlparser",
     "arrow",
     "parquet",
+    "dashmap",
 ]
 datafusion-ext = ["datafusion"]
 gcs = ["object_store/gcp"]
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -21,6 +21,7 @@ async-trait = "0.1"
 bytes = "1"
 chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
 cfg-if = "1"
+dashmap = "5"
 datafusion-objectstore-hdfs = { version = "0.1.3", default-features = false, features = [
     "hdfs3",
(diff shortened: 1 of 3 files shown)


rust/Cargo.toml
```diff
@@ -21,6 +21,7 @@ async-trait = "0.1"
 bytes = "1"
 chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
 cfg-if = "1"
+dashmap = "5"
```
rtyler, line 24: This is only needed for Datafusion, so can it be made optional

rust/src/data_catalog/storage/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// assuming it contains valid deltalake data, i.e a `_delta_log` folder:
+/// s3://host.example.com:3000/data/tpch/customer/_delta_log/
+pub struct ListingSchemaProvider {
+    authority: String,
+    path: object_store::path::Path,
+    /// Underlying object store
+    store: Arc<dyn ObjectStore>,
+    /// A map of table names to a fully quilfied storage location
+    tables: DashMap<String, String>,
```
rtyler, line 38: Is there an artifact of implementing this `SchemaProvider` that requires the use of DashMap? Since this adds a new dependency, I'm curious what forces this requirement since it's not clear to me

roeap: We could also go the "classic" route using `Arc<Mutex<...>>`, however since datafusion internally switched in most places, we already have that as a transient dependency if we use datafusion.

It does however simplify all code where we use this.

rust/Cargo.toml
roeap: I initially planned to also implement our trait. but then also thought we should maybe slightly update ours .. but yes, this can (and should) right now be optional.

rust/src/data_catalog/storage/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+impl ListingSchemaProvider {
+    /// Create a new [`ListingSchemaProvider`]
+    pub fn try_new(
+        root_uri: impl AsRef<str>,
+        storage_options: Option<HashMap<String, String>>,
+    ) -> DeltaResult<Self> {
+        let uri = ensure_table_uri(root_uri)?;
+        let authority = uri.to_string();
```
rtyler, line 49: Why is `uri` turned into a String and then parsed again to be passed into `configure_store`? Can `authority` be removed here and the function just pass along `uri` for configuring the store?

rust/src/data_catalog/storage/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+        let storage_options = storage_options.unwrap_or_default().into();
+        // We already parsed the url, so unwrapping is safe.
+        let store = configure_store(&url::Url::parse(&authority).unwrap(), &storage_options)?;
+        Ok(Self {
+            authority,
```
rtyler, line 54:
```suggestion
            authority: uri.to_string(),
```

rust/src/data_catalog/storage/mod.rs
roeap: my best defence is that this is an artifact from an earlier code version 😆.

wjones127 (approved):

PR #1383, opened 2023-05-21
feat: allow concurrent file compaction

### LINKED ISSUE #1171: parallel processing in Optimize command
# Description

Since moving the async reader when optimizing (binning) files, we can now leverage parallel processing to speed up the operation. Specifically the loops when executing the `MergePlan` are straight forward to parallelize.

**Use Case**

**Related Issue(s)**

wjones127:
# Description

Refactors such that:

1. Runs compaction tasks in parallel, with parallelism controlled by the user but defaulting to number of cpus. (The `num_cpu` crate is used by `tokio`, so we already have it transitively.)
2. Turns on zstd compression by default at level 4. In a future PR, we can make this configurable for Python and maybe benchmark different levels.
3. Initial prep to have other types of optimize commands.

However, the writer isn't very good at writing for a target row size, because the code that checks the size of the written file only knows the size of the serialized row groups and not the current row group. So if your row groups are 100MB in size, and you target 150MB, you will get 200MB files. There is upstream work in https://github.com/apache/arrow-rs/pull/4280 that will allow us to write much more exactly sized files, so this will improve in the near future.

# Related Issue(s)

closes #1171

# Documentation


rust/src/operations/optimize.rs
```diff
@@ -264,99 +282,130 @@ pub struct MergePlan {
...
+            .collect::<Result<Vec<_>, DeltaTableError>>()?;
+
+        let mut files_removed = MetricDetails::default();
+
+        for file in files.iter() {
+            files_removed.total_files += 1;
+            files_removed.total_size += file.size as i64;
+            files_removed.max = std::cmp::max(files_removed.max, file.size as i64);
+            files_removed.min = std::cmp::min(files_removed.min, file.size as i64);
+        }
```
roeap, line 321:
```suggestion
        let files_removed = files
            .iter()
            .fold(MetricDetails::default(), |mut curr, file| {
                curr.total_files += 1;
                curr.total_size += file.size as i64;
                curr.max = std::cmp::max(files_removed.max, file.size as i64);
                curr.min = std::cmp::min(files_removed.min, file.size as i64);
                curr
            });
```

wjones127:
FYI @Blajda

wjones127:
TBH, I'm not seeing much performance difference locally between 1 and 5 concurrent tasks. But I can at least confirm with the debug logs it is working concurrently. My current working theory is I am maxing out the write IO on my SSD.

roeap:
As you implied, I would assume that this could have the most benefit using object stores, presumably even more when deployed. I guess especially allowing out of order could have a benefit, when we are reading on multiple threads as well?

I guess we are about at that point where we maybe need to formalize the benchmarking setup a bit? Not sure if we reaches out to the databricks folks already, if they maybe have some "real" cloud infra that can be used in CI/CD?

Blajda:
I think we would still keep a form of bin packing and use this approach whenever a z-order is specified.

Based on my current understanding it looks like each worker is responsible for a single partition. Typically in my workflows partitions are dates and the latest dates are hot (i.e have a lot of small files that can be compacted) and optimization is executed daily after the date closes. Using bins increases the granularity of the work and keeps all the workers active.

Another trade off is that files near the target limit size are rewritten where bin packing would prune them when no coalescing is achieved.  

I imagine this is also not idempotent. Can you confirm if run optimize again on partition there are zero new actions?

wjones127:
> I imagine this is also not idempotent. Can you confirm if run optimize again on partition there are zero new actions?

It is. It passes the unit test you wrote earlier. :) Every time optimize is run you will end up with at most 1 file that is below the target size. If we only detect one file that is smaller than the target size, then we skip that partition.

> Another trade off is that files near the target limit size are rewritten where bin packing would prune them when no coalescing is achieved.

Yeah I think this is a convincing point. If you have two files of 100MB and a target of 110MB, then it will change them to one with 110MB and one with 90MB, which isn't really an improvement. So it's probably worth it to create planners specific to each optimize type.

I'll go ahead and bring back the binning.

rust/src/operations/optimize.rs
```diff
@@ -262,101 +333,153 @@ pub struct MergePlan {
...
+                        .await;
+                        match rewrite_result {
+                            Ok(Ok(data)) => Ok(data),
+                            Ok(Err(e)) => Err(e),
+                            Err(e) => Err(DeltaTableError::GenericError {
+                                source: Box::new(e),
+                            }),
+                        }
+                    })
+                    .buffer_unordered(self.max_concurrent_tasks)
```
wjones127, line 464: I'm finding this is a good pattern for parallel processing a stream:

1. Map tasks into `tokio::task::spawn()` (which allows them to be sent to other threads)
2. Then follow with `buffer_unordered` to control the number of tasks being spawned at a time.

wjones127:
So for 5 max tasks it's now about 3 times faster in the benchmark. I think the reason it's slower for a single thread is that I added compression by default.

Local benchmark results:

New results:

```
--------------------------------------------------------------------------------------- benchmark 'optimize': 2 tests ---------------------------------------------------------------------------------------
Name (time in ms)                     Min                   Max                  Mean              StdDev                Median                 IQR            Outliers     OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_optimize[5]       779.3044 (1.0)      1,052.4848 (1.0)        842.6561 (1.0)      117.7213 (1.0)        792.2957 (1.0)       84.6014 (1.0)           1;1  1.1867 (1.0)           5           1
test_benchmark_optimize[1]     2,559.4057 (3.28)     3,065.0166 (2.91)     2,704.3611 (3.21)     207.4415 (1.76)     2,644.0822 (3.34)     203.8526 (2.41)          1;0  0.3698 (0.31)          5           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Results from main:

```
---------------------------------------- benchmark 'optimize': 1 tests -----------------------------------------
Name (time in s)               Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------
test_benchmark_optimize     2.2828  2.6773  2.3809  0.1674  2.3108  0.1399       1;1  0.4200       5           1
----------------------------------------------------------------------------------------------------------------
```

rust/src/operations/optimize.rs
```diff
@@ -162,14 +218,18 @@ impl<'a> std::future::IntoFuture for OptimizeBuilder<'a> {
...
         Box::pin(async move {
-            let writer_properties = this
-                .writer_properties
-                .unwrap_or_else(|| WriterProperties::builder().build());
+            let writer_properties = this.writer_properties.unwrap_or_else(|| {
+                WriterProperties::builder()
+                    .set_compression(Compression::ZSTD(ZstdLevel::try_new(4).unwrap()))
+                    .set_created_by(format!("delta-rs version {}", crate_version()))
+                    .build()
+            });
```
Blajda, line 226: Since the files are rewritten using ZSTD don't we need to update the filenames?

https://github.com/delta-io/delta-rs/blob/8288fdd4e2b51982caeffb71e2456ac779e8b7b9/rust/src/operations/writer.rs#L303

rust/src/operations/optimize.rs
```diff
@@ -111,6 +149,10 @@ pub struct OptimizeBuilder<'a> {
     writer_properties: Option<WriterProperties>,
     /// Additional metadata to be added to commit
     app_metadata: Option<HashMap<String, serde_json::Value>>,
+    /// Whether to preserve insertion order within files (default false)
+    preserve_insertion_order: bool,
+    /// Max number of concurrent tasks (defeault 10)
```
Blajda, line 154: Update doc to mention num_cpus::get() is used to determine concurrency

rust/src/operations/optimize.rs
wjones127: Oh yeah. TBH I don't know why we write the compression algorithm in the filename. We don't handle this currently if a user passes in a different compression algorithm themselves.

roeap: > TBH I don't know why we write the compression algorithm in the filename

This is again one of these relics from the early days i think. I do remember comments floating around, about figuring out the various parts in names written by spark, so I guess short answer is "b/c spark did it" 😆.

That said I do believe this to be somewhat of a "soft convention" in many parquet usages, but I may also be wrong.

roeap (approved):
LGTM!

roeap:
re-ran the integration tests several times to no avail. Getting the HDFS integration to pass is often quite cumbersome, thus, I opened an issue regarding the failing integration tests with HDFS #1428.

rust/src/operations/optimize.rs
Blajda: Yeah this can be fixed in a future PR. I believe it should not break anything.

roeap (approved):

rust/src/operations/optimize.rs
houqp: yes, this the convention. the reason behind that is it can be used as an optimization to setup the parquet reader with proper decompression decoder based on the list api response alone, without having to read the first couple of bytes from the object store to detect the compression algo.

PR #1388, opened 2023-05-23
refactor!: remove `DeltaDataType` aliases

cmackenzie1:
# Description

Removing the aliases in favor of the actual type to improve readability of the code and reduce confusion around the existence of the types.

`DeltaDataTypeLong` replaced with `i64`
`DeltaDataTypeVersion` replaced with `i64`
`DeltaDataTypeTimestamp` replaced with `i64`
`DeltaDataTypeInt` replaced with `i32`

# Related Issue(s)
N/A

# Documentation

N/A

### DIFF
--- a/aws/delta-checkpoint/src/main.rs
+++ b/aws/delta-checkpoint/src/main.rs
@@ -10,7 +10,6 @@
 
 use deltalake::checkpoints;
 use deltalake::checkpoints::CheckpointError;
-use deltalake::DeltaDataTypeVersion;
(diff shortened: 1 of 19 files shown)


roeap (approved):
Thanks @cmackenzie1, I agree that this is an improvement, but I'll leave it open a bit for other opinions :) 

cc @wjones127 @rtyler @fvaleye

wjones127 (approved):
I'm good with this as well. I've always found them confusing. I still have to look up which bit width "long" is 🤣

PR #1390, opened 2023-05-24
feat: implement update operation

### LINKED ISSUE #1126: Support Update operation
# Description

Add update to the operations supported in https://github.com/delta-io/delta-rs/tree/main/rust/src/operations

See the `write` operation implementation for inspiration: https://github.com/delta-io/delta-rs/blob/main/rust/src/operations/write.rs

**Use Case**

**Related Issue(s)**

Blajda:
# Description
Users can now update data that matches a predicate. 

This operation should be encouraged over the replace write operation since update determines which values require rewriting based on the supplied predicate.

# Related Issue(s)
- closes #1126


Blajda:
This implementation follows the same structure as the delete command. `find_files` is used to determine which files contains records the satisfy the predicate. Once the files are identified 3 Datafusion projections are performed to first mark if the record needs to be updated, second add internal columns calculated based on if the column is to be updated, and finally renames the internal columns to their original names.

Something different from previous operations is that a new ExecutionPlan implementation is created to expose a count of how many records were updated. I wanted to avoid creating this count in a scan/loop and was able to take advantage of null counts.

rust/src/operations/update.rs
```diff
@@ -0,0 +1,921 @@
...
+
+use super::{transaction::commit, write::write_execution_plan};
+
+/// Used to represent user input of either a Datafusion expression or string expression
+pub enum Expression {
+    /// Datafusion Expression
+    E(Expr),
+    /// String Expression
+    S(String),
+}
```
wjones127, line 68: nit: please don't use single character variable names

```suggestion
pub enum Expression {
    /// Datafusion Expression
    DataFusion(Expr),
    /// String Expression
    String(String),
}
```

rust/src/operations/update.rs
```diff
@@ -0,0 +1,921 @@
...
+
+impl From<&str> for Expression {
+    fn from(val: &str) -> Self {
+        Expression::S(val.to_string())
+    }
+}
+
+/// Updates records in the Delta Table.
+/// See this module's documentaiton for more information
```
wjones127, line 83:
```suggestion
/// See this module's documentation for more information
```

rust/src/operations/update.rs
```diff
@@ -0,0 +1,921 @@
...
+            partition_values: Some(action.partition_values),
+            size: Some(action.size),
+            tags: None,
+        }))
+    }
+
+    metrics.execution_time_ms = Instant::now().duration_since(exec_start).as_micros();
+
+    // Do not make a commit when there are zero updates to the state
+    if !actions.is_empty() {
```
wjones127, line 413: Shouldn't we have eliminating this possibility when finding files?

rust/src/operations/update.rs
```diff
@@ -0,0 +1,921 @@
...
+            Some(Ok(batch)) => {
+                let array = batch.column_by_name("__delta_rs_update_predicate").unwrap();
+                let copied_rows = array.null_count();
+                let num_updated = array.len() - copied_rows;
+                let c1 = MetricBuilder::new(&self.metrics).global_counter("num_updated_rows");
+                c1.add(num_updated);
+
+                let c2 = MetricBuilder::new(&self.metrics).global_counter("num_copied_rows");
+                c2.add(copied_rows);
+                Some(Ok(batch))
```
wjones127, line 556: This is cool!

rust/src/operations/update.rs
```diff
@@ -0,0 +1,921 @@
...
+        expressions.push((
+            Arc::new(expressions::Column::new(field.name(), i)),
+            field.name().to_owned(),
+        ));
+    }
+
+    // Take advantage of how null counts are tracked in arrow arrays use the
+    // null count to track how many records do NOT statisfy the predicate.  The
+    // count is then exposed through the metrics through the `UpdateCountExec`
+    // execution plan
```
wjones127, line 288: Clever!

rust/src/operations/update.rs
```diff
@@ -0,0 +1,921 @@
...
+            "| 2     |",
+            "| 4     |",
+            "+-------+",
+        ];
+        let actual = get_data(&table).await;
+        assert_batches_sorted_eq!(&expected, &actual);
+    }
+
+    #[tokio::test]
+    async fn test_str_expressions() {}
```
wjones127, line 881: This a TODO?

wjones127 (changes requested):
This is impressive. I appreciate the variety of test cases :)

Had various questions and corrections throughout. I'll be excited to release this.

rust/src/delta_datafusion.rs
```diff
@@ -988,6 +1002,382 @@ impl TableProviderFactory for DeltaTableFactory {
...
+    let filter: Arc<dyn ExecutionPlan> =
+        Arc::new(FilterExec::try_new(predicate_expr, parquet_scan.clone())?);
+    let limit: Arc<dyn ExecutionPlan> = Arc::new(LocalLimitExec::new(filter, 1));
+
+    let task_ctx = Arc::new(TaskContext::from(state));
+    let partitions = limit.output_partitioning().partition_count();
+    let mut tasks = Vec::with_capacity(partitions);
+
+    for i in 0..partitions {
+        let stream = limit.execute(i, task_ctx.clone())?;
```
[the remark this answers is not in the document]
Blajda: I changed this area to push parallelism concerns onto Datafusion. I instead call collect and then process each batch for the path that was discovered. It cleans it up quite a bit

rust/src/operations/update.rs
Blajda: Yes good point.  I made a change to return early when find files determines zero candidates. Added a test to ensure the metrics are correct too.

rust/src/operations/update.rs
Blajda: Yeah my bad. I wanted to add a test to demonstrate that a str can be used for the predicate and update expression but it seemed like a test with little value.

I've updated the null tests to that.
FYI using strings for expressions will require additional work in future PRs.
If you have a predicate like `value < 2 or value > 2` DataFusion will return an error about being unable to compare `int32` to `int64`

roeap: > If you have a predicate like value < 2 or value > 2 DataFusion will return an error about being unable to compare int32 to int64

I stumbled across this one as well, not sure if this is maybe even a datafusion bug, since the expression parser seems to ignore the information in the schema passed to it.

rust/src/operations/update.rs
```diff
@@ -0,0 +1,936 @@
...
+
+    let mut _updates = HashMap::new();
+    for (key, expr) in updates {
+        let expr = match expr {
+            Expression::DataFusion(e) => e,
+            Expression::String(s) => snapshot.parse_predicate_expression(s)?,
+        };
+        _updates.insert(key, expr);
+    }
+    let updates = _updates;
```
roeap, line 225:
```suggestion
    let updates = updates
        .into_iter()
        .map(|(key, expr)| {
            let expr = match expr {
                Expression::DataFusion(e) => e,
                Expression::String(s) => snapshot.parse_predicate_expression(s)?,
            };
            Ok((key, expr))
        })
        .collect::<Result<HashMap<_, _>>, _>()?;
```

I think this might also work a be a little bit more idiomatic?

rust/src/operations/update.rs
roeap: Indeed it is!

roeap (commented):
Impressive work!

Left some minor comments, and there are some `unwraps` floating around that we maybe can have a look at if we can avoid them...

roeap (commented):
Impressive work!

Left some minor comments, and there are some `unwraps` floating around that we maybe can have a look at if we can avoid them...

rust/src/operations/update.rs
Blajda: Thanks it took me a bit to arrive to this solution but I'm glad ended up this simple.

rust/src/operations/update.rs
```diff
@@ -0,0 +1,936 @@
...
+    String(String),
+}
+
+impl From<Expr> for Expression {
+    fn from(val: Expr) -> Self {
+        Expression::DataFusion(val)
+    }
+}
+
+impl From<&str> for Expression {
```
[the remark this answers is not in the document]
Blajda: We can't do Into<String> since `Expr` implements `Display` and will cause a conflict with our `Expr` implementation. And yeah we can't do `AsRef<str>` since `Expr` may implement someday.

roeap (approved):
Looking good! - I'll leave it open for @wjones127 to look at, since he did the bulk of the review.

rust/src/delta_datafusion.rs
```diff
@@ -988,6 +999,363 @@ impl TableProviderFactory for DeltaTableFactory {
...
+
+    for batch in path_batches {
+        if batch.num_rows() > 1 {
+            return Err(DeltaTableError::Generic(
+                "Find files returned multiple records for batch".to_owned(),
+            ));
+        }
+        let array = batch
+            .column_by_name(PATH_COLUMN)
+            .unwrap()
```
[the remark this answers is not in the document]
Blajda: :disappointed: Sorry about that.
I factored that entire section into function since two places had the logic should be cleaner now.

rust/src/delta_datafusion.rs
```diff
@@ -988,6 +999,345 @@ impl TableProviderFactory for DeltaTableFactory {
...
+                statistics: snapshot.datafusion_table_statistics(),
+                projection: None,
+                limit: None,
+                table_partition_cols,
+                infinite_source: false,
+                output_ordering: None,
+            },
+            None,
+        )
+        .await?;
```
wjones127, line 1192: I found this Parquet scan was pulling all the columns, even ones that weren't involved in the predicate. These changes should make it only pull the ones needed for the predicate, plus the path column.

```suggestion
    // Identify which columns we need to project
    let mut used_columns = expression
        .to_columns()
        ?
        .into_iter()
        .map(|column| input_schema.index_of(&column.name))
        .collect::<Result<Vec<usize>, ArrowError>>()
        .unwrap();
    // Add path column
    used_columns.push(input_schema.fields().len());

    let parquet_scan = ParquetFormat::new()
        .create_physical_plan(
            state,
            FileScanConfig {
                object_store_url: store.object_store_url(),
                file_schema,
                file_groups: file_groups.into_values().collect(),
                statistics: snapshot.datafusion_table_statistics(),
                projection: Some(used_columns),
                limit: None,
                table_partition_cols,
                infinite_source: false,
                output_ordering: None,
            },
            None,
        )
        .await?;

    let input_schema = parquet_scan.schema();
    let input_dfschema: DFSchema = input_schema.clone().as_ref().clone().try_into()?;
    let predicate_expr = create_physical_expr(
        &Expr::IsTrue(Box::new(expression.clone())),
        &input_dfschema,
        &input_schema,
        state.execution_props(),
    )?;
```

rust/src/operations/update.rs
```diff
@@ -0,0 +1,947 @@
...
+            "| A  | 100   | 2023-05-14 |",
+            "| B  | 10    | 2021-02-02 |",
+            "+----+-------+------------+",
+        ];
+        let actual = get_data(&table).await;
+        assert_batches_sorted_eq!(&expected, &actual);
+    }
+
+    #[tokio::test]
+    async fn test_update_partitions() {
```
wjones127, line 749: Since a partition-only predicate is a separate code path from one with a mix of partition and normal columns, I think we are missing some coverage on partition column handling in the case where we need to scan in find files.

Could you add either a separate test or just another part in this test where you update a partitioned table with a predicate that is on the partition column *and* a normal column?

wjones127 (changes requested):
Almost there! I have a few performance-related suggestions, and one more test case I think we want. After that, I think this is good to go :)

rust/src/operations/update.rs
```diff
@@ -0,0 +1,947 @@
...
+    /// Number of files removed.
+    pub num_removed_files: usize,
+    /// Number of rows updated.
+    pub num_updated_rows: usize,
+    /// Number of rows just copied over in the process of updating files.
+    pub num_copied_rows: usize,
+    /// Time taken to execute the entire operation.
+    pub execution_time_ms: u128,
+    /// Time taken to scan the files for matches.
+    pub scan_time_ms: u128,
```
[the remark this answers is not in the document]
Blajda: So the expected unit is actually ms since I'm just pulling the metrics [here](https://learn.microsoft.com/en-us/azure/databricks/delta/history#--operation-metrics-keys) 
I've changed the call sites to use `as_millis()` which returns a u128. Should I really explicitly downcast that to a u64?

rust/src/delta_datafusion.rs
```diff
@@ -988,6 +999,345 @@ impl TableProviderFactory for DeltaTableFactory {
...
+                PATH_COLUMN
+            )))?;
+        for path in array {
+            let path = path.ok_or(DeltaTableError::Generic(format!(
+                "{} cannot be null",
+                PATH_COLUMN
+            )))?;
+
+            match actions.remove(path) {
+                Some(action) => files.push(action),
```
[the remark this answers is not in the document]
Blajda: Yes the  current implementation does guarantee that paths are unique but I want to make be defensive against any unexpected changes from Datafusion or future refactoring. My assumption is that removal from the hashmap is O(1) and that rust does not realloc the underlying array when active items go below a certain threshold.

rust/src/operations/update.rs
Blajda: Done

rust/src/operations/update.rs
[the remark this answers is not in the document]
wjones127: Yes that will always be safe on the timescales we care about.

wjones127 (approved):

PR #1395, opened 2023-05-27
feat: dynamodb lock reorganization

DnOberon:
# Description
This is the first PR for the work on providing a second locking provider such as PostgreSQL.  I wanted to get this one in early as it changes function signatures and touches the delta lake package. 

- renamed dynamodb_lock to distributed_lock
- pulled dynamodb specific code out into its own module
- pulled tests out into their own folders
- made the `LockItem` and `LockClient` part of the root `distributed_lock` module in preparation for changing the `s3` module


roeap:
Thanks for doing this @DnOberon!

Just wondering out loud here. The current implementation of the S3 is a bit of a "zombie" as it is somewhere between our current way of configuring things and the way we used to do things before the `object_store` crate. In essence there is also nothing S3 specific about it, It already just takes a `DynObjectStore` as inner. So we can also use it for other `ObjectStore` implementations. 

Would it therefore maybe make sense to just rip out the whole thing, rename it to `LockedObjectStore`, so the delta-rs crate needs no knowledge about the specific of how locking is handled?

Maybe we can eventually even go as far as moving the locked store upstream to the object store crate where it can take advantage of the various credentials.

Generally though I am a bit hesitant to rename the crate just yet, since we need to publish the crate for others to build delta-rs with s3 support enabled. So maybe we should just get it to work end to end under the current - albeit confusing - name, and then rename and publish it in one go?

DnOberon:
I’m fine with that as long as you’re fine with the other parts of the
reorganization so far.

roeap:
Did not mean to scare you off - hope it was not perceived as such 😄.

In the end its not just my opinion that matters, if you have thoughts around this happy to hear! Also do @rtyler @wjones127 @fvaleye have opinions?

wjones127:
I think what Robert said makes sense to me. I haven’t used the locking clients much so don’t have too strong opinions.

DnOberon:
It wasn’t :) You made a valid point and I was closing the pr so it wouldn’t
clutter anything up if you’re going to keep the same name for right now.

PR #1396, opened 2023-05-27
perf: improve record batch partitioning

roeap:
# Description

Some time ago, the arrow-row crate was published, specifically with use cases like multi-column ordering / sorting in mind. We need to do this whenever we write data to a table to partition batches according to their partition values. In this PR we adopt this. As a drive-by I also updated the imports in the module to import from the respective sub-crates.

# Related Issue(s)

# Documentation

### DIFF
--- a/rust/src/writer/record_batch.rs
+++ b/rust/src/writer/record_batch.rs
@@ -26,31 +26,30 @@
 //!     }))
 //! }
 //! ```
-use super::{
-    stats::{create_add, NullCounts},
-    utils::{
-        arrow_schema_without_partitions, next_data_path, record_batch_without_partitions,
-        stringified_partition_value, PartitionPath,
-    },
-    DeltaWriter, DeltaWriterError,
-};
-use crate::builder::DeltaTableBuilder;
-use crate::writer::stats::apply_null_counts;
-use crate::writer::utils::ShareableBuffer;
-use crate::DeltaTableError;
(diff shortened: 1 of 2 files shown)


rust/src/writer/record_batch.rs
```diff
@@ -410,6 +403,18 @@ pub(crate) fn divide_by_partition_values(
     Ok(partitions)
 }
 
+fn lexsort_to_indices(arrays: &[ArrayRef]) -> UInt32Array {
```
rtyler, line 406: I am not following why this function is being used here instead of the previous import from `arrow::compute` as far as I can tell the only difference is this version doesn't take the optional `limit` parameter?

:confused:

roeap: This version uses arrow-row rather then the "regular" compute kernels. https://github.com/apache/arrow-rs/blob/77aa8f5b2645a91724048f5c1d644c6b52880028/arrow-ord/src/sort.rs#L1081-L1082

wjones127 (approved):
LGTM!

rust/src/writer/record_batch.rs
rtyler, line 406:
```suggestion
/*
 * This version uses arrow-row rather then the "regular" compute kernels. 
 */
fn lexsort_to_indices(arrays: &[ArrayRef]) -> UInt32Array {
```

rtyler (approved):

PR #1397, opened 2023-05-27
feat: more efficient parquet writer and more statistics

### LINKED ISSUE #1208: Ensure decimal statistics are written correctly in Rust
We are currently writing the underlying i128 value, but we should be writing the equivalent floating point value.

Here is an example from Spark:

https://github.com/delta-io/delta-rs/blob/860e272c9484fcd9841e13d33d48d119319ffcd5/rust/tests/data/delta-1.2.1-only-struct-stats/_delta_log/00000000000000000002.json#L2

### LINKED ISSUE #1209: Support writing statistics for date columns in Rust
# Description

**Use Case**

**Related Issue(s)**

### LINKED ISSUE #1394: Rust writer in operations makes a lot of data copies
# Environment

We keep the Parquet file being written as a `Vec<u8>`. The `write_batch` method seems to clone the vec each time it is called, and the batch size for many operations is quite small, defaulting at 1024 rows. 

https://github.com/delta-io/delta-rs/blob/930d16e583fbd784f5f3d4bb8bcd753e2d6488a6/rust/src/operations/writer.rs#L326-L343

@roeap Do you remember why we do this? I'm wondering whether we can just rip out this error handling, or if we actually do some sort of retry with this.

wjones127:
# Description

* Removed the data copies in a tight loop, which were extremely bad for performance when writing files > 100MB.
* Rewrote statistics handling to collect null values from metadata, just like min and max.
* Added support for more types in statistics.

# Related Issue(s)

- closes #1394
- closes #1209 
- closes #1208

# Documentation

### DIFF
--- a/rust/src/operations/writer.rs
+++ b/rust/src/operations/writer.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use crate::action::Add;
 use crate::storage::ObjectStoreRef;
 use crate::writer::record_batch::{divide_by_partition_values, PartitionResult};
-use crate::writer::stats::{apply_null_counts, create_add, NullCounts};
+use crate::writer::stats::create_add;
 use crate::writer::utils::{
     arrow_schema_without_partitions, record_batch_without_partitions, PartitionPath,
     ShareableBuffer,
@@ -16,7 +16,6 @@ use arrow::datatypes::SchemaRef as ArrowSchemaRef;
 use arrow::error::ArrowError;
 use arrow::record_batch::RecordBatch;
 use bytes::Bytes;
-use log::warn;
 use object_store::{path::Path, ObjectStore};
 use parquet::arrow::ArrowWriter;
 use parquet::basic::Compression;
@@ -269,7 +268,6 @@ pub(crate) struct PartitionWriter {
     buffer: ShareableBuffer,
     arrow_writer: ArrowWriter<ShareableBuffer>,
     part_counter: usize,
-    null_counts: NullCounts,
     files_written: Vec<Add>,
 }
 
@@ -293,7 +291,6 @@ impl PartitionWriter {
             buffer,
             arrow_writer,
             part_counter: 0,
-            null_counts: NullCounts::new(),
             files_written: Vec::new(),
         })
     }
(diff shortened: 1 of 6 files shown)


rust/src/operations/writer.rs
```diff
@@ -324,40 +318,27 @@ impl PartitionWriter {
     }
 
     fn write_batch(&mut self, batch: &RecordBatch) -> DeltaResult<()> {
-        // copy current cursor bytes so we can recover from failures
-        // TODO is copying this something we should be doing?
-        let buffer_bytes = self.buffer.to_vec();
```
wjones127, line 329: This was the offending line causing a lot of data copying.

rust/src/writer/stats.rs
```diff
@@ -528,7 +698,7 @@ mod tests {
                 ("some_bool", ColumnCountStat::Value(v)) => assert_eq!(100, *v),
                 ("some_string", ColumnCountStat::Value(v)) => assert_eq!(100, *v),
                 ("some_list", ColumnCountStat::Value(v)) => assert_eq!(100, *v),
-                ("some_nested_list", ColumnCountStat::Value(v)) => assert_eq!(0, *v),
+                ("some_nested_list", ColumnCountStat::Value(v)) => assert_eq!(100, *v),
```
wjones127, line 701: I believe this test was incorrect, given the null values here:

https://github.com/delta-io/delta-rs/blob/0ee692cf74bed6ce259a7e883c34c3f0fe407569/rust/src/writer/stats.rs#L885

rtyler (approved):

rust/src/writer/stats.rs
```diff
@@ -448,6 +615,8 @@ mod tests {
         let mut null_count_keys = vec!["some_list", "some_nested_list"];
         null_count_keys.extend_from_slice(min_max_keys.as_slice());
 
+        dbg!(&stats);
```
roeap, line 618: debug artifact?

rust/src/writer/stats.rs
```diff
@@ -148,45 +67,338 @@ fn min_max_values_from_file_metadata(
...
+                min: None,
+                max: None,
+                null_count,
+            }
+        }
+    }
+}
+
+impl AddAssign for AggregatedStats {
+    fn add_assign(&mut self, rhs: Self) {
```
roeap, line 306: 👍

roeap (dismissed):
Really looking forward to see this in action! Feels like this a great step forward for our write experience!

roeap (dismissed):

roeap (approved):

PR #1407, opened 2023-05-29
chore: type-check friendlier exports

roeap:
# Description

Some static type checker (i.e. pyright) consider imports as internal to the module unless explicitly stated otherwise - by adding it to `__all__` or using the "redundant alias" (not sure that's an official term 😆) form. I opted for the second, as I think it looks a bit nicer..

We were also exporting `RawDeltaTable` at the root, which I thogtht would be an internal class, that we do not want to expose to users?

Took the opportunity to also add a few more docstrings.

# Related Issue(s)

# Documentation

### DIFF
--- a/python/deltalake/_internal.pyi
+++ b/python/deltalake/_internal.pyi
@@ -95,18 +95,34 @@ class Field:
         *,
         nullable: bool = True,
         metadata: Optional[Dict[str, Any]] = None,
-    ) -> None: ...
+    ) -> None:
+        """A named field, with a data type, nullability, and optional metadata."""
     name: str
+    """The field name."""
     type: DataType
+    """The field data type."""
     nullable: bool
+    """The field nullability."""
     metadata: Dict[str, Any]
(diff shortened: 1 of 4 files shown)


python/deltalake/_internal.pyi
```diff
@@ -95,18 +95,34 @@ class Field:
...
+    ) -> None:
+        """A named field, with a data type, nullability, and optional metadata."""
     name: str
+    """The field name."""
     type: DataType
+    """The field data type."""
     nullable: bool
+    """The field nullability."""
     metadata: Dict[str, Any]
+    """The field metadata."""
```
wjones127, line 107: Does this show up in the API reference correctly? I had never seen this syntax before.

wjones127 (approved):

python/deltalake/_internal.pyi
roeap: Checked again, and it seems to be valid.

PR #1409, opened 2023-05-29
feat: explicit python exceptions

roeap:
# Description

Working on an integration, I felt that our error handling in python was a bit opaque and handling errors on the python side would have to rely on inspecting the error message, which can easily silently break, if not covered by tests.

The proposed approach would be to introduce a few more python exceptions and - as we already do - rely on existing python exceptions where appropriate.

As this would be a bigger breaking change on the python side I guess, looking for some feedback, if you think this is worthwhile.

@wjones127 @fvaleye @rtyler 

# Related Issue(s)

# Documentation


wjones127:
My overall opinion on Python errors is that we should be mapping to standard Python exceptions in most cases, except:

* `TableNotFoundError`: for when we can't find the Delta Table at a path
* `DeltaProtocolError`: for when an operation was tried that isn't allowed by the delta protocol (e.g. trying to write to a delta table with a writer version we don't yet support)
* `CommitFailedError`: When a commit write fails.

But most other things could be Python exceptions:

* Tokio errors: `RuntimeError`
* Chrono parse error: `ValueError`
* Object store not found: `FileNotFoundError`
* `std::io::Error`: `IOError`

The purpose of these errors should be making it as easy as possible for Python users to write error handling code. If I were making breaking changes to the Python exceptions, I would drop the generic `PyDeltaTableError` and only make custom errors for the specific cases that aren't already covered by built-in exceptions.

roeap:
> My overall opinion on Python errors is that we should be mapping to standard Python exceptions in most cases

Completely agree! Previously had a look into the datafusion python repo, and they already did a lot of work mapping to python exceptions. We should be able to adopt much of that since our stacks overlap quite a bit ...

I also like the custom error you proposed, along with being generally cautious of adding additional custom ones.

roeap:
@wjones127 - converted all the errors and did not find any that would warrant a new custom error type other then the ones you mentioned. At times I may have been too liberal with the `DeltaProtocolError` as I interpreted it as thing specified in the protocol. I "think" the mention to atomicity is alos in that document so e.g. the unsafe rename error is also a protocol error.

Main think I took away from this though is that we have to a) name / use errors more consistently throughout delta-rs, and b) simplify our errors quite a bit, but this goes beyond the scope of this PR ...

python/deltalake/writer.py
```diff
@@ -403,16 +404,18 @@ def try_get_deltatable(
...
+    # except PyDeltaTableError as err:
+    #     # TODO: There has got to be a better way...
+    #     if "Not a Delta table" in str(err):
+    #         return None
+    #     elif "cannot find" in str(err):
+    #         return None
+    #     elif "No such file or directory" in str(err):
+    #         return None
+    #     else:
+    #         raise
```
rtyler, line 418:
```suggestion
```

I don't see a reason to keep this commented code around

python/src/error.rs
```diff
@@ -0,0 +1,96 @@
...
+        // commit erros
+        DeltaTableError::Transaction { source } => CommitFailedError::new_err(source.to_string()),
+
+        // python exceptions
+        DeltaTableError::ObjectStore { source } => object_store_to_py(source),
+        DeltaTableError::Io { source } => PyIOError::new_err(source.to_string()),
+
+        DeltaTableError::Arrow { source } => arrow_to_py(source),
+
+        // catach all
```
rtyler, line 35:
```suggestion
        // catch all
```

python/src/error.rs
```diff
@@ -0,0 +1,96 @@
...
+        DeltaTableError::InvalidTableLocation(msg) => TableNotFoundError::new_err(msg),
+
+        // protocol errors
+        DeltaTableError::InvalidJsonLog { .. } => DeltaProtocolError::new_err(err.to_string()),
+        DeltaTableError::InvalidStatsJson { .. } => DeltaProtocolError::new_err(err.to_string()),
+        DeltaTableError::InvalidData { violations } => {
+            DeltaProtocolError::new_err(format!("Inaviant violations: {:?}", violations))
+        }
+
+        // commit erros
```
rtyler, line 26:
```suggestion
        // commit errors
```

rtyler (commented):
I like this cleanup, especially the ergonomics of mapping errors to `PythonError::from`.

Some minor suggestions, but I think once @wjones127 approves this should be landed

python/src/error.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+fn object_store_to_py(err: ObjectStoreError) -> PyErr {
+    match err {
+        ObjectStoreError::NotFound { .. } => PyFileNotFoundError::new_err(err.to_string()),
+        ObjectStoreError::Generic { source, .. }
+            if source.to_string().contains("AWS_S3_ALLOW_UNSAFE_RENAME") =>
+        {
+            DeltaProtocolError::new_err(source.to_string())
+        }
```
wjones127, line 47: 👍

python/src/filesystem.rs
```diff
@@ -61,7 +61,7 @@ impl DeltaFileSystemHandler {
         let to_path = Path::from(dest);
         self.rt
             .block_on(self.inner.copy(&from_path, &to_path))
-            .map_err(PyDeltaTableError::from_object_store)?;
+            .map_err(PythonError::from)?;
```
wjones127, line 64: 🤔 I wonder if we implemented:

```rust
impl<T: Into<PythonError>> From<T> for PyErr { ... }
```

could we remove the `map_err` entirely?

wjones127 (commented):
This is awesome! Besides existing comments had one thought about whether we could remove some more code.

python/src/filesystem.rs
roeap: Cool idea, and gave it a quick go, but the compiler is complaining that `Into` is not our trait, and neither is `PyErr`, so it won't allow it :(

rtyler (approved):

PR #1423, opened 2023-06-02
chore: remove the delta-checkpoint lambda which I have moved to a new repo

rtyler:
This can now be found in https://github.com/delta-incubator/lambdas where it can be iterated upon and released as a downstream project of deltalake crate releases


houqp:
thoughts on mentioning the downstream project in the readme for awareness?

rtyler:
@houqp Not particularly, we had not mentioned it previously :shrug: Based on real-world experience the checkpoint lambda in its current form suffered from some memory problems in production anyways that need to be ironed out :laughing:

houqp (approved):

PR #1424, opened 2023-06-02
chore: remove no-longer-necessary build.rs for Rust bindings

rtyler:
The build.rs is a long-since-unnecessary customization since we adopted the object_store crate and no longer manage storage ourselves. This change also allows us to purge the glibc-version build dependency which makes the build a teensy bit faster too.


roeap (approved):
👍

PR #1430, opened 2023-06-03
refactor: move checkpoint and errors into separate module

roeap:
# Description

While doing #1409 it became evident, that our errors are somewhat organically grown and could do with some pruning. At the same time we are hoping to reorganize (#1136) delta-rs a bit, to make it easier to reason about.

This PR is a first attempt to introduce a more explicit approach to how we model our errors. Here I'd like to propose we work towards the approach taken in the object store crate - specifically have very specific errors in submodules, but do not surface those errors to the user. Rather collapse them into eventually a single error type.

Since delta-rs does much more things then object store, I believe we will eventually end up with some more top level errors, or have maybe a two level hierarch.

As a first step in this PR:
- move `checkpoints.rs` into actions module (as proposed in #1136)
- move top level errors into new `errors` module - at lest the ones defined in `delta.rs`
- try and consolidate `ActionError` and `CheckpointError` which are now part of the `action` module.

As I needed to touch a bunch of imports anyhow, I took the liberty to organize them according to what to the best of my knowledge is the [leading contender](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#StdExternalCrate%5C%3A) for what rust-fmt might do.

@wjones127 @rtyler @Blajda - opening this as a draft to get some feedback, as this is a somewhat intrusive change to the overall crate, in case you want to go another way. 

# Related Issue(s)

part of: #1136

# Documentation

### DIFF
--- a/python/src/error.rs
+++ b/python/src/error.rs
@@ -1,6 +1,6 @@
 use arrow_schema::ArrowError;
 use deltalake::checkpoints::CheckpointError;
-use deltalake::{DeltaTableError, ObjectStoreError};
+use deltalake::{errors::DeltaTableError, ObjectStoreError};
 use pyo3::exceptions::{
     PyException, PyFileNotFoundError, PyIOError, PyNotImplementedError, PyValueError,
 };
--- a/python/src/error.rs
+++ b/python/src/error.rs
@@ -1,5 +1,5 @@
 use arrow_schema::ArrowError;
-use deltalake::checkpoints::CheckpointError;
+use deltalake::action::ProtocolError;
 use deltalake::{errors::DeltaTableError, ObjectStoreError};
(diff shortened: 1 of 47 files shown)


roeap:
The refactoring went reasonably well, at the very least `CheckpointError` is contained in its module. I went ahead and renamed `ActionError` to `ProtocolError`, since I thought its reasonable that we should end up with an error type as such.

If folks agree I'd like to get this merged to do some follow ups, that should be much easier to reason about.

wjones127 (approved):
This all looks reasonable.

Is there a script yet we can run to order the use statements? Or are we doing that manually for now?

roeap:
I think i tried it once, but it required unstable rust. In principle though its part of rustfmt.

PR #1431, opened 2023-06-03
feat: update writers to include compression method in file name

Blajda:
# Description
The compression name was hard-coded to include snappy however users can now  specify their own methods which will cause a disconnect in the the name and the method used.

The naming convention is used as a hint by spark and hive to create the correct reader without reading having read the header of the file.

### DIFF
--- a/rust/src/writer/utils.rs
+++ b/rust/src/writer/utils.rs
@@ -90,14 +90,16 @@ pub(crate) fn next_data_path(
 ) -> Path {
     fn compression_to_str(compression: &Compression) -> &str {
         match compression {
+            // This is to match HADOOP's convention
+            // https://github.com/apache/parquet-mr/blob/c4977579ab3b149ea045a177b039f055b5408e8f/parquet-common/src/main/java/org/apache/parquet/hadoop/metadata/CompressionCodecName.java#L27-L34
             Compression::UNCOMPRESSED => "",
             Compression::SNAPPY => ".snappy",
             Compression::GZIP(_) => ".gz",
             Compression::LZO => ".lzo",
-            Compression::BROTLI(_) => ".brotli",
+            Compression::BROTLI(_) => ".br",
             Compression::LZ4 => ".lz4",
             Compression::ZSTD(_) => ".zstd",
-            Compression::LZ4_RAW => ".lz4_raw",
+            Compression::LZ4_RAW => ".lz4raw",
         }
     }
 
--- a/rust/src/writer/utils.rs
+++ b/rust/src/writer/utils.rs
@@ -371,7 +371,6 @@ mod tests {
         let uuid = Uuid::parse_str("02f09a3f-1624-3b1d-8409-44eff7708208").unwrap();
 
         // Validated against Spark
-
         let props = WriterProperties::builder()
             .set_compression(Compression::UNCOMPRESSED)
             .build();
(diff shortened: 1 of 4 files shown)


wjones127:
Thanks for fixing this. 😄 

> The naming convention is used as a hint by spark and hive to create the correct reader without reading having read the header of the file.

I'd like to know more about how they use this. Each column is allowed to have a different compression in Parquet, so the Java readers need to be ready to use any compression decoder anyways. If there's not much benefit, it seems like it would be better to just write out a plain `.parquet` file, right? Is there any code or docs we can point to where Spark and Hive use this?

Blajda:
@houqp since you provided the original insight in the other PR can you help direct us to the source of that claim?

rust/src/writer/utils.rs
```diff
@@ -75,33 +78,44 @@ impl Display for PartitionPath {
...
+    fn compression_to_str(compression: &Compression) -> &str {
+        match compression {
+            Compression::UNCOMPRESSED => "",
+            Compression::SNAPPY => ".snappy",
+            Compression::GZIP(_) => ".gz",
+            Compression::LZO => ".lzo",
+            Compression::BROTLI(_) => ".brotli",
+            Compression::LZ4 => ".lz4",
+            Compression::ZSTD(_) => ".zstd",
+            Compression::LZ4_RAW => ".lz4_raw",
```
wjones127, line 100:
```suggestion
            // This is to match HADOOP's convention
            // https://github.com/apache/parquet-mr/blob/c4977579ab3b149ea045a177b039f055b5408e8f/parquet-common/src/main/java/org/apache/parquet/hadoop/metadata/CompressionCodecName.java#L27-L34
            Compression::UNCOMPRESSED => "",
            Compression::SNAPPY => ".snappy",
            Compression::GZIP(_) => ".gz",
            Compression::LZO => ".lzo",
            Compression::BROTLI(_) => ".br",
            Compression::LZ4 => ".lz4",
            Compression::ZSTD(_) => ".zstd",
            Compression::LZ4_RAW => ".lz4raw",
```

I *think* this is the right convention, but if you are referencing another source, would be good to know.

rust/src/writer/utils.rs
```diff
@@ -75,33 +78,44 @@ impl Display for PartitionPath {
...
 
-    let partition_key = PartitionPath::from_hashmap(partition_columns, partition_values)?;
-    Ok(Path::from(format!("{partition_key}/{file_name}")))
+    // TODO: what does c000 mean?
+    let file_name = format!(
+        "part-{}-{}-c000{}.parquet",
+        part,
+        writer_id,
+        compression_to_str(&compression)
+    );
```
wjones127, line 117: We seem to be trying to copy Spark. This information is mostly added to make debugging easier. For example, they have a writer id because they might want to see all the files written by a particular node in a Spark cluster. But we don't have nodes or retries like Spark does. So I think we can have our own convention.

At a minimum, I think all we need is `{uuid}.parquet`. This is just to make sure we don't have collisions between files. Adding the compression codec can be nice for debugging. If we find more things that's are useful to delta-rs, we can consider adding them too.

Blajda: Thanks for locating the Spark source. I tried searching myself but only found the mapping for orc files. 
I left some of the comments / TODO that were scattered in the implementation just in case if we wanted to completely copy the Spark approach. Since all writers now use this function it should be easier in the future to make our own convention as you mentioned.

wjones127 (approved):

PR #1432, opened 2023-06-04
refactor: remove `LoadCheckpointError` and `ApplyLogError`

roeap:
# Description

This PR removes `LoadCheckpointError` and  `ApplyLogError` as well as moves some more checkpoint related functions into the checkpoint module.

Aside from cleaning up errors, this moves us a little bit closer to isolating log handling related logic in preparation for some upcoming work.

# Related Issue(s)

part of: #1136

### DIFF
--- a/python/src/error.rs
+++ b/python/src/error.rs
@@ -64,6 +64,7 @@ fn checkpoint_to_py(err: ProtocolError) -> PyErr {
         ProtocolError::Arrow { source } => arrow_to_py(source),
         ProtocolError::ObjectStore { source } => object_store_to_py(source),
         ProtocolError::NoMetaData => DeltaProtocolError::new_err("Table metadata missing"),
+        ProtocolError::CheckpointNotFound => DeltaProtocolError::new_err(err.to_string()),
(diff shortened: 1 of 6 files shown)


python/src/error.rs
```diff
@@ -63,11 +63,14 @@ fn checkpoint_to_py(err: ProtocolError) -> PyErr {
     match err {
         ProtocolError::Arrow { source } => arrow_to_py(source),
         ProtocolError::ObjectStore { source } => object_store_to_py(source),
+        ProtocolError::EndOfLog => DeltaProtocolError::new_err("Enf of log"),
```
cmackenzie1, line 66:
```suggestion
        ProtocolError::EndOfLog => DeltaProtocolError::new_err("End of log"),
```

cmackenzie1:
> and everyting in delta-rs just returns a DeltaError ...

@roeap I think that sounds great. Right now there are a few things that return different errors so you can't use `?` if they are in the same scope.

Is there the possibility to create a class of errors `Retryable`? Things for transient errors like 5xx from object stores?

wjones127 (approved):
All seems reasonable :)

PR #1434, opened 2023-06-04
refactor: re-organize top level modules

roeap:
# Description

~This contains changes from #1432, will rebase once that's merged.~

This PR constitutes the bulk of re-organising our top level modules.
- move `DeltaTable*` structs into new `table` module
- move table configuration into `table` module
- move schema related modules into `schema` module
- rename `action` module to `protocol` - hoping to isolate everything that can one day be the log kernel.

~It also removes the deprecated commit logic from `DeltaTable`  and updates call sites and tests accordingly.~

I am planning one more follow up, where I hope to make `transactions` currently within `operations` a top level module. While the number of touched files here is already massive, I want to do this in a follow up, as it will also include some updates to the transactions itself, that should be more carefully reviewed.

# Related Issue(s)

closes: #1136

# Documentation


python/src/error.rs
```diff
@@ -63,11 +63,14 @@ fn checkpoint_to_py(err: ProtocolError) -> PyErr {
     match err {
         ProtocolError::Arrow { source } => arrow_to_py(source),
         ProtocolError::ObjectStore { source } => object_store_to_py(source),
+        ProtocolError::EndOfLog => DeltaProtocolError::new_err("Enf of log"),
```
cmackenzie1, line 66:
```suggestion
        ProtocolError::EndOfLog => DeltaProtocolError::new_err("End of log"),
```

rust/src/operations/transaction/mod.rs
```diff
@@ -164,7 +164,8 @@ pub async fn commit(
 ) -> DeltaResult<i64> {
     let tmp_commit = prepare_commit(storage, &operation, actions, app_metadata).await?;
 
-    let max_attempts = 5;
+    // TODO make max attempts configurable
+    let max_attempts = 25;
```
roeap, line 168: Increasing this was needed to make our concurrent writer tests pass.

rust/tests/integration_commit.rs
roeap, line 1: these tests we just covering the old commit logic.

rust/src/operations/transaction/mod.rs
wjones127: Huh. I guess 5 workers trying to make 3 commits. That mean one potentially has to fail 14 times?

wjones127 (approved):
This looks good. I like that there aren't as many top level modules in the end. We should do a round of Rust documentation improvements after this.

I'd like @rtyler's feedback before we merge though.

rust/src/operations/transaction/mod.rs
roeap: 10 was not cutting it for s3, so at least that...

roeap:
> We should do a round of Rust documentation improvements after this

Was thinking the same. At the very lest we should be able to say what each module does ...

roeap:
@rtyler - I'd love to get this merged, as I have some work lined up dependent on this. Before we'd love to hear your opinion.

The number of changed files may look daunting, but its mostly renames 😉.

roeap:
@rtyler @wjones127 - any chance we can get this merged?

IF you feel this should be abandoned, let me know as well, then I'll stop using this as a baseline for follow on work 😆.

ion-elgreco:
> I I would really like you to do some refactoring. I'm currently working on implementing CDF, but I feel that if the source code is more structured with clearer guidelines, it will be easier to determine where to implement it.

How is adding CDF going? :D Would be awesome to have

wjones127 (approved):

PR #1461, opened 2023-06-14
feat: handle larger z-order jobs with streaming output and spilling

### LINKED ISSUE #1459: Error when Z Ordering a larger dataset
# Environment

**Delta-rs version**: 0.10.0

**Binding**: Python

**Environment**:
- **Cloud provider**: Localhost
- **OS**: Macbook 
- **Other**: Macbook M1 with 64 GB of RAM

***
# Bug

**What happened**: Z Order command worked on 5 GB h2o groupby dataset (1e8), but errors out of 50 GB dataset (1e9)

**What you expected to happen**: I expected the Z Ordering to work

**How to reproduce it**: [This notebook](https://github.com/delta-io/delta-examples/blob/master/notebooks/python-deltalake/z-order-delta-table.ipynb) shows the computations working well on the 1e8 dataset, but erroring out on the 1e9 dataset.

**More details**: I'm Z Ordering on a single column.  Here's the error message:

```
thread 'tokio-runtime-worker' panicked at 'overflow', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/arrow-select-39.0.0/src/interleave.rs:172:56
DeltaError                                Traceback (most recent call last)
File <timed eval>:1

File ~/opt/miniconda3/envs/deltalake-0100/lib/python3.9/site-packages/deltalake/table.py:697, in TableOptimizer.z_order(self, columns, partition_filters, target_size, max_concurrent_tasks)
    675 def z_order(
    676     self,
    677     columns: Iterable[str],
   (...)
    680     max_concurrent_tasks: Optional[int] = None,
    681 ) -> Dict[str, Any]:
    682     """
    683     Reorders the data using a Z-order curve to improve data skipping.
    684 
   (...)
    695     :return: the metrics from optimize
    696     """
--> 697     metrics = self.table._table.z_order_optimize(
    698         list(columns), partition_filters, target_size, max_concurrent_tasks
    699     )
    700     self.table.update_incremental()
    701     return json.loads(metrics)

DeltaError: Generic error: task 334 panicked
```

### LINKED ISSUE #1460: Support out-of-core Z-order using DataFusion

wjones127:
# Description

Fixes the base implementation so that is doesn't materialize the entire result in one record batch. It will still require materializing the full input for each partition in memory. This is mostly a problem for unpartitioned table, since that means materializing the entire table in memory.

Adds a new datafusion-based implementation enabled by the `datafusion` feature. In theory, this should support spilling to disk.

# Related Issue(s)

For example:

- closes #1459
- closes #1460

# Documentation

### DIFF
--- a/rust/src/operations/optimize.rs
+++ b/rust/src/operations/optimize.rs
@@ -447,7 +447,7 @@ impl MergePlan {
         columns: Arc<Vec<String>>,
         files: MergeBin,
         object_store: ObjectStoreRef,
-        _use_threads: bool,
+        use_threads: bool,
     ) -> Result<BoxStream<'static, Result<RecordBatch, ParquetError>>, DeltaTableError> {
         use arrow_array::cast::as_generic_binary_array;
         use arrow_array::ArrayRef;
@@ -502,9 +502,12 @@ impl MergePlan {
             .collect_vec();
 
         // Interleave the batches
-        Ok(util::interleave_batches(batches, indices, 10_000)
-            .map_err(|err| ParquetError::General(format!("Failed to reorder data: {:?}", err)))
-            .boxed())
+        Ok(
+            util::interleave_batches(batches, indices, 10_000, use_threads)
+                .await
+                .map_err(|err| ParquetError::General(format!("Failed to reorder data: {:?}", err)))
+                .boxed(),
+        )
     }
 
     /// Datafusion-based z-order read.
@@ -939,13 +942,15 @@ pub(super) mod util {
 
     /// Interleaves a vector of record batches based on a set of indices
     #[cfg(not(feature = "datafusion"))]
-    pub fn interleave_batches(
+    pub async fn interleave_batches(
         batches: Vec<RecordBatch>,
         indices: Vec<(usize, usize)>,
         batch_size: usize,
+        use_threads: bool,
     ) -> BoxStream<'static, Result<RecordBatch, DeltaTableError>> {
-        use arrow_schema::ArrowError;
+        use arrow_array::ArrayRef;
         use arrow_select::interleave::interleave;
+        use futures::TryFutureExt;
 
         if batches.is_empty() {
             return futures::stream::empty().boxed();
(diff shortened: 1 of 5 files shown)


rust/src/operations/optimize.rs
```diff
@@ -498,11 +502,77 @@ impl MergePlan {
...
+        use datafusion::execution::runtime_env::{RuntimeConfig, RuntimeEnv};
+        // TODO: usethreads
+        use datafusion::prelude::{col, ParquetReadOptions, SessionConfig};
+
+        // TODO: make this configurable.
+        // TODO: push this up and share between bins or maybe an overall runtime.
+        let memory_pool = FairSpillPool::new(8 * 1024 * 1024 * 1024);
+        let config = RuntimeConfig::new().with_memory_pool(Arc::new(memory_pool));
+        let runtime = Arc::new(RuntimeEnv::new(config)?);
+        runtime.register_object_store(&Url::parse("delta-rs://").unwrap(), object_store);
```
wjones127, line 531: I'm a little stumped here. Where should this configuration live?

It makes sense to have the memory pool as a global config, IMO. But the object store should be local to here, since it's specific to this particular table's root.

roeap: Had some similar question a little while back. For a while I thought the `DeltaOps` struct could be a place where session state is stored, but the way we use that right now, it's even shorter lived then the `DeltaTable`.

I guess the only thing that "survives" is the `DeltaTableState`. Maybe a way to go is to have that as an optional field on the`DeltaTable`. Then once the first operation is performed, we populate that field in the returned table and re-use if for subsequent operations. This may also give us a way to capture user configuration for such things that would take a similar route through the code as the storage options do right now.

roeap (approved):
LGTM!

As you mentioned in one of the comments, going forward we should find a more central place to handle the configuration, and also harmonize how we register object stores on the datafusion runtime.

[Elsewhere](https://github.com/delta-incubator/hydrofoil/tree/main/deltalake/sql), I have been experimenting with more formally integrating with the datafusion query planning, in the hopes that we may adopt parts of this here.

roeap (approved):

PR #1463, opened 2023-06-14
fix(python): match Field signatures

guilhem-dvr:
# Description

Field signatures are not identical on the `type` field, which makes instantiating deltalake Fields with the `type` keyword impossible in python - and the `name` keyword as well for that matter.

# Related Issue(s)

# Documentation


wjones127 (approved):
Thanks for the fix!

PR #1477, opened 2023-06-19
fix: add `sizeInBytes` to _last_checkpoint and change `size` to # of actions

### LINKED ISSUE #1468: _last_checkpoint size field is incorrect
The `size` field in the `_last_checkpoint` file should be the number of actions that are stored in the checkpoint.  The code is instead writing the size of the file (which could be written to the optional field `sizeInBytes`).

[field definitions](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#last-checkpoint-file)
```
size | The number of actions that are stored in the checkpoint.
...
sizeInBytes | The number of bytes of the checkpoint. This field is optional.
```

https://github.com/delta-io/delta-rs/blob/0dda99bbb080d5cf8a6ecc350d0a789e7fa56549/rust/src/action/checkpoints.rs#LL122C43-L122C43
```
    let size = parquet_bytes.len() as i64;
    let checkpoint = CheckPoint::new(version, size, None);
```

cmackenzie1:
# Description
The `size` field should be the number of actions stored in the checkpoint while `sizeInBytes` is used for the total size in bytes.

Added `CheckPointBuilder` to make the creation of these easier to use.

# Related Issue(s)
- Closes #1468 

# Documentation

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#last-checkpoint-file

### DIFF
--- a/rust/src/action/mod.rs
+++ b/rust/src/action/mod.rs
@@ -791,11 +791,7 @@ pub(crate) async fn find_latest_check_point_for_version(
                 continue;
             }
             if cp.is_none() || curr_ver > cp.unwrap().version {
-                cp = Some(CheckPoint {
-                    version: curr_ver,
-                    size: 0,
-                    parts: None,
-                });
+                cp = Some(CheckPoint::new(curr_ver, 0, None));
             }
             continue;
         }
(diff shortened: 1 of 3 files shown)


wjones127 (dismissed):

rust/src/action/mod.rs
```diff
@@ -791,11 +791,7 @@ pub(crate) async fn find_latest_check_point_for_version(
                 continue;
             }
             if cp.is_none() || curr_ver > cp.unwrap().version {
-                cp = Some(CheckPoint {
-                    version: curr_ver,
-                    size: 0,
-                    parts: None,
-                });
+                cp = Some(CheckPoint::new(version, 0, None));
```
cmackenzie1, line 794: @wjones127 this was the culprit. Fixing...

```suggestion
                cp = Some(CheckPoint::new(curr_ver, 0, None));
```

wjones127 (approved):

PR #1482, opened 2023-06-21
feat: expose create_add to the public

### LINKED ISSUE #1458: Exposing create_add in the API
# Description
It would be nice if [create_add](https://github.com/delta-io/delta-rs/blob/8294c5f0e6fe441fa735a10726a81fc74e506c97/rust/src/writer/stats.rs#LL16C1-L16C26) was exposed somehow in the API to be used on its own, instead of keeping it public in the crate. Is this possible?

**Use Case**
We're specifically interested in extracting the [stats value](https://github.com/delta-io/delta-rs/blob/8294c5f0e6fe441fa735a10726a81fc74e506c97/rust/src/writer/stats.rs#LL37C8-L37C8) that's returned from this function and pass it to our implementation directly.

atefsaw:
# Description
This PR exposes create_add to the public

# Related Issue(s)
- closes #1458

# Documentation

### DIFF
--- a/rust/src/writer/mod.rs
+++ b/rust/src/writer/mod.rs
@@ -13,10 +13,11 @@ use crate::DeltaTable;
 
 pub use json::JsonWriter;
 pub use record_batch::RecordBatchWriter;
+pub use stats::create_add;
 
 pub mod json;
 pub mod record_batch;
-pub mod stats;
+pub(crate) mod stats;
 pub mod utils;
 
 #[cfg(test)]
--- a/rust/src/writer/mod.rs
+++ b/rust/src/writer/mod.rs
@@ -24,7 +24,7 @@ pub mod test_utils;
 
 /// Enum representing an error when calling [`DeltaWriter`].
 #[derive(thiserror::Error, Debug)]
-pub enum DeltaWriterError {
+pub(crate) enum DeltaWriterError {
     /// Partition column is missing in a record written to delta.
     #[error("Missing partition column: {0}")]
     MissingPartitionColumn(String),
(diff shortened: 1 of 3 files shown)


rust/src/writer/mod.rs
```diff
@@ -16,15 +16,15 @@ pub use record_batch::RecordBatchWriter;
...
+pub mod stats;
 pub mod utils;
 
 #[cfg(test)]
 pub mod test_utils;
 
 /// Enum representing an error when calling [`DeltaWriter`].
 #[derive(thiserror::Error, Debug)]
-pub(crate) enum DeltaWriterError {
+pub enum DeltaWriterError {
```
roeap, line 27: Not exposing too many error variants is actually deliberate :). In the end we want to get to a place where we only expose `DeltaError` to the outside.

rust/src/writer/stats.rs
```diff
@@ -13,6 +14,7 @@ use parquet::{
 use super::*;
 use crate::action::{Add, ColumnValueStat, Stats};
 
+/// Creates an [`Add`] log action struct.
 pub(crate) fn create_add(
```
roeap, line 18: this is still private to the crate, right? Was it more about the stats module?

atefsaw: Sorry, that was a mistake. I am interested in making create_add public.

rust/src/writer/mod.rs
atefsaw: I understand. Should I return DeltaTableError instead of DeltaWriterError in create_add or do you suggest something else?

roeap:
Thanks for making the updates @atefsaw!

Right now we are refactoring a lot of things, so I'd like to keep new things we expose to a minimum. Creating an Add action from parquet metadata is something the I agree is useful to others, but maybe we could make the `stats` module itself again private to the crate, and just re-export `create_add` at the writer level via.

```rust
pub use stats::create_add;
```

sorry for the extra effort.

atefsaw:
> Thanks for making the updates @atefsaw!
> 
> Right now we are refactoring a lot of things, so I'd like to keep new things we expose to a minimum. Creating an Add action from parquet metadata is something the I agree is useful to others, but maybe we could make the `stats` module itself again private to the crate, and just re-export `create_add` at the writer level via.
> 
> ```rust
> pub use stats::create_add;
> ```
> 
> sorry for the extra effort.

Hi @roeap , I've made the changes. can you please review? Thanks!

wjones127 (approved):

PR #1497, opened 2023-06-27
feat(rust): expose WriterProperties method on RecordBatchWriter and DeltaWriter

### LINKED ISSUE #1235: Support configuring parquet compression
# Description

We should probably make this configurable at the table level, if there's some standard for that.

Note: need to change the filename format:

https://github.com/delta-io/delta-rs/blob/8b854694d7eca2ef563601aac03f167fb7f0449c/rust/src/writer/utils.rs#L96-L97

Also, should we consider using zstd by default? It produces smaller files than snappy, but is faster than gzip (the Iceberg default). It was one of the better ones it my tests of Parquet compression, and I've seen suggested in several other venues for Parquet tables.

**Use Case**

**Related Issue(s)**

### LINKED ISSUE #1465: Expose compression options in optimize and write_deltalake

### LINKED ISSUE #1469: Expose Compression Options or WriterProperties for writing to Delta
Per the discussion linked... The current delta table writer does not expose setting compression currently and requires so more difficult workarounds. Exposing either and compression argument to the `write` method or allowing a custom `WriterProperties` or builder to the method would improve ease of use. See discussion for relevant code links.

### Discussed in https://github.com/delta-io/delta-rs/discussions/1464

<sup>Originally posted by **theelderbeever** June 14, 2023</sup>
As the question states above I would like to use zstd instead of snappy for writing parquet files to delta lake using the rust crate. I haven't been able to find anything that mentions customizing this in the writer. Any clarity would be super helpful.</div>

theelderbeever:
# Description
Adds the capability to pass a configured WriterProperties to the `RecordBatchWriter` and `DeltaWriter` similar to how the `OptimizeBuilder` can be updated.

# Related Issue(s)
- closes #1469 
- closes #1235 

# Documentation

### DIFF
--- a/rust/examples/recordbatch-writer.rs
+++ b/rust/examples/recordbatch-writer.rs
@@ -14,8 +14,12 @@ use deltalake::errors::DeltaTableError;
 use deltalake::writer::{DeltaWriter, RecordBatchWriter};
 use deltalake::*;
 use log::*;
-
 use object_store::path::Path;
+use parquet::{
+    basic::{Compression, ZstdLevel},
+    file::properties::WriterProperties,
+};
+
 use std::collections::HashMap;
 use std::sync::Arc;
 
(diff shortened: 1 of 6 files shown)


theelderbeever:
This is my first time contributing to a Rust repo so happy for any and all feedback and to be told I am on the completely wrong path.

theelderbeever:
Not sure why the python lint is suddenly not passing... Seems to be a mypy issue with a `# type: ignore` that it is incorrectly saying isn't necessary. There is a global workaround by changing updating the `make check-python` command to run `mypy --no-warn-unused-ignores`.

wjones127:
@theelderbeever don't worry about the mypy. It's from a new version of the checker. I have a fix in another PR, so that'll be green once that PR is merged.

theelderbeever:
@wjones127 Gotcha makes sense. I can revert my last commit that adds the ignore flag then.

rust/examples/recordbatch-writer.rs
wjones127, line 1: Thanks for adding this to the example :)

wjones127 (approved):
Thanks for doing this. Looks good to me!

PR #1498, opened 2023-06-28
fix: avoid writing statistics for binary columns to fix JSON error

### LINKED ISSUE #1493: Json error: Binary is not supported by JSON when writing checkpoint files
# Environment

**Delta-rs version**: Latest (master)

**Environment**: 
- **Cloud provider**: AWS
- **OS**: OSX, docker debian
- **Other**:

***
# Bug

**What happened**:

When writing checkpoints for a table with binary columns via:

```
        checkpoints::create_checkpoint(&table).await?;
```

I get the error

```
Error: Failed to convert into Arrow schema: Json error: Binary is not supported by JSON

Caused by:
    Json error: Binary is not supported by JSON
```

The checkpoint file should be written

**How to reproduce it**:

Write a table with binary columns and try to write a checkpoint

**More details**:

Here's the relevant logs from before the error:

```
[2023-06-26T14:00:57Z DEBUG deltalake::checkpoints] Writing parquet bytes to checkpoint buffer.
[2023-06-26T14:00:57Z DEBUG deltalake::checkpoints] Writing to checkpoint parquet buffer...
```

ChewingGlass:
# Description
Avoid writing statistics for binary columns to fix JSON error thrown by Arrow

# Related Issue(s)
 - closes #1493

### DIFF
--- a/rust/src/action/checkpoints.rs
+++ b/rust/src/action/checkpoints.rs
@@ -287,7 +287,7 @@ fn filter_binary(schema: &Schema) -> Schema {
     Schema::new(
         schema
             .get_fields()
-            .into_iter()
+            .iter()
             .flat_map(|f| match f.get_type() {
                 SchemaDataType::primitive(p) => {
                     if p != "binary" {
--- a/rust/src/action/checkpoints.rs
+++ b/rust/src/action/checkpoints.rs
@@ -283,7 +283,6 @@ pub async fn cleanup_expired_logs_for(
 
 // Filter binary from the schema so that it isn't serialized into JSON,
 // as arrow currently does not support this.
-// https://github.com/delta-io/delta-rs/issues/1493
 fn filter_binary(schema: &Schema) -> Schema {
     Schema::new(
         schema
@@ -309,7 +308,9 @@ fn filter_binary(schema: &Schema) -> Schema {
     )
 }
 
-fn parquet_bytes_from_state(state: &DeltaTableState) -> Result<bytes::Bytes, ProtocolError> {
+fn parquet_bytes_from_state(
+    state: &DeltaTableState,
+) -> Result<(CheckPoint, bytes::Bytes), ProtocolError> {
     let current_metadata = state.current_metadata().ok_or(ProtocolError::NoMetaData)?;
 
     let partition_col_data_types = current_metadata.get_partition_col_data_types();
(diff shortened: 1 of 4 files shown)


wjones127 (commented):
This file is a bit messy, but your changes make sense. :/

I *think* you could accomplish the same thing in a more targeted way by removing and binary columns using `collect_stats_conversion` and `apply_stats_conversion`. Use `collect_stats_conversion` to get which columns are binary, and then apply to remove them from the value (can use `serde_json::Value::remove()`). LMK if that works instead.

ChewingGlass:
Unfortunately you can't do this with stats conversion, the error occurs when making the ReaderBuilder, not when applying the values.

rust/src/writer/stats.rs
```diff
@@ -238,8 +239,16 @@ impl StatsScalar {
...
-                Ok(Self::Decimal(val))
+
+                let str_val = val.to_string();
+                let decimal_string = if str_val.len() > *scale as usize {
+                    let (integer_part, fractional_part) =
+                        str_val.split_at(str_val.len() - *scale as usize);
+                    format!("{}.{}", integer_part, fractional_part)
+                } else {
+                    format!("0.{}", str_val)
+                };
```
wjones127, line 250: I think we also want to have negative scales. Also, could we unit test this directly in this file?

ChewingGlass: Added support for negative scales. Not going to be able to add tests for a bit. I added you to my fork so you can commit if you want to take a crack at it.

wjones127:
We have a new failure for decimal, but I'm a little stumped on it for now.

wjones127:
I think for decimal, we might be able to start using the `arbitrary_precision` feature of `serde_json`. Not sure how nicely it plays with decimals yet, but it seems to be the suggested route.

rust/src/action/checkpoints.rs
```diff
@@ -281,6 +281,33 @@ pub async fn cleanup_expired_logs_for(
     }
 }
 
+// Filter binary from the schema so that it isn't serialized into JSON,
+// as arrow currently does not support this.
```
rtyler, line 285:
```suggestion
/// Filter binary from the schema so that it isn't serialized into JSON,
/// as arrow currently does not support this.
```

rtyler (commented):
Looks promising to me, I'm looking forward to this being merged

roeap:
@rtyler - i fear it may have been a bit too early to merge this., since the rust test are failing for this. since 0.13 was just released, we should look into releasing a fix asap. bot sure if @wjones127 already knows whats going on? Otherwise I might look into it once available again.

rtyler:
@roeap Yes, I'm looking at this now. I must have been moving too quickly through some PRs. I now see that the test was failing but I'm not sure why I merged this :facepalm:  I must have mistaken this pull request for another one which I had open at the time.

I am going to land a revert into `main` directly, and cut a 0.13.1 with that revert just to avoid causing anybody problems

Issue #837, opened 2022-09-23
Implement restore command


wjones127:
# Description

This is another easy to implement command that shouldn't require writing data files. It makes the history command much more useful!

See docs: https://docs.databricks.com/delta/delta-utility.html#restore-delta-table

**Use Case**

**Related Issue(s)**

Bernolt:
Hello,
I will start working on this.
I will use a similar approach as for Vacuum implementation.

roeap:
Hi @Bernolt - there is a PR open (#852) which will intorduce some changes to the operations APIs. It is similar to the way how vacuum is implemented right now, but leverages the `IntoFuture` trait stabilized in 1.64. Not sure how far along you are, but if still feasible maybe consider this :).

loleek:
Is this issue still available?

wjones127:
Yes, we haven't gotten a PR for this yet, but would be happy to review one. I haven't heard from Bernolt recently.

loleek:
OK.I will look into it and try to file a PR.

loleek:
Been busy with other things lately, now back to this feature. I've already read delta's implement, and will try to file recent days.

ion-elgreco:
@wjones127 it's available in rust and python now, so we can close it :)

PR #1502, opened 2023-06-29
feat: implement restore operation

loleek:
# Description
This is a implementation of the Restore Command.

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/837

### DIFF
--- a/rust/src/operations/restore.rs
+++ b/rust/src/operations/restore.rs
@@ -183,7 +183,7 @@ async fn execute(
                 extended_file_metadata: Some(false),
                 partition_values: Some(a.partition_values.clone()),
                 size: Some(a.size),
-                tags: a.tags.clone(),
+                tags: a.tags,
             }
         })
         .collect();
--- a/rust/src/operations/restore.rs
+++ b/rust/src/operations/restore.rs
@@ -16,7 +16,7 @@
 //! # Example
 //! ```rust ignore
 //! let table = open_table("../path/to/table")?;
-//! let (table, metrics) = RestoreBuilder::new(table.object_store(), table.state).restore_to_version(1).await?;
+//! let (table, metrics) = RestoreBuilder::new(table.object_store(), table.state).with_version_to_restore(1).await?;
 //! ````
 
 use std::cmp::max;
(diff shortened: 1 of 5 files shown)


loleek:
I will file python binding implement in another PR. Is that OK?

loleek:
I saw integration test failed. But it looks like not caused by my PR.

rust/src/operations/restore.rs
```diff
@@ -0,0 +1,288 @@
...
+            store,
+            version_to_restore: None,
+            datetime_to_restore: None,
+            ignore_missing_files: false,
+            protocol_downgrade_allowed: false,
+        }
+    }
+
+    /// Set the version to restore
+    pub fn restore_to_version(mut self, version: i64) -> Self {
```
roeap, line 99: For consisitency with other builders it would be great to name the methods to configure options `with_*`

rust/src/operations/restore.rs
```diff
@@ -0,0 +1,288 @@
...
+                partition_values: Some(a.partition_values.clone()),
+                size: Some(a.size),
+                tags: a.tags,
+            }
+        })
+        .collect();
+
+    if !ignore_missing_files {
+        check_files_available(object_store.as_ref(), &files_to_add).await?;
+    }
```
roeap, line 193: interesting - would not have thought that it is useful, to delivberately create a state with missing files, but it seems the reference implementation also supports this :).

rust/src/operations/restore.rs
```diff
@@ -0,0 +1,288 @@
...
+    match try_commit_transaction(object_store.as_ref(), &commit, commit_version).await {
+        Ok(_) => {}
+        Err(err @ TransactionError::VersionAlreadyExists(_)) => {
+            return Err(err.into());
+        }
+        Err(err) => {
+            object_store.delete(&commit).await?;
+            return Err(err.into());
+        }
+    }
```
roeap, line 242: Just making sure I understand. For restore we always want to fail, if any other operation was performed on the table?

roeap (changes requested):
Thaks for taking care of this @loleek - left a minor comment re naming, otherwise looking great!

roeap:
> I saw integration test failed. But it looks like not caused by my PR.

Indeed it is not, unfortunately we have some flakiness in the tests due to memory issues.

rust/src/operations/restore.rs
loleek: done

rust/src/operations/restore.rs
loleek: The data file may missing for many reason, deleted unintentionally or by vacuum operation.This will cause data missing if always allow to restore, so we must make users aware of the problem.

rust/src/operations/restore.rs
loleek: Yes, if any other operation was performed, the restore is invalid because the ADD and REMOVE files are computed based on latest snapshot state and state of restored version.

roeap (approved):
Thanks for the great contribution, @loleek!

PR #1511, opened 2023-07-03
chore: remove deprecated table functions

roeap:
# Description

Removing the deprecated commit logic on `DeltaTable` along with some other deprecations. A large chunk of the changes is just related to adjusting all test code use current table creation and commit logic.


roeap:
@rtyler - integration tests are passing now, our friend memory at play 😆.

rtyler (approved):

PR #1512, opened 2023-07-04
refactor: unify with_predicate for delete ops

Blajda:
# Description
A small refactor on the delete op to have the same interface as the update op

Allows a single method to be used for predicates represented as an Datafusion `Expr` or as a string.


roeap (approved):
Great improvement - thanks @Blajda!

PR #1519, opened 2023-07-06
feat: add nested struct supports

### LINKED ISSUE #1518: Add nested struct supports
# Description
Add nested struct supports.

**Use Case**

**Related Issue(s)**

haruband:
# Description
This PR adds nested struct supports.

# Related Issue(s)
- closes #1518 

# Documentation

### DIFF
--- a/rust/src/operations/write.rs
+++ b/rust/src/operations/write.rs
@@ -545,7 +545,10 @@ mod tests {
     use super::*;
     use crate::operations::DeltaOps;
     use crate::writer::test_utils::datafusion::get_data;
-    use crate::writer::test_utils::{get_delta_schema, get_record_batch};
+    use crate::writer::test_utils::{
+        get_delta_schema, get_delta_schema_with_nested_struct, get_record_batch,
+        get_record_batch_with_nested_struct,
+    };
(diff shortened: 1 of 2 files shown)


rust/src/operations/write.rs
```diff
@@ -473,19 +473,59 @@ impl std::future::IntoFuture for WriteBuilder {
...
+            }
         } else {
             false
         }
     })
 }
 
+fn can_cast_batch(from_schema: &ArrowSchema, to_schema: &ArrowSchema) -> bool {
+    can_cast_batch_fields(from_schema.fields(), to_schema.fields())
+}
```
rtyler, line 498:
```suggestion
```
Since there is only one callsite of this method and it is not a public API, I figure it makes sense to just remove it

rtyler (commented):
I think this mostly looks good and can be merged with some minor tweaks. If you don't happen to get to it, i'll knock it out tomorrow

rust/src/operations/write.rs
haruband: I agree with you. I pushed new commit.
Thanks for the review.

rtyler (approved):

PR #1520, opened 2023-07-08
fix: handle nulls in file-level stats

### LINKED ISSUE #1496: Pyarrow is_null filter not working as expected after loading using deltalake
# Environment

**Delta-rs version**: 0.10.0

**Binding**: Python

**Environment**: Seemed to have the same issue across Windows 10 and Amazon Workspaces
- **Cloud provider**:
- **OS**: 
- **Other**:

***
# Bug
Explained in the following stackoverflow question: https://stackoverflow.com/questions/76557635/possible-bug-in-using-pyarrow-is-null-function-with-delta-tables
**What happened**: The is_null filter only returned rows from partitions where all rows have null values for the column in question i.e. any rows with null values that belonged to a partition that had other non-null values was not returned. Using pa.dataset.dataset("/path/to/table") works as expected.

**What you expected to happen**: I expected all rows with null values to be filtered.

**How to reproduce it**:
```
import pyarrow as pa
import pyarrow.compute as pc

partition_column = pa.array(['a', 'a', 'b', 'b', 'c', 'c'], type=pa.string())
int_value = pa.array([12, None, None, None, 15, 67], type=pa.int64())
names = ["partition_column", "int_value"]

table = pa.Table.from_arrays([partition_column, int_value], names=names)
write_deltalake("./resources/is_null_test_table", table, partition_by="partition_column")

ds = DeltaTable("./resources/is_null_test_table").to_pyarrow_dataset()
pat = ds.to_table(filter=(pc.field("int_value").is_valid()))
print(pat.to_pandas())
```

I'd expect this code to retrieve the 3 rows with None values but it only retrieves 2 of them - the ones which belong to the same partition ('b'). 

**More details**:

wjones127:
# Description

Fixes issue where predicate pushdown isn't working for null values. This adds tests for both columns and partition columns.

# Related Issue(s)

- closes #1496

# Documentation

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -36,7 +36,7 @@ jobs:
         run: make check-rust
 
   test-minimal:
-    name: Python Build (Python 3.7 PyArrow 7.0.0)
+    name: Python Build (Python 3.7 PyArrow 8.0.0)
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: "-C debuginfo=0"
@@ -70,7 +70,7 @@ jobs:
           source venv/bin/activate
           make setup
           # Install minimum PyArrow version
-          pip install -e .[pandas,devel] pyarrow==7.0.0
+          pip install -e .[pandas,devel] pyarrow==8.0.0
         env:
           RUSTFLAGS: "-C debuginfo=0"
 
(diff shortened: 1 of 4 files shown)


wjones127:
Needed to upgrade the minimal PyArrow version, I think due to https://issues.apache.org/jira/browse/ARROW-12659

roeap (approved):
LGTM!

Issue #850, opened 2022-09-26
Implement merge command


darabos:
If I understand Delta's docs (https://docs.delta.io/latest/delta-update.html#upsert-into-a-table-using-merge) correctly, _merge_ is different from a generic write. Rather than appending to or replacing an existing table, it updates (rewrites) the affected target files. Looks like it is also implemented separately. ([MergeIntoCommand.scala](https://github.com/delta-io/delta/blob/v2.1.0/core/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommand.scala))

**Use Case**

Writing a new table to Delta is not so exciting for me. I could just write a Parquet table. But updating an existing table has no substitute. I'm working on integrating a non-Spark computation system with Delta, and the merge command looks like the best way to put back results into the original table.

**Related Issue(s)**

#542 added basic writing support.

darabos:
From MergeIntoCommand.scala it looks like a really complex operation. I may be able to work on this. It would be great if you could help me estimate how much effort this would be. Thanks for any input!

wjones127:
Merge is indeed a complex operation. We started with just append/overwrite because that doesn't require a query engine; you can just pass some data and write to parquet. But for things like delete and merge, we'll have to integrate a query engine and/or allow plugging one in. (Since we need a way to query for which rows match.) We're in the early stages of designing that, so it might be a couple months before we are even ready to start implementing merge.

That being said, we'd definitely like to support this and would welcome contributions 😄 . Are you planning on building your integration on top of the Rust library or the Python library? And for your use case are you more interested in having the query engine parts provided for you (probably DataFusion) or being able to pass in your own implementation?

darabos:
Thanks! We would be using the Python interface and already have a Dask system on hand. It would be great if we could use that!

nkarpov:
[...]

PR #1522, opened 2023-07-09
feat: merge operation

Blajda:
# Description
Implement the Merge operation using Datafusion.

Currently the implementation rewrites the entire DeltaTable limiting the files that are rewritten will be performed in future work. 

# Related Issue(s)
- progresses #850

# Documentation


ion-elgreco:
@Blajda @wjones127 does that mean the full table will be loaded in memory first so it can rewrite the whole table?

Blajda:
>  does that mean the full table will be loaded in memory first so it can rewrite the whole table?

Yes since this is using a nested loop join both the source and target need to be fully loaded into memory. There would need to be follow up improvements to perform pruning and to use a Hash Join / Merge Join.

ion-elgreco:
> >  does that mean the full table will be loaded in memory first so it can rewrite the whole table?
> 
> Yes since this is using a nested loop join both the source and target need to be fully loaded into memory. There would need to be follow up improvements to perform pruning and to use a Hash Join / Merge Join.

Good to know, still useful for datasets that fit in memory then. Hopefully it can be `merged` soon ;) Then I could try taking a go on the Python bindings and start using it as :)

rust/src/operations/merge.rs
```diff
@@ -0,0 +1,1614 @@
...
+    /// Ther order of match clauses matter.
+    ///
+    /// #Example
+    /// ```rust ignore
+    /// let table = open_table("../path/to/table")?;
+    /// let (table, metrics) = DeltaOps(table)
+    ///     .merge(source, col("id").eq(col("source.id")))
+    ///     .with_source_alias("source")
+    ///     .when_matched_delete(|delete| {
+    ///         update
```
wjones127, line 203: This meant to be `delete`?

rust/src/operations/merge.rs
```diff
@@ -0,0 +1,1614 @@
...
+    ///             .predicate(col("source.delete"))
+    ///     })?
+    ///     .when_matched_update(|update| {
+    ///         update
+    ///             .update("value", col("source.value"))
+    ///             .update("modified", col("source.modified"))
+    ///     })?
```
wjones127, line 210: should this be delete or is this meant to show how to mix delete and update?

rust/src/operations/merge.rs
```diff
@@ -0,0 +1,1614 @@
...
+    /// Specify how to insert a source record when it does not match with a
+    /// target record
+    ///
+    /// Multiple not match clasues can be specified and their predicates are
+    /// evaluated to determine if the corresponding operation are performed.
+    /// Only the first clause that results in an satisfy predicate is executed.
+    /// Ther order of not match clauses matter.
+    ///
+    /// #Example
+    /// ```rust ignore
```
wjones127, line 237: I don't think the example uses this method. Does it?

rust/src/operations/merge.rs
```diff
@@ -0,0 +1,1614 @@
...
+    }
+    expressions.push((
+        Arc::new(expressions::Literal::new(true.into())),
+        "__delta_rs_target".to_owned(),
+    ));
+    let target = Arc::new(ProjectionExec::try_new(expressions, target.clone())?);
+
+    // TODO: Currently a NestedLoopJoin is used but we should target to support SortMergeJoin
+    // This would require rewriting the join predicate to only contain equality between left and right columns and pushing some filters down
+    // Ideally it would be nice if the optimizer / planner can pick the best join so maybe explore rewriting the entire operation using logical plans.
```
wjones127, line 610: Agreed. It would be nice if we could write this instead with the DataFrame API or as a logical plan, rather than having to write it as a physical one.

rust/src/operations/merge.rs
```diff
@@ -0,0 +1,1614 @@
...
+//! predicate.  A full outer join is performed which results in source and
+//! target records that match, source records that do not match, or target
+//! records that do not match.
+//!
+//! Users can specify update, delete, and insert operations for these categories
+//! and specify additional predicates for finer control. The order of operations
+//! specified matter.  See [`MergeBuilder`] for more information
+//!
+//! *WARNING* The current implementation rewrites the entire delta table so only
+//! use on small to medium sized tables. Enhancements tracked at #850
```
wjones127, line 11: Also worth noting that it currently loads the entire thing into memory and uses only 1 thread, IIUC.

rust/src/operations/merge.rs
```diff
@@ -0,0 +1,1614 @@
...
+        );
+        assert_eq!(
+            parameters["notMatchedPredicates"],
+            json!(r#"[{"actionType":"insert"}]"#)
+        );
+        // Todo: Expected this predicate to actually be 'value = 1'. Predicate should contain a valid sql expression
+        assert_eq!(
+            parameters["notMatchedBySourcePredicates"],
+            json!(r#"[{"actionType":"update","predicate":"value = Int32(1)"}]"#)
+        );
```
wjones127, line 1246: Should this be fixed before merging?

rust/src/operations/update.rs
```diff
@@ -291,7 +288,22 @@ async fn execute(
     let projection_predicate: Arc<dyn ExecutionPlan> =
         Arc::new(ProjectionExec::try_new(expressions, parquet_scan)?);
 
-    let count_plan = Arc::new(UpdateCountExec::new(projection_predicate.clone()));
+    let count_plan = Arc::new(MetricObserverExec::new(
```
wjones127, line 291: Nice job abstraction this!

wjones127 (changes requested):
Sorry it took me so long to get to this.

This is a big operation, thanks for starting this.

I think since we are using the inner loop join, I don't think we can expect great performance out of this except in very small tables. But it will take some work to move towards logical plans that can choose more efficient joins. So for this PR, I think it's best we mostly focus on correctness. The main thing I see is that the record SQL statement isn't quite correct.

rust/src/operations/merge.rs
Blajda: Yeah I think we can hold on merging this until it's resolved. I raised an ask on the DF tracker here: https://github.com/apache/arrow-datafusion/issues/7165
This same issue exits for delete and update too so they'll need to be fixed too.

rust/src/operations/update.rs
Blajda: Thanks :)

Blajda:
No worries about the delays I understand efforts on this project are mostly volunteer and I appreciate your time :)

I've updated the docs with the correct methods and refined description for each method.

What do you think about putting this operation behind a feature flag? Anyone using it would opt in having understood the current issues and it allow smaller PRs to be submitted for the enhancements.

For next steps, I have another branch where I introduce a `DeltaScanBuilder` and updated `DeltaScan` to include a  optional column that maintains which file a particular record came from. The implementation is similar to the `FindFiles` and would be used here to determine which files actually have modifications.

wjones127:
> What do you think about putting this operation behind a feature flag?

I don't feature flag is necessary, but we could label it "experimental" maybe. I think as long as the docs give up-to-date notes on limitations this should be fine.

wjones127 (approved):

Blajda:
Thanks @wjones127  for taking the time to review this. I'll break future enhancements into smaller chunks to make it easier.

roeap:
Hoping that merging main will resolve some of the failures, which seem to be related to recently added deletion vector fields...

Blajda:
Updated code to add new dv field. hopefully it passes all test now.

roeap:
There is one more failure - probably related to a mypy update o.a. - we could just `# type: ignore` that line for now.

Other then that we are good to go :)

Blajda:
:confused: Seems to fail with the ignore and without. check-python is passes locally for me

roeap:
yea, this is a bit annoying :( - unfortunately I do not have permissions to disable or bypass required checks ... 

Should we try once more without? Local checks are passing for me as well without the comment.

Alternatively we could try `# noqa`, which is quite common, but not sure if mypy honors it.

roeap:
or maybe even both, not sure how make works. i.e. would it fail early if now the ruff check is complaining? otherwise `# noqa` would silence ruff ...

Blajda:
@roeap @wjones127 Please provide approval on the workflows. The lint change should hopefully allow everything to pass now.

ion-elgreco:
@roeap @wjones127 can you approve the workflow, so this can get merged in the next release?

Blajda:
@wjones127 Please approve the workflow. Merged with the latest changes in main.

wjones127 (approved):

Blajda:
@wjones127 Thanks for reviewing and merging.

PR #1529, opened 2023-07-11
feat: add restore command in python binding

loleek:
# Description
This is a implementation of the Restore Command for python binding.

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/837


python/src/lib.rs
```diff
@@ -318,6 +319,35 @@ impl RawDeltaTable {
...
+        version_to_restore: Option<i64>,
+        datetime_to_restore: Option<&str>,
+        ignore_missing_files: bool,
+        protocol_downgrade_allowed: bool,
+    ) -> PyResult<String> {
+        let mut cmd = RestoreBuilder::new(self._table.object_store(), self._table.state.clone());
+        if let Some(version) = version_to_restore {
+            cmd = cmd.with_version_to_restore(version)
+        }
+        if let Some(ds) = datetime_to_restore {
```
rtyler, line 335: Since both of these parameters are optional there is a potential for a user to inadvertently provide _both_ arguments and that would result in two restore "commands" being passed into the builder, and the `datetime` one is the one which will win in that case.

I think it might be useful to raise an error here to indicate to the user that they should not provide both parameters, or issue a warning that the `version_to_restore` will be ignored.

loleek: This check has been done here.[parameter check](https://github.com/delta-io/delta-rs/blob/main/rust/src/operations/restore.rs#L132). Should I do double check?

wjones127: I think the check is good. But TBH, for the Python interface, I think we should combine the parameters into one. See my other comment.

python/deltalake/table.py
```diff
@@ -459,6 +459,30 @@ def pyarrow_schema(self) -> pyarrow.Schema:
         )
         return self.schema().to_pyarrow()
 
+    def restore(
+        self,
+        version_to_restore: Optional[int] = None,
+        datetime_to_restore: Optional[str] = None,
+        ignore_missing_files: bool = False,
+        protocol_downgrade_allowed: bool = False,
+    ) -> Dict[str, Any]:
```
wjones127, line 468: Thanks for working on this! I think there's a few UX changes I'd like to see to make this better for Python users:

 1. Combine the first two arguments so the user can always specify just the first argument, rather than sometimes having to go `table.restore(datetime_to_restore="2020-05-01T00:47:31-07:00")`.
 2. Support passing in a `datetime` instead of just a string. That way users don't have to guess which format they need to specify it in.
 3. Make the other arguments keyword-only. It's all boolean arguments and those are terrible without labels (`table.restore(1, True, True)` 😱 )

```suggestion
    def restore(
        self,
        target: Optional[int | datetime | str] = None,
        *,
        ignore_missing_files: bool = False,
        protocol_downgrade_allowed: bool = False,
    ) -> Dict[str, Any]:
```

wjones127 (changes requested):
I appreciate the high-quality tests. I think there's some UX stuff on the Python API that could be improved, then I think this is ready to be merged.

python/deltalake/table.py
loleek: Unfortunately find a problem when I do local test. This caused by a copy action on delta table, the last_modified timestamp change and [we mapping version with timestamp by the file meta](https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L437). Is this expected behavior? @wjones127 Maybe we should mapping the version with timestamp in commit info.

wjones127: I think that's a hack to make sure the timestamps are monotonically increasing. We could switch to doing what Spark does, but that requires some complicated logic to ensure the parsed timestamps are indeed monotonic. But it would probably work better with the copy action I guess.

loleek: There has a sphinx doc build step failed in checks. I'm not sure how to fix it. @wjones127 
Here is the error message.
Warning, treated as error:
/home/runner/work/delta-rs/delta-rs/python/deltalake/table.py:docstring of deltalake.table.DeltaTable.restore::py:class reference target not found: datetime.datetime
make: *** [Makefile:87: build-documentation] Error 2
Error: Process completed with exit code 2.

wjones127: Hi @loleek sorry for the delay. I've created a PR into your branch with the fixes for that CI failure: https://github.com/loleek/delta-rs/pull/1

loleek: Thx.

loleek: All request changed resolved. Would you mind do a re-review? @wjones127

wjones127 (approved):

PR #1536, opened 2023-07-15
chore: fix the latest clippy warnings with the newer rustc's

rtyler:


rtyler:
I am going to land this without review, since everybody is probably asleep right now :smile: This will fix `main` which I believe was broken when rustc 1.71 was released

PR #1543, opened 2023-07-17
refactor: enable s3 feature by default

### LINKED ISSUE #270: Enable S3 storage feature by default
# Description

**Use Case**

Follow up for https://github.com/delta-io/delta-rs/discussions/208. We have decided to enable s3 storage backend by default and provide a new feature flag to disable s3 storage backend if they desire.

xuky76:
# Description

Commit switches s3 feature to disable-s3 feature. Changes are reflected in documentation

# Related Issue(s)
- closes #270

### DIFF
--- a/rust/src/delta.rs
+++ b/rust/src/delta.rs
@@ -909,7 +909,7 @@ pub fn crate_version() -> &'static str {
 #[cfg(test)]
 mod tests {
     use super::*;
-    #[cfg(any(feature = "s3", feature = "s3-native-tls"))]
+    #[cfg(any(not(feature = "disable-s3"), feature = "s3-native-tls"))]
     use crate::builder::DeltaTableBuilder;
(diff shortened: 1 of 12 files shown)


roeap:
Thanks for your contribution @xuky76.

In this case I guess the more straight forward way to achieve this, would be to add "s3" to the default features in the cargo.toml file.

Since this issue was originally created a lot has changed though, and I am not sure there is any more benefit in having S3 enabled by default, as we have a similar number of azure users by now.

More generally, we are also moving some of the additional locking logic into a separate place, so we can reuse it for other object stores, which also require an external locking mechanism.

xuky76:
Thanks for the prompt feedback @roeap.  That makes sense, I'll withdraw the PR and make sure to comment beforehand to see if an issue is current in future 😊

PR #1544, opened 2023-07-19
fix: revert premature merge of an attempted fix for binary column statistics

rtyler:
Mea culpa regarding #1498


roeap (approved):
Thanks for fixing this!

PR #1549, opened 2023-07-20
feat: buffered reading of transaction logs

eeroel:
# Description
This PR introduces two changes to how the commit log is read:
- When reading the commit log, get requests are made concurrently in a buffered stream. This decreases latency when there are many files in the log since the last checkpoint.
- Before scanning the log, the latest table version is found by listing with `list_with_offset`. This helps avoid a potential infinite loop in a slow reader/fast writer scenario, and makes it possible to buffer GET requests without making any unnecessary ones. This listing logic is now also used for the time-travel reading.

# Related Issue(s)

# Documentation


rust/src/delta.rs
```diff
@@ -540,8 +550,35 @@ impl DeltaTable {
...
+            let loc = commit_uri_from_version(i).clone();
+            async move {
+                let out = store2.get(&loc);
+                (i, out.await)
+            }
+        });
+
+        let buf_size = 50; // TODO
+
+        // why is mut needed here and is it OK?
```
rtyler, line 566: To answer the question in the comment here, the `mut` is needed because of the `log_buffer.next()` below which will consume from the buffer

rust/src/delta.rs
```diff
@@ -540,8 +550,35 @@ impl DeltaTable {
...
+        let mut log_buffer = {
+            match max_version {
+                Some(n) => log_stream.take((n - self.version() + 2).try_into().unwrap()).buffered(buf_size),
```
rtyler, line 569:
```suggestion
                Some(n) => log_stream.take((n - self.version() + 2).try_into()?).buffered(buf_size),
```

I'm not sure off the top of my head, but I think the `?` will work and help ensure that errors get propagated up to the `DeltaResult` returned by the function

rust/src/delta.rs
```diff
@@ -540,8 +550,35 @@ impl DeltaTable {
...
+                None => log_stream.take(usize::MAX).buffered(buf_size) // need take to have compatible types
+            }
+        };
+        
+        while let Some((new_version, actions)) = {
+            let next_commit = log_buffer.next().await.unwrap();
```
rtyler, line 575:
```suggestion
            let next_commit = log_buffer.next().await?;
```

This might be warranted as well, unwraps will panic the caller on errors which is not something we want

eeroel: In this case the `await` returned an `Option` so I just handled it as such in the match below. Now the case when the stream ends before expected is explicitly an error

rust/src/delta.rs
eeroel: This didn't work directly, but I added explicit handling for the error cases

rust/tests/read_delta_test.rs
```diff
@@ -527,6 +528,40 @@ async fn test_poll_table_commits() {
     assert!(matches!(peek, PeekCommit::UpToDate));
 }
 
+#[tokio::test]
```
eeroel, line 531: @rtyler Here I'm testing that when buffering > 1 the processing is faster. Is there a nice way to patch the `get` method to insert a delay there, so that the test would work reliably? And do you think this kind of slightly stochastic test is OK or necessary to have? Or is it in the benchmark territory?

rtyler: The simplest path would be to implement a wrapper struct which implements `ObjectStore` and then implement the delay you seek in its `get` implementation.

This test would need to be changed to construct the table in a different fashion, such as with the table builder .

rust/src/delta.rs
```diff
@@ -887,6 +938,20 @@ pub async fn open_table_with_version(
...
+    log_buffer_size: usize,
+) -> Result<DeltaTable, DeltaTableError> {
+    let table = DeltaTableBuilder::from_uri(table_uri)
+        .with_version(version)
+        .with_log_buffer_size(log_buffer_size)
+        .load()
+        .await?;
+    Ok(table)
+}
+
```
rtyler, line 954: We have been discouraging more use of these top level methods, I would prefer this function not be added and instead users who are interested in buffering to use the table builder method

rust/src/builder.rs
```diff
@@ -153,6 +169,12 @@ impl DeltaTableBuilder {
         self
     }
 
+    /// Sets `log_buffer_size` to the builder
+    pub fn with_log_buffer_size(mut self, log_buffer_size: usize) -> Self {
```
rtyler, line 173:
```suggestion
    pub fn with_buffer(mut self, log_buffer_size: usize) -> Self {
```

IMHO this can be more concise

eeroel:
Did some refactoring and revisions according to suggestions, and also exposed the parameter to Python. Confirmed the speed-up locally against a table in S3, `from_data_catalog` takes 13s with no concurrency and <2s with log_buffer_size=30

rust/src/builder.rs
```diff
@@ -365,7 +388,7 @@ lazy_static::lazy_static! {
 /// Extra slashes will be removed from the end path as well.
 ///
 /// Will return an error if the location is not valid. For example,
-pub(crate) fn ensure_table_uri(table_uri: impl AsRef<str>) -> DeltaResult<Url> {
+pub fn ensure_table_uri(table_uri: impl AsRef<str>) -> DeltaResult<Url> {
```
eeroel, line 391: This I figured might not be a good idea, but is there some other way of using the function in the test?

rust/src/delta.rs
```diff
@@ -540,8 +550,49 @@ impl DeltaTable {
...
+            }
+        };
+        
+        while let Some((new_version, actions)) = {
+            let next_commit = log_buffer.next().await;
+            match next_commit {
+                Some((v, Ok(x))) => Ok(Some((v, self.get_actions(v,  x.bytes().await?).await?))),
+                Some((_, Err(ObjectStoreError::NotFound { .. }))) => Ok(None),
+                Some((_, Err(err))) => Err(DeltaTableError::GenericError { source: Box::new(err) }),  // TODO ??
+                None => Err(DeltaTableError::Generic(String::from("Log stream closed unexpectedly!")))
```
rtyler, line 593: My understanding is perhaps a bit hazy, but isn't a `None` at the end of the stream an expected condition? Won't this always happen once the stream is at the end?

rust/tests/fs_common/mod.rs
```diff
@@ -121,3 +125,129 @@ pub async fn commit_actions(
     table.update().await.unwrap();
     version
 }
+
+
+#[derive(Debug)]
+pub struct SlowStore {
+    inner: DeltaObjectStore
+}
```
rtyler, line 133: :smiley:

rust/tests/read_delta_test.rs
```diff
@@ -527,6 +528,48 @@ async fn test_poll_table_commits() {
...
+async fn test_log_buffering() {
+    let path = "./tests/data/simple_table_with_no_checkpoint";
+    let max_iter = 10;
+    let buf_size = 10;
+
+    let location = deltalake::builder::ensure_table_uri(path).unwrap();
```
rtyler, line 537: I'm on the fence about letting this API leak out here, but let's roll with it for now

rust/src/delta.rs
eeroel: Yeah you're right, this is expected. I made this an Ok case and also updated the `.take` call to take `max_version - self.version()` elements instead of +2. Had some confusion there earlier due to this error.

rust/src/delta.rs
```diff
@@ -540,8 +550,55 @@ impl DeltaTable {
             self.version(),
         );
 
-        while let PeekCommit::New(new_version, actions) =
-            self.peek_next_commit(self.version()).await?
+        let max_version = max_version.and_then(|x| {
+            if x <= self.version() {
```
eeroel, line 554: @rtyler I refactored this max_version handling a bit so it's clearer. Here I made explicit the behavior that's on `main` currently: i.e. the table will be updated to the latest version if max_version is smaller or equal to the current version.

eeroel:
Did some refactoring and added a test to check that results are correct with different buffer sizes. I will still think a bit about how to limit the number of requests, as now there can be potentially a lot of unnecessary ones if max_version is not set

rust/src/delta.rs
```diff
@@ -540,14 +550,78 @@ impl DeltaTable {
             self.version(),
         );
 
-        while let PeekCommit::New(new_version, actions) =
-            self.peek_next_commit(self.version()).await?
+        // update to latest version if given max_version is not larger than current version
```
eeroel, line 553: @rtyler I ended up adding this bigger change to list the files to find out what the latest version is. The code is mostly copied over from another module so it needs refactoring. It's somewhat orthogonal to the buffered reading changes, but unless I'm missing something
- it's necessary in order to make sure the load actually completes
- it lets get requests to be buffered without creating any wasted requests (and thus the user doesn't need to be made aware of potential extra costs)

I understand though that this is a more major change because listing can be costly with big logs and in particular with no prefix filtering. And as far as I understood the protocol doesn't prescribe how to do this, but it does sort of hint at listing first being the way to read tables.

eeroel:
I think the implementation is starting to be done. The scope of the PR widened a bit from the original idea, but hope I'm not being silly and it's the right direction... Ideally the performance could be benchmarked using tables in s3, azure and gcp with 1. a huge number of log entries (to see the effect of listing), and 2. varying checkpoint intervals, 10-100 for example. 

Some remaining doubts I have:
- I currently set the buffer size limit arbitrarily to 50 but I wonder if such a limit is needed, or if it should be considerably higher? It's also not a very user-friendly knob as ideally the system would just work with with maximum performance.
- Is the original way of reading without listing something that should be retained as another mode? I imagine it's been implemented the way it this to avoid the cost of listing, and I can imagine it being useful when the log is huge and the underlying storage doesn't support `list_with_offset` style prefix filtering.

rust/src/builder.rs
wjones127: To keep it private, we can keep the tests within this file, in the tests section at the bottom.

rust/src/delta.rs
```diff
@@ -540,14 +556,49 @@ impl DeltaTable {
...
+            }
+        });
+        
+        let mut log_buffer = {
+            let n_commits = usize::try_from(max_version - self.version());
+            match n_commits {
+                Ok(n) => log_stream.take(n).buffered(buf_size),
+                Err(err) => return Err(DeltaTableError::GenericError { source: Box::new(err) })
+            }
+        };
```
wjones127, line 585: This seems a little complicated. Why couldn't the original iterator be `self.version() + 1..max_version`? And you can just early return if `self.version() >= max_version`?

rust/tests/data/simple_table_with_no_checkpoint/.part-00000-136c36f5-639d-4e95-bb0f-15cde3fb14eb-c000.snappy.parquet.crc
wjones127, line 1: If it's possible to generate this table using delta-rs, I'd strongly prefer we just created the table dynamically in a temp directory as part of the test setup. We usually check in tables as a last resort when there's something the Spark writer does that we can't read.

rust/src/builder.rs
```diff
@@ -153,6 +170,17 @@ impl DeltaTableBuilder {
         self
     }
 
+    /// Sets `log_buffer_size` to the builder
+    pub fn with_buffer(mut self, log_buffer_size: usize) -> DeltaResult<Self> {
+        if log_buffer_size > 50 {
+            return Err(DeltaTableError::Generic(String::from("Log buffer size should smaller than 51")));
```
wjones127, line 176: Why make this limit? Practically I don't think many will want to increase this, but seems fine to not have an explicit limit. In most cases, because of checkpoints there is already an implicit limit on the number of calls.

rust/src/builder.rs
```diff
@@ -111,6 +127,7 @@ impl DeltaTableLoadOptions {
             storage_backend: None,
             require_tombstones: true,
             require_files: true,
+            log_buffer_size: 1,
```
wjones127, line 130: I'm not sure we want to default to the least performant option here. I think I'd prefer:

```suggestion
            log_buffer_size: num_cpus::get() * 4,
```

Usually the memory allotted is proportional to number of CPUs, as is the capability of the machine to handle many concurrent requests, so it's a good heuristic.

rust/src/builder.rs
```diff
@@ -67,13 +67,22 @@ pub struct DeltaTableConfig {
...
+    /// last checkpoint.
+    pub log_buffer_size: usize
 }
 
 impl Default for DeltaTableConfig {
     fn default() -> Self {
         Self {
             require_tombstones: true,
             require_files: true,
+            log_buffer_size: 1
```
wjones127, line 85:
```suggestion
            log_buffer_size: num_cpus::get() * 4,
```

rust/benches/read_checkpoint.rs
```diff
@@ -13,6 +13,7 @@ fn read_null_partitions_checkpoint(c: &mut Criterion) {
     let config = DeltaTableConfig {
         require_tombstones: true,
         require_files: true,
+        log_buffer_size: 1
```
wjones127, line 16:
```suggestion
        log_buffer_size: num_cpus::get() * 4,
```

rust/src/builder.rs
```diff
@@ -67,13 +67,22 @@ pub struct DeltaTableConfig {
     /// Some append-only applications might have no need of tracking any files.
     /// Hence, DeltaTable will be loaded with significant memory reduction.
     pub require_files: bool,
+    /// Controls how many files to buffer from the commit log when updating the table.
+    /// The maximum possible value is 50.
+    /// This defaults to 1
+    ///
+    /// Setting a value greater than 1 results in concurrent calls to the storage api.
+    /// This can decrease latency if there are many files in the log since the 
+    /// last checkpoint.
```
wjones127, line 76: It's important to note the tradeoff, otherwise it sounds like we should always just choose the max `log_buffer_size`. Often in these instances, it's a tradeoff between speed and memory usage. Though if you think there's another tradeoff worth mentioning, feel free to add that.

```suggestion
    /// Setting a value greater than 1 results in concurrent calls to the storage api.
    /// This can decrease latency if there are many files in the log since the 
    /// last checkpoint, but will also increase memory usage.
```

rust/src/delta.rs
```diff
@@ -540,14 +556,49 @@ impl DeltaTable {
...
+        
+        while let Some((new_version, actions)) = {
+            let next_commit = log_buffer.next().await;
+            match next_commit {
+                Some((v, Ok(x))) => Ok(Some((v, self.get_actions(v,  x.bytes().await?).await?))),
+                Some((_, Err(ObjectStoreError::NotFound { .. }))) => Ok(None), // no more files in the log
+                Some((_, Err(err))) => Err(DeltaTableError::GenericError { source: Box::new(err) }),
+                None => Ok(None) // reached end of stream at max_version
+            }
+        }?
```
wjones127, line 595: It's a little hard to grok when there are so many lines on the RHS of a `while let`, that also ends with a `?`. Could we split out some of that logic above? Perhaps something like:

```suggestion
        let log_buffer = log_buffer
            .and_then(|(version, maybe_bytes)| async {
                let bytes = maybe_bytes?.await?;
                let actions = self.get_actions(version, bytes).await?;
                Ok((version, actions))
            });
        while let Some((new_version, actions)) = log_buffer.next().await? {
```

Also, shouldn't `NotFound` still be an error at this point since we previously verified the log file does exist?

wjones127 (changes requested):
Thanks for working on this.

rust/src/builder.rs
eeroel: Yep, this was relevant in the earlier version where max_version was not known, because in that case you would have 49 wasted requests if there was only one file. But I removed that now.

rust/src/builder.rs
eeroel: Agreed and changed this as suggested. The only thing that crossed my mind was if the underlying storage system had some concurrency limits, but I don't think that's a problem at least with the major cloud providers?

rust/src/delta.rs
eeroel: Yeah that's a good point about the while let. I think I tried to do this earlier, but for some reason I can't get the `and_then` method to work here even after importing `TryStreamExt`, I get complaints about not satisfying trait bounds.

The NotFound is an Ok case here if the max_version passed to the function is larger than the actual max_version, that's how it works currently on `main`. Not sure if that behavior is relied upon anywhere?

rust/src/builder.rs
eeroel: Hmm, then I have some trouble with the `mod fs_common` where I put the SlowStore implementation. And the tests are not really testing builder per se. Would it make sense to just make a copy of this function in `fs_common` for the purposes of the test?

rust/tests/data/simple_table_with_no_checkpoint/.part-00000-136c36f5-639d-4e95-bb0f-15cde3fb14eb-c000.snappy.parquet.crc
eeroel: Thanks, I didn't realize this. The tests now use `create_table`

rust/src/builder.rs
wjones127: That's true, rate limits also apply. Feel free to add a comment about that too if you'd like.

rust/src/delta.rs
wjones127: Ah I forgot `max_version` is passed in. That part's fine then.

wjones127: This is a bit tricky, but this is what I'd propose (tested locally):

```rust
        let mut log_stream = futures::stream::iter(self.version() + 1..max_version + 1)
            .map(|version| {
                let store = store.clone();
                let loc = commit_uri_from_version(version);
                async move {
                    let data = store.get(&loc).await?.bytes().await?;
                    let actions = Self::get_actions(version, data).await?;
                    Ok((version, actions))
                }
            })
            .buffered(buf_size);

        while let Some(res) = log_stream.next().await {
            let (new_version, actions) = match res {
                Ok((version, actions)) => (version, actions),
                Err(DeltaTableError::ObjectStore {
                    source: ObjectStoreError::NotFound { .. },
                }) => break, // no more files in the log
                Err(err) => return Err(DeltaTableError::from(err)),
            };
```

As part of this, need to change `get_actions` to not take `&self`. I don't know why it does, given it does use `self` at all.

eeroel: Nice, updated the code to this. The `&self` was just an oversight from my part when factoring out the function.

rust/src/builder.rs
eeroel: Added something now: https://github.com/delta-io/delta-rs/pull/1549/files#diff-f81b8b91e9721dc0a3235b6674a57c0700153b92b59f3988d357950a9c8c4760R240
Also updated the docs on the rust side to match the current version

python/deltalake/table.py
```diff
@@ -234,13 +235,20 @@ def __init__(
         :param without_files: If True, will load table without tracking files.
                               Some append-only applications might have no need of tracking any files. So, the
                               DeltaTable will be loaded with a significant memory reduction.
+        :param log_buffer_size: Number of files to buffer when reading the commit log. A positive integer.
```
wjones127, line 238: nit: could you also mention what the default is?

wjones127 (commented):
This is looking good. Just one nit on the docs and need some fixes for CI. Then I'm happy to merge :)

eeroel:
> This is looking good. Just one nit on the docs and need some fixes for CI. Then I'm happy to merge :)

Thanks! Updated the docs and fixed formatting + linter issues

rust/src/builder.rs
```diff
@@ -153,6 +171,17 @@ impl DeltaTableBuilder {
         self
     }
 
+    /// Sets `log_buffer_size` to the builder
+    pub fn with_buffer(mut self, log_buffer_size: usize) -> DeltaResult<Self> {
```
cmackenzie1, line 175:
```suggestion
    pub fn with_log_buffer_size(mut self, log_buffer_size: usize) -> DeltaResult<Self> {
```

eeroel: 👍 I think this is better as it's more explicit

wjones127 (approved):
Looks good. That last failure will be fixed once #1615 is merged.

eeroel:
Rebased on main

PR #1552, opened 2023-07-21
chore: address some integration test bloat of disk usage for development

### LINKED ISSUE #1550: Excessive integration test sizes causing builds to fail
# Environment 

**Delta-rs version**: v0.13.1

**Binding**: rust

***
# Bug

**What happened**:

When running `cargo test -p deltalake --no-default-features --features integration_test,s3-native-tls,datafusion` an excessive amount of disk space is consumed  When running with `RUSTFLAGS="-C debuginfo=1" CARGO_INCREMENTAL=0` as CI does, the local `target/` directory generates about 3.3GB of junk for me, with each binary linked for the `tests/` coming in between 400-500MB.

`cargo` not to shred my disks that much.

**How to reproduce it**:

Run zee tests

**More details**:

:grimacing:

rtyler:
# Description

This change attempts to further reduce our build/link time by changing `RUSTFLAGS` for CI and consolidating some integration tests into singular files, while moving more unit-style tests (those which just rely on `tests/data/`) into unit tests.

The previous use of "debuginfo=1" would still result in significant file sizes. When building integration tests locally a setting of "1" leads to a total target/ directory size of 3.1GB, whereas "line-tables-only" results in a total clean build target/ directory size of 2.1GB.

The "line-tables-only" feature provides:

    Generates the minimal amount of debug info for backtraces with filename/line
    number info, but not anything else, i.e. no variable or function parameter info.

Setting this to "0" does reduce the size further, but for panics and stacktraces in CI, we need some level of information to be present in order for the backtraces to be useful.

# Related Issue(s)

Closes #1550 

# Documentation 

The `line-tables-only` value for `debuginfo` was introduced in [Rust 1.70](https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#named-levels-of-debug-information)


omkar-foss:
Thanks for this PR @rtyler. As it has optimized not just on disk usage, but also significantly reduced RAM consumption and total test runtime. Peace for local development especially for newcomers to this project (like myself).

**My test run comparisons as below for reference:**
Test Run Instance: Ubuntu 22.04, 8GB RAM, 4 vCPUs, SSD
- Current `main` (HEAD 1bb6c3da) with clean `target`, updated flags in this PR:
~8GB RAM consumption (lots of swapping), ~15 mins runtime, ~19GB size of `target` after completion.
- This PR (HEAD 20da846f) with clean `target`, old flags:
~6.5 RAM consumption (reduced swapping), ~8 mins runtime, ~6GB size of `target` after completion.

**Below is the new test command that I've constructed from this PR's CI yaml, for reference:**
```bash
RUSTFLAGS="-C debuginfo=line-tables-only" CARGO_INCREMENTAL=0 RUST_BACKTRACE=1 CARGO_NET_GIT_FETCH_WITH_CLI="true" cargo test -p deltalake --no-default-features --features integration_test,s3-native-tls,datafusion
```

wjones127 (approved):
Thank you for doing this. LGTM!

PR #1565, opened 2023-07-27
ci: install newer rust for macos python release

wjones127:
The bleeding edge features we are using caused the macos Python release to fail, because the GH runners are still on 1.70. https://github.com/delta-io/delta-rs/actions/runs/5676317320/job/15382866806

The Windows runner already has 1.71 and the Linux are fresh installing in a manylinux container already, so it's just MacOS that's the problem.

### DIFF
--- a/.github/workflows/python_release.yml
+++ b/.github/workflows/python_release.yml
@@ -39,6 +39,14 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      # We use extra recent Cargo.toml syntax, so we need at least Rust 1.71.0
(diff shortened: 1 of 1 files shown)


wjones127:
FYI @rtyler it was 1.71 not 1.70 that introduced the "line-tables-only" value.

> Note that these named options are not yet available to be used via Cargo.toml. Support for that will be available in the next release 1.71.

https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#named-levels-of-debug-information

rtyler (approved):

PR #1575, opened 2023-07-31
chore: move deps to `[workspace.dependencies]`

### LINKED ISSUE #1572: Define common dependencies in Cargo Workspace
# Description

By defining a set of common dependecies in the root `Cargo.toml`, updates to those crates versions only need to be applied in one location instead of in `rust/` and `python/` workspaces. 

See https://github.com/delta-io/delta-rs/pull/1571 as an example of something that would benefit from using `[workspace.dependencies]`

**Use Case**

**Related Issue(s)**

cmackenzie1:
# Description

This commit moves a common set of dependencies from `rust/` and `python/` Cargo.toml files in favor of using the `workspace = true` variation.

The goal of this commit if merely organization. Pruning features, bumping versions are not goals of this commit.

# Related Issue(s)

- Closes #1572
# Documentation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,5 @@
 [workspace]
-members = [
-    "rust",
-    "python",
-]
+members = ["rust", "python"]
 exclude = ["proofs", "delta-inspect"]
 
 [profile.release-with-debug]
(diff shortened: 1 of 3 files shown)


roeap (approved):
👍

rtyler (approved):

Issue #1094, opened 2023-01-24
Support deletion vector


wjones127:
# Description

For protocol version 3, will want to support deletion vector. 

* [ ] Supporting reading with deletion vector
* [ ] Support delete operations using delete vector

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#deletion-vectors

Question: how do we decide to rewrite vs use delete vector?

**Use Case**

This enables much faster deletes.

**Related Issue(s)**

Prerequisites:
 * #930 
 * #832

houqp:
> Question: how do we decide to rewrite vs use delete vector?

This looks like a tradeoff between faster read performance v.s. faster write that need to be decided case by case? If so, might be better to just let the user decide depending on the expected workload pattern.

guyrt:
+1 to supporting user-owned tradeoff decision. I'm investigating this feature internally and update patterns in individual tables likely dictate the right decision.

For instance, in many dimension tables, edits may be spread randomly through existing data and merge on read will be more efficient. For fact tables with mostly append pattern (but occasional fact updates), judicious partition plus copy on write may be superior.

aersam:
Don't know if this helps, just tried to read a deletion vector file, and this seems to be working with the roaring crate:

```rust
use roaring::RoaringTreemap;
use std::io::Take;
use std::{fs::File, io::Read};

fn main() {
    let mut file =
        File::open("faker/deletion_vector_b3dfb787-c930-49c1-b78d-8333173f4db4.bin").unwrap();
    let mut buf = vec![0; 2];
    file.read(&mut buf).unwrap();
    let version = u16::from_le_bytes(buf.clone().try_into().unwrap());
    assert_eq!(version, 1);
    loop {
        let mut buf = vec![0; 3];
        let nrread = file.read(&mut buf).unwrap();
        if nrread == 0 {
            break;
        }

        let size_buf = [&[0], &buf[0..3]].concat();
        println!("size: {:?}", size_buf);
        let datasize = u32::from_be_bytes(size_buf.try_into().unwrap());
        println!("datasize: {}", datasize);
        if datasize == 0 {
            break;
        }
        let mut buf = vec![0; 4];
        file.read(&mut buf).unwrap();
        let magic = i32::from_le_bytes(buf.try_into().unwrap());
        assert!(magic == 1681511377);
        let take: Take<&File> = (&file).take(datasize as u64);
        let rdr = RoaringTreemap::deserialize_from(take).unwrap();
[...]

PR #1583, opened 2023-08-03
feat: add metadata for deletion vectors

aersam:
# Description
This just adds the deletion vector metadata to the actions. It does not interpret those yet, reading / writing deletion vectors is not supported with this. Still it enables use cases where you use delta-rs just for metadata retrieval

I have to add that I'm still learning rust and I expect this to take some iterations until code quality is sufficient

# Related Issue(s)
Part of #1094 : Adds the required metadata

# Documentation

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#deletion-vectors


aersam:
As a next step, one could add the absolute path of the deletion vector to the get_actions method. This requires a dependency on z85

aersam:
I'll update the PR in the next days to address the checks

aersam:
> I'll update the PR in the next days to address the checks

Should be ok now

wjones127 (changes requested):
Hi @aersam! Thanks for making this PR.

I think you need to run format, and then I have one requested change. Otherwise looks pretty good.

aersam:
> Hi @aersam! Thanks for making this PR.
> 
> I think you need to run format, and then I have one requested change. Otherwise looks pretty good.

That sounds good! I tried making the change in an idiomatic way and I hope I did nothing too bad :) Though I'm new to rust it feels a bit like "Programming done right", but I'm sure I still have to learn a lot there

wjones127 (approved):

wjones127:
@aersam Sorry for the delay in re-reviewing. Looks good. Just needs to be rebased and pass CI to merge :)

aersam:
> @aersam Sorry for the delay in re-reviewing. Looks good. Just needs to be rebased and pass CI to merge :)

No problem, thanks for re-reviewing! I clicked that "Sync" Button :) , should be rebased now

PR #1592, opened 2023-08-16
feat: use url parsing from object store

### LINKED ISSUE #1356: add support for atomic rename in R2
add support for atomic rename in R2

roeap:
# Description

This PR moves the url parsing and options to the object store implementation where applicable. This also provides a fix for a subtle bug we had in our configuration handling. Specifically in certain scenarios we may choose an unexpected credential. If the user passed in configuration for a credential, but there is another credential defined in the environment, the choice of credential depends on the order credentials are checked within object store.

While not completely robust, we now try to create a credential without info added from the environment, and if that fails, augment the options with specs from the environment.

closes #1356


rust/src/storage/config.rs
```diff
@@ -31,6 +30,65 @@ use object_store::gcp::{GoogleCloudStorageBuilder, GoogleConfigKey};
...
+            ("s3" | "s3a", Some(_)) => (Self::AmazonS3, url.path()),
+            ("gs", Some(_)) => (Self::GoogleCloudStorage, url.path()),
+            ("az" | "adl" | "azure" | "abfs" | "abfss", Some(_)) => {
+                (Self::MicrosoftAzure, url.path())
+            }
+            ("http", Some(_)) => (Self::Http, url.path()),
+            ("hdfs", Some(_)) => (Self::Hdfs, url.path()),
+            ("https", Some(host)) => {
+                if host.ends_with("dfs.core.windows.net") || host.ends_with("blob.core.windows.net")
+                {
```
vipeller, line 71: this does not seem to have the latest object store changes for fabric support

roeap: will definately add that !

rtyler (approved):

PR #1600, opened 2023-08-25
fix: address clippy errors on the latest Rust 1.72

rtyler:
There is still some documentation to add before this comes out of draft, but this is breaking CI for other pull requests at the moment

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -219,8 +219,10 @@ def write_deltalake(
         # PyArrow added support for written_file.size in 9.0.0
         if PYARROW_MAJOR_VERSION >= 9:
             size = written_file.size
-        else:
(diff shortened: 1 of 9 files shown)


python/deltalake/writer.py
```diff
@@ -219,8 +219,10 @@ def visitor(written_file: Any) -> None:
         # PyArrow added support for written_file.size in 9.0.0
         if PYARROW_MAJOR_VERSION >= 9:
             size = written_file.size
-        else:
+        else if filesystem is not None:
```
wjones127, line 222:
```suggestion
        elif filesystem is not None:
```

wjones127:
FYI PyArrow 13.0.0 will cause some test failures. I'm working on fixing them in another PR.

PR #1607, opened 2023-08-29
fix: stop using deprecated from_utc method

jacksonrnewhouse:
# Description
HEAD is currently failing `cargo check` because a new deprecation was just pushed to chrono, deprecating from_utc: https://github.com/chronotope/chrono/pull/1175. Since we fail on warnings, this breaks the build.

Might be worth relaxing some of these checks, as typically marking something as deprecated is done specifically to _not_ break downstream projects.


wjones127:
@jacksonrnewhouse 
> Might be worth relaxing some of these checks, as typically marking something as deprecated is done specifically to not break downstream projects.

Do you know how to do that? I looked around the clippy docs and google but couldn't find an obvious solution.

Issue #1569, opened 2023-07-28
Very slow read from a super massive table in Azure Data Lake Gen2


WaterKnight1998:
First of all thanks for creating this awesome library and eliminating the need to use spark for everything related to deltalake.

I have been trying to read a table from Azure Data Lake gen 2 and it is super slow.

The table hast lots of versions but when instantiating DeltaTable I specify the latest one

```python
dt=DeltaTable("adl://container1/table1/", version=33, storage_options=storage_options)
```

This above code is almost fast 6s, but when I run `to_pyarrow_table` and `to_pyarrow_dataset` to use the data later in DuckDB it is taking super long. Is it possible to increase the number of threads it uses for reading data?

Any other possible solution?

Thanks in advance!

roeap:
We have some ideas to improve read performance using our internal file system implementaions (which are in use here).

In case you have some predicates that can be applied to you data, you may be able to avoid reading some of the data into memory, by using `to_pyarrow_dataset` and subsequently the `scanner(...)` methods. I "think" DuckDB can also do predicate pushdown for pyarrow types, but I am not sure about it. In case that does not happen, or happen only partially, applying some predicated to the scanner should give you some performance boost.

Yacobolo:
We have also experienced really slow reads with ADLS gen 2 and a 70 mb file taking 12 minutes.

mjducutoae:
Hi there! 

We found a faster approach for this one if your goal is only read the current data of the delta table. 
Steps are:

1. Create a DeltaTable instance. 
``` python 
dt = DeltaTable("adl://container1/table1/", storage_options=storage_options)
``` 

2. After this, the goal is to grab the path of the current parquet files used by the latest version of the delta table: 
```python
files = dt.file_uris()
```
3. Create a pyarrow dataset using these files and read into a pyarrow table. You also need to pass the fs used: 
```python 
import pyarrow.parquet as pq
import pyarrow as pa

pq_data = pq.ParquetDataset(path_or_paths= files, filesystem= fs)
table= pq_data.read()
```

Special thanks to @Yacobolo for this one!

wjones127:
I think this is likely due to some GIL-locking in the filesystem implementation.

If you pass a PyArrow filesystem down instead, does it perform as expected?

```python
dt = DeltaTable("adl://container1/table1/")
dt.to_pyarrow_dataset(filesystem=fs)
```

ion-elgreco:
[...]

PR #1615, opened 2023-09-05
perf: avoid holding GIL in DeltaFileSystemHandler

wjones127:
# Description

I'm not sure yet this is the root cause of the performance issues we are seeing in #1569, but this seems like an obvious change that might help.

Any method that takes `&mut self` implicitly holds the GIL, so we'll need to allow threads explicitly in those methods.

# Related Issue(s)

# Documentation

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -66,6 +66,8 @@ jobs:
 
       - name: Build and install deltalake
         run: |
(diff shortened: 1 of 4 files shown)


rtyler (approved):
This makes sense to me but I cannot determine whether this Is A Good Way to solve this problem :laughing: 

ship it

PR #1613, opened 2023-09-03
fix: don't re-encode paths

wjones127:
# Description

In the delta log, paths are percent encoded. We decode them here:

https://github.com/delta-io/delta-rs/blob/787c13a63efa9ada96d303c10c093424215aaa80/rust/src/action/mod.rs#L435-L437

Which is good. But then we've been re-encoding them with `Path::from`. This PR changes to use `Path::parse` when possible instead. Instead of propagating errors, we just fallback to `Path::from` for now. Read more here: https://docs.rs/object_store/0.7.0/object_store/path/struct.Path.html#encode

# Related Issue(s)

* closes #1533
* closes #1446 
* closes #1079
* closes #1393
* closes #1228

# Documentation

PR #1618, opened 2023-09-07
fix: Parse paths before encoding them via a utility.

### LINKED ISSUE #1228: Error reading partitioned table with slash in partition key

mahinshaw:
# Description
The use of `Path::from` double encodes strings in some cases. This is seen when reading tables with a timestamp as the partition for instance. When written to the table the `:` is encoded as %3A. Calling Path::from on such a table double encodes the aready % encoded fields.

This PR merely adds a utility that first calls `parse` on the string, which validates that it has indeed been encoded correctly. In the case of failure, the string is then handed to Path::from to be handled.

# Related Issue(s)
- fixes #1228

# Documentation

[object_store::path::Path](https://docs.rs/object_store/latest/object_store/path/struct.Path.html)

### DIFF
--- a/python/src/filesystem.rs
+++ b/python/src/filesystem.rs
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use deltalake::storage::{DynObjectStore, ListResult, MultipartId, ObjectStoreError, Path};
+use deltalake::storage::{
+    get_path, DynObjectStore, ListResult, MultipartId, ObjectStoreError, Path,
+};
 use deltalake::DeltaTableBuilder;
 use pyo3::exceptions::{PyIOError, PyNotImplementedError, PyValueError};
 use pyo3::prelude::*;
@@ -57,8 +59,8 @@ impl DeltaFileSystemHandler {
     }
 
     fn copy_file(&self, src: String, dest: String) -> PyResult<()> {
-        let from_path = Path::from(src);
-        let to_path = Path::from(dest);
+        let from_path = get_path(src);
+        let to_path = get_path(dest);
         self.rt
             .block_on(self.inner.copy(&from_path, &to_path))
             .map_err(PythonError::from)?;
@@ -71,7 +73,7 @@ impl DeltaFileSystemHandler {
     }
 
     fn delete_dir(&self, path: String) -> PyResult<()> {
-        let path = Path::from(path);
+        let path = get_path(path);
         self.rt
             .block_on(delete_dir(self.inner.as_ref(), &path))
             .map_err(PythonError::from)?;
@@ -79,7 +81,7 @@ impl DeltaFileSystemHandler {
     }
 
     fn delete_file(&self, path: String) -> PyResult<()> {
-        let path = Path::from(path);
+        let path = get_path(path);
         self.rt
             .block_on(self.inner.delete(&path))
             .map_err(PythonError::from)?;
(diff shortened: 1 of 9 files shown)


rtyler:
Thanks for the pull request!

This seems similar to #1613 so I'm curious from @wjones127 how he wants to handle this change?

wjones127:
Hi @mahinshaw, thanks for the PR. This seems identical to the changes in #1613. Would you be willing to review and try out that PR instead?

mahinshaw:
Thanks both! @wjones127 I will pull it down and see if it solves my use case. I'll close if it covers my case.

mahinshaw:
I ran some tests with both branches and I don't see any difference in behavior for both reading data from existing delta schemas (written by spark to delta in azure) and written out locally with the same schema and re read again. There are some strange behaviors around creating a table via `DeltaOps::try_from_uri` with a `Path::from(./testing)`. But I think that's buried in `object_store`. Another curious thing Is whether or not the changes I made in `src/operations/{writer,restore}.rs` are important. However, I've yet to experience an issue that would lead me to believe the changes are necessary.

I'm curious what the plan is on getting the other PR merged, as it will unblock us on a number of use cases.

Thanks for the quick feedback! I'll close this PR in favor of #1613. Thanks!

wjones127:
> I'm curious what the plan is on getting the other PR merged, as it will unblock us on a number of use cases.
It will be merged as soon as it gets an approving review. We are unfortunately a little low on active reviewers at the moment.

PR #1620, opened 2023-09-08
fix: change map nullable value to false

### LINKED ISSUE #1619: Internal Parquet panic when using a Map type.
# Environment

**Delta-rs version**: 0.14

**Binding**: rust

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

**What happened**:

Writing JSON records with a `Map(String, String)` fails panics during parquet flush and close. 

```
thread 'local::test_issue_parquet_bit_util' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', /Users/cole/.cargo/registry/src/index.crates.io-6f17d22bba15001f/parquet-45.0.0/src/util/bit_util.rs:272:9
```

Not to panic and error returned if applicable. 

**How to reproduce it**:

```rust 
let fields: Vec<SchemaField> = vec![SchemaField::new(
    "metadata".to_string(),
    SchemaDataType::map(SchemaTypeMap::new(
        Box::new(SchemaDataType::primitive("string".to_string())),
        Box::new(SchemaDataType::primitive("string".to_string())),
        true,
    )),
    true,
    HashMap::new(),
)];
let schema = deltalake::Schema::new(fields);
let table = deltalake::DeltaTableBuilder::from_uri("./tests/data/map-null").build()?;
let _ = DeltaOps::from(table)
    .create()
    .with_columns(schema.get_fields().to_owned())
    .await?;

let mut table = deltalake::open_table("./tests/data/map-null").await?;

let mut writer = JsonWriter::for_table(&table).unwrap();
let _ = writer
    .write(vec![
        serde_json::json!({"metadata": {"hello": "world", "something": null}}),
    ])
    .await
    .unwrap();
writer.flush_and_commit(&mut table).await.unwrap();
```

**More details**:

This value is `true` https://github.com/delta-io/delta-rs/blob/main/rust/src/delta_arrow.rs#L170C21-L170C25 , where arrow defines it as `false` https://github.com/apache/arrow-rs/blob/master/arrow-schema/src/field.rs#L230. This is also described in https://github.com/apache/arrow-rs/issues/1697.

cmackenzie1:
# Description

This value was true but where arrow defines it as always false https://github.com/apache/arrow-rs/blob/master/arrow-schema/src/field.rs#L230.

This is also described in apache/arrow-rs#1697.

This also replaces `key_value` as the struct name with `entries` to remain consistent with https://github.com/apache/arrow-rs/blob/878217b9e330b4f1ed13e798a214ea11fbeb2bbb/arrow-schema/src/datatype.rs#L247-L250

# Related Issue(s)

- closes #1619 

# Documentation

### DIFF
--- a/rust/src/delta_arrow.rs
+++ b/rust/src/delta_arrow.rs
@@ -63,20 +63,17 @@ impl TryFrom<&schema::SchemaTypeMap> for ArrowField {
     type Error = ArrowError;
 
     fn try_from(a: &schema::SchemaTypeMap) -> Result<Self, ArrowError> {
-        Ok(ArrowField::new(
+        Ok(ArrowField::new_map(
+            "entires",
             "entries",
-            ArrowDataType::Struct(
-                vec![
-                    ArrowField::new("key", ArrowDataType::try_from(a.get_key_type())?, false),
-                    ArrowField::new(
-                        "value",
-                        ArrowDataType::try_from(a.get_value_type())?,
-                        a.get_value_contains_null(),
-                    ),
-                ]
-                .into(),
+            ArrowField::new("key", ArrowDataType::try_from(a.get_key_type())?, false),
+            ArrowField::new(
+                "value",
+                ArrowDataType::try_from(a.get_value_type())?,
+                a.get_value_contains_null(),
             ),
-            false, // always non-null
+            false,
+            false,
         ))
     }
 }
@@ -167,7 +164,7 @@ impl TryFrom<&schema::SchemaDataType> for ArrowDataType {
                         ]
                         .into(),
                     ),
-                    true,
+                    false,
                 )),
                 false,
             )),
@@ -305,7 +302,7 @@ macro_rules! arrow_map {
             stringify!($fieldname),
             ArrowDataType::Map(
                 Arc::new(ArrowField::new(
-                    "key_value",
+                    "entries",
                     ArrowDataType::Struct(
                         vec![
                             ArrowField::new("key", ArrowDataType::Utf8, false),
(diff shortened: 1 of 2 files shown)


rust/tests/integration_datafusion.rs
```diff
@@ -1027,6 +1078,8 @@ async fn simple_query(context: &IntegrationContext) -> TestResult {
 }
 
 mod date_partitions {
+    use deltalake::SchemaField;
```
cmackenzie1, line 1081: will remove.

rust/src/delta_arrow.rs
wjones127, line 1: This makes sense. Box entries and keys are required to be non-nullable by the Arrow spec :) 

https://github.com/apache/arrow/blob/c4b01c60fba85bbfc3a1b1510e179153c0f79515/format/Schema.fbs#L124

wjones127:
I think we are hitting an upstream bug in `MapArray::new_from_strings()`: https://github.com/apache/arrow-rs/issues/4807

cmackenzie1:
@wjones127 I reverted the change to `parquet2` - not sure what is up with that one. Also I yanked the `MapArray::new_from_strings` and applied the patch inline for now. We can clean that up once arrow is upgraded here.

rust/src/delta_arrow.rs
```diff
@@ -63,20 +63,17 @@ impl TryFrom<&schema::SchemaTypeMap> for ArrowField {
     type Error = ArrowError;
 
     fn try_from(a: &schema::SchemaTypeMap) -> Result<Self, ArrowError> {
-        Ok(ArrowField::new(
+        Ok(ArrowField::new_map(
+            "entires",
```
wjones127, line 67:
```suggestion
            "entries",
```

wjones127 (approved):

PR #1621, opened 2023-09-09
feat: allow multiple incremental commits in optimize

kvap:
Currently "optimize" executes the whole plan in one commit, which might fail. The larger the table, the more likely it is to fail and the more expensive the failure is.

Add an option in OptimizeBuilder that allows specifying a commit interval. If that is provided, the plan executor will periodically commit the accumulated actions.

### DIFF
--- a/rust/src/operations/optimize.rs
+++ b/rust/src/operations/optimize.rs
@@ -22,7 +22,7 @@
 
 use std::collections::HashMap;
 use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use arrow::datatypes::{Schema as ArrowSchema, SchemaRef as ArrowSchemaRef};
 use arrow_array::RecordBatch;
@@ -172,6 +172,7 @@ pub struct OptimizeBuilder<'a> {
     max_spill_size: usize,
     /// Optimize type
     optimize_type: OptimizeType,
+    min_commit_interval: Option<Duration>,
 }
 
 impl<'a> OptimizeBuilder<'a> {
@@ -188,6 +189,7 @@ impl<'a> OptimizeBuilder<'a> {
             max_concurrent_tasks: num_cpus::get(),
             max_spill_size: 20 * 1024 * 1024 * 2014, // 20 GB.
             optimize_type: OptimizeType::Compact,
+            min_commit_interval: None,
         }
     }
 
@@ -241,6 +243,12 @@ impl<'a> OptimizeBuilder<'a> {
         self.max_spill_size = max_spill_size;
         self
     }
+
+    /// Min commit interval
+    pub fn with_min_commit_interval(mut self, min_commit_interval: Duration) -> Self {
+        self.min_commit_interval = Some(min_commit_interval);
+        self
+    }
 }
 
 impl<'a> std::future::IntoFuture for OptimizeBuilder<'a> {
@@ -263,10 +271,16 @@ impl<'a> std::future::IntoFuture for OptimizeBuilder<'a> {
                 this.filters,
                 this.target_size.to_owned(),
                 writer_properties,
-                this.max_concurrent_tasks,
-                this.max_spill_size,
             )?;
-            let metrics = plan.execute(this.store.clone(), &this.snapshot).await?;
+            let metrics = plan
+                .execute(
+                    this.store.clone(),
+                    &this.snapshot,
+                    this.max_concurrent_tasks,
+                    this.max_spill_size,
+                    this.min_commit_interval,
+                )
+                .await?;
             let mut table = DeltaTable::new_with_state(this.store, this.snapshot);
             table.update().await?;
             Ok((table, metrics))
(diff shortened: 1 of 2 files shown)


rust/src/operations/optimize.rs
```diff
@@ -641,61 +651,86 @@ impl MergePlan {
...
-                &actions,
-                self.task_parameters.input_parameters.clone().into(),
-                snapshot,
-                Some(metadata),
-            )
-            .await?;
+            let now = Instant::now();
+            if !actions.is_empty() && (self.min_commit_interval.map_or(false, |i| now.duration_since(last_commit) > i) || end) {
+                let actions = std::mem::take(&mut actions);
+                last_commit = now;
```
cmackenzie1, line 695: Does it matter if this is updated before the commit success?

kvap: I don't think it matters much, because `min_commit_interval` is supposed to be much larger than the time it takes to do a commit.

kvap:
`clippy` is complaining that `create_merge_plan` has too many arguments. After looking at that function, and at the `MergePlan` struct, I noticed that they take and store some things which, rather than describing the plan itself, describe how to execute it: `max_concurrent_tasks`, `max_spill_size`, and `min_commit_interval`.

Should we move those out of `MergePlan` and `create_merge_plan`, and add them as arguments to `MergePlan.execute()`?

rust/src/operations/optimize.rs
```diff
@@ -641,61 +650,90 @@ impl MergePlan {
...
-            let maybe_map_metrics = serde_json::to_value(metrics.clone());
-            if let Ok(map) = maybe_map_metrics {
-                metadata.insert("operationMetrics".to_owned(), map);
+        let mut table = DeltaTable::new_with_state(object_store.clone(), snapshot.clone());
+
+        let mut last_commit = Instant::now();
+        loop {
+            let next = stream.next().await.transpose()?;
+
+            let end = next.is_none();
```
wjones127, line 682: Why aren't we just using `while let Some(next) = stream.next()`?

wjones127:
> Should we move those out of MergePlan and create_merge_plan, and add them as arguments to MergePlan.execute()?

Yeah that sounds like a good idea 👍

rust/src/operations/optimize.rs
kvap: Didn't want to repeat the code to do the final commit when next() returns None.

rust/src/operations/optimize.rs
```diff
@@ -641,61 +650,90 @@ impl MergePlan {
...
-            let mut metadata = Map::new();
-            metadata.insert("readVersion".to_owned(), self.read_table_version.into());
-            let maybe_map_metrics = serde_json::to_value(metrics.clone());
-            if let Ok(map) = maybe_map_metrics {
-                metadata.insert("operationMetrics".to_owned(), map);
+        let mut table = DeltaTable::new_with_state(object_store.clone(), snapshot.clone());
+
+        let mut last_commit = Instant::now();
+        loop {
+            let next = stream.next().await.transpose()?;
```
wjones127, line 680: nit: For clarity, let's initialize the variables we are collecting right above the loop where we are collecting them.

```suggestion
        // Actions buffered so far. These will be flushed either at the end
        // or when we reach the commit interval.
        let mut buffered_actions = vec![];
        
        // Each time we commit, we'll reset buffered_metrics to orig_metrics.
        let orig_metrics = std::mem::take(&mut self.metrics);
        let mut buffered_metrics = orig_metrics.clone();
        let mut total_metrics = orig_metrics.clone();
        
        let mut last_commit = Instant::now();
        loop {
            let next = stream.next().await.transpose()?;
```

kvap: Applied this.

kvap:
> > Should we move those out of MergePlan and create_merge_plan, and add them as arguments to MergePlan.execute()?
> 
> Yeah that sounds like a good idea 👍

Done.

ion-elgreco:
@kvap is this option also exposed to the python bindings?

rust/src/operations/optimize.rs
```diff
@@ -576,54 +584,51 @@ impl MergePlan {
         mut self,
         object_store: ObjectStoreRef,
         snapshot: &DeltaTableState,
+        max_concurrent_tasks: usize,
+        #[allow(unused_variables)] // used behind a feature flag
+        max_spill_size: usize,
+        min_commit_interval: Option<Duration>,
```
wjones127, line 590: This is fine for now, but eventually we might want to put these execution settings in a struct.

rust/tests/command_optimize.rs
```diff
@@ -307,7 +305,10 @@ async fn test_conflict_for_remove_actions() -> Result<(), Box<dyn Error>> {
     )
     .await?;
 
-    let maybe_metrics = plan.execute(dt.object_store(), &dt.state).await;
+    let maybe_metrics = plan
+        .execute(dt.object_store(), &dt.state, 1, 20, None)
+        .await;
```
wjones127, line 310: Could you add at least one test where we pass in a commit interval? even if it doesn't make an intermediate commit, it would be good to know those code paths run through.

wjones127 (changes requested):
I'd like to see one more change to the tests before merging.

@ion-elgreco I've created a follow up issue for the Python bindings: https://github.com/delta-io/delta-rs/issues/1640

rust/tests/command_optimize.rs
kvap: Added a test where an intermediate commit is actually created.

wjones127 (approved):

PR #1630, opened 2023-09-15
feat: pass known file sizes to filesystem in Python

eeroel:
# Description
Currently the Filesystem implementation always makes a HEAD request when opening a file, to determine the file size. The proposed change is to read the file sizes from the delta log instead, and to pass them down to the `open_input_file` call, eliminating the HEAD request.

# Related Issue(s)

# Documentation

### DIFF
--- a/python/deltalake/fs.py
+++ b/python/deltalake/fs.py
@@ -12,47 +12,23 @@ class DeltaStorageHandler(DeltaFileSystemHandler, FileSystemHandler):
     DeltaStorageHandler is a concrete implementations of a PyArrow FileSystemHandler.
     """
 
-    known_sizes: Dict[str, int] = {}
-
-    def __new__(  # type:ignore
-        cls,
-        table_uri: str,
-        storage_options: Optional[Dict[str, str]] = None,
-        known_sizes: Optional[Dict[str, int]] = None,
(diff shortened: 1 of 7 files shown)


python/deltalake/fs.py
```diff
@@ -12,23 +12,47 @@ class DeltaStorageHandler(DeltaFileSystemHandler, FileSystemHandler):
...
+
+    def __init__(
+        self,
+        table_uri: str,
+        storage_options: Optional[Dict[str, str]] = None,
+        known_sizes: Optional[Dict[str, int]] = None,
+    ):
+        if known_sizes:
+            self.known_sizes = known_sizes
+        return
```
wjones127, line 35: Could we push the `known_sizes` mapping down into the Rust `DeltaFileSystemHandler`? Then we won't need to change the signatures of `open_input_file` or `open_input_stream`.

eeroel: Good point, fixed this 👍

eeroel:
Fixed formatting issue and rebased on main

eeroel:
Fixed remaining issue with type annotation

wjones127 (approved):
This looks good. I appreciate the tests :)

PR #1635, opened 2023-09-17
chore: relax chrono pin to 0.4

houqp:
# Description

relax chrono pin to improve downstream compatibility.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ serde_json = "1"
 
 # "stdlib"
 bytes = { version = "1" }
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This seems reasonable to me. cc @rtyler

rtyler:
@wjones127 unfortunately this does not accomplish the objective of relaxing our chrono dependency. The method `timestamp_nanos_opt()` did not exist before 0.4.31. The `0.4` version constraint allows for `Cargo.lock` files with 0.4.30 (etc) to still compile file without an update, and will then fail on our code.

I tried to be more explicit in #1641, but golly this chrono change was annoying.

PR #1638, opened 2023-09-17
fix: include in-progress row group when calculating in-memory buffer length

### LINKED ISSUE #1637: `PartitionWriter`'s `buffer_len` doesn't include incomplete row groups
# Environment

**Delta-rs version**: 0.15

**Binding**: Rust

***
# Bug

**What happened**:
`PartitionWriter`'s `buffer_len` function does not include the length of an unflushed row group in its returned length. This means for writers without a complete row group, `buffer_len` returns  `0`.  

`buffer_len` should return `completed row groups length + in progress row group length` to better capture the current memory use. 

**How to reproduce it**:
A test like this reproduces the issue (and currently fails):

```rust
let batch = get_record_batch(None, false);
let table = create_initialized_table(&vec![]).await;
let mut writer = RecordBatchWriter::for_table(&table).unwrap();

// Write the given batch. This is less than the configured row group size
writer.write(batch).await.unwrap();

// This returns 0, even though we have some data buffered in memory (the incomplete row group).
assert!(writer.buffer_len() > 0);
```

BnMcG:
# Description
`PartitionWriter.buffer_len()` is documented as returning: 

> the current byte length of the in memory buffer.

However, this doesn't currently include the length of the in-progress row group. This means that until a row group is flushed, `buffer_len()` returns `0`. Based on the documented description, its length should probably include the bytes currently in-memory as part of an unflushed row group. 

`buffered_record_batch_count` _does_ include in-progress row groups, so this change also means record count and buffered bytes are reported consistently.

# Related Issue(s)

- closes #1637

# Documentation

[`buffer_len` on `RecordBatchWriter`](https://docs.rs/deltalake/0.15.0/deltalake/writer/record_batch/struct.RecordBatchWriter.html#method.buffer_len)


wjones127 (commented):
Thanks @BnMcG this looks great!

wjones127 (approved):

PR #1639, opened 2023-09-17
refactor!: update operations to use delta scan

Blajda:
# Description
Recently implemented operations did not use `DeltaScan` it had some gaps. These gaps would make it harder switch towards logical plans which is required for merge.

Gaps:
- It was not possible to include file lineage in the result
- The subset of files to be scanned is known ahead of time. Users had to reconstruct a parquet scan based on those files

The PR introduces a `DeltaScanBuilder` that allow users to specify which files to use when constructing the scan, if the scan should be enhanced to include additional metadata columns, and allows a projection to be specified. It also retains previous functionality of pruning based on the provided filter when files to scan are not provided.

`DeltaScanConfig` is also introduced which allows users to deterministic obtain the names of any added metadata columns or allows them to specify the name if required.

The public interface for `find_files` has changed but functionality remains the same.

A new table provider was introduced which accepts an `DeltaScanConfig`. This is required for future merge enhancements so unmodified files can be pruned pruned prior to writes.


Blajda:
Can be merged after #1705

wjones127 (dismissed):

Blajda:
@wjones127 This is ready to be merged

rust/src/delta_datafusion/mod.rs
```diff
@@ -362,66 +364,299 @@ pub(crate) fn register_store(store: ObjectStoreRef, env: Arc<RuntimeEnv>) {
...
+            arrow_schema::DataType::Utf8,
+            true,
+        )));
+    }
+
+    Ok(Arc::new(ArrowSchema::new(fields)))
+}
+
+#[derive(Debug, Clone, Default)]
+/// Used to specify it additonal metadat columns are exposed to the user
```
roeap, line 389:
```suggestion
/// Used to specify if additonal metadat columns are exposed to the user
```

rust/src/delta_datafusion/mod.rs
```diff
@@ -362,66 +364,299 @@ pub(crate) fn register_store(store: ObjectStoreRef, env: Arc<RuntimeEnv>) {
...
+}
+
+impl DeltaScanConfigBuilder {
+    /// Construct a new instance of `DeltaScanConfigBuilder`
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Indicate that a column containing a records file path is included.
+    /// Column name is randomly generated and can be determined once this Config is built
```
roeap, line 407:
```suggestion
    /// Column name is generated and can be determined once this Config is built
```

roeap (commented):
Looking great so far! Running out of time for now, but will finish asap!

roeap (approved):
LGTM! operations are really coming along :).

PR #1564, opened 2023-07-26
fix: add support for microsoft onelake

vmuddassir-msft:

PR #1642, opened 2023-09-19
fix: introduce support for Microsoft OneLake

rtyler:
# Description

This is a rebase and update of #1564 which already has some robust discussion. 

# Related Issue(s)

Closes #1564

### DIFF
--- a/rust/src/storage/config.rs
+++ b/rust/src/storage/config.rs
@@ -124,6 +124,8 @@ pub(crate) fn configure_store(
                 try_configure_s3(url, options)
             } else if host.contains("dfs.core.windows.net")
                 || host.contains("blob.core.windows.net")
+                || host.contains("dfs.fabric.microsoft.com")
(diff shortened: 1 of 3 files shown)


wjones127 (approved):
Seems okay. Would ideally like the tests run in CI and use the full `test_object_store` suite.

rtyler:
@wjones127 Me too, but the only way I think we do that is if I set up an Azure account and plonk some secrets into CI like I have done in the past for S3.


I'll think about it :laughing:

vmuddassir-msft:
@rtyler - Thanks for getting this merged , since a dynamically generated jwt token is used for now it will might not be possible to store this Github secrets. Can you please let me know the release date for deltalake package

https://crates.io/crates/deltalake/versions - The last update seems to have been published on 2 weeks ago

rtyler:
@vmuddassir-msft please watch the repository's releases 
image

PR #1645, opened 2023-09-19
feat: expose min_commit_interval to `optimize.compact` and `optimize.z_order`

### LINKED ISSUE #1640: Expose Optimize option min_commit_interval in Python
# Description

Follow up for #1621.

**Use Case**

**Related Issue(s)**

ion-elgreco:
# Description
Exposes min_commit_interval in the Python API to `optimize.compact` and `optimize.z_order`. Added one test-case to verify the min_commit_interval.

# Related Issue(s)
closes #1640


python/tests/test_optimize.py
```diff
@@ -46,3 +46,21 @@ def test_z_order_optimize(
...
+    write_deltalake(tmp_path, sample_data, partition_by="utf8", mode="append")
+
+    dt = DeltaTable(tmp_path)
+    old_version = dt.version()
+
+    dt.optimize.z_order(["date32", "timestamp"], min_commit_interval=0)
+
+    last_action = dt.history(1)[0]
+    assert last_action["operation"] == "OPTIMIZE"
+    assert dt.version() == old_version + 5
```
wjones127, line 66: This line isn't particularly obvious, so worth adding a comment here:

```suggestion
    # The table has 5 distinct partitions, each of which are Z-ordered
    # independently. So with min_commit_interval=0, each will get its
    # own commit.
    assert dt.version() == old_version + 5
```

python/deltalake/table.py
```diff
@@ -691,6 +691,7 @@ def compact(
         partition_filters: Optional[FilterType] = None,
         target_size: Optional[int] = None,
         max_concurrent_tasks: Optional[int] = None,
+        min_commit_interval: Optional[int] = None,
```
wjones127, line 694: Not important, but it might be nice if this accepted a timedelta as well. That way users could write out `timedelta(minutes=5)` instead of `5 * 60`, which can be a little more cryptic.

wjones127 (changes requested):
Overall looks good. Would just like the test to have a descriptive comment.

python/deltalake/table.py
ion-elgreco: @wjones127 Done! You can pass timedelta now and it gets converted to seconds before being passed over to Rust

python/deltalake/table.py
```diff
@@ -738,14 +757,32 @@ def z_order(
...
+
+        Use a timedelta object to specify the seconds, minutes or hours of the interval.
+        >>> from deltalake import DeltaTable
+        >>> from datetime import timedelta
+        >>> dt = DeltaTable("tmp")
+        >>> time_delta = timedelta(minutes=10)
+        >>> dt.optimize.compact(min_commit_interval=time_delta)
         """
+        if isinstance(min_commit_interval, timedelta):
+            commit_interval = min_commit_interval.seconds
```
wjones127, line 775: I think this isn't quite right:

```python
>>> from datetime import timedelta
>>> timedelta(days=5).seconds
0
>>> timedelta(days=5).total_seconds()
432000.0
```

```suggestion
            commit_interval = min_commit_interval.total_seconds()
```

ion-elgreco: @wjones127 Strange, for minutes and seconds it gave the correct amount. Anyway I've changed it to that method, and fixed the issue where I was using | operator (my local python is always 3.10, so it became second nature to use that always). All checks should pass now.

wjones127 (dismissed):

wjones127 (approved):
Thanks @ion-elgreco!

carlos6moya:
Hello, I am conducting tests with Z-order optimization and the min_commit_interval. It seems that while optimize respects the min_commit_interval, the Z-order operation does not use this setting as expected.

Each Z-order operation is creating a new entry in the Delta log, even when min_commit_interval is set to a higher timedelta value. This is not happening with optimize.compact.

My expectation was that Z-order would batch commits based on min_commit_interval, similar to how optimize works, but this does not appear to be the case.

Environment:

Delta Lake version: 0.21.0

PR #1646, opened 2023-09-20
chore: proposed updated CODEOWNERS to allow better review notifications

### LINKED ISSUE #1553: Update the CODEOWNERS to capture the current reviewers and contributors
We have a few folks who were once very active but no longer spend much time with us :hugs: 

I believe we should remove the following from the `*` rule in `CODEOWNERS` to make our reviews cleaner and to spare their inboxes and notifications:

* @mosyp 
* @houqp 
* @xianwill 

I think once #1434 is completed, we can and should add some more folks as `CODEOWNERS` with more granularity :partying_face:

rtyler:
Based on current pull request feedback and maintenance trends I'm suggesting these rules to get the right people on the reviews by default.

Closes #1553


wjones127 (commented):
It looks like we might want to add @MrPowers as a maintainer if we want to add him to this list.

Otherwise looks good to me. Worth keeping this open for a few days to collect feedback though.

MrPowers:
Yea would love to be a maintainer to help out more!

roeap (approved):
Looking good - I think we need to give @MrPowers access to the repo, at least GH ist complaining :)

PR #1305, opened 2023-04-22
fix: read log from DLT logs

roeap:
# Description

This PR fixes an issue reading tables created by DLT, specifically, when the metadata action does not contain schema information.

This is a draft, since I want to first clarify if this is intended. In the protocol the `schemaString` property is marked as required.

# Related Issue(s)

- closes #1302

# Documentation

Issue #1302, opened 2023-04-20
Invalid JSON in log record missing field `schemaString` for DLT tables


abhilashsharma1992:
# Environment
Databricks DLT 

**Delta-rs version**: [rust-v0.8.0](https://github.com/delta-io/delta-rs/releases/tag/rust-v0.8.0)

**Binding**:  [python-v0.8.1](https://github.com/delta-io/delta-rs/releases/tag/python-v0.8.1)

**Environment**:
- **Cloud provider**: Azure
- **OS**: Ubuntu
- **Other**: Specifically observing for Delta Live Tables

***
# Bug
**What happened**:  Seeing an error while loading a ADLS delta table, please find the code snippet
    storage_options = {

        "AZURE_STORAGE_ACCOUNT_KEY": tenant_storage_access_key,

        "AZURE_STORAGE_ACCOUNT_NAME": tenant_storage_account_name

    }

    dt = DeltaTable(deltaPathName,storage_options=storage_options)

Exception Stack trace:

Traceback (most recent call last): File "<string>", line 1, in <module> File "/home/ankit/spc-data-analytics/src/lakehousechangedatacapture/.venv/lib/python3.8/site-packages/deltalake/table.py", line 122, in __init__ self._table = RawDeltaTable( deltalake.PyDeltaTableError: Invalid JSON in log record, version=0, line=`{"metaData":{"id":"ac0a0120-970e-4d8c-ae92-b5244b055d6e","format":{"provider":"parquet","options":{}},"partitionColumns":[],"configuration":{},"createdTime":1681798577757}}`, err=`missing field `schemaString` at line 1 column 171`

**What you expected to happen**: DeltaTable class should load the ADLS delta table, all other delta tables other than DLT tables are working as expected

**How to reproduce it**: Create a DLT table in ADLS and read using python binding(as shown in code above).

**More details**:
Summary: Azure Databricks DLT pipeline used to write to ADLS everyday, subsequently read from Delta-rs python binding which is failing to load.

roeap:
Personally I never worked with live tables, but it is surprising that we do not have a schema defined in the shown log record.

Could you check the log for version 1 to see if that contains a schema definition?

abhilashsharma1992:
Please find the first two log files for the delta table:
**1st:**
[...]

Issue #1357, opened 2023-05-11
Support merge (upsert) in Python


hongbo-miao:
# Description

**Use Case**

I have dataframe with a few columns `timestamp`, `current`, `voltage`, `temperature`.
I hope `timestamp` to be unique, like a primary key.

Currently below code will keep appending if I run same code on same dataframe.

```py
write_deltalake("s3a://my-bucket/my-delta-tables/motor", df, mode="append")
```

It would be great to support merge (upsert) like this https://docs.databricks.com/delta/merge.html
Thanks! 😃

**Related Issue(s)**

Originally asked at https://github.com/delta-io/delta-rs/discussions/1355

mohitbansal-gep:
Hi Team , any updates, of when can this be made available in python

ion-elgreco:
@wjones127 can you assign this to me? Would like to give this a try next few weekends :)

pfwnicks:
any updates on this? This would be an amazing feature to have out of the box for python. Any open issues blocking this still?

ion-elgreco:
> any updates on this? This would be an amazing feature to have out of the box for python. Any open issues blocking this still?

I started last week, I'm still getting the hang of rust so it will probably take a couple weeks

ion-elgreco:
> any updates on this? This would be an amazing feature to have out of the box for python. Any open issues blocking this still?

I'm working on it but I expect it will take a couple weeks. I only have time to work on it in weekends

mohitbansal-gep:
When this will be released to pypi?

ion-elgreco:
> When this will be released to pypi?

Somewhere this week

pfwnicks:
awesome job @ion-elgreco! Thanks for cranking this out!

ion-elgreco:
> @ion-elgreco Thanks for enabling this!
> Does this feature also enable schema merge? Something similar to this: https://docs.databricks.com/en/delta/update-schema.html#merge-schema-evolution

No only Merge operation.

mohitbansal-gep:
Is the usage updated in Docs?

ion-elgreco:
@mohitbansal-gep nope not yet. The API has examples for each command and you can find some more clear examples in the blog I wrote: https://delta.io/blog/2023-10-22-delta-rs-python-v0.12.0/

PR #1647, opened 2023-09-20
fix: compensate for invalid log files created by Delta Live Tables

rtyler:
It would appear that in some cases Delta Live Tables will create a Delta table which does not adhere to the Delta Table protocol.

The metaData action as a **required** `schemaString` property which simply doesn't exist. Since it appears that this only exists at version zero of the transaction log, and the _actual_ schema exists in the following versions of the table (e.g. 1), this change introduces a default deserializer on the MetaData action which provides a simple empty schema.

This is an alternative implementation to #1305 which is a bit more invasive and makes our schema_string struct member `Option<String>` which I do not believe is worth it for this unfortunate compatibility issue

Closes #1305, #1302


wjones127:
Removing mention of #1357 from PR description, since I think tagging that was accidental.

wjones127 (approved):

Issue #1215, opened 2023-03-10
Handle illegal charactes for Windows paths


Blajda:
# Environment

**Delta-rs version**: Rust
**Binding**: main

**Environment**:
- **OS**: Windows

***
# Bug

**What happened**:
When delta-rs tries to write a timestamp partition on Windows it fails with the following error.

```
test timestamp
thread 'test_files_scanned' panicked at 'called `Result::unwrap()` on an `Err` value: ObjectStore { source: Generic { store: "LocalFileSystem", source: UnableToCreateFile { path: "C:\\Users\\runneradmin\\AppData\\Local\\Temp\\.tmpKm0Y2T\\timestamp=1970-01-01 00:00:00\\part-00000-537f4856-39e3-4058-895a-a726fcf769ae-c000.snappy.parquet", err: Os { code: 123, kind: InvalidFilename, message: "The filename, directory name, or volume label syntax is incorrect." } } } }', rust\tests\datafusion_test.rs:89:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This occurs since the following characters are illegal for a path in windows. < > ? [ ] : | *
This likely is also an issue with the string column too

**What you expected to happen**:
Writing timestamp columns on Windows should work. We also need to handle any other illegal value in the path.
The solution should conform to Spark's behavior and end users should only use the partition value from the Delta log

PR #1661, opened 2023-09-24
fix: more consistent handling of partition values and file paths

### LINKED ISSUE #1111: Python: support partitioning a table on a timestamp column
# Description

**Use Case**

Would like to be able to partition by a timestamp column, but it seems like we might not be serializing them right.

**Related Issue(s)**

### LINKED ISSUE #1299: Special characters in partition path not handled locally
# Environment

**Delta-rs version**: 0.8.1

**Binding**: 

**Environment**:
- **OS**: Ubuntu 22.04.2 LTS 
- **Python**: 3.10.6

***
# Bug

**What happened**:
Parquet file were not found 

I expected `to_pandas` to load the parquet file

**How to reproduce it**:

```python
from pandas import DataFrame
df = DataFrame(
    [
        ["Pierre", "Python", 24, "R&D"], # special character: &
        ["David", "Python", 33, "R&D"],
        ["Cyril", "Typescript", 26, "R&D"],
        ["Marie", "Excel", 36, "Commerce"],
    ],
    columns=["prenom", "skill", "age", "department"],
)
write_deltalake("./test/tables/garbage.delta", df, partition_by=["department"])
dt = DeltaTable("./test/tables/garbage.delta")

dt.to_pandas()
```

**More details**:

```
  File "/home/zar3bski/Documents/Code/octaave/deltastic/test/minimally_reproductible.py", line 18, in <module>
    dt.to_pandas()
  File "/home/zar3bski/.cache/pypoetry/virtualenvs/deltastic-GlE5VuQW-py3.10/lib/python3.10/site-packages/deltalake/table.py", line 418, in to_pandas
    return self.to_pyarrow_table(
  File "/home/zar3bski/.cache/pypoetry/virtualenvs/deltastic-GlE5VuQW-py3.10/lib/python3.10/site-packages/deltalake/table.py", line 400, in to_pyarrow_table
  File "/home/zar3bski/.cache/pypoetry/virtualenvs/deltastic-GlE5VuQW-py3.10/lib/python3.10/site-packages/deltalake/fs.py", line 22, in open_input_file
deltalake.PyDeltaTableError: Object at location /home/zar3bski/Documents/Code/octaave/deltastic/test/tables/garbage.delta/department=R&D/0-0294291a-0d31-410b-8b04-115377a6f9a2-0.parquet not found: No such file or directory (os error 2)
terminate called recursively
terminate called without an active exception
[1]    189090 IOT instruction (core dumped)  poetry run python test/minimally_reproductible.py
```
When I look in my project files, I find the file in `test/tables/garbage.delta/department=R%2526D/0-0294291a-0d31-410b-8b04-115377a6f9a2-0.parquet` There seems to be a problem with the URL encoding of **&** that should not be handled as %2526` in a local context`

### LINKED ISSUE #1591: partition_by with "$" on local filesystem
# Environment

**Delta-rs version**: 0.10.1

**Binding**:  Python

**Environment**: 
- **Cloud provider**: n/a
- **OS**: Linux/Ubuntu/WSL
- **Other**: local filesystem (ext4)

***
# Bug

**What happened**: 
error in reading DeltaTable (deltalake.table.DeltaTable) - FileNotFoundError

Load of DeltaTable

**How to reproduce it**:
Create a DeltaLake on local filesystem which contains a partition with "$" - deltalake.writer.write_deltalake(path, df, name = name, partition_by = partition). Try to read it dt = deltalake.table.DeltaTable(path) and access the columns df = dt.to_pandas(columns=cols).

**More details**:
When the DeltaLake is written with a partition_by the "$" char is replaced with "%2524". On read the $ is kept, a file not found error is thrown. In my case:
On write where partition is "Exchange_Code" which contains a value of "ID$", the directory created is "Exchange_Code=ID%2524" which contains a file 10-38fbb976-1e37-4fe1-b008-31194f837d9c-0.parquet.
On read the error is: "FileNotFoundError: Object at location path/Exchange_Code=ID$/10-38fbb976-1e37-4fe1-b008-31194f837d9c-0.parquet not found: No such file or directory (os error 2)"

The parquet file itself is accessible from OS level.

### LINKED ISSUE #1651: Unable to interface with data written from Spark Databricks
# Environment

**Delta-rs version**: 0.10.2

**Binding**: python / rust

**Environment**:
* Azure
* Locally
* Spark Databricks + Azure

***
# Bug

**What happened**:
When attempting to interface with Databricks we're getting inconsistent results for the encoding of the partition resulting in inability to interface across clients.

After the fix from https://github.com/delta-io/delta-rs/pull/1613 we're getting closer but still not consistent with Databricks.

### Python
When writing from python the partitions are formatted as:
```partition_date=2023-09-15%2000%3A00%3A00.000000```

### Rust
When writing from rust we see it as:
```partition_date=2023-09-15 00:00:00```

### Databricks
When writing from Spark Databricks to Delta Lake we see partial encoding:
```partition_date=2023-09-15 00%3A00$3A00```

Would expect to have consistent encoding across platforms.

**How to reproduce it**:

Write to Azure using Databricks, see partition layout.

Sample Python run locally:

```python
import datetime
import pyarrow as pa

data = pa.table({"data": pa.array(["mydata"]),
                 "inserted_at": pa.array([datetime.datetime.now()]),
                 "partition_column": pa.array([datetime.datetime(2023, 9, 15)])})



write_deltalake(table_or_uri="./unqueryable_table2", \
  mode="append", \
  data=data, \
  partition_by=["partition_column"]
)
```

### LINKED ISSUE #1653: [python] Different stringification of partition values in reader and writer
# Environment

**Delta-rs version**: python v0.10.2

***
# Bug

The delta reader and writer treat differently partition values which are not strings. You can clearly see it from:
https://github.com/delta-io/delta-rs/blob/a74589be7c39315360925049c716d1d70b906970/python/deltalake/table.py#L614-L627

https://github.com/delta-io/delta-rs/blob/a74589be7c39315360925049c716d1d70b906970/python/deltalake/writer.py#L494-L509

An easy fix would be to make the second function accessible, removing the double underscore, and using it inside the reader.

roeap:
# Description

This PR does some more cleanup in how we handle encoding of file paths and partition values.

specifically, the following should be fixed

- we were sometimes writing url-encoded strings as partition values in add actions
- we were writing un-encoded paths to add actions

Building on the updates in #1613, this PR moves the encoding/decoding for the `path` fields in add / remove / cdc actions directly to serde,which covers all current code paths where create actions.

Last, we never explictly encoded the values when creating a partition path, but let `object_store::Path` handle that for us, which looks a bit different to what spark and pyarrow are doing - not saying they are doing the same thing :).#

While debugging #1591, I think I came across some situations where I don not see a clear path how we can make that fully work. I.e. `object_store` is more restrictive in the caracters it allows in a path then sparks implementation is. While file paths are somewhat arbitrary in delta, we cannot process paths, that contain characters object, store deems illegal. As a consequence - I think - there are certain paths that spark will write, that object store will not process, Some example characters include: `}|<>~`.

Similarly pyarrows partitioning can produce paths the we cannot process - i.e. if a partition value contains some of the chars above. If we create the same table using the rust writer, we can load it again, but for some reason spark cannot.

As a follow-up we should investigate a bit deeper. While we may not be able to cover all cases, I believe there is room for improvements by finetuning what chararcters we encode and where. When we do this we can also cover all of  #1215 - at tlest I think we can.

# Related Issue(s)

- closes #1651
- closes #1653
- closes #1591
- closes #1299
- closes #1111


ion-elgreco:
@roeap could https://github.com/apache/incubator-opendal perhaps be an alternative to object_store with regards to that url parsing?

roeap:
> could https://github.com/apache/incubator-opendal perhaps be an alternative to object_store with regards to that url parsing?

Maybe :) - generally speaking though are we very happy with obejct_store and there is ahigh likelyhood that similar things would arise with opendal, as there are even more systems that are supposed to behave the same behind a sahred API ... and getting it fully integrated would like be a quite massive amount fo work.

rtyler (approved):

natinimni:
Hi @roeap, it seems delta-rs does not encode file paths, at least not completely - when trying to add a file with path: `/str_col=hello world/file.parquet`, the space character in the path is not encoded (nor the '=' character), and unfortunately when the path is provided "pre encoded", i.e. `/str_col=hello%20world/file.parquet` the result is double encoding, as the '%' character is getting encoded to "%25", resulting with: `/str_col=hello%2520world/file.parquet`.
I am writing this in the context of this PR, because it seems like spaces and some other characters are omitted from the list of "INVALID" characters, and I wonder if that was somehow on purpose?  
AFAIK, the [delta protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file) dictates that add/remove actions' 'path' field should be kept URI-encoded: 
> A relative path to a data file from the root of the table or an absolute path to a file that should be added to the table. The path is a URI as specified by [RFC 2396 URI Generic Syntax](https://www.ietf.org/rfc/rfc2396.txt), **which needs to be decoded to get the data file path**.

roeap:
Thanks for the report @natinimni. The handling of path encoding has recently been at least been consolidated a bit, but to improve testing I opened this [issue](https://github.com/delta-io/delta-rs/issues/2146).

One of the issues is, that the object_store crate which actually does the encoding took a different approach focussing on an encoding that would produce valid paths for all object store variants. In the next release with the next version of object store included, the encoding was moved a bit closed to at least what spark does...

That said, there is one thought I had that I'd appreciate your opinion on. Since delta actually does not use the partition values encoded in the file paths for anything, I planned on making the partitioning scheme configurable and defaulting to not including any partitions values in the file paths - thereby avoiding the issue all together...

natinimni:
Thank you @roeap for your prompt reply and for opening a new issue on this.
Indeed I see that the "serde_path" code was moved to under "/core/src/kernel/models/actions.rs". 

Would you consider approving a PR that adds a few key characters to the "INVALID" list (the characters that undergo encoding)? 
I understand that it won't be a complete solution, mainly because "string" partition can assume any value, which may certainly contain specials characters that are ought to be encoded, but I think that even just adding the `:` and ` ` (whitespace) characters to that list would go a long way. WDYT?  

As for you suggestion of not using partition values in paths - it absolutely makes sense. We actually use delta-rs only to read/write the delta log, while using other proprietary tools to write the data files and manage the table, so we won't directly benefit from such a change. I did however considered doing exactly that in our system, and decouple partition values from the path. 
In any case, this would still be just a workaround as we do want users to be able to configure paths, and it sounds you intend to to the same in delta-rs.

roeap:
Hey @natinimni,

certainly would. Unfortunately also realized I may have made a mistake in a recent refactor w.r.t. path encoding. Essentially we also need to take care of these lines

https://github.com/delta-io/delta-rs/blob/467afc57a2fa95ac37ee4241666040cf7c87b31d/crates/core/src/kernel/snapshot/log_data.rs#L161-L168

If you are up for it two things would be great. 

1) running encoding through the same function for the `Add` action and in the location above.
2) having a roundtrip tests with spark for these cases, since we had situations where we might end up not being able to read each others tables.

Since you mention `:` explicitly, I assume this may also be windows related where this is a legal (or illegal I don't remember) character in a path?

I am happy to help out if this is beyond what you were planning.

natinimni:
Thanks @roeap - I can sure try to include these in a PR, but I would need some more context :)
Regarding # 1 - are you suggesting decorating `path` property of `LogicalFile<'a>`  with `serde_path.encode #[serde(with = "serde_path")]` ?
Regarding # 2 - do we currently have similar spark-roundtrip tests I can base these new tests of?

My motivation of encoding `:` characters is actually for HTTP URI paths :)
By the way, do have an estimate on when the next delta-rs version would be available? the last one was published months ago..

PR #1668, opened 2023-09-26
feat: allow to set large dtypes for the schema check in `write_deltalake`

ion-elgreco:
# Description
Currently it was always checking the schema for non-large types, I didn't know before we could change it so in polars we added some schema casting from large to non-large, this however became a problem today when I wanted to write 200M records at once because the array was too big the fit in normal string type.

```python
ArrowInvalid: Failed casting from large_string to string: input array too large
```

Adding this flag will allow libraries like polars to write directly with their large dtypes in arrow. If this is merged, I can work on fix in polars to remove the schema casting for these large types.

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -142,6 +142,7 @@ def write_deltalake(
     :param overwrite_schema: If True, allows updating the schema of the table.
     :param storage_options: options passed to the native delta filesystem. Unused if 'filesystem' is defined.
     :param partition_filters: the partition filters that will be used for partition overwrite.
+    :param large_dtypes: If True, the table schema is checked against large_dtypes
     """
     if _has_pandas and isinstance(data, pd.DataFrame):
         if schema is not None:
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -92,6 +92,7 @@ def write_deltalake(
     overwrite_schema: bool = False,
     storage_options: Optional[Dict[str, str]] = None,
     partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
+    large_dtypes: bool = False,
 ) -> None:
     """Write to a Delta Lake table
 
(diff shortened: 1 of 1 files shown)


wjones127:
> a problem today when I wanted to write 200M records at once because the array was too big the fit in normal string type.

It looks like this happens here: https://github.com/pola-rs/polars/blob/f885aceb2d9a9bb8a5a715042b4c2b4573bddd40/py-polars/polars/dataframe/frame.py#L3609-L3617


As a workaround, you should be able to convert in smaller batches:

```python
data = self.to_arrow()

schema = delta_write_options.pop("schema", None)
if schema is None:
    schema = _convert_pa_schema_to_delta(data.schema)

def batch_iter():
    for batch in data.to_batches(batch_size=10_000)
        batch = pa.Table.from_batches([batch]).cast(schema).to_batches()[0]
        yield batch

data = pa.RecordBatchReader.from_batches(batch_iter(), schema)
```

ion-elgreco:
> > a problem today when I wanted to write 200M records at once because the array was too big the fit in normal string type.
> 
> It looks like this happens here: https://github.com/pola-rs/polars/blob/f885aceb2d9a9bb8a5a715042b4c2b4573bddd40/py-polars/polars/dataframe/frame.py#L3609-L3617
> 
> As a workaround, you should be able to convert in smaller batches:
> 
> ```python
> data = self.to_arrow()
> 
> schema = delta_write_options.pop("schema", None)
> if schema is None:
>     schema = _convert_pa_schema_to_delta(data.schema)
> 
> def batch_iter():
>     for batch in data.to_batches(batch_size=10_000)
>         batch = pa.Table.from_batches([batch]).cast(schema).to_batches()[0]
>         yield batch
> 
> data = pa.RecordBatchReader.from_batches(batch_iter(), schema)
> ```

That's a useful workaround. Could be a containment until, we can use next version of delta-rs with write_deltalake(large_dtypes=True)

wjones127 (approved):

PR #1671, opened 2023-09-26
fix: change partitioning schema from large to normal string for pyarrow<12

### LINKED ISSUE #1669: Writing with large arrow types (e.g. large_utf8), writes wrong partition encoding
# Environment

**Delta-rs version**:
0.11.0
0.10.2
0.10.1

**Binding**: 
python

**Environment**:
- **Cloud provider**: Azure
- **OS**: WSL/Ubuntu
- **Other**:

***
# Bug

**What happened**:
When you write arrow table with large_utf8 the written partition encoding is incorrect.
image

Write the correct string encoding just like it does with small utf8 type.
image

**How to reproduce it**:
```python
import polars as pl
import pyarrow as pa
from polars.io.delta import _convert_pa_schema_to_delta

df = pl.DataFrame({
    "foo": ["A","B","C"],
    "bar": ["1","2","3"]
})
arrow_data = df.to_arrow()

small_schema = _convert_pa_schema_to_delta(arrow_data.schema)
arrow_data_small = arrow_data.cast(small_schema)

write_deltalake('TEMP_large_utf8', arrow_data, partition_by=['foo'])
write_deltalake('TEMP_utf8', arrow_data_small, partition_by=['foo'])

```

**More details**:

ion-elgreco:
# Description
If pyarrow is below v12.0.0 it changes the partitioning schema fields from large_string to string.

# Related Issue(s)
closes #1669 

# Documentation
https://github.com/apache/arrow/issues/34546#issuecomment-1466587443

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -207,7 +207,7 @@ def write_deltalake(
         current_version = -1
 
     dtype_map = {
-        pa.large_string(): pa.string(),
+        pa.large_string(): pa.string(), # type: ignore
     }
 
     def _large_to_normal_dtype(dtype: pa.DataType) -> pa.DataType:
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -206,8 +206,26 @@ def write_deltalake(
     else:  # creating a new table
         current_version = -1
 
+    dtype_map = {
+        pa.large_string(): pa.string(),
+    }
+
+    def _large_to_normal_dtype(dtype: pa.DataType) -> pa.DataType:
+        try:
+            return dtype_map[dtype]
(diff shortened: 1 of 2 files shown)


wjones127 (changes requested):
Could we add a unit test for this? we do run CI with PyArrow at the minimal supported version, so it would be good to have this checked there :)

ion-elgreco:
@wjones127 I've added the test. I also got a linting error on mypy with regards to pyarrow. I just put a `# type: ignore`  😛. I didn't see how I could not solve that, since if you run mypy on the tests it gives the same errors but there they are ignored in the toml..

wjones127 (approved):

PR #1672, opened 2023-09-27
fix: use epoch instead of ce for date stats

### LINKED ISSUE #1670: min & max values are wrong when writing `Date32` dtype
# Environment

**Delta-rs version**:
```toml
[dependencies]
datafusion = "31.0.0"
tokio = { version = "1.32.0", features = ["rt-multi-thread"] }

[dependencies.deltalake]
git = "https://github.com/delta-io/delta-rs.git"
rev = "93eb9cefba67fcb91ba99eb9cbc7874f089303bb"
features = ["s3", "gcs", "azure", "datafusion", "arrow", "parquet"]
```
**Binding**: 
rust 

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

**What happened**:
datafusion predicate pushdowns fail for `Date` parquet datatype

result set is not empty & the predicate is properly applied

**How to reproduce it**:
fully reproducible example here 
https://github.com/GlareDB/glaredb/issues/1834

simplified example here
```rust
use datafusion::prelude::SessionContext;
use std::sync::Arc;

#[tokio::main]
async fn main() {
    let mut ctx = SessionContext::new();
    // assuming the lineitem dataset as created from parquet
    let table = deltalake::open_table("./lineitems/").await.unwrap();
    println!("{:?}", table);
    ctx.register_table("lineitem", Arc::new(table)).unwrap();

    let batches = ctx
        .sql("SELECT * FROM lineitem where l_shipdate > date '1995-03-09';")
        .await
        .unwrap()
        .collect()
        .await
        .unwrap();
    println!("batches {:?}", batches)
}
```
**More details**:

the datafusion plan shows that the predicate is getting pushed down to delta exec

```sql
explain SELECT l_shipdate FROM lineitem2 where l_shipdate = date '1995-03-09';
```

It actually looks like the file groups are empty? 
```
    CoalesceBatchesExec: target_batch_size=8192
      RepartitionExec: partitioning=UnknownPartitioning(1), input_partitions=0
        LocalLimitExec: fetch=1
          CoalesceBatchesExec: target_batch_size=8192
            FilterExec: l_shipdate@0 = 9198
              ParquetExec: 
                file_groups={0 groups: []}, 
                projection=[l_shipdate], 
                predicate=l_shipdate@10 = 9198, 
                pruning_predicate=l_shipdate_min@0 <= 9198 AND 9198 <= l_shipdate_max@1
```

the same query on parquet returns the proper file groups
```
    CoalescePartitionsExec
      CoalesceBatchesExec: target_batch_size=8192
        FilterExec: l_shipdate@0 = 9198
          ParquetExec: 
            file_groups={12 groups: [
                [....lineitem/00.parquet:0..22567114], 
                [....lineitem/00.parquet:22567114..45134228], 
                [....lineitem/00.parquet:45134228..67701342], 
                [....lineitem/00.parquet:67701342..90268456], 
                [....lineitem/00.parquet:90268456..112835570], 
                ...
            
            ]}, 
            projection=[l_shipdate], 
            predicate=l_shipdate@10 = 9198, 
            pruning_predicate=l_shipdate_min@0 <= 9198 AND 9198 <= l_shipdate_max@1
```

universalmind303:
# Description
date32 statistics logic was subjectively wrong. It was using `from_num_days_from_ce_opt` which 
> Makes a new NaiveDate from a day's number in the proleptic Gregorian calendar, with January 1, 1 being day 1.

while date32 is commonly  represented as days since UNIX epoch (1970-01-01)

# Related Issue(s)
closes #1670

# Documentation
It doesn't seem like parquet actually has a spec for what a `date` should be, but many other tools seem to use the epoch logic.

duckdb, and polars seem to use epoch instead of gregorian. 

Also arrow spec states that date32 should be epoch.

for example, if i write using polars
```py
import polars as pl

# %%
df = pl.DataFrame(
    {
        "a": [
            10561,
            9200,
            9201,
            9202,
            9203,
            9204,
            9205,
            9206,
            9207,
            9208,
            9199,
        ]
    }
)
# %%

df.select(pl.col("a").cast(pl.Date)).write_delta("./db/polars/")
```
the stats are correctly interpreted
```
{"add":{"path":"0-7b8f11ab-a259-4673-be06-9deedeec34ff-0.parquet","size":557,"partitionValues":{},"modificationTime":1695779554372,"dataChange":true,"stats":"{\"numRecords\": 11, \"minValues\": {\"a\": \"1995-03-10\"}, \"maxValues\": {\"a\": \"1998-12-01\"}, \"nullCount\": {\"a\": 0}}"}}
```


wjones127 (approved):

wjones127:
Parquet spec seems to be epoch: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#date

Thanks for catching this!

universalmind303:
@wjones127 seems like the `test_restore_by_datetime` test is failing. 

not sure if this is due to some older delta versions using the prior logic, or if it's something else. I'm unable to reproduce it locally.

universalmind303:
@wjones127 can you re-run the workflows? it was passing but had to merge in latest changes from main

PR #1673, opened 2023-09-27
fix: unify environment variables referenced by Databricks docs

### LINKED ISSUE #1627: Standardize Databricks environment variable names
# Description

In the [documentation](https://delta-io.github.io/delta-rs/python/usage.html#loading-a-delta-table) I can see that in order to authenticate to Databricks Unity Catalog I need to use environment variables `DATABRICKS_WORKSPACE_URL` and `DATABRICKS_ACCESS_TOKEN`

However, Databricks has worked recently to standardize the environment variables that hold this information with different names, they use `DATABRICKS_HOST` and `DATABRICKS_TOKEN` as stated in [section 2.e. of this documentation](https://learn.microsoft.com/en-us/azure/databricks/dev-tools/databricks-connect-ref#--step-2-configure-connection-properties)

It would be great to Unify environment variable names that `deltalake` used with the ones Databricks uses so that users don't have to maintain 2 different sets of environment variables

rtyler:
Long-term fix will be for Databricks to release a Rust SDK for Unity :smile:

Fixes #1627

### DIFF
--- a/rust/src/data_catalog/unity/mod.rs
+++ b/rust/src/data_catalog/unity/mod.rs
@@ -84,7 +84,11 @@ pub enum UnityCatalogConfigKey {
     /// - `unity_workspace_url`
     /// - `databricks_workspace_url`
     /// - `workspace_url`
+    #[deprecated(since = "0.17.0", note = "Please use the DATABRICKS_HOST env variable")]
     WorkspaceUrl,
+
+    /// Host of the Databricks workspace
+    Host,
 
     /// Access token to authorize API requests
     ///
@@ -92,7 +96,14 @@ pub enum UnityCatalogConfigKey {
     /// - `unity_access_token`
     /// - `databricks_access_token`
     /// - `access_token`
+    #[deprecated(
+        since = "0.17.0",
+        note = "Please use the DATABRICKS_TOKEN env variable"
+    )]
     AccessToken,
+
+    /// Token to use for Databricks Unity
+    Token,
 
     /// Service principal client id for authorizing requests
     ///
@@ -167,6 +178,7 @@ pub enum UnityCatalogConfigKey {
 impl FromStr for UnityCatalogConfigKey {
     type Err = DataCatalogError;
 
+    #[allow(deprecated)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "access_token" | "unity_access_token" | "databricks_access_token" => {
@@ -187,6 +199,7 @@ impl FromStr for UnityCatalogConfigKey {
             "federated_token_file"
             | "unity_federated_token_file"
             | "databricks_federated_token_file" => Ok(UnityCatalogConfigKey::FederatedTokenFile),
+            "host" => Ok(UnityCatalogConfigKey::Host),
             "msi_endpoint" | "unity_msi_endpoint" | "databricks_msi_endpoint" => {
                 Ok(UnityCatalogConfigKey::MsiEndpoint)
             }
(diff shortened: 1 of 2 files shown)


python/docs/source/usage.rst
```diff
@@ -71,15 +71,15 @@ For AWS Glue catalog, use AWS environment variables to authenticate.
     {'id': [5, 7, 9, 5, 6, 7, 8, 9]}
 
 For Databricks Unity Catalog authentication, use environment variables:
-  * DATABRICKS_WORKSPACE_URL (e.g. https://adb-62800498333851.30.azuredatabricks.net)
-  * DATABRICKS_ACCESS_TOKEN
+  * DATABRICKS_HOSAT (e.g. https://adb-62800498333851.30.azuredatabricks.net)
```
wjones127, line 74:
```suggestion
  * DATABRICKS_HOST (e.g. https://adb-62800498333851.30.azuredatabricks.net)
```

rust/src/data_catalog/unity/mod.rs
```diff
@@ -81,18 +81,16 @@ pub enum UnityCatalogConfigKey {
...
 
     /// Access token to authorize API requests
     ///
     /// Supported keys:
-    /// - `unity_access_token`
-    /// - `databricks_access_token`
-    /// - `access_token`
-    AccessToken,
+    /// - `unity_token
+    /// - `databricks_token
```
wjones127, line 92:
```suggestion
    /// - `unity_token`
    /// - `databricks_token`
```

rust/src/data_catalog/unity/mod.rs
```diff
@@ -169,9 +167,7 @@ impl FromStr for UnityCatalogConfigKey {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
-            "access_token" | "unity_access_token" | "databricks_access_token" => {
-                Ok(UnityCatalogConfigKey::AccessToken)
-            }
+            "unity_token" | "databricks_token" => Ok(UnityCatalogConfigKey::Token),
```
wjones127, line 170: Hmmm can we leave the old ones in but emit a deprecation warning? Something like:

```rust
// TODO: Remove these variables
"access_token" | "unity_access_token" | "databricks_access_token" => {
    log::warn!("Environment variable '{}' is deprecated and will be remove in a future veresion. Please use UNITY_TOKEN or DATABRICKS_TOKEN instead.");
    Ok(UnityCatalogConfigKey::Token)
}
```

This will just save us some work answering questions about why their environment variables suddenly don't work anymore.

rust/src/data_catalog/unity/mod.rs
```diff
@@ -199,9 +195,7 @@ impl FromStr for UnityCatalogConfigKey {
             "use_azure_cli" | "unity_use_azure_cli" | "databricks_use_azure_cli" => {
                 Ok(UnityCatalogConfigKey::UseAzureCli)
             }
-            "workspace_url" | "unity_workspace_url" | "databricks_workspace_url" => {
-                Ok(UnityCatalogConfigKey::WorkspaceUrl)
-            }
+            "unity_host" | "databricks_host" => Ok(UnityCatalogConfigKey::Host),
```
wjones127, line 198: Same thing here, would prefer to provide a deprecation warning first.

rust/src/data_catalog/unity/mod.rs
```diff
@@ -81,18 +81,16 @@ pub enum UnityCatalogConfigKey {
     /// Url of a Databricks workspace
     ///
     /// Supported keys:
-    /// - `unity_workspace_url`
-    /// - `databricks_workspace_url`
-    /// - `workspace_url`
-    WorkspaceUrl,
+    /// - `unity_host
+    /// - `databricks_host
```
wjones127, line 85:
```suggestion
    /// - `unity_host`
    /// - `databricks_host`
```

wjones127 (changes requested):
Let's deprecate the old ones before we make the breaking change.

rust/src/data_catalog/unity/mod.rs
rtyler: I don't disagree but don't feel like doing that work today, so I'm going to drop this out of the milestone and release 0.16

wjones127 (approved):

PR #1679, opened 2023-09-28
feat: write support for Ray data sets in Python

### LINKED ISSUE #1593: Delta-rs Write with Ray DataSet
# Description
Currently delta-rs write takes pyarrow.Table, BatchRecord and pandas . ray dataset is a distributed dataset with underlying record is ObjectRef of pa.Table 

**Use Case**
Ability to Write from ray dataset directly to delta lake 

I have added my changes in fork repo 
https://github.com/raviranak/delta-rs

raviranak:
# Description
1. Ability to write to delta lake with input format of Ray Dataset 
2. Ability to read to delta lake with Ray Dataset format 

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/1593

# Documentation

Usage 
`
from deltalake import write_table
write_table.write_deltalake_ray(<s3_path_location>, df, mode="overwrite")
`

`
 from deltalake import write_table
write_table.write_deltalake_ray(<s3_path_location>, df, mode="append")
`
https://docs.ray.io/en/latest/data/data.html


ion-elgreco:
Shouldn't this be part of Ray API, otherwise you can make the argument to include Polars, dask, DuckDB etc

raviranak:
> Shouldn't this be part of Ray API? otherwise you can make the argument to include Polars, dask, DuckDB etc

Agreed. But currently i don't see strait forward way to build this on ray as most functions are not abstracted and are tied to deltalake internally . Would you recommend to spend time to work on abstraction on deltas-rs to be able to build on ray ?

ion-elgreco:
> > Shouldn't this be part of Ray API? otherwise you can make the argument to include Polars, dask, DuckDB etc
> 
> Agreed. But currently i don't see strait forward way to build this on ray as most functions are not abstracted and are tied to deltalake internally . Would you recommend to spend time to work on abstraction on deltas-rs to be able to build on ray ?

I like how it's done inside Polars, where delta is just another format the user can interact with from within Polars api:
 https://github.com/pola-rs/polars/blob/fda3073826eb99dcfcb8f35bebd3eae74310d670/py-polars/polars/io/delta.py#L19-L142

Pandas should imho also have this directly in their api instead of in delta-rs.

raviranak:
> We have Pandas support in the package which only works if Pandas is available, I think this could be promising to live in deltalake and will start to take a look through the code with the other maintainers

sure

python/tests/.coverage.D11-ravi-rana.71300.240387
roeap, line 1: please remove this file.

python/tests/.coverage.D11-ravi-rana.71311.033930
roeap, line 1: please remove this file.

python/deltalake/test_ray_deltars.ipynb
roeap, line 1: please remove this file.

python/wheels/deltalake-0.10.2-cp37-abi3-macosx_11_0_arm64.whl
roeap, line 1: please remove this file.

python/deltalake/write_table.py
```diff
@@ -0,0 +1,546 @@
...
+        max_partitions: Optional[int] = None,
+        name: Optional[str] = None,
+        description: Optional[str] = None,
+        configuration: Optional[Mapping[str, Optional[str]]] = None,
+        overwrite_schema: bool = False,
+        storage_options: Optional[Dict[str, str]] = None,
+        partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
+) -> None:
+    import os
+    os.environ["AWS_S3_ALLOW_UNSAFE_RENAME"] = "true"
```
roeap, line 197: defaulting to usafe behaivour is probably something we should avoid :)

roeap (changes requested):
Thanks for tackling this @raviranak. 

I have a few thoughts and would also love to hear @wjones127 's opinion on this. With the changes in this PR, we introduce quite some new API surface, as well as code duplication, that I think we can avoid.

* The `read_delta` function could become a `to_ray` member function on `DeltaTable` itself - much like we do for pandas.
* Ideally, we would also leverage some introspection to figure out the type of dataframe passed to `write_deltalake`, so we can just have that one entry point for writing data. On first glance it seems the parameters are more or less identical.

To properly distribute this, we would also need to make sure, ray is not a required dependency, so everything is safe to import also when ray is not available in the environment. Pandas can server as a template here as well. To install it we would then add an extra ...

Lastly, there are some files, that should not be committed to the repo.

python/tests/.coverage.D11-ravi-rana.71300.240387
raviranak: done

python/tests/.coverage.D11-ravi-rana.71311.033930
raviranak: done

python/deltalake/test_ray_deltars.ipynb
raviranak: done

python/wheels/deltalake-0.10.2-cp37-abi3-macosx_11_0_arm64.whl
raviranak: done

python/deltalake/write_table.py
raviranak: cleaned

raviranak:
> Thanks for tackling this @raviranak.
> 
> I have a few thoughts and would also love to hear @wjones127 's opinion on this. With the changes in this PR, we introduce quite some new API surface, as well as code duplication, that I think we can avoid.
> 
> * The `read_delta` function could become a `to_ray` member function on `DeltaTable` itself - much like we do for pandas.
> * Ideally, we would also leverage some introspection to figure out the type of dataframe passed to `write_deltalake`, so we can just have that one entry point for writing data. On first glance it seems the parameters are more or less identical to `write_deltalake_ray`.
> 
> To properly distribute this, we would also need to make sure, ray is not a required dependency, so everything is safe to import also when ray is not available in the environment. Pandas can server as a template here as well. To install it we would then add an extra ...
> 
> Lastly, there are some files, that should not be committed to the repo.

1. Modified for read_data to to_ray similar to to_pandas 
2. Would it be better if i can use the write_deltalake with input of one more dataframe of tye ray Dataset like 

Union[
        "pd.DataFrame",
        "ray.data.dataset.Dataset",
        pa.Table,
        pa.RecordBatch,
        Iterable[pa.RecordBatch],
        RecordBatchReader,
    ]

and use the flow of write_table as private member function ? As I feel the schema , write behaviour is significantly different so accommodating these logic within current write_deltalake would lead to complicated code. Wanted to know about your thoughts here as well 

3. cleaned the import for ray

python/deltalake/table.py
```diff
@@ -607,6 +614,35 @@ def to_pandas(
...
+            filesystem: filesystem implementation to read from
+            columns: list of columns to read
+            parallelism: requested parallelism of read, may be limited by number of files
+
+        Returns:
+            Dataset holding Arrow records read from the Delta Lake Table
+        """
+
+        return read_parquet(
+            paths=self.file_uris(),
```
wjones127, line 639: I'd be reticent to merge this, because this isn't future-proof to new versions of the Delta Lake format. The idea that a Delta Table is just a list of Parquet files is outdated, given features like [column mapping](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-mapping) and [deletion vectors](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#deletion-vectors).

If you do want to go through with this, then we at least need to place a guard limiting the protocol `minReaderVersion` to 1.

python/deltalake/write_table.py
wjones127, line 1: It seems like you are writing a distributed writer, but one that seems specific to Ray. I think we should think about how we can make this generic so it's not tied to a particular engine.

Could be something like:

```python
def write_data_files(data: pa.RecordBatchReader, table_uri: str) -> List[AddAction]: ...

def commit_append(add_actions: List[AddAction]]): ...
```

We have something like this in Lance:

https://lancedb.github.io/lance/api/python/lance.html#lance.LanceOperation.Append
https://lancedb.github.io/lance/api/python/lance.html#lance.LanceDataset.commit

wjones127 (changes requested):
> Would you recommend to spend time to work on abstraction on deltas-rs to be able to build on ray ?

Yeah I'd generally prefer to make sure we have good abstractions here that could be build outside.

We have "pandas" integration, but it doesn't do much Pandas-specific other than call `to_pandas()` on Arrow data structures. Whereas, as-is this PR has a lot of Ray-specific code.

python/deltalake/table.py
raviranak: sure , will add guard against this.

python/deltalake/write_table.py
raviranak: sure. will work on this

python/deltalake/table.py
raviranak: added

MrPowers:
I commented on the issue, but also dropping a comment here that there is a [deltaray](https://github.com/delta-incubator/deltaray) project.

MrPowers:
Let's add this to the deltaray project to avoid the Ray dependency in delta-rs, thank you!!

PR #1686, opened 2023-10-01
feat: add convert_to_delta

### LINKED ISSUE #1041: Add convertToDelta command
The `convertToDelta` command is a good way to attract users because it makes Delta an easy switch.  See [this blog post](https://delta.io/blog/2022-09-23-convert-parquet-to-delta/) for details.

The command adds a `_delta_log` directory to an existing Parquet data lake.  It doesn't change any of the underlying data files - it's an in-place operation.

Would be a nice addition to this library 💪

### LINKED ISSUE #1682: CONVERT TO DELTA as a pure delta-rs API
# Description
The equivalent API in delta-rs to `CONVERT TO DELTA parquet.'s3://my-bucket/parquet-data';` directly in Python.

It is https://docs.delta.io/latest/api/python/index.html#delta.tables.DeltaTable.convertToDelta for delta-rs without Spark.

**Use Case**
Let's say that we have a bunch of Parquet directories which can be quickly/effectively converted to Delta.

2nd request: generate Uniform 3.0 manifest via delta-rs. That will be great.

**Related Issue(s)**
#1041

junjunjd:
# Description
Add a convert_to_delta operation for converting a Parquet table to a Delta Table in place. 

# Related Issue(s)
- closes #1041
- closes #1682

# Documentation

### DIFF
--- a/crates/deltalake-core/src/operations/convert_to_delta.rs
+++ b/crates/deltalake-core/src/operations/convert_to_delta.rs
@@ -1,0 +1,842 @@
+//! Command for converting a Parquet table to a Delta table in place
+// https://github.com/delta-io/delta/blob/1d5dd774111395b0c4dc1a69c94abc169b1c83b6/spark/src/main/scala/org/apache/spark/sql/delta/commands/ConvertToDeltaCommand.scala
+
+use crate::{
+    kernel::{Action, Add, Schema, StructField},
+    logstore::{LogStore, LogStoreRef},
+    operations::create::CreateBuilder,
+    protocol::SaveMode,
+    storage::config::configure_log_store,
+    table::builder::ensure_table_uri,
+    table::config::DeltaConfigKey,
+    DeltaResult, DeltaTable, DeltaTableError, DeltaTablePartition, ObjectStoreError,
+    NULL_PARTITION_VALUE_DATA_PATH,
+};
+use arrow::{datatypes::Schema as ArrowSchema, error::ArrowError};
+use futures::{
+    future::{self, BoxFuture},
+    TryStreamExt,
+};
+use log::{debug, info};
+use parquet::{
+    arrow::async_reader::{ParquetObjectReader, ParquetRecordBatchStreamBuilder},
+    errors::ParquetError,
+};
+use percent_encoding::percent_decode_str;
+use serde_json::{Map, Value};
+use std::{
+    collections::{HashMap, HashSet},
+    num::TryFromIntError,
+    str::Utf8Error,
+    sync::Arc,
+};
+
+/// Error converting a Parquet table to a Delta table
+#[derive(Debug, thiserror::Error)]
+enum Error {
+    #[error("Object store error: {0}")]
+    ObjectStore(#[from] ObjectStoreError),
+    #[error("Arrow error: {0}")]
+    Arrow(#[from] ArrowError),
+    #[error("Parquet error: {0}")]
+    Parquet(#[from] ParquetError),
+    #[error("DeltaTable error: {0}")]
+    DeltaTable(#[from] DeltaTableError),
+    #[error("Error percent-decoding as UTF-8: {0}")]
+    PercentDecode(#[from] Utf8Error),
+    #[error("Error converting usize to i64: {0}")]
+    TryFromUsize(#[from] TryFromIntError),
+    #[error("No parquet file is found in the given location")]
+    ParquetFileNotFound,
+    #[error("The schema of partition columns must be provided to convert a Parquet table to a Delta table")]
+    MissingPartitionSchema,
+    #[error("Partition column provided by the user does not exist in the parquet files")]
+    PartitionColumnNotExist(HashSet<StructField>),
+    #[error("The given location is already a delta table location")]
+    DeltaTableAlready,
+    #[error("Location must be provided to convert a Parquet table to a Delta table")]
+    MissingLocation,
+}
+
+impl From<Error> for DeltaTableError {
+    fn from(err: Error) -> Self {
+        match err {
+            Error::ObjectStore(e) => DeltaTableError::ObjectStore { source: e },
+            Error::Arrow(e) => DeltaTableError::Arrow { source: e },
+            Error::Parquet(e) => DeltaTableError::Parquet { source: e },
+            Error::DeltaTable(e) => e,
+            _ => DeltaTableError::GenericError {
+                source: Box::new(err),
+            },
+        }
+    }
+}
+
+/// The partition strategy used by the Parquet table
+/// Currently only hive-partitioning is supproted for Parquet paths
+#[non_exhaustive]
+#[derive(Default)]
+pub enum PartitionStrategy {
+    /// Hive-partitioning
+    #[default]
+    Hive,
+}
+
+/// Build an operation to convert a Parquet table to a [`DeltaTable`] in place
+pub struct ConvertToDeltaBuilder {
+    log_store: Option<LogStoreRef>,
+    location: Option<String>,
+    storage_options: Option<HashMap<String, String>>,
+    partition_schema: HashSet<StructField>,
+    partition_strategy: PartitionStrategy,
+    mode: SaveMode,
+    name: Option<String>,
+    comment: Option<String>,
+    configuration: HashMap<String, Option<String>>,
+    metadata: Option<Map<String, Value>>,
+}
+
+impl Default for ConvertToDeltaBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ConvertToDeltaBuilder {
+    /// Create a new [`ConvertToDeltaBuilder`]
+    pub fn new() -> Self {
+        Self {
+            log_store: None,
+            location: None,
+            storage_options: None,
+            partition_schema: Default::default(),
+            partition_strategy: Default::default(),
+            mode: SaveMode::ErrorIfExists,
+            name: None,
+            comment: None,
+            configuration: Default::default(),
+            metadata: Default::default(),
+        }
+    }
+
+    /// Provide a [`LogStore`] instance, that points at table location
+    pub fn with_log_store(mut self, log_store: Arc<dyn LogStore>) -> Self {
+        self.log_store = Some(log_store);
+        self
+    }
+
+    /// Specify the path to the location where table data is stored,
+    /// which could be a path on distributed storage.
+    ///
+    /// If an object store is also passed using `with_log_store()`, this path will be ignored.
+    pub fn with_location(mut self, location: impl Into<String>) -> Self {
+        self.location = Some(location.into());
+        self
+    }
+
+    /// Set options used to initialize storage backend
+    ///
+    /// Options may be passed in the HashMap or set as environment variables.
+    ///
+    /// [crate::table::builder::s3_storage_options] describes the available options for the AWS or S3-compliant backend.
+    /// [dynamodb_lock::DynamoDbLockClient] describes additional options for the AWS atomic rename client.
+    ///
+    /// If an object store is also passed using `with_log_store()`, these options will be ignored.
+    pub fn with_storage_options(mut self, storage_options: HashMap<String, String>) -> Self {
+        self.storage_options = Some(storage_options);
+        self
+    }
+
+    /// Specify the partition schema of the Parquet table
+    pub fn with_partition_schema(
+        mut self,
+        partition_schema: impl IntoIterator<Item = StructField>,
+    ) -> Self {
+        self.partition_schema = HashSet::from_iter(partition_schema);
+        self
+    }
+
+    /// Specify the partition strategy of the Parquet table
+    /// Currently only hive-partitioning is supproted for Parquet paths
+    pub fn with_partition_strategy(mut self, strategy: PartitionStrategy) -> Self {
+        self.partition_strategy = strategy;
+        self
+    }
+    /// Specify the behavior when a table exists at location
+    pub fn with_save_mode(mut self, save_mode: SaveMode) -> Self {
+        self.mode = save_mode;
+        self
+    }
+
+    /// Specify the table name. Optionally qualified with
+    /// a database name [database_name.] table_name.
+    pub fn with_table_name(mut self, name: impl Into<String>) -> Self {
+        self.name = Some(name.into());
+        self
+    }
+
+    /// Comment to describe the table.
+    pub fn with_comment(mut self, comment: impl Into<String>) -> Self {
+        self.comment = Some(comment.into());
+        self
+    }
+
+    /// Set configuration on created table
+    pub fn with_configuration(
+        mut self,
+        configuration: impl IntoIterator<Item = (impl Into<String>, Option<impl Into<String>>)>,
+    ) -> Self {
+        self.configuration = configuration
+            .into_iter()
+            .map(|(k, v)| (k.into(), v.map(|s| s.into())))
+            .collect();
+        self
+    }
+
+    /// Specify a table property in the table configuration
+    pub fn with_configuration_property(
+        mut self,
+        key: DeltaConfigKey,
+        value: Option<impl Into<String>>,
+    ) -> Self {
+        self.configuration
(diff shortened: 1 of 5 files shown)


rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,277 @@
...
+) -> Result<DeltaTable, Error> {
+    let (mut schemas, mut actions, mut partition_columns, mut partition_schema_fields) =
+        (Vec::new(), Vec::new(), HashSet::new(), Vec::new());
+    for file in files {
+        debug!("Processing parquet file: {:#?}", file.location);
+        schemas.push(arrow_schema(&file, store.clone()).await?);
+        let partition_values =
+            partitions(&file, &mut partition_columns, &mut partition_schema_fields)?;
+        actions.push(add_action(&file, partition_values)?);
+    }
```
rtyler, line 175: Based on my read of the module added, this code will read every single `.parquet` file that exists int he configured object store and mash their schemas together. Is there particular concern taking the latest file and calling that the table's schema?

rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// Error canonicalizing the given path
+    #[error("Error canonicalizing the given path: {0}")]
+    CanonicalizePath(#[from] std::io::Error),
+    /// Error converting the path into an URL
+    #[error("Error converting the path into an URL")]
+    UrlFromFilePath,
+}
+
+/// Convert a Parquet table to a Delta table in place
+pub async fn convert_to_delta(path: &str) -> Result<DeltaTable, Error> {
```
rtyler, line 54: I would recommend this API taking an already configured `ObjectStore`. The object store can have a number of configured parameters that exist outside of the `path` We have plenty of other places where URL and path parsing exist and have caused us enough grief.

rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,277 @@
...
+            Path::from("part-00001-911a94a2-43f6-4acb-8620-5e68c2654989-c000.snappy.parquet"),
+        ]
+    )]
+    #[test_case("tests/data/delta-0.8.0-numeric-partition" =>
+        vec![
+            Path::from("x=9/y=9.9/part-00007-3c50fba1-4264-446c-9c67-d8e24a1ccf83.c000.snappy.parquet"),
+            Path::from("x=10/y=10.0/part-00015-24eb4845-2d25-4448-b3bb-5ed7f12635ab.c000.snappy.parquet"),
+        ]
+    )]
+    #[tokio::test]
```
rtyler, line 221: I'm not sure I understand the test_case crate and what this macro is doing that function calls couldn't just do? :confused:

rtyler (commented):
Thanks for the pull request! This is a great start and I've left some comments. I can help bring this further if you need it, but also feel free to ping me on Slack if you need any further help preparing this for merge!

MrPowers:
@junjunjd - did you see the comments from @rtyler?  Anything else we can do to help you with this one?

rust/src/table/convert_to_delta.rs
junjunjd: Yes, this code will read all parquet files and smash their schemas together. This is because timestamp will not be preserved when copying files in S3. A newer timestamp does not necessarily mean a newer version of the parquet file.
Some discussions https://github.com/aws/aws-cli/issues/3415, https://github.com/aws/aws-cli/discussions/7071.

rust/src/table/convert_to_delta.rs
junjunjd: Thanks for the recommendation. I updated the function to take an `ObjectStoreRef` as argument. I also added a `convert_to_delta_from_path` function that takes a path as argument, since [the `convertToDelta` command](https://docs.delta.io/latest/api/python/index.html#delta.tables.DeltaTable.convertToDelta) supports path. Would it make sense to support both functions in the API?

rust/src/table/convert_to_delta.rs
junjunjd: Removed the `test_case` macros. Agree that function calls would also work. I used `test_case` because I think it improves the readability of the tests by combining the test function and the function call into one parametrized function.

junjunjd:
> @junjunjd - did you see the comments from @rtyler? Anything else we can do to help you with this one?

I've replied to comments from @rtyler. This MR is ready for review.

I also created three follow-up issues. I've started working on the first issue #1718.
- Allow users to define the data type and nullability of a partition column when converting a Parquet table to Delta table #1718
- Collect stats on parquet files when converting a Parquet table to a Delta table #1719
- Support parquet files with nanoseconds timestamps when converting a Parquet table to a Delta table #1721

junjunjd:
Follow-up issue #1718 is done. The API now takes a vector of `SchemaField` for partition columns defined by user.
```
pub async fn convert_to_delta(
    storage: ObjectStoreRef,
    partition_schema: Vec<SchemaField>,
) -> Result<DeltaTable, Error> 
```
The changes are pushed to this MR.

junjunjd:
@rtyler @wjones127 @roeap This MR is ready for final review.

junjunjd:
@rtyler @wjones127 @roeap Is there anything I can do to help get this merged?

junjunjd:
@rtyler @wjones127 @roeap Just a gentle ping for review.

rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,765 @@
...
+pub async fn convert_to_delta(
+    storage: ObjectStoreRef,
+    partition_schema: Vec<SchemaField>,
+) -> Result<DeltaTable, Error> {
+    if storage.is_delta_table_location().await? {
+        info!("A Delta table already exists in the given object store");
+        let mut table = DeltaTable::new(storage, DeltaTableConfig::default());
+        table.load().await?;
+        return Ok(table);
+    }
```
roeap, line 77: Just thinking out lod here... Should this maybe return an error. From the perspective of converting the goven location, the operation is infeasible. On the other hand, converting delta to delta is in fact a no-op :).

rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,765 @@
...
+///
+/// * `path` - A path with "/" pointing at delta table root (i.e. where `_delta_log` is located).
+/// * `partition_schema`
+///     - A vector of [`SchemaField`] describing the data type, nullibility and metadata of partion columns.
+///     - If the schema of a partition column is not provided, schema data type is set to string and the field is set to nullable.
+///     - Errors if any partition column provided by the user does not exist in the parquet files.
+pub async fn convert_to_delta_from_path(
+    path: &str,
+    partition_schema: Vec<SchemaField>,
+) -> Result<DeltaTable, Error> {
```
roeap, line 98: We should provide a code path where we pass through storage options, otherwise this wont work for any remote store, or rather force config via the env.

rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,765 @@
...
+        }
+    }
+}
+
+fn object_store(path: &str) -> Result<ObjectStoreRef, Error> {
+    debug!("Creating an object store for path: {path:?}");
+    let location = ensure_table_uri(path)?;
+    Ok(Arc::new(DeltaObjectStore::try_new(
+        location,
+        HashMap::new(),
```
roeap, line 118: see above comment re storage options.

rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,765 @@
...
+}
+
+fn add_action(
+    file: &ObjectMeta,
+    partition_values: HashMap<String, Option<String>>,
+) -> Result<Action, Error> {
+    Ok(Action::add(Add {
+        path: percent_decode_str(file.location.as_ref())
+            .decode_utf8()?
+            .to_string(),
```
roeap, line 159: are we sure this location needs to be decoded before we pipe it through the add action?

rust/src/lib.rs
```diff
@@ -126,6 +126,9 @@ pub use protocol::checkpoints;
 #[cfg(feature = "integration_test")]
 pub mod test_utils;
 
+#[cfg(feature = "parquet")]
+const NULL_PARTITION_VALUE_DATA_PATH: &str = "__HIVE_DEFAULT_PARTITION__";
```
roeap, line 130: why does this need to be defined in the top level?

roeap (commented):
Thanks for this great wortk @junjunjd! Left some comments in the code.

More generally speaking, I believe we should model `CONVET TO DELTA` as an operation (i.e. config builder / `IntoFuture` Trait ...) and also place it in the operations module rather then next to the table. We are slowly but surely working towards also modelling things as logical plans and exposing sql APIs so convert to delta should be compatible whith the other operations we expose.

rust/src/lib.rs
junjunjd: This const used to be defined in `src/writer/utils.rs`. I moved it to `src/lib.rs` because it is also used in `operations/convert_to_delta`. IMO it doesn't make a lot of sense to make `convert_to_delta` depend on `writer/utils.rs`, thus I defined this const in the top level. Do you have any suggestions on where to place this definition?

rust/src/table/convert_to_delta.rs
junjunjd: Done.

rust/src/table/convert_to_delta.rs
junjunjd: Done.

rust/src/table/convert_to_delta.rs
junjunjd: Updated to return an error when the location is already a Delta table location. This aligns with `ConvertToDeltaCommand` documentation https://books.japila.pl/delta-lake-internals/commands/convert/ConvertToDeltaCommand/#executing-command`

rust/src/table/convert_to_delta.rs
```diff
@@ -0,0 +1,765 @@
...
+use std::{
+    collections::{HashMap, HashSet},
+    num::TryFromIntError,
+    str::Utf8Error,
+    sync::Arc,
+};
+
+/// Error converting a Parquet table to a Delta table
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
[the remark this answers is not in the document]
junjunjd: Done.

rust/src/table/convert_to_delta.rs
junjunjd: Percent decoding as UTF8 is needed here because the path of Add action is used to generate a `ScalarValue::Utf8` in the `build` method of `DeltaScanBuilder` https://github.com/delta-io/delta-rs/blob/main/crates/deltalake-core/src/delta_datafusion/mod.rs#L594-L595

Also, the `path` field has a [`#[serde(with = "serde_path")]` attribute](https://github.com/delta-io/delta-rs/blob/main/rust/src/protocol/mod.rs#L350-L351) applied to it. `path` will be decoded during [deserialization](https://github.com/delta-io/delta-rs/blob/main/rust/src/protocol/serde_path.rs#L11).

junjunjd:
@roeap Thanks for the comment. I updated the MR to model CONVET TO DELTA as an operation. This MR is ready for another review.

crates/deltalake-core/src/operations/convert_to_delta.rs
```diff
@@ -0,0 +1,807 @@
...
+            .await?
+            .try_for_each_concurrent(10, |meta| {
+                if Some("parquet") == meta.location.extension() {
+                    debug!("Found parquet file {:#?}", meta.location);
+                    files.push(meta);
+                }
+                future::ready(Ok(()))
+            });
+        // Return immediately with an error if the stream returns an error
+        fut.await?;
```
roeap, line 227:
```suggestion
        object_store
            .list(None)
            .await?
            .try_for_each_concurrent(10, |meta| {
                if Some("parquet") == meta.location.extension() {
                    debug!("Found parquet file {:#?}", meta.location);
                    files.push(meta);
                }
                future::ready(Ok(()))
            })
            .await?;
```

crates/deltalake-core/src/operations/convert_to_delta.rs
```diff
@@ -0,0 +1,807 @@
...
+                        percent_decode_str(subpath).decode_utf8()?.as_ref(),
+                    )?;
+                    debug!(
+                        "Found partition {partition:#?} in parquet file {:#?}",
+                        file.location
+                    );
+                    let (key, val) = (partition.key, partition.value);
+                    partition_values.insert(
+                        key.clone(),
+                        if val == NULL_PARTITION_VALUE_DATA_PATH {
```
roeap, line 257: Haven't dealt with many alternative partitioning schemes, but I guess this would limit `ConvertToDelta` to only use hive-partitioned tables. Not sure about the implications, but maybe we could make the partitioning scheme configurable, but have only hive partitioning as an option for now?

crates/deltalake-core/src/operations/mod.rs
```diff
@@ -14,6 +14,8 @@ use crate::errors::{DeltaResult, DeltaTableError};
 use crate::table::builder::DeltaTableBuilder;
 use crate::DeltaTable;
 
+#[cfg(feature = "parquet")]
```
roeap, line 17: i think we also need to enable the `arrow` feature.

roeap (commented):
Great work @junjunjd and thanks for sticking with it.

Just some minor nits to look at, but overall I think we are good to go!

crates/deltalake-core/src/operations/convert_to_delta.rs
```diff
@@ -0,0 +1,807 @@
...
+    MissingLocation,
+}
+
+impl From<Error> for DeltaTableError {
+    fn from(err: Error) -> Self {
+        DeltaTableError::GenericError {
+            source: Box::new(err),
+        }
+    }
+}
```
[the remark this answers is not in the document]
junjunjd: Updated to map errors to the more concrete types if there is a corresponding variant in `DeltaTableError`.

crates/deltalake-core/src/operations/convert_to_delta.rs
junjunjd: Done.

crates/deltalake-core/src/operations/mod.rs
junjunjd: Added the `arrow` feature.

rust/src/lib.rs
junjunjd: Moved `NULL_PARTITION_VALUE_DATA_PATH` to `crate::schema::partitions`

crates/deltalake-core/src/operations/convert_to_delta.rs
junjunjd: Good idea. I added a `PartitionStrategy` enum so that the partitioning scheme is configurable. For now, the enum only has a `Hive` variant. It appears that [`ConvertToDeltaCommand.scala`](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/commands/ConvertToDeltaCommand.scala#L319) also only supports hive-partitioning for parquet paths. If a parquet path is non-hive, `ConvertToDeltaCommand.scala` will throw an error. This aligns with [`DeltaTablePartition::try_from`](https://github.com/delta-io/delta-rs/pull/1686/files#diff-55016b8b61e61c4b987804307f47377cda9bde37a3dcb0e014ce4f27710bd212R252). 
I created a follow-up [issue](https://github.com/delta-io/delta-rs/issues/1828) to extend `PartitionStrategy` to support other popular partitioning schemes such as the directory scheme like `./partitioned/2000/part-0.parquet`. User will need to provide the column name when they use this scheme.

roeap:
It seems we do have some compiler errors going on. I would disregard the ones related to a future not being `Send` at first until the other issues are fixed, they may just resolve themselves :).

junjunjd:
@roeap Thanks for the review! CI is fixed. MR is ready for final review/merge.

roeap (approved):
Looking great @junjunjd, thanks for sticking with it!

PR #1687, opened 2023-10-02
feat(python): expose delete operation

### LINKED ISSUE #1417: Support delete operation in Python
# Description
The delete operation has been implemented in rust (#832) . It would be nice to support it as well in python. 
Unless I'm mistaken I haven't seen any work on that yet. 

I'd be happy to start working on the feature, as it would be my first contribution I'd probably need support along the way.

**Use Case**

**Related Issue(s)**

guilhem-dvr:
# Description
Naively expose the delete operation, with the option to provide a predicate.

I first tried to expose a richer API with the Python `FilterType` and DNF expressions, but from what I understand delta-rs doesn't implement generic filters but only `PartitionFilter`. The `DeleteBuilder` also only accepts datafusion expressions. So Instead of hacking my way around or proposing a refactor I went for the simpler approach of sending a string predicate to the rust lib.

If this implementation is OK I will add tests.

# Related Issue(s)
- closes #1417

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -199,9 +199,9 @@ _DNF_filter_doc = """
 Predicates are expressed in disjunctive normal form (DNF), like [("x", "=", "a"), ...].
 DNF allows arbitrary boolean logical combinations of single partition predicates.
 The innermost tuples each describe a single partition predicate. The list of inner
-predicates is interpreted as a conjunction (AND), forming a more selective and 
-multiple partition predicates. Each tuple has format: (key, op, value) and compares 
-the key with the value. The supported op are: `=`, `!=`, `in`, and `not in`. If 
+predicates is interpreted as a conjunction (AND), forming a more selective and
+multiple partition predicates. Each tuple has format: (key, op, value) and compares
+the key with the value. The supported op are: `=`, `!=`, `in`, and `not in`. If
 the op is in or not in, the value must be a collection such as a list, a set or a tuple.
 The supported type for value is str. Use empty string `''` for Null partition value.
 
@@ -302,13 +302,13 @@ class DeltaTable:
 
     files.__doc__ = f"""
 Get the .parquet files of the DeltaTable.
-    
+
 The paths are as they are saved in the delta log, which may either be
 relative to the table root or absolute URIs.
 
-:param partition_filters: the partition filters that will be used for 
+:param partition_filters: the partition filters that will be used for
     getting the matched files
-:return: list of the .parquet files referenced for the current version 
+:return: list of the .parquet files referenced for the current version
     of the DeltaTable
 {_DNF_filter_doc}
     """
@@ -665,6 +665,20 @@ given filters.
         2  x=1/0-91820cbf-f698-45fb-886d-5d5f5669530b-0.p...         565 1970-01-20 08:40:08.071         True            1            1             0      4      4
         """
         return self._table.get_add_actions(flatten)
+
+    def delete(self, predicate: Optional[str] = None) -> Dict[str, Any]:
+        """Delete records from a Delta Table that statisfy a predicate.
+
+        When a predicate is not provided then all records are deleted from the Delta
+        Table. Otherwise a scan of the Delta table is performed to mark any files
+        that contain records that satisfy the predicate. Once files are determined
+        they are rewritten without the records.
+
+        :param predicate: a logical expression, defaults to None
+        :return: the metrics from delete
+        """
+        metrics = self._table.delete(predicate)
+        return json.loads(metrics)
 
 
 class TableOptimizer:
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -674,8 +674,8 @@ given filters.
         that contain records that satisfy the predicate. Once files are determined
         they are rewritten without the records.
 
-        :param predicate: a logical expression, defaults to None
-        :return: the metrics from delete
+        :param predicate: a SQL where clause. If not passed, will delete all rows.
+        :return: the metrics from delete.
         """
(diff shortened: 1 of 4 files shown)


ion-elgreco:
@guilhem-dvr can you also expose the writer_properties to be passed to delete?

guilhem-dvr:
@ion-elgreco what would be a good way to expose writer_properties? I see in https://github.com/delta-io/delta-rs/pull/1685/files that I could look for every key in a HashMap, but that would be very verbose and not reusable. 

I'm new to python bindings with rust, would there be a way to expose `WriterProperties` and all the structs it relies on to the python module?

ion-elgreco:
> @ion-elgreco what would be a good way to expose writer_properties? I see in https://github.com/delta-io/delta-rs/pull/1685/files that I could look for every key in a HashMap, but that would be very verbose and not reusable.
> 
> I'm new to python bindings with rust, would there be a way to expose `WriterProperties` and all the structs it relies on to the python module?

The thing is you need to create a WriterProperties struct with the builder, so I am not sure if there is a way around checking if the value of a key exists and then calling that method to add with the builder. 




> > @guilhem-dvr can you also expose the writer_properties to be passed to delete?
> 
> Should I expose the metadata as well?

Probably good idea, I also need to add in the MERGE operation.

wjones127:
> I first tried to expose a richer API with the Python FilterType and DNF expressions, but from what I understand delta-rs doesn't implement generic filters but only PartitionFilter. The DeleteBuilder also only accepts datafusion expressions. So Instead of hacking my way around or proposing a refactor I went for the simpler approach of sending a string predicate to the rust lib.

Sorry our expressions/predicates are a mess. We have these DNF filters, PyArrow expressions, and DataFusion expressions.

As an aside, there is some work to make these all easier to convert between through [Substrait](https://github.com/substrait-io/substrait). PyArrow expressions [will have methods to convert to/from Substrait ones in 14.0.0](https://arrow.apache.org/docs/dev/python/generated/pyarrow.dataset.Expression.html#pyarrow.dataset.Expression.from_substrait), and my colleague Weston Pace is working on [similar support in DataFusion](https://github.com/apache/arrow-datafusion/issues/7687).

python/deltalake/table.py
```diff
@@ -666,6 +666,20 @@ def get_add_actions(self, flatten: bool = False) -> pyarrow.RecordBatch:
...
 
+    def delete(self, predicate: Optional[str] = None) -> Dict[str, Any]:
+        """Delete records from a Delta Table that statisfy a predicate.
+
+        When a predicate is not provided then all records are deleted from the Delta
+        Table. Otherwise a scan of the Delta table is performed to mark any files
+        that contain records that satisfy the predicate. Once files are determined
+        they are rewritten without the records.
+
+        :param predicate: a logical expression, defaults to None
```
wjones127, line 677:
```suggestion
        :param predicate: a SQL where clause. If not passed, will delete all rows.
```

wjones127 (changes requested):
I'm fine with taking a SQL string for now. It's messy, but we can migrate away from it once we have a better expression situation.

Could you add a unit test in Python for this? One for case of deleting some rows, one for deleting a partition, one for deleting all rows.

ion-elgreco:
> > I first tried to expose a richer API with the Python FilterType and DNF expressions, but from what I understand delta-rs doesn't implement generic filters but only PartitionFilter. The DeleteBuilder also only accepts datafusion expressions. So Instead of hacking my way around or proposing a refactor I went for the simpler approach of sending a string predicate to the rust lib.
> 
> Sorry our expressions/predicates are a mess. We have these DNF filters, PyArrow expressions, and DataFusion expressions.
> 
> As an aside, there is some work to make these all easier to convert between through [Substrait](https://github.com/substrait-io/substrait). PyArrow expressions [will have methods to convert to/from Substrait ones in 14.0.0](https://arrow.apache.org/docs/dev/python/generated/pyarrow.dataset.Expression.html#pyarrow.dataset.Expression.from_substrait), and my colleague Weston Pace is working on [similar support in DataFusion](https://github.com/apache/arrow-datafusion/issues/7687).
> 
> 

That's great, does that mean once substrait support is there in data fusion, we could essentially support any library in Python that can convert expressions to substraits (for example Polars)?

guilhem-dvr:
@ion-elgreco I attempted to expose parquet's `WriterProperties` to Python - with a proper class through the binding - but ultimately not all fields are easily parseable. It would be doable but I'm not sure it would belong to this CL, and I also wonder if some of the features that are needed shouldn't be part of parquet instead of deltalake.

Also, I had issues with the app metadata between pyo3 and serde_json - but that's more down to my Rust knowledge I think.

python/tests/test_delete.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_delete_a_partition(tmp_path: pathlib.Path, sample_data: pa.Table):
+    write_deltalake(tmp_path, sample_data, partition_by=["bool"])
+
+    dt = DeltaTable(tmp_path)
+    old_version = dt.version()
+
+    expr = pc.field("bool") != pc.scalar(True)
+    expected_table = sample_data.filter(expr)
```
wjones127, line 31: This doesn't work in older versions of PyArrow, which we test with. Perhaps instead it would be sufficient to do:

```suggestion
    mask = pc.equal(sample_data['bool'], False)
    expected_table = sample_data.filter(make)
```

wjones127 (commented):
One minor change, then I think this is ready to go.

wjones127 (dismissed):

wjones127 (dismissed):

wjones127 (approved):

PR #1694, opened 2023-10-03
feat(python): expose UPDATE operation

### LINKED ISSUE #1505: Expose Update & Delete operations from Rust into Python
# Description
Since the Update, Delete operations have been developed in Rust, would be nice to expose them also in Python.

ion-elgreco:
# Description

- Exposes UPDATE operation to Python.
- Added two test cases, with predicate and without
- Took some learnings in simplifying the code (will apply it in MERGE PR as well)

# Related Issue(s)

Closes #1505


python/deltalake/table.py
```diff
@@ -456,6 +456,27 @@ def vacuum(
...
+
+        :param updates: what values to update
+        :param predicate: a logical expression, defaults to None
+        :writer_properties: Pass writer properties to the Rust parquet writer, see options https://arrow.apache.org/rust/parquet/file/properties/struct.WriterProperties.html
+        :strict_cast: specify if data types need to be casted strictly or not :default = True
+        :return: the metrics from delete
+        """
+
+        metrics = self._table.update(
+            updates, predicate, writer_properties, safe_cast=not strict_cast
```
wjones127, line 476: It doesn't seem right to me that `safe_cast` and `strict_cast` would be antonyms. Do we have a test validating this?

python/deltalake/table.py
```diff
@@ -456,6 +456,27 @@ def vacuum(
             enforce_retention_duration,
         )
 
+    def update(
+        self,
+        updates: Dict[str, str],
+        predicate: Optional[str] = None,
+        writer_properties: Optional[Dict[str, int]] = None,
```
wjones127, line 463: Once we [drop support for Python 3.7](https://github.com/delta-io/delta-rs/issues/1703), I'd love for this to be a [TypedDict](https://peps.python.org/pep-0589/)

python/deltalake/table.py
```diff
@@ -456,6 +456,27 @@ def vacuum(
...
+        strict_cast: bool = True,
+    ) -> Dict[str, Any]:
+        """UPDATE records in the Delta Table that matches an optional predicate.
```
wjones127, line 466: Could we add an examples section ([like this](https://github.com/delta-io/delta-rs/blob/4da7d66d06985d386e61d3fb124cad6680594dcc/python/deltalake/table.py#L649-L665))? I feel like it might not be obvious how to use the `updates` param in particular.

If you are willing, it would be awesome to have a section in https://github.com/delta-io/delta-rs/blob/main/python/docs/source/usage.rst

python/deltalake/table.py
```diff
@@ -456,6 +456,27 @@ def vacuum(
...
+
+        :param updates: what values to update
```
wjones127, line 468: This could be expanded upon, but I think at a minimum we should say what the keys and values are supposed to be:

```suggestion
        :param updates: a mapping of column name to update SQL expression.
```

python/deltalake/table.py
```diff
@@ -456,6 +456,27 @@ def vacuum(
             enforce_retention_duration,
         )
 
+    def update(
+        self,
+        updates: Dict[str, str],
+        predicate: Optional[str] = None,
+        writer_properties: Optional[Dict[str, int]] = None,
+        strict_cast: bool = True,
```
wjones127, line 464: Maybe a better name is `error_on_type_mismatch: boal = True`? As a reader of this, I'm not sure what updating values has to do with casting.

python/deltalake/table.py
```diff
@@ -456,6 +456,27 @@ def vacuum(
...
+    ) -> Dict[str, Any]:
+        """UPDATE records in the Delta Table that matches an optional predicate.
+
+        :param updates: what values to update
+        :param predicate: a logical expression, defaults to None
+        :writer_properties: Pass writer properties to the Rust parquet writer, see options https://arrow.apache.org/rust/parquet/file/properties/struct.WriterProperties.html
+        :strict_cast: specify if data types need to be casted strictly or not :default = True
+        :return: the metrics from delete
```
wjones127, line 472: Perhaps you could say what fields are in the metrics? Or at least the important ones?

python/tests/test_update.py
```diff
@@ -0,0 +1,67 @@
...
+    expected = pa.table(
+        {
+            "id": pa.array(["1", "2", "3", "4", "5"]),
+            "price": pa.array(list(range(nrows)), pa.int64()),
+            "sold": pa.array(list(range(nrows)), pa.int32()),
+            "deleted": pa.array([False, False, False, False, True]),
+        }
+    )
+
+    dt.update(updates={"deleted": "True"}, predicate="price > 3")
```
wjones127, line 37: I'm confused. `4` should satisfy the predicate `price > 3`. Why isn't that row modified?

wjones127 (changes requested):
Thanks for creating this PR. This will be quite valuable, though we should probably make sure we document and test this well first.

python/tests/test_update.py
ion-elgreco: Because Python range(5) will start at 0. (0..4)

python/deltalake/table.py
ion-elgreco: I will add the test. Reason why I named it strict_cast is that I found safe_cast in rust confusing. Safe_cast = False means it will give an error when the types are not the same.

python/deltalake/table.py
ion-elgreco: I'll make a note of it, let me know if there is a PR that is dropping the support so I can track it.

python/deltalake/table.py
ion-elgreco: Will do! Forgot to add it

python/deltalake/table.py
ion-elgreco: Good one, that would be clear enough!

python/tests/test_update.py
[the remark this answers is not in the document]
ion-elgreco: I've added a test for both.

python/deltalake/table.py
ion-elgreco: Added also section in the docs.

python/deltalake/table.py
```diff
@@ -456,6 +456,50 @@ def vacuum(
...
+        :param updates: a mapping of column name to update SQL expression.
+        :param predicate: a logical expression, defaults to None
+        :writer_properties: Pass writer properties to the Rust parquet writer, see options https://arrow.apache.org/rust/parquet/file/properties/struct.WriterProperties.html,
+            only the fields: data_page_size_limit, dictionary_page_size_limit, data_page_row_count_limit, write_batch_size, max_row_group_size are supported.
+        :error_on_type_mismatch: specify if merge will return error if data types are mismatching :default = True
+        :return: the metrics from delete
+
+        Examples:
+
+        Update some row values with SQL predicate
```
wjones127, line 477: For clarity, we should probably show the equivalent SQL:
```suggestion
        Update some row values with SQL predicate. This is equivalent to
        ``UPDATE table SET deleted = true WHERE id = '5'``
```

python/deltalake/table.py
```diff
@@ -456,6 +456,50 @@ def vacuum(
...
+
+        >>> from deltalake import DeltaTable
+        >>> dt = DeltaTable("tmp")
+        >>> dt.update(predicate="id = '5'",
+        ...           updates = {
+        ...             "deleted": True,
+        ...             }
+        ...         )
+
+        Update all row values
```
wjones127, line 487:
```suggestion
        Update all row values. This is equivalent to
        ``UPDATE table SET id = concat(id, '_old')``.
```

python/docs/source/usage.rst
```diff
@@ -484,6 +484,53 @@ the data passed to it differs from the existing table's schema. If you wish to
...
+Updating Delta Tables
+---------------------
+
+.. py:currentmodule:: deltalake.table
+
+Row values in an existing delta table can be updated with the :meth:`DeltaTable.update` command. A update
+dictionary has to be passed, where they key is the column you wish to update, and the value is a
+Expression in string format.
+
+Update all the rows for the column "deleted" to the value True.
```
wjones127, line 496: TBH, I don't love this example, since the values might confuse a user who doesn't know the different between this and row deletion. Could you change the column name `deleted` to something else? Perhaps `processed`?

wjones127 (changes requested):
Just some minor suggestions on the docs

wjones127:
BTW there's some Rust failures which can be fixed if you rebase / update your branch.

wjones127 (approved):

PR #1697, opened 2023-10-03
Dictionary type support

alexwilcoxson-rel:
# Description
* Adds support for converting Arrow to/from Delta schemas when using Arrow Dictionary types.
* Encodes the Arrow type as Delta column metadata so the Arrow types that don't have direct analog in Delta, such as Dictionary, are used when converting back to Arrow.

[Slack thread](https://delta-users.slack.com/archives/C013LCAEB98/p1695055237800099)

### DIFF
--- a/rust/src/schema/arrow_convert.rs
+++ b/rust/src/schema/arrow_convert.rs
@@ -8,6 +8,7 @@ use arrow::datatypes::{
 use arrow::error::ArrowError;
 use lazy_static::lazy_static;
 use regex::Regex;
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::sync::Arc;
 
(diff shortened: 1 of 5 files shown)


rust/src/schema/arrow_convert.rs
```diff
@@ -8,9 +8,13 @@ use arrow::datatypes::{
 use arrow::error::ArrowError;
 use lazy_static::lazy_static;
 use regex::Regex;
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::sync::Arc;
 
+/// Column metadata that stores JSON serialized Arrow DataType
+pub const ARROW_TYPE_METADATA_KEY: &str = "ARROW:schema";
```
alexwilcoxson-rel, line 16: this is the same key that parquet arrow writer uses, could change to a `delta.` prefixed key but wanted to get buy in as those are reserved.

alexwilcoxson-rel:
I decided to go for an opt-in approach to adding the metadata when converting from an arrow schema for now.

I wasn't comfortable adding it to the TryFrom impl given how the python tests were failing. I'd want someone with more experience to dig into that. I could open another issue to expose this functionality via Python API.

rust/src/schema/arrow_convert.rs
wjones127: Hmm if we are going to use that key, could we instead make it at the schema level (not the field level) and use the same format as is in Parquet?

As described [here](https://arrow.apache.org/docs/cpp/parquet.html#roundtripping-arrow-types):

> The Arrow schema is serialized as a [Arrow IPC](https://arrow.apache.org/docs/format/Columnar.html#format-ipc) schema message, then base64-encoded and stored under the ARROW:schema metadata key in the Parquet file metadata.

alexwilcoxson-rel: I could change the key name.

I went this route because I don't see metadata on delta's SchemaTypeStruct. And used JSON as the metadata values on Field are serde_json::Value's

alexwilcoxson-rel:
this was deprioritized on my end and would need to be revisited as this PR is out of date

PR #1705, opened 2023-10-06
fix: merge operation with string predicates

### LINKED ISSUE #1699: String predicates do not work for Merge
# Environment

**Delta-rs version**: latest

**Binding**: rust

***
# Bug

**What happened**:
When string predicates are used and the predicate contains a reference to the source table, the merge operation fails

**How to reproduce it**:

**More details**:
`parse_prediacte_expression` only has the schema of the target table and is unable to lookup the columns from the source table.
To fix this allow passing the joined schema of source and target with the source prefix applied.

Blajda:
# Description
Fixes an issue when users use string predicates with the merge operation.

Parsing a string predicate did not properly handle table references and would always assume a bare table with a table name of the empty string. Now the qualifier is `None` however a `DFSchema` with qualifiers can be supplied where it makes sense.

Now users must provide source and target aliases whenever both sides share a column name otherwise the operation will error out.

Minor refactoring of the expression parser was also done and allowed using of case expressions.

# Related Issue(s)
- closes #1699


ion-elgreco:
@Blajda I tried the fix, but it's not yet fully working, or this is another issue, I am getting these errors:

`Schema error: Schema contains qualified field name source.id and unqualified field name id which would be ambiguous`

Blajda:
@ion-elgreco This change now requires columns references to be unambiguous. The previous behavior of `source.id = id` won't work anymore and aligns with spark.
I added a new method called `with_target_alias` which allow specifying a table name for columns from the underlying table. You'll will need to update your predicates with the target alias. E.g `source.id = target.id`
This only needs to be done when source and target share columns with the same name.

ion-elgreco:
> @ion-elgreco This change now requires columns references to be unambiguous. The previous behavior of `source.id = id` won't work anymore and aligns with spark. I added a new method called `with_target_alias` which allow specifying a table name for columns from the underlying table. You'll will need to update your predicates with the target alias. E.g `source.id = target.id` This only needs to be done when source and target share columns with the same name.

I see, that makes sense! Thanks for explaining, let me try it out then with also provide target_alias

ion-elgreco:
@Blajda I've check it, I can update however it's not fully the way I was expecting it.

In the updates if I add no alias for the target, it properly updates everything.
```python
dt.merge(
    source=source_table, source_alias='source', target_alias='target', predicate="target.id = source.id"
).when_matched_update({"price": "source.price", "sold": "source.sold"}).execute()
```
However, I was expecting you had to do this.
```python
dt.merge(
    source=source_table, source_alias='source', target_alias='target', predicate="target.id = source.id"
).when_matched_update({"target.price": "source.price", "target.sold": "source.sold"}).execute()
```

Blajda:
> @Blajda I've checked it, I can update however it's not fully the way I was expecting it.
> 
> In the updates if I add no alias for the target, it properly updates everything.

Yes that's expected only the predicates themselves require require the table qualifiers so your first example should be used.

ion-elgreco:
> > @Blajda I've checked it, I can update however it's not fully the way I was expecting it.
> > 
> > In the updates if I add no alias for the target, it properly updates everything.
> 
> Yes that's expected only the predicates themselves require require the table qualifiers so your first example should be used.
> 
> 

From what I can see in the delta-spark API, they also have examples where the updates or sets are using the target alias as well, however not always, so I guess that implies that it will work either way.

I think it makes sense if you add a target alias in the updates or sets it also would work.

Blajda:
> From what I can see in the delta-spark API, they also have examples where the updates or sets are using the target alias as well, however not always, so I guess that implies that it will work either way.

Good catch. I've enabled that functionality and added tests to ensure the behavior is the same.

ion-elgreco:
> > From what I can see in the delta-spark API, they also have examples where the updates or sets are using the target alias as well, however not always, so I guess that implies that it will work either way.
> 
> Good catch. I've enabled that functionality and added tests to ensure the behavior is the same.
> 

Nice work! :)

Blajda:
@wjones127 please look over this PR when you are able.

rust/src/operations/merge.rs
```diff
@@ -452,28 +457,71 @@ struct MergeOperation {
...
+                        } => {
+                            if table.eq(alias) {
+                                Column {
+                                    relation: Some(r),
+                                    name,
+                                }
+                            } else {
+                                return Err(DeltaTableError::Generic(
+                                    "Column must reference column in Delta table".into(),
+                                ));
```
wjones127, line 493: Given the feedback from Ion, seems like this might be confusing. Perhaps we should give a message with a hint:

```suggestion
                                return Err(DeltaTableError::Generic(
                                    format!("Table alias '{table}' in column reference '{table}.{name}' unknown. Hint: You must reference the Delta Table with alias '{alias}'.")
                                ));
```

wjones127 (dismissed):
All looks reasonable, though I think this one error message could be improved.

wjones127 (approved):

PR #1709, opened 2023-10-08
feat: improve read performance by 7x with prebuffer

ion-elgreco:
# Description
Enable prebuffer in the pyarrow.dataset.ParquetFragmentScanOptions. Relevant PR in [Arrow ](https://github.com/apache/arrow/pull/37854)repo, where they changed it to be default behavior. However, this won't be the case for older versions for PyArrow, so we need to set it to True.: 

It improves read speed by 6-7x on Azure in one dataset that I have.

Before:
1min 4s ± 3.48 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

After:
8.99 s ± 786 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Related Issue(s)

Closes #https://github.com/delta-io/delta-rs/issues/1569

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -24,6 +24,7 @@ from pyarrow.dataset import (
     Expression,
     FileSystemDataset,
     ParquetFileFormat,
+    ParquetFragmentScanOptions,
     ParquetReadOptions,
 )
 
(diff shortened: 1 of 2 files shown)


rtyler (approved):
There is a memory tradeoff, but I think upstream defaulting indicates it's well worth the tradeoff for a default behavior.

Going to approve, thanks for another solid improvement @ion-elgreco

Issue #1561, opened 2023-07-26
Adding version number to .history()


jakubkovac:
# It would be helpful to have version number alongside each entry in the .history() method

**Use Case**
similarly to delta.DeltaTable.forPath(...).history() that returns a sparkDataFrame with version number, it would be useful to have this directly in the history method.

Currently I'm using the spark history to detect the specific version number corresponding to a datetime, since version is the only argument with polars.read_delta.

If I understood delta correctly, history might not have the complete history of versions, because of retention period - but available history is probably coming from the logs in the _delta_log folder right?

Is there a better way to get version numbers and their corresponding timestamp through deltalake library?

Hanspagh:
I am facing the same issue

ion-elgreco:
Looks pretty straightforward, I could look into it soon.

PR #1710, opened 2023-10-09
feat: add version number in `.history()` and display in reversed chronological order

### LINKED ISSUE #1680: Documentation for .history says results should be in reverse chronological order, but results are returned in chronological order
# Environment
MacOS

**Delta-rs version**:
delta-rs.0.15.0

**Binding**: 
Python

***
# Bug
History is returned in chronological order while documentation says the opposite 

**What happened**:

**How to reproduce it**:
df = pd.DataFrame({"x": [1, 2, 3, 4]})

write_deltalake("table.delta", df, mode="append")
write_deltalake("table.delta", df, mode="append")
write_deltalake("table.delta", df, mode="append")
write_deltalake("table.delta", df, mode="append")

**More details**:
DeltaTable("table.delta").history()

```
[{'timestamp': 1695978312893,
  'operation': 'CREATE TABLE',
  'operationParameters': {'metadata': '{"configuration":{},"created_time":1695978312893,"description":null,"format":{"options":{},"provider":"parquet"},"id":"9adb7a7d-e4b6-4f1a-b325-c60f45685bd9","name":null,"partition_columns":[],"schema":{"fields":[{"metadata":{},"name":"x","nullable":true,"type":"long"}],"type":"struct"}}',
   'mode': 'ErrorIfExists',
   'protocol': '{"minReaderVersion":1,"minWriterVersion":1}',
   'location': 'file:///Users/hans.pagh/Code/pippi/pippi-base/sovs.delta'},
  'clientVersion': 'delta-rs.0.15.0'},
 {'timestamp': 1695978319475,
  'operation': 'WRITE',
  'operationParameters': {'mode': 'Append', 'partitionBy': '[]'},
  'clientVersion': 'delta-rs.0.15.0'},
 {'timestamp': 1695978899613,
  'operation': 'WRITE',
  'operationParameters': {'mode': 'Append', 'partitionBy': '[]'},
  'clientVersion': 'delta-rs.0.15.0'}]
```

ion-elgreco:
# Description
Adds the version number to each commit info.

# Related Issue(s)

- Closes https://github.com/delta-io/delta-rs/issues/1561
- Closes #1680


rtyler (dismissed):

python/deltalake/table.py
```diff
@@ -428,10 +428,19 @@ def history(self, limit: Optional[int] = None) -> List[Dict[str, Any]]:
         :param limit: the commit info limit to return
         :return: list of the commit infos registered in the transaction log
         """
-        return [
-            json.loads(commit_info_raw)
-            for commit_info_raw in self._table.history(limit)
-        ]
+        if (limit is None) or (self.version() + 1 <= limit):
+            start = 0
```
wjones127, line 432: It seems like this could be incorrect if `limit` is None and there are log files in the beginning that were deleted, right?

Although IIRC, our history function can't look past the checkpoints at the moment?

ion-elgreco: > It seems like this could be incorrect if `limit` is None and there are log files in the beginning that were deleted, right?
> 
> Although IIRC, our history function can't look past the checkpoints at the moment?

Hm I wasn't aware of this possibility. Then this won't work for that case. I'll check how I can make it work otherwise, maybe it should be reverse order and counting down

ion-elgreco: @wjones127 @rtyler I added a reverse countdown, this actually also fixes immediately: https://github.com/delta-io/delta-rs/issues/1680

However! It relies on using `DeltaTable.version()` which means in these two cases the count won't be correct:

- A user does `DeltaTable.load_version( old_ver )`, then this version is used to count downwards
- Someone writes to the deltalake, without using `write_deltalake `and passing the `DeltaTable `object. 

In any of these two cases the highest version number of the table can't be fetched from the API. I think this can't be fixed until we have a function that fetches the highest available table version.

ion-elgreco: @rtyler @wjones127 the counting will be correct now for all cases and in reversed order. 🕺

wjones127 (approved):
Looks better. Thanks for looking into that.

Hanspagh:
Fwi, this is a breaking change for user depending on the current order of the history.

ion-elgreco:
> Fwi, this is a breaking change for user depending on the current order of the history. 

We should make a mention I guess in the changelog that this could break, but it's actually easily resolved with just doing `reversed(DeltaTable.history())`

Hanspagh:
> > Fwi, this is a breaking change for user depending on the current order of the history.
> 
> We should make a mention I guess in the changelog that this could be a breaking change, but it's actually easily resolved with just doing `reversed(DeltaTable.history())`

Agreed, it's is easy to fix, hopefully people will read the release notes before updating.

Issue #1708, opened 2023-10-07
Address various issues in mkdocs for Python


wjones127:
# Description

We want to use mkdocs instead, but need to address some issues in the draft.

## Issue 1: cross reference aren't working

```
WARNING -  mkdocs_autorefs: usage/index.md: Could not find cross-reference target '[DeltaTable]'
WARNING -  mkdocs_autorefs: usage/examining-table.md: Could not find cross-reference target '[DeltaTable.metadata()]'
WARNING -  mkdocs_autorefs: usage/examining-table.md: Could not find cross-reference target '[DeltaTable.schema]'
WARNING -  mkdocs_autorefs: usage/examining-table.md: Could not find cross-reference target '[deltalake.schema.Schema.from_json()]'
WARNING -  mkdocs_autorefs: usage/examining-table.md: Could not find cross-reference target '[deltalake.schema.Schema.to_pyarrow()]'
```

## Issue 2: API Reference table of contents is too flat for schema types

![image](https://github.com/delta-io/delta-rs/assets/5488879/2643d75a-832a-4fbf-b71f-ab995b57ffd2)

One you get the schema module, it stops nesting the methods

## Issue 3: API reference contains random dunder methods for schema types

It shows all the `__eq__()`, `__ge__()`, etc. methods for `PrimitiveType`, for example.

## Issue 4: Old sphinx directives in docstrings need to be replaced

We use Sphinx directives like `.. deprecated::` in the docstrings. They need to be replaced with the mkdocs alternative.

## Issue 5: Examples don't render properly

<img width="755" alt="Screenshot 2023-10-07 at 2 22 40 PM" src="https://github.com/delta-io/delta-rs/assets/5488879/4b79dc1f-3398-4d6b-8c50-3bdba65f5d06">

## Issue 6: docstring fields aren't being parsed

Things like `:param` and `:returns` are just shown raw.

# TODO

* [ ] Fix cross references in mkdocs
* [ ] Fix API reference table of contents hierarchy in mkdocs
* [ ] Hide dunder methods in mkdocs API reference
* [ ] Replace old sphinx directives with mkdocs equivalents in docstrings
* [ ] Make docstring examples display properly in mkdocs
* [ ] Format docstrings so mkdocs can understand parameters and return types.

r3stl355:
I'll track my progress here

- [ ] Fix API reference table of contents hierarchy in mkdocs
- [x] Hide dunder methods in mkdocs API reference
[...]

PR #1724, opened 2023-10-13
docs: initial attempt to convert docs to use mkdocs

r3stl355:
# Description
I fixed most of the tasks in the issue #1708 but did not make all the changes to docstrings. Submitting this to get early feedback before making larger changes. Please review `API Reference -> Delta Table` for most changes. 

I also suggest changing theme from `readthedocs` to `mkdocs` - both built in but latter looks sleeker. Just change the value in `mkdocs.yaml` to see.

# Related Issue(s)
#1708


r3stl355:
Please ignore the deleted code when reviewing this - looks like i messed up my merge, focus on docs changes only. I'll make sure all is properly merged before I submit the full PR

PR #1725, opened 2023-10-13
refactor!: remove a layer of lifetimes from PartitionFilter

### LINKED ISSUE #1501: Make `PartitionFilters` easier to use
# Description

**Use Case**

For example, I want to convert `HashMap<String, String>` into a `Vec<PartitionField<_>>` to supply to the `OptimizeBuilder::with_filters` method, where each entry is used as a `PartitionValue::Equal` comparison. In it's current state, due to the explicit lifetime on `PartitionField`, it is very hard to achieve this. 

When talking about it in the `#delta-rs` channel on Slack, one suggestion is to use `Cow`, `Arc`, or `Box` values instead. 

**Related Issue(s)**

cmackenzie1:
# Description
This commit removes a bunch of lifetime restrictions on the `PartitionFilter` and `PartitionFilterValue` classes to make them easier to use. While the original discussion in Slack and #1501 made mention of using a reference type, there doesn't seem to a need for it. A particular instance of a `PartitionFilter` is created once and just borrowed and read for the remainder of its life.

Functions, when necessary continue to accept the non-container types (i.e, `&str` and `&[&str]`), allowing their containerized counterparts to continue working with them without needing to borrow or clone the containers (i.e, `String` and `Vec<String>`).

# Related Issue(s)
- resolves #1501 

# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -193,18 +193,17 @@ impl RawDeltaTable {
         &self,
         partitions_filters: Vec<(&str, &str, PartitionFilterValue)>,
     ) -> PyResult<Vec<String>> {
-        let partition_filters: Result<Vec<PartitionFilter<&str>>, DeltaTableError> =
-            partitions_filters
-                .into_iter()
-                .map(|filter| match filter {
-                    (key, op, PartitionFilterValue::Single(v)) => {
-                        PartitionFilter::try_from((key, op, v))
-                    }
-                    (key, op, PartitionFilterValue::Multiple(v)) => {
-                        PartitionFilter::try_from((key, op, v))
-                    }
-                })
-                .collect();
+        let partition_filters: Result<Vec<PartitionFilter>, DeltaTableError> = partitions_filters
+            .into_iter()
+            .map(|filter| match filter {
+                (key, op, PartitionFilterValue::Single(v)) => {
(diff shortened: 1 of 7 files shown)


cmackenzie1:
Also took the liberty to remove `<T>` as partition keys and values are always serialized as a string anyway. https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization

cmackenzie1:
For an example on the current difficulty on using these.

```rust
let optimization = match config.lookback_window {
    Some(lookback_window) => {
        let partition_filters = Box::new(date_partition_filters(lookback_window));

        OptimizeBuilder::new(delta_table.object_store(), delta_table.state)
            .with_max_concurrent_tasks(config.concurrency)
            .with_target_size(config.target_bytes.try_into().unwrap())
            .with_min_commit_interval(config.commit_interval)
            .with_filters(&partition_filters) // <--- partition_filters does not live long enough
            .into_future()
    }
    None => OptimizeBuilder::new(delta_table.object_store(), delta_table.state)
        .with_max_concurrent_tasks(config.concurrency)
        .with_target_size(config.target_bytes.try_into().unwrap())
        .with_min_commit_interval(config.commit_interval)
        .into_future(),
};
```

wjones127 (approved):
This seems reasonable.

Issue #1726, opened 2023-10-14
VACUUM doesn't create a commit in transaction log


ion-elgreco:
When you vacuum a table with delta-rs, it  doesn't create a commit in the delta transaction log, stating the VACUUM operation.

wjones127:
Does it do that in Spark? I wouldn't expect it to create a transaction.

ion-elgreco:
@wjones127 apparently they do since this PR: https://github.com/delta-io/delta/pull/1552

ion-elgreco:
@wjones127 I checked my tables written with databricks, and they all have the VACUUM START/END commits

![image](https://github.com/delta-io/delta-rs/assets/15728914/d95ebf87-0030-4d71-a362-56675eecc2cc)

PR #1728, opened 2023-10-14
feat: add VACUUM operation as commit in transaction log

ion-elgreco:
# Description
Since deltalake 2.3.0 VACUUM operations are added in the transaction log, with `VACUUM START`, and `VACUUM END`. This PR will add these changes to the transaction log now for Delta-RS 🕺:

Examples of the commits:

```
{"commitInfo":{"timestamp":1697298818104,"operation":"VACUUM START","operationParameters":{"defaultRetentionMillis":"604800000","retentionCheckEnabled":"false","specifiedRetentionMillis":"0"},"clientVersion":"delta-rs.0.17.0","operationMetrics":{"numFilesToDelete":7, "sizeOfDataToDelete":1000}}}
```

```
{"commitInfo":{"timestamp":1697298818105,"operation":"VACUUM END","operationParameters":{"status":"COMPLETED"},"clientVersion":"delta-rs.0.17.0","operationMetrics":{"numDeletedFiles":7,"numVacuumedDirectories":0}}}
```

# Related Issue(s)

- closes https://github.com/delta-io/delta-rs/issues/1726

# Documentation

PR in delta repo: https://github.com/delta-io/delta/pull/1552


rust/src/operations/vacuum.rs
```diff
@@ -214,18 +244,102 @@ impl std::future::IntoFuture for VacuumBuilder {
...
+            retention_check_enabled: self.retention_check_enabled,
+            specified_retention_millis: self.specified_retention_millis,
+            default_retention_millis: self.default_retention_millis,
+        };
+
+        let end_operation = DeltaOperation::VacuumEnd {
+            status: String::from("COMPLETED"),
+        };
+
+        // let mut total_size = 0;
```
wjones127, line 279: I think if you want to collect this information, we should keep around the [ObjectMeta](https://docs.rs/object_store/latest/object_store/struct.ObjectMeta.html) here:

https://github.com/delta-io/delta-rs/blob/4226983f595cff004e905c3fde3c933aaae848d0/rust/src/operations/vacuum.rs#L203

instead of just getting the path. I think that should give you all the info you need, right?

ion-elgreco: I'll take a look tonight! Thanks 💯

ion-elgreco: I updated it now, one thing I noticed before is that we don't vacuum directories. I noticed it while vacuum a table with old partition folders. So I just keep the number of vacuumdirectories at 0.

ion-elgreco:
@rtyler or @wjones127 could you take a look at the last changes if it's ok now?

wjones127 (approved):
I wish there was a description somewhere of how this operation is meant to be used.

PR #1730, opened 2023-10-15
feat(python): expose FSCK (repair) operation 

### LINKED ISSUE #1727: Expose FSCK (Repair table) operation into Python bindings
# Description
FSCK can be a useful operation to fix corrupted or by accident deleted files, to add them as remove actions in the delta table. It's already available in Rust, we just need to expose it into the Python Api.

ion-elgreco:
# Description
This PR exposes the FSCK operation as a `repair` method under the `DeltaTable `class.

# Related Issue(s)

- closes #1727


python/deltalake/table.py
```diff
@@ -746,6 +746,31 @@ def delete(self, predicate: Optional[str] = None) -> Dict[str, Any]:
         metrics = self._table.delete(predicate)
         return json.loads(metrics)
 
+    def repair(self, dry_run: bool = True) -> Dict[str, Any]:
```
wjones127, line 749: I wonder if we should make `dry_run` `False` by default. It's not a dangerous operation right?

ion-elgreco: It's not a dangerous operation, I just set it to True for consistency because with Vacuum we also do dry runs first.

wjones127: Let's make it `True` by default. `Vacuum` is false because it is actually dangerous; you could accidentally delete files you meant to keep. This one is reversible (IIUC) using restore.

ion-elgreco: I think you meant the other way around. :) False by default 😄

ion-elgreco: @wjones127 can we merge this?

python/deltalake/table.py
```diff
@@ -826,6 +826,31 @@ def delete(self, predicate: Optional[str] = None) -> Dict[str, Any]:
         metrics = self._table.delete(predicate)
         return json.loads(metrics)
 
+    def repair(self, dry_run: bool = False) -> Dict[str, Any]:
+        """It repairs the Delta Table by auditting active files that do not exist in the underlying
```
wjones127, line 830:
```suggestion
        """Repair the Delta Table by auditing active files that do not exist in the underlying
```

wjones127 (approved):

PR #1733, opened 2023-10-17
fix: reorder encode_partition_value() checks and add tests

ldacey:
# Description
The `isinstance(val, datetime)` check was after `isinstance(val, date)` which meant that it was never found. I added a test for each encoding type.

### DIFF
--- a/python/tests/test_table_read.py
+++ b/python/tests/test_table_read.py
@@ -697,6 +697,8 @@ def test_issue_1653_filter_bool_partition(tmp_path: Path):
         )
         == 1
     )
+
+
 @pytest.mark.parametrize(
     "input_value, expected",
     [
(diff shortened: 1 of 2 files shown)


wjones127:
```python
>>> from datetime import date, datetime
>>> isinstance(datetime(2021, 1, 1), date)
True
```
TIL 🤯

python/tests/test_table_read.py
wjones127, line 1: Thank you for thoroughly testing this 🙌

wjones127:
@ldacey if you run `make format` in the Python directory it should fix all formatting issues.

ldacey:
> @ldacey if you run `make format` in the Python directory it should fix all formatting issues.

Nice - thanks. I was struggling with that because of my new neovim setup which I am still tinkering with.

roeap (approved):

PR #1741, opened 2023-10-19
chore: update datafusion

roeap:
# Description

Updates arrow and datafusion dependencies to latest.

# Related Issue(s)

# Documentation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,23 +15,23 @@ debug = "line-tables-only"
 
 [workspace.dependencies]
 # arrow
-arrow = { version = "46" }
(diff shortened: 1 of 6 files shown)


rtyler (approved):

roeap:
as discussed with @wjones127, I updated the zorder tests to no longer validate the binary representation of the z-order keys, as it may change between arrow releases due to changes in the `RowEncoder`. as it was a bit of a challenge to mentally predict the byte representation of the encoded data and what would be the right resulting order, we went for crafting a few examples and seeing that the generated ordering is reasonable ...

rtyler (approved):

PR #1743, opened 2023-10-20
fix: relax `pyarrow` pin

### LINKED ISSUE #1701: Unlock pyarrow patches in python dependencies
# Description

Python binding of delta-rs is closing pyarrow to version 12. However, the line in the `pyproject.toml` that sets the bind is formulated as follows:
https://github.com/delta-io/delta-rs/blob/4da7d66d06985d386e61d3fb124cad6680594dcc/python/pyproject.toml#L20
This allows installing only pyarrow==12.0.0, and patches (such as 12.0.1) are outside the bounds specified.

I would formulate the line as follows:
either
```py
    "pyarrow>=8,<13",
```

or, to be safer
```py
    "pyarrow>=8,<12.1",
```

dhirschfeld:
# Description
Relaxes the `pyarrow` pin to allow *any* version 12 as it was likely unintended to *only* allow 12.0.0 and no other 12.* versions.

# Related Issue(s)

Resolves #1701

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11"
 ]
 dependencies = [
-    "pyarrow>=8,<=12",
+    "pyarrow>=8,<13",
     'typing-extensions;python_version<"3.8"',
 ]
 


dhirschfeld:
Cheers @wjones127 - I was just browsing the conventional commit docs 😄 

I know of the convention, but have never used it myself.

wjones127 (approved):

PR #1747, opened 2023-10-20
feat: honor appendOnly table config

### LINKED ISSUE #352: Honor appendOnly table config in table writes
# Environment

**Delta-rs version**: all

**Binding**: all

# Bug

Currently we are not taking `appendOnly` table config into account when performing delta table writes. This config needs to be honored at two levels. In the low level delta transaction commit code path, we need to throw an error if a remove file action is included in a transaction. In the high level delta write API (not available in delta-rs yet), we need to throw an error if user query will result in in place data update.

junjunjd:
# Description
Throw an error if a transaction includes Remove action with data change but the Delta Table is append-only.

# Related Issue(s)
- closes #352

### DIFF
--- a/rust/src/operations/transaction/mod.rs
+++ b/rust/src/operations/transaction/mod.rs
@@ -48,6 +48,11 @@ pub enum TransactionError {
     /// Error returned when maximum number of commit trioals is exceeded
     #[error("Failed to commit transaction: {0}")]
     MaxCommitAttempts(i32),
+    /// The transaction includes Remove action with data change but Delta table is append-only
+    #[error(
+        "The transaction includes Remove action with data change but Delta table is append-only"
+    )]
+    DeltaTableAppendOnly,
 }
 
 impl From<TransactionError> for DeltaTableError {
@@ -68,9 +73,18 @@ impl From<TransactionError> for DeltaTableError {
 // Convert actions to their json representation
 fn log_entry_from_actions<'a>(
     actions: impl IntoIterator<Item = &'a Action>,
+    read_snapshot: &DeltaTableState,
 ) -> Result<String, TransactionError> {
+    let append_only = read_snapshot.table_config().append_only();
     let mut jsons = Vec::<String>::new();
     for action in actions {
+        if append_only {
+            if let Action::remove(remove) = action {
+                if remove.data_change {
+                    return Err(TransactionError::DeltaTableAppendOnly);
+                }
+            }
+        }
         let json = serde_json::to_string(action)
             .map_err(|e| TransactionError::SerializeLogJson { json_err: e })?;
         jsons.push(json);
@@ -81,6 +95,7 @@ fn log_entry_from_actions<'a>(
 pub(crate) fn get_commit_bytes(
     operation: &DeltaOperation,
     actions: &Vec<Action>,
+    read_snapshot: &DeltaTableState,
     app_metadata: Option<Map<String, Value>>,
 ) -> Result<bytes::Bytes, TransactionError> {
     if !actions.iter().any(|a| matches!(a, Action::commitInfo(..))) {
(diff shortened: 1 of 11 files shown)


rust/src/operations/transaction/mod.rs
```diff
@@ -68,9 +71,25 @@ impl From<TransactionError> for DeltaTableError {
 // Convert actions to their json representation
 fn log_entry_from_actions<'a>(
     actions: impl IntoIterator<Item = &'a Action>,
+    read_snapshot: &DeltaTableState,
 ) -> Result<String, TransactionError> {
+    let append_only = read_snapshot.current_metadata().map_or(false, |meta| {
+        meta.configuration
+            .get("delta.appendOnly")
```
rtyler, line 78:
```suggestion
            .get(DeltaConfigKey::AppendOnly)
```

Since we already have this constant defined elsewhere, it would make sense to me to just refer to it

rust/src/operations/transaction/mod.rs
```diff
@@ -232,13 +259,36 @@ mod tests {
...
         // writes every action to a line
         assert_eq!(actions.len(), lines.len())
     }
 
+    fn remove_action_exists_when_delta_table_is_append_only(
+        data_change: bool,
+    ) -> Result<String, TransactionError> {
+        let remove = create_remove_action("test_append_only", data_change);
+        let mut actions = init_table_actions(Some(HashMap::from([(
+            "delta.appendOnly".to_string(),
```
rtyler, line 275:
```suggestion
            DeltaConfigKey::ApendOnly,
```

rust/src/operations/transaction/mod.rs
roeap: I would go a bit further - hidden ion the depth of our code, we have a `TableConfig` type, that abstracts reading table configuration. Its accessible via `table_config()` on the snapshot.

roeap (changes requested):
Thanks for taking care of this @junjunjd!

Looking great, I especially appreciate the test coverage!

There are a few smaller things we may consider. I believe the `commit` or `commit_with_retry` function may be a little bit more fitting place to do the check?

On amore general level - and this is completely optional for this PR. For the different validations we have been creating `*Checker` structs - like the `DataChecker` or the `ConflictChecker`. I am currently working on broader protocol support, and planned to introduce something like a `PreconditionChecker` which handles more pre-commit checks.

Just for append-only this may be overkill, so I leave it up to you if you want to adopt this here.

rust/src/operations/transaction/mod.rs
junjunjd: Updated.

junjunjd:
> There are a few smaller things we may consider. I believe the commit or commit_with_retry function may be a little bit more fitting place to do the check?
@roeap I agree logically `commit` or `commit_with_retry function` would be the right place to do the check. However, the code that iterates the Action vector exists in `log_entry_from_actions`. `commit_with_retry` calls `log_entry_from_actions` (and a bunch of intermediate functions) to prepare the commit. I think it makes sense to put the check in `log_entry_from_actions` so that the Action vector is only iterated once.

>  I am currently working on broader protocol support, and planned to introduce something like a PreconditionChecker which handles more pre-commit checks - e.g. writer version compatibility.
This is a good idea to build the high-level checker. For append-only, we can add high-level checking to delete, merge and update operations to return error immediately. I have created a follow-up issue to add high-level checking https://github.com/delta-io/delta-rs/issues/1759.

junjunjd:
@rtyler @roeap @wjones127 I have addressed the comments. This MR is ready for final review.

roeap:
> I think it makes sense to put the check in log_entry_from_actions so that the Action vector is only iterated once.

i see your point, however we discussed with databricks and even in their environments commits rarely get larger then low 10s of actions (with some extreme outliers like shallow copy). So all in all i would guess the cost of one traversal is negligible compared to all the io we just did writing files and the potentially many traversals  in conflict checking ...

That said, we can revisit this when we create a higher level helper.

roeap (commented):
LGTM! thanks for this work!!

there is one more thing i realised. could we update the readme to refelct the feature support in the tables towards the end?

junjunjd:
> there is one more thing i realised. could we update the readme to refelct the feature support in the tables towards the end?
@roeap README is updated. Thanks for the review.

roeap (approved):

Issue #1740, opened 2023-10-19
update don't seems to be working


djouallah:
# Environment

0.12 , google colab

using this code 

```
import pandas as pd
from deltalake.writer import write_deltalake
df = pd.DataFrame(
    {
        "a": [1, 2, 3, 4, 5],
        "fruits": ["banana", "orange", "mango", "apple", "mandarin"],
    }
)
write_deltalake('./fruits', df,mode='overwrite')
from deltalake import DeltaTable
fruits_dl   = DeltaTable('./fruits')
fruits_dl.update(predicate="a = 5",updates={"fruits": "banana"})
```
I get this error

```
DeltaError                                Traceback (most recent call last)
[<ipython-input-28-8d483d464424>](https://localhost:8080/#) in <cell line: 1>()
----> 1 fruits_dl.update(predicate="a = 5",updates={"fruits": "banana"})

[/usr/local/lib/python3.10/dist-packages/deltalake/table.py](https://localhost:8080/#) in update(self, updates, predicate, writer_properties, error_on_type_mismatch)
    481         """
    482 
--> 483         metrics = self._table.update(
    484             updates, predicate, writer_properties, safe_cast=not error_on_type_mismatch
    485         )

DeltaError: Generic DeltaTable error: Schema error: No field named banana. Valid fields are a, fruits.

```

wjones127:
The values of the `updates` parameter are SQL expressions. `banana` without any quotes is an identifier, which is why it says it's not a valid column. If you wrap it in single quotes, it will work as expected:
```python
>>> fruits_dl.update(predicate="a = 5",updates={"fruits": "'banana'"})
{'num_added_files': 1, 'num_removed_files': 1, 'num_updated_rows': 1, 'num_copied_rows': 4, 'execution_time_ms': 26, 'scan_time_ms': 17}
>>> fruits_dl.to_pandas()
   a  fruits
0  1  banana
1  2  orange
2  3   mango
3  4   apple
4  5  banana
```

wjones127:
It does feel like it would be nice to be able to pass python objects here though. Something like:

```python
table.update(
    predicate="a = 5",
    updates={
        "fruits": "banana",
        "count": 2,
        "last_updated": datetime.now()
    }
)
```

But this should be provided by a different parameter. Otherwise we could end up interpreting the SQL strings as straight string values, and that wouldn't be great. cc @ion-elgreco

ion-elgreco:
[...]

PR #1749, opened 2023-10-21
feat(python): allow python objects to be passed as new values in `.update()`

ion-elgreco:
# Description
A user can now add a new_values dictionary that contains python objects as a value. 

Some weird behavior's I noticed, probably related to datafusion, updating a timestamp column has to be done by providing a unix timestamp in microseconds. I personally find this very confusing, I was expecting to be able to pass "2012-10-01" for example in the updates.

Another weird behaviour is with list of string columns. I can pass `{"list_of_string_col":"[1,2,3]"}` or `{"list_of_string_col":"['1','2','3']"}` and both will work. I expect the first one to raise an exception on invalid datatypes. Combined datatypes `"[1,2,'3']"`  luckily do raise an error by  datafusion.

# Related Issue(s)

- closes https://github.com/delta-io/delta-rs/issues/1740


python/deltalake/table.py
```diff
@@ -464,7 +469,7 @@ def update(
         >>> dt = DeltaTable("tmp")
         >>> dt.update(predicate="id = '5'",
         ...           updates = {
-        ...             "deleted": True,
+        ...             "deleted": 'True',
```
wjones127, line 472: hmm we should add a doctest CI job. 🤔 https://github.com/delta-io/delta-rs/issues/1783

python/deltalake/table.py
```diff
@@ -473,15 +478,48 @@ def update(
...
         ...             "id": "concat(id, '_old')"
         ...             }
         ...         )
 
+        Update some row values with python object. This is equivalent to
+        ``UPDATE table SET price = 150.10 WHERE id = '5'``
+        >>> from deltalake import DeltaTable
+        >>> dt = DeltaTable("tmp")
+        >>> dt.update(predicate="id = '5'",
+        ...           updates = {
```
wjones127, line 491:
```suggestion
        ...           new_values = {
```

python/deltalake/table.py
```diff
@@ -473,15 +478,48 @@ def update(
         >>> from deltalake import DeltaTable
         >>> dt = DeltaTable("tmp")
         >>> dt.update(updates = {
-        ...             "deleted": True,
+        ...             "deleted": 'True',
         ...             "id": "concat(id, '_old')"
         ...             }
         ...         )
 
+        Update some row values with python object. This is equivalent to
```
wjones127, line 486:
```suggestion
        To use Python objects instead of SQL strings, use the `new_values` parameter
        instead of the `updates` parameter. For example, this is equivalent to
```

python/deltalake/table.py
```diff
@@ -473,15 +478,48 @@ def update(
...
+        ...           updates = {
+        ...             "price": 150.10,
+        ...             }
+        ...         )
         """
-
+        if updates is None and new_values is not None:
+            updates = {}
+            for key, value in new_values.items():
+                if isinstance(value, (int, float, bool, list)):
```
wjones127, line 499: Are we sure this works for `list`? I believe DataFusion uses PostgreSQL syntax, which uses `{}` and not `[]` delimiters for arrays: https://www.postgresql.org/docs/current/arrays.html
Regardless, we should make sure to have a unit test for all of these types.

python/deltalake/table.py
```diff
@@ -473,15 +478,48 @@ def update(
...
+                    value = str(value)
+                elif isinstance(value, str):
+                    value = f"'{value}'"
+                elif isinstance(value, datetime):
+                    value = str(
+                        int(value.timestamp() * 1000 * 1000)
+                    )  # convert to microseconds
+                updates[key] = value
```
wjones127, line 507: I think we should have an `else` clause that raises `TypeError` given it's probably an unsupported type, right?

python/deltalake/table.py
```diff
@@ -473,15 +478,48 @@ def update(
...
+        elif updates is not None and new_values is None:
+            pass
```
wjones127, line 509: Probably worth validating here that `updates` values are all strings. Can provide a helpful message saying "The values of the updates parameter must all be SQL strings. Got {}. Did you mean to use the new_values parameter?"

python/deltalake/table.py
wjones127: Huh I guess you do have a test. I'm somewhat surprised it works, but I'll take it.

wjones127 (changes requested):
Looks like a nice interface. I have some suggestions to make some of the parameter handling clearer.

python/deltalake/table.py
ion-elgreco: Yeah we should, then this bug would have been catched the first PR.

python/deltalake/table.py
ion-elgreco: There is also another weird quirk, I described it in the description of the PR.

python/deltalake/table.py
ion-elgreco: I agree, it's added now.

python/deltalake/table.py
ion-elgreco: Done :)

ion-elgreco:
@wjones127 I've integrated all your feedback in the code

python/deltalake/table.py
```diff
@@ -522,18 +526,63 @@ def update(
         ```
         dt = DeltaTable("tmp")
-        dt.update(predicate="id = '5'", updates = {"deleted": True})
+        dt.update(predicate="id = '5'", updates = {"deleted": 'True'})
         ```
 
-        Update all row values. This is equivalent to `UPDATE table SET id = concat(id, '_old')`.
+        Update all row values. This is equivalent to
+        ``UPDATE table SET id = concat(id, '_old')``.
```
wjones127, line 533:
```suggestion
        ``UPDATE table SET deleted = true, id = concat(id, '_old')``.
```

wjones127 (approved):
One small nit on the docs, otherwise looks good :)

ion-elgreco:
> One small nit on the docs, otherwise looks good :)

Your approval got dismissed with the commit :stuck_out_tongue:

wjones127 (approved):

franz101:
Am I correct this does not work with dictionaries?
```
data = pa.table({"x": [1, 2, 3], "y": [{"a":"a"}, {"a":"b"}, {"a":"c"}]},schema=dt.schema().to_pyarrow(False))
write_deltalake("tmp", data, mode="overwrite", overwrite_schema=True)
dt.update(predicate="id = 2",updates={"y": {"a":"banana"}})
```

Throws:
`DeltaError: Generic DeltaTable error: Schema error: No field named id. Valid fields are x, y.`

ion-elgreco:
@franz101 if you want to use python types you need to use the new_values parameter, and that one doesn't supports dictionaries.

PR #1754, opened 2023-10-22
feat: add metadata for reader\writer features

michaelbrichko:
# Description
Rust client should be able to understand if given table contains any reader\writer features

# Use case
Given delta file that was written by other client (possibly a more advanced one), try to understand if it requres any of the advanced features that might be specified in reader\writer features section of the protocol

# Documentation

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#protocol-evolution

### DIFF
--- a/rust/src/table/state.rs
+++ b/rust/src/table/state.rs
@@ -41,6 +41,8 @@ pub struct DeltaTableState {
     app_transaction_version: HashMap<String, i64>,
     min_reader_version: i32,
     min_writer_version: i32,
+    reader_features: HashSet<String>,
+    writer_features: HashSet<String>,
     // table metadata corresponding to current version
     current_metadata: Option<DeltaTableMetaData>,
     // retention period for tombstones in milli-seconds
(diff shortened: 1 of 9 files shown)


michaelbrichko:
Please review.
Thank you in advance!

rust/src/table/state.rs
```diff
@@ -290,6 +302,14 @@ impl DeltaTableState {
             self.min_writer_version = new_state.min_writer_version;
         }
 
+        for feature in new_state.reader_features.iter() {
+            self.reader_features.insert(feature.to_owned());
+        }
+
+        for feature in new_state.writer_features.iter() {
+            self.writer_features.insert(feature.to_owned());
+        }
```
wjones127, line 311: It's unclear to me why we should take the union, and not just adopt the latest set of features. This would mean you could never drop features, right?

wjones127 (commented):
I think I prefer the approach with enums in #1755, though this is fine.

michaelbrichko:
> I think I prefer the approach with enums in #1755, though this is fine.

Ok, then should I abondon this one ? 
I would like to have this feature available ASAP so hopefully it won't be too long untill the other PR will be checked-in.
Thank you.

michaelbrichko:
Abondoning....

PR #1756, opened 2023-10-22
feat: adopt kernel schemas and improve protocol support

roeap:
# Description

~~this PR depends on #1741.~~

Migrating the implementation of actions and schema over from kernel. The schema is much more complete in terms of the more recent delta features and more rigorously leverages the rust type system.

# Related Issue(s)

# Documentation


rust/src/kernel/actions/arrow.rs
```diff
@@ -0,0 +1,275 @@
...
+                        }
+                    }
+                    PrimitiveType::Date => {
+                        // A calendar date, represented as a year-month-day triple without a
+                        // timezone. Stored as 4 bytes integer representing days since 1970-01-01
+                        Ok(ArrowDataType::Date32)
+                    }
+                    PrimitiveType::Timestamp => {
+                        // Issue: https://github.com/delta-io/delta/issues/643
+                        Ok(ArrowDataType::Timestamp(TimeUnit::Microsecond, None))
```
ion-elgreco, line 125: What about timestamps that were written with the UTC timezone? Are we doing anything with the IsAdjustedToUtc flag in the parquet timestamp column?

roeap: right now this is not handled. Within delta this is also gated behind the `timestampNtz` feature. As this PR is quite hard to review due to its size, I focussed on not adjusting any business logic but rather only update the schema / action definitions and getting things to build. So support for that feature should be in a follow-up PR.

roeap:
@wjones127 @rtyler - There are some failing python tests related to how we handle schema metadata, where we need to make a decision on the kernel side. Other then that, I think it's ready for review.

Since this PR is again quite sizable, I focussed on not making any changes to the logic and limit impact on the APIs. As such no new feature support in this PR :).

rust/src/operations/create.rs
```diff
@@ -10,8 +10,9 @@ use serde_json::{Map, Value};
 use super::transaction::commit;
 use super::{MAX_SUPPORTED_READER_VERSION, MAX_SUPPORTED_WRITER_VERSION};
 use crate::errors::{DeltaResult, DeltaTableError};
-use crate::protocol::{Action, DeltaOperation, MetaData, Protocol, SaveMode};
-use crate::schema::{SchemaDataType, SchemaField, SchemaTypeStruct};
+use crate::kernel::{Action, DataType, Metadata, Protocol, StructField, StructType};
+// use crate::kernel::{SchemaDataType, SchemaField, SchemaTypeStruct};
```
wjones127, line 14: drop this?

wjones127 (approved):
Thanks for moving this forward. This will be good for unblocking us on making progress on newer protocol versions :)

rust/src/kernel/schema.rs
```diff
@@ -0,0 +1,788 @@
...
+        })?;
+
+    Ok((precision, scale))
+}
+
+impl Display for PrimitiveType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PrimitiveType::String => write!(f, "string"),
+            PrimitiveType::Long => write!(f, "bigint"),
```
wjones127, line 450: Oh I think this is a bug: [Delta Protocol says "long"](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#schema-serialization-format)

wjones127 (changes requested):
We should fix the `bigint` bug.

The other Python failure about `metadata` is a result of the metadata keys being strings. That seems like a case where the test might have been wrong, right?

rust/src/kernel/schema.rs
```diff
@@ -0,0 +1,788 @@
...
+}
+
+fn default_true() -> bool {
+    true
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+#[serde(rename_all = "camelCase")]
+/// Primitive types supported by Delta
+pub enum PrimitiveType {
```
cmackenzie1, line 376: I love this. ❤️

roeap:
@wjones127 - made display for primitive types consistent with protocol...

> The other Python failure about metadata is a result of the metadata keys being strings. That seems like a case where the test might have been wrong, right?

This is an open question to me and mainly motivated by not wanting `serde_json` in kernel. in practice it seems spark/delta encodes nested structures as a json string. We have been allowing arbitrary json data in our metadata fields.

Raised this in the kernel sync - i.e. does delta metadata support nested structures? - but jury is still out in that :). As for going forward here I would appreciate your opinion. Continuing to use `Value` for now or refactoring the test are both fairly easily done.

wjones127:
> Raised this in the kernel sync - i.e. does delta metadata support nested structures? - but jury is still out in that :). As for going forward here I would appreciate your opinion. Continuing to use Value for now or refactoring the test are both fairly easily done.

Absent any evidence, I don't know either. My slight inclination is to move toward strings.

roeap:
> My slight inclination is to move toward strings.

I share that inclination, so jut went ahead and made the change :).

wjones127 (approved):

PR #1763, opened 2023-10-23
refactor: perform bulk deletes during metadata cleanup

### LINKED ISSUE #1761: Use bulk deletes where possible
# Description

Underlying `object_store` crate has a `delete_stream` method which if more performant / efficient than calling `delete` for many files. Where possible, prefer usage of this function.

For example: https://github.com/delta-io/delta-rs/blob/a67140c6b13a6f820f4879f31eb927d140327746/rust/src/protocol/checkpoints.rs#L154-L162

**Use Case**

**Related Issue(s)**

cmackenzie1:
# Description
In addition to doing bulk deletes, I removed what seems like (at least to me) unnecessary code. At it's core, files are considered up for deletion when their last_modified time is older than the cutoff time AND the version is less than the specific version (usually the latest version).

# Related Issue(s)
- closes #1761 

# Documentation

- https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.delete_stream


rtyler:
@cmackenzie1 I looked this over and thought "coo coo" but I didn't like the intermediate iterators, so I put my thinking cap on and wondered if we could just feed streams into streams (sup dawg).

I haven't done any testing other running the existing test suite, but I think this works. If you don't like it feel free to revert my commit out and we can discuss more :smile:

rust/src/protocol/checkpoints.rs
```diff
@@ -141,148 +140,55 @@ pub async fn create_checkpoint_for(
...
+    // Feed a stream of candidate deletion files directly into the delete_stream
+    // function to try to improve the speed of cleanup and reduce the need for
+    // intermediate memory.
+    let deleted = storage
+        .delete_stream(
+            storage
+                .list(Some(storage.log_path()))
+                .await?
+                // Pass along only the Ok results from storage.list
+                .filter(|res| futures::future::ready(res.is_ok()))
```
cmackenzie1, line 164: This was the only reason I didn't take this approach first. If we are ok with not handling the error here that's fine with me!

cmackenzie1 (commented):
Looks good! Thanks!

rust/src/protocol/checkpoints.rs
rtyler: Given the nature of the function, I don't think we would want to halt execution on a single error would we? 

Assuming that's correct, I I will modify the code to log errors but continue processing

rtyler:
Since I contributed here, I would like @wjones127 or @roeap to tap this one into main after a review

wjones127 (approved):
Looks good 👍

PR #1769, opened 2023-10-25
chore: fix building/running tests without the datafusion feature

rtyler:
This looks like an oversight that our CI didn't test because we have the datafusion feature typically enabled for our tests. The build error would only show up when building tests without it.


roeap (approved):
LGTM!

PR #1774, opened 2023-10-26
chore: refactor into the deltalake meta crate and deltalake-core crates

rtyler:
This puts the groundwork in place for starting to partition into smaller crates in a simpler and more manageable fashion.

See #1713


wjones127 (approved):
Thanks this is a good start. Easy to follow so far :)

PR #1775, opened 2023-10-26
chore: upgrade datafusion 33

Blajda:
# Description

I've had to live on the bleeding edge of datafusion since I've discovered multiple bugs while implementing merge enhancements. Creating this PR to contain changes necessary to use it. 

Datafusion has made significant changes with how table statistics are represented. I refactored and was able to trim a significant amount of code.
There were some bugs with how we presented statistics for tables that do not contain column metadata (deleta-0.2.0) where we stated the number of records for a file is 0.

## Fixes

The null-ability status for partition columns are now accurately captured by Datafusion. Before if a partition column contained a null value an error would be returned. This should be resolved now.


Blajda:
Requires a fix for the following regression https://github.com/apache/arrow-rs/issues/5038 This was found by the checkpoint writer tests

wjones127 (approved):
Thanks for working on this!

Issue #1721, opened 2023-10-13
Support parquet files with nanoseconds timestamps when converting a Parquet table to a Delta table


junjunjd:
# Description

When converting from `ArrowSchema` to `deltalake::Schema`, the `try_from` method will return an error if the arrow schema has a `Timestamp(nanosecond)`. [The `convert_to_delta` function](https://github.com/delta-io/delta-rs/pull/1686) will fail. It makes sense to support parquet files with nanoseconds timestamps when converting a Parquet table to a Delta table.

**Use Case**

Convert parquet files in `rust/tests/data/delta-1.2.1-only-struct-stats` to a Delta table.

**Related Issue(s)**
https://github.com/delta-io/delta-rs/issues/1041, https://github.com/delta-io/delta-rs/issues/1682, https://github.com/delta-io/delta-rs/pull/1686

hntd187:
Take

MrPowers:
Sweet, assigned it to you @hntd187

aditanase:
@rtyler @ion-elgreco I am hitting a variant of this when doing partition pruning on Timestamp columns, see related issue and discussion on delta-kernel: https://github.com/delta-io/delta-kernel-rs/issues/921.
In a nutshell, partition pruning does not work if the partition column has Timestamp type, on tables written with spark defaults.

I have a patch that fixes the `partitionValues_parsed` in the record batches loaded from the checkpoint, which I'd be happy to contribute in this form if you think it's useful as a temporary workaround:
https://github.com/hstack/delta-rs/commit/1b502724d9c1ac3524b41803ba56ae3164d3ffe9

If you are planning to address this globally in delta-rs or delta-kernel, or simply by upgrading to the correct arrow parquet versions, I'm fine keeping this as a a patch on our fork for now.

ion-elgreco:
@aditanase this workaround is not need since we are already on kernel-0.10 and arrow 55

aditanase:
@ion-elgreco got a chance to test this after rebasing on the newest versions, it is still not resolved.
Unfortunately the type hint doesn't get triggered, since the table in question is produced by spark (3.4.3), and the type hint is [probably only added by the parquet.rs writers](https://arrow.apache.org/rust/parquet/arrow/index.html#schema-conversion).
As a result, the changelog batches still come back with Nanos resolution, and schema mapping errors silently, leading to no pruning.

What is your recommended path forward, since we don't always control the table producers.

PR #1782, opened 2023-10-28
feat: add hint to support parquet with nanosecond timestamps

hntd187:
# Description
This gives a hint on the schema convert telling convert to delta that the timestamps are nanosecond and need to be converted to the appropriate microseconds, at which point the schema written should use the standard `SchemaDataType::primitive("timestamp")` style.

# Related Issue(s)
For #1721 @junjunjd 

# Documentation


rtyler:
Thanks for the pull request, I'm still mulling this over and kind of hoping somebody else chimes in with an opinion on the approach :thinking:

hntd187:
My own two cents is this is a very meh solution. But it's the only obvious way I saw to carry some context to someone converting an arrow schema to do a `conver_to_delta` It may make more sense for the conversion to delta to not use the standard schema conversion facilities and have something more purpose fit that gives the context on what if anything has to be converted.

PR #1789, opened 2023-10-31
fix: delta scan partition ordering bug

### LINKED ISSUE #1787: `MERGE` works incorrectly with partitioned table if the data column order is not same as table column order
# Environment

**Delta-rs version**: 0.12.0

**Binding**: Python

***
# Bug

**What happened**: When you do a merge on a table that has partition columns in a different order than the source data it will cause the wrong values to be updated in the partition columns.

Update the columns correctly irrespective of the order.

**How to reproduce it**:
Write table first.
```python
import polars as pl
from polars.io.delta import _convert_pa_schema_to_delta

df = pl.DataFrame({
    "foo": ['foo_value'],
    "bar": ['bar_value'],
    'value': [10]
})

arrow_data = df.to_arrow()
arrow_data = arrow_data.cast(_convert_pa_schema_to_delta(arrow_data.schema))
write_deltalake("test_table", arrow_data,  partition_by=['bar','foo'])

```
Merge the result, the partition column order is `bar` `foo`. Data order is `foo` & `bar

```python
df_merge = pl.DataFrame({
    "foo": ['foo_value'],
    "bar": ['bar_value'],
    'value': [20]
})

arrow_data_merge = df_merge.to_arrow()
arrow_data_merge = arrow_data_merge.cast(_convert_pa_schema_to_delta(arrow_data_merge.schema))

dt = DeltaTable('test_table')

dt.merge(arrow_data_merge, 
         predicate='s.value = t.value', 
         source_alias='s', target_alias='t').when_not_matched_insert_all().execute()
```

The result is incorrect as you can see:

```python
dt = DeltaTable('test_table')

print(pl.from_arrow(dt.to_pyarrow_table()))

┌───────────┬───────────┬───────┐
│ foo       ┆ bar       ┆ value │
│ ---       ┆ ---       ┆ ---   │
│ str       ┆ str       ┆ i64   │
╞═══════════╪═══════════╪═══════╡
│ bar_value ┆ foo_value ┆ 10    │
│ foo_value ┆ bar_value ┆ 20    │
└───────────┴───────────┴───────┘
```

Blajda:
# Description
Sometimes the order of partition columns in our delta schema does not match the order of partition columns in the deltatable metadata. 
This would cause `DeltaScan` to provide incorrect values for partition columns.
This is fixed by having `DeltaScan` use the metadata as the source of truth.

# Related Issue(s)
- closes #1787


mightyshazam:
take

Blajda:
@mightyshazam @MrPowers Why is there a take on this PR? it's mostly done the only things that needs to be done is removing some unwraps.

MrPowers:
@Blajda - sorry for the miscommunication.  @mightyshazam will help out with a review.  I assigned it to him cause I can't assign him as a reviewer, but now realize that was misleading.  So yea, let @mightyshazam know when you're done so he can take a look, thanks!!

Blajda:
@MrPowers 
ah thanks for the clarification!

Blajda:
@mightyshazam this is ready for review. Thanks for taking the initiative

crates/deltalake-core/src/delta_datafusion/mod.rs
```diff
@@ -586,8 +586,15 @@ impl<'a> DeltaScanBuilder<'a> {
         // However we may want to do some additional balancing in case we are far off from the above.
         let mut file_groups: HashMap<Vec<ScalarValue>, Vec<PartitionedFile>> = HashMap::new();
 
+        let table_partition_cols = self
```
mightyshazam, line 589: You can borrow here with `&` instead of cloning. Then you can pass that reference directory to `partitioned_file_from_action` instead of borrowing it for the call. It will save you the clone.

mightyshazam (approved):
Looks good. I pointed out a pretty trivial optimization to take or leave.

crates/deltalake-core/src/delta_datafusion/mod.rs
Blajda: great catch. I'll take the optimization.

mightyshazam (approved):

wjones127 (approved):

PR #1792, opened 2023-11-01
feat: add configuration key to allow filtering stats written by column

emcake:
# Description
I have a use case with a table containing large binary objects in a column (approx. 1MB per object). When writing the stats for the Add action for the delta log, we compute the min and max sizes for this column, and then serialize it as JSON. Once we have serialized the min and max, and then heavily escaped both of these, the delta log files approach 30MB individually. If left for too long (this is quite a large dataset with a large number of commits) the delta log can become impossible to read - 1000 append commits means taking 30GB to just parse the delta log actions.

This commit allows configuring which columns to generate stats for when creating add actions. the config key `delta-rs.writeStatsFilter` can be specified as a JSON object indicating includes or excludes:

```json
{
  "include": ["col_a", "col_b"]
}
```

Where `include` will only generate stats for listed columns, and `exclude` will generate stats for columns except those listed. Nested columns can also be indexed via structs:

```json
{
  "exclude": {
    "col_a" : {},
    "struct_col" : {
      "col_b" : {},
      "doubly_nested": ["col_c"]
    }
}
```

# Related Issue(s)

# Documentation


wjones127:
>  When writing the stats for the Add action for the delta log, we compute the min and max sizes for this column, and then serialize it as JSON.

To be clear, are you doing this yourself or is this happening inside delta-rs. We should be truncating large binary values, and if we aren't I think a better solution is to turn that on, since it will apply automatically without user input. This should have been completed upstream in: https://github.com/apache/arrow-rs/pull/4389

emcake:
Oh, nice - yes, that would be a satisfactory solution. The stats were coming from delta, yes. I was observing very large delta logs with 0.16 - let me try getting a local version doing truncation on stats correctly and I’ll submit a new PR with any changes needed.

PR #1796, opened 2023-11-02
feat: initial table features implementation

hntd187:
# Description
Hi again, this is table features again, but albeit with less work done on the main branch like the smart guy I am.

# Related Issue(s)
#1755 prior review and information

# Documentation
https://delta.io/blog/2023-07-27-delta-lake-table-features/
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features


crates/deltalake-core/src/protocol/mod.rs
```diff
@@ -640,6 +640,190 @@ pub struct Protocol {
...
+    DELETION_VECTORS,
+    /// timestamps without timezone support
+    #[serde(alias = "timestampNtz")]
+    TIMESTAMP_WITHOUT_TIMEZONE,
+    /// version 2 of checkpointing
+    #[serde(alias = "v2Checkpoint")]
+    V2_CHECKPOINT,
+    /// If we do not match any other reader features
+    #[serde(other)]
+    OTHER,
```
cmackenzie1, line 671: Should we capture this value using an untagged variant instead? 

Something like this

```rust
#[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
pub enum Example {
    Foo,
    Bar,
    #[serde(untagged)]
    Unknown(String),
}

#[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
struct Features {
    reader: Vec<Example>
}

fn main() {
    let features = Features { 
        reader: vec![Example::Foo, Example::Bar, Example::Unknown("Hello".to_string())],
    };
    println!("{}", serde_json::to_string(&features).unwrap());
    let features: Features = serde_json::from_str(r#"{"reader":["Foo","Bar","Hello"]}"#).unwrap();
    println!("{:#?}", features);
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ef8134eeaf36d87ca8fbb38b57d18765

hntd187: Actually read my mind, I like this more than other.

wjones127 (dismissed):

wjones127 (approved):

Issue #1744, opened 2023-10-20
Support `pyarrow>=13`


dhirschfeld:
# Description
As it say's in the title.

https://github.com/delta-io/delta-rs/blob/16910304e64792410fcefeac01507fb621fc8f1e/python/pyproject.toml#L20

***xref:*** #1743

wjones127:
Started to look into this in https://github.com/delta-io/delta-rs/pull/1602, FYI

PR #1602, opened 2023-08-27
fix: various issues from PyArrow 13.0.0 release

wjones127:
# Description
The description of the main changes of your pull request

# Related Issue(s)

# Documentation

PR #1804, opened 2023-11-04
fix(python): add support for pyarrow 13+

ion-elgreco:
# Description

I build on top of the branch of @wjones127 https://github.com/delta-io/delta-rs/pull/1602. In pyarrow v13+ the ParquetWriter by default uses the `compliant_nested_types = True` (see related PR: https://github.com/apache/arrow/pull/35146/files)and the docs: https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html).

In arrow/parquet-rs it fails when it compares schemas because it expected the old non-compliant ones. For now we can have pyarrow 13+ supported by disabling it or updating the file options provided by a user.

# Related Issue(s)

- Closes https://github.com/delta-io/delta-rs/issues/1744

# Documentation


python/deltalake/table.py
```diff
@@ -1435,3 +1435,18 @@ def z_order(
...
+    """
+    Cast a batch to a schema, if it is already considered equal.
+
+    This is mostly for mapping things like list field names, which arrow-rs
+    checks when looking at schema equality, but pyarrow does not.
+    """
+    if batch.schema == schema:
+        return pyarrow.Table.from_batches([batch]).cast(schema).to_batches()[0]
+    else:
+        return batch
```
wjones127, line 1452: @ion-elgreco If we take this out, do the tests still pass?

ion-elgreco: Yes, everything passes still. Shall I take it out? 

Or would we require it later once we enable the compliant types again.

wjones127: Ideally I think we shouldn’t need it. This was an earlier attempt at fixing the compliant types thing

ion-elgreco: We should have backwards compatibility during reads and writes when we use the compliant types, I think spark-delta also doesn't write these compliant types but I may be wrong here

wjones127 (approved):

wjones127:
Thanks for finishing this up!

PR #1807, opened 2023-11-05
feat: handle protocol compatibility

roeap:
# Description

In preparation for further improvements to protocol support, this PR introduces a `ProtocolChecker` which validates that we can read form / write to / commit to a specific table. So far everything is expressed in table features in the hopes, that this keeps on giving correct behaviours as we add more table features. 

The existing support for append only is integrated and extended to handle enablement logic according to the protocol.

### DIFF
--- a/crates/deltalake-core/benches/read_checkpoint.rs
+++ b/crates/deltalake-core/benches/read_checkpoint.rs
@@ -1,6 +1,6 @@
 use criterion::{criterion_group, criterion_main, Criterion};
-use deltalake::table::state::DeltaTableState;
-use deltalake::DeltaTableConfig;
+use deltalake_core::table::state::DeltaTableState;
+use deltalake_core::DeltaTableConfig;
 use std::fs::File;
 use std::io::Read;
 
--- a/crates/deltalake-core/src/operations/transaction/protocol.rs
+++ b/crates/deltalake-core/src/operations/transaction/protocol.rs
@@ -3,7 +3,7 @@ use std::collections::HashSet;
 use once_cell::sync::Lazy;
 
 use super::TransactionError;
-use crate::kernel::{ReaderFeatures, WriterFeatures};
+use crate::kernel::{Action, ReaderFeatures, WriterFeatures};
 use crate::table::state::DeltaTableState;
 
 static READER_V2: Lazy<HashSet<ReaderFeatures>> = Lazy::new(|| {
@@ -62,7 +62,7 @@ impl ProtocolChecker {
     }
 
     pub fn default_writer_version(&self) -> i32 {
-        1
+        2
     }
 
     /// Check if delta-rs can read form the given delta table.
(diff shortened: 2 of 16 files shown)


crates/deltalake-core/benches/read_checkpoint.rs
roeap, line 1: this is just a drive-by fix as builds were failing for benches.

crates/deltalake-core/src/operations/transaction/protocol.rs
```diff
@@ -0,0 +1,396 @@
...
+    writer_features.insert(WriterFeatures::AppendOnly);
+    writer_features.insert(WriterFeatures::Invariants);
+    // writer_features.insert(WriterFeatures::CheckConstraints);
+    // writer_features.insert(WriterFeatures::ChangeDataFeed);
+    // writer_features.insert(WriterFeatures::GeneratedColumns);
+    // writer_features.insert(WriterFeatures::ColumnMapping);
+    // writer_features.insert(WriterFeatures::IdentityColumns);
+
+    ProtocolChecker::new(reader_features, writer_features)
+});
```
rtyler, line 162: Overall I like the API of `ProtocolChecker` and I think it's a great addition to the library. The motivation for this global I am struggling with however. I thought it might be to avoid allocations, but as far as I can tell, the common paths through `can_read` and `can_write_to` will clone a number of the related HashSets anyways, which has me perplexed.

I think I'm just not understanding, would you mind putting some more commentary above `INSTANCE` on the motivation for it so that I may be enlightened :smile:

roeap: Happy to add some comments, just really quick here as well.

The cloning bothered me as well, and I'll see that I can fix that. The main motivation though is that with table features protocol support is essentially a negotiation between the client and the table. Since the table features we support lock in at compile time, there really is exactly one correct way the instance should look like, so i felt its less error prone if we just create it globally and track support for features in one place.

roeap: added some comments and avoided cloning the sets.

hntd187:
Ayyo @MrPowers

crates/deltalake-core/src/operations/transaction/protocol.rs
```diff
@@ -0,0 +1,404 @@
+use std::collections::HashSet;
+
+use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
+
+use super::TransactionError;
+use crate::kernel::{Action, ReaderFeatures, WriterFeatures};
+use crate::table::state::DeltaTableState;
+
+lazy_static! {
```
hntd187, line 10: What is this used for? The table reader and writers have `Into<usize>` impls so you should just be able to cast them to find their associated reader / writer version.

roeap: Main intent is to have a uniform way of handling compatibility for tables on 3/7 and the ones below.

hntd187: Sorry I think I might not understand the check here. Are you saying these are for the chain of required features? Like check constraints for example requires append only and invariants? So maybe it's the name that throws me off if that's the case.

roeap: as usual, it depends 😆 - tables features as a concept are only available at reader/writer version 3/7. AT that version, clients can validate that they implement all features the table requires as stated in the `Protocol` action. The respective fields must be available at these versions. For any reader/writer version below that, clients must assume that they need to implement all features up to the respective version.

hntd187: Ahhh I see, so this only matters when we are not dealing with actual table features and meant to enforce the "all or nothing" style of protocol versioning prior?

roeap: Exactly.

hntd187 (approved):

roeap:
@rtyler @wjones127 - any chance you could approve this so we can land it on main? There seems to be quite some activity around table features right now, so i think it would be valualble to establish a baseline...

wjones127 (approved):

PR #1810, opened 2023-11-06
chore: update read me

ion-elgreco:
# Description
Marking FSCK ~and Table features~ as done and MERGE as semi-done.


roeap:
Hmm... I don't think we can mark table features as done just yet. In #1807 we are introducing the basic wiring beyond reading the info and starting to do some checking. Haven't thought it through yet, but i wouldn't want to claim, that we cover all relevant code paths yet.

ion-elgreco:
> Hmm... I don't think we can mark table features as done just yet. In #1807 we are introducing the basic wiring beyond reading the info and starting to do some checking. Haven't thought it through yet, but i wouldn't want to claim, that we cover all relevant code paths yet.

Ah my bad, I thought with https://github.com/delta-io/delta-rs/pull/1796 we got the support to at least parse the features while reading.

roeap (approved):

Issue #868, opened 2022-10-09
Make load_version return DeltaTable object


MrPowers:
# Description

The Python `load_version` method returns `None`, so you can't do this: 

```python
dt = DeltaTable("tmp/some_delta_lake")
dt.load_version(0).to_pandas()
```

You have to do this:

```python
dt = DeltaTable("tmp/some_delta_lake")
dt.load_version(0)
dt.to_pandas()
```

Would it be alright to update the API to allow for the fluent style of programming?

ion-elgreco:
Closing it as per short discussion with Will in the above mentioned pr

PR #1811, opened 2023-11-06
feat(python): return deltatable object with load methods

ion-elgreco:
# Description
This would make it possible to chain to_pyarrow_dataset etc, like @MrPowers proposed. 

The only thing that I am doubting here is, that one could assume if I don't store the object in a variable, you expect the original deltatable object to be loaded with the previous version. This may be confusing, since the version loading is done in place, but let me know your thoughts.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/868

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -379,7 +379,7 @@ class DeltaTable:
 
     file_uris.__doc__ = ""
 
-    def load_version(self, version: int) -> None:
+    def load_version(self, version: int) -> "DeltaTable":
         """
         Load a DeltaTable with a specified version.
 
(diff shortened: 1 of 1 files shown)


wjones127:
> The only thing that I am doubting here is, that one could assume if I don't store the object in a variable, you expect the original deltatable object to be loaded with the previous version. This may be confusing, since the version loading is done in place, but let me know your thoughts.

Yeah, if I were being a purist, I think the maxim:

> functions should either mutate their argument or return a value, but not both

applies here.

We could instead define a new method `DeltaTable.at_version()` maybe? I also think the original example could also just be written:

```python
dt = DeltaTable("tmp/some_delta_lake", version=0).to_pandas()
```

which looks fine and fluent to me 🤷

ion-elgreco:
@wjones127 Yes, good point. I'll abandon this then since it doesn't sit right

PR #1813, opened 2023-11-07
chore: fix typos

xiaolou86:


roeap (approved):
Thanks for fixing these @xiaolou86!

Issue #227, opened 2021-04-29
Unit and integration tests for the vacuum command


rtyler:
Our preliminary support for vacuum would benefit from some more unit and integration tests to ensure that it's only deleting the right files whenever the vacuum is invoked.

danmx:
Referencing @mrk-its [comment](https://github.com/delta-io/delta-rs/issues/97#issuecomment-1173174328) from #97 so it's not lost.

> @rtyler @fvaleye It looks like there are still two serious issues with vacuum implementation:
> 
>     * vacuum lists all files in dataset using `StorageBackend.list_objs`. The problem is this function returns all files (including these in subdirectories) on `s3` backend and `gcs` backend (althrough I'm not sure about gcs). On `file` and `azure` backends this function  lists only first-level files (without recursing to subdirectories).
> 
>     * vacuum ignores files not referenced by delta log at all (so not included on DeltaTableState.files() and DeltaTableState.all_tombstones() lists).

ion-elgreco:
This should be well tested at this point, also with introduction of VACUUM START, VACUUM END, I added a couple tests on the Python side

PR #1817, opened 2023-11-07
fix: add handling for unmanaged files to vacuum command

Jan-Schweizer:
# Description
Adds handling for unmanaged files to the VACUUM command. The VACUUM command should correctly handle files that are not managed by a Delta table, e.g. files and directories that start with `_` or `.`.
Additionally fixes test `test_non_managed_files`.

# Related Issue(s)
- closes [#227](https://github.com/delta-io/delta-rs/issues/227) (Unit and integration tests for the vacuum command)


wjones127 (dismissed):
Looks good. Thanks for finishing this.

wjones127 (commented):
It looks like there is a test assertion that just needs to be updated, since `_change_data` and `_delta_index` are exceptions in `is_hidden_directory()`.

roeap:
We may have some additional aspects into consideration when selecting the files to be vacuumed. As it stands I think we may accidentally delete deletion vector files for files that are still part of the log. We should however be able to get the corresponding files names from the add action and add them to the managed files.

Some `.crc` files might also get deleted pre-maturely, but these are always optional, so no harm to the table data.

crates/deltalake-core/src/operations/vacuum.rs
```diff
@@ -195,12 +195,48 @@ impl VacuumBuilder {
...
+                            None
+                        }
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                };
+            }))
+            .collect::<HashSet<String>>();
```
Jan-Schweizer, line 224: Since the deletion vector file path is being constructed on the fly by the `absolut_path` function, it now must be a `HashSet<String>` instead of `HashSet<&str>`. This also means that we now have to clone the String of the Add and Remove path. Is this fine or is there a better way to handle this?

Also, this nested handling of the deletion vectors doesn't look nice, can you show me a better/more idiomatic way to achieve this?

crates/deltalake-core/src/operations/vacuum.rs
```diff
@@ -195,12 +195,48 @@ impl VacuumBuilder {
...
+
         while let Some(obj_meta) = all_files.next().await {
             // TODO should we allow NotFound here in case we have a temporary commit file in the list
             let obj_meta = obj_meta.map_err(DeltaTableError::from)?;
-            if valid_files.contains(&obj_meta.location) // file is still being tracked in table
-            || !expired_tombstones.contains(obj_meta.location.as_ref()) // file is not an expired tombstone
-            || is_hidden_directory(partition_columns, &obj_meta.location)?
+
+            if is_hidden_file(partition_columns, &obj_meta.location)? {
```
Jan-Schweizer, line 230: I renamed this to is_hidden_file since we are not only ignoring directories but also single files

crates/deltalake-core/src/operations/vacuum.rs
```diff
@@ -362,11 +398,15 @@ impl VacuumPlan {
 /// Names of the form partitionCol=[value] are partition directories, and should be
 /// deleted even if they'd normally be hidden. The _db_index directory contains (bloom filter)
 /// indexes and these must be deleted when the data they are tied to is deleted.
-fn is_hidden_directory(partition_columns: &[String], path: &Path) -> Result<bool, DeltaTableError> {
-    let path_name = path.to_string();
-    Ok((path_name.starts_with('.') || path_name.starts_with('_'))
-        && !path_name.starts_with("_delta_index")
-        && !path_name.starts_with("_change_data")
+fn is_hidden_file(partition_columns: &[String], path: &Path) -> Result<bool, DeltaTableError> {
```
Jan-Schweizer, line 401: I renamed this to `is_hidden_file` since we are not only ignoring directories but also single files

crates/deltalake-core/src/operations/vacuum.rs
```diff
@@ -217,6 +263,16 @@ impl VacuumBuilder {
...
+
+    /// Whether a file is contained within the set of managed files.
+    fn is_file_managed(&self, managed_files: &HashSet<String>, file: &Path) -> bool {
+        return if is_absolute_path(file.as_ref()) {
+            managed_files.contains(file.as_ref())
+        } else {
+            let path = format!("{}{}", self.log_store.root_uri(), file.as_ref());
+            managed_files.contains(&path)
+        };
+    }
```
Jan-Schweizer, line 275: Can the `location` of an object in the log_store ever be absolute? Or is it always relative to the table root? If so, then we wouldn't need the `is_absolute_path` check

rtyler: @Jan-Schweizer the location can be an absolute URL, and that's the case with shallow cloned tables. The [add action](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file) describes the `path` key. 

So for better or worse, we need to check for an absolute path  :smile:

rtyler (dismissed):

ion-elgreco:
I'll pick this up next week since we getting often issues about vacuum not cleaning up all required files.

rtyler:
I have added a `VacuumMode::Full` which should help with these types of scenarios in #3368   :thinking: 

I am going to close this because it's gotten pretty out of date but I'm hopeful that we may have improved support for newer vacuum needs since this was originally created

PR #1819, opened 2023-11-07
fix: pin dynamodb_lock to 0.4

jacksonrnewhouse:
# Description
Having the dependency pointed at 0 means anything depending on deltalake uses the latest version of dynamodb_lock. There are breaking changes between 0.4 and 0.6, causing all prior versions of deltalake to fail to build, along with projects depending on them. This is pointed at main, but should probably be ported back as a patch release to all prior deltalake versions.


kerryeon:
Same idea. In particular, starting from version 0.6, the default constructor for `DynamoDbLockClient` has been removed and replaced with a factory pattern that causes unnecessary copying and wrapping. This is an issue that can only be resolved by updating the code on the `dynamodb_lock` crate side.

Issue #1467, opened 2023-06-15
fix(python): make sure we always write microsecond precision timestamps


wjones127:
# Description

Right now I think we are be accident. But PyArrow will change default to nanoseconds so we should specify this.

See: https://github.com/apache/arrow/issues/35746

**Use Case**

**Related Issue(s)**

anjakefala:
The goal is to get https://github.com/apache/arrow/issues/35746 merged in by 13.0.0!

ion-elgreco:
@wjones127 I can give this a go, should be straightforward and similar to the delta_arrow_schema_from_pandas logic, right?

ion-elgreco:
@wjones127 I was thinking this could work: https://github.com/delta-io/delta-rs/compare/main...ion-elgreco:delta-rs:fix/cast-timestamp-always-to-us-precision

But after checking the PyArrow docs and testing what a RecordBatchReader does, it apparently doesn't cast the schema when you pass a different one. It will throw an error that the schema that is passed is not the same as the one that was expected, so it only works on pa.Table.. 

Do you have any idea how you can cast/read a recordBatch to a new schema?

Also, one thing that won't be caught are all the fields or structs that contain nested date-times with different precision. This may possibly be the case for the method where the precision is fixed for pd.DataFrame.

Issue #1753, opened 2023-10-22
Writing with large arrow types in MERGE


ion-elgreco:
# Description
It seems it's not possible to write with large arrow types in `.merge()` yet, however there is write support with `write_deltalake(large_types=True)`, we should add this also in merge.

**Use Case**
Converting Polars dataframes to arrow and then merging immediately instead of casting to normal arrow types which may not fit if the arrays are too large.

```
DeltaError: Generic DeltaTable error: Execution error: Fail to build join indices in NestedLoopJoinExec, error:Arrow error: Invalid argument error: Invalid comparison operation: LargeUtf8 == Utf8
```

PR #1820, opened 2023-11-07
feat(python): add pyarrow to delta compatible schema conversion in writer/merge

ion-elgreco:
# Description
This ports some functionality that @stinodego and I had worked on in Polars. Where we converted a pyarrow schema to a compatible delta schema. It converts the following:

- uint -> int
- timestamp(any timeunit) -> timestamp(us) 

I adjusted the functionality to do schema conversion from large to normal when necessary, which is still needed in MERGE as workaround https://github.com/delta-io/delta-rs/issues/1753. 

Additional things I've added:

- Schema conversion for every input in write_deltalake/merge
- Add Pandas dataframe conversion
- Add Pandas dataframe as input in merge

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/686
- closes https://github.com/delta-io/delta-rs/issues/1467


ion-elgreco:
With https://github.com/delta-io/delta-rs/pull/1820/commits/843c1d8fa9b8168980641a92a3edeaa2af7cf518 it's robust against schema's that have mixed Large and normal types. The schema will either be down casted to normal types or everything to large types, depending on user input.

python/deltalake/schema.py
```diff
@@ -17,34 +17,91 @@
...
-    data: "pd.DataFrame",
-) -> Tuple[pa.Table, pa.Schema]:
-    """
-    Infers the schema for the delta table from the Pandas DataFrame.
-    Necessary because of issues such as:  https://github.com/delta-io/delta-rs/issues/686
-
-    Args:
-        data: Data to write.
+### Inspired from Pola-rs repo - licensed with MIT License, see licenses folder.###
+def _convert_pa_schema_to_delta(
```
wjones127, line 21: Perhaps we should add additional unit tests to enforce that these conversions are expected behavior now?

python/deltalake/schema.py
```diff
@@ -17,34 +17,91 @@
...
+        fields = [dtype.field(i) for i in range(dtype.num_fields)]
+        fields_cast = [pa.field(f.name, dtype_to_delta_dtype(f.type)) for f in fields]
+        return pa.struct(fields_cast)
+
+    return pa.schema([pa.field(f.name, dtype_to_delta_dtype(f.type)) for f in schema])
+
+
+def convert_pyarrow_recordbatchreader(
+    data: pa.RecordBatchReader, large_dtypes: bool
+) -> pa.RecordBatchReader:
```
wjones127, line 82: These return types do not match what is actually returned.

python/deltalake/schema.py
```diff
@@ -17,34 +17,91 @@
...
+    """Converts a PyArrow RecordBatchReader to a PyArrow RecordBatchReader with a compatible delta schema"""
+    schema = _convert_pa_schema_to_delta(data.schema, large_dtypes=large_dtypes)
+    data = pa.RecordBatchReader.from_batches(
+        schema,
+        data.read_all().cast(schema).to_batches(),
```
wjones127, line 87: It would be nice to achieve this with a generator rather that loading the entire reader into memory.

python/deltalake/schema.py
ion-elgreco: How so? I'm a bit less familiar with the recordbatchreader

ion-elgreco: Ah I see what you mean now :) I am returning an iterable of recordbatches hehe

python/deltalake/schema.py
ion-elgreco: I was looking into it but I don't see a way on how I could change the schema of the generator inside the reader. I tried to go from reader -> scanner but that's a dead end. Reader -> dataset, should theoretically work but `pa.dataset.dataset `mentions RecordBatchReader is an unsupported type even though the docs mentions it is a valid input type.

I am also not sure if materializing is even an issue because later before write or merge we do `valide_batch` which seems to materialize the results.

python/deltalake/schema.py
ion-elgreco: Done, I've added many test cases for:

- large -> normal
- normal -> large
- mixed large/normal -> large
- mixed large/normal -> normal
- uint -> int
- nested lists with string/binary/int
- timestamp -> timestamp(us)
- struct with nested fields

ion-elgreco:
@wjones127 do you know how we can cast schemas onto a pyarrow dataset or recordbatchreader without materializing? `replace_schema` on Dataset doesn't work apparently. 

Also I am not sure if it's even wrong to not materialize since we have the validate_batch to check the invariants at the end

python/deltalake/schema.py
wjones127: `validate_batch` is run per batch within a reader. `pa.RecordBatchReader.from_batches` will not materialize the batches if you pass a generator, like we do here:

https://github.com/delta-io/delta-rs/blob/9b938308490a60808443ee36b4f7b2c3055a6c15/python/deltalake/writer.py#L341-L343

So this should work:

```python
def cast_to_schema(reader):
    for batch in reader:
        yield pa.Table.from_batches([batch]).cast(schema).to_batches()[0]
data = pa.RecordBatchReader.from_batches(schema, cast_to_schema(schema))
```

python/deltalake/schema.py
wjones127: No, I meant you claim it returns `pa.RecordBatchReader` but it actually returns `Tuple[pa.RecordBatchReader, pa.Schema]`.

Also, why are you returning the schema? It's already a part of the RBR?

ion-elgreco: Ah indeed, wasn't fully awake ^^

I'm returning the schema for the data schema vs table schema check.

python/deltalake/schema.py
ion-elgreco: I see, I thought going through the reader is materializing the results. Ben then I should be able to do the same with the dataset once I convert it to a reader through the scanner.

python/deltalake/schema.py
wjones127: you can just access `data.schema` for that

https://arrow.apache.org/docs/python/generated/pyarrow.RecordBatchReader.html#pyarrow.RecordBatchReader.schema

ion-elgreco: Yeah, I'll make a commit to just directly access the schema

python/deltalake/schema.py
ion-elgreco: @wjones127 should be good now

python/deltalake/schema.py
```diff
@@ -17,34 +18,109 @@
...
-    table = table.cast(target_schema=schema)
-    return table, schema
+            return pa.list_(nested_dtype_cast)
+
+    def struct_to_delta_dtype(dtype: pa.StructType) -> pa.StructType:
+        fields = [dtype[i] for i in range(dtype.num_fields)]
+        fields_cast = [pa.field(f.name, dtype_to_delta_dtype(f.type)) for f in fields]
+        return pa.struct(fields_cast)
+
+    return pa.schema([pa.field(f.name, dtype_to_delta_dtype(f.type)) for f in schema])
```
wjones127, line 80: As written, this loses the nullability information. We can use [with_type()](https://arrow.apache.org/docs/python/generated/pyarrow.Field.html#pyarrow.Field.with_type) to preserve everything but the type:
```suggestion
    return pa.schema([f.with_type(dtype_to_delta_dtype(f.type)) for f in schema])
```

python/deltalake/schema.py
```diff
@@ -17,34 +18,109 @@
...
         else:
-            schema_out.append(field)
-    schema = pa.schema(schema_out, metadata=schema.metadata)
-    table = table.cast(target_schema=schema)
-    return table, schema
+            return pa.list_(nested_dtype_cast)
+
+    def struct_to_delta_dtype(dtype: pa.StructType) -> pa.StructType:
+        fields = [dtype[i] for i in range(dtype.num_fields)]
+        fields_cast = [pa.field(f.name, dtype_to_delta_dtype(f.type)) for f in fields]
```
wjones127, line 77: Same here:
```suggestion
        fields_cast = [f.with_type(dtype_to_delta_dtype(f.type)) for f in fields]
```

python/tests/test_schema.py
```diff
@@ -213,3 +220,206 @@ def test_delta_schema():
     schema_without_metadata = schema = Schema(fields)
     pa_schema = schema_without_metadata.to_pyarrow()
     assert schema_without_metadata == Schema.from_pyarrow(pa_schema)
+
+
+@pytest.mark.parametrize(
+    "schema,expected_schema,large_dtypes",
```
wjones127, line 226: I like these test :)

python/tests/test_writer.py
```diff
@@ -304,6 +304,62 @@ def test_write_iterator(
...
+
+    write_deltalake(tmp_path, dataset, mode="overwrite", large_dtypes=True)
+    assert DeltaTable(tmp_path).to_pyarrow_table() == sample_data
+
+
+def test_write_recordbatch(
+    tmp_path: pathlib.Path, existing_table: DeltaTable, sample_data: pa.Table
+):
+    batch = existing_table.to_pyarrow_table().to_batches()
+    print(len(batch))
```
wjones127, line 347:
```suggestion
```

python/tests/test_writer.py
```diff
@@ -304,6 +304,62 @@ def test_write_iterator(
...
+
+    write_deltalake(tmp_path, batch[0], mode="overwrite")
+    assert DeltaTable(tmp_path).to_pyarrow_table() == sample_data
+
+
+def test_write_recordbatch_large_dtypes(
+    tmp_path: pathlib.Path, existing_table: DeltaTable, sample_data: pa.Table
+):
+    batch = existing_table.to_pyarrow_table().to_batches()
+    print(len(batch))
```
wjones127, line 357:
```suggestion
```

python/tests/test_writer.py
```diff
@@ -304,6 +304,62 @@ def test_write_iterator(
     assert DeltaTable(tmp_path).to_pyarrow_table() == sample_data
 
 
+def test_write_dataset(
+    tmp_path: pathlib.Path, existing_table: DeltaTable, sample_data: pa.Table
+):
+    dataset = existing_table.to_pyarrow_dataset()
+
+    write_deltalake(tmp_path, dataset, mode="overwrite")
```
wjones127, line 312: This is possibly a bit convoluted, but you could instead do something like:

```suggestion
@pytest.parametrize("large_dtypes", [True, False])
@pytest.parametrize("constructor", [
    lambda table: table.to_pyarrow_dataset(),
    lambda table: table.to_pyarrow_table(),
    lambda table: table.to_pyarrow_table().to_batches()[0]
])
def test_write_dataset(
    tmp_path: pathlib.Path, existing_table: DeltaTable, sample_data: pa.Table,
    large_dtypes: bool, constructor
):
    dataset = constructor(existing_table)

    write_deltalake(tmp_path, dataset, mode="overwrite", large_dtypes=large_dtypes)
```

wjones127 (dismissed):
This looking good. Just a few minor changes and it's ready to ship :)

python/tests/test_schema.py
ion-elgreco: : D

python/deltalake/schema.py
ion-elgreco: Oh neat, didn't knew about this!

ion-elgreco: I've also added two tests to check if we maintain nullability.

ion-elgreco:
@wjones127 I've applied all the requested changes

python/deltalake/writer.py
roeap, line 1: should this file live at the same level as the regular license and be included via the `pyproject.toml`

wjones127: Here's what I would propose:

1. Move the license to `python/licenses/polars_license.txt` and reference that path in code comment.
2. Create a file `python/licenses/readme.md` and include a list of which license apply to which code.
3. Add MIT license to pyproject.toml

I think it's possible we will borrow other polars code in the futures, so nice to name the file `polars_license.txt`.

Does that sound good @ion-elgreco ?

roeap: just realized I commented on a completely unrelated file - luckily the question seems to have been understood regardless 😆.

Personally, I like @wjones127's suggestion.

ion-elgreco: @wjones127 can you take a look at point 3, I wasn't sure on how I could add another license file in the toml for it to work with maturin.

ion-elgreco: Are we ready to merge this? 😀

python/deltalake/schema.py
```diff
@@ -1,9 +1,10 @@
-from typing import TYPE_CHECKING, Tuple, Union
+from typing import TYPE_CHECKING, Generator, Union
 
 import pyarrow as pa
+import pyarrow.dataset as ds
 
 if TYPE_CHECKING:
-    import pandas as pd
+    pass
```
roeap, line 7: remove this?

python/deltalake/table.py
```diff
@@ -617,25 +624,38 @@ def merge(
...
+
         if isinstance(source, pyarrow.RecordBatchReader):
-            schema = source.schema
+            source = convert_pyarrow_recordbatchreader(source, large_dtypes=False)
         elif isinstance(source, pyarrow.RecordBatch):
-            schema = source.schema
-            source = [source]
+            source = convert_pyarrow_recordbatch(
+                source, large_dtypes=False
+            )  # TODO(ion): set to True once MERGE uses logical plan
```
roeap, line 639: logical plans for MERGE are merged, do you we want to update this?

roeap (commented):
two more small comments / question - overall looking great.!

python/deltalake/schema.py
ion-elgreco: Will do :) not needed anymore

python/deltalake/table.py
ion-elgreco: Yes forgot about this one, will set it to large.

Just a quick note, I noticed the rust engine writer is more flexible in large and normal types, so my thought is to deprecate the part of the schema conversion from large to normal and vice versa, once the python writer uses the rust engine exclusively.

ion-elgreco:
@roeap resolved the two comments!

(Coding on a phone is tricky with no formatting :rofl: )

roeap (approved):
👍

ion-elgreco:
@roeap had to force push :) can you approve again?

roeap (approved):

PR #1823, opened 2023-11-08
feat: add lazy loading of tables

PengLiVectra:
# Description
Add lazy loading of tables. When preforming streaming operations we don't need any version of the table loaded. Large tables are slow to load, so we see a huge performance boost by avoiding CPU time spent loading the table metadata.

# Related Issue(s)

# Documentation

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -229,6 +229,7 @@ class DeltaTable:
         storage_options: Optional[Dict[str, str]] = None,
         without_files: bool = False,
         log_buffer_size: Optional[int] = None,
+        lazy_load: bool = False
     ):
         """
         Create the Delta Table from a path with an optional version.
@@ -247,7 +248,7 @@ class DeltaTable:
                                 This can decrease latency if there are many files in the log since the last checkpoint,
                                 but will also increase memory usage. Possible rate limits of the storage backend should
                                 also be considered for optimal performance. Defaults to 4 * number of cpus.
-
+        :param lazy_load: when true the table metadata isn't loaded
         """
         self._storage_options = storage_options
         self._table = RawDeltaTable(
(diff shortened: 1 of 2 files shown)


python/deltalake/table.py
```diff
@@ -258,6 +259,15 @@ def __init__(
...
         self._metadata = Metadata(self._table)
+        
+        if lazy_load:
+            self._table = RawDeltaTable.load_lazy(table_uri)
+            self._metadata = None
+            return
+        self._table = RawDeltaTable(
+            table_uri, version=version, storage_options=storage_options
+        )
+        self._metadata = Metadata(self._table)
```
ion-elgreco, line 270: I think it would be cleaner if you do if-else here:

```suggestion
        if lazy_load:
            self._table = RawDeltaTable.load_lazy(table_uri)
            self._metadata = None
        else:
            self._table = RawDeltaTable(
                    table_uri, version=version, storage_options=storage_options
            )
            self._metadata = Metadata(self._table)
```

python/deltalake/table.py
```diff
@@ -258,6 +259,15 @@ def __init__(
             log_buffer_size=log_buffer_size,
         )
         self._metadata = Metadata(self._table)
+        
+        if lazy_load:
+            self._table = RawDeltaTable.load_lazy(table_uri)
```
ion-elgreco, line 264: How about the storage options here? We probably want to lazy load also tables in cloud storage

python/deltalake/table.py
```diff
@@ -247,7 +248,7 @@ def __init__(
                                 This can decrease latency if there are many files in the log since the last checkpoint,
                                 but will also increase memory usage. Possible rate limits of the storage backend should
                                 also be considered for optimal performance. Defaults to 4 * number of cpus.
-
+        :param lazy_load: when true the table metadata isn't loaded
         """
         self._storage_options = storage_options
         self._table = RawDeltaTable(
```
ion-elgreco, line 254: I think something went wrong here, since you're loading it first with everything

PR #1824, opened 2023-11-08
feat: use pyo3-asyncio to get a fresh tokio runtime

PengLiVectra:
# Description
Use pyo3-asyncio to get a fresh tokio runtime.

# Related Issue(s)

# Documentation

### DIFF
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -45,6 +45,10 @@ reqwest = { version = "*", features = ["native-tls-vendored"] }
 version = "0.19"
 features = ["extension-module", "abi3", "abi3-py37"]
 
+[dependencies.pyo3-asyncio]
+version = "0.16.0"
+features = ["tokio-runtime"]
+
 [dependencies.deltalake]
 path = "../crates/deltalake"
 version = "0"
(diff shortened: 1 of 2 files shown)


roeap:
Thanks @PengLiVectra for looking into this.

I haven't looked into how pyo3-asyncio manages the tokio runtime, so I was wondering if you have any thoughts on how the datafusion python library [mamanges their lifetime](https://github.com/apache/arrow-datafusion-python/blob/e4942b897472ec2f0c59682db4954d0aa2d5c185/src/lib.rs#L77-L80), and if it may be an alternative that does not require the extra dependency...

wjones127:
Hi @PengLiVectra. Thanks for submitting a PR. Before we make changes like these, we like to open issues to discuss them. Without any description in the PR, it's hard to tell what problem you are trying to solve here. But I'm assuming this has to do with the issue that we recreate a runtime each time `rt()` is called.

In Lance, we wrote our own wrapper around the tokio runtime that let's use do things like propagate `KeyboardInterrupt` and release the GIL:

https://github.com/lancedb/lance/blob/main/python/src/executor.rs

We keep this reference as a Lazy static here:

https://github.com/lancedb/lance/blob/d4813fccb46fefc00c48f2b277bfb56bf3f1e31d/python/src/lib.rs#L85-L87

We could consider implementing something similar here.

Issue #1601, opened 2023-08-25
Remove rusoto in favor of the official AWS SDK for Rust


rtyler:
# Description

[Rusoto has been in maintenance mode](https://github.com/rusoto/rusoto#maintenance-status) for the last 2.5 years. In the mean time the official [AWS SDK for Rust is in developer preview](https://aws.amazon.com/sdk-for-rust/) and I would guess that they'll call it "GA" at re:invent this year.

Either way developer preview is better than "not really maintained".

This issue is to track the migration away from rusoto.

**Use Case**

**Related Issue(s)**

tustvold:
FWIW object_store now exposes a request authoriser, if you wanted to keep dependencies down, and aren't using an extensive number of APIs - https://docs.rs/object_store/latest/object_store/aws/struct.AwsAuthorizer.html

PR #1825, opened 2023-11-08
chore: break Glue support into its own crate without rusoto

rtyler:
This change also pilots a removal of Rusoto in favor of the AWS SDK for Rust which AWS is now supporting and funding the development of.

The API surface is largely the same, but this move I believe will aso ensure  that we're much more consistent on handling AWS environment variables for some  things.

Related to #1601


cmackenzie1:
I just have a comment about the crate name. Is the goal to have `deltalake-catalog-<glue|unity|etc>` or should it just be `deltalake-catalog` with feature flags for `glue`, `unity` and the like? I've never interacted with the catalogs and not sure if the more detailed crates going to be a pain or blessing.

rtyler:
@cmackenzie1 Yes the plan is to put each catalog provider into its own crate, there's not really any common code between something like Unity, which will likely have dependencies on Databricks SDKs, and Glue which has AWS SDK dependencies.

Some of the theory between parcelling these things out is that we can rev releases of these less frequently and use semver to make it easier to pull in updates.

rtyler:
The benchmark test failure I don't understand at all, but also don't think it has any relation to this pull request :confused:

crates/deltalake-core/src/data_catalog/mod.rs
```diff
@@ -25,11 +23,14 @@ pub enum DataCatalogError {
...
         catalog: &'static str,
-
         /// Error message
         source: Box<dyn std::error::Error + Send + Sync + 'static>,
     },
 
+    /// Error from a specific catalog provider
+    #[error("Catalog implementation error: {0}")]
+    Error(Box<dyn std::error::Error + Send + Sync + 'static>),
+
```
roeap, line 33: should integrations raise the above `Generic` error, which also contains an identifier from the specific catalogue? This pattern worked quite nicely in object store :).

crates/deltalake/Cargo.toml
```diff
@@ -24,8 +25,8 @@ default = ["arrow", "deltalake-core/parquet"]
 datafusion = ["deltalake-core/datafusion"]
 datafusion-ext = ["datafusion"]
 gcs = ["deltalake-core/gcs"]
-glue = ["deltalake-core/glue"]
-glue-native-tls = ["deltalake-core/glue-native-tls"]
+glue = ["deltalake-catalog-glue"]
+glue-native-tls = ["deltalake-catalog-glue/native-tls"]
```
roeap, line 29: was just wondering if there maybe is a way to propagate this from the general native-tls feature, and if this would even be wanted?.

roeap (commented):
Looking good! Exciting to see the split-up take shape!

Main question is arund the build scripts, while I don't feel too strongly about it, elsewhere we use makefiles rather then individual scripts. Then again locally I have a justfile 😆.

crates/deltalake-core/src/data_catalog/mod.rs
rtyler: Can you link me to what you're referring to @roeap ? I can make the changes here once I understand what you're suggesting :smile:

crates/deltalake/Cargo.toml
rtyler: I have not found a good way to propagate features, it's a little tedious book-keeping :cry:

crates/deltalake-core/src/data_catalog/mod.rs
roeap: certainly :). https://github.com/delta-io/delta-rs/blob/2b913b37e71ed96212dcec8c3fc8e865754ced82/crates/deltalake-core/src/data_catalog/mod.rs#L25-L31

crates/deltalake/Cargo.toml
roeap: not entirely sure, but I found the [example](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) below.

```toml
[dependencies]
serde = { version = "1.0.133", optional = true }
rgb = { version = "0.8.25", optional = true }

[features]
serde = ["dep:serde", "rgb?/serde"]
```

The docs there say.. 
> In this example, enabling the serde feature will enable the serde dependency. It will also enable the serde feature for the rgb dependency, but only if something else has enabled the rgb dependency.

So if I read that correctly, we may be able to add `"deltalake-catalog-glue?/native-tls"` to the `native-tls` feature and it only takes effect, if glue is enabled?

rtyler: Coming back to this, I think this is a good idea, but it will require a features rework in the metacarte since there we also have `s3-native-tls` for example. I'll tackle that in a follow up pull request

crates/deltalake-catalog-glue/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl std::fmt::Debug for GlueDataCatalog {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(fmt, "GlueDataCatalog")
+    }
+}
+
+// Placeholder suffix created by Spark in the Glue Data Catalog Location
+const PLACEHOLDER_SUFFIX: &str = "-__PLACEHOLDER__";
```
roeap, line 60:
```suggestion
const PLACEHOLDER_SUFFIX: &str = "__PLACEHOLDER__";
```

roeap (commented):
I think we may have one relevant typo, otherwise looking good!

crates/deltalake-catalog-glue/src/lib.rs
roeap: ... or rather looking at the name "suffix" this may actually be intended, but making sure nonetheless 😆.

roeap (approved):

roeap:
@rtyler - meant to resolve merge conflicts, but broke formatting :(

roeap (approved):

PR #1829, opened 2023-11-09
[WIP] chore: add kernel ExpressionHandler

roeap:
based on #1807

# Description

In the effort to advance protocol support and move our internal APIs closer to the kernel library, it is advantageous to leverage the expression handling logic from kernel specifically for filtering actions etc.


Blajda:
Once this is merged in do you suggest we remove the CNF expressions we use for optimize? Seem like we should depreciate that interface and also create a bridge between string expressions and maybe extra code to convert to an equivalent Datafusion expression.

roeap:
From the kernel side, the general idea is to have a way of defining expressions without additional dependencies. So yes, we should ten likely define conversions to datafusion expressions. I only looked a bit into our partition filters, if we could remove those too... From the python API side at least we probably want to keep CNF expressions around?

wjones127 (approved):

PR #1830, opened 2023-11-09
fix: run integration tests in CI

roeap:
# Description

For our integrations tests we we selecting the `deltalake` crate. After splitting the crate into smaller crates, integration tests are no longer defined there, and thus are currently not executed in CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -139,11 +139,11 @@ jobs:
 
       - name: Run tests with rustls (default)
         run: |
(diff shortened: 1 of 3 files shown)


rtyler (approved):

roeap:
@rtyler @wjones127 - it seems the s3 specific `repair_when_worker_pauses_{after,before}_rename_test`s are hanging indefinately right now. I disabled them for the currently passing test. Could this be related to the recent dynamo-db lock update? WHich bear the question should we maintain tests tests here, or does it make more sense to move them to the lock crate repo?

rtyler (approved):

PR #1836, opened 2023-11-10
fix: use physical name for column name lookup in partitions

### LINKED ISSUE #1835: get_add_actions fails with deltalake 0.13
# Environment

**Delta-rs version**: 0.13.0

**Binding**: Python

**Environment**:
- **Cloud provider**:
- **OS**: Windows Server / Windows 10 (both tested)
- **Other**:

***
# Bug

**What happened**:

`get_add_actions` fails, with both flatten=True/False:

```python
dt = DeltaTable("tbl")
dt.get_add_actions()
```

gives:

```
thread '<unnamed>' panicked at crates\deltalake-core\src\table\state_arrow.rs:175:53:
called Option::unwrap() on a None value
stack backtrace:
   0:     0x7ffc6523beca - BrotliDecoderVersion
   1:     0x7ffc65263dbb - BrotliDecoderVersion
   2:     0x7ffc65237231 - BrotliDecoderVersion
   3:     0x7ffc6523bc4a - BrotliDecoderVersion
   4:     0x7ffc6523e6ba - BrotliDecoderVersion
   5:     0x7ffc6523e328 - BrotliDecoderVersion
   6:     0x7ffc6523ed6e - BrotliDecoderVersion
   7:     0x7ffc6523ec1a - BrotliDecoderVersion
   8:     0x7ffc6523cb89 - BrotliDecoderVersion
   9:     0x7ffc6523e960 - BrotliDecoderVersion
  10:     0x7ffc653a5365 - BrotliDecoderVersion
  11:     0x7ffc653a5412 - BrotliDecoderVersion
  12:     0x7ffc62951a14 - PyInit__internal
  13:     0x7ffc62746857 - BrotliDecoderSetParameter
  14:     0x7ffc6271b8ce - BrotliDecoderSetParameter
  15:     0x7ffc627377c1 - BrotliDecoderSetParameter
  16:     0x7ffca3f8961e - PyUnicode_ToDecimalDigit
  17:     0x7ffca3ed494b - PyObject_Vectorcall
  18:     0x7ffca3ed5da4 - PyEval_EvalFrameDefault
  19:     0x7ffca3ef2f73 - PyMapping_Check
  20:     0x7ffca3ef27db - PyEval_EvalCode
  21:     0x7ffca3f62582 - Py_SourceAsString
  22:     0x7ffca3f624fe - Py_SourceAsString
  23:     0x7ffca40a3df4 - PyThread_tss_is_created
  24:     0x7ffca4006b99 - PyRun_SimpleFileObject
  25:     0x7ffca40078e0 - PyRun_AnyFileObject
  26:     0x7ffca4007387 - PyDict_DelItemString
  27:     0x7ffca4007243 - PyDict_DelItemString
  28:     0x7ffca3fc0d80 - Py_RunMain
  29:     0x7ffca3fc0a2d - Py_RunMain
  30:     0x7ffca3ec81b5 - Py_Main
  31:     0x7ff62cf51230 - <unknown>
  32:     0x7ffcca5684d4 - BaseThreadInitThunk
  33:     0x7ffccbf61791 - RtlUserThreadStart
  File "k:\Software\AZCopy\CSVGetter\test_ds.py", line 5, in <module>
    dt.get_add_actions()
  File "K:\Software\AZCopy\CSVGetter\.venv\Lib\site-packages\deltalake\table.py", line 859, in get_add_actions
    return self._table.get_add_actions(flatten)

           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called Option::unwrap() on a None value
```

Well, it should not panic

**How to reproduce it**:

Repo is here: [tester.zip](https://github.com/delta-io/delta-rs/files/13317653/tester.zip) 

**More details**:

aersam:
# Description
get_actions wrongly assumes that partition_columns from schema and partitionValues from log must be the same. This is not true since partition_columns are logical column names while partitionValues are physical column names.

Tests pending

# Related Issue(s)

- closes #1835

# Documentation
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-column-mapping 
"Track partition values and column level statistics with the physical name of the column in the transaction log."


aersam:
seems to be compiling in CI/CD. 

Where would you propose to add a test for this?

crates/deltalake-core/src/table/state_arrow.rs
```diff
@@ -167,13 +167,33 @@ impl DeltaTableState {
...
+            .iter()
+            .map(|name| {
+                let physical_name = metadata
+                    .schema
+                    .field_with_name(name)
+                    .unwrap()
+                    .physical_name();
+                (physical_name, name)
+            })
+            .collect::<HashMap<&String, &String>>();
```
wjones127, line 181: small nit: `&String` is almost never what we want, as it's a pointer to a pointer. You can just use `&str` here.

```suggestion
                (physical_name, name.as_str())
            })
            .collect::<HashMap<&str, &str>>();
```

crates/deltalake-core/src/kernel/schema.rs
```diff
@@ -150,6 +150,16 @@ impl StructField {
         self.nullable
     }
 
+    /// Returns the phyiscal name
+    pub fn physical_name(&self) -> &String {
```
wjones127, line 154:
```suggestion
    pub fn physical_name(&self) -> &str {
```

wjones127 (commented):
For testing, I think the tests belong here: https://github.com/delta-io/delta-rs/blob/9b938308490a60808443ee36b4f7b2c3055a6c15/crates/deltalake-core/src/protocol/mod.rs#L858

aersam:
> For testing, I think the tests belong here:
> 
> https://github.com/delta-io/delta-rs/blob/9b938308490a60808443ee36b4f7b2c3055a6c15/crates/deltalake-core/src/protocol/mod.rs#L858

Ok, I'll add some tests. How am I supposed to execute those locally? Always get strange compile errors locally.  I'm using Windows

roeap:
> Ok, I'll add some tests. How am I supposed to execute those locally? Always get strange compile errors locally. I'm using Windows

There is currently a build issue on main which may be related. Could you try with the `datafusion` flag enabled?

aersam:
> > Always get strange compile errors locally. I'm using Windows
> 
> There is currently a build issue on main which may be related. Could you try with the `datafusion` flag enabled?

Does not help, however in WSL it works just with `cargo test`. Tests seem to be passing now, so if you want to review ;)

aersam:
I don't think the build error is related to my changes? At least I would not understand it 🙂

crates/deltalake-core/src/table/state_arrow.rs
```diff
@@ -167,13 +167,27 @@ impl DeltaTableState {
...
             .collect::<HashMap<&str, _>>();
 
+        let physical_name_to_logical_name = metadata
+            .partition_columns
+            .iter()
+            .map(|name| {
+                let physical_name = metadata
+                    .schema
+                    .field_with_name(name)
+                    .unwrap()
```
roeap, line 177: maybe add a quick comment why it is ok to unwrap here.

roeap:
> I don't think the build error is related to my changes? At least I would not understand it 🙂

Agreed, this seems very unlikely to be related 😄.

crates/deltalake-core/src/kernel/schema.rs
```diff
@@ -166,6 +166,16 @@ impl StructField {
         self.nullable
     }
 
+    /// Returns the phyiscal name
+    pub fn physical_name(&self) -> &str {
+        let phys_name = self.get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName);
+        match phys_name {
+            None => &self.name,
+            Some(MetadataValue::String(s)) => s,
+            _ => panic!("Unexpected value for physical name"),
```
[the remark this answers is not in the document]
aersam: Makes sense but here a number is really nonsense, no?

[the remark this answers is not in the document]
aersam: I changed the code a bit so that it's more clear that I panic on number only. Also I added a comment

crates/deltalake-core/src/table/state_arrow.rs
aersam: Added that comment

aersam:
> Thanks @aersam for getting us started on column mapping!
> 
> Left a comment around error handling, but overall looking good.
> 
> One thing I noticed is that there is a lot going on in the example data log. Would it maybe be an option to set up the column mapping on creating the table, which makes it a bit easier to reason about the relevant commit.

I now simplified the test file, it's a lot less files now.
So it's ready for the next review :)

aersam:
> > Always get strange compile errors locally. I'm using Windows
> 
> There is currently a build issue on main which may be related. Could you try with the `datafusion` flag enabled?

One comment here: the error disappeared on Windows when doing `cargo update`

crates/deltalake-core/src/kernel/schema.rs
```diff
@@ -166,6 +166,16 @@ impl StructField {
         self.nullable
     }
 
+    /// Returns the phyiscal name
+    pub fn physical_name(&self) -> &str {
+        let phys_name = self.get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName);
+        match phys_name {
+            None => &self.name,
+            Some(MetadataValue::String(s)) => s,
+            Some(MetadataValue::Number(_)) => panic!("Unexpected type for physical name"), // Number makes no sense here and is against the spec
```
roeap, line 175: What I meant here is not to panic at all but return a `Result`. 

Assuming a server running this code, if we hit this error in the cofig, we would crash the entire server. But this is an error i would deem recoverable and a server may just want to return a message to the user.

roeap (changes requested):
Thanks for the log clean-up - much easier to read now 👍. 

Clarified a bit on why to remove the panic and handle via a `Result`.

crates/deltalake-core/src/kernel/schema.rs
aersam: I corrected that, I must say I'm still learning a bit about rusty error handling, I hope it's better now.
One thing I noticed is that there are two error enum's, one in the kernel and one outside. I adjusted both, however it's kind of duplicate. Not sure I did everything how it's supposed to be.

aersam:
> Thanks for the log clean-up - much easier to read now 👍.
> 
> Clarified a bit on why to remove the panic and handle via a `Result`.

In addition to a lot more error handling, I also added a check for delta.columnMapping.mode which matches the spec. If column mapping mode=none, we should ignore physical name entirely

crates/deltalake-core/src/table/mod.rs
```diff
@@ -208,6 +220,21 @@ impl DeltaTableMetaData {
...
+            None => Ok(ColumnMappingMode::None),
+            Some(None) => Ok(ColumnMappingMode::None),
+            Some(Some(s)) => match s.as_str() {
+                "none" => Ok(ColumnMappingMode::None),
+                "id" => Ok(ColumnMappingMode::Id),
+                "name" => Ok(ColumnMappingMode::Name),
+                _ => Err(DeltaTableError::InvalidColumnMappingMode(s.clone())),
+            },
+        }
+    }
```
roeap, line 237: Could we integrate this with the rest of the config in `crate::table::config`? After that we would ideally be reading that config via `state.table_config().column_mapping_mode()`

roeap (changes requested):
Looking really good!

One more ask is to handle the column mapping analogous to the rest of the table configuration.

crates/deltalake-core/src/table/mod.rs
aersam: I did that, and tried to align to code more with IsolationLevel/CheckpointPolicy
The error handling is a bit worse now, since the methods in table_config() always do unwrap_or_default instead of returning a Result, but I'd regard this as a separate topic.

roeap (approved):
Thanks for sticking with it and your patience @aersam!

aersam:
> Thanks for sticking with it and your patience @aersam!

Thanks for your guidance and for merging it!

Issue #1776, opened 2023-10-26
[RFC] Use arrow for checkpoint reading and state handling


roeap:
# Description

With the switch to updated schema / types in #1756 we also got quite some code to read checkpoints and more generally handle the table state as arrow record batches. In the past we tried several times to switch to arrow-based state.

I believe now is a good time, as we did a lot of the required work in kernel already. In this context I would also question the value of further maintaining a parquet2 variant of the checkpoint parsing or rather an arrow2 equivalent. IIRC, polars is currently vendoring large parts of the arrow2 codebase and as they are currently the main maintainers, there is likely little maintenance going forward.

AS there is a zero-copy interop between arrow and arrow2, I think we can just remove that support from our codebase, which would lead to - i be lieve - significantly reduced complexity.

**Use Case**

**Related Issue(s)**

ion-elgreco:
Polars also cloned pieces of arrow2 into Polars since the main maintainer of arrow2 pulled back. It's called polars-arrow now. So any activity on that arrow will probably also likely only happen into polars-arrow, unless folks back port stuff to arrow2.

houqp:
I agree with deleting the parquet2 support code to keep things simple given the fork. That said, I recommend taking the columnar based parsing approach in the current parquet2 implement for the new arrow-rs based state implementation for better performance.

roeap:
>  I recommend taking the columnar based parsing approach

Fully agree. My thinking was to start with loading most of the state in a first pass, and then moving to something I recently learned is called `PnM-Query` - i.e. trying our best to get the protocol and metadata as efficiently as possible. i.e. scanning all commit files and then selectively getting the data from the checkpoint if not present in any commit json file. and loadig file actions only when needed ...

Issue #425, opened 2021-09-07
Optionally ignore tombstones for read-only use cases


dispanser:
# Description

In a discussion for PR #420, https://github.com/delta-io/delta-rs/pull/420#discussion_r703270551 an idea came up to have a config option to ignore all tombstone entries entirely. Applications that are only reading do not actually use tombstones.
 
**Use Case**

This would reduce memory footprint, in particular for read-only applications.

@houqp writes:
> Letting read-only application specifying whether tombstones should be kept at all sounds like a neat trick to reduce memory usage :) I think it's better handled as a separate self-contained PR though because it's an optimization specific to a subset of readers, while the problem we are dealing where applies to both readers and write

I'd like to explore that idea further. We actually have a delta table that is mostly tombstones, which would see a major decrease in memory consumption.

houqp:
Extending on @dispanser 's idea, I think we could even extend it to all writers that doesn't handle checkpoints. By moving checkpoints to a dedicated writer/lambda function, we could even apply this optimization to all data writers.

mosyp:
An easy way would be just an additional option/flag. And if set, then we'd keep tomstones list empty, destipe it's writer or reader

dispanser:
I agree with @mosyp , let the user decide on instantiating the delta table.

However, it would be interesting if we're able to detect the misuse, i.e. panic / warn (?) when a writer attempts to write a checkpoint despite not having tombstones available. Not sure if that's easy to implement.

mosyp:
@dispanser If we stick with an optional flag, then if it's set, then prevent of creating a checkpoint and telling user a meaningful error

dispanser:
I'd make a quick foray into that topic if someone assigns this to me.

houqp:
yep, it should be easy to prevent side effect of this optimization by checking for the flag in both checkpoint and vacuum code path.

dispanser:
Now that I'm able to load my large tombstone-filled delta table, I plan on revisiting this issue. Still considering myself new to rust, I'm wondering how this feature could be represented in the public API. I assume that `pub async fn open_table(table_uri: &str)` and related functions form the public interface. If this was scala, I'd just tuck on an additional method parameter `ignoreTombstones: Boolen = false`, which would be backwards compatible but expose the new feature to everyone who wants to use it. What's the best practice in rust?

houqp:
[...]

Issue #288, opened 2021-06-10
Multi-part checkpoints


xianwill:
# Description

https://github.com/delta-io/delta-rs/pull/280 adds checkpoint write support, but does not write multipart checkpoints currently. We should do some additional research to identify a good split strategy and implement it.

Issue #1598, opened 2023-08-22
IsAdjustedToUtc flag not used while while reading DeltaTable


ion-elgreco:
# Environment

**Delta-rs version**: 0.10.1

**Binding**:  Python

**Environment**:
- **Cloud provider**: Azure
- **OS**: Linux
- **Other**:

***
# Bug

**What happened**:
When you write a PyArrow table with UTC timezone datetimes to delta table, the timezone information get's removed in the final table schema. See example below:

**What you expected to happen**:

Maintain UTC information by looking at the IsAdjustedToUTC flag in the parquet file and passing this back into the schema while reading to arrow. The information is there because when you read the partition directly with Polars, the timezone information is read as UTC:

```python
print(pl.read_parquet('/home/<redacted>/polars/py-polars/test/0-f7fe1e96-4adc-425a-b75c-80d4fdb1336c-0.parquet'))

shape: (1, 1)
┌─────────────────────────┐
│ datetime                │
│ ---                     │
│ datetime[μs, UTC]       │
╞═════════════════════════╡
│ 2010-01-01 00:00:00 UTC │
└─────────────────────────┘

```
**How to reproduce it**:
```python
import polars as pl
from deltalake import DeltaTable

df = pl.select(pl.datetime(2010, 1, 1, time_unit="us", time_zone="UTC"))

print(df.to_arrow().schema)
datetime: timestamp[us, tz=UTC]

df.write_delta('test')


dt = DeltaTable('test')

print(dt.schema().to_pyarrow())
datetime: timestamp[us]



```
**More details**:

ldacey:
Ah, just ran into this myself. I thought UTC timestamps were supported so I did:

```python
utc_tz = cs.datetime(time_zone="UTC")
non_utc_tz = cs.datetime() - utc_tz

df = df.with_columns(non_utc_tz.cast(pl.Datetime("us")), utc_tz.cast(pl.Datetime("us", "UTC")), cs.categorical().cast(pl.Utf8))
```
That adjusted my "ns" timestamp columns and I was hopefully going to leave the UTC timestamp intact. Generally I am saving data with the naive version (localized to whatever time zone the client is using with the TZ info stripped) and a UTC version.

My understanding is that at least UTC timestamps are supported, but just not with the delta-rs library currently?

ion-elgreco:
> Ah, just ran into this myself. I thought UTC timestamps were supported so I did:
> 
> ```python
[...]

PR #1837, opened 2023-11-11
feat: arrow based table state and checkpoint handling

roeap:
this is based on #1829 which is based on #1807 

# Description

Allows for reading and writing checkpoints more in line with current spec and updates the internal representation of the table state to use Arrow record batches. This is mainly in pursuit of trying to be more aligned with the ongoing work in kernel as well as moving our protocol support forward.

Still very much a work in progress, but opening it for feedback and discussion.

As a follow up I am planning on moving some of the updated code back to kernel.

# Related Issue(s)

supersedes: #454
closes: #1776
closes: #425 (should also be addressed in the current implementation)
closes: #288 (multi-part checkpoints are deprecated)
related: #1598 (maybe we can even close this)
related: #435

we may also want to consider closing: #125

# Documentation

On order to retain the parquet2 functionality and ease the transition I defined a `Snapshot` trait, that looks similar to a kernel snapshot. The new arrow backed state as well as the existing `DeltaTableState` implement this trait.

```rs
/// A snapshot of a Delta table at a given version.
pub trait Snapshot: std::fmt::Display + Send + Sync + std::fmt::Debug + 'static {
    /// The version of the table at this [`Snapshot`].
    fn version(&self) -> i64;

    /// Table [`Schema`](crate::kernel::schema::StructType) at this [`Snapshot`]'s version.
    fn schema(&self) -> Option<&StructType>;

    /// Table [`Metadata`] at this [`Snapshot`]'s version.
    fn metadata(&self) -> DeltaResult<Metadata>;

    /// Table [`Protocol`] at this [`Snapshot`]'s version.
    fn protocol(&self) -> DeltaResult<Protocol>;

    /// Iterator over the [`Add`] actions at this [`Snapshot`]'s version.
    fn files(&self) -> DeltaResult<Box<dyn Iterator<Item = Add> + '_>>;

    /// Well known table [configuration](crate::table::config::TableConfig).
    fn table_config(&self) -> TableConfig<'_>;
}
``` 

## log replay / v2 checkpoints

* handle compaction actions
* parse checkpoints with v2 metadata (`CheckpointMetadata` and `Sidecar` actions)
* add a new log path abstraction to set us on a path for supporting urls in delta log.

```rs
pub enum LogPath {
    ObjectStore(Path),
    Url(Url),
}
```
Usage is still fairly contained to the inner parts of delta-rs. Main usage right now is providing helpers for the various kinds of files in the log and using URLs will mostly just error. Nonetheless i felt it advantageous to start modelling this. 

## ~~timestamps~~ (moved to follow-up PR)

> non surprisingly, changes to timestamp handling propagate quite a bit, so better to do a follow up. Keeping the notes for reference.

The delta protocol states, that `timestamp` types should be written to parquet with the `isAdjustedToUtc` flag set to `true`. Right now delta-rs does not honour this. Right now I opted for:

* converting delta timestamps to arrow timestamps with tz="UTC"
* allow for converting arrow timestamps with utc to delta timestamps (previously only timestamps with tz = None were parsed)
* maintaining the behaviour of converting arrow timestamps with tz = `None` to delta timestamps until we implement `timestampNtz`.


houqp:
Really exciting work, glad to see this finally happen! Any early memory consumption benchmark results to share?

roeap:
superseded by #2037

PR #1840, opened 2023-11-11
ci: run doctest in CI for Python API examples

### LINKED ISSUE #1783: Run doctest in CI for Python API examples
# Description

I don't think our Python examples are being run in CI. We should add this to make sure they are validated on PRs. Pytest has support for it: https://docs.pytest.org/en/7.1.x/how-to/doctest.html

**Use Case**

**Related Issue(s)**

marijncv:
# Description
Enables running of doctest for python examples in docstrings. Also updates the docstrings where needed to make the test pass.

# Related Issue(s)
closes #1783

# Documentation


python/docs/source/usage.rst
```diff
@@ -476,8 +530,8 @@ to append pass in ``mode='append'``:
 
 .. code-block:: python
 
-    >>> write_deltalake('path/to/table', df, mode='overwrite')
-    >>> write_deltalake('path/to/table', df, mode='append')
+    >>> write_deltalake('path/to/table', df, mode='overwrite') # doctest: +SKIP
+    >>> write_deltalake('path/to/table', df, mode='append') # doctest: +SKIP
```
wjones127, line 534: Given the library is largely about reading and write from disk, it seems like it might be worth the effort to make this work. I was able to do this in lancedb/lance by configuring all doc tests to run in a temporary directory. This is just a few lines.

https://github.com/lancedb/lance/blob/84e4ce55f869a833c42239bfd6ba6176aa5840d3/python/python/lance/conftest.py#L18

Although the drawback is this might break the other tests that rely on reading specific pre-written tables in the repo. 🤔 I think that's a tradeoff I'd be willing to make. Let me know if you disagree.

python/docs/source/usage.rst
```diff
@@ -225,16 +223,42 @@ specified by the table configuration ``delta.logRetentionDuration``.
 To view the available history, use :meth:`DeltaTable.history`:
 
 .. code-block:: python
-
+    
+    >>> from pprint import pprint
```
wjones127, line 227: I like the usage of pprint btw. 👍

wjones127 (commented):
BTW this doctest applies to the user guide docs, but I was hoping to also have them run the ones within the docstrings, such as:

https://github.com/delta-io/delta-rs/blob/9b938308490a60808443ee36b4f7b2c3055a6c15/python/deltalake/table.py#L834-L858

roeap:
just wondering out loud, as I am not fully aware of the latest state. IIRC, we did merge some code migrating docs to mkdocs/markdown recently, albeit it not being released yet. Was the plan to deprecate the sphinx/rst docs once the switch is made? And would this work for the mkdocs docs as well?

python/docs/source/usage.rst
marijncv: Great suggestion! It seems to work well in [commit](https://github.com/delta-io/delta-rs/pull/1840/commits/dc71553f210036ab6556a19066baae4cc8fed7e9) (on my machine, not sure about CI machine yet)

marijncv:
> BTW this doctest applies to the user guide docs, but I was hoping to also have them run the ones within the docstrings,

Enabled the docstring tests in the later commits. Although perhaps it actually makes more sense to scope this PR to only the docstrings and leave out the sphinx docs (as probably was the intention of https://github.com/delta-io/delta-rs/issues/1783). Also given the comment by @roeap and the release of the [mkdocs](https://delta-io.github.io/delta-rs/). It would make the conftest.py setup a lot simpler too. Curious what you think!

roeap:
Agreed, the sphinx docs by now are officially no longer updated - at least we now display a header that says so 😆.

roeap (approved):

PR #1844, opened 2023-11-12
chore: upgrade dynamodb for rust-v0.16.x

rtyler:


roeap:
@rtyler - I think it would be best if we disable the repair_worker tests on that PR as well

roeap:
just a second too late - saw it merged as i typed 😆

PR #1848, opened 2023-11-13
fix: checkpoint error with Azure Synapse

### LINKED ISSUE #1847: Checkpoint error with Azure Synapse
# Environment

**Delta-rs version**: 0.12.0

**Environment**:
  - Azure Synapse
  - WSL2 - Debian
  - Python 3.11

***
# Bug
A deltatable is hosted in an Azure DataLake. When trying to read the data via a SQL query performed in a Synapse workspace, an error is raised due to certain fields in the checkpoint file.
The problematic fields are those whose values are "null".

# Solution
The solution is not to show fields with a 'null' value in the checkpoint json. 
To achieve this, optional fields in the Checkpoint struct must not be serialized if they are null or None.

PierreDubrulle:
# Description
Modify Checkpoint stuct not to serialize optional fields if null.

# Related Issue(s)
- closes #1847

### DIFF
--- a/crates/deltalake-core/src/operations/update.rs
+++ b/crates/deltalake-core/src/operations/update.rs
@@ -737,7 +737,6 @@ mod tests {
             .with_update("value", col("value") + lit(1))
             .await
             .unwrap();
-
         assert_eq!(table.version(), 1);
         assert_eq!(table.get_file_uris().count(), 1);
         assert_eq!(metrics.num_added_files, 1);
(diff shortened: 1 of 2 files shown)


rtyler (approved):
I double-checked [the protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#last-checkpoint-file-schema) for this one, and I think the skipping on serialization is fine. 

Kind of annoying that Azure Synapse is being doof, this [has also come up in discussions](https://github.com/delta-io/delta-rs/discussions/1611)

I trust that you've tested this change with Azure Synapse?

rtyler:
I have fixed some minor issues and rebased on top of `main`

PierreDubrulle:
@rtyler 
Yes, it's really the fields set to null that prevent the table from being read in Azure Synapse.

Skip_serializing_if is surely the best option to mitigate this problem, as it has the least impact on the rest of the code.

rtyler (approved):

PR #1850, opened 2023-11-13
docs: add docs on small file compaction with optimize

MrPowers:
Adds documentation page for small file compaction with optimize.


rtyler (approved):
Haven't fully processed the documentation but more is better at this point! :vulcan_salute:

PR #1854, opened 2023-11-13
fix: schema comparaison in writer

### LINKED ISSUE #1853: Schema comparaison in writer
# Environment

**Delta-rs version**: 0.12.0

**Environment**:
  - WSL 2
  - Python 3.10

***
# Bug
Schema comparison between the table schema and the schema passed as an argument to the write_deltalake function is too restrictive.
If schema two contains the same columns and values, but in a different order, then the exception ValueError: Schema of data does not match table schema is raised.

# Reproducible example
```import pandas as pd
import pyarrow as pa

testing_schema1: pa.lib.Schema = pa.schema(
    [
        pa.field("a", pa.int64()),
        pa.field("b", pa.int64()),
        
    ]
)

testing_schema2: pa.lib.Schema = pa.schema(
    [
        pa.field("b", pa.int64()),
        pa.field("a", pa.int64()),
        
    ]
)

df = pd.DataFrame({'b': [4, 5, 6, 7], 'a': [1, 2, 3, 5]})

write_deltalake(table_or_uri="delta.garbage", data=df, mode="append", schema=testing_schema1)
write_deltalake(table_or_uri="delta.garbage", data=df, mode="append", schema=testing_schema2)
```

***
# Solution
It's important not to make a naive comparison of the two schemas (the one in the table and the one provided for writing). It's better to hash the two by sorting them in order to compare them effectively.
It would also be interesting to be able to expose these hashes via a property of the DeltaTable class.

PierreDubrulle:
# Description
Improving scheme comparison

# Related Issue(s)

- closes #1853

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -4,6 +4,7 @@ import uuid
 from dataclasses import dataclass
 from datetime import date, datetime
 from decimal import Decimal
+from hashlib import md5
 from math import inf
 from pathlib import Path
 from typing import (
@@ -19,7 +20,6 @@ from typing import (
     Union,
 )
 from urllib.parse import unquote
-from hashlib import md5
 
 from deltalake.fs import DeltaStorageHandler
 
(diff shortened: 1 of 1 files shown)


PierreDubrulle:
Add a property to the DeltaTable class to expose the table schema hash

python/deltalake/writer.py
```diff
@@ -195,7 +204,12 @@ def write_deltalake(
         partition_by = [partition_by]
 
     if table:  # already exists
-        if schema != table.schema().to_pyarrow(as_large_types=large_dtypes) and not (
+        hash_table_schema = _hash_schema(
+            table.schema().to_pyarrow(as_large_types=large_dtypes)
+        )
+        hash_schema_provided = _hash_schema(schema)
+
+        if hash_schema_provided != hash_table_schema and not (
```
wjones127, line 212: I think we should compare actual objects and not just hashes. It is still possible for hashes to collide, despite that being unlikely.

You could write a method to normalize schema order:

```python
def normalize_schema_order(schema: pa.Schema) -> pa.Schema:
    fields = sorted(iter(schema), key=lambda field: field.name)
    return pa.schema(fields)
```

And then compare those. How does that sound?

PR #1859, opened 2023-11-14
feat: drop python 3.7 and adopt 3.12

### LINKED ISSUE #1703: Drop support for Python 3.7, test 3.12
# Description

Numpy and Pandas don't even support 3.8 anymore. We are safe dropping 3.7, and maybe can consider dropping 3.8 support.

We should definitely start testing with the latest Python version.

https://pypi.org/project/numpy/#files

**Use Case**

**Related Issue(s)**

roeap:
# Description

This change is mainly motivated by the current build failures on main. Since 3.7 is EOL and we have been meaning to drop it for a while, I thought maybe its good to give that a try before diving deeper into the inners of openssl build failures.

Closes #1703

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
 
       - name: Check Python
         run: |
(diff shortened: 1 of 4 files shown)


roeap:
@rtyler - i believe you may have to take that one, as it requires updating the required tests ...

In the end we were running the 3.7 tests in a manylinux container that we used to use also for builds. we have since then changed to the github action, so i felt using the approach we use in the other tests would be OK.

Just hoping that we don't run into the same issues again in the next release, as the action may as well use a similar manylinux image. But hopefully one, where all build dependencies etc for openssl are compatible with the versions we need in the build...

ion-elgreco (approved):

.github/workflows/python_build.yml
```diff
@@ -36,20 +36,19 @@ jobs:
...
       CARGO_INCREMENTAL: 0
 
-    # use the same environment we have for python release
-    container: quay.io/pypa/manylinux2014_x86_64:2022-09-24-4f086d0
     steps:
-      # actions/checkout@v3 is a node action, which runs on a fairly new
-      # version of node. however, manylinux environment's glibc is too old for
-      # that version of the node. so we will have to use v1 instead, which is a
-      # docker based action.
-      - uses: actions/checkout@v1
```
wjones127, line 52: Can leave this as a TODO, but I think the proper thing to do would be use the Maturin action to build in manylinux like we do for the release jobs.

wjones127 (approved):

PR #1863, opened 2023-11-14
fix: prevent writing checkpoints with a version that does not exist in table state

rtyler:
In some situations where the same writer is issuing append transactions using the operations API, which returns the newly created version, such as 10.

If the caller then attempts to create a checkpoint for version 10, the operation will produce an inconsistency in the `_last_checkpoint` file, if the callers in-memory table state has *not* been reloaded since the append operation completed.

In this scenario the _delta_log/ directory may contain: .
    ├── 00000000000000000000.json
    ├── 00000000000000000001.json
    ├── 00000000000000000002.json
    ├── 00000000000000000003.json
    ├── 00000000000000000004.json
    ├── 00000000000000000005.json
    ├── 00000000000000000006.json
    ├── 00000000000000000007.json
    ├── 00000000000000000008.json
    ├── 00000000000000000009.json
    ├── 00000000000000000010.checkpoint.parquet
    ├── 00000000000000000010.json
    └── _last_checkpoint

While `_last_checkpoint` contains the following:
    {"num_of_add_files":null,"parts":null,"size":342,"size_in_bytes":95104,"version":9}

This will result in an error on any attempts to read the Delta table:

    >>> dt = DeltaTable('.')
    [2023-11-14T18:05:59Z DEBUG deltalake_core::protocol] loading checkpoint from _delta_log/_last_checkpoint
    [2023-11-14T18:05:59Z DEBUG deltalake_core::table] update with latest checkpoint CheckPoint { version: 9, size: 342, parts: None, size_in_bytes: Some(95104), num_of_add_files: None }
    File "/home/tyler/venv/lib64/python3.11/site-packages/deltalake/table.py", line 253, in __init__
        self._table = RawDeltaTable(
                    ^^^^^^^^^^^^^^
    FileNotFoundError: Object at location /home/tyler/corrupted-table/_delta_log/00000000000000000009.checkpoint.parquet not found: No such file or directory (os error 2)
    >>>

To prevent this error condition, the create_checkpoint_for() function should ensure that the provided checkpoint version (used to write the `.checkpoint.parquet` file) matches the table state's version (used to write the `_last_checkpoint` file).

This has the added benefit of helping prevent the caller from passing in a nonsensical version number that could also lead to a broken table.

### DIFF
--- a/crates/deltalake-core/src/protocol/checkpoints.rs
+++ b/crates/deltalake-core/src/protocol/checkpoints.rs
@@ -38,6 +38,10 @@ enum CheckpointError {
     #[error("Partition value {0} cannot be parsed from string.")]
     PartitionValueNotParseable(String),
 
+    /// Caller attempt to create a checkpoint for a version which does not exist on the table state
+    #[error("Attempted to create a checkpoint for a version {0} that does not match the table state {1}")]
+    StaleTableVersion(i64, i64),
+
     /// Error returned when the parquet writer fails while writing the checkpoint.
     #[error("Failed to write parquet: {}", .source)]
     Parquet {
@@ -60,6 +64,7 @@ impl From<CheckpointError> for ProtocolError {
         match value {
             CheckpointError::PartitionValueNotParseable(_) => Self::InvalidField(value.to_string()),
             CheckpointError::Arrow { source } => Self::Arrow { source },
+            CheckpointError::StaleTableVersion(..) => Self::Generic(value.to_string()),
             CheckpointError::Parquet { source } => Self::ParquetParseError { source },
         }
     }
@@ -117,6 +122,14 @@ pub async fn create_checkpoint_for(
     state: &DeltaTableState,
     log_store: &dyn LogStore,
 ) -> Result<(), ProtocolError> {
+    if version != state.version() {
+        error!(
+            "create_checkpoint_for called with version {version} but table state contains: {}. The table state may need to be reloaded",
+            state.version()
+        );
+        return Err(CheckpointError::StaleTableVersion(version, state.version()).into());
+    }
+
     // TODO: checkpoints _can_ be multi-part... haven't actually found a good reference for
     // an appropriate split point yet though so only writing a single part currently.
     // See https://github.com/delta-io/delta-rs/issues/288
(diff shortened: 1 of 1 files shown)


crates/deltalake-core/src/protocol/checkpoints.rs
```diff
@@ -117,6 +122,14 @@ pub async fn create_checkpoint_for(
     state: &DeltaTableState,
     log_store: &dyn LogStore,
 ) -> Result<(), ProtocolError> {
+    if version != state.version() {
+        error!(
+            "create_checkpoint_for called with version {version} but table state contains: {}. The table state may need to be reloaded",
+            state.version()
+        );
+        return Err(CheckpointError::StaleTableVersion(version, state.version()).into());
```
wjones127, line 130: Why can't we just checkout the correct version?

rtyler: @wjones127 how do you mean "checkout?" I am less versed in the nuance of checkpoints. If the last version in the table is 100, is it valid to call this with `_last_checkpoint` of `version: 5`?

I honestly don't know whether we should even allow a user-specified version here at all, rather than creating a checkpoint with the version of the table state.

HOWEVER, if a writer has a stale table state creates `_last_checkpoint` after _another_ writer has created a later checkpoint, this becomes confusing and out of sync anyways :confounded:

rtyler: I guess to expand, I like the failure option here because the caller is asking us to do something here that will result in incorrect state, and we don't know what they intend or what performance concerns they have. I.e. I don't think we should reload the state here, as an example, because that would be surprising and may actually result in a later version than the user specified too

xianwill: I feel like (a) failing is the easiest solution that avoids any potential new bug introduction and (b) reloading correct version could have unexpected perf impact, esp for things like streaming workloads that are managing their own table state

rtyler: oh one more thing since my thoughts are disjointed here :laughing: I think we already have the path you describe covered with `create_checkpoint()`  :smile:

wjones127: Okay re-read the signature. Erroring makes sense to me.

I was thinking in the situation you were describing in the issue report, it sounds like the correct thing to do is to checkout the desired version first then create the checkpoint.

wjones127 (approved):

roeap (approved):
makes a lot of sense :).

Currently there also is a draft PR in the works to be able to handle v2 checkpoints and honour the latest recommendations from the protocol.

Issue #1858, opened 2023-11-14
When will upgrade object_store to 0.8?


RobinLin666:
# Description

**Use Case**

There are a lot of features and bug fixes in object_store 0.8, when it will upgrade?

**Related Issue(s)**

rtyler:
We're more or less beholden to our transitive dependencies like  [datafusion](https://crates.io/crates/datafusion/32.0.0/dependencies) upgrading unfortunately.

I'm hoping to see a datafusion 33 in the next week or two

RobinLin666:
> We're more or less beholden to our transitive dependencies like [datafusion](https://crates.io/crates/datafusion/32.0.0/dependencies) upgrading unfortunately.
> 
> I'm hoping to see a datafusion 33 in the next week or two

Hi @rtyler, I saw the [datafusion](https://crates.io/crates/datafusion/33.0.0/dependencies) 33 is ready, but it still use 0.7 object_store, it means that delta-rs can only use 0.7 version of object_store? :-(

ion-elgreco:
@rtyler :)

Issue #1765, opened 2023-10-24
Support writes to mounted local path with blobfuse


RobinLin666:
# Description

**Use Case**
Re-open an issue for tracking https://github.com/delta-io/delta-rs/issues/1418#issuecomment-1769921086. Thank you.

```python
import pandas as pd
from deltalake.writer import write_deltalake

df = pd.DataFrame(
    {
        "a": [1, 2, 3, 4, 5],
        "fruits": ["banana", "orange", "mango", "apple", "banana"],
    }
)

write_deltalake('/lakehouse/default/Files/fruits', df,mode='overwrite')
```
<img width="816" alt="image" src="https://github.com/delta-io/delta-rs/assets/128118209/56da2a0b-2414-4a6a-86ae-57f3762d60a0">

**Related Issue(s)**
https://github.com/delta-io/delta-rs/issues/1418#issuecomment-1769847878
https://github.com/GlareDB/glaredb/issues/1809

scsmithr:
For what it's worth, I was testing with this change to object store: https://github.com/GlareDB/arrow-rs/pull/2/files

The first diff chunk explicitly drops the file so that the metadata gets flushed to blobstore before the rename.

The second chunk I'm not sure is necessary. I was just trying out explicitly syncing the file, but I don't recall it actually fixing the issue.

RobinLin666:
Thanks @scsmithr, I tried to mitigate the issue from blobfuse side.

for #1 rename issue can be mitigated. I upload the tmp file before renaming.

for #2 you said blobfuse doesn't support hard linking, `copy_if_not_exists` will fail. I don't quite understand how this aspect is affected and whether it is the cause of the error I encountered above. `OSError: Generic DeltaLocalObjectStore error: Function not implemented (os error 38)`  If possible, please let me know where it need a hard link and we can think about any workarounds available.

I saw the same error when writing to DBFS, perhaps it's one of the reasons?
https://github.com/delta-io/delta-rs/issues/987

Issue #1376, opened 2023-05-17
dbfs paths not supported


MrPowers:
# Environment

**Delta-rs version**: 0.9.0

**Binding**:  Python

**Environment**:
- **Cloud provider**: Databricks
- **OS**: ? 
- **Other**: ?

***
# Bug

**What happened**: Tried to instantiate a DeltaTable from a DBFS path, like this: `deltalake.DeltaTable("dbfs:/some-thing/some_dir")`

**What you expected to happen**: I expected this to work.  This works: `spark.read.format("delta").load("dbfs:/some-thing/some_dir").show()`

**How to reproduce it**: Create a Delta table in Databricks with a DBFS path and then try to instantiate a `deltalake.DeltaTable`.  Should be relatively easy to reproduce.

**More details**: N/A

rtyler:
@MrPowers to the best of my knowledge there is not a REST API for DBFS or any such open "file system provider" for what DBFS actually. Does Databricks make it possible for third party interoperability with DBFS

MrPowers:
@rtyler - yea, I'm not sure.  Perhaps I have to figure out another way to get the path to the data.

Lundez:
I have the same issue when using a mounted ADSL2 in a Azure ML Studio job. I wish to write, and it fails on writing the log. The parquet-file is correctly written.

This is ADSL2 with Hierarchial Storage.

ion-elgreco:
> I have the same issue when using a mounted ADSL2 in a Azure ML Studio job. I wish to write, and it fails on writing the log. The parquet-file is correctly written.
> 
> This is ADSL2 with Hierarchial Storage.

I also ran into this issue with AML, writing to mounted storage is not supported. 

The way I do it now is I don't mount but write to the adls2 container directly.

Lundez:
> > I have the same issue when using a mounted ADSL2 in a Azure ML Studio job. I wish to write, and it fails on writing the log. The parquet-file is correctly written.
> > 
> > This is ADSL2 with Hierarchial Storage.
> 
> I also ran into this issue with AML, writing to mounted storage is not supported. 
> 
> The way I do it now is I don't mount but write to the adls2 container directly.

I solved it the same way, but that means my jobs aren't as clear (output is not job output but a hidden API call) 😅

Thanks for responding!

ion-elgreco:
Should work now for mounted storage with change by https://github.com/delta-io/delta-rs/pull/1868

PR #1868, opened 2023-11-16
fix: compatible to write to local file systems that do not support hard link

RobinLin666:
compatible to write to local file systems that do not support hard link.

# Description

When we write to the local file system, sometimes hard link is not supported, such as blobfuse, goofys, s3fs, so deal with it with compatibility.

It is important to note that:
There is another problem with blobfuse, that is, when it comes to rename, it will report errors. Because rename did not release the file handle before.
See here for details: https://github.com/delta-io/delta-rs/issues/1765

Arrow-rs is required to cooperate with the modification, for example: https://github.com/GlareDB/arrow-rs/pull/2/files
Because object_store has been upgraded to 0.8, there are a lot of breaking change, so I haven't changed this one for the time being. Will fix it after upgrading to 0.8 https://github.com/delta-io/delta-rs/issues/1858

# Related Issue(s)

#1765
 
#1376 

# Documentation

### DIFF
--- a/crates/deltalake/Cargo.toml
+++ b/crates/deltalake/Cargo.toml
@@ -14,7 +14,7 @@ rust-version.workspace = true
 [package.metadata.docs.rs]
 # We cannot use all_features because TLS features are mutually exclusive.
 # We cannot use hdfs feature because it requires Java to be installed.
-features = ["azure", "datafusion", "gcs", "hdfs", "json", "python", "s3", "unity-experimental"]
+features = ["azure", "datafusion", "gcs", "hdfs", "json", "mount", "python", "s3", "unity-experimental"]
 
 [dependencies]
 deltalake-core = { version = "0.17.1", path = "../core" }
@@ -35,11 +35,11 @@ gcs = ["deltalake-gcp"]
 glue = ["deltalake-catalog-glue"]
 hdfs = []
 json = ["deltalake-core/json"]
+mount = ["deltalake-mount"]
 python = ["deltalake-core/python"]
 s3-native-tls = ["deltalake-aws/native-tls"]
 s3 = ["deltalake-aws/rustls"]
 unity-experimental = ["deltalake-core/unity-experimental"]
-mount = ["deltalake-mount"]
 
 [dev-dependencies]
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
--- a/crates/deltalake/Cargo.toml
+++ b/crates/deltalake/Cargo.toml
@@ -22,6 +22,7 @@ deltalake-aws = { version = "0.1.0", path = "../aws", default-features = false, 
 deltalake-azure = { version = "0.1.0", path = "../azure", optional = true }
 deltalake-gcp = { version = "0.1.0", path = "../gcp", optional = true }
 deltalake-catalog-glue = { version = "0.1.0", path = "../catalog-glue", optional = true }
+deltalake-mount = { version = "0.1.0", path = "../mount", optional = true }
 
 [features]
 # All of these features are just reflected into the core crate until that
@@ -38,6 +39,7 @@ python = ["deltalake-core/python"]
 s3-native-tls = ["deltalake-aws/native-tls"]
 s3 = ["deltalake-aws/rustls"]
 unity-experimental = ["deltalake-core/unity-experimental"]
+mount = ["deltalake-mount"]
 
 [dev-dependencies]
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
--- a/crates/deltalake/src/lib.rs
+++ b/crates/deltalake/src/lib.rs
@@ -9,3 +9,5 @@ pub use deltalake_aws as aws;
 pub use deltalake_azure as azure;
 #[cfg(feature = "gcs")]
 pub use deltalake_gcp as gcp;
+#[cfg(feature = "mount")]
+pub use deltalake_mount as mount;
--- a/crates/mount/Cargo.toml
+++ b/crates/mount/Cargo.toml
@@ -1,0 +1,43 @@
+[package]
+name = "deltalake-mount"
+version = "0.1.0"
+authors.workspace = true
+keywords.workspace = true
+readme.workspace = true
+edition.workspace = true
+homepage.workspace = true
+description.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+deltalake-core = { version = "0.17.0", path = "../core", features = [
+    "datafusion",
+] }
+lazy_static = "1"
+errno = "0.3"
+
+# workspace depenndecies
+async-trait = { workspace = true }
+bytes = { workspace = true }
+futures = { workspace = true }
+tracing = { workspace = true }
+object_store = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true }
+regex = { workspace = true }
+url = { workspace = true }
+
+[dev-dependencies]
+chrono = { workspace = true }
+serial_test = "3"
+deltalake-test = { path = "../test" }
+pretty_env_logger = "0.5.0"
+rand = "0.8"
+serde_json = { workspace = true }
+tempfile = "3"
+fs_extra = "1.3.0"
+
+[features]
+integration_test = []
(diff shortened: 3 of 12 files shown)


roeap (commented):
Thanks for looking into this @RobinLin666.

AFAIK, regular rename will relace the original file if it already exists. As such it is not safe to use for delta commits.

For S3 lock stores we have an setting `ALLOW_UNSAFE_REANME` to allow writing to S3 without a lock client. I would be fine adding this if we also require that setting to be set, such that users are aware they themselves have to make sure there are no concurrent writers to the table.

RobinLin666:
Thanks for your suggestion, @roeap.

I've added a new option, `FILE_ALLOW_UNSAFE_RENAME`, to control whether to enable the `rename` operation. Additionally, I wanted to confirm that the `rename` operation is atomic and doesn't alter the file's inode. Could you provide more details on the specific scenarios where safety concerns might arise? 

Looking forward to your insights.

roeap:
Certainly ... `std::fs::rename` will [replace](https://doc.rust-lang.org/std/fs/fn.rename.html) the file at the target location if it already exists. The concurrency mechanics in delta heavily rely on that not happening :).

The basic idea to make commits atomic is to write a temporary file, renaming it to the desired commit file, and relying on the file system to error if the file already exists.

RobinLin666:
Yes, if the target already exists, it will replace it. But in this scenario, we require that the target file does not exist, if it does, it will report an error directly, and the next rename operation will not be performed, so I think it is still atomic.

roeap:
Well, we went over this several times, also with the core team and general consensus was that an a-priori check without any lock mechanism would not prevent any race conditions.

I think you might be able to validate this by adopting the s3 concurrency tests.

RobinLin666:
Yes! You are right! Maybe I should consider use `nix::fcntl::{flock, FlockArg};` to have a simple file lock.

RobinLin666:
Hi @roeap, do you have any other concerns? I can modify them at any time. Please feel free to review, thanks!

ion-elgreco:
@RobinLin666 what's the impact if multiple people write at the same time to a mounted storage?

RobinLin666:
> @RobinLin666 what's the impact if multiple people write at the same time to a mounted storage?

Hi @ion-elgreco 
Here is the doc of blobfuse,
image

writing to different directories only from different instances" will work if you set --use-attr-cache=false and --file-cache-timeout-in-seconds=0. But cannot write to a same blob/file.

Here is some detail: https://github.com/Azure/azure-storage-fuse/issues/366 .

Lyndon1994:
kindly ping

RobinLin666:
Hi team, just wanted to kindly bump this PR for review. Thank you!

djouallah:
it will be really nice to have this functionality as it is very useful in Fabric notebook please :)

RobinLin666:
Since delta-rs has been refactored, I have also rearranged the code here. Please review it again. Thank you!

Note: To completely fix this problem, you need to upgrade arrow-rs object_store to version 0.9. I wrote delta table to mounted path in my local test and it passed.

RobinLin666:
Hi @MrPowers, could you please help to review the PR, since you refined the code last week. Thanks

RobinLin666:
Hi guys, we are preparing a new Fabric Notebook which has a pure python environment, so we want to depend on this library to provide users with an enjoyable experience to operate Lakehouse Tables, whether through abfss path or local mounted path.
So, can you sign off this PR?

jimdowling:
This would be a really useful PR for other file systems like HDFS.

RobinLin666:
Excuse me, Hi @roeap / @wjones127 / @rtyler / @fvaleye / @ion-elgreco / @MrPowers Could you please help to review the PR? Thank you very much!

RobinLin666:
Excuse me, Hi @roeap / @wjones127 / @rtyler / @fvaleye / @ion-elgreco / @MrPowers 
Does this PR still have a chance to join?
If it has a chance, can you please help review, because I have resolved many conflicts, it can't go on like this, right?
If it doesn't have a chance, please tell me why, or is there any other plan? So I can do it in time. Because our new project is very dependent on this open-source project.
Thanks!

RobinLin666:
Excuse me, Hi @roeap / @wjones127 / @rtyler / @fvaleye / @ion-elgreco / @MrPowers 
Does this PR still have a chance to join?
If it has a chance, can you please help review, because I have resolved many conflicts, it can't go on like this, right?
If it doesn't have a chance, please tell me why, or is there any other plan? So I can do it in time. Because our new project is very dependent on this open source project.

RobinLin666:
😊

RobinLin666:
Hi @ion-elgreco can you help to sign off the PR? Thanks!

ion-elgreco:
@RobinLin666 can you update the correct docs pages? The .rst file are the old docs

RobinLin666:
> @RobinLin666 can you update the correct docs pages? The .rst file are the old docs

Nice to see your reply. I just updated it. Thank you!

crates/core/src/storage/file.rs
```diff
@@ -375,6 +472,10 @@ mod imp {
         }
     }
 
+    pub(super) async fn regular_rename(from: &str, to: &str) -> Result<(), LocalFileSystemError> {
+        rename_noreplace(from, to).await
```
ion-elgreco, line 476: Why is this calling rename_noreplace here?

RobinLin666: For macOS, platform_specific_rename function will be invoked instead of utilizing hard links, as macOS does not support fuse file systems. Therefore, it's advisable to use rename_noreplace in this context.

RobinLin666:
Hi @ion-elgreco Have a good day!
Is there any problem/ question for now?

RobinLin666:
> HEy @RobinLin666 @ion-elgreco, sorry for being MIA for a while.
> 
> Since blob-fuse is a valid use case, I guess we have to do simething about it. generally I was hoping to get rind of all custom file system implementation in delta-rs, but it seems we will have to retain something after all.
> 
> One thing though - could we make the config handling analogus to how we handle configuration in object store and also our itnernal config - e.g. `table/config.rs`?

Hi @roeap thank you for your suggestion, I wrote it with reference to the implementation of aws, supporting environment variables or passing parameters to pass the configuration. 
I don't fully understand the effect you want, maybe I'm not very familiar with delta-rs code. I hope you can explain it in more detail. Thank you. If possible, I'd like to go in this time and reconstruct it later.

RobinLin666:
Hi @roeap ?

RobinLin666:
Hello @roeap is there any advice ?

RobinLin666:
Hi @roeap  , we greatly rely on this PR as many users strongly demand the ability to write data to the mount point, not only with blobfuse but also potentially with tools like s3fs, Databricks' dbfs, and others. If you believe that the code needs refactoring, could you provide more detailed guidance on how it should be refactored? Alternatively, it would be ideal if we could proceed with this PR first, and then I can prepare a refactoring PR. I prefer not to privately compile a delta-rs installation for our product, as it would be cumbersome to maintain and not practical. Thank you.

ion-elgreco:
@RobinLin666 I think what Robert means it to make this akin to crates/azure/src/config.rs, in parsing and setting configuration. Does that help?

RobinLin666:
Hi @roeap / @ion-elgreco , I have refine the code, please help to review, thanks !

ion-elgreco:
@RobinLin666 Looks good! Can you fix the test failures?

RobinLin666:
Hi @roeap / @ion-elgreco  can you help to review again?
Thank you very much!

ion-elgreco:
@RobinLin666 I don't see a test anymore in Python, could you add one?

RobinLin666:
> @RobinLin666 I don't see a test anymore in Python, could you add one? Once that's there we are good to go.

Thanks @ion-elgreco, done.

ion-elgreco (dismissed):
@RobinLin666 thanks!

RobinLin666:
Hi @ion-elgreco thank you very much for your approval!!!
I just fixed a test case for dbfs, because it cannot create dir in /dbfs.
image

ion-elgreco (approved):

Issue #1867, opened 2023-11-15
Fix docs deployment


wjones127:
# Description

https://github.com/delta-io/delta-rs/actions/runs/6880582444/job/18715123279

Likely a matter of which credentials the job is running under.

**Use Case**

**Related Issue(s)**

r3stl355:
I need to do some experiments with this.

@wjones127 do you have write access to `gh-pages` branch? (I presume you do). 

The failing action is https://github.com/JamesIves/github-pages-deploy-action, responsible for deploying the build results to a `gh-pages` branch. According to the action's docs, `you must ... provide Read and Write Permissions to the provided GITHUB_TOKEN` but I am not sure which GITHUB_TOKEN was used when the action was triggered manually. 
Previous deployment scripts used `github.GITHUB_TOKEN` to do the same.

The action has a `token` property which I can try setting to `github.GITHUB_TOKEN` but I am not sure it will make any difference as it seems to be a default value.

The other alternative is to add `permissions: contents: write` to the action - that should work but will result in a longer script as I'd have create a separate step for deployment.

Lastly, during the release process docs release is run as a dispatched/async job so even if it fails, the main release should not be effected.

r3stl355:
take

PR #1869, opened 2023-11-16
fix: docs deployment action

r3stl355:
# Description
Moved the requirement for contents write permission from release action to docs deployment. The same permission was  used in the previous release action (sphinx) so I'm not introducing any extra permission requirements. 

One thing I don't really like here is that such permission is only needed for the `Deploy` step but the lowest level it can be defined is job. This should be fine because other steps in the job don't do any git operations.

Alternative would be creating a separate Deploy job but then I'd have to duplicate all the steps from the Build part which would effectively be the same thing as now but with a longer/redundant config.

Deployment logic remains the same - docs are deployed only when the Build Documentation is run manually or via the Release Python action.

# Related Issue(s)
 #1867

I am unable to fully test this in my environment because the failing version was actually working in my repo - I don't see how things are configured in the main repo.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,13 +35,15 @@ jobs:
         with:
           src: docs/src/python
 
-  build:
+  build-deploy:
     needs:
       [
         lint,
         markdown-link-check,
       ]
     runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
I think this should work.

Issue #1861, opened 2023-11-14
Add a Rust-backed engine for write_deltalake


wjones127:
# Description

Right now we've built on top of the PyArrow writers. This requires a lot of complex code that is essentially duplicating logic in Rust. The main motivation for writing it was that the Rust implementation wasn't ready, so it was faster to build on top of PyArrow. That might not be true anymore.

We can update the signature of `write_deltalake()` to take an `engine` parameter (sort of like how Pandas [read_parquet](https://pandas.pydata.org/docs/reference/api/pandas.read_parquet.html) has this parameter), which would let users choose to use the pyarrow engine or the Rust engine for now. Eventually we can switch the default and deprecate the pyarrow implementation.

We should be on the lookout for issues that block this. First, we need to make sure the same unit tests pass with the new writer. So we should parametrize all tests by the engine.

Second, we should be on the lookout for performance issues. We have a set of benchmarks here: https://github.com/delta-io/delta-rs/blob/56a97289201dc05ca2a1a94953b82efdc6bb3b5f/python/tests/test_benchmark.py#L28, which we can add to. One known issue (that might have been solved) is: https://github.com/delta-io/delta-rs/issues/1225

**Use Case**

**Related Issue(s)**

roeap:
we should use this opportunity to also consolidate our writer implementations. right now we have one in `/writer` as well as `/operations/writer`. The one in operations being the newer one. Originally it was meant to have two implementations only for a short while, but this has been the case for many months now 😆.

The `/writer` one was mainly kept around to keep supporting the kafka-delta-ingest use case to also be able to writer json data.

ion-elgreco:
@roeap I am mainly exposing operations/write to Python, not writer

roeap:
I'll re-open this to keep track until we have a more full-featured rust writer?

ion-elgreco:
@roeap yeah good one!

PR #1872, opened 2023-11-16
feat(python): expose rust writer as additional engine

ion-elgreco:
# Description
First version of a functional python binding to rust writer. Lots of work is happening around the writer. This one we should merge first: https://github.com/delta-io/delta-rs/pull/1820.

A couple gaps will exist between current Rust writer and pyarrow writer. We will have to solve this in a later PR,:
- Overwrite schema in Rust 
- Replacewhere (partition filter / predicate) overwrite

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1861


ion-elgreco:
@wjones127 Any early feedback on how to tackle the missing features and whether I am overseeing something would be much appreciated 😄

python/src/lib.rs
```diff
@@ -1127,6 +1128,71 @@ impl From<&PyAddAction> for Add {
...
+    mode: String,
+    max_rows_per_group: i64,
+    overwrite_schema: bool,
+    partition_by: Option<Vec<String>>,
+    name: Option<String>,
+    description: Option<String>,
+    configuration: Option<HashMap<String, Option<String>>>,
+    storage_options: Option<HashMap<String, String>>,
+) -> PyResult<()> {
+    let batches = data.0.map(|batch| batch.unwrap()).collect::<Vec<_>>();
```
roeap, line 1146: We should probably try to not fully materialize the RBR into memory. This would entail wrapping the RBR in an `ExecutionPlan`, which can be passed to the `CreateBuilder`. Maybe [`DatasetExec`](https://github.com/apache/arrow-datafusion-python/blob/main/src/dataset_exec.rs) from datafusion-python can provide some inspiration.

python/src/lib.rs
```diff
@@ -1127,6 +1128,71 @@ impl From<&PyAddAction> for Add {
...
+        .block_on(DeltaOps::try_from_uri_with_storage_options(
+            &table_uri, options,
+        ))
+        .map_err(PythonError::from)?;
+
+    let mut builder = table
+        .write(batches)
+        .with_save_mode(save_mode)
+        .with_overwrite_schema(overwrite_schema)
+        .with_metadata(metadata)
```
roeap, line 1181: This name is a little bit misleading. It does not refer to the metadata action, but rather additional application-specific metadata. Took a quick look at the code, but it seems table configuration it not passed through to the create operation in the "create if not exists" path.

__update__: just saw you mentioned that already :).

crates/deltalake-core/src/operations/write.rs
```diff
@@ -108,6 +108,8 @@ pub struct WriteBuilder {
     write_batch_size: Option<usize>,
     /// RecordBatches to be written into the table
     batches: Option<Vec<RecordBatch>>,
+    /// whether to overwrite the schema
+    overwrite_schema: bool,
```
roeap, line 112: is this meant for schema evolution? If so, I'd recommend moving that to a follow-up PR as it would likely blow up this PR quite a bit.

roeap (commented):
Certainly :). 

Overall I believe this is going in the right direction. one thing that may be a larger piece of work is not collecting the iterator, but rather wrapping it in an execution plan. Maybe we could also live with collect for now to get the wiring right, and leave the execution plan to a follow-up.

python/deltalake/writer.py
wjones127, line 1: To make the diff as clean as possible, could we instead change `write_deltalake` to dispatch to either `_write_deltalake_pyarrow` (current impl) or `_write_deltalake_rust` (new impl)?

python/tests/test_writer.py
```diff
@@ -29,24 +29,30 @@
...
 
     with pytest.raises(OSError) as exception:
-        write_deltalake(tmp_path, sample_data, mode="overwrite")
+        write_deltalake(tmp_path, sample_data, mode="overwrite", engine=engine)
 
     assert "directory is not empty" in str(exception)
 
 
-def test_roundtrip_basic(tmp_path: pathlib.Path, sample_data: pa.Table):
+@pytest.mark.parametrize("engine", ["pyarrow", "rust"])
```
wjones127, line 48: Thanks, this is exactly what I had in mind 😄

crates/deltalake-core/src/operations/write.rs
wjones127: +1. I think it's fine if we let that return `NotImplementedError` for now.

python/src/lib.rs
ion-elgreco: Maybe we can do that in a second PR then, since I was also doing it like this for MERGE python binding. Then I can improve these two at the same time.

crates/deltalake-core/src/operations/write.rs
ion-elgreco: It was an quick attempt for schema evolution. I was able to write except it didn't write the columns that were not part of the original schema, so I need to dig through the code more.

Ok, let's do this as improvement in another update.

python/deltalake/writer.py
ion-elgreco: Yes can do that!

crates/deltalake-core/src/operations/write.rs
roeap: we also would need to update all read path to always add null columns for columns non-existent in older parquet files. Haven't looked into it, but this would likely require some larger refactoring particularly in the datafusion `DeltaScan`. Saying this we likely need to validate that added columns are always nullable as well.

ion-elgreco: I think that would be schema evolution for also appends.

The PyArrow writer can do schema evolution but only combined with an `overwrite` mode. 

I think that's purely a metadata action then. Would this be doable with the existing deltalake-core crate?

roeap: not sure, we may end up with unreadble tables if we do this... if we replace the whole table this might work.

ion-elgreco: With PyArrow it only works together with overwrite so it should be safe. Is there a way to adjust the commit that's written during a write?

ion-elgreco:
@roeap @wjones127 it's ready for a full review now. That one failing test will be solved when we merge this PR https://github.com/delta-io/delta-rs/pull/1820, I've also updated the PR description, basically we have two gaps with pyarrow and rust writer.

roeap:
While looking through this I realized we may have a bug with the schema overwrite right now.

When we added this, overwrite would always replace the entire table. Since then we added selectively overwriting partitions, in which case I believe we may be corrupting the table, or at least make it unreadable to our readers, since we end up with parquet files with different schemas in the table.

Not sure if the pyarrow Dataset can handle that, but almost certain, our rust readers would not be able to handle that.

Should we include as an additional check, that no partition filters are supplied, if the schema is overwritten?

ion-elgreco:
> While looking through this I realized we may have a bug with the schema overwrite right now.
> 
> When we added this, overwrite would always replace the entire table. Since then we added selectively overwriting partitions, in which case I believe we may be corrupting the table, or at least make it unreadable to our readers, since we end up with parquet files with different schemas in the table.
> 
> Not sure if the pyarrow Dataset can handle that, but almost certain, our rust readers would not be able to handle that.
> 
> Should we include as an additional check, that no partition filters are supplied, if the schema is overwritten?

For the rust writer? Because currently only pyarrow is using them.

roeap:
> For the rust writer? Because currently only pyarrow is using them.

Yes, and the pyarrow writer may be creating corrupted tables that we cannot read from rust and maybe even python if partition filters are supplied.

ion-elgreco:
Closing this in favour of: https://github.com/delta-io/delta-rs/pull/1891

Issue #1568, opened 2023-07-27
Path to reader version 3, writer version 7 transaction log protocol support


MrPowers:
# Description

Delta Lake transitioned to [table features](https://delta.io/blog/2023-07-27-delta-lake-table-features/) and now has a better upgrade path for connectors.

delta-rs can upgrade to protocol version 3,7 (skipping the intermediate protocol versions) and then cherry pick the table features that will be supported.

Adding table feature support may also make it safer when reading Delta tables with delta-rs.  Does delta-rs gracefully error out when reading a table with table features that are enabled that delta-rs does not currently support?

The purpose of this issue is to organize the remaining work that's needed to get protocol version 3,7 support.  Seems like delta-rs is close.

wjones127:
> Does delta-rs gracefully error out when reading a table with table features that are enabled that delta-rs does not currently support?

This should happen right now, so if anyone is getting incorrect data back from delta-rs, we should consider that a bug.

MrPowers:
Once [this](https://github.com/delta-io/delta-rs/pull/1807) is merged, we should have full support for append only tables.

roeap:
> does that mean the rust writer automatically checks it?

This check is done in the `commit` function, since we are also committing actions to the protocol via that function it should cover all code paths...

hntd187:
Okay so I have to check the write paths are respected for invariants and fix that if necessary to ensure we have invariant support throughout. Then once that is done I'll start implementing check constraints.

MrPowers:
@hntd187 - can we claim reader version 3, writer version 7 transaction log protocol support now?

hntd187:
I believe what I signed up to do is done and dusted, so barring any other support we need to support it, I believe so.

PR #1882, opened 2023-11-18
feat: add enforcement for invariants to commit checker

### LINKED ISSUE #1880: Check Invariants are respecting table features for write paths
# Description

Check Invariants are respecting table features for write paths, @MrPowers if you can assign this to me so others know.

**Use Case**

This is for proper 3,7 support from #1568 

**Related Issue(s)**

#1568

hntd187:
# Description
Added invariant checking in the commit checker.

# Related Issue(s)
#1568 closes #1880

# Documentation


crates/deltalake-core/src/operations/transaction/protocol.rs
```diff
@@ -142,6 +142,25 @@ impl ProtocolChecker {
             })?;
         }
 
+        // Does the table have invariants?
+        let invariants = snapshot
+            .current_metadata()
+            .and_then(|meta| meta.schema.get_invariants().ok())
+            .unwrap_or_default()
+            .is_empty();
```
wjones127, line 150: `invariants = false` meaning "it does have invariants" does feel very intuitive. Could we reverse it, and rename so it's a little clearer?

```suggestion
        let has_invariants = !snapshot
            .current_metadata()
            .and_then(|meta| meta.schema.get_invariants().ok())
            .unwrap_or_default()
            .is_empty();
```

crates/deltalake-core/src/operations/transaction/protocol.rs
```diff
@@ -142,6 +142,25 @@ impl ProtocolChecker {
...
+
+        // The table is at least version 7, has actual invariants and has the proper writer feature
+        // enabled
+        if snapshot.min_writer_version() >= 7
+            && !invariants
```
wjones127, line 155:
```suggestion
            && has_invariants
```

wjones127 (changes requested):

crates/deltalake-core/src/operations/transaction/protocol.rs
hntd187: done

crates/deltalake-core/src/operations/transaction/protocol.rs
hntd187: also done

crates/deltalake-core/src/kernel/actions/types.rs
```diff
@@ -274,6 +274,26 @@ pub enum WriterFeatures {
...
+            | WriterFeatures::DeletionVectors
+            | WriterFeatures::RowTracking
+            | WriterFeatures::TimestampWithoutTimezone
+            | WriterFeatures::DomainMetadata
+            | WriterFeatures::V2Checkpoint
+            | WriterFeatures::IcebergCompatV1 => 7,
+        }
+    }
+}
+
```
roeap, line 296: Are we using this somewhere? As per our discussion on Slack, not entirely sure how and where we would use this.

roeap (changes requested):
There may have been a bit of miscommunication :).

The thing we need to decide is, if we do the invariant check or not - i.e. are invariants enabled. This is the case for `2 <= writer_version < 7` and if `writer_version == 7 & 'invariants' in writer_features`. And only if invariants are enabled we should inspect the schema to see if there are any invariants to validate.

One way to do this would be to make sure we only pass invariants to the [`DeltaDataChecker`](https://github.com/delta-io/delta-rs/blob/dd6b45362a14c0f127b32c4b81afc15d17f710d5/crates/deltalake-core/src/delta_datafusion/mod.rs#L1021-L1024) if they are enabled.

It also means we likely need to error if invariants are enabled and we have some defined, but the datafusion features and with that the checker is not available.

crates/deltalake-core/src/kernel/actions/types.rs
hntd187: No, I think this was me speculating about what we might need for determining say minimum versions across the entire hashset of features. I don't have any reason to say it should stay so I'm fine removing it if you would like.

hntd187:
According to the protocol invariants for 2<=>6 are always enabled, so you'd only care to check if you were 7 or higher. Are you able to disable them prior to writer feature checks?

roeap:
> Are you able to disable them prior to writer feature checks?

No, as you said, between 2 and 6 they are always enabled. The main point is that it is not an error if they are configured on the schema, so we should not error, rather if if they are not enabled, we just kind of pretend they don't exist and do nothing w.r.t. to them ...

hntd187:
I added this in check constraints anyways so this PR isn't needed anymore.

PR #1890, opened 2023-11-20
feat(python): expose rust writer v2 as additional engine

ion-elgreco:
# Description
Builds on top of the other PR by adding schema overwrite.


ion-elgreco:
Messed up my branch .. starting over

PR #1891, opened 2023-11-20
feat(python): expose rust writer as additional engine v2

ion-elgreco:
# Description
- Adds rust writer as additional engine in python
- Adds overwrite schema functionality to the rust writer. @roeap feel free to point out improvements 😄 

A couple gaps will exist between current Rust writer and pyarrow writer. We will have to solve this in a later PR:
- Replacewhere (partition filter / predicate) overwrite  
(users however can solve this by doing DeltaTabel.delete and then append)

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1861


rtyler:
@ion-elgreco should #1872 be closed with this being its replacement?

ion-elgreco:
@rtyler  yes will do that!

crates/deltalake-core/src/operations/write.rs
```diff
@@ -415,12 +476,31 @@ impl std::future::IntoFuture for WriteBuilder {
...
+
+                if schema != table_schema {
+                    let mut metadata = this
+                        .snapshot
+                        .current_metadata()
+                        .ok_or(DeltaTableError::NoMetadata)?
+                        .clone();
+                    metadata.schema = schema.clone().try_into().unwrap();
+                    let metadata_action = Metadata::try_from(metadata).unwrap();
+                    actions.push(Action::Metadata(metadata_action));
```
roeap, line 502: let's handle these unwraps.

python/deltalake/writer.py
```diff
@@ -162,15 +228,27 @@ def write_deltalake(
...
-
     table, table_uri = try_get_table_and_table_uri(table_or_uri, storage_options)
+    if table is not None:
+        storage_options = table._storage_options or {}
+        storage_options.update(storage_options or {})
 
-    # We need to write against the latest table version
-    if table:
         table.update_incremental()
+        if table.protocol().min_writer_version > MAX_SUPPORTED_WRITER_VERSION:
```
roeap, line 241: we should use the global `PROTOCOL` instance (specifically `can_write_to`) from the transaction module to check if we can write to a table.

roeap (commented):
looking good, just some minor comments :)

python/deltalake/writer.py
ion-elgreco: @roeap I'll take a look!  I'll probably push this check down to the pyarrow engine only since we still need it but for rust engine we can check with the protocol instance


@wjones127 With respect to all these changes around the protocol support and soon having check constraints, should we support this in pyarrow engine, or we can communicate to users, moving forward rust engine will get all the new features?

ion-elgreco: @roeap ah it seems, it's already there in the `fn check_preconditions`, which is executed when the builder is converted IntoFuture

roeap: true :), then we should just delete that check here.

crates/deltalake-core/src/operations/write.rs
ion-elgreco: @roeap Can you check if the change is ok? I think you want that it returns the error, so I used the question mark operator

python/deltalake/writer.py
ion-elgreco: Yup :) Removed it, check its only done for pyarrow writer engine in python.

crates/deltalake-core/src/operations/write.rs
roeap: yes, exactly right :)

rtyler (approved):

crates/deltalake-core/src/operations/write.rs
```diff
@@ -415,12 +476,31 @@ impl std::future::IntoFuture for WriteBuilder {
...
             if matches!(this.mode, SaveMode::Overwrite) {
+                // Update metadata with new schema
+                let table_schema = this
+                    .snapshot
+                    .physical_arrow_schema(this.log_store.object_store().clone())
+                    .await
+                    .or_else(|_| this.snapshot.arrow_schema())
+                    .unwrap_or(schema.clone());
+
+                if schema != table_schema {
```
scarman-db, line 494: Do the schemas here have Eq or PartialEq on them? This might not be as straight forward as this.

ion-elgreco: It has PartialEq on them. This piece was also reused from src/lib.rs in the python module

PR #1895, opened 2023-11-21
feat: compare timestamp partition values as timestamps instead of strings

sigorbor:
# Description
Delta protocol specifies 2 possible formats for timestamp partitions: {year}-{month}-{day} {hour}:{minute}:{second} or {year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}

However, string comparison of partition filter value and partition values was performed, which rendered timestamps like 2020-12-31 23:59:59.000000 and 2020-12-31 23:59:59 as different.

This change uses timestamp comparison instead of string comparison.

# Documentation
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization


crates/deltalake-core/src/schema/partitions.rs
```diff
@@ -40,6 +42,20 @@ pub struct PartitionFilter {
     pub value: PartitionValue,
 }
 
+fn parse_timestamp(timestamp_str: &str) -> Option<NaiveDateTime> {
+    // Timestamp formats as per https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization
+    let formats = &["%Y-%m-%d %H:%M:%S", "%Y-%m-%d %H:%M:%S%.6f"];
```
natinimni, line 47: I believe you can leave just this format, as it will successfully capture both cases (with or without microseconds).
Moreover, you can omit the "6" constraint (i.e. "%Y-%m-%d %H:%M:%S%.f") because it does not add anything - timestamp strings that include more/less/no microseconds will still be successfully parsed, both with "%.6f" and with "%.f".

sigorbor: Done

sigorbor:
@rtyler, @roeap, @wjones127 - guys, could you please review?

rtyler (approved):

PR #1897, opened 2023-11-21
docs: on append, overwrite, delete and z-ordering

MrPowers:
# Description

Adds docs on how to append, overwrite, delete rows, and Z Order Delta tables.

Will add much more detailed pages in the future.  Just getting the high-level skeleton of the docs developed.


docs/usage/appending-overwriting-delta-lake-table.md
```diff
@@ -0,0 +1,76 @@
...
+|     2 | b        |
+|     3 | c        |
++-------+----------+
+```
+
+Append two additional rows of data to the table:
+
+```python
+df = pd.DataFrame({"num": [8, 9], "letter": ["dd", "ee"]})
+dl.writer.write_deltalake("tmp/some-table", df, mode="append")
```
ion-elgreco, line 23: Should we showcase examples where you import it like as dl. 

I think we generally import directly, right?

docs/usage/create-delta-lake-table.md
```diff
@@ -0,0 +1,26 @@
+# Creating a Delta Lake Table
+
+This section explains how to create a Delta Lake table.
+
+You can easily write a DataFrame to a Delta table.
+
+```python
+import deltalake as dl
```
ion-elgreco, line 8: I think you forgot to remove this import as well

docs/usage/create-delta-lake-table.md
```diff
@@ -0,0 +1,26 @@
...
+from deltalake.writer import write_deltalake
```
ion-elgreco, line 9: We expose write_deltalake also on the deltalake module directly. Also in other docs we import directly from there, maybe do it here as well for consistency

ion-elgreco (changes requested):

docs/usage/create-delta-lake-table.md
MrPowers: I get this error: `cannot import name 'DeltaLake' from 'deltalake'`

You sure `write_deltalake` is exposed on `deltalake`?

docs/usage/optimize/delta-lake-z-order.md
```diff
@@ -0,0 +1,16 @@
+# Delta Lake Z Order
+
+This section explains how to Z Order a Delta table.
+
+Z Ordering colocates similar data in the same files, which allows for more better file skipping and faster queries.
```
thomasfrederikhoeck, line 5: `more` or `better`, right? not both

MrPowers: Fixed, thanks

docs/usage/create-delta-lake-table.md
MrPowers: I had a typo. You're right.  Updated.

docs/usage/create-delta-lake-table.md
MrPowers: Updated.

PR #1905, opened 2023-11-25
docs: update docs home page and add pandas integration

MrPowers:
# Description

Updated the docs home page to include more information about the project.

Also added a pandas integration page.  Lots more integration pages to come soon hopefully!


docs/integrations/delta-lake-pandas.md
```diff
@@ -0,0 +1,279 @@
...
+|     2 | b        |
+|     3 | c        |
+|     8 | dd       |
+|     9 | ee       |
++-------+----------+
+```
+
+Time travel is a powerful feature that pandas users cannot access with CSV or Parquet.
+
+## Schema enforcement
```
ion-elgreco, line 224: Do we perhaps want to have a short section on how you can overwrite the schema and the table at same time.

docs/index.md
```diff
@@ -1,7 +1,35 @@
...
\ No newline at end of file
+## Important terminology
+
+* "Rust deltalake" refers to the Rust API of delta-rs (no Spark dependency)
+* "Python deltalake" refers to the Python API of delta-rs (no Spark dependency)
+* "Delta Spark" refers to the Scala impementation of the Delta Lake transaction log protocol.  This depends on Spark and Java.
+
+## Why implement the Delta Lake transaction log protocol in Rust and Scala?
+
+Delta Spark depends on Java and Spark, which is fine for many use cases, but not all Delta Lake users want to depend on these libraries.
```
roeap, line 15: should we also mention that this allows using delta in Rust or other native projects? in many of these cases using a JVM is not an option.

docs/integrations/delta-lake-pandas.md
MrPowers: @ion-elgreco - I actually just tried to do `write_deltalake("tmp/some-table", df, mode="overwrite")` thinking that would overwrite the scheme and table at the same time and it still gave a `ValueError` surprisingly.  Is there another syntax?  Should I also create an issue to add schema evolution?

docs/index.md
MrPowers: I added a sentence to clarify this.  Good idea!

docs/integrations/delta-lake-pandas.md
ion-elgreco: You need to also pass overwrite_schema = True

PR #1913, opened 2023-11-26
feat: add more info for contributors

### LINKED ISSUE #1893: Add documentation on how to build and run documentation locally
# Description

The documentation can be run locally when the right packages are installed and commands are run.

Here's my understanding of the steps:

```
cd python
make develop
cd ..
mkdocs serve
```

We should document these steps, perhaps in the [CONTRIBUTING.md](https://github.com/delta-io/delta-rs/blob/main/CONTRIBUTING.md) file?

r3stl355:
# Description
Added what I learned when I started contributing here, could save some time to newcomers
Also added a `make` action to build new docs. Old (Sphinx) scripts are still there, we should be good to remove those as we are not longer building those docs, will do that separately

# Related Issue(s)
resolves #1893


ion-elgreco (approved):
Thanks @r3stl355!

Issue #1881, opened 2023-11-18
Implementing CHECK constraints 


hntd187:
# Description

As part of 3,7 support, implement check constraints on tables. #1568 @MrPowers could you please assign me this so others know

**Use Case**

**Related Issue(s)**
#1568

ion-elgreco:
@hntd187 you can write 'take'

hntd187:
take

MrPowers:
@r3stl355 - I love "take".  Thank you!!

PR #1915, opened 2023-11-26
feat: check constraints

hntd187:
# Description
This PR adds CHECK constraints on delta tables. I still have some outstanding work to do on this, but this is my working draft PR for this feature.

```rust
let constraint = table.add_constraint().with_constraint("id", "id < 100");
constraint.await?;
```

# Related Issue(s)
#1881 

# Documentation


ion-elgreco:
@hntd187 does the delta protocol allow to add constraints during create operation?

hntd187:
> @hntd187 does the delta protocol allow to add constraints during create operation?

I'm not sure, the documentation only seems to add them post creation.

hntd187:
Yeah, do an initial create, then just do the alter to add the constraint should be easy since the table is empty.

roeap:
@hntd187 - could we reduce the number of files in the example table to only contain a single add action per commit (unless we need more for the case), to keep entropy down.

crates/deltalake-core/src/operations/constraints.rs
```diff
@@ -0,0 +1,263 @@
...
+                min_reader_version: if old_protocol.min_reader_version > 1 {
+                    old_protocol.min_reader_version
+                } else {
+                    1
+                },
+                min_writer_version: if old_protocol.min_reader_version > 3 {
+                    old_protocol.min_reader_version
+                } else {
+                    3
+                },
```
roeap, line 144: We have some mix-ups here when to use reader and writer version?

Also maybe add a comment that this would be the least required versions for the check feature.

crates/deltalake-core/src/operations/constraints.rs
```diff
@@ -0,0 +1,263 @@
...
+            ]);
+            let commit_info = CommitInfo {
+                timestamp: Some(Utc::now().timestamp_millis()),
+                operation: Some("ADD CONSTRAINT".to_string()),
+                operation_parameters: Some(operational_parameters),
+                read_version: Some(this.snapshot.version()),
+                isolation_level: Some(IsolationLevel::Serializable),
+                is_blind_append: Some(false),
+                ..Default::default()
+            };
```
roeap, line 161: We usually create the `CommitInfo` via the assoaciated `commit_info` method on the `DeltaOperation`.

crates/deltalake-core/src/operations/constraints.rs
```diff
@@ -0,0 +1,263 @@
...
+            let _version = commit(
+                this.log_store.as_ref(),
+                &actions,
+                operations,
+                &this.snapshot,
+                None,
+            )
+            .await?;
+
+            Ok(DeltaTable::new_with_state(this.log_store, this.snapshot))
```
roeap, line 183: I don't think, this snapshot will reflect he latest table state including the updates we just did, as the `commit` function will not update the snapshot.

roeap (commented):
Thanks for looking into this @hntd187 - left a few comments.

Would it make sense to also make sure we check this on the python side?

crates/deltalake-core/src/delta_datafusion/mod.rs
```diff
@@ -1016,15 +1017,43 @@ pub(crate) fn logical_expr_to_physical_expr(
...
+        }
+    }
+
     /// Create a new DeltaDataChecker
-    pub fn new(invariants: Vec<Invariant>) -> Self {
+    pub fn new(snapshot: &DeltaTableState) -> Self {
+        let metadata = snapshot.metadata();
+
+        let invariants = metadata
+            .and_then(|meta| meta.schema.get_invariants().ok())
```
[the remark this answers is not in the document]
scarman-db: This was copied from the place it originally was prior to invariant enforcement in the data checker. Perhaps since this is not related to check constraints and slightly alters what would happen normally can we have a followup PR on it?

crates/deltalake-core/src/delta_datafusion/mod.rs
```diff
@@ -1016,15 +1017,43 @@ pub(crate) fn logical_expr_to_physical_expr(
...
 #[derive(Clone)]
 pub struct DeltaDataChecker {
+    constraints: Vec<Constraint>,
     invariants: Vec<Invariant>,
     ctx: SessionContext,
 }
 
 impl DeltaDataChecker {
+    /// Create a new DeltaDataChecker with a specified set of invariants
+    pub fn with_invariants(invariants: Vec<Invariant>) -> Self {
```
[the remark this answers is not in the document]
scarman-db: Done for both

crates/deltalake-core/src/operations/constraints.rs
scarman-db: Done and done

crates/deltalake-core/src/operations/constraints.rs
scarman-db: So the correct thing here @roeap is to do a merge on the states after the commit right?

roeap: Exactly, we should add the actions we just committed to the table also to the new state.

crates/deltalake-core/src/operations/constraints.rs
scarman-db: This doesn't seem better than what I'm doing here, I'd still have to set most of this myself. I took this info from the spark implementation when I created it. Is that okay?

crates/deltalake-core/src/operations/constraints.rs
```diff
@@ -0,0 +1,263 @@
...
+                .build()
+                .await?;
+
+            let task_ctx = Arc::new(TaskContext::from(&state));
+            let record_stream: SendableRecordBatchStream = scan.execute(0, task_ctx)?;
+            let records = collect_sendable_stream(record_stream).await?;
+
+            for batch in records {
+                checker.check_batch(&batch).await?;
+            }
```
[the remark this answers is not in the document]
roeap: Should look similar to what we do when we write the stream to disk - sans the writing stuff :)

https://github.com/delta-io/delta-rs/blob/e54996ae3bc4a0d1b838e6a584b2a636171f387e/crates/deltalake-core/src/operations/write.rs#L324-L351

To avoid having to iterate through all the partitions in a loop, You could also wrap it in a `CoalesceExec`, where you would always end up with a single partition.

crates/deltalake-core/src/operations/constraints.rs
roeap: Yeah its fine .. the creation of the commit info itself is not so critical.

The only really valuable thing over there is that we have a single place where we can define the string representation for an operation, in this case `ADD CONSTRAINT`. In erarlier version of delta-rs we hade some inconsistencies in that regard.

Come to think of it, most information - like `read_version`, `isolation_level` etc. we should probably set in a central place within the `commit` function anyhow. Especially since there are some cases where we can (and do) relax the isolation level for the actual commit. Should we maybe add an issue for that?

crates/deltalake-core/src/delta_datafusion/mod.rs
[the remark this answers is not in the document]
roeap: sure. but maybe add a TODO?

crates/deltalake-core/src/operations/constraints.rs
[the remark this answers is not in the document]
scarman-db: I used exactly this to reflect the ask. :-)

crates/deltalake-core/src/operations/constraints.rs
```diff
@@ -0,0 +1,318 @@
...
+                ..Default::default()
+            };
+
+            let actions = vec![
+                Action::CommitInfo(commit_info),
+                Action::Metadata(Metadata::try_from(metadata)?),
+                Action::Protocol(protocol),
+            ];
+
+            let _version = commit(
```
roeap, line 189: just a minor nit, but since we use it these is no more need for the underscore.

crates/deltalake-core/src/operations/transaction/protocol.rs
```diff
@@ -170,7 +170,7 @@ pub static INSTANCE: Lazy<ProtocolChecker> = Lazy::new(|| {
     let mut writer_features = HashSet::new();
     writer_features.insert(WriterFeatures::AppendOnly);
     writer_features.insert(WriterFeatures::Invariants);
-    // writer_features.insert(WriterFeatures::CheckConstraints);
+    writer_features.insert(WriterFeatures::CheckConstraints);
```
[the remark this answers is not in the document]
roeap: @hntd187 @scarman-db - one last thing, could we hide `Invariants` and `CheckConstraints` behind the `datafusion` features, as they will not work otherwise ...

roeap (commented):
Just a minor nit and a small ask, then one more sync with main and we are good to go!

roeap (approved):
Thanks for sticking with it, great work @hntd187!

PR #1917, opened 2023-11-28
fix: fail fast for opening non-existent path

### LINKED ISSUE #1916: Load DeltaTable from non-existing folder causing empty folder creation
# Environment

**Delta-rs version**: 0.17.0

**Binding**: rust

**Environment**:
- **Cloud provider**:
- **OS**: macos
- **Other**:

***
# Bug

**What happened**:
New folder created [at the beginning](https://github.com/delta-io/delta-rs/blob/main/crates/deltalake-core/src/table/builder.rs#L250) of `deltalake::open_table("./tests/data/folder_doesnt_exist").await?` which isn't expected and only then it fails with `NotATable` error which is expected. It uses `ensure_table_uri` function in the `build_storage` function during load of Delta Table. And `ensure_table_uri` creates new folder (recursively) if folder doesn't exist. Moreover it doesn't remove that folder at the end so you end up with failed function call and new folder.
I guess it will be nice to fail fast in case user is trying to load some path which isn't even exists not to mention it can't be a table 🙂...
**How to reproduce it**:
`deltalake::open_table("./tests/data/folder_doesnt_exist").await?` and you will see that `folder_not_exists` will be created.
**More details**:
image

`crates/deltalake-core/src/lib.rs`
```
    #[tokio::test()]
    #[should_panic(expected = "NotATable(\"No snapshot or version 0 found")]
    async fn test_load_non_existing_folder() {
        use std::path::Path as FolderPath;
        use std::fs;

        let path_str = "./tests/data/folder_doesnt_exist";

        // Check that there is no such path at the beginning
        let path_doesnt_exist = !FolderPath::new(path_str).exists();
        assert!(path_doesnt_exist);

        let result = match crate::open_table(path_str).await {
            Ok(table) => Ok(table),
            Err(e) => {
                let newly_created_folder_path = FolderPath::new(path_str);
                assert!(newly_created_folder_path.exists());
                // This needs because it actually creates a folder and we need to remove it at the end of the test 😁
                fs::remove_dir(newly_created_folder_path)
                    .expect("can't remove folder for some reason");
                Err(e)
            }
        };
        result.unwrap();
    }
```

dimonchik-suvorov:
# Description
Save user from ending up with failed `open_table` function call and new folder created - failing fast in case user is trying to load some path that doesn't exist

# Related Issue(s)
- closes #1916


crates/deltalake-core/src/table/builder.rs
```diff
@@ -391,6 +409,30 @@ lazy_static::lazy_static! {
         ]);
 }
 
+/// Utility function to figure out whether string representation of the path
+/// is either local path or some kind or URL.
+///
+/// Will return an error if the path is not valid.
+fn resolve_uri_type(table_uri: impl AsRef<str>) -> DeltaResult<UriType> {
```
dimonchik-suvorov, line 416: basically pulled out from `ensure_table_uri` to it's own function

rtyler (approved):

roeap:
@dimonchik-suvorov @rtyler -  I see we have a panic in this PR - generally we want to be panic free and let callers handle errors without crashing everything ... could we fix that?

dimonchik-suvorov:
> @dimonchik-suvorov @rtyler -  I see we have a panic in this PR - generally we want to be panic free and let callers handle errors without crashing everything ... could we fix that?

Sure, will do. It's just my justification was that wrong path it's a pretty wrong to do and I've decided to panic in this case. But sure, I'll rewrite it to return error instead

roeap:
> that wrong path it's a pretty wrong to do and I've decided to panic in this case

Maybe a bit more reasoning behind that :) - assuming someone uses the delta integration in a server and a user provides a erroneous input. In that case a panic would potentially crash the entire server. In that case we would want to allow integrators to forward that error to the user.

Since we are a library, I believe that callers should be able to decide if they can recover from such errors.

dimonchik-suvorov:
> Since we are a library, I believe that callers should be able to decide if they can recover from such errors.

Make sense for the library, agree, will fix 🫡

Issue #1889, opened 2023-11-20
optimize.compact() fails with bad schema after updating to pyarrow 8.0


sebdiem:
# Environment

**Delta-rs version**: 0.12.0

**Binding**: python

**Environment**:
- **Cloud provider**:
- **OS**: Apple M2 Sonoma 14.1.1
- **Other**:

***
# Bug

**What happened**: When I perform an `optimize.compact()` operation following an append to a table, I encounter an exception. Oddly, this optimize.compact() succeeds after the initial append to an empty table but fails after subsequent appends. This is the exception I get:

```_internal.DeltaError: Data does not match the schema or partitions of the table: Unexpected Arrow schema: got: Field { name: "name", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "surname", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, expected: Field { name: "name", data_type: Utf8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "surname", data_type: Utf8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }```

The diff in the exception shows that there are expected non nullable values that are nullable in the Arrow schema.
~The issue does not manifest when using `pyarrow == 7.0.0`.~ Actually it does

**What you expected to happen**: The `optimize.compact()` operation should work.

**How to reproduce it**:  

```python

# The problem exists with pyarrow >= 8 and deltalake = 0.12.0

import pandas as pd
import pyarrow as pa
from deltalake import DeltaTable, write_deltalake
from deltalake.schema import Field, Schema

path = "/tmp/reproduce_delta"

delta_schema = Schema(
    [
        Field("name", "string", nullable=False),
        Field("surname", "string", nullable=False),
    ]
)
pyarrow_schema = pa.schema(field.to_pyarrow() for field in delta_schema.fields)

try:
    table = DeltaTable(path)
except:
    write_deltalake(
        path,
        data=[],
        schema=pyarrow_schema,
        name="test_table",
        partition_by=[],
        mode="error",
        overwrite_schema=False,
        configuration={
            "delta.isolationLevel": "Serializable",
            "delta.enableChangeDataFeed": "false",
            "delta.autoOptimize.autoCompact": "true",
            "delta.autoOptimize.optimizeWrite": "true",
            "delta.targetFileSize": str(128 * 1024 * 1024),
[...]

PR #1926, opened 2023-11-30
fix(rust/python): `optimize.compact` not working with tables with mixed large/normal arrow

ion-elgreco:
# Description
- Fixes optimize.compact not working when a table has parquet files with large and normal arrow types. Basically it cast the recordbatch to normal arrow types

# Issues
- closes https://github.com/delta-io/delta-rs/issues/1889


rtyler (approved):
The specific issue being addressed here I don't have an opinion on, but making `cast_record_batch` a public API is fine to me.

ion-elgreco:
@rtyler yeah the cast_record_batch is needed since we have writers who can write large arrow data into a parquet. According to Will, the arrow writers serialize the arrow schema in the metadata of the parquet, so when we re-read these parquets, there is a chance some recordbatches will have the large dtypes while the others aren't.

PR #1928, opened 2023-11-30
fix: get rid of panic in during table

dimonchik-suvorov:
# Description
This is a continuation of the discussion in the https://github.com/delta-io/delta-rs/pull/1917 Getting rid of panic in the library crate in favor of returning an error so lib users could handle it in a way they see it Test changes accordingly


dimonchik-suvorov:
@roeap here is the fix we ware speaking of in https://github.com/delta-io/delta-rs/pull/1917#issuecomment-1832669833 🙂

roeap (approved):
Thanks so much for the update @dimonchik-suvorov!

PR #1935, opened 2023-12-03
refactor: prefer usage of metadata and protocol fields

roeap:
# Description

Another PR on our way to more kernel compatibility, less API surface, and arrow backed table state. Along also comes a minor update to our Add actions to include the recently added `clusteringProvider` field.

Mainly we try to use `metadata` and `protocol` directly, rather then accessing individual fields via dedicated methods on our top level `DeltaTable` struct.

# Related Issue(s)

related: #1776


crates/deltalake-core/src/table/state.rs
```diff
@@ -41,7 +41,8 @@ pub struct DeltaTableState {
     app_transaction_version: HashMap<String, i64>,
     // table metadata corresponding to current version
     current_metadata: Option<DeltaTableMetaData>,
```
ion-elgreco, line 43: Should this still be there? Should the old fn metadata point to the new metadata field instead of this one? So you don't store it twice

python/src/lib.rs
```diff
@@ -1122,6 +1118,7 @@ impl From<&PyAddAction> for Add {
             deletion_vector: None,
             base_row_id: None,
             default_row_commit_version: None,
+            clustering_provider: None,
```
ion-elgreco, line 1121: What's the clustering provider?

PR #1936, opened 2023-12-03
docs: add logo, dark mode, boost search

ion-elgreco:
- Adds transparent logo (we need a higher quality one), I've pinged @fvaleye on Slack to check whether he has one.
- Add dark mode toggle
- Boost some search results since it was prioritizing the guides over the API reference


roeap:
if the logo is transparent, should we use it in the README as well?

w.r.t. boosting API reference, depending on the audience, the examples might be more helpful? maybe @MrPowers has an opinion?

MrPowers:
LGTM, but not sure about the search boosting...  My guess is that we want the usage/integration pages ranked above the API docs (and the usage pages linking to the API docs).  Open to thoughts.  Normally I don't think API docs give the best SEO experience.

roeap:
@dennyglee made me aware of the [assets](https://github.com/delta-io/delta-docs/tree/main/static/images/logos/_icons/delta-rust) available in the delta-docs repo. Maybe we should use these? There are high quality PNGs as well as SVG versions available.

ion-elgreco:
Awesome @roeap @dennyglee! That was what I was looking for :D

Issue #1466, opened 2023-06-15
DataCatalog does not work for the Python binding


everdark:
# Environment

**Delta-rs version**: 0.10.0

**Binding**: Python

**Environment**:
- **Cloud provider**: AWS
- **OS**: macOS 13.4 (Intel chip)

***
# Bug

**What happened**:

When running 

```python
DeltaTable.from_data_catalog(DataCatalog.AWS, "dbname", "tablename")
```
the following error is raised:

```
--> 264         table_uri = RawDeltaTable.get_table_uri_from_data_catalog(
    265             data_catalog=data_catalog.value,
    266             data_catalog_id=data_catalog_id,

ValueError: Catalog 'glue' not available.
```

AWS credentials have been properly setup as environment variable.
Delta table has been properly registered by crawler.
And within the same session, we have no issue fecthing the table info using AWS SDK:

```python
import boto3
client = boto3.client('glue')
client.get_table(DatabaseName="dbname", Name="tablename")
```

Also a direct instantiation from s3 path works as well:

```python
DeltaTable("s3://path/to/table")
```

**What you expected to happen**:

`DeltaTable.from_data_catalog` should return a `DeltaTable` instance based on catalog supported. In this case, it is AWS Glue.

As a side note, I also hear from other folks facing the same error when using UNITY catalog.

arabe91:
i can confirm that I have exactly the same issue while using the UNITY catalog

wjones127:
Thanks for reporting this. I think we'll need some help from folks who know more about catalogs; we don't have integration tests with them yet since we aren't sure how to set them up.

rtyler:
I'm going to close this, since we removed that functionality.

PR #1941, opened 2023-12-05
fix: remove the get_data_catalog() function

### LINKED ISSUE #1860: Add DataCatalog support
# Environment

**Delta-rs version**: 0.13.0

**Binding**: Python (`deltalake-0.13.0-cp37-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`)

**Environment**:
- **OS**: Linux/amd64
- **Other**:

***
# Bug

When trying to use either flavor of `DataCatalog` a `ValueError` is thrown.

**What happened**:

```
❯ python3
Python 3.11.4 (main, Jun 28 2023, 19:51:46) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from deltalake import DeltaTable, DataCatalog
>>> dt = DeltaTable.from_data_catalog(DataCatalog.AWS, 'db', 'table')
  File "/home/tyler/source/github/noviconnect/venv/lib64/python3.11/site-packages/deltalake/table.py", line 287, in from_data_catalog
    table_uri = RawDeltaTable.get_table_uri_from_data_catalog(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Catalog 'glue' not available.
>>> dt = DeltaTable.from_data_catalog(DataCatalog.UNITY, 'db', 'table')
  File "/home/tyler/source/github/noviconnect/venv/lib64/python3.11/site-packages/deltalake/table.py", line 287, in from_data_catalog
    table_uri = RawDeltaTable.get_table_uri_from_data_catalog(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Catalog 'unity' not available.
>>>
```

C'mon son.

**How to reproduce it**:

**More details**:

rtyler:
This is related to #1713 and #1860 insofar that it removes the `get_data_catalog()` function. This is being done separately from the sub-crates work since it's easy enough to remove and frankly doesn't work properly anyways for Python users, which is the only caller we currently have in our codebase

Fixes #1860


rtyler:
@ion-elgreco I dunno, I strongly doubt people are using that functionality. The Unity catalog functionality is practically worthless IMHO since you have to have credentials for Unity catalog to lookup a table's location, and then a _second_ set of credentials to actually access that storage.

IMHO it is not worth slowing changes for a couple reasons:

* I think Unity is _highly_ unlikely to be used by Python users in this way
* We're not a 1.0 crate or wheel yet

I however defer to @wjones127 here since he's much more familiar with the Python ecossytem than I am.

I do need this to merge before #1825 because there's this presumption in this API that all available catalogs will be compiled into deltalake-core which is I need to remove for sub-crates.

roeap (approved):
LGTM.

as you mentioned the catalogs in python are somewhat non functional right now anyhow ...

wjones127:
Given they are not functional and don't have a maintainer right now, I'm fine removing them for now.

https://github.com/delta-io/delta-rs/issues/1466

ion-elgreco:
We need to remove the python code paths as well.

dhirschfeld:
I'm trying/struggling to understand the current state of Unity Catalog support.

* Is it true that UC support has been *removed* from 0.15?
* If so, is it planned to bring it back in future so I could connect to a table via UC without having direct access to the S3 bucket?
* If it is planned to bring it back, are there any estimates of when that will be - e.g. 2024? Or are there no firm plans to bring back UC support at this stage?

ion-elgreco:
@dhirschfeld I would love to get UC support but this is all in the hands of databricks. They simply made it impossible to read data that is behind a catalog outside of databricks...

dhirschfeld:
> @dhirschfeld I would love to get UC support but this is all in the hands of databricks. They simply made it impossible to read data that is behind a catalog outside of databricks...

Well, that's disappointing 😞  
I thought it might be possible as they have a lot of docs around managing storage credentials - e.g.
* https://docs.databricks.com/en/connect/unity-catalog/index.html
* https://docs.databricks.com/api/account/accountstoragecredentials/get

...but maybe that just gives access from withing the DataBricks UI and doesn't actually expose the underlying storage token 😞

ion-elgreco:
@dhirschfeld One of the requirements of grabbing the storage credential is: ` have some permission on the storage credential`.

 I don't think that would be the same as read access to a certain table, so you still need some form of high level access, and I assume the people who have access to the managed credential they likely already have access directly to the object store.

If you find another way, we can definetely add it! :) I'm personally also hampered by the fact that I can't read delta tables in UC with delta-rs

dhirschfeld:
> *If you find another way, we can definetely add it! :)*

I don't know much about Databricks, and less about Unity Catalog, but I'll have a poke about to see what I can find. If you have been unable to find a way to do it, I'm not that hopeful, but we'll see...

I opened https://github.com/delta-io/delta-rs/discussions/2066 to track my progress! 😄

Issue #1361, opened 2023-05-12
refactor: load table state lazily


wjones127:
Right now, when we instantiate a Delta table, we load the entire table state into memory. For many workloads, we often don't need to have all of it, especially if we are only querying certain partitions at a time. Instead, we should ignore the add and remove actions when instantiating, and only load them as needed during scans.

When we read the log files, we should cache them on disk so we can quickly scan them again later for add and remove actions.

## Instantiate table

1. Verify `_delta_log` exists
2. Either identify most recent version, or verify version requested exists
3. Download relevant log files
4. Scan for the table-level actions, such as metadata and protocol.

This gives you an instance of `DeltaTable` you can inspect and get table-level metadata. When asked for files, will run scan process below.

## Scan table

1. Using the partition filter, scan the log files to get the add and remove actions that are relevant.
2. Use the actions to generate the file set
3. Prune the files based on stats
4. Scan all the files.

This kind of operation will need to run any time asked for a files.

## In-memory caching

Below a certain threshold (which we could make configurable), it's probably fine to keep all the table state in-memory. So we might find some data structure that we can use as a memory-limited in-memory cache for table state. 🤔

PR #1943, opened 2023-12-05
feat: add lazy loading of tables

PengLiVectra:
# Description
Add lazy loading of tables. When preforming streaming operations we don't need any version of the table loaded. Large tables are slow to load, so we see a huge performance boost by avoiding CPU time spent loading the table metadata.

# Related Issue(s)
Partly with https://github.com/delta-io/delta-rs/issues/1361

# Testing
Adding Unit tests

# Breaking Change
Not a breaking change.

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -85,6 +85,28 @@ struct RawDeltaTableMetaData {
     configuration: HashMap<String, Option<String>>,
 }
 
+#[inline]
+fn build_table(
+    table_uri: &str,
+    storage_options: Option<HashMap<String, String>>,
+    without_files: bool,
+    log_buffer_size: Option<usize>,
+) -> Result<DeltaTableBuilder, PythonError> {
+    let mut builder = deltalake::DeltaTableBuilder::from_uri(table_uri);
+    if let Some(storage_options) = storage_options {
+        builder = builder.with_storage_options(storage_options)
+    }
+    if without_files {
+        builder = builder.without_files()
+    }
+    if let Some(buf_size) = log_buffer_size {
+        builder = builder
+            .with_log_buffer_size(buf_size)
+            .map_err(PythonError::from)?;
+    }
+    Ok(builder)
+}
+
 #[pymethods]
 impl RawDeltaTable {
     #[new]
@@ -96,24 +118,34 @@ impl RawDeltaTable {
         without_files: bool,
         log_buffer_size: Option<usize>,
     ) -> PyResult<Self> {
-        let mut builder = deltalake::DeltaTableBuilder::from_uri(table_uri);
         let options = storage_options.clone().unwrap_or_default();
-        if let Some(storage_options) = storage_options {
-            builder = builder.with_storage_options(storage_options)
-        }
+        let mut builder = build_table(table_uri, storage_options, without_files, log_buffer_size);
         if let Some(version) = version {
-            builder = builder.with_version(version)
-        }
-        if without_files {
-            builder = builder.without_files()
-        }
-        if let Some(buf_size) = log_buffer_size {
-            builder = builder
-                .with_log_buffer_size(buf_size)
-                .map_err(PythonError::from)?;
+            builder = Ok(builder?.with_version(version))
         }
 
-        let table = rt()?.block_on(builder.load()).map_err(PythonError::from)?;
+        let table = rt()?.block_on(builder?.load()).map_err(PythonError::from)?;
+        Ok(RawDeltaTable {
+            _table: table,
+            _config: FsConfig {
+                root_url: table_uri.into(),
+                options,
+            },
+        })
+    }
+
+    #[classmethod]
+    #[pyo3(signature = (table_uri, storage_options = None, without_files = false, log_buffer_size = None))]
+    fn load_lazy(
+        _cls: &PyType,
+        table_uri: &str,
+        storage_options: Option<HashMap<String, String>>,
+        without_files: bool,
+        log_buffer_size: Option<usize>,
+    ) -> PyResult<Self> {
+        let options = storage_options.clone().unwrap_or_default();
+        let builder = build_table(table_uri, storage_options, without_files, log_buffer_size);
+        let table = builder?.build().map_err(PythonError::from)?;
         Ok(RawDeltaTable {
             _table: table,
             _config: FsConfig {
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -85,67 +85,39 @@ struct RawDeltaTableMetaData {
     configuration: HashMap<String, Option<String>>,
 }
 
-#[inline]
-fn build_table(
-    table_uri: &str,
-    storage_options: Option<HashMap<String, String>>,
-    without_files: bool,
-    log_buffer_size: Option<usize>,
-) -> Result<DeltaTableBuilder, PythonError> {
-    let mut builder = deltalake::DeltaTableBuilder::from_uri(table_uri);
-    if let Some(storage_options) = storage_options {
-        builder = builder.with_storage_options(storage_options)
-    }
-    if without_files {
-        builder = builder.without_files()
-    }
-    if let Some(buf_size) = log_buffer_size {
-        builder = builder
-            .with_log_buffer_size(buf_size)
-            .map_err(PythonError::from)?;
-    }
-    Ok(builder)
-}
-
 #[pymethods]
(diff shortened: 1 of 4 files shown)


python/src/lib.rs
```diff
@@ -85,6 +85,28 @@ struct RawDeltaTableMetaData {
...
+    if let Some(buf_size) = log_buffer_size {
+        builder = builder
+            .with_log_buffer_size(buf_size)
+            .map_err(PythonError::from)?;
+    }
+    Ok(builder)
+}
+
 #[pymethods]
 impl RawDeltaTable {
```
ion-elgreco, line 111: I think this makes sense to add, just two comments, I think we can simplify it a lot. We can just add the load_lazy argument as input for the `new` method, something like this: 

```rust
   fn new(
        table_uri: &str,
        version: Option<i64>,
        storage_options: Option<HashMap<String, String>>,
        without_files: bool,
        log_buffer_size: Option<usize>,
        load_lazy: bool,
    ) -> PyResult<Self> {
        let options = storage_options.clone().unwrap_or_default();
        let mut builder = deltalake::DeltaTableBuilder::from_uri(table_uri);
        if let Some(storage_options) = storage_options {
            builder = builder.with_storage_options(storage_options)
        }
        if without_files {
            builder = builder.without_files()
        }
        if let Some(buf_size) = log_buffer_size {
            builder = builder
                .with_log_buffer_size(buf_size)
                .map_err(PythonError::from)?;
        }
        if let Some(version) = version {
            builder = builder.with_version(version)
        }
        let table = if !load_lazy {
            rt()?.block_on(builder.load()).map_err(PythonError::from)?
        } else {
            builder.build().map_err(PythonError::from)?
        };
        Ok(RawDeltaTable {
            _table: table,
            _config: FsConfig {
                root_url: table_uri.into(),
                options,
            },
        })
```
and in python we just do an if else to set the metadata:
```python
        if load_lazy:
            self._metadata = None
        else:
            self._metadata = Metadata(self._table)
```

roeap:
@PengLiVectra - could you explain the use case a bit? 

For typical scenarios, without protocol and metadata information there is not much we can (legally) do with the table.

To avoid loading the state into memory we can already pass options to omit files and tombstones. Currently we are also working on making that more lazy - i.e. just getting protocol and metadata asap, but avoid loading anything else.

PengLiVectra:
> @PengLiVectra - could you explain the use case a bit?
> 
> For typical scenarios, without protocol and metadata information there is not much we can (legally) do with the table.
> 
> To avoid loading the state into memory we can already pass options to omit files and tombstones. Currently we are also working on making that more lazy - i.e. just getting protocol and metadata asap, but avoid loading anything else.

Yes, when using in streaming operations, sometimes the task has no data, it takes too much time to load large tables. We did a simple test with `load_lazy`:
# cat test.py
`import deltalake
x = deltalake.DeltaTable(s3path, lazy_load=True)
print(x.peek_next_commit(4690))`

If `load_lazy = True`, result is `python test.py  0.89s user 0.39s system 84% cpu 1.526 total`
If `load_lazy = False`, result is `python test.py  97.53s user 3.74s system 87% cpu 1:56.23 total`

ion-elgreco:
@PengLiVectra if the task has no data, and you can check this why not simply load the table state if there is data, if not you skip the step.

roeap:
I still don't quite understand the use case :) 

If your task receives no data, then there should be no need to lad the table at all, right? But when does hte need to peek the next table version arise?

delta-rs was initially build to support large scale log ingestion into delta via kafka streams (this is also why the kafka-delta-ingest project exists)

To handle the sometimes very large state that gets generated there are two options that can be passed to the table `require_files` and `require_tombstones` which will cause the internals to not load the file actions (we don't want to do a scan) and the tombstones (we also don't want to vacuum).

for any other meaningful interaction with the table, we always need the protocol and metadata actions. for the above scenario it would also need to be an `appendOnly` table.

So from the litle I know about your use case, I would suggest: 
1. configure table as append only.
2. check you data if you need to write to the table at all
3. if you need to write  data, load the table with `require_files` and `require_tombstones` set to false?

Would that meet your needs?

PengLiVectra:
> 3\. require_files

Sorry for the misleading @ion-elgreco and @roeap , we didn't know if there is no data prior, we need to load the table and check if there is a new commit. So basically we need to save the time in the step of loading table.

ion-elgreco:
@PengLiVectra also in that case, wouldn't initializing the table without files give the same result combined with get_latest_version or peek next commit

PengLiVectra:
> @PengLiVectra if the task has no data, and you can check this why not simply load the table state if there is data, if not you skip the step.



> @PengLiVectra also in that case, wouldn't initializing the table without files give the same result combined with get_latest_version or peek next commit



> @PengLiVectra also in that case, wouldn't initializing the table without files give the same result combined with get_latest_version or peek next commit

Thanks, I'm still not very clear. Could you give an example about it in Python?

PengLiVectra:
> I still don't quite understand the use case :)
> 
> If your task receives no data, then there should be no need to lad the table at all, right? But when does hte need to peek the next table version arise?
> 
> delta-rs was initially build to support large scale log ingestion into delta via kafka streams (this is also why the kafka-delta-ingest project exists)
> 
> To handle the sometimes very large state that gets generated there are two options that can be passed to the table `require_files` and `require_tombstones` which will cause the internals to not load the file actions (we don't want to do a scan) and the tombstones (we also don't want to vacuum).
> 
> for any other meaningful interaction with the table, we always need the protocol and metadata actions. for the above scenario it would also need to be an `appendOnly` table.
> 
> So from the litle I know about your use case, I would suggest:
> 
> 1. configure table as append only.
> 2. check you data if you need to write to the table at all
> 3. if you need to write  data, load the table with `require_files` and `require_tombstones` set to false?
> 
> Would that meet your needs?

`require_files` and `require_tombstones` are in `build` function, not in `load` function
https://github.com/delta-io/delta-rs/blob/main/crates/deltalake-core/src/table/builder.rs#L280-L300.

It seems only `load` functions is exposed in python (https://github.com/delta-io/delta-rs/blob/main/python/src/lib.rs#L116).

Is there a way to use this in Python?

ion-elgreco:
@PengLiVectra that's likely because it doesn't load any state and also doesn't validates the file path unlike the normal load would do

@roeap please correct me if I'm wrong

roeap:
@PengLiVectra - sorry for the large delay.

We are currently working towards release 0.17 which will include many breaking changes. One large piece of work I am currently addressing is to optimize the whole log handling and be as minimal as possible in getting to a useful state of the delta table. We are also leaking a lot of internals that we cannot commit to long term.

Why am I telling you this? :) 

> TL;DR; would it be OK to wait with a way forward a bit until some internal changes have landed? 

Your use case makes a lot of sense and we certainly have the opportunity to provide an implementation that streaming use cases can benefit from - much like you proposed.

The lazy-load in this case is more on the object store level, since if we do not have any metadata and without it we cannot do anything meaningful specific to delta. In fact the fact that the metadata and protool are currently optional is a major nuicance, that permeates the entire codebase ... IIRC, you are most interested if a new version of the table exists, right?

In that case we will also have facilities to inspect the state of the log from a storage perspective. and we may expose some of this for use cases like yours.

PengLiVectra:
> @PengLiVectra - sorry for the large delay.
> 
> We are currently working towards release 0.17 which will include many breaking changes. One large piece of work I am currently addressing is to optimize the whole log handling and be as minimal as possible in getting to a useful state of the delta table. We are also leaking a lot of internals that we cannot commit to long term.
> 
> Why am I telling you this? :)
> 
> > TL;DR; would it be OK to wait with a way forward a bit until some internal changes have landed?
> 
> Your use case makes a lot of sense and we certainly have the opportunity to provide an implementation that streaming use cases can benefit from - much like you proposed.
> 
> The lazy-load in this case is more on the object store level, since if we do not have any metadata and without it we cannot do anything meaningful specific to delta. In fact the fact that the metadata and protool are currently optional is a major nuicance, that permeates the entire codebase ... IIRC, you are most interested if a new version of the table exists, right?
> 
> In that case we will also have facilities to inspect the state of the log from a storage perspective. and we may expose some of this for use cases like yours.

Thanks, @roeap. Do you have a timeline for release 0.17?

ion-elgreco:
@PengLiVectra now that we already using the new log replay in python as well, I believe this can be closed. Does it meet your needs?

PR #1950, opened 2023-12-06
feat: use pyo3-asyncio to get a fresh tokio runtime

PengLiVectra:
# Description
This PR greatly reduces network connections and dns request volume by the delta-rs library when using Python bindings. The approach here is to utilize pyo3-asyncio's tokio-runtime feature as the source of the Runtime. This yields the same runtime across function calls which preserves connections in the connection pool. The previous code created a new runtime per python function call, which established all new socket connections and issued new DNS requests.

# Related Issue(s)
Partly https://github.com/delta-io/delta-rs/pull/1315

# Testing:
Ran a script that called hundreds of delta operations, and watched tcpdump. Only saw one dns request.

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -4,6 +4,7 @@ mod error;
 mod filesystem;
 mod schema;
 mod utils;
+extern crate pyo3_asyncio;
 
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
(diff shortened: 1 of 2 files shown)


python/src/lib.rs
```diff
@@ -52,10 +53,15 @@ use crate::filesystem::FsConfig;
...
 #[inline]
-fn rt() -> PyResult<tokio::runtime::Runtime> {
+fn rt_pyo3() -> PyResult<tokio::runtime::Runtime> {
     tokio::runtime::Runtime::new().map_err(|err| PyRuntimeError::new_err(err.to_string()))
 }
 
+#[inline]
+fn rt() -> &'static tokio::runtime::Runtime {
+    pyo3_asyncio::tokio::get_runtime()
+}
```
wjones127, line 63: Why don't we just create one global one in a lazy static instead? I don't think we need the `pyo3-asyncio` dependency.

ion-elgreco:
@PengLiVectra can you add the requested suggestion from Will?

PR #1953, opened 2023-12-08
fix: Fixing example in README

blais:
fix: Fixing the Python example in the README, which fails.

df == df2 will trigger element-size comparison and fail to evaluated in Python 's assert.


blais:
I don't understand the format. Other things to do. Retracting.

PR #1959, opened 2023-12-12
feat: cast list items to default before write with different item names

JonasDev1:
# Description
Delta-rs always uses `item` as the list item name when writing lists. If you read data which is for example written by Spark, the item name is `element`, in the current implemantation it's not possible to write RecordBatches with a different item name. This leads for example to the problem that you cann't optimize tables which are written by Spark and contain a List column. 
In this MR I add condition which will intiate a cast if the list item name of the record batch is different to the target schema one. 
I have also tried to explain this behaviour in the tests, but unfortunately creating the test data has become complicated (Happy to get feedback)

This is my first MR in this project 

# Related Issue(s)
https://github.com/delta-io/delta-rs/blob/main/crates/deltalake-core/src/kernel/arrow/mod.rs#L58
https://github.com/delta-io/delta-rs/pull/684/files#r940790524
https://delta-users.slack.com/archives/C013LCAEB98/p1701885637615699


crates/deltalake-core/src/operations/cast.rs
```diff
@@ -37,6 +38,16 @@ fn cast_record_batch_columns(
         .collect::<Result<Vec<_>, _>>()
 }
 
+fn is_cast_required(a: &DataType, b: &DataType) -> bool {
+    match (a, b) {
+        (DataType::List(a_item), DataType::List(b_item)) => {
+            // If list item name is not the default('item') the list must be casted
+            !a.equals_datatype(b) || a_item.name() != b_item.name()
```
ion-elgreco, line 45: I think just checking by name can give false positives, especially since we don't check the nested datatypes

ion-elgreco: I also mean the other way around. This will return True if you only check by name:

        let field1 = DataType::List(FieldRef::from(Field::new("item", DataType::Int32, false)));
        let field2 = DataType::List(FieldRef::from(Field::new("element",  DataType::Float64,   false, )));

ion-elgreco: One thing I dont follow, how come  !a.equals_datatype(b) not already return here True without a_item.name, Does it only check the top-level datatype with the name of the inner field?

JonasDev1: They only check the datatypes, but not the name.

`Compares the datatype with another, ignoring nested field names`
[It would be best to adjust this function, but it is native arrow](https://github.com/apache/arrow-rs/blob/master/arrow-schema/src/datatype.rs#L415)

ion-elgreco: I see, then this fix should be fine since it does check nested types.

ion-elgreco (approved):

PR #1962, opened 2023-12-12
docs: add auto-release when docs are merged to main

r3stl355:
# Description
This will trigger a release action for docs when a pull request to docs (`mkdocs.yml` or anything in `docs/` is changed) is merged to main. This would help with more frequent releases to docs without waiting for a full release or bothering admins to run the docs release manually, especially handy at this time when lots of changes are being made to the docs.

One caveat is that this will replace the current public docs (which is what we intend to achieve with this) including the Python API docs which may not be a desirable outcome for API docstrings which are on the `main` branch but not in the latest Python release.

### DIFF
--- a/.github/workflows/docs_release.yml
+++ b/.github/workflows/docs_release.yml
@@ -1,0 +1,24 @@
+name: docs_release
+
+on:
+  pull_request:
+    types:
+      - closed
+    branches: [main]
+    paths:
+      - docs/**
+      - mkdocs.yml
+
+jobs:
+  release-docs:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger the docs release event
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          event-type: release-docs
+          client-payload: >
+            {
+              "tag": "${{ github.ref_name }}"
+            }


mightyshazam (commented):
Looks good.

mightyshazam (approved):

ion-elgreco:
@r3stl355 can we perhaps combine this with a versioning system. Otherwise indeed what you mention it can become confusing since definitions will be changed and don't reflect to the actual release.

rtyler (approved):

r3stl355:
I couldn't come up with a good suggestion for this versioning @ion-elgreco. I consider this `docs_release` a temporary solution for simplifying the docs release cycle and suggest disabling it once the new docs catch up so that docs release will be part of python release. 
I guess we could pull the next release version so it is clearly reflected in the docs header (docs release are configurable so I can simply add it as a config value) but that in turn will also be misleading because the next version is not released yet.
A 3rd option would be making a release to a sub-folder (as was suggested by @mightyshazam in the yesterday's call) with a link from the main docs pointing to it. Also easy to achieve and would not pollute the current released logs

ion-elgreco:
@r3stl355 I got a PR open, where I got the versioning working. The only caveat is that it will mean the old urls won't work since everything is under a version name.

This also theoretically should work with the change you made, so we should have a main which is the latest, and when we do a release that will be pushed as new version and then will be marked as latest, until it's superseded by a PR in main

PR #1964, opened 2023-12-13
fix: properly deserialize percent-encoded file paths of Remove actions, to make sure tombstone and file paths match

sigorbor:
# Description
Percent-encoded file paths of Remove actions were not properly deserialized, and when compared to active file paths, the paths didn't match, which caused tombstones to be recognized as active files (be kept in the state)


sigorbor:
@roeap @rtyler @wjones127 -  a test is failing likely because I'm not supposed to manually "fix" test data delta logs. Can simply push the fix without the change in the test data? The scenario is simple - Remove transaction doesn't match Add transaction if there are percent encoded paths.

natinimni (approved):

Issue #1967, opened 2023-12-13
.load_with_datetime() is incorrectly rounding to nearest second


echai58:
# Environment

**Delta-rs version**: 0.14.0

**Binding**: python

***
# Bug

**What happened**:
When you call `load_with_datetime`, it calls `datetime.timestamp()`, https://github.com/delta-io/delta-rs/blob/fee96a7b31e89d994593f303a3dc8d3cca49b061/crates/deltalake-core/src/table/mod.rs#L835 - which returns the number of **seconds**. This is incorrect, because the version history has millisecond granularity https://github.com/delta-io/delta-rs/blob/fee96a7b31e89d994593f303a3dc8d3cca49b061/crates/deltalake-core/src/table/mod.rs#L180 , so it will return the wrong version if you are trying to read/write at sub-second granularity. 

**What you expected to happen**:
It should call `datetime.timestamp_millis()`.

**How to reproduce it**:
Write a table, check the transaction time with `.history()`, and try to read the version as of something in in the same second, but with a greater milliseconds. It will return the incorrect version.

Issue #1910, opened 2023-11-25
Rethink the load_version and load_with_datetime interfaces


MrPowers:
# Description

There are currently a few different interfaces for time traveling:

Option 1:

```python
DeltaTable("some_path", version=1)
```

Option 2:

```python
dt = DeltaTable("some_path")
dt.load_version(1)
```

Option 3:

```python
dt = DeltaTable("some_path")
dt.load_with_datetime("2018-01-26T18:30:09Z")
```

We should consider providing a single interface:

```python
dt = DeltaTable("some_path").time_travel(version=1)
dt = DeltaTable("some_path").time_travel(version_as_of="2018-01-26T18:30:09Z")
```

I don't think methods like these should be limited to only taking string arguments: `load_with_datetime(datetime_string: str)`

Users should be able to pass in Python date and datetime objects as well.

ion-elgreco:
Good one, we can perhaps consolidate the load_version and load_with_datetime into a single function:

`DeltaTable.load_to(version: int | datetime)`

PR #1968, opened 2023-12-14
feat(python): combine load_version/load_with_datetime into `load_as_version`

ion-elgreco:
# Description
Combines the two functions into one.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1910
- closes https://github.com/delta-io/delta-rs/issues/1967


python/deltalake/table.py
```diff
@@ -453,20 +453,68 @@ def file_uris(
...
+            ```
+            datetime(2023,1,1)
+            ```
+
+            **Use a datetime in string format**
+            ```
+            "2018-01-26T18:30:09Z"
+            "2018-12-19T16:39:57-08:00"
+            "2018-01-26T18:30:09.453+00:00"
+            ```
```
roeap, line 480: should these examples show the actual function invocaton?

roeap (commented):
thanks for this refactor.

somehow I am not a fan of the name `load_to`, but also don't have a better one ready.

roeap:
maybe an opportunity to fix #1967 as well?

python/deltalake/table.py
ion-elgreco: Yeah, that would be better

ion-elgreco:
> thanks for this refactor.
> 
> somehow I am not a fan of the name `load_to`, but also don't have a better one ready.

@roeap maybe we can just name it `load` instead?

roeap:
I try to always see if it works as a short sentence ... "load [table] to version" just somehow hit my ear wrong 😆. 

Would "load [table] as version" work? `load` alone may be confusion as well, as this may be misconsrued as the table still needs to be loaded?

ion-elgreco:
@roeap Yeah when you read it like that it doesn't sound right haha. Hmm, that would translate into load_as_version or load_asof?

roeap:
Not sure myself, but what would you think about `at` ... 

```
table.at(version)
```

ion-elgreco:
@roeap I think `at` is to unclear. Would the previously mentioned `load_state` not be more clear? Since it's doing that?

ion-elgreco:
> maybe an opportunity to fix #1967 as well?

@roeap fixing this, causes now one of the restore test in Rusts to fail, not sure what to do here because all python tests pass fine.

roeap (approved):
👍

PR #1970, opened 2023-12-14
fix: properly decode percent-encoded file paths coming from parquet checkpoints

sigorbor:
# Description

When read from parquet checkpoints, the Add and Remove file paths are not percent-decoded, in contrary to when read from JSON transaction logs. That causes file paths mismatch (e.g. percent-encoded Add file path is read from checkpoint while the same tombstone path is read from JSON), and can result in tombstone files being counted as active.


crates/deltalake-core/src/protocol/serde_path.rs
```diff
@@ -1,89 +0,0 @@
...
-            ),
-            (
-                "string=$%25&%2F()%3D%5E%22<>~%5B%5D%7B}`%23|%2A%3F%2F%5Cr%5Cn.%3A/part-00023-e0a68495-8098-40a6-be5f-b502b111b789.c000.snappy.parquet",
-                "string=$%2525&%252F()%253D%255E%2522%3C%3E~%255B%255D%257B%7D%60%2523%7C%252A%253F%252F%255Cr%255Cn.%253A/part-00023-e0a68495-8098-40a6-be5f-b502b111b789.c000.snappy.parquet"
-            ),
-            (
-                "string=$%25&%2F()%3D%5E%22<>~%5B%5D%7B}`%23|%2A%3F%2F%5Cr%5Cn.%3A_-/part-00023-346b6795-dafa-4948-bda5-ecdf4baa4445.c000.snappy.parquet",
-                "string=$%2525&%252F()%253D%255E%2522%3C%3E~%255B%255D%257B%7D%60%2523%7C%252A%253F%252F%255Cr%255Cn.%253A_-/part-00023-346b6795-dafa-4948-bda5-ecdf4baa4445.c000.snappy.parquet"
-            )
-        ];
```
wjones127, line 80: It seems we are losing test coverage here. Why are we deleting this file?

sigorbor: This file is duplicated in src\kernel\actions\serde_path.rs, including the test. Likely a leftover.

natinimni (approved):

rtyler (approved):

PR #1973, opened 2023-12-14
feat(python): expose `add constraint` operation

ion-elgreco:
# Description
Currently limited to a single constraint at once. I didn't want to chance the function defintion for this so I just check if more than 1 constraint is passed it will raise an error. Once we have the multiple constraints possibility in Rust, I can simply remove this size check.


ion-elgreco:
Errors will be fixed if we merge this refactor first: https://github.com/delta-io/delta-rs/pull/1979

roeap (approved):
Nice! Should we also add an issue to track documenting this feature?

ion-elgreco:
@roeap regarding the single constraint addition per time?

roeap:
@ion-elgreco - yes, not sure that there is a question o.a. here ... 😆

ion-elgreco:
@roeap :) I'll create an issue for it.

@hntd187  mentioned he would add that possibility

PR #1974, opened 2023-12-14
docs: add versioning

ion-elgreco:
- Adds versioning to the docs


ion-elgreco:
I think this will break the url though.

MrPowers:
@ion-elgreco - what kind of versioning?  Like different docs pages for v0.14 and v0.15?

ion-elgreco:
@MrPowers yes, different docs pages for each version, see example below:

image

MrPowers:
@ion-elgreco - let's chat about this one on a call.  Lots of factors to consider for this one in particular.  This is an issue I have thought about extensively.  Lots of pros/cons to the different approaches.

ion-elgreco:
@MrPowers sure, maybe next week?

.github/workflows/docs.yml
```diff
@@ -77,7 +77,9 @@ jobs:
       - name: Build
         run: |
           source python/venv/bin/activate
-          mkdocs build
+          var=$(echo `${{ github.event.client_payload.tag || github.ref_name }` | sed 's/^python-v//')
```
NikoUlmas-db, line 80: I believe there is an error here with a missing second `}`

.github/workflows/docs.yml
```diff
@@ -11,11 +11,11 @@ on:
     types:
       - release-docs
   # Allow manual trigger for now
-  workflow_dispatch:
+  # workflow_dispatch:
```
r3stl355, line 14: Any reason for removing this? It's handy for running ad-hoc manual releases

.github/workflows/docs.yml
r3stl355: Ooops, too many github accounts :)

r3stl355:
@ion-elgreco  I'm not sure if this versioning is going to work as expected because the versioning is used to build the docs but it is not aware of the deployment location so effectively will always show the same content, that dropdown will be used only as a label

.github/workflows/docs.yml
ion-elgreco: It's not needed anymore I believe with the change you made. It now triggers after every PR with main and it will trigger with every release.

.github/workflows/docs.yml
ion-elgreco: Typo :) Fixed it now

ion-elgreco:
@r3stl355 I checked, the tabbed code snippets still works. If you change the code snippet it doesn't affect the previous doc versions.

r3stl355:
Yes @ion-elgreco the snippets will work but the links about those code snippets will be pointing to the main docs only, this is driven by a single config value in the `mkdocs.yml` file (e.g. `python_api_url: https://delta-io.github.io/delta-rs/api/`

ion-elgreco:
@r3stl355 Ah I see what you mean now, yeah those should be dynamic..

PR #1975, opened 2023-12-15
feat: retry with exponential backoff for DynamoDb interaction

dispanser:
# Description

We use an external crate, [backoff](https://crates.io/crates/backoff), to retry DynamoDb read and write operations when the error in the response is `ProvisionedThroughPutExceeded`, indicating an overload of DynamoDb wrt to the configured read and write capacity.

### DIFF
--- a/crates/deltalake-aws/src/lib.rs
+++ b/crates/deltalake-aws/src/lib.rs
@@ -21,8 +21,8 @@ use rusoto_core::{HttpClient, Region, RusotoError};
 use rusoto_credential::AutoRefreshingProvider;
 use rusoto_dynamodb::{
     AttributeDefinition, AttributeValue, CreateTableError, CreateTableInput, DynamoDb,
-    DynamoDbClient, GetItemInput, KeySchemaElement, PutItemError, PutItemInput, QueryInput,
-    UpdateItemError, UpdateItemInput,
+    DynamoDbClient, GetItemError, GetItemInput, KeySchemaElement, PutItemError, PutItemInput,
+    QueryError, QueryInput, UpdateItemError, UpdateItemInput,
 };
 use rusoto_sts::WebIdentityProvider;
 use url::Url;
@@ -156,7 +156,7 @@ impl DynamoDbLockClient {
 
     /// Create the lock table where DynamoDb stores the commit information for all delta tables.
     ///
-    /// Transparently handles the case where that table already exists, so it's to call.
+    /// Transparently handles the case where that table already exists, so it's safe to call.
     /// After `create_table` operation is executed, the table state in DynamoDb is `creating`, and
     /// it's not immediately useable. This method does not wait for the table state to become
     /// `active`, so transient failures might occurr when immediately using the lock client.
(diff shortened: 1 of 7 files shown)


dispanser:
I decided against using the existing retry logic from `crates/deltalake-core/src/data_catalog/client/retry.rs` that was suggested by @roeap:
- with the current crate structure, it's not possible to access from `deltalake-aws` crate
- there's not much that can be directly reused, most of the logic resides in `impl RetryExt for reqwest::RequestBuilder`
- the model (extension trait `RetryExt`) isn't really viable because we're retrying four different operations, so we'd be duplicating the actual retry handling (`loop`, considering `max_retries`, actual sleeping, etc) multiple times
- I think the approach taking by backoff, marking errors as either transient or permanent, is self-documenting and easy to understand

@rtyler: We're currently only retrying `ProvisionedThroughputExceeded`, but for most of these error enums there's also the `RequestLimitExceeded` variant, documented via:
> Throughput exceeds the current throughput quota for your account. Please contact AWS Support at <a href="https://aws.amazon.com/support">AWS Support</a> to request a quota increase

In theory, this indicates that the error is also transient, but I'm not an AWS expert enough to understand if it's sensible to handle this as well.

crates/deltalake-aws/src/lib.rs
```diff
@@ -266,16 +288,37 @@ impl DynamoDbLockClient {
...
 
+async fn retry<I, E, Fn, Fut>(operation: Fn) -> Result<I, E>
+where
+    Fn: FnMut() -> Fut,
+    Fut: std::future::Future<Output = Result<I, backoff::Error<E>>>,
+{
+    let backoff = backoff::ExponentialBackoffBuilder::new()
+        .with_multiplier(2.)
+        .with_max_interval(Duration::from_secs(15))
+        .with_max_elapsed_time(Some(Duration::from_secs(60)))
```
rtyler, line 317: The max elapsed interval of 60 I am assuming is just a number that you've picked? 

I think this is something that will likely need to be configurable since different systems would have a different tolerance here. Based on how `retry` is being invoked I don't see a clear path to pulling configuration down into this call :thinking:

rtyler (commented):
I think pulling the backoff crate into deltalake-aws is reasonable. The nice thing about the subcrates is that this new dependency only affects AWS users :smile:

crates/deltalake-aws/src/lib.rs
dispanser: Yes, these numbers would ideally be configurable. It seems other aspects of `delta-rs`, including the locking provider choice itself, are being configured via environment variables, so I guess that's the way to go.

My question would mostly be if we want all five of possible parameters to be configurable:
- max elapsed time
- max interval
- multiplier
- initial interval (using defaults in this PR)
- randomization factor (using defaults in this PR)

The most obvious one is probably the one you singled out, max elapsed time, but there's probably a user out there for any of these :).

rtyler: @dispanser I agree there are folks likely that would want to tune every single one of these...maybe.

I think making an environment variable for max elapsed time is the only one really important to implement to merge this. The others people may want, or not, but they should make themselves known in the future :smile:

dispanser: I added an environment variable for the max elapsed time setting.

I'd address the diverging docs in a separate PR, they are already way off anyways :)

dispanser:
I think the failing test (macos, `restore_by_datetime`) is unrelated.

rtyler:
@dispanser please rebase this on the latest main with the mega refactor :smile:

dispanser:
@rtyler rebase is done, all tests pass locally.

rtyler (approved):

PR #1979, opened 2023-12-17
refactor: trigger metadata retrieval only during `DeltaTable.metadata`

ion-elgreco:
# Description
Triggers metadata retrieval only on metadata call, this is a better approach otherwise we need to add it after each method that does an alteration to the table config. Also now it will only be triggered if the user actually wants to retrieve the metadata.


r3stl355:
Looks good to me

roeap (approved):
👍

PR #1985, opened 2023-12-19
fix: implement consistent formatting for constraint expressions

### LINKED ISSUE #1971: Constraint expr not formatted during commit action
Currently when you write a constraint expr, the expr is not formatted on write, so you could end up with these expressions in the log: 

`'delta.constraints.value_gt_5': 'foo         >0',`

Maybe it's good if we format them before writing.

Blajda:
# Description
Implements consistent formatting for constraint expressions so something like `value      <             1000` is normalized to `value < 1000`

Also includes drive by improvements.
 1. Test & Fix that Datafusion expressions can actually be used when adding a constraint
 2. Test & Fix that constraints can be added to column with capitalization
 
# Related Issue(s)
- closes #1971

### DIFF
--- a/crates/deltalake-core/src/operations/constraints.rs
+++ b/crates/deltalake-core/src/operations/constraints.rs
@@ -8,11 +8,15 @@ use datafusion::execution::context::SessionState;
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::prelude::SessionContext;
+use datafusion_common::ToDFSchema;
 use futures::future::BoxFuture;
 use futures::StreamExt;
 use serde_json::json;
 
-use crate::delta_datafusion::{register_store, DeltaDataChecker, DeltaScanBuilder};
+use crate::delta_datafusion::expr::fmt_expr_to_sql;
+use crate::delta_datafusion::{
+    register_store, DeltaDataChecker, DeltaScanBuilder, DeltaSessionContext,
+};
 use crate::kernel::{Action, CommitInfo, IsolationLevel, Metadata, Protocol};
 use crate::logstore::LogStoreRef;
 use crate::operations::datafusion_utils::Expression;
@@ -22,6 +26,8 @@ use crate::table::state::DeltaTableState;
 use crate::table::Constraint;
 use crate::DeltaTable;
 use crate::{DeltaResult, DeltaTableError};
+
+use super::datafusion_utils::into_expr;
 
 /// Build a constraint to add to a table
 pub struct ConstraintBuilder {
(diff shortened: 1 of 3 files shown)


crates/deltalake-core/src/operations/constraints.rs
```diff
@@ -94,23 +95,29 @@ impl std::future::IntoFuture for ConstraintBuilder {
...
 
-            // Checker built here with the one time constraint to check.
-            let checker = DeltaDataChecker::new_with_constraints(vec![Constraint::new("*", &expr)]);
             let scan = DeltaScanBuilder::new(&this.snapshot, this.log_store.clone(), &state)
                 .build()
                 .await?;
 
+            let schema = scan.schema().to_dfschema()?;
+            let expr = into_expr(expr, &schema, &state)?;
+            let expr_str = fmt_expr_to_sql(&expr)?;
```
ion-elgreco, line 115: Is this going from SQL string expr to Datafusion Expr and then back to sql string expression?

ion-elgreco (approved):
Great addition!

Left only one comment just for my own understanding

crates/deltalake-core/src/operations/constraints.rs
Blajda: Yes exactly. 
If provided expression is already a Datafusion expression then `into_expr` will simply return it otherwise it will parse the string expression.
Either way if a String is provided or a proper DF expression it needs to be normalized somewhere.

ion-elgreco: Which SQL dialect does Datafusion actually use?

roeap (commented):
❤️

crates/deltalake-core/src/operations/constraints.rs
roeap: postgres ...  but the actual dialect expressions in the delta log should comply with is as of yet unspecified :).

ion-elgreco: Should we put SQLglot in between on the python side then, so we can allow more flexibility there?

Blajda: We use the [generic dialect](https://github.com/delta-io/delta-rs/blob/4ece26d0da666b59df206515d24ab8797329b46f/crates/deltalake-core/src/delta_datafusion/expr.rs#L87) which in my mind is similar to postgres. Given we are only parsing expressions and not DML or DDL statements we should be fine.
https://docs.rs/sqlparser/latest/sqlparser/dialect/struct.GenericDialect.html

Blajda: > Should we put SQLglot in between on the python side then, so we can allow more flexibility there?

I'm indifferent to that feature. My only concern is if we start exposing operations to other languages (i.e delta-core) then we should have fairly consistent interfaces. 
A decision / documentation needs to be made on which sql dialect must be used when exposing expressions to the log. If I had to guess they would pick hive sql.

PR #1987, opened 2023-12-20
fix: flakey gcs test

roeap:
# Description

The GCS tests seem to be failing sometimes, looked around and hoping this change fixes that.


ion-elgreco (approved):
Ofcourse:)

PR #1988, opened 2023-12-21
feat(test): integration test for S3 log store with pyspark

dispanser:
# Description

This adds an integration test to make sure that delta-rs and pyspark can write to a shared delta table concurrently, using the same lock table in a compatible way.

Due to the way pyspark instantiates the `SparkSession` there's only one session for the entire test run, so we must set up the S3-relevant configs globally, for all tests. This doesn't seem to interfere with the other tests as they write locally and ignore any S3-related configuration.


dispanser:
@rtyler : the new test doesn't work out of the box, as it requires S3 credentials in the form of environment variables:
- `AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY`
- `AWS_REGION`

`AWS_PROFILE` might work as well instead, but I haven't tested that.

dispanser:
it may be necessary to mark this test with a separate tag so it can be excluded from a normal CI run, unless we have some credentials we want to stick into GitHub

rtyler:
I think I can add some credentials to an AWS account I can control and structure specifically for this purpose. That said, it looks like I'm going to have to create an IAM user with a brought swath of permissions, have you been testing basically with an admin-level IAM user, or do you have a specific set of IAM permissions in mind?

dispanser:
I completely forgot about permissions and their implications while writing this. :tired_face: 

I've been using a very unrestricted user, indeed. The way it currently works is by creating a bucket and a table for each run, but that's purely for the sake of setup-free reproducability. To reduce the required permissions, we could modify the tests to accept an existing bucket + table via environment , which would allows to get away with very limited permissions - basically reading from + writing to a specific bucket and table. That table could also be pre-configured with a proper TTL and very small provisioned throughput for neatly fitting into any free AWS tier.

rtyler:
@dispanser The big refactor just landed, can you rebase and rework this pull request accordingly? :smile:

dispanser:
@rtyler : rebase done. What do you think about my proposal in the previous comment?

rtyler:
I'm going to close this. It's too old, and unfortunately I was never able to get the changes into Delta/Spark to make this easier to do integration testing with :unamused:

Issue #1838, opened 2023-11-11
add statistics to arrow datasets


djouallah:
I am using delta_rs python with DuckDB, it works rather and thanks for that, but I did notice that it is substantially slower than reading directly from Parquet, I asked the DuckDB devs and their answer is that arrow dataset does not contains any statistics, so a lot of queries that involved join orders for example become problematics 

is there a way to fix that ?

wjones127:
@djouallah Did the DuckDB devs say how external libraries can provide statistics to DuckDB? We have various kinds of statistics; I just don't know how to provide them to DuckDB and which it needs for join orders.

r3stl355:
As far as I know the flow is Deltalake->Arrow->DuckDB, if they need stats in DuckDB does it mean they are not pushing down anything to Deltalake?

roeap:
pyarrow `Dataset`s are made up of `Fragments` (I think thats what they call them), which for us corresponds to files. These can also contain statistics, but this should already be set. IF we pass a reader to DuckDB, I think the only way is to actually iterate through all the batches. IN case of table, all is materialized.

So as far as I know, the `Dataset` generated by delta-rs should contain statistics, if DuckDB ran leverage them in this form is a different question :).

@djouallah, can you confirm you used `to_pyarrow_dataset` for this?

ion-elgreco:
Polars is able to do pushdowns through pyarrow dataset, so theoretically duckdb should be able to do that as well

wjones127:
My interpretation of "their answer is that arrow dataset does not contains any statistics" is that it's not about pushdown but about surfacing information to their query planner. For example, they would consider the estimated number of rows in each source table to determine order. So I think this is separate from statistics-based page / file pruning, which we do indeed perform.

djouallah:
Yes exactly

wjones127:
Understood. In DataFusion we have this information here: https://docs.rs/datafusion/latest/datafusion/physical_plan/struct.Statistics.html

In DuckDB the closest thing I can find is: https://github.com/duckdb/duckdb/blob/a00b28f5d453ff7ec3b3837385f083d0887124ad/src/include/duckdb/storage/statistics/node_statistics.hpp#L16

I don't think Polars has any notion of this. 

I'll think about ways this could be put in some abstraction that can be read by DuckDB.

djouallah:
link to duckdb devs comments
https://github.com/duckdb/duckdb/discussions/4636#discussioncomment-7497063

PR #1993, opened 2023-12-26
docs: datafusion integration

MrPowers:
Purpose: document the DataFusion integration page.

Need to figure out why Delta Lake depends on DataFusion and put a little info in this guide before merging.

### DIFF
--- a/docs/integrations/delta-lake-arrow.md
+++ b/docs/integrations/delta-lake-arrow.md
@@ -2,7 +2,7 @@
 
 Delta Lake tables can be exposed as Arrow tables and Arrow datasets, which allows for interoperability with a variety of query engines.
 
-This page shows you how to convert Delta tables to Arrow data structures and teaches you the difference between Arrow tables and Arrow datasets.
+This page shows you how to convert Delta tables to Arrow data structures and teaches you the difference between Arrow tables and Arrow datasets.  Tables are "eager" and datasets are "lazy", which has important performance implications, keep reading to learn more!
 
 ## Delta Lake to Arrow Dataset
 
--- a/docs/integrations/delta-lake-datafusion.md
+++ b/docs/integrations/delta-lake-datafusion.md
@@ -1,0 +1,80 @@
+# Using Delta Lake with DataFusion
+
+This page explains how to use Delta Lake with DataFusion.
+
+Delta Lake offers DataFusion users better performance and more features compared to other formats like CSV or Parquet.
+
+Delta Lake works well with the DataFusion Rust API and the DataFusion Python API.  It's a great option for all DataFusion users.
+
+Delta Lake also depends on DataFusion to implement some functionality under the hood.  We will also discuss this dependency at the end of this guide in case you're interested in learning more about the symbiotic relationship between the two libraries.
+
+## Delta Lake performance benefits for DataFusion users
+
+Let's run some DataFusion queries on a Parquet file and a Delta table with the same data to learn more about the performance benefits of Delta Lake.
+
+Suppose you have the following dataset with 1 billion rows and 9 columns.  Here are the first three rows of data:
+
+```
++-------+-------+--------------+-------+-------+--------+------+------+---------+
+| id1   | id2   | id3          |   id4 |   id5 |    id6 |   v1 |   v2 |      v3 |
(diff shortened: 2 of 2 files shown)


djouallah:
I think it may be useful to clarify that using arrow dataset to read delta table is just a workaround with some serious limitations,  currently as far as i can tell stats are not passed, using a simple benchmarks, reading Parquet directly is substantially Faster
image

https://colab.research.google.com/drive/1sJD7w6l7RUjRHoPKoM4EQGcfwRqfKCqk#scrollTo=KMX-DymJKIh4

ion-elgreco:
@djouallah which pyarrow and deltalake version did you use?

djouallah:
> @djouallah which pyarrow and deltalake version did you use?

14.0.2, 0.14

djouallah:
just as a reference
https://github.com/delta-io/delta-rs/issues/1838

ion-elgreco:
@djouallah could you add polars and duckdb in the mix so we can compare across engines?

djouallah:
> @djouallah could you add polars and duckdb in the mix so we can compare across engines?

Polars does not support arbitrary SQL, so I can't use it, I will add Duckdb parquet vs delta, but same problem,  total rows are not passed, so duckdb end up with weird Query plans

ion-elgreco:
> > @djouallah could you add polars and duckdb in the mix so we can compare across engines?
> 
> Polars does not support arbitrary SQL, so I can't use it, I will add Duckdb parquet vs delta, but same problem, total rows are not passed, so duckdb end up with weird Query plans

Polars has a sql context: https://pola-rs.github.io/polars/user-guide/sql/intro/#execute-queries-from-multiple-sources

djouallah:
> > > @djouallah could you add polars and duckdb in the mix so we can compare across engines?
> > 
> > 
> > Polars does not support arbitrary SQL, so I can't use it, I will add Duckdb parquet vs delta, but same problem, total rows are not passed, so duckdb end up with weird Query plans
> 
> Polars has a sql context: https://pola-rs.github.io/polars/user-guide/sql/intro/#execute-queries-from-multiple-sources

I know but last time, i tried, it did not support the whole 22 Queries

MrPowers:
@djouallah - stats are passed.  The runtime on Delta Lake vs Parquet for a small dataset is quite volatile and really depends how the data is distributed.  For example, suppose you need to query 1% of the data and the entire dataset is 50GB.

The Delta Table could have 99% data skipping or 0% data skipping.

The Parquet table could also have 99% data skipping (only the relevant data is in one of the row groups) or 0% data skipping.

In order to make an apples:apples Parquet:Delta Lake comparison, the file distribution should be similar to the row group distribution.  

I am running these queries on a 50 GB dataset locally (I have a Macbook M1 with 64 GB of RAM).  The query runs in ~5 seconds - pretty fast!

djouallah:
> > > @djouallah could you add polars and duckdb in the mix so we can compare across engines?
> > 
> > 
> > Polars does not support arbitrary SQL, so I can't use it, I will add Duckdb parquet vs delta, but same problem, total rows are not passed, so duckdb end up with weird Query plans
> 
> Polars has a sql context: https://pola-rs.github.io/polars/user-guide/sql/intro/#execute-queries-from-multiple-sources
https://colab.research.google.com/drive/1cfWgQW4LoP9RN9rUkfVclku3qfSP3P2w
image

djouallah:
> @djouallah - stats are passed. The runtime on Delta Lake vs Parquet for a small dataset is quite volatile and really depends how the data is distributed. For example, suppose you need to query 1% of the data and the entire dataset is 50GB.
> 
> The Delta Table could have 99% data skipping or 0% data skipping.
> 
> The Parquet table could also have 99% data skipping (only the relevant data is in one of the row groups) or 0% data skipping.
> 
> In order to make an apples:apples Parquet:Delta Lake comparison, the file distribution should be similar to the row group distribution.
> 
> I am running these queries on a 50 GB dataset locally (I have a Macbook M1 with 64 GB of RAM). The query runs in ~5 seconds - pretty fast!

@MrPowers , feel free to check the notebook, it is totally reproducible, if you increase sf to 5, the performance difference will be substantial , running a Query on a single table may not show the issue, it is more about joins reordering I think.

ion-elgreco (approved):

Issue #1990, opened 2023-12-22
store userMetadata on write operations


resulyrt93:
# Description
With spark deltalake we can store userMetadata with write operations, thus we can access them in delta history. It can be an arbitrary json data. Is there a way to store it with delta-rs python binding? Doc [link](https://books.japila.pl/delta-lake-internals/delta/options/#usermetadata)

**Use Case**
In our case, we want to store user data with write operation to match transaction with user.

**Related Issue(s)**
https://github.com/delta-io/delta-rs/issues/133

ion-elgreco:
Should be trivial to expose, I can take a look at it after new years eve

r3stl355:
Can `userMetadata` be added to `app_metadata` before being passed to `commit` (unless it is already the same thing and just needs to be exposed to Python)? Adding a new parameter to `commit` would touch many operations. 
I am working on https://github.com/delta-io/delta-rs/issues/1957 so can also look into this after I'm done there

cekicbaris:
+ 
this is exactly what we need.

ion-elgreco:
@r3stl355 it's the same thing, see my PR: https://github.com/delta-io/delta-rs/pull/1994

![image](https://github.com/delta-io/delta-rs/assets/15728914/a3701544-0663-4c51-899a-736945ed97a0)

r3stl355:
cool, so this doesn't touch Rust at all, nice

PR #1994, opened 2023-12-27
feat(python): expose custom metadata to writers

ion-elgreco:
# Description
- exposes the custom_metadata to pyarrow and rust writer
- addresses a bug in the create operation, we were not passing the app_metadata to the actual commit

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1990

### DIFF
--- a/crates/deltalake-core/src/operations/create.rs
+++ b/crates/deltalake-core/src/operations/create.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use futures::future::BoxFuture;
-use serde_json::{Map, Value};
+use serde_json::Value;
 
 use super::transaction::{commit, PROTOCOL};
 use crate::errors::{DeltaResult, DeltaTableError};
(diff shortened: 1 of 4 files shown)


crates/deltalake-core/src/operations/create.rs
```diff
@@ -181,8 +181,11 @@ impl CreateBuilder {
     ///
     /// This might include provenance information such as an id of the
     /// user that made the commit or the program that created it.
-    pub fn with_metadata(mut self, metadata: Map<String, Value>) -> Self {
-        self.metadata = Some(metadata);
+    pub fn with_metadata(
+        mut self,
+        metadata: impl IntoIterator<Item = (String, serde_json::Value)>,
```
r3stl355, line 186: This looks interesting, would love to understand the reason for changing. Is this to allow more than just Map?

r3stl355:
LGTM

crates/deltalake-core/src/operations/create.rs
ion-elgreco: Mostly to streamline between the other implementations, we were using HashMap there, also the commit takes a hashmap as input and not a map

rtyler (approved):

PR #1995, opened 2023-12-27
chore: refactoring AWS code out of the core crate

rtyler:
This change was quite time-consuming and as such changes a _lot_ of files to ensure that unit and integration tests are cleaned up properly.

What should be done in a follow up pull request is the creation of the `deltalake-gcs` and `deltalake-azure` crates to contain their objectstore registration code and integration tests. I consider that out of the scope of this change.

This is a prerequisite for my work on #1601


crates/deltalake-core/src/logstore/mod.rs
```diff
@@ -16,19 +18,71 @@ use crate::{
...
+        },
+    ))
+}
+
+#[derive(Clone, Debug, Default)]
+struct DefaultLogStoreFactory {}
+impl LogStoreFactory for DefaultLogStoreFactory {}
+
+/// Registry of [LogStoreFactory] instances
+pub type FactoryRegistry = Arc<DashMap<Url, Arc<dyn LogStoreFactory>>>;
```
ion-elgreco, line 66: Just for my own understanding, what does DashMap solve here instead of a normal hashmap? I read the docs but I don't grasp what it's purpose is here : P

crates/deltalake-core/src/table/builder.rs
```diff
@@ -150,31 +146,48 @@ pub struct DeltaTableBuilder {
...
+                    DeltaTableError::InvalidTableLocation(table_uri.as_ref().to_string())
+                })?;
+                ensure_file_location_exists(path)?;
             }
+        } else {
+            ensure_file_location_exists(PathBuf::from(table_uri.as_ref()))?;
         }
 
-        Ok(DeltaTableBuilder::from_uri(table_uri))
+        let url = ensure_table_uri(&table_uri).expect("The specified table_uri is not valid");
```
ion-elgreco, line 181: This seems like unnecessarily check if you do `from_uri` already

crates/deltalake-core/src/table/builder.rs
```diff
@@ -483,6 +411,8 @@ mod tests {
         assert!(uri.is_ok());
         let _uri = ensure_table_uri("./nonexistent");
         assert!(uri.is_ok());
+        let _uri = ensure_table_uri("file:///tmp/nonexistent/some/path");
+        assert!(uri.is_ok());
```
ion-elgreco, line 415: It's not checking _uri here

crates/deltalake-test/src/utils.rs
```diff
@@ -183,88 +144,6 @@ impl IntegrationContext {
...
-    Amazon,
-    Microsoft,
-    Onelake,
-    Google,
-    Local,
-    Hdfs,
-    OnelakeAbfs,
-}
-
-impl StorageIntegration {
```
ion-elgreco, line 221: Are these things not used anymore for the other cloud stores?

crates/deltalake-core/Cargo.toml
```diff
@@ -156,29 +143,13 @@ datafusion = [
     "sqlparser",
     "arrow",
     "parquet",
-    "dashmap",
 ]
 datafusion-ext = ["datafusion"]
 gcs = ["object_store/gcp"]
-hdfs = ["datafusion-objectstore-hdfs"]
```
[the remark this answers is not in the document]
rtyler: Lots of the functionality tied to the `hdfs` feature is totally untested, it's ignored code, so I removed the feature for now. The `hdfs` code needs to be moved into its own crate and probably its own repo tbh

crates/deltalake-core/src/logstore/mod.rs
rtyler: Dashmap supports lock-free concurrency, so it can be used across threads in the process. HashMap would need to be put into a `Mutex<T>` first in order to do the same.

crates/deltalake-test/src/utils.rs
rtyler: For the other cloud stores they should be implementing the trait. I don't actually know why our integration tests were implemented the way they that they were, basically with everything in kind of hard-coded lists with feature flags.

The model used in `deltalake-aws/tests/common.rs` is what I would expect to see more of for the other cloud provider traits

crates/deltalake-core/src/table/builder.rs
rtyler: The module level `ensure_table_uri` does some behaviors that needed to be preserved for file URLs in particular. `DeltaTableBuilder::from_uri` and `from_valid_uri` both have slightly different functionality which I do think we might want to get rid of before 1.0

crates/deltalake-aws/src/storage.rs
```diff
@@ -0,0 +1,597 @@
...
+use tokio::io::AsyncWrite;
+use url::Url;
+
+const STORE_NAME: &str = "DeltaS3ObjectStore";
+
+#[derive(Clone, Default, Debug)]
+pub struct S3ObjectStoreFactory {}
+
+impl S3ObjectStoreFactory {
+    fn with_env_s3(&self, options: &StorageOptions) -> StorageOptions {
```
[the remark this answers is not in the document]
rtyler: The specifics of why this code does things these way are a little outside of my realm of understanding :laughing: 

I was _mostly_ moving code around, with some refactoring here and there. This code is largely doing the same thing it did when it lived in deltalake-core.  That said, this takes `StorageOptions` rather than `S3StorageOptions` because the result of this is intended to be passed into the `ObjectStoreFactory.parse_url_opts` which is a more generic trait and just operates on the `StorageOptions` struct

roeap:
@rtyler - took the liberty to disable auto-merge since I think we should squash that one. The indivisula commits don't seem to be valuable on their own. 

More importantly though, I started to do a proper review. Off the bat I was wondering if we need a dedicated test crate to place all the data, or if we can just keep that in the core crate, and then test integrations.

In any case i the way things are going, it makes sense to just move the example table to the root of the repository, rather then in a crate. Eventually we would likely add or even replace with the DAT tables.

crates/deltalake-core/src/protocol/time_utils.rs
```diff
@@ -96,25 +24,12 @@ pub fn timestamp_to_delta_stats_string(n: i64, time_unit: &TimeUnit) -> Option<S
...
-        TimeUnit::Milliseconds => temporal_conversions::timestamp_ms_to_datetime(n),
-        TimeUnit::Microseconds => temporal_conversions::timestamp_us_to_datetime(n),
-        TimeUnit::Nanoseconds => temporal_conversions::timestamp_ns_to_datetime(n),
-    }?;
-
-    Some(format!("{}", dt.format("%Y-%m-%dT%H:%M:%S%.3fZ")))
-}
-
+#[cfg(feature = "parquet")]
 #[cfg(test)]
```
roeap, line 28:
```suggestion
#[cfg(all(test, feature = "parquet")]
```

crates/deltalake-core/src/table/builder.rs
```diff
@@ -266,11 +279,22 @@ impl DeltaTableBuilder {
...
-            &self.options.table_uri,
-            self.storage_options(),
-            self.options.storage_backend,
-        )
+        debug!("build_storage() with {}", &self.options.table_uri);
+        let location = Url::parse(&self.options.table_uri).map_err(|_| {
+            DeltaTableError::NotATable(format!(
+                "Could not turn {} into a URL",
+                self.options.table_uri
+            ))
```
roeap, line 287: seems like in practice we should never hit this, if the builder was crated via a rout that ensure URIs are fine ... since we do have to dela with relative paths and the likes, how about we make the field on `DeltaLoadOptios` a URL, and avoid throwing here?

crates/deltalake-core/src/logstore/mod.rs
```diff
@@ -37,6 +91,58 @@ lazy_static! {
...
+
+    if let Some(entry) = crate::storage::factories().get(&scheme) {
+        debug!("Found a storage provider for {scheme} ({location})");
+        let (store, _prefix) = entry
+            .value()
+            .parse_url_opts(&location, &options.clone().into())?;
+        return logstore_with(store, location, options);
+    }
+    Err(DeltaTableError::InvalidTableLocation(
+        location.clone().into(),
```
roeap, line 121: do we need this clone, as this is the last usage?

crates/deltalake-core/src/logstore/mod.rs
```diff
@@ -37,6 +91,58 @@ lazy_static! {
...
+/// Return the [LogStoreRef] using the given [ObjectStoreRef]
+pub fn logstore_with(
+    store: ObjectStoreRef,
+    location: Url,
+    options: impl Into<StorageOptions> + Clone,
+) -> DeltaResult<LogStoreRef> {
+    let scheme = Url::parse(&format!("{}://", location.scheme()))
+        .map_err(|_| DeltaTableError::InvalidTableLocation(location.clone().into()))?;
+
+    if let Some(factory) = logstores().get(&scheme) {
```
roeap, line 134: i think we need to consider both scheme and host in the registries. Otherwise we cannot register stores for differnt buckets with differnet credentials.

The `file:///` stores will always have an empty host, except in some arcane windows cases, that I have not yet encountered in the wild...

crates/deltalake-core/src/storage/mod.rs
```diff
@@ -22,11 +28,147 @@ lazy_static! {
...
+            _ => Err(DeltaTableError::InvalidTableLocation(url.clone().into())),
+        }
+    }
+}
+
+/// TODO
+pub type FactoryRegistry = Arc<DashMap<Url, Arc<dyn ObjectStoreFactory>>>;
+
+/// TODO
+pub fn factories() -> FactoryRegistry {
```
roeap, line 74: do we need to create global state for this? Ie. could we just place this on `DeltaTable`? 

My main concer is the integration with datafusion and other query engines, where such state is usually handled on a session. I also plan to to leverage the datafusion interns etc for our SCAN et. al. operations. In think in these cases having a global registry might limit hopw we can integrate.

roeap (changes requested):
Exited to see we are working to untagle all this!!

Left some initial comments mainly orounf how we set up the registries.

Haven't full worked throught the factiories / registires yet, but at first glace it seems like we may be doing a bit too much work inside there.. My thinking would be users need to register a fuly functioning store for a given `ObjectStoreUrl` (just scheme and host) - where we may or may not wnat to define our own `ObjectStoreUrl`.

Maybe it looks already like that, just dumping some thought :).

roeap (changes requested):
Exited to see we are working to untagle all this!!

Left some initial comments mainly orounf how we set up the registries.

Haven't full worked throught the factiories / registires yet, but at first glace it seems like we may be doing a bit too much work inside there.. My thinking would be users need to register a fuly functioning store for a given `ObjectStoreUrl` (just scheme and host) - where we may or may not wnat to define our own `ObjectStoreUrl`.

Maybe it looks already like that, just dumping some thought :).

roeap (approved):
👍

Issue #1957, opened 2023-12-11
Add replaceWhere functionality


MrPowers:
# Description

PySpark has a cool `replaceWhere` function that lets you override existing data in a Delta table that matches a predicate with new data.  Here's an example of the `replaceWhere` functionality:

```python
df2 = spark.createDataFrame(
    [
        ("x", 7),
        ("y", 8),
        ("z", 9),
    ]
).toDF("letter", "number")

(
    df2.write.format("delta")
    .option("replaceWhere", "number >= 2")
    .mode("overwrite")
    .save("tmp/my_data")
)
```

What do folks think about adding `replaceWhere` functionality to Python deltalake.

It's possible that the Rust `predicate` argument in `write_deltalake` already exposes this functionality.

ion-elgreco:
I exposed the predicate parameter for the rust engine writer but it's currently not doing anything because the functionality in Rust is not built yet

r3stl355:
take

r3stl355:
I'll give this a try

r3stl355:
WriteBuilder uses `predicate: Option<String>` but has no implementation for it yet whereas DeleteBuilder uses `predicate: Option<Expression>`. I suggest harmonising by changing WriteBuilder to use `predicate: Option<Expression>`. Though this is a breaking change, predicate handling is not implemented in WriteBuilder so changing the type should not cause issues

roeap:
It would be great to do this usig logical expressions rather then the physical ones - much like @Blajda recently updated for merge. The good thing there is we get some type coercion for free, which has been a hassle with expressions.

In python we will likely have to accept strings and do the parsing..

ion-elgreco:
@roeap I think we can start allowing arrow expressions as input, which we can serialize as substrait and then deserialize with Datafusion-substrait

roeap:
This would be a great goal, but I would say lets be consistent in that and make a deliberate API choice.

I.e not have substrait supported in one method but not the other...

Good news is substrait plans are of course logical plans :)

r3stl355:
I'll try that @roeap. As for 

> It would be great to do this usig logical expressions rather then the physical ones - much like @Blajda recently updated for merge.

is this the David's PR you are referring to? https://github.com/delta-io/delta-rs/pull/1969

ion-elgreco:
@roeap we should be able to add this to merge, update, delete and write and then just add the conversion inside the pyo3 binding, so it's a Python only feature.

[...]

PR #1996, opened 2023-12-28
feat: implementation for replaceWhere

r3stl355:
# Description
First/naive implementation of `replaceWhere` for `write`. Code compiles and there is a test to verify the outcome. I would appreciate any feedback on improving the structure/implementation. For example, I copied the part of code from `delete` operation because there is no way to call that code in `delete` directly from `write` - should I look into extracting that code from `delete` to somewhere central?

Seems to also works with partitions columns.

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/1957

# Documentation
Added a section in docs

### DIFF
--- a/python/tests/test_writer.py
+++ b/python/tests/test_writer.py
@@ -869,6 +869,8 @@ def test_replace_where_overwrite(
     value_type: pa.DataType,
     filter_string: str,
 ):
+    table_path = tmp_path
+
     sample_data = pa.table(
         {
             "p1": pa.array(["1", "1", "2", "2"], pa.string()),
@@ -876,9 +878,9 @@ def test_replace_where_overwrite(
             "val": pa.array([1, 1, 1, 1], pa.int64()),
         }
     )
-    write_deltalake(tmp_path, sample_data, mode="overwrite", partition_by=["p1", "p2"])
+    write_deltalake(table_path, sample_data, mode="overwrite")
 
-    delta_table = DeltaTable(tmp_path)
+    delta_table = DeltaTable(table_path)
     assert (
         delta_table.to_pyarrow_table().sort_by(
             [("p1", "ascending"), ("p2", "ascending")]
@@ -890,36 +892,101 @@ def test_replace_where_overwrite(
         {
             "p1": pa.array(["1", "1"], pa.string()),
             "p2": pa.array([value_2, value_1], value_type),
-            "val": pa.array([2, 2], pa.int64()),
+            "val": pa.array([2, 3], pa.int64()),
         }
     )
     expected_data = pa.table(
         {
             "p1": pa.array(["1", "1", "2", "2"], pa.string()),
             "p2": pa.array([value_1, value_2, value_1, value_2], value_type),
-            "val": pa.array([2, 2, 1, 1], pa.int64()),
+            "val": pa.array([3, 2, 1, 1], pa.int64()),
         }
     )
 
-    with pytest.raises(
-        DeltaError,
-        match="Generic DeltaTable error: Overwriting data based on predicate is not yet implemented",
-    ):
-        write_deltalake(
-            tmp_path,
-            sample_data,
-            mode="overwrite",
-            predicate="`p1` = 1",
-            engine="rust",
-        )
+    write_deltalake(
+        table_path,
+        sample_data,
+        mode="overwrite",
+        predicate="p1 = '1'",
+        engine="rust",
+    )
 
-        delta_table.update_incremental()
-        assert (
-            delta_table.to_pyarrow_table().sort_by(
-                [("p1", "ascending"), ("p2", "ascending")]
-            )
-            == expected_data
+    delta_table.update_incremental()
+    assert (
+        delta_table.to_pyarrow_table().sort_by(
+            [("p1", "ascending"), ("p2", "ascending")]
         )
+        == expected_data
+    )
+
+
+@pytest.mark.parametrize(
+    "value_1,value_2,value_type,filter_string",
+    [
+        (1, 2, pa.int64(), "1"),
+        (False, True, pa.bool_(), "false"),
+        (date(2022, 1, 1), date(2022, 1, 2), pa.date32(), "2022-01-01"),
+    ],
+)
+def test_replace_where_overwrite_partitioned(
+    tmp_path: pathlib.Path,
+    value_1: Any,
+    value_2: Any,
+    value_type: pa.DataType,
+    filter_string: str,
+):
+    table_path = tmp_path
+
+    sample_data = pa.table(
+        {
+            "p1": pa.array(["1", "1", "2", "2"], pa.string()),
+            "p2": pa.array([value_1, value_2, value_1, value_2], value_type),
+            "val": pa.array([1, 1, 1, 1], pa.int64()),
+        }
+    )
+    write_deltalake(
+        table_path, sample_data, mode="overwrite", partition_by=["p1", "p2"]
+    )
+
+    delta_table = DeltaTable(table_path)
+    assert (
+        delta_table.to_pyarrow_table().sort_by(
+            [("p1", "ascending"), ("p2", "ascending")]
+        )
+        == sample_data
+    )
+
+    replace_data = pa.table(
+        {
+            "p1": pa.array(["1", "1"], pa.string()),
+            "p2": pa.array([value_2, value_1], value_type),
+            "val": pa.array([2, 3], pa.int64()),
+        }
+    )
+    expected_data = pa.table(
+        {
+            "p1": pa.array(["1", "1", "2", "2"], pa.string()),
+            "p2": pa.array([value_1, value_2, value_1, value_2], value_type),
+            "val": pa.array([3, 2, 1, 1], pa.int64()),
+        }
+    )
+
+    write_deltalake(
+        table_path,
+        replace_data,
+        mode="overwrite",
+        partition_by=["p1", "p2"],
+        predicate="p1 = '1'",
+        engine="rust",
+    )
+
+    delta_table.update_incremental()
+    assert (
+        delta_table.to_pyarrow_table().sort_by(
+            [("p1", "ascending"), ("p2", "ascending")]
+        )
+        == expected_data
+    )
 
 
 def test_partition_overwrite_with_new_partition(
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -878,6 +878,12 @@ impl DeltaDataChecker {
         self
     }
 
+    /// Add the specified set of constraints to the current DeltaDataChecker's constraints
+    pub fn with_extra_constraints(mut self, constraints: Vec<Constraint>) -> Self {
+        self.constraints.extend(constraints);
+        self
+    }
+
     /// Create a new DeltaDataChecker
     pub fn new(snapshot: &DeltaTableState) -> Self {
         let invariants = snapshot.schema().get_invariants().unwrap_or_default();
--- a/crates/core/src/operations/write.rs
+++ b/crates/core/src/operations/write.rs
@@ -27,26 +27,36 @@
 
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use arrow_array::RecordBatch;
 use arrow_cast::can_cast_types;
 use arrow_schema::{DataType, Fields, SchemaRef as ArrowSchemaRef};
 use datafusion::execution::context::{SessionContext, SessionState, TaskContext};
+use datafusion::physical_expr::create_physical_expr;
+use datafusion::physical_plan::filter::FilterExec;
 use datafusion::physical_plan::{memory::MemoryExec, ExecutionPlan};
+use datafusion_common::DFSchema;
+use datafusion_expr::Expr;
 use futures::future::BoxFuture;
 use futures::StreamExt;
 use parquet::file::properties::WriterProperties;
 
+use super::datafusion_utils::Expression;
 use super::transaction::PROTOCOL;
 use super::writer::{DeltaWriter, WriterConfig};
 use super::{transaction::commit, CreateBuilder};
+use crate::delta_datafusion::expr::fmt_expr_to_sql;
+use crate::delta_datafusion::expr::parse_predicate_expression;
 use crate::delta_datafusion::DeltaDataChecker;
+use crate::delta_datafusion::{find_files, register_store, DeltaScanBuilder};
 use crate::errors::{DeltaResult, DeltaTableError};
-use crate::kernel::{Action, Add, PartitionsExt, StructType};
+use crate::kernel::{Action, Add, PartitionsExt, Remove, StructType};
 use crate::logstore::LogStoreRef;
 use crate::protocol::{DeltaOperation, SaveMode};
 use crate::storage::ObjectStoreRef;
 use crate::table::state::DeltaTableState;
+use crate::table::Constraint as DeltaConstraint;
 use crate::writer::record_batch::divide_by_partition_values;
 use crate::DeltaTable;
 
@@ -79,7 +89,6 @@ impl From<WriteError> for DeltaTableError {
 }
 
 /// Write data into a DeltaTable
-#[derive(Debug, Clone)]
 pub struct WriteBuilder {
     /// A snapshot of the to-be-loaded table's state
     snapshot: Option<DeltaTableState>,
@@ -94,7 +103,7 @@ pub struct WriteBuilder {
     /// Column names for table partitioning
     partition_columns: Option<Vec<String>>,
     /// When using `Overwrite` mode, replace data that matches a predicate
-    predicate: Option<String>,
+    predicate: Option<Expression>,
     /// Size above which we will write a buffered parquet file to disk.
     target_file_size: Option<usize>,
     /// Number of records to be written in single batch to underlying writer
@@ -154,7 +163,7 @@ impl WriteBuilder {
     }
 
     /// When using `Overwrite` mode, replace data that matches a predicate
-    pub fn with_replace_where(mut self, predicate: impl Into<String>) -> Self {
+    pub fn with_replace_where(mut self, predicate: impl Into<Expression>) -> Self {
         self.predicate = Some(predicate.into());
         self
     }
@@ -292,7 +301,8 @@ impl WriteBuilder {
 }
 
 #[allow(clippy::too_many_arguments)]
-pub(crate) async fn write_execution_plan(
+async fn write_execution_plan_with_predicate(
+    predicate: Option<Expr>,
     snapshot: Option<&DeltaTableState>,
     state: SessionState,
     plan: Arc<dyn ExecutionPlan>,
@@ -317,6 +327,14 @@ pub(crate) async fn write_execution_plan(
         DeltaDataChecker::new(snapshot)
     } else {
         DeltaDataChecker::empty()
+    };
+    let checker = match predicate {
+        Some(pred) => {
+            // TODO: get the name of the outer-most column? `*` will also work but would it be slower?
+            let chk = DeltaConstraint::new("*", &fmt_expr_to_sql(&pred)?);
+            checker.with_extra_constraints(vec![chk])
+        }
+        _ => checker,
     };
 
     // Write data to disk
@@ -361,6 +379,134 @@ pub(crate) async fn write_execution_plan(
         .concat()
         .into_iter()
         .collect::<Vec<_>>())
+}
+
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn write_execution_plan(
+    snapshot: Option<&DeltaTableState>,
+    state: SessionState,
+    plan: Arc<dyn ExecutionPlan>,
+    partition_columns: Vec<String>,
+    object_store: ObjectStoreRef,
+    target_file_size: Option<usize>,
+    write_batch_size: Option<usize>,
+    writer_properties: Option<WriterProperties>,
+    safe_cast: bool,
+    overwrite_schema: bool,
+) -> DeltaResult<Vec<Add>> {
+    write_execution_plan_with_predicate(
+        None,
+        snapshot,
+        state,
+        plan,
+        partition_columns,
+        object_store,
+        target_file_size,
+        write_batch_size,
+        writer_properties,
+        safe_cast,
+        overwrite_schema,
+    )
+    .await
+}
+
+async fn execute_non_empty_expr(
+    snapshot: &DeltaTableState,
+    log_store: LogStoreRef,
+    state: SessionState,
+    partition_columns: Vec<String>,
+    expression: &Expr,
+    rewrite: &[Add],
+    writer_properties: Option<WriterProperties>,
+) -> DeltaResult<Vec<Add>> {
+    // For each identified file perform a parquet scan + filter + limit (1) + count.
+    // If returned count is not zero then append the file to be rewritten and removed from the log. Otherwise do nothing to the file.
+
+    let input_schema = snapshot.input_schema()?;
+    let input_dfschema: DFSchema = input_schema.clone().as_ref().clone().try_into()?;
+
+    let scan = DeltaScanBuilder::new(snapshot, log_store.clone(), &state)
+        .with_files(rewrite)
+        .build()
+        .await?;
+    let scan = Arc::new(scan);
+
+    // Apply the negation of the filter and rewrite files
+    let negated_expression = Expr::Not(Box::new(Expr::IsTrue(Box::new(expression.clone()))));
+
+    let predicate_expr = create_physical_expr(
+        &negated_expression,
+        &input_dfschema,
+        &input_schema,
+        state.execution_props(),
+    )?;
+    let filter: Arc<dyn ExecutionPlan> =
+        Arc::new(FilterExec::try_new(predicate_expr, scan.clone())?);
+
+    // We don't want to verify the predicate against existing data
+    let add_actions = write_execution_plan(
+        Some(snapshot),
+        state,
+        filter,
+        partition_columns,
+        log_store.object_store(),
+        Some(snapshot.table_config().target_file_size() as usize),
+        None,
+        writer_properties,
+        false,
+        false,
+    )
+    .await?;
+
+    Ok(add_actions)
+}
+
+// This should only be called wth a valid predicate
+async fn prepare_predicate_actions(
+    predicate: Expr,
+    log_store: LogStoreRef,
+    snapshot: &DeltaTableState,
+    state: SessionState,
+    partition_columns: Vec<String>,
+    writer_properties: Option<WriterProperties>,
+    deletion_timestamp: i64,
+) -> DeltaResult<Vec<Action>> {
+    let candidates =
+        find_files(snapshot, log_store.clone(), &state, Some(predicate.clone())).await?;
+
+    let add = if candidates.partition_scan {
+        Vec::new()
+    } else {
+        execute_non_empty_expr(
+            snapshot,
+            log_store,
+            state,
+            partition_columns,
+            &predicate,
+            &candidates.candidates,
+            writer_properties,
+        )
+        .await?
+    };
+    let remove = candidates.candidates;
+
+    let mut actions: Vec<Action> = add.into_iter().map(Action::Add).collect();
+
+    for action in remove {
+        actions.push(Action::Remove(Remove {
+            path: action.path,
+            deletion_timestamp: Some(deletion_timestamp),
+            data_change: true,
+            extended_file_metadata: Some(true),
+            partition_values: Some(action.partition_values),
+            size: Some(action.size),
+            deletion_vector: action.deletion_vector,
+            tags: None,
+            base_row_id: action.base_row_id,
+            default_row_commit_version: action.default_row_commit_version,
+        }))
+    }
+    Ok(actions)
 }
 
 impl std::future::IntoFuture for WriteBuilder {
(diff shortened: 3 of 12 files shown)


crates/deltalake-core/src/operations/write.rs
```diff
@@ -314,6 +327,14 @@ pub(crate) async fn write_execution_plan(
     };
 
     let checker = DeltaDataChecker::new(snapshot);
+    let checker = match predicate {
+        Some(pred) => {
+            // TODO: get the name of the outer-most column? `*` will also work but would be slower
+            let chk = DeltaConstraint::new("*", &fmt_expr_to_sql(&pred)?);
```
ion-elgreco, line 333: Why is the predicate added as a delta constraint?

r3stl355: this is to check that incoming data also conforms to the predicate (this is how the Spark Delta works). I could create a separate checker for it but this checker is only used in the context of this function so I thought integrating with existing checks could be another option. Happy to separate of course which would actually let me throw a more meaningful error in case of predicate check failure (at the moment it's only the generic check error message)

r3stl355: Perhaps we should support this type of behaviour as well, either with the same configuration or a delta-rs table setting rust.delta.replaceWhere.constraintCheck.enabled ?

Good point but maybe deserves a separate issue? Delta is using `REPLACEWHERE_CONSTRAINT_CHECK_ENABLED` but default value is `true` so maybe add it in a separate PR?

r3stl355: Check is enabled by default (I came across the behavior incidentally when I tried writing to a new table with no properties set and it didn't let me). Setting is here: 
https://github.com/delta-io/delta/blob/9f9993db8a51fd83168186bba24041feffdbb566/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala#L753

ion-elgreco: Ah, then it would make sense to also set this by default when we create a new table with the create api. And handle the cases where in some tables this setting is missing, which I guess means the same as if it was set to False?

r3stl355: I disagree @ion-elgreco. Default value in Delta is `true` so there is no need to set it by default on create, and only set it if the caller has provided an explicit value. Well, it actually makes no harm to always setting it on create but

> And handle the cases where in some tables this setting is missing, which I guess means the same as if it was set to False?

Is not correct. In delta, if the setting is missing then the value is assumed `true` so we should adopt the same logic, right?

r3stl355: which also means that if I don't implement that table setting for this PR then I'll be matching the behavior of the Delta by default

r3stl355: Also, I presume we need to use the full config value as `spark.databricks.delta.replaceWhere.constraintCheck.enabled` if Delta is using that, right? (i've not touched the table properties yet)

r3stl355: Actually, this config does not look like a table property but only an engine configuration, not sure how we are supposed to handle it: https://github.com/delta-io/delta/blob/9f9993db8a51fd83168186bba24041feffdbb566/docs/source/delta-batch.md?plain=1#L714C12-L714C12

r3stl355:
It is passing the CI and with added docs I'm removing the draft (one of the Python tests is flaky, need to be fixed)

r3stl355:
looks like few extra checks were introduced recently so lint is picking up changes which are not mine, e.g. crates/deltalake-core/src/delta_datafusion/mod.rs:149:37

ion-elgreco:
> looks like few extra checks were introduced recently so lint is picking up changes which are not mine, e.g. crates/deltalake-core/src/delta_datafusion/mod.rs:149:37

 Could you create a separate PR to address that so main is fixed?

r3stl355:
> Could you create a separate PR to address that so main is fixed?

Ahh, fixed and pushed here before seeing your response. Can do another PR if they still break but leaving that for another day, done for today

r3stl355:
Looks like I am close to removing the draft, only need to fix this  flaky`tests/test_writer.py::test_replace_where_overwrite` - sometimes it fails, sometimes not even on my machine

python/tests/test_writer.py
```diff
@@ -871,6 +872,11 @@ def test_replace_where_overwrite(
     filter_string: str,
 ):
     table_path = tmp_path / sub_path
+
+    # Sometimes the path is already there resulting in failed test
+    if table_path.exists():
```
ion-elgreco, line 877: For some reason replacewhere overwrite does not work with a partitioned table, if you remove the partioning then it won't throw this table already exist erorr.

r3stl355: Not necessarily. It only happens sometimes, not always. See here the other run on Python 3.10 - 2 passed, 1 failed: https://github.com/delta-io/delta-rs/actions/runs/7356976767/job/20027915530?pr=1996#step:9:261
I thought it was because of existing path because that fixed the issue on my machine but looks like not

r3stl355: it could be that there is a race condition somewhere related to partitioning. I'll investigate but for now I'll comment out the paritioning to get the full build

Blajda:
> I copied the part of code from `delete` operation because there is no way to call that code in `delete` directly from `write` - should I look into extracting that code from `delete` to somewhere central?

I think this should be done. Delete is one of the things we should keep DRY which will help when deletion vectors are implemented. All operations will benefit instead of needing to update code in different location.

My 2 cents is that we shouldn't modify the writer to implement this operation but rather should have a brand new operation structured in a manner similar to merge, delete, and update. 

Some builder interface like below
```
DeltaOps(table)
    .replace_where(predicate, source_data)
    .with_check_constraint(false)
    .await?
```

crates/deltalake-core/src/operations/write.rs
```diff
@@ -960,4 +1132,95 @@ mod tests {
...
+            Arc::clone(&schema),
+            vec![
+                Arc::new(arrow::array::StringArray::from(vec!["C"])),
+                Arc::new(arrow::array::Int32Array::from(vec![50])),
+                Arc::new(arrow::array::StringArray::from(vec!["2023-01-01"])),
+            ],
+        )
+        .unwrap();
+
+        let table = DeltaOps(table)
```
Blajda, line 1183: This test case should be split into two.
1. Happy path
2. Failure due to constraint violation

There should also be a test using a table with partition column since it's a common source of bugs

r3stl355: Sure, no problem, was already planning to do both

r3stl355:
My 2 cents is that we shouldn't modify the writer to implement this operation but rather should have a brand new operation structured in a manner similar to merge, delete, and update.

@Blajda I understand that you suggesting a standalone operator just for `replaceWhere`, correct? Which means:
1. Removing a `predicate` from `write`, which is fine as it wasn't implemented anyway
2. Exposing a function from`delete` which behaves similar to `writer::write_execution_plan` by returning a set of actions
3. [MAYBE] Possibly exposing a version of `write_execution_plan` function that takes a predicate from `write`

If this is what you were suggesting I can give it a try

Blajda:
@r3stl355 Yes I agree with point 1. For 2. one approach is to factor out a function that returns a tuple of files to be removed and an optional stream of record batches. I'm not sure on what 3. would provide.

r3stl355:
Hmm, this was unexpected but it is only a problem with the test itself, not the library. 

Looks like `utime::set_file_times` is expecting timestamp in seconds whereas our timestamps are in millis (for some reason it only broke on Windows).

r3stl355:
> @r3stl355 Yes I agree with point 1. For 2. one approach is to factor out a function that returns a tuple of files to be removed and an optional stream of record batches. I'm not sure on what 3. would provide.

The 3. would allow to check if the new data conforms to the `replaceWhere` predicate. Current implementation of the public`write_execution_plan` does not do that and if I don't exposed a public function checking that the new data conforms to a predicate (e.g. `write_execution_plan_with_predicate`) then I would have to implement it within the `replaceWhere` op effectively copying the code

ion-elgreco:
I'm not sure if I agree with taking this out of write. It's still a write operation

python/tests/test_writer.py
r3stl355: Problem found, was an existing code for constraints, fix is on the way

crates/deltalake-core/src/operations/write.rs
r3stl355: Done both

python/tests/test_writer.py
ion-elgreco: Nice find!

MrPowers:
What's the user-facing public interface for Python?

r3stl355:
Good old `write_deltalake` with `predicate` parameter. It was already there before this PR but was not implemented. All should be working with this PR

r3stl355:
I also added docs and examples in docs for both Python and Rust

MrPowers:
@r3stl355 - ah, got it.  Here's the delta-spark syntax:

```python
(
    df.write.format("delta")
    .option("replaceWhere", "number > 2")
    .mode("overwrite")
    .save("tmp/my_data")
)
```

Looks like the proposed syntax is as follows:

```python
write_deltalake(
    table_path,
    data,
    mode="overwrite",
    predicate="id = '1'",
    engine="rust",
)
```

I almost feel like the `mode` should be `replace_where`.  Thoughts?

ion-elgreco:
@MrPowers I don't think it makes sense to introduce another mode because replace where without a predicate wouldnt be possible

docs/usage/writing-delta-tables.md
```diff
@@ -50,4 +50,17 @@ that partition or else the method will raise an error.
 ```
 
 This method could also be used to insert a new partition if one doesn't
-already exist, making this operation idempotent.
\ No newline at end of file
+already exist, making this operation idempotent.
+
+## Overwriting part of the table data with `replaceWhere`
```
ion-elgreco, line 55: Using the replaceWhere naming is a bit odd, it's not spark-delta and our api interface uses predicate solely (or the partition_filter which we will eventually phase out).

ion-elgreco (commented):
Can you one more test cases to match the pyarrow partition filter overwrite test cases:

A test case where the data is not meeting the constraint. Also I wonder if you could capture that constraint error and give a better error msg

ion-elgreco:
@MrPowers either we throw an error in Python or ignore it. 

We can also add an overload specifically for overwrite mode to show the predicate, and other modes not this should also help indicating that.

MrPowers:
Yea, I'm not sure about this design then.

Maybe I have PTSD from pandas [to_csv](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html), haha.

They have `float_format` and `decimal` arguments and have to document stuff like "Format string for floating point numbers. If a Callable is given, it takes precedence over other numeric formatting parameters, like decimal."

I prefer no dependencies between the different arguments of public interface methods.

docs/usage/writing-delta-tables.md
r3stl355: Agree, it's not ideal because it doesn't match the Python's parameter name, more like a concept, happy to change if you have a better name in mind

crates/deltalake-core/src/operations/write.rs
```diff
@@ -81,7 +89,7 @@ impl From<WriteError> for DeltaTableError {
 }
 
 /// Write data into a DeltaTable
-#[derive(Debug, Clone)]
+// #[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
r3stl355: I removed it because it was compiling when I changed a predicate type from String to Expression. I checked and and didn't find a place where those traits were used so instead of adding them to Expressions I thought it would make sense to remove them from here. I actually need to delete instead of commenting out
Besides, other builders don't have that either

crates/deltalake-core/src/operations/write.rs
```diff
@@ -314,6 +324,14 @@ pub(crate) async fn write_execution_plan(
     };
 
     let checker = DeltaDataChecker::new(snapshot);
```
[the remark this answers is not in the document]
r3stl355: Couple of reasons I had in mind, which made me choose this approach, happy to reconsider if I missed something more important that would make the mutable implementation more important

1. DeltaDataChecker is using a builder style we used in many places which does not require the builder variable itself to be mutable
2. At the moment I am creating a checker here but if later for some reason it was provided then I would not want to change it in place.

r3stl355:
> Can you add one more test cases to match the pyarrow partition filter overwrite test cases? 

Will do

> A test case where the data is not meeting the constraint. Also I wonder if you could capture that constraint error and give a better error msg

The test case with data not meeting the constraint is already there

Using a more specific error would be tricky and would require creating a separate checker. I was already thinking about that but could not find a sleek way of doing so

crates/deltalake-core/src/operations/write.rs
```diff
@@ -96,7 +104,8 @@ pub struct WriteBuilder {
     /// Column names for table partitioning
     partition_columns: Option<Vec<String>>,
     /// When using `Overwrite` mode, replace data that matches a predicate
-    predicate: Option<String>,
+    // predicate: Option<String>,
```
ion-elgreco, line 107: Also need to remove this comment : )

crates/deltalake-core/src/operations/write.rs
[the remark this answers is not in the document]
ion-elgreco: Because it uses this builder style it leans itself wel to do this kind of structure, see here: 

https://github.com/delta-io/delta-rs/blob/6d41b37de413f17b28f2b8ece8eb72b0b1d27215/python/src/lib.rs#L102-L106

crates/deltalake-core/tests/command_restore.rs
```diff
@@ -128,7 +131,20 @@ async fn test_restore_by_datetime() -> Result<(), Box<dyn Error>> {
         .await?;
     let timestamp = meta.last_modified.timestamp_millis();
     let naive = NaiveDateTime::from_timestamp_millis(timestamp).unwrap();
-    let datetime: DateTime<Utc> = Utc.from_utc_datetime(&naive);
+    let datetime: DateTime<Utc> = Utc.from_utc_datetime(&naive); // - Duration::hours(1);
```
ion-elgreco, line 134: Is this still relevant after your fix of the test?

docs/usage/writing-delta-tables.md
ion-elgreco: This fits better:

Overwriting part of the table data with a SQL `predicate`

crates/deltalake-core/tests/command_restore.rs
r3stl355: Oh yeah, missed that, will delete

r3stl355:
@MrPowers re
I almost feel like the mode should be replace_where. Thoughts?

@Blajda suggested another approach - creating a dedicated operator for `replace_where` and taking it out of `write`, which I am happy to implement but would require a consensus that this is a good idea

crates/deltalake-core/src/operations/write.rs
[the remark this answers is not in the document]
r3stl355: Yes I understand what you mean, it is less verbose but it's just a style in my opinion, but I'd end up with a mutable. Again, happy to change if that's important

docs/usage/writing-delta-tables.md
r3stl355: @MrPowers - your take on this?

crates/deltalake-core/src/operations/write.rs
```diff
@@ -960,4 +1132,177 @@ mod tests {
...
+            "+----+-------+------------+",
+        ];
+        let actual = get_data(&table).await;
+        assert_batches_sorted_eq!(&expected, &actual);
+    }
+
+    #[tokio::test]
+    async fn test_replace_where_fail_not_matching_predicate() {
+        let tmp_dir = tempdir::TempDir::new("test").unwrap();
+        let tmp_path = std::fs::canonicalize(tmp_dir.path()).unwrap();
```
Blajda, line 1203: Instead of writing to disk the suggested approach is to use a memory table.
See: https://github.com/delta-io/delta-rs/blob/main/crates/deltalake-core/src/operations/merge/mod.rs#L1527-L1528

crates/deltalake-core/src/writer/test_utils.rs
```diff
@@ -334,6 +334,21 @@ pub mod datafusion {
             .unwrap()
     }
 
+    pub async fn get_data_sorted(table: &DeltaTable, columns: &str) -> Vec<RecordBatch> {
```
Blajda, line 337: You don't need to sort the data yourself and can reuse this utility: https://github.com/delta-io/delta-rs/blob/main/crates/deltalake-core/src/writer/test_utils.rs#L325

Blajda (commented):
Just a few nitpick on how to setup test tables and reusing existing test methods. I don't think the request to have ReplaceWhere as a separate operation should block merging.

crates/deltalake-core/src/operations/write.rs
r3stl355: The reason I used a disk location was because I could not find a way to load the table from memory after a failed write op. Few lines liner I have `table.is_err()` so all my attempt to reload it after that were unsuccessful so I ended up using file system. Is there a way?

crates/deltalake-core/src/writer/test_utils.rs
r3stl355: The reason I added that extra method was that, unlike a straightforward write, after `replaceWhere` `get_data` was returning rows in some mixed order and `assert_batches_sorted_eq` was not able to match them. I could of course set the `expected` to be in the same order as returned by `get_data` but I thought that it could still be vulnerable to some future changes that could for example change the order returned by `get_data`. 

I can remove this method and use `get_data` with `expected` being in a specific order as I am getting today if you think that order will never change.

crates/deltalake-core/src/operations/write.rs
Blajda: It might be possible to create a clone of the table prior to failure then update the clone to the latest version after failure. not 100% sure on that.

crates/deltalake-core/src/writer/test_utils.rs
Blajda: I'm a bit surprised it was failing. The order given by `get_data` should not matter since the assert marco performs a sort on the string representation of the batches. 
Your not supposed to set the `expected` string to whatever `get_data` returns. The order depends on the string representation. When using this macro I typically put some garbage string for `expected`. let the test fail, manually validate the `actual` results (which will be ordered) and then update `expected` with actual.

r3stl355: That's exactly what I did to find out what the `actual` was. I'll try again using only `get_data` and `assert_batches_sorted_eq`, maybe I messed something up when I tried before

crates/deltalake-core/src/operations/write.rs
r3stl355: Cool, I'll give that a try

crates/deltalake-core/src/writer/test_utils.rs
r3stl355: Ahh, I remembered the other reason now, `get_data` was messing up the column order too so I needed a version of `get_data` where I could pass the column names (`get_data` uses `*`). Once I added that function I decided to include `order by` too. Here is what I get in my test if I use `get_data`:
```
expected:

[
    "+----+-------+------------+",
    "| id | value | modified   |",
    "+----+-------+------------+",
    "| A  | 11    | 2024-02-02 |",
    "| A  | 13    | 2024-02-02 |",
    "| A  | 15    | 2024-02-01 |",
    "| B  | 2     | 2021-02-02 |",
    "| B  | 4     | 2021-02-01 |",
    "| B  | 8     | 2021-02-01 |",
    "| B  | 9     | 2021-02-01 |",
    "+----+-------+------------+",
]
actual:

[
    "+------------+----+-------+",
    "| modified   | id | value |",
    "+------------+----+-------+",
    "| 2021-02-01 | B  | 4     |",
    "| 2021-02-01 | B  | 8     |",
    "| 2021-02-01 | B  | 9     |",
    "| 2021-02-02 | B  | 2     |",
    "| 2024-02-01 | A  | 15    |",
    "| 2024-02-02 | A  | 11    |",
    "| 2024-02-02 | A  | 13    |",
    "+------------+----+-------+",
]
```

Blajda: Got it. Yeah this is just how datafusion functions with partition columns. You can keep that utility function to maintain column order.

crates/deltalake-core/src/operations/write.rs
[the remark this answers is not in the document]
r3stl355: I'm leaving this as is for now, it's a small change, can be easily done as a small PR if needed

crates/deltalake-core/src/operations/write.rs
r3stl355: @Blajda I tried few options as you suggested but to no avail. DeltaTable is not clone-able. The closest I could get was serializing/de-ser but looks like table memory is lost in the process so I can no longer refer to files (even though ` table.get_latest_version()` of the deserialized table returns a correct value, this is likely just a deserialised `self.version` value without touching the table in memory)

r3stl355:
ohh crap, what just happened here, how all these commits ended up on my branch 🤷

ion-elgreco:
@r3stl355 maybe something went wrong with rebasing?

r3stl355:
> maybe something went wrong with rebasing?

@ion-elgreco merge had not issues, the only strange thing was git saying that I has 1 diverted commit from my remote and recommended a put then push - I did that 🤦

ion-elgreco:
Strange.. then I guess you have to force push a previous state

r3stl355:
Yep, git just piled older commits onto the HEAD, strange to see commits from December appear ahead of mine :) ```
commit 0544b9591210ade3bd3a7a863bbe2d205df4361f
Author: Robert Pack <robstar.pack@gmail.com>
Date:   Wed Jan 3 10:42:39 2024 +0100

    fix: windows tests

commit 0ab026a13371e47ae41780bdbc8526c6d198eddd
Author: R. Tyler Croy <rtyler@brokenco.de>
Date:   Mon Dec 11 10:00:47 2023 -0800

    Refactor AWS specific code into the deltalake-aws crate
    
    There are a number of changes here to untangle the coupling inside of
    deltalake-core to allow deltalake-aws to be separated properly

commit 66262064b8a8bd4847630ce8955eeaff5beaa5a9
Author: Nikolay Ulmasov <ulmasov@hotmail.com>
Date:   Wed Jan 10 21:55:01 2024 +0000

    tidy up
    
    Signed-off-by: Nikolay Ulmasov <ulmasov@hotmail.com>
```

**Comment by r3stl355** [comment]
Is it OK if I reset and force-push?

**Comment by r3stl355** [comment]
Actually, that does not look simple to do so I may just choose an easier way and just create a new PR

**Comment by r3stl355** [comment]
OK, looks better now. There were some warnings and errors on my local build unrelated to my commit so the build may fail but that's for an upstream to fix. @ion-elgreco can you please start the workflow? thanks

File: crates/deltalake-core/src/operations/write.rs (hunk 1 above)
Comment (ion-elgreco) [reply crates/deltalake-core/src/operations/write.rs:328]: @roeap  or @rtyler, I don't recall who exactly, mentioned we should support these session settings as well. We could simply extend this in a future PR where we parse rust.delta environment variables to check which mode of predicate overwrite needs to be executed.

**Comment by ion-elgreco** [comment]
@r3stl355 can you add an overload to write_deltalake for mode='overwrite' and engine='rust' so that only that specific overload allows the predicate as input. Type checkers should then throw warnings if you pass predicate to any other mode or engine.

**Comment by r3stl355** [comment]
Do you mean something like this (only showing relevant lines)?
```
@overload
def write_deltalake(
    ...
    mode: Literal["overwrite"],
    ...
    engine: Literal["rust"],
    ...
```
I'm not sure if that really works. Won't the main overload catch it all? I tried it quick here, no warnings

**Comment by ion-elgreco** [comment]
@r3stl355 here are the overloads: https://delta-users.slack.com/archives/C013LCAEB98/p1704303063159459?thread_ts=1704286042.601189&cid=C013LCAEB98

**Comment by r3stl355** [comment]
@ion-elgreco I know where the overloads are, just don't really see a point in having them, maybe I'm missing something fundamental. In the end, there the main catch-all overload which will always be acceptable

**Comment by ion-elgreco** [comment]
@r3stl355 the link includes the additional overload I was requesting. 

Without an overload all combinations of inputs are valid but that's not the case since we have parameters that are specific for a engine and a mode, so you need a function overload to hint which param combinations are valid.

**Comment by ion-elgreco** [comment]
@r3stl355 I think we are good to go, but we likely want to wait on merging so that the arrow log handling of @roeap can land first

**Comment by r3stl355** [comment]
I re-based using @roeap's PR but had to force-push. On a positive, it also looks like I can squash easily once the CI passes.

I am a bit less happy with my code now after `snapshot` was made an `Option`, so maybe another round of reviews would be appropriate.

**Review by roeap — COMMENTED** [review]
Sorry it took so long, but finally got around to do some reviewing.

> I am a bit less happy with my code now after snapshot was made an Option

This is actually also one of my biggest concerns annoyances right now, and kind of boild down to what purpose each of these structs serve in our codebase. Previously, `DeltaTableState` took no opinion on if `Metadata` and `Protocol` exist. However if we cannot get these, there really is not anything meaningful we can do with a delta table. However `DeltaTable` has some operations that work on empty locations etc. As a result, we were raising errors all throughout our operations if the back then optionsal metadata / protocol actions were not available. While this requires some more refactoring, all operations  (except the ones that can cretate a table) should always require a valid snapshot.

btw. `DeltaTableState` is by now also just a thin wrapper around a snapshot, which I also hope to remove soon :).

File: crates/deltalake-core/src/operations/write.rs [hunk 11]
```diff
@@ -363,6 +381,133 @@ pub(crate) async fn write_execution_plan(
         .collect::<Vec<_>>())
 }
 
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn write_execution_plan(
+    snapshot: Option<&DeltaTableState>,
+    state: SessionState,
+    plan: Arc<dyn ExecutionPlan>,
+    partition_columns: Vec<String>,
```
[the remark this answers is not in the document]
roeap: after reading your comment around the `Option`, it seems this is something we should look into. i.e. we can write something when we create a new table, which is why an option is allowed here. but having to deal with that option for other operations that write, but do require a snapshot seems annoying. Haven't looked into it yet, but my feeling is we somehow need to accommodate that.

On the operations where we do create a table, we of course also create everything we need from a snapshot. SO maybe we should try and somehow isolate the code paths where we may not have one?

r3stl355:
Thank you for the review @roeap

crates/deltalake-core/src/operations/write.rs
```diff
@@ -940,4 +1118,168 @@ mod tests {
...
+            "+----+-------+------------+",
+        ];
+        let actual = get_data(&table).await;
+        assert_batches_sorted_eq!(&expected, &actual);
+    }
+
+    #[tokio::test]
+    async fn test_replace_where_fail_not_matching_predicate() {
+        // Using a temp file-based table instead of `memory://` table because we need to reload the table
+        // after a failed write attempt, which seem to be impossible for memory tables
```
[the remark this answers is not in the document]
r3stl355: It looks like something that didn't work few weeks ago works now so I'll refactor

TLDR;

After L1222-L1227 the table is in error state so I need to reload it to check it's current state. I do that on L1229 but if I use `memory://` I get a `NotATable("no log files")` error. 

I tried cloning the table's `state` and `logstore` before L1222 but that was causing a different error when I was trying to read the table in L1230. I wanted to get the exact error just now, tried it again and ... it works. Something may have changed since my last try because I am using the same code that didn't work few weeks ago.

crates/deltalake-core/src/operations/write.rs
```diff
@@ -940,4 +1118,168 @@ mod tests {
...
+            ],
+        )
+        .unwrap();
+
+        let table = DeltaOps(table)
+            .write(vec![batch_fail])
+            .with_save_mode(SaveMode::Overwrite)
+            .with_replace_where(col("id").eq(lit("C")))
+            .await;
+        assert!(table.is_err());
```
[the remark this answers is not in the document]
r3stl355: This section tests the scenario when an incoming data does not conform to predicate - predicate need to replace records with `C` but the new data has `D` in it. This throws an error, and it's a default behavior (also in Spark Delta), so I discussed this with @ion-elgreco  and made it default behavior too. 

Later we may need to add a flag to allow non-predicate-matching records to be inserted (Spark has that) but that's an engine/operation property, not a table property, and I didn't do that for now as it means I'd be touching more code. Default behavior should suffice for now (and i put that into the docs too)

crates/deltalake-core/src/operations/write.rs
```diff
@@ -363,6 +381,133 @@ pub(crate) async fn write_execution_plan(
...
+pub(crate) async fn write_execution_plan(
+    snapshot: Option<&DeltaTableState>,
+    state: SessionState,
+    plan: Arc<dyn ExecutionPlan>,
+    partition_columns: Vec<String>,
+    object_store: ObjectStoreRef,
+    target_file_size: Option<usize>,
+    write_batch_size: Option<usize>,
+    writer_properties: Option<WriterProperties>,
+    safe_cast: bool,
```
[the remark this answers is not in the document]
r3stl355: I agree but perhaps it deserves a dedicated issues because `write_execution_plan` is used by many operations and probably not a good idea to make it part of this PR. I'll create a new issue for this

crates/deltalake-core/src/operations/write.rs
```diff
@@ -363,6 +381,133 @@ pub(crate) async fn write_execution_plan(
...
+    )?;
+    let filter: Arc<dyn ExecutionPlan> =
+        Arc::new(FilterExec::try_new(predicate_expr, scan.clone())?);
+
+    // We don't want to verify the predicate against existing data
+    let add_actions = write_execution_plan(
+        Some(snapshot),
+        state.clone(),
+        filter.clone(),
+        table_partition_cols.clone(),
```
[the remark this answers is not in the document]
r3stl355: done

crates/deltalake-core/src/operations/write.rs
```diff
@@ -363,6 +381,133 @@ pub(crate) async fn write_execution_plan(
...
+        log_store.object_store(),
+        Some(snapshot.table_config().target_file_size() as usize),
+        None,
```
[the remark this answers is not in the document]
r3stl355: I've done some refactoring to reduce recreation/cloning of partition_columns but until we change `write_execution_plan` to either accept an Option for partition_columns (or completely drop that parameter and get `partition_columns` from `snapshot` as you suggested in the other comment), I don't think there is much I can do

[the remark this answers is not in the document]
roeap: I opened some issues regarding that, and you are right, this is beyond the scope of this PR.

roeap (approved):
Thanks for your patience with this PR @r3stl355, and great work!

I'll leave it open for a bit, since @MrPowers had some questions around the APIs, but form my end passing the predicate to `write_deltalake` seems reasonable. While this function already takes a lot of parameters, I feel this PR does not make it worse :).

ion-elgreco:
@roeap once we drop the Pyarrow writer the write_deltalake will be much leaner

r3stl355:
> this PR does not make it worse :).

minimum standard achieved 😁 . I'll create couple of related issues

r3stl355:
@MrPowers do you still have those questions around API? If not then this should be ready to merge (and I squashed all my commits into one as agreed)

roeap:
@r3stl355 - it seems we have seen a bigger rename of files before I could get that in :( - sorry oyu have to resolve once more.

Since we haven't heard back, I would just merge once we are green again ...

r3stl355:
Thanks @roeap , should be good now

ion-elgreco (approved):

ion-elgreco (approved):
Thanks Niko! Great feature :) One step closer to deprecating the PyArrow writer now

PR #1999, opened 2023-12-29
chore: fix CI breaking lint issues

r3stl355:
# Description
Fixed few warnings picked up by lint


r3stl355:
@ion-elgreco this is the fix for those lint issues. Oddly, some were not getting picked up on my laptop but did in github actions run, hence the reason for multiple commits

ion-elgreco (approved):

Issue #1925, opened 2023-11-30
CI integration test fails randomly:  test_restore_by_datetime


ion-elgreco:
One of our tests (test_restore_by_datetime) in the CI - integration tests sometimes randomly fails: https://github.com/delta-io/delta-rs/actions/runs/7044053111/job/19170981661?pr=1924

r3stl355:
There are 3 writes to the table in that test, could all 3 be within the same millisecond? The test is picking up v1 timestamp then retrieving a version for it. The error can be thrown if the snapshot version also has the same timestamp

r3stl355:
Interesting results after some extra tracing within this test

All the history timestamps are [1703869558237, 1703869559278, 1703869560403, 1703869561411] showing that they are in sequence with 1 sec gap between them so I doubt the issue can be fixed by adding a gap between writes so I am going to remove the current delay between writes from the test

Test is looking for a timestamp 1703869559278, which is a correct value at index 1. I confirmed that is gets converted to a correct DateTime (converting it to timestamps gives a correct value: 1703869559278)

So looks like there is something wrong with the lookup logic

r3stl355:
Problem located - this is not a flaky team IMO but a bug. 

It starts here with creating a new DeltaTable with no history/timestamps loaded: https://github.com/delta-io/delta-rs/blob/02e26e5ecd7a8a64609bd21e8cb30e9bad4cc17b/crates/deltalake-core/src/operations/restore.rs#L142C44-L142C44

The consequences surface here: https://github.com/delta-io/delta-rs/blob/02e26e5ecd7a8a64609bd21e8cb30e9bad4cc17b/crates/deltalake-core/src/table/mod.rs#L884C28-L884C49 in turn leading to https://github.com/delta-io/delta-rs/blob/02e26e5ecd7a8a64609bd21e8cb30e9bad4cc17b/crates/deltalake-core/src/table/mod.rs#L643

Consequences: If your writes are fast enough to millisecond between the start of write and commit to storage - you are good, otherwise not and the restore to DataTime will not works as expected, possibly resulting in an earlier version being restored.

I'll submit a fix in the next days

rtyler:
I believe @r3stl355 fixed this

PR #2000, opened 2023-12-29
fix: failing test est_restore_by_datetime

r3stl355:
# Description
This started as a fix for the issue occurring on CI Mac only. Turns out it is not Mac only but could also happen in other systems if commit takes more than a millisecond. Fix changes the following:

1. Previous logic: `get_version_timestamp` would get the timestamp from the `last_modified` property of the log entry file
2. New logic: `get_version_timestamp` will first attempt to get the timestamp from `CommitInfo` and will revert to previous logic of getting the timestamp from file's `last_modified` if `CommitInfo` does not contain a timestamp. This will also help restoring a correct version for tables which get copied to a different location (e.g. for in case of DR)

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/1925

### DIFF
--- a/crates/deltalake-core/src/table/mod.rs
+++ b/crates/deltalake-core/src/table/mod.rs
@@ -586,6 +586,7 @@ impl DeltaTable {
 
             debug!("merging table state with version: {new_version}");
             let s = DeltaTableState::from_actions(actions, new_version)?;
+
             self.state
                 .merge(s, self.config.require_tombstones, self.config.require_files);
             if self.version() == max_version {
@@ -640,14 +641,23 @@ impl DeltaTable {
         match self.version_timestamp.get(&version) {
             Some(ts) => Ok(*ts),
             None => {
-                let meta = self
-                    .object_store()
-                    .head(&commit_uri_from_version(version))
-                    .await?;
-                let ts = meta.last_modified.timestamp_millis();
+                let timestamp: Option<i64> = if !self.state.commit_infos().is_empty() {
+                    self.state.commit_infos().last().unwrap().timestamp
+                } else {
+                    None
+                };
+                let ts = match timestamp {
+                    Some(ts) => ts,
+                    _ => {
+                        let meta = self
+                            .object_store()
+                            .head(&commit_uri_from_version(version))
+                            .await?;
+                        meta.last_modified.timestamp_millis()
+                    }
+                };
                 // also cache timestamp for version
                 self.version_timestamp.insert(version, ts);
-
                 Ok(ts)
             }
         }
@@ -881,6 +891,7 @@ impl DeltaTable {
         while min_version <= max_version {
             let pivot = (max_version + min_version) / 2;
             version = pivot;
+            self.load_version(version).await?;
             let pts = self.get_version_timestamp(pivot).await?;
             match pts.cmp(&target_ts) {
                 Ordering::Equal => {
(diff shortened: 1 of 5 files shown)


r3stl355:
It wasn't meant to pass the CI 🤦 as I'm still trying to trace but I think I know where the issue is - during the version search a blank table is initiated which has no timestamp stored and `get_version_timestamp` is called on it which ends up retrieving the timestamp from the file which I believe leads to an issue. 
I'll do further tests in my GitHub, will be faster and less noisy

r3stl355:
@ion-elgreco are you able to start the workflows here?

r3stl355:
There is good in every bad - different tests are failing now. Of course I didn't run them locally 🤦

r3stl355:
Hmm, this was unexpected overflow failure here but it is only a problem with the test itself, not the library.

Looks like utime::set_file_times is expecting timestamp in seconds whereas our timestamps are in millis (for some reason it only broke on Windows).

crates/deltalake-core/src/table/mod.rs
```diff
@@ -637,17 +638,31 @@ impl DeltaTable {
...
-                let ts = meta.last_modified.timestamp_millis();
+                // Load the version specified
+                if self.version() != version {
+                    self.load_version(version).await?;
+                }
+                let timestamp: Option<i64> = if !self.state.commit_infos().is_empty() {
+                    self.state.commit_infos().last().unwrap().timestamp
+                } else {
+                    None
+                };
```
ion-elgreco, line 653: This looks obsolete because `self.version_timestamp` is a HashMap of the table versions and their timestamps which to my understanding are coming from the transaction log.

crates/deltalake-core/tests/time_travel.rs
```diff
@@ -12,9 +12,19 @@ async fn time_travel_by_ds() {
...
-        let ts = ds_to_ts(ds);
-        utime::set_file_times(Path::new(log_dir).join(fname), ts, ts).unwrap();
+        let (file_ts, commit_ts) = ds_to_ts(ds);
+        let file = log_dir_path.join(fname);
+        let contents = std::fs::read_to_string(&file).unwrap();
+        let my_reg = regex::Regex::new(r#"("timestamp"):\s*\d+([\s,}])"#).unwrap();
+        let new_contents = my_reg
+            .replace(&contents, format!("$1:{}$2", commit_ts))
+            .to_string();
+        std::fs::write(&file, new_contents).unwrap();
```
ion-elgreco, line 26: Why are we rewriting the commit files?

crates/deltalake-core/src/table/mod.rs
ion-elgreco: Ah it seems we are not initializing any timestamp:version in the hashmap, only when it's not there its cached.

crates/deltalake-core/src/table/mod.rs
```diff
@@ -637,17 +638,31 @@ impl DeltaTable {
...
         match self.version_timestamp.get(&version) {
             Some(ts) => Ok(*ts),
             None => {
-                let meta = self
-                    .object_store()
-                    .head(&commit_uri_from_version(version))
-                    .await?;
-                let ts = meta.last_modified.timestamp_millis();
+                // Load the version specified
+                if self.version() != version {
```
ion-elgreco, line 646: Instead of loading the version each time, we could also just do this

`self.state.commit_infos().get(version).unwrap().timestamp`

ion-elgreco: Just think aloud here but in general you shouldn't be able to restore to a version of a table that's above the current loaded table version

r3stl355:
Closing this because a simpler fix for the test only was implemented in #2010

PR #2002, opened 2023-12-30
docs: add writer properties to docs

ion-elgreco:
# Description
Forgot to add WriterProperties to the docs page and mark a deprecation in the docs.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -1176,7 +1176,10 @@ class TableMerger:
         write_batch_size: Optional[int] = None,
         max_row_group_size: Optional[int] = None,
     ) -> "TableMerger":
-        """Pass writer properties to the Rust parquet writer, see options https://arrow.apache.org/rust/parquet/file/properties/struct.WriterProperties.html:
+        """
+        !!! warning "Deprecated"
+            Use `.merge(writer_properties = WriterProperties())` instead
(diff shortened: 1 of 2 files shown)


python/deltalake/table.py
```diff
@@ -1176,7 +1176,10 @@ def with_writer_properties(
         write_batch_size: Optional[int] = None,
         max_row_group_size: Optional[int] = None,
     ) -> "TableMerger":
-        """Pass writer properties to the Rust parquet writer, see options https://arrow.apache.org/rust/parquet/file/properties/struct.WriterProperties.html:
+        """
+        !!! warning "Deprecated"
+            Use `.merge(writer_properties = WriterProperties())` instead
+        Pass writer properties to the Rust parquet writer, see options https://arrow.apache.org/rust/parquet/file/properties/struct.WriterProperties.html:
```
MrPowers, line 1182: Can we use something other than "Rust Parquet writer" for the Python API docs?

I would rather link to our docs than a third party page?  Can we add a "Python version" of all these options to the documentation?

ion-elgreco: There is a short description of each argument, I only kept the rust docs link in there in case people want to request more properties to be exposed.

ion-elgreco: @MrPowers what shall we do? Would be good to merge this since the new release contains the WriterProperties but the docs not yet if we don't merge this : P

MrPowers: @ion-elgreco - merge away!  We can always improve the docs later!

PR #2009, opened 2024-01-01
feat(rust): add more commit info to most operations

ion-elgreco:
# Description
Some operations were not writing the readVersion and operationMetrics to the commitInfo. This adds that to those operations as well.


ion-elgreco:
@roeap I can also undo the change of the commit in restore, then we can look at that another time?

ion-elgreco:
@rtyler @roeap I've removed the high level commit and reverted back to the old version for restore : )

crates/deltalake-core/src/operations/delete.rs
```diff
@@ -249,6 +249,19 @@ async fn execute(
...
+    };
+
+    app_metadata.insert("readVersion".to_owned(), snapshot.version().into());
+
+    let delete_metrics = serde_json::to_value(metrics.clone());
+
+    if let Ok(map) = delete_metrics {
+        app_metadata.insert("operationMetrics".to_owned(), map);
+    }
+
```
roeap, line 264: I think we can avoid cloning here.

```suggestion
    app_metadata.insert("readVersion".to_owned(), snapshot.version().into());

    if let Ok(map) = serde_json::to_value(&metrics) {
        app_metadata.insert("operationMetrics".to_owned(), map);
    }

```

roeap (commented):
Looking good so far.

I do think we can avoid cloning these metrics, and with that also don't need the `Clone` implementations.

roeap (approved):

PR #2024, opened 2024-01-03
feat: introduce schema evolution on RecordBatchWriter

### LINKED ISSUE #1386: Schema evolution `mergeSchema` support
# Description

**Use Case**

rtyler:
This commit introduces the `WriteMode` enum and the ability to specify writes which should enable [schema
evolution](https://delta.io/blog/2023-02-08-delta-lake-schema-evolution/).

The result of this is a new `metaData` action added to the transaction log with the write which reflects the updated schema

There are some caveats however such as all writes must include non-nullable columns.

Fixes #1386

Sponsored-by: Raft, LLC.

### DIFF
--- a/crates/core/src/writer/json.rs
+++ b/crates/core/src/writer/json.rs
@@ -22,7 +22,7 @@ use super::utils::{
     arrow_schema_without_partitions, next_data_path, record_batch_from_message,
     record_batch_without_partitions,
 };
-use super::{DeltaWriter, DeltaWriterError};
+use super::{DeltaWriter, DeltaWriterError, WriteMode};
 use crate::errors::DeltaTableError;
 use crate::kernel::{Add, PartitionsExt, Scalar, StructType};
 use crate::table::builder::DeltaTableBuilder;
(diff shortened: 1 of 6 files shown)


michaelsauget:
Very excited to see this feature merged into the lib, this is the last blocker to adopting delta-rs for processing small to medium datasets outside of Spark! :)

crates/core/src/writer/json.rs
```diff
@@ -544,4 +556,103 @@ mod tests {
...
+            );
+
+            writer.write(vec![data]).await.unwrap();
+            let add_actions = writer.flush().await.unwrap();
+            assert_eq!(add_actions.len(), 1);
+
+            let second_data = serde_json::json!(
+                {
+                    "id" : 1,
+                    "name" : "Ion"
```
ion-elgreco, line 595: Hehe : )

ion-elgreco (approved):
Great work Tyler! :) We do likely need to port some stuff over to creates/core/src/operations/writer for it to become available to the operations directly

Issue #2025, opened 2024-01-03
Github actions fail to trigger the docs release


r3stl355:
# Environment
GitHub

# Bug

There are 2 GitHub actions which can trigger the docs deployment process, these are failing at the moment. 
- https://github.com/delta-io/delta-rs/actions/workflows/docs_release.yml
- https://github.com/delta-io/delta-rs/actions/workflows/python_release.yml

The docs deployment can also be triggered manually which is working as expected (here is example of a run: https://github.com/delta-io/delta-rs/actions/runs/7387684941)

Error is `Resource not accessible by integration`, likely related to permissions

r3stl355:
take

r3stl355:
Keeping this open until at least one successful run

PR #2026, opened 2024-01-03
fix: github actions for releasing docs

r3stl355:
# Description
Added a `write` permission to jobs triggering the docs release process. The job itself already had it but those triggering it did not, which was likely the cause for them faling

# Related Issue(s)
#2025


ion-elgreco (approved):
Thanks @r3stl355!

r3stl355:
@ion-elgreco  looks like we have some broken tests

rtyler (approved):

PR #2030, opened 2024-01-04
feat: refactor WriterProperties class

ion-elgreco:
# Description
- Add better typehint
- Better description of the compression levels
- Gracefully handle wrong compression levels, rust parquet errors were not clear enough


rtyler (approved):

python/deltalake/table.py
```diff
@@ -62,39 +62,70 @@ def __init__(
...
+            Literal[
+                "UNCOMPRESSED",
+                "SNAPPY",
+                "GZIP",
+                "BROTLI",
+                "LZ4",
+                "ZSTD",
+                "LZ4_RAW",
+            ]
+        ] = None,
```
roeap, line 75: Do not have too strong opinions about this, just thinking :) - could / should this also be an enum?

roeap (approved):

python/tests/test_writerproperties.py
```diff
@@ -0,0 +1,89 @@
...
+
+def test_write_with_writerproperties(
+    tmp_path: pathlib.Path, sample_table: pa.Table, writer_properties: WriterProperties
+):
+    write_deltalake(
+        tmp_path, sample_table, engine="rust", writer_properties=writer_properties
+    )
+
+    parquet_path = DeltaTable(tmp_path).file_uris()[0]
+    print(parquet_path)
```
roeap, line 86: debug artifact?

python/deltalake/table.py
```diff
@@ -51,6 +52,59 @@
...
+    UNCOMPRESSED = "UNCOMPRESSED"
+    SNAPPY = "SNAPPY"
+    GZIP = "GZIP"
+    BROTLI = "BROTLI"
+    LZ4 = "LZ4"
+    ZSTD = "ZSTD"
+    LZ4_RAW = "LZ4_RAW"
+
+    @classmethod
+    def __from_str__(cls, value: str) -> "Compression":
```
roeap, line 65: is this an "official" dunder method? if not, could it be just `from_str`?

python/tests/test_writerproperties.py
ion-elgreco: Ah yup, good spot

python/deltalake/table.py
ion-elgreco: It's not, will change it

roeap (approved):

PR #2033, opened 2024-01-04
fix: remove casts of structs to record batch

### LINKED ISSUE #2019: Rust panicking through Python library when a delete predicate uses a nullable field
# Environment

**Delta-rs version**: `python-v0.15.0`

**Environment**:
- **OS**: MacOS, localstack simulating S3 in Docker

***
# Bug

**What happened**:

We're using the delta-rs python library to setup the ability to delete from our existing Deltalake tables in S3. They're simple WHERE clauses like the following:

`table.delete(predicate="properties.pageUrl = 'testing.com'")`

However, I'm seeing a Rust panic:
```
thread 'tokio-runtime-worker' panicked at /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-48.0.1/src/record_batch.rs:487:9:
assertion `left == right` failed: Cannot convert nullable StructArray to RecordBatch, see StructArray documentation
  left: 1
 right: 0
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[ERROR] DeltaError: Generic error: Failed to execute write task: task 48 panicked
  File "/var/task/aws_lambda_powertools/logging/logger.py", line 449, in decorate
    return lambda_handler(event, context, *args, **kwargs)
  File "/var/task/newrelic/api/lambda_handler.py", line 135, in _nr_lambda_handler_wrapper_
    result = wrapped(*args, **kwargs)
  File "/var/task/lambda.py", line 60, in lambda_handler
    dt.delete(predicate=predicate)
  File "/var/task/deltalake/table.py", line 1107, in delete
    metrics = self._table.delete(
END RequestId: 8df76707-d13c-45b2-b112-d1801161785c
```

This is happening on a field that is nullable.. 

When a field is null / nullable, either a success or at least some specific error explaining why you can't delete on nullable fields.

**How to reproduce it**:

Setup a delta table with a nullable field, and attempt to perform a deletion on that field using deltatable's `delete` method.

**More details**:

Blajda:
# Description
Fixes an issue where the writer attempts to convert a Arrow `Struct` into a `RecordBatch`. This cannot be done since it will drop the validity array and would prevents structs with a value of `null` from being stored correctly.

This PR also extends the predicate representation for struct field access, list index access, and list range access.

# Related Issue(s)
- closes #2019


rtyler (commented):
I am not familiar enough with this code so I would like to ask @roeap to help here :cry:

liamphmurphy:
Hi all, I reported the linked issue. I pulled down the changes locally and can confirm delete's are now working for me as expected!

crates/deltalake-core/src/operations/cast.rs
```diff
@@ -8,29 +8,27 @@ use std::sync::Arc;
...
     cast_options: &CastOptions,
 ) -> Result<Vec<Arc<(dyn Array)>>, arrow_schema::ArrowError> {
     fields
         .iter()
-        .map(|f| {
-            let col = batch.column_by_name(f.name()).unwrap();
-
+        .map(|field| {
+            let col = struct_array.column_by_name(field.name()).unwrap();
             if let (DataType::Struct(_), DataType::Struct(child_fields)) =
```
ion-elgreco, line 20: There is one branch missing I guess and that's if the data type is List(Struct), so perhaps we need to add:

```rust
else if (DataType::List(_), DataType::List(to_inner_field)) { 
    match to_inner_field.data_type()  {
         Struct(child_fields) => .... do struct casting here,
         _ => ... continue normal flow
}
}

```

ion-elgreco (changes requested):
Looks good! Just left one comment on a possible branch we may need to cover

crates/deltalake-core/src/operations/cast.rs
Blajda: I think this should be a follow up PR. Primary scope for this is to remove the panic / unsoundness that occurred with null-able structs. If a `List<Struct>` is provided and LHS does not match RHS then a proper error value is returned by `cast_with_options`.

ion-elgreco: Alright let's have this in a separate PR :)

ion-elgreco (approved):

PR #2037, opened 2024-01-05
feat: arrow backed log replay and table state

roeap:
# Description

This is still very much a work in progress, opening it up for visibility and discussion.

Finally I do hope that we can make the switch to arrow based log handling. Aside from hopefully advantages in the memory footprint, I also believe it opens us up to many future optimizations as well.

To make the transition we introduce two new structs 

  - `Snapshot` - a half lazy version of the Snapshot, which only tries to get `Protocol` & `Metadata` actions ASAP. Of course these drive all our planning activities and without them there is not much we can do.
  - `EagerSnapshot` - An intermediary structure, which eagerly loads file actions and does log replay to serve as a compatibility laver for the current `DeltaTable` APIs.

One conceptually larger change is related to how we view the availability of information. Up until now `DeltaTableState` could be initialized empty, containing no useful information for any code to work with. State (snapshots) now always needs to be created valid. The thing that may not yet be initialized is the `DeltaTable`, which now only carries the table configuration and the `LogStore`. the state / snapshot is now optional. Consequently all code that works against a snapshot no longer needs to handle that matadata / schema etc may not be available.

This also has implications for the datafusion integration. We already are working against snapshots mostly, but should abolish most traits implemented for `DeltaTable` as this does not provide the information (and never has) that is al least required to execute a query.

Some larger notable changes include:

* remove `DeltaTableMetadata` and always use `Metadata` action.
* arrow and parquet are now required, as such the features got removed. Personalyl I would also argue, that if you cannot read checkpoints, you cannot read delta tables :). - so hopefully users weren't using arrow-free versions.

### Major follow-ups:

* (pre-0.17) review integration with `log_store` and `object_store`. Currently we make use mostly of `ObjectStore` inside the state handling. What we really use is `head` / `list_from` / `get` - my hope would be that we end up with a single abstraction...
* test cleanup - we are currently dealing with test flakiness and have several approaches to scaffolding tests. SInce we have the `deltalake-test` crate now, this can be reconciled.
* ...
* do more processing on borrowed data ...
* perform file-heavy operations on arrow data
* update checkpoint writing to leverage new state handling and arrow ...
* switch to exposing URL in public APIs

## Questions

* should paths be percent-encoded when written to checkpoint?

# Related Issue(s)

supersedes: #454
supersedes: #1837
closes: #1776
closes: #425 (should also be addressed in the current implementation)
closes: #288 (multi-part checkpoints are deprecated)
related: #435

# Documentation


Blajda:
@roeap This looks great! I initial had concerns about on how to update from one snapshot to another and the potential costs. Assuming checkpoints are written every 10 commits, the total of the overall `LogSegment` structure will be fairly small so the cost of cloning a segment will be relatively cheap. 

Seems like the primary trade off with this approach is a significant redaction in overall memory footprint but it comes with a potential of additional reads to the underlying store.

crates/deltalake-core/src/kernel/snapshot/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use super::{Metadata, Protocol};
+
+mod extract;
+mod log_segment;
+mod replay;
+
+/// A snapshot of a Delta table
+pub struct Snapshot {
+    log_segment: LogSegment,
```
Blajda, line 22: Will there be any benefit from wrapping this in an Arc? If this will replace the table state struct then it be cloned a bit.

roeap: We shall see 😆.

IIRC, we tried to avoid cloning the table whenever possible b/c can be quite large, but then eventually caved and made it `Clone`. A goal is to end up in a place where we can do guilt-free clones and (de-)serialization of the `DeltaTable` and / or `Snapshot`. Whether we are already getting there in this PR is another question 😆.

roeap:
> Seems like the primary trade off with this approach is a significant redaction in overall memory footprint but it comes with a potential of additional reads to the underlying store.

Exactly @Blajda. Especially for the non-query cases (vacuum etc) we may see more reads, and if not using the `EagerSnapshot` we may also do some duplicate reading. In this PR we are not yet leveraging some more optimization that can be done with regards to that.

As you mentioned, if users take care of their tables (e.g. checkpoints...) then reading from parquet is the most interesting thing. Thus, the main thing I want to do as a immediate follow up, is to introduce some optimizations for parquet reads, which I hope will come in handy for the work in #2006 - this can make reads from checkpoint files very efficient.

crates/deltalake-core/src/kernel/actions/schemas.rs
```diff
@@ -9,236 +9,241 @@ lazy_static! {
     // https://github.com/delta-io/delta/blob/master/PROTOCOL.md#change-metadata
     static ref METADATA_FIELD: StructField = StructField::new(
         "metaData",
-        DataType::Struct(Box::new(StructType::new(vec![
```
rtyler, line 12: These changes look like they include nullable column changes, which will require a little finer review than I have time for today before the merge. I don't see any big problems, but I do see some differences that I need to noodle on :thinking:

crates/deltalake-core/src/kernel/actions/types.rs
```diff
@@ -76,18 +76,16 @@ pub struct Metadata {
...
-        configuration: Option<HashMap<String, Option<String>>>,
+        configuration: HashMap<String, Option<String>>,
     ) -> Self {
         Self {
-            id: id.into(),
-            format,
-            schema_string: schema_string.into(),
+            id: uuid::Uuid::new_v4().to_string(),
+            format: Default::default(),
+            schema_string: serde_json::to_string(&schema).unwrap(),
```
rtyler, line 86: It gives me the willies to see an unwrap in the constructor of `Metadata`, but since  this is accepting a `StructType`, is there any possible way for this to panic?

crates/deltalake-core/src/kernel/snapshot/extract.rs
```diff
@@ -0,0 +1,177 @@
...
+                    //         "Map {} has {} columns, expected 2",
+                    //         path_step,
+                    //         maparr.entries().num_columns()
+                    //     )));
+                    // }
+                    // if next_path_step == *maparr.entries().column_names().first().unwrap() {
+                    //     Ok(maparr.entries().column(0))
+                    // } else {
+                    //     Ok(maparr.entries().column(1))
+                    // }
```
rtyler, line 82: Any reason to keep this code around? I'm fine deleting if we're not using :smile:

crates/deltalake-core/src/kernel/snapshot/extract.rs
```diff
@@ -0,0 +1,177 @@
...
+
+use arrow_array::{
+    Array, ArrowNativeTypeOp, ArrowNumericType, BooleanArray, ListArray, MapArray, PrimitiveArray,
+    RecordBatch, StringArray, StructArray,
+};
+use arrow_schema::{ArrowError, DataType};
+
+use crate::{DeltaResult, DeltaTableError};
+
+pub(crate) trait ProvidesColumnByName {
```
rtyler, line 11: This file would benefit from a lot of Rustdoc, I can follow it right now because I have a lot of this state in my head, but I guarantee I'll forget why this all exists in 2 months :laughing:

crates/deltalake-core/src/kernel/snapshot/mod.rs
```diff
@@ -0,0 +1,669 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct Snapshot {
+    log_segment: LogSegment,
+    config: DeltaTableConfig,
+    protocol: Protocol,
+    metadata: Metadata,
+    schema: StructType,
+    // TODO make this an URL
+    /// path of the table root within the object store
+    table_url: String,
```
rtyler, line 39: :pray:

rtyler (commented):
I understand the motivation to land this as soon as possible, this is going to get lots of conflicts if we don't. 

I am generally fine with the direction, the commitment to arrow here, and want to get this in as soon as possible. We should target the 0.17 release since that's already going to have a lot of fun stuff in it :laughing:

crates/deltalake-core/src/kernel/snapshot/log_segment.rs
```diff
@@ -0,0 +1,665 @@
...
+    e_tag: Option<String>,
+    version: Option<i64>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(super) struct LogSegment {
+    pub(super) version: i64,
+    pub(super) commit_files: VecDeque<ObjectMeta>,
+    pub(super) checkpoint_files: Vec<ObjectMeta>,
+}
```
Blajda, line 107: I've been trying to wrap my head around this and the benefit obtained from storing the ObjectMeta for commit files.
Suppose we want to load a table at version 60. If there is no checkpoint then we can infer &check the json files from 0 to 60 exist and then in the LogSegment store this range.
If the table had a checkpoint at version 50 then we can store & check the files files from 51 to 60 and store that range. 
From what I've seen we can remove `commit_files` and replace it with `commit_json_start` which is the first json file that must be read after the snapshot. If the snapshot matches the version then it can be None.

crates/deltalake-core/src/kernel/actions/schemas.rs
roeap: they do - this is mainly due to the fact that these schemas represent what we read from checkpoint / log files, and as such the nullability of the column in the whole record-batch. Thus here almost everything is nullable.

The actual nullability of the data structures is only enforced once we parse that data. there we then error for missing required fields.

crates/deltalake-core/src/kernel/actions/types.rs
roeap: rightfully so :) - this is removed locally now, and will be fixed in the next push...

crates/deltalake-core/src/kernel/snapshot/log_segment.rs
roeap: This gets a bit more complex, as we need to know which checkpoint files to read. I.e. while deprecated we should still be able to read multi-part checkpoints. With V2 checkpoints (not yet supported) we also need to be able to read the sidecar files.

More broadly speaking, for parquet files we always require the file size to read it, so we avoid an extra `head` if we store the `ObjectMeta` from list. 

I do see the point though that we may want to see if we can avoid storing some of these files in the final snapshot.

crates/deltalake-core/src/kernel/snapshot/extract.rs
roeap: I guess this is true for most (all) added files. I will see that we have some comments for us.

Once larger goal is to only expose the `*Snapshot` to the outside world, and leave how we resolve all of this internal to the core crate.

crates/deltalake-core/src/kernel/snapshot/log_segment.rs
roeap: Wrinting some comments I realized we also do this to support time travel right now. There we rely on the modification time of the log files to identify the right version number.

crates/deltalake-core/src/kernel/actions/schemas.rs
```diff
@@ -9,236 +9,241 @@ lazy_static! {
...
+                MapType::new(
+                    DataType::STRING,
+                    DataType::STRING,
                     true,
-                ))),
+                ),
                 true,
             ),
             StructField::new(
                 "operationMetrics",
```
ion-elgreco, line 90: This wouldn't be valid according to the protocol and our current operations since we also write operationMetrics as   MapType::new(
                    DataType::STRING,
                    DataType::INTEGER)

ion-elgreco:
@roeap I think you can fix the python errors if you change `use_compliant_nested_type=True` for the pyarrow writer.

crates/deltalake-core/src/kernel/arrow/mod.rs
```diff
@@ -8,12 +8,25 @@ use arrow_schema::{
...
+};
 
 pub mod schemas;
 
-const MAP_KEYS_NAME: &str = "keys";
-const MAP_VALUES_NAME: &str = "values";
+const MAP_ROOT_DEFAULT: &str = "entries";
+const MAP_KEY_DEFAULT: &str = "keys";
+const MAP_VALUE_DEFAULT: &str = "values";
+const LIST_ROOT_DEFAULT: &str = "element";
```
ion-elgreco, line 21: Will this be backwards compatible? Old writers wrote `item` instead of element, this may become problematic when a table schema is constructed from the StructType and the ArrowSchema is actually the same except for the nested field name.

crates/deltalake-core/src/kernel/arrow/mod.rs
```diff
@@ -219,12 +210,12 @@ impl TryFrom<&ArrowDataType> for DataType {
...
-            ))),
+            ArrowDataType::Decimal128(p, s) => {
+                Ok(DataType::Primitive(PrimitiveType::Decimal(*p, *s)))
+            }
+            ArrowDataType::Decimal256(p, s) => {
+                Ok(DataType::Primitive(PrimitiveType::Decimal(*p, *s)))
+            }
             ArrowDataType::Date32 => Ok(DataType::Primitive(PrimitiveType::Date)),
             ArrowDataType::Date64 => Ok(DataType::Primitive(PrimitiveType::Date)),
             ArrowDataType::Timestamp(TimeUnit::Microsecond, None) => {
```
ion-elgreco, line 221: Shound't this return a TimestampNtz instead since no timezone is part of the arrow datatype.

crates/deltalake-core/src/kernel/schema.rs
```diff
@@ -521,65 +566,40 @@ pub enum DataType {
...
+    pub const LONG: Self = DataType::Primitive(PrimitiveType::Long);
+    pub const INTEGER: Self = DataType::Primitive(PrimitiveType::Integer);
+    pub const SHORT: Self = DataType::Primitive(PrimitiveType::Short);
+    pub const BYTE: Self = DataType::Primitive(PrimitiveType::Byte);
+    pub const FLOAT: Self = DataType::Primitive(PrimitiveType::Float);
+    pub const DOUBLE: Self = DataType::Primitive(PrimitiveType::Double);
+    pub const BOOLEAN: Self = DataType::Primitive(PrimitiveType::Boolean);
+    pub const BINARY: Self = DataType::Primitive(PrimitiveType::Binary);
+    pub const DATE: Self = DataType::Primitive(PrimitiveType::Date);
+    pub const TIMESTAMP: Self = DataType::Primitive(PrimitiveType::Timestamp);
```
ion-elgreco, line 599: TIMESTAMPNTZ? :)

crates/deltalake-core/src/kernel/snapshot/extract.rs
```diff
@@ -0,0 +1,167 @@
...
+                    extract_column(
+                        cast_column_as::<StructArray>(next_path_step, &Some(listarr.values()))?,
+                        next_path,
+                        remaining_path_steps,
+                    )
+                } else {
+                    Ok(listarr.values())
+                }
+            }
+            _ => extract_column(
```
ion-elgreco, line 106: Won't this fail when you have a field that is called random.field and the the actual datatype is a non-nested dtype? Maybe we should the if `dataype.is_nested()` first

crates/deltalake-core/src/table/config.rs
```diff
@@ -302,6 +304,20 @@ impl<'a> TableConfig<'a> {
             .and_then(|o| o.as_ref().and_then(|v| v.parse().ok()))
             .unwrap_or_default()
     }
+
+    /// Return the check constraints on the current table
+    pub fn get_constraints(&self) -> Vec<Constraint> {
```
ion-elgreco, line 309: Should there also be `get_invariants` on the TableConfig?

crates/deltalake-core/src/kernel/arrow/mod.rs
```diff
@@ -142,24 +150,7 @@ impl TryFrom<&DataType> for ArrowDataType {
                     .into(),
             )),
             DataType::Array(a) => Ok(ArrowDataType::List(Arc::new(a.as_ref().try_into()?))),
```
ion-elgreco, line 152: This one should actually be 'UTC' timezone. I did not notice before the error of our totally not supporting writing UTC Tz issues was coming from this line :P

PrimitiveType::Timestamp => {
                        // Issue: https://github.com/delta-io/delta/issues/643
                        Ok(ArrowDataType::Timestamp(TimeUnit::Microsecond, None))
                    }

crates/deltalake-core/tests/integration.rs
```diff
@@ -61,23 +61,34 @@ async fn test_action_reconciliation() {
...
+            .snapshot()
+            .unwrap()
+            .all_tombstones(table.object_store().clone())
+            .await
+            .unwrap()
+            .map(|r| r.path.clone())
             .collect::<Vec<_>>(),
-        vec![a.path.as_str()]
+        vec![a.path.clone()]
     );
```
roeap, line 75: While this part of the test does not pass anymore, i feel this is the desired behavior.

What this tests, if we can add an already deleted file back into the table, using the same path etc. i.e. un-doing a remove. The thing to do would be to time travel, or add the same data under a new file name...

crates/deltalake-core/src/kernel/arrow/mod.rs
roeap: in the sense that all our tests are passing, yes. while the written parquet files might look different , but this is already the case as different writers handle this differently. (e.g. spark and arrow).

For this PR this change also is not too important, as we do not yet try to enforce this schema speciufically on parquet reads. I have an improved version for parquet reading in preparation where we can go deeper into the discussion...

crates/deltalake-core/src/kernel/arrow/mod.rs
roeap: yes, this is the current behavior. since we had so much fun with timestamps I did not want to break this as well. In any case something we can hopefully fix before `0.,17`.

crates/deltalake-core/src/kernel/arrow/mod.rs
roeap: this is the current behavior, but in principle yes. of course this would have to be guarded by the `TimestampNtz` feature ...

crates/deltalake-core/src/kernel/snapshot/extract.rs
roeap: `.` is only allowed in column names if collumn mapping is enabled. Since we do not yet support that feature, I wanted to leave this for a follow up ...

crates/deltalake-core/src/kernel/arrow/mod.rs
ion-elgreco: Alright sounds good!

crates/deltalake-core/src/kernel/snapshot/serde.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("struct EagerSnapshot")
+    }
+
+    fn visit_seq<V>(self, mut seq: V) -> Result<EagerSnapshot, V::Error>
+    where
+        V: SeqAccess<'de>,
+    {
+        println!("eager: {:?}", "start");
```
Blajda, line 156: Forgot to remove a print :)

crates/deltalake-core/src/writer/json.rs
```diff
@@ -46,7 +45,7 @@ pub(crate) struct DataArrowWriter {
     writer_properties: WriterProperties,
     buffer: ShareableBuffer,
     arrow_writer: ArrowWriter<ShareableBuffer>,
-    partition_values: HashMap<String, Option<String>>,
+    partition_values: BTreeMap<String, Scalar>,
```
Blajda, line 48: Why the change from a HashMap to BTree?

Blajda (commented):
Overall a very high quality PR and the changes made to operations makes sense to me.

crates/deltalake-core/src/writer/json.rs
roeap: For the partition values, ordering matters in several places. Previously we were using a combination of the 'partition_values' on metadata and the partition values HashMap. With this change we can rely on the insertion order and no longer need to track the values vector.

rtyler (approved):

rtyler:
I am going to merge this because we need it merged, main is already broken :unamused:  So the repair test will need to get fixed shortly

PR #2041, opened 2024-01-06
chore: update documentation for S3 / DynamoDb log store configuration

dispanser:
# Description

My attempt at rewording the documentation to address the changes related to `S3DynamoDbLogStore` and its configuration.

Not sure how our documentation works in the first place, and I'm not a word smith either, but at least I hope this makes things factually correct.

### DIFF
--- a/python/docs/source/usage.rst
+++ b/python/docs/source/usage.rst
@@ -492,30 +492,39 @@ locking provider at the moment in delta-rs. To enable DynamoDB as the
 locking provider, you need to set the **AWS_S3_LOCKING_PROVIDER** to 'dynamodb'
 as a ``storage_options`` or as an environment variable.
 
-Additionally, you must create a DynamoDB table with the name ``delta_rs_lock_table``
+Additionally, you must create a DynamoDB table with the name ``delta_log``
 so that it can be automatically recognized by delta-rs. Alternatively, you can
-use a table name of your choice, but you must set the **DYNAMO_LOCK_TABLE_NAME**
+use a table name of your choice, but you must set the **DELTA_DYNAMO_TABLE_NAME**
 variable to match your chosen table name. The required schema for the DynamoDB
 table is as follows:
 
 .. code-block:: json
 
 
-       {
+
+        "Table": {
             "AttributeDefinitions": [
                 {
-                    "AttributeName": "key",
+                    "AttributeName": "fileName",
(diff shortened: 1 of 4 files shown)


rtyler:
Test failure looks unrelated, we're seeing that on main too :frowning:

python/docs/source/usage.rst
```diff
@@ -492,47 +492,77 @@ locking provider at the moment in delta-rs. To enable DynamoDB as the
 locking provider, you need to set the **AWS_S3_LOCKING_PROVIDER** to 'dynamodb'
 as a ``storage_options`` or as an environment variable.
 
-Additionally, you must create a DynamoDB table with the name ``delta_rs_lock_table``
+Additionally, you must create a DynamoDB table with the name ``delta_log``
 so that it can be automatically recognized by delta-rs. Alternatively, you can
-use a table name of your choice, but you must set the **DYNAMO_LOCK_TABLE_NAME**
+use a table name of your choice, but you must set the **DELTA_DYNAMO_TABLE_NAME**
```
rtyler, line 497: I don't recall from the pull request reviews previously, but why did this name change @dispanser ?

dispanser: In the initial implementation path, both locking solutions where applied at the same time, so the table name had to be different (due to incompatible schemas). After throwing out the old locking logic, I simply did not think about changing this.
Given that a user switching to the new library has to tear down the old lock table and set up the new one anyways, it may not be breaking any more than the logic change already does, and I do think having `DELTA` in there is actually better.

I now realize that it's not very fortunate, but with the release already made (?), changing back might make more harm than good.

rtyler:
@dispanser I can buy that argument, are there other environment variable names which might be helpful to rename to help users recognize the necessary transition?

ion-elgreco:
@dispanser @rtyler it seems the wrong docs were updated.. I was trying to see where they were but I noticed the previous sphinx docs were edited instead of the new mkdocs markdowns, I will bring it back in to mkdcs and also remove these old docs so it can't happen again :)

dispanser:
Aw, I'm so sorry about that. I basically searched our code base for the relevant pieces (mostly env var names for the configuration). I should have asked.

ion-elgreco:
@dispanser no worries! :) It's confusing there are two docs in there still

PR #2042, opened 2024-01-06
docs: update docs for merge

### LINKED ISSUE #850: Implement merge command

Blajda:
# Description
Update documentation to reflect the current state of merge. Since merge now supports upserts without performing a full rewrite I'd let to mark it as "done".

There is of course further optimizations that can be performed but it is now in a usable state.

# Related Issue(s)
- closes #850

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ of features outlined in the Delta [protocol][protocol] is also [tracked](#protoc
 | Delete - predicates   |         ![done]          |         ![done]          | Delete data based on a predicate            |
 | Optimize - compaction |         ![done]          |         ![done]          | Harmonize the size of data file             |
 | Optimize - Z-order    |         ![done]          |         ![done]          | Place similar data into the same file       |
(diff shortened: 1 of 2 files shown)


ion-elgreco (approved):
🚀🚀

PR #2048, opened 2024-01-07
feat: cdf reader for delta tables

hntd187:
# Description
This PR is the initial work for Change Data Feed (CDF) readers for delta tables. This PR looks a lot larger than it really is because a physical test table is checked in with this which will be removed once the loop is closed on CDF reading/writing. 

# Related Issue(s)

# Documentation
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#change-data-files
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-cdc-file


crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,318 @@
...
+                let remove_actions = version_actions
+                    .iter()
+                    .filter_map(|a| match a {
+                        Action::Remove(r) if r.data_change => Some(r.clone()),
+                        _ => None,
+                    })
+                    .collect::<Vec<Remove>>();
+
+                if !add_actions.is_empty() {
+                    add_files.push(CdcDataSpec {
```
ion-elgreco, line 134: Where is this CdcDataSpec coming from?

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        let this_partition_values = self
+            .partition_values
+            .iter()
+            .map(|name| self.schema.field_with_name(name).map(|f| f.to_owned()))
+            .collect::<Result<Vec<_>, ArrowError>>()?;
+
+        // Setup for the Read Schemas of each kind of file, CDC files include commit action type so they need a slightly
+        // different schema than standard add file reads
+        let cdc_file_schema = create_cdc_schema(schema_fields.clone(), true);
```
ion-elgreco, line 181: Here you add CHANGE_TYPE_COL but in CDC_PARTITION_SCHEMA it's not there? I am a bit confused about this

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,318 @@
...
+        let cdc_file_groups =
+            create_partition_values(self.schema.clone(), cdc, &self.partition_values, None)?;
+        let add_file_groups = create_partition_values(
+            self.schema.clone(),
+            add,
+            &self.partition_values,
+            Self::get_add_action_type(),
+        )?;
+
+        // Create the parquet scans for each associated type of file. I am not sure when we would use removes yet, but
```
ion-elgreco, line 203: Aren't remove actions then read and appended to the result with `_change_type: delete`?

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,318 @@
...
+        // tests self contained and not rely on a physical table with them
+        let _table = DeltaOps::try_from_uri("./tests/data/cdf-table").await?;
+        let schema = _table.0.state.arrow_schema()?;
+        let partition_cols = _table.0.metadata()?.clone().partition_columns.clone();
+
+        let scan = DeltaCdfScan::new(_table.0.log_store.clone(), 0, schema, partition_cols);
+
+        let results = scan.scan().await?;
+        let data: Vec<RecordBatch> = collect_sendable_stream(results).await?;
+        print_batches(&data)?;
```
ion-elgreco, line 314: Can you add an assert of the output of data?

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,318 @@
...
+        let add_file_groups = create_partition_values(
+            self.schema.clone(),
+            add,
+            &self.partition_values,
+            Self::get_add_action_type(),
+        )?;
+
+        // Create the parquet scans for each associated type of file. I am not sure when we would use removes yet, but
+        // they would be here if / when they are necessary
+        let cdc_scan = ParquetFormat::new()
```
ion-elgreco, line 205: Have you considered ctx.read_parquet?

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
hntd187: It's in the mod.rs for the module

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
hntd187: because in cdc partitions those files already have the change type where as adds don't have them so the schema needs to account for this.

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
hntd187: I'll flesh tests out to be more complete when we are more complete in the initial impl

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
hntd187: You lose some control when you use that facility and I'm not even sure that this would work given it looks like it's more expecting a directory and not explicit file paths. It might work, I haven't explored it extensively to be honest, but at a glance it doesn't look like it would provide the control necessary.

ion-elgreco: It does accept a list of filepaths. Comparing the two, it looks like these params are the same (I think)

ParquetReadOptions.schema  == FileScanConfig.file_schema
table_paths == FileScanConfig.file_groups
ParquetReadOptions.table_partition_cols == FileScanConfig.table_partition_cols 

Project and limit I guess you do after read_parquet on the dataframe (I assume it's lazy).

The only gap is the statistics part, but you don't seem to use it.  Could be perhaps interesting to check if there is any performance difference between the two?

crates/deltalake-core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+pub struct DeltaCdfScan {
+    log_store: LogStoreRef,
+    starting_version: i64,
+    schema: SchemaRef,
+    partition_values: Vec<String>,
+}
+
+impl DeltaCdfScan {
```
roeap, line 54: Did not have time for a proper review, just as a quick comment - we should implement `ExecutionPlan` for this scan, so it can be used in datafusion as an operation.

crates/core/src/operations/load_cdf.rs
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    #[inline]
+    fn get_add_action_type() -> Option<ScalarValue> {
+        Some(ScalarValue::Utf8(Some(String::from("insert"))))
+    }
+
+    /// Executes the scan
+    pub async fn build(&self) -> DeltaResult<DeltaCdfScan> {
+        let (cdc, add, _remove) = self.determine_files_to_read().await?;
```
ion-elgreco, line 217: Why is nothing done with the remove actions?

crates/core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,93 @@
...
+pub const COMMIT_VERSION_COL: &str = "_commit_version";
+/// Commit Timestamp column name
+pub const COMMIT_TIMESTAMP_COL: &str = "_commit_timestamp";
+
+lazy_static! {
+    pub static ref CDC_PARTITION_SCHEMA: Vec<Field> = vec![
+        Field::new(COMMIT_VERSION_COL, DataType::Int64, true),
+        Field::new(
+            COMMIT_TIMESTAMP_COL,
+            DataType::Timestamp(TimeUnit::Millisecond, None),
```
ion-elgreco, line 23: Should be UTC since it's a delta timestamp

crates/core/src/operations/load_cdf.rs
```diff
@@ -0,0 +1,471 @@
...
+            }
+
+            for action in &version_actions {
+                match action {
+                    Action::Cdc(f) => cdc_actions.push(f.clone()),
+                    Action::Metadata(md) => {
+                        log::info!("Metadata: {:?}", &md);
+                        if let Some(Some(key)) = &md.configuration.get("delta.enableChangeDataFeed")
+                        {
+                            if key.to_lowercase() == "false" {
```
ion-elgreco, line 147: Shouldn't you also check the protocol versions here in case its at v3,7

hntd187: I'm not entirely sure there. The scala impl doesn't it merely checks for the table feature, so I'm assuming that since it's above the table features needed version it just checks that.

crates/core/src/delta_datafusion/cdf/scan.rs
hntd187: Oddly Spark does not write the CDF timestamp with any timezone, so this doesn't actually work

crates/core/src/operations/load_cdf.rs
ion-elgreco: That's weird. Then they likely set the tableConfig while you enable the feature : S

crates/core/src/operations/load_cdf.rs
hntd187: It's a left over from when I thought I'd have to do coarse grained cdc with this, but to my knowledge without coarse grained removes are not used. I just wanted to leave it for now incase I misunderstood this, but I'll remove it if it holds to not be needed.

crates/core/src/delta_datafusion/cdf/scan.rs
ion-elgreco: Isn't that then a bug. They mention timestamp in the docs, and protocol states a timestamp is an UTC datetime

ion-elgreco:
Probably best to also just add the pyo3 binding as well in this PR, we could either add a static method on DeltaTable called `read_change_feed` or have a function similar to `write_deltalake`, called `read_delta_cdf`.

hntd187:
Good idea @ion-elgreco lemme start doing that

crates/test/tests/data/cdf-table-non-partitioned/_delta_log/00000000000000000004.json
ion-elgreco, line 1: To maybe increase the coverage, also have a table with a merge operation that has done delete, insert and update?

crates/core/src/operations/load_cdf.rs
```diff
@@ -0,0 +1,471 @@
...
+                    }
+                }
+            }
+
+            for action in &version_actions {
+                match action {
+                    Action::Cdc(f) => cdc_actions.push(f.clone()),
+                    Action::Metadata(md) => {
+                        log::info!("Metadata: {:?}", &md);
+                        if let Some(Some(key)) = &md.configuration.get("delta.enableChangeDataFeed")
```
nicklan, line 145: Drive by comment, but the spark version seems to check that the starting version has this enabled, and then comments [here](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/commands/cdc/CDCReader.scala#L524C1-L526C90) that the equivalent of this check is enough to ensure the total range has it enabled.

I don't see a check for the starting version anywhere in this PR though. Do you need to add one?

python/src/lib.rs
```diff
@@ -520,6 +525,67 @@ impl RawDeltaTable {
...
+                .into_iter()
+                .flatten()
+                .collect::<Result<Vec<Vec<_>>, _>>()
+                .unwrap()
+                .into_iter()
+                .flatten()
+                .collect();
+
+            let final_batch = concat_batches(&batches[0].schema(), &batches).unwrap();
+            Ok(PyArrowType(final_batch))
```
ion-elgreco, line 585: I think we can also convert it as a reader again

ion-elgreco (approved):

ion-elgreco (changes requested):
I send you some comments on slack on the python signature.

 Regarding the starting_version it always seems to be required but in spark sql it can also be just timestamps. I think it would make sense to have the starting version optional as well, and just raise if starting version nor starting timestamp is provided

python/src/lib.rs
hntd187: I'm sorry, what do you mean?

hntd187:
Well, the rust side of it defaults to 0 for the starting version if you don't provide one, so in theory you can give nothing and it's just the full CDF feed for the table since the beginning. I don't know why I required python here to provide a starting version, but you are right we can mark it as optional.

python/src/lib.rs
ion-elgreco: impl IntoPyArrow for ArrowArrayStreamReader
Convert a ArrowArrayStreamReader into a pyarrow.RecordBatchReader.

https://docs.rs/arrow/latest/arrow/ffi_stream/struct.ArrowArrayStreamReader.html

I mean the above. You can take an arrowArrayStreamReader and move it into python as pyarrow.RecordBatchReader

crates/core/src/delta_datafusion/cdf/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn partition_values(&self) -> HashMap<String, Option<String>>;
+    /// Physical Path to the data
+    fn path(&self) -> String;
+    /// Byte size of the physical file
+    fn size(&self) -> usize;
+}
+
+impl FileAction for Add {
+    fn partition_values(&self) -> HashMap<String, Option<String>> {
+        self.partition_values.clone()
```
nicklan, line 42: seems like you only use this to `get` out of the map, so you could probably return `&HashMap<String, Option<String>>` and not have to clone (or just provide a `get_partition_value` which does the probe internally (which makes it easier to not have to clone for remove).

crates/core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn output_partitioning(&self) -> Partitioning {
+        self.plan.output_partitioning()
+    }
+
+    fn output_ordering(&self) -> Option<&[PhysicalSortExpr]> {
+        self.plan.output_ordering()
+    }
+
+    fn children(&self) -> Vec<Arc<dyn ExecutionPlan>> {
+        vec![self.plan.clone()]
```
nicklan, line 76: this is probably my own ignorance regarding all these apis, but all the other things (like `output_ordering` etc) seem to delegate to `self.plan`, but this one says that `plan` is the child. In particular, if this returns `plan` I would have expected `with_new_children` to _replace_ `plan`, but it instead makes the new children children of `plan`.

Anyway, just something to think of :)

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -0,0 +1,99 @@
...
+                    .unwrap_or(Some(ScalarValue::Null))
+                    .unwrap_or(ScalarValue::Null),
+                    None => get_null_of_arrow_type(field.data_type()).unwrap_or(ScalarValue::Null),
+                })
+                .unwrap_or(ScalarValue::Null)
+        })
+        .unwrap_or(ScalarValue::Null)
+}
+
+pub fn create_spec_partition_values<F: FileAction>(
```
nicklan, line 42: should have some docs for a `pub` func

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -0,0 +1,99 @@
...
+use object_store::path::Path;
+use object_store::ObjectMeta;
+use serde_json::Value;
+
+use crate::delta_datafusion::cdf::scan::CHANGE_TYPE_COL;
+use crate::delta_datafusion::cdf::{CdcDataSpec, FileAction};
+use crate::delta_datafusion::{get_null_of_arrow_type, to_correct_scalar_value};
+use crate::DeltaResult;
+
+pub fn map_action_to_scalar<F: FileAction>(
```
nicklan, line 17: docs, as it's pub

nicklan (commented):
couple more comments

crates/core/src/delta_datafusion/cdf/scan.rs
hntd187: You are right here, since this is a scan, it should be an empty vec

crates/core/src/delta_datafusion/cdf/scan_utils.rs
hntd187: These are pub in their module but are actually only exported as `pub(crate)` outside the module. There actually is a deny lint set for publicly reachable things that don't have docs

python/src/lib.rs
ion-elgreco: Create a RecordBatchReader first from the `batches` then the stream

```
let stream = FFI_ArrowArrayStream::new(reader)
Ok(PyArrowType(stream))
```

With this as the return type: PyResult<PyArrowType<ArrowArrayStreamReader>>

ion-elgreco (dismissed):

ion-elgreco (approved):

ion-elgreco (approved):

crates/core/src/delta_datafusion/cdf/scan.rs
```diff
@@ -0,0 +1,59 @@
...
+        self.plan.properties()
+    }
+
+    fn children(&self) -> Vec<Arc<dyn ExecutionPlan>> {
+        vec![]
+    }
+
+    fn with_new_children(
+        self: Arc<Self>,
+        _children: Vec<Arc<dyn ExecutionPlan>>,
```
nicklan, line 47: nit: you do use this, so why `_children` and not `children`?

nicklan (approved):
nice, lgtm!

crates/core/src/delta_datafusion/cdf/scan.rs
hntd187: A side effect of me having this stubbed to start as `todo!()` and then not fixing it properly.

PR #2059, opened 2024-01-08
fix: allow checkpoints to contain metadata actions without a createdTime value

rtyler:
From the protocol, createdTime is of type Option[Long] and:

    The time when this metadata action is created, in milliseconds since the Unix epoch

Not required, but this only shows up when checkpoints are created and then read back into on the table load


roeap:
The affected code is pending removal in #2037 there the created time field is treated as optional - or rather I'll make sure :).

roeap (approved):
LGTM 👍

Issue #2070, opened 2024-01-13
Add `drop constraint`


ion-elgreco:
**Use Case**
We can add them but we should also be able to drop them : )

PR #2071, opened 2024-01-13
feat(rust, python): add `drop constraint` operation

ion-elgreco:
# Description
Adds `drop constraint` as an operation, and also exposed it to python under the alter namespace 😄 

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2070

### DIFF
--- a/crates/core/src/operations/drop_constraints.rs
+++ b/crates/core/src/operations/drop_constraints.rs
@@ -1,0 +1,256 @@
+//! Add a check constraint to a table
+
+use std::collections::HashMap;
+
+use chrono::Utc;
+use futures::future::BoxFuture;
+use serde_json::json;
+
+use crate::kernel::{Action, CommitInfo, IsolationLevel, Protocol};
+use crate::logstore::LogStoreRef;
+use crate::operations::transaction::commit;
+use crate::protocol::DeltaOperation;
+use crate::table::state::DeltaTableState;
+use crate::DeltaTable;
+use crate::{DeltaResult, DeltaTableError};
+
+/// Remove constraints from the table
+pub struct DropConstraintBuilder {
+    /// A snapshot of the table's state
+    snapshot: DeltaTableState,
+    /// Name of the constraint
+    name: Option<String>,
+    /// Raise if constraint doesn't exist
+    raise_if_not_exists: bool,
+    /// Delta object store for handling data files
+    log_store: LogStoreRef,
+    /// Additional metadata to be added to commit
+    app_metadata: Option<HashMap<String, serde_json::Value>>,
+}
+
+impl DropConstraintBuilder {
+    /// Create a new builder
+    pub fn new(log_store: LogStoreRef, snapshot: DeltaTableState) -> Self {
+        Self {
+            name: None,
+            raise_if_not_exists: true,
+            snapshot,
+            log_store,
+            app_metadata: None,
+        }
+    }
+
+    /// Specify the constraint to be removed
+    pub fn with_constraint<S: Into<String>>(mut self, name: S) -> Self {
+        self.name = Some(name.into());
+        self
+    }
+
+    /// Specify if you want to raise if the constraint does not exist
+    pub fn with_raise_if_not_exists(mut self, raise: bool) -> Self {
+        self.raise_if_not_exists = raise;
+        self
+    }
+
+    /// Additional metadata to be added to commit info
+    pub fn with_metadata(
+        mut self,
+        metadata: impl IntoIterator<Item = (String, serde_json::Value)>,
+    ) -> Self {
+        self.app_metadata = Some(HashMap::from_iter(metadata));
+        self
+    }
+}
+
+impl std::future::IntoFuture for DropConstraintBuilder {
+    type Output = DeltaResult<DeltaTable>;
+
+    type IntoFuture = BoxFuture<'static, Self::Output>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        let mut this = self;
+
+        Box::pin(async move {
+            let name = match this.name {
+                Some(v) => v,
+                None => return Err(DeltaTableError::Generic("No name provided".to_string())),
+            };
+
+            let mut metadata = this.snapshot.metadata()?.clone();
+            let configuration_key = format!("delta.constraints.{}", name);
+
+            let found_constraint = metadata.configuration.contains_key(&configuration_key);
+            if this.raise_if_not_exists && !found_constraint {
+                return Err(DeltaTableError::Generic(format!(
+                    "Constraint with name: {} doesn't exists",
+                    name
+                )));
+            } else if !this.raise_if_not_exists && !found_constraint {
+                return Ok(DeltaTable::new_with_state(this.log_store, this.snapshot));
+            }
+
+            metadata
+                .configuration
+                .remove(format!("delta.constraints.{}", name).as_str());
+
+            let mut contains_constraints: bool = false;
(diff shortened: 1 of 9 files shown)


ion-elgreco:
@hntd187 do you want to review this one?

crates/deltalake-core/src/operations/drop_constraints.rs
```diff
@@ -0,0 +1,247 @@
...
+                )));
+            } else if !this.raise_if_not_exists && !found_constraint {
+                return Ok(DeltaTable::new_with_state(this.log_store, this.snapshot));
+            }
+
+            metadata
+                .configuration
+                .remove(format!("delta.constraints.{}", name).as_str());
+
+            let mut contains_constraints: bool = false;
```
hntd187, line 96: you could instead write this as `let contains_constraints = metadata.configuration.keys().any(|k| k.starts_with("delta.constraints"));`

crates/deltalake-core/src/operations/drop_constraints.rs
```diff
@@ -0,0 +1,247 @@
...
+                } else {
+                    continue;
+                }
+            }
+
+            let operational_parameters = HashMap::from_iter([("name".to_string(), json!(&name))]);
+
+            let operations = DeltaOperation::DropConstraint { name: name.clone() };
+
+            let app_metadata = match this.app_metadata {
```
hntd187, line 110: `.unwrap_or_default()` instead

crates/deltalake-core/src/operations/drop_constraints.rs
```diff
@@ -0,0 +1,247 @@
...
+                operation: Some(operations.name().to_string()),
+                operation_parameters: Some(operational_parameters),
+                read_version: Some(this.snapshot.version()),
+                isolation_level: Some(IsolationLevel::Serializable),
+                is_blind_append: Some(false),
+                info: app_metadata,
+                ..Default::default()
+            };
+
+            let protocol = if !contains_constraints {
```
hntd187, line 126: Do you know if other implementations downgrade the protocol as a result of dropping the constraint? I can check this for you, but I don't think the protocol says anything about whether or not it's required to do this.

crates/deltalake-core/src/operations/drop_constraints.rs
```diff
@@ -0,0 +1,247 @@
...
+                    reader_features: old_protocol.reader_features.clone(),
+                    writer_features: old_protocol.writer_features.clone(),
+                })
+            } else {
+                None
+            };
+
+            let mut actions = vec![Action::CommitInfo(commit_info), Action::Metadata(metadata)];
+
+            protocol.map(|protocol| actions.push(Action::Protocol(protocol)));
```
hntd187, line 149: while it's more code a `if let Some(protocol) = protocol {` makes it clearer on what you are intending here.

hntd187 (changes requested):

crates/deltalake-core/src/operations/drop_constraints.rs
ion-elgreco: I just assumed here it's safe to do since the protocol version was only for constraints, and if there are no constraints imho it should not be left there. 

You could check what spark-delta here does :) But it should be fine I think, I don't believe it violates the protocol

crates/deltalake-core/src/operations/drop_constraints.rs
ion-elgreco: Ah of course there is something for that :P Thanks!

crates/deltalake-core/src/operations/drop_constraints.rs
ion-elgreco: Ah didn't know this method existed.

crates/deltalake-core/src/operations/drop_constraints.rs
ion-elgreco: Discussed with @roeap on slack and it wouldn't be safe to do this, so I've removed it in ths operation, but we may look at adding an explicit operation to downgrade

crates/core/src/operations/drop_constraints.rs
```diff
@@ -0,0 +1,216 @@
...
+                    "Constraint with name: {} doesn't exists",
+                    name
+                )));
+            } else if !this.raise_if_not_exists && !found_constraint {
+                return Ok(DeltaTable::new_with_state(this.log_store, this.snapshot));
+            }
+
+            metadata
+                .configuration
+                .remove(format!("delta.constraints.{}", name).as_str());
```
roeap, line 94: I think we can simplify this a bit by leveraging that remove will return the removed value if the key existed.

 ```suggestion
            if metadata.configuration.remove(&configuration_key).is_none() {
                if this.raise_if_not_exists {
                    return Err(DeltaTableError::Generic(format!(
                        "Constraint with name: {} doesn't exists",
                        name
                    )));
                }
                return Ok(DeltaTable::new_with_state(this.log_store, this.snapshot));
            }
```

File: crates/core/src/operations/drop_constraints.rs [hunk 6]
```diff
@@ -0,0 +1,216 @@
...
+
+            let contains_constraints = metadata
+                .configuration
+                .keys()
+                .any(|k| k.starts_with("delta.constraints"));
```
roeap, line 99: do we still need this?

python/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-python"
-version = "0.15.2"
+version = "0.16.0"
```
roeap, line 3: Should we do version bump in a separate PR? Or rather should we come up with a strategy how we usually do this?

roeap (commented):
looking good, left a few mostly style related comments.

crates/core/src/operations/drop_constraints.rs
```diff
@@ -0,0 +1,216 @@
...
+    type IntoFuture = BoxFuture<'static, Self::Output>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        let mut this = self;
+
+        Box::pin(async move {
+            let name = match this.name {
+                Some(v) => v,
+                None => return Err(DeltaTableError::Generic("No name provided".to_string())),
+            };
```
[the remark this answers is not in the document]
ion-elgreco: Ah there seems to be an inbuilt pattern for all these things already 😆

python/Cargo.toml
ion-elgreco: Usually it's a separate one, but wanted to save time here in case this ended up in main this week : )

ion-elgreco:
@roeap thanks for the feedback! I've resolved the comments

hntd187 (approved):
LGMT

crates/core/src/operations/mod.rs
```diff
@@ -199,6 +201,12 @@ impl DeltaOps {
     pub fn add_constraint(self) -> ConstraintBuilder {
         ConstraintBuilder::new(self.0.log_store, self.0.state.unwrap())
     }
+
+    /// Drops constraints from a table
+    #[must_use]
```
rtyler, line 206:
```suggestion
   #[cfg(feature = "datafusion")]
    #[must_use]
```

This should be guarded on the `datafusion` feature

rtyler (commented):
Small change and then we're ready to roll :guitar:

crates/core/src/operations/mod.rs
ion-elgreco: Ah oops :)

rtyler (approved):

Issue #2073, opened 2024-01-13
Add `UNSET TBLPROPERTIES`


ion-elgreco:

abhiaagarwal:
Will this be related to the "alter" namespace proposed in #1909? I need to implement a few of them for an application I'm developing

ion-elgreco:
@abhiaagarwal yup, already have a PR open for this

abhiaagarwal:
@ion-elgreco  Yeah, I read through it, but it doesn't explicitly namespace everything to "alter", on either the Python or the Rust side. Is that still planned? I want to develop Alter Column, I want to know what's the preferred format for developing a something in `core/operations'

abhiaagarwal:
Sorry, I was unclear — I see that it does exist in Python, but will the idea of the "alter namespace" be exported to rust as well? I'm using delta-rs primarily in rust. I'm wondering how to group all Alter-related expressions together on the rust side to be exported into the alter namespace on the python, or to just handle it on the python side exclusively.

ion-elgreco:
You don't really need to group them in rust, the grouping is mainly done in python under TableAlterer

ion-elgreco:
You don't really need to group them in rust, the grouping is mainly done in python under TableAlterer, you just expose it operation as a method under RawDeltaTable

PR #2075, opened 2024-01-13
feat(rust,python): add `unset tblproperties` operation

ion-elgreco:
# Description
It adds the unset tblproperties operation. 

I'm currently parsing the string value as DeltaConfigKey to check if the properties that are going to be unset are valid. The reason for that is to prevent unsetting of other configurations like delta.constraints etc.  

The downside here is that our ConfigKey enum should be up to date, on the other hand we could also check each property against a "protected" list of properties that can only be changed by their respective operation. If a user tries to unset a property that is in this protected list it will raise. 

While writing this, I noticed that ChangeDataFeed and AppendOnly are WriterFeatures and part of the DeltaConfigKey, not sure if that can be problematic

@roeap can you give some feedback here on the approach?

I will add the rust tests once it's a bit clear which way want to go :) 

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2073


roeap:
> While writing this, I noticed that ChangeDataFeed and AppendOnly are WriterFeatures and part of the DeltaConfigKey, not sure if that can be problematic

@ion-elgreco, not entirely sure if this is what you are referring to, but generally speaking we have to distinguish two things. 

1) If a table supports a feature, which means readers / writers have to adhere to the specs of that feature in the protocol. This is what is reflected via `readerFeatures` and `writerFeatures`.

2) If a feature is enabled - this is what is in some cases reflected via table properties - like CDC or deletion vectors.

> The downside here is that our ConfigKey enum should be up to date

I expect that the configuration keys are eventually up-streamed to kernel, where they always have to be valid anyhow. A related question would be if we want to allw setting things that delta-rs does not yet support. E.g. setting `delta.deletionVecotrsEnabled` to `true` would make the table un-readable for us right now. (though hopefully not for too long :))

ion-elgreco:
@roeap on your last question I think we should allow that because I can think of usecases where users quickly adjust table features and properties with delta-rs to then just write with spark-delta.


Back on the main point of my note, what order is done to check for example whether CDC is enabled below table feature protocol and above?

Let's take this example:

If we are below V7 protocol we only check the config so we can safely unset the property.

But what if you have a table at V6 with enableChangeDataFeed=true, then we upgrade to V7. Will that config setting be translated into a CDC writer feature enabled?


Also what if we have a table with table feature CDC enabled, then I assume any setting or unsetting the table property: enableChangeDataFeed will not affect it

crates/core/src/operations/unset_tbl_properties.rs
```diff
@@ -0,0 +1,169 @@
...
+use crate::table::state::DeltaTableState;
+use crate::DeltaTable;
+use crate::{DeltaResult, DeltaTableError};
+
+/// Remove constraints from the table
+pub struct UnsetTablePropertiesBuilder {
+    /// A snapshot of the table's state
+    snapshot: DeltaTableState,
+    /// Name of the property
+    properties: Option<Vec<String>>,
```
roeap, line 24: we could just make this a `Vec<DeltaConfigKey>` and omit the `Option` checking and key validation logic below.

crates/core/src/operations/unset_tbl_properties.rs
```diff
@@ -0,0 +1,169 @@
...
+                HashMap::from_iter([("properties".to_string(), json!(&properties))]);
+
+            let operations = DeltaOperation::UnsetTblProperties {
+                properties: properties,
+            };
+
+            let app_metadata = match this.app_metadata {
+                Some(metadata) => metadata,
+                None => HashMap::default(),
+            };
```
roeap, line 141:
```suggestion
            let app_metadata = this.app_metadata.unwrap_or_default();
```

crates/core/src/protocol/mod.rs
```diff
@@ -435,6 +435,12 @@ pub enum DeltaOperation {
         /// The status of the operation
         status: String,
     },
+
+    /// Unset table properties
+    UnsetTblProperties {
```
roeap, line 440: nit: maybe write this out `UnsetTableProperties `

crates/core/src/operations/unset_tbl_properties.rs
```diff
@@ -0,0 +1,169 @@
...
+use serde_json::json;
+
+use crate::kernel::{Action, CommitInfo, IsolationLevel, Protocol};
+use crate::logstore::LogStoreRef;
+use crate::operations::transaction::commit;
+use crate::protocol::DeltaOperation;
+use crate::table::config::DeltaConfigKey;
+use crate::table::state::DeltaTableState;
+use crate::DeltaTable;
+use crate::{DeltaResult, DeltaTableError};
```
roeap, line 17: nit:

```suggestion
use crate::{DeltaResult, DeltaTable, DeltaTableError};
```

crates/core/src/operations/unset_tbl_properties.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn with_property(mut self, name: String) -> Self {
+        match self.properties {
+            Some(mut vec_name) => {
+                vec_name.push(name);
+                self.properties = Some(vec_name);
+            }
+            None => self.properties = Some(vec![name]),
+        }
+        self
+    }
```
roeap, line 55: here is was a little bit unsure what behaviour I would expect. From the name, my first instinct was that this would set the property, not append it to any existing ones.

What do you think?

crates/core/src/operations/unset_tbl_properties.rs
```diff
@@ -0,0 +1,169 @@
...
+
+            if this.raise_if_not_exists && incorrect_config_names.len() > 0 {
+                return Err(DeltaTableError::Generic(format!(
+                    "table properties with names: {:?} don't exists",
+                    incorrect_config_names
+                )));
+            } else if !this.raise_if_not_exists && incorrect_config_names.len() == properties.len()
+            {
+                return Ok(DeltaTable::new_with_state(this.log_store, this.snapshot));
+            }
```
roeap, line 125: could we check (or did we already) how spark behaves here? while this is a noop, it may be that we still want to create a commit to recognise that the operation was executed, but not sure.

roeap (commented):
Thanks, left a few small comments.

crates/core/src/operations/unset_tbl_properties.rs
ion-elgreco: Sure, I'll check this one over the weekend!

crates/core/src/operations/unset_tbl_properties.rs
ion-elgreco: That's also fine, I will still have validate the keys than in the Python binding

crates/core/src/operations/unset_tbl_properties.rs
ion-elgreco: So, I was thinking people could either chain multiple `.with_property().with_property()` to add the configkeys or they use the `with_properties `to pass a Vec.

crates/core/src/operations/unset_tbl_properties.rs
ion-elgreco: @roeap sorry for the very late response :) But Spark-delta still does the commit, so I will update the logic to do that as well

ion-elgreco:
Putting this in draft until I address some properties that are apparently not allowed to be unset :)

PR #2080, opened 2024-01-14
docs: make an overview tab visible in docs

r3stl355:
# Description
Currently  Usage/Overview points as `usage/index.md` and therefore it does not appear in the left-hand-side menu (it is hidden behind the `Usage` header of the sub-menu. Renaming the files fixes that


MrPowers (approved):
LGTM

PR #2083, opened 2024-01-16
fix: reinstate copy-if-not-exists passthrough

emcake:
# Description
If the `AWS_COPY_IF_NOT_EXISTS` key exists, then we don't require instantiating a lock client or setting allow-unsafe-rename.

### DIFF
--- a/crates/deltalake-aws/src/lib.rs
+++ b/crates/deltalake-aws/src/lib.rs
@@ -43,6 +43,13 @@ impl LogStoreFactory for S3LogStoreFactory {
         let store = url_prefix_handler(store, Path::parse(location.path())?)?;
         let s3_options = S3StorageOptions::from_map(&options.0);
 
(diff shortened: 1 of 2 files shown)


roeap (commented):
Thanks for fixing this!

One thought, should we just check for the [corresponding key](https://docs.rs/object_store/0.9.0/object_store/aws/enum.AmazonS3ConfigKey.html#variant.CopyIfNotExists) in the storage config?

This would not require adding an extra config key, and mirrors the previous behaviour.

emcake:
> Thanks for fixing this!
> 
> One thought, should we just check for the [corresponding key](https://docs.rs/object_store/0.9.0/object_store/aws/enum.AmazonS3ConfigKey.html#variant.CopyIfNotExists) in the storage config?
> 
> This would not require adding an extra config key, and mirrors the previous behaviour.

Yep, I'll make that change.

rtyler (approved):

PR #2093, opened 2024-01-20
docs: move dynamo docs into new docs page

ion-elgreco:
# Description
Adds the dynamo docs into our new docs, within the python write_deltalake I am pointing to the guide since it's quite extensive and only for S3 users.

@rtyler @dispanser


MrPowers:
LGTM, but I think we might want to organize this content a bit differently.

Here's the current structure:

* Usage
  * Writing a table

Here's what you're proposing:

* Usage
  * Writing
    * Writing to S3 with S3 locking provider

We may want to organize the content as follows:

* Clouds
  * AWS
    * S3 locking with DynamoDB
    * AWS Lambda
    * AWS Sagemaker
  * Azure
    * Azure functions

ion-elgreco:
We could reorganize that a bit later, currently I am only moving some docs that landed by accident in the sphinx back in to the current docs.

In terms of terminology, I would say object store probably better and S3 API is also used outside of AWS so maybe best to not to mention AWS specifically here

MrPowers:
@ion-elgreco - yea, in general Google doesn't like when pages get moved to new URLs.  Perhaps we can add redirects for the links we're breaking? https://github.com/mkdocs/mkdocs-redirects

ion-elgreco:
@MrPowers a nice one, let's do that in the next reorganisation of the pages.

Then I'll also write a small example where you write from Azure

MrPowers (approved):
LGTM

PR #2111, opened 2024-01-23
chore: removed unnecessary print statement from update method

### LINKED ISSUE #2110: Is this print necessary? Can we remove this.
https://github.com/delta-io/delta-rs/blob/1a984ceff38583143d0b7e0098cb2ffc4100b018/python/deltalake/table.py#L844

I'm using this `update` function but this print statement is printing the unnecessary logs.

LilMonk:
# Description
There is a print statement in the update method of tables.py. It is not needed now. So I have removed it.

# Related Issue(s)

- closes #2110

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -841,7 +841,6 @@ class DeltaTable:
                 updates[key] = value
         elif updates is not None and new_values is None:
             for key, value in updates.items():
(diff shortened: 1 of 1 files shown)


ion-elgreco (approved):

ion-elgreco:
@LilMonk thanks!

PR #2116, opened 2024-01-23
fix: add missing pandas import

### LINKED ISSUE #2112: Pandas import error when merging tables
# Environment
Local 

**Delta-rs version**:
v0.15.1

**Binding**: 
Python

**Environment**:
- **OS**: Windows 

***
# Bug

**What happened**:
Trying to get the merging example from [the docs](https://delta-io.github.io/delta-rs/api/delta_table/delta_table_merger/), but using pandas instead of pyarrow for the tabkle. This gives a `NameError` that pandas is not defined, as in the `table.py` pandas is only imported when type checking:

https://github.com/delta-io/delta-rs/blob/1a984ceff38583143d0b7e0098cb2ffc4100b018/python/deltalake/table.py#L941

As per my limited understanding of this construct, pandas is only imported for static type checking. But seeing as it is actually used in the code, should it be imported normally? 

That the pandas dataframe gets merged in the existing delta table :). 

**How to reproduce it**:
Using the example from the docs, but with a pandas DataFrame instead of a pyarrow table:
```python
import pandas as pd

data = pd.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
write_deltalake("tmp", data)
dt = DeltaTable("tmp")
new_data = pd.DataFrame({"x": [1], "y": [7]})

(
     dt.merge(
         source=new_data,
         predicate="target.x = source.x",
         source_alias="source",
         target_alias="target")
     .when_matched_update(updates={"x": "source.x", "y": "source.y"})
     .execute()
)

dt.to_pandas()
```
results in:
`NameError: name 'pandas' is not defined`

**More details**:
I'm not very familiar with static type checking or this code, so please let me know if I did something wrong and this is not a bug!

Tim-Haarman:
# Description
Pandas was only imported for static type checking, but was actually used in the code of the module leading to import errors. Added the import similarly to `writer.py`, and updated the pandas alias in the type hints from 'pandas' to 'pd'.

# Related Issue(s)
- closes #2112 

# Documentation
x

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -37,8 +37,6 @@ from pyarrow.dataset import (
 if TYPE_CHECKING:
     import os
 
-    import pandas
-
 from deltalake._internal import DeltaDataChecker as _DeltaDataChecker
 from deltalake._internal import RawDeltaTable
 from deltalake._internal import create_deltalake as _create_deltalake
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@Tim-Haarman Thanks!

ion-elgreco (approved):

PR #2119, opened 2024-01-26
fix: set partition values for added files when building compaction plan

alexwilcoxson-rel:
# Description
After the table state refactor optimize, compaction, was not writing files to the partition directory properly nor setting the partitionValues field in the Add action.


roeap (approved):
thanks for fixing this!

PR #2122, opened 2024-01-26
docs: add dask page to integration docs

rrpelgrim:


ion-elgreco:
@MrPowers do we want to show that in our docs?

It seems dask-deltalake hasn't been updated for over 6 months and relies on an older python version of deltalake

MrPowers:
@ion-elgreco - yea, I think it's good to highlight.  

Does it depend on a specific version of `deltalake`?  Here's the dependency I could spot: https://github.com/dask-contrib/dask-deltatable/blob/main/requirements.txt#L2

ion-elgreco:
@MrPowers well it doesn't depend so our new releases broke their package since they are using private methods, which ofcourse is at someone's own risk :)

MrPowers:
@ion-elgreco - yep, well we should help them make fixes I think 😄 

I kind of view us as being responsible for the whole Delta Lake ecosystem that depends on delta-rs too.

PR #2123, opened 2024-01-26
fix: do not write empty parquet file/add on writer close; accurately …

alexwilcoxson-rel:
…track unflushed row group size

# Description
When writing a batch with the writer, if that batch would result in a flush and no more batches follow, then on close an empty parquet file (no rows) is written. This includes an Add in the log that looks like

```
    Add {
        path: "part-00002-3da49db6-e5e9-4426-8839-0092a56cc155-c000.parquet",
        partition_values: {},
        size: 346,
        modification_time: 1706297596165,
        data_change: true,
        stats: Some(
            "{\"numRecords\":0,\"minValues\":{},\"maxValues\":{},\"nullCount\":{}}",
        ),
        tags: None,
        deletion_vector: None,
        base_row_id: None,
        clustering_provider: None,
        stats_parsed: None,
    },
```

The empty stats structs causes issues with scans in datafusion.

Also changed it so the writer tracking internal buffers includes the parquet writer buffering within its row group writer.

### DIFF
--- a/crates/deltalake-core/src/operations/writer.rs
+++ b/crates/deltalake-core/src/operations/writer.rs
@@ -374,7 +374,10 @@ impl PartitionWriter {
             // flush currently buffered data to disk once we meet or exceed the target file size.
             let estimated_size = self.buffer.len() + self.arrow_writer.in_progress_size();
             if estimated_size >= self.config.target_file_size {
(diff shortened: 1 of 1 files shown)


roeap (approved):
LGTM 👍 - one more merge from main and good to go!

PR #2144, opened 2024-01-29
fix(#2143): keep specific error type when writing fails

### LINKED ISSUE #2143: JsonWriter converts structured parsing error into plain string
# Environment

**Delta-rs version**:
0.16.5 and `main` branch

**Binding**: 
Rust

**Environment**:
- **Cloud provider**:
- **OS**: macOS Ventura 13.2.1
- **Other**:

***
# Bug

**What happened**:
When parsing fails, JsonWriter converts `DeltaWriterError::DeltaTable(DeltaTableError::Arrow { source: ArrowError::JsonError(_) })` to a `DeltaTableError::Generic(string)`. This makes it impossible to handle error cases in client code without parsing strings.

JsonWriter should return `DeltaTableError::Arrow{ source: ArrowError::JsonError(_) }`.

**How to reproduce it**:
Add the following unit test in `json.rs`:
```
    #[tokio::test]
    async fn test_parsing_error() {
        let table_dir = tempfile::tempdir().unwrap();
        let schema = get_delta_schema();
        let path = table_dir.path().to_str().unwrap().to_string();

        let arrow_schema = <ArrowSchema as TryFrom<&StructType>>::try_from(&schema).unwrap();
        let mut writer = JsonWriter::try_new(
            path.clone(),
            Arc::new(arrow_schema),
            Some(vec!["modified".to_string()]),
            None,
        )
        .unwrap();

        let data = serde_json::json!(
            {
                "id" : "A",
                "value": "abc",
                "modified": "2021-02-01"
            }
        );

        let res = writer.write(vec![data]).await;
        // This assertion fails because the result is actually an `Err(DeltaTableError::Generic(string))`
        assert!(matches!(
            res,
            Err(DeltaTableError::Arrow {
                source: ArrowError::JsonError(_)
            })
        ));
    }
```

**More details**:

abaerptc:
# Description
Add case to error conversion implementation to preserve structured parsing error instead of converting to plain string.

# Related Issue(s)
- closes #2143

### DIFF
--- a/crates/core/src/writer/json.rs
+++ b/crates/core/src/writer/json.rs
@@ -415,6 +415,7 @@ fn extract_partition_values(
 
 #[cfg(test)]
 mod tests {
+    use arrow_schema::ArrowError;
     use parquet::file::reader::FileReader;
     use parquet::file::serialized_reader::SerializedFileReader;
     use std::fs::File;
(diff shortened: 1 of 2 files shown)


rtyler:
Thanks for the pull request @abaerptc, unfortunately I just dropped some folder renames into `main`, would you mind rebasing please?

rtyler (approved):

PR #2147, opened 2024-01-30
docs: add delta lake best practices

MrPowers:
Adds a docs page on the Delta Lake best practices.

This is a first pass and I think this should evolve over time.

This is some of the most important content for our users IMO.

### DIFF
--- a/docs/delta-lake-best-practices.md
+++ b/docs/delta-lake-best-practices.md
@@ -1,0 +1,131 @@
+# Delta Lake Best Practices
+
+This page outlines Delta Lake best practices.
+
+You should consider several factors to optimize the performance of your Delta tables and minimize costs.
+
+The Delta Lake best practices depend on your data ingestion into the Delta table and query patterns.  You must understand your data and how users run queries to best leverage Delta Lake.
+
+## Compacting small files
+
+Delta tables work best when the files are “right-sized”.  Files that are too small create I/O overhead.  Files that are too large limit the impact of file skipping (a critical query optimization).
+
+Delta tables can accumulate a lot of small files, especially if you’re frequently writing small amounts of data.  If your table has many small files, you should run a small compaction operation to consolidate all the tiny files into “right-sized” files.
+
+It’s generally best for files in a Delta table to be between 100MB and 1GB, but that can vary based on the overall size of the table and the query patterns.
+
+Delta Lake makes it easy to [compact the small files](https://delta-io.github.io/delta-rs/usage/optimize/small-file-compaction-with-optimize/).
+
+## Optimizing table layout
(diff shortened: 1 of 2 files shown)


docs/delta-lake-best-practices.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Partitioning datasets
+
+You can partition your Delta tables, which separates the data by one or more partition keys into separate folders.  Partitioning can be an excellent performance optimization (when you filter on the partition key) and is a good way to sidestep concurrency conflict issues.
+
+Hive-style partitioning also has some significant downsides.
+
+* It’s only suitable for low-cardinality columns.
+* It can create many small files, especially if you use the wrong partition key or frequently update the Delta table.
+* It can cause some queries that don’t rely on the partition key to run slower (because of the excessive number of small files)
```
ion-elgreco, line 37: Maybe highlight that a lot of small files is problematic for IO throughput

docs/delta-lake-best-practices.md
```diff
@@ -0,0 +1,131 @@
...
+
+Keeping old versions of Delta tables in storage is often desirable because it allows for versioned data, time travel, and rolling back tables to a previous state.
+
+If you don’t want to leverage older versions of a table, then you should remove the legacy files from storage with the vacuum command.  Vacuum will remove all files older than the table retention period and marked for removal in the transaction log.
+
+You only need to vacuum when you perform operations that mark files for removal in the transaction log.  An append-only table doesn’t create legacy files that need to be vacuumed.
+
+Create a good vacuum strategy for your tables to minimize your storage costs.
+
+## Registering tables in a metastore/catalog
```
ion-elgreco, line 83: I think we need to make a huge note here because none of the catalogs allow it to be read outside of the vendors ecosystem, so delta-rs cannot ready any catalog.

ion-elgreco (commented):
Nice docs as always, just left two comment :)

It may also be worthwhile to mention that for data science usecases you would want to keep the logs indefinitely or as long as you want a period of traceability. Because the moment you remove the logs you lose the ability to time travel or check the change data feed, which could mean you cannot explain to business or legal why your model said Y in 2022 for example

docs/delta-lake-best-practices.md
MrPowers: Yea, I am just going to take out this whole section for now.  Hopefully we have a better catalog story soon.

ion-elgreco: It would be really nice if databricks opens up the unity catalog to be accessible from the outside, this would solve the disjunction between the delta-rs and databricks spark -delta world

PR #2150, opened 2024-01-31
fix: replace BTreeMap with IndexMap to preserve insertion order

roeap:
# Description

When switching to using `BTreeMap` for representing partition values, I introduced a bug, thinking `BTreeMap` would preserve insertion order. While it is ordered, it is ordered based on keys.

This PR moves to using `IndexMap`, which actually preserves insertion order.


rtyler (approved):
I wish there was a test to prove that the orderingness of IndexMap fixed a problem with BTreeMap, but oh well, let's roll :smile:

Issue #2131, opened 2024-01-28
Refactor commit interface to be a Builder


Blajda:
# Description
Currently  [commit](https://github.com/delta-io/delta-rs/blob/0f6790fd90be869990f55379514339eb09280ba3/crates/deltalake-core/src/operations/transaction/mod.rs#L174-L180) is not a builder which results in a brittle interface whenever new commit options are added. Refactor to be a builder so code changes are isolated to the module and don't require updates in all operations. 

Operation should be updated to delegate to this builder but must keep some members like `log_store` and `operation` private.

**Related Issue(s)**
#2130 - Implementing Txn identifiers would require adding a new parameter to the current function and break various call sites.

hntd187:
take

PR #2154, opened 2024-02-01
refactor!: use builder for commit interface

Blajda:
# Description
Refactor the commit function to instead be accessed through a builder interface to prevent breaking call sites when new  commit attributes are added.

This also introduces a new public struct called `CommitProperties` to be used by any delta operation and to ensure a consistent interface for end users. With this change we should break usage of `app_metadata` on operations and instead have users use this struct. 

# Related Issue(s)
- closes: #2131


crates/core/src/operations/transaction/mod.rs
```diff
@@ -167,94 +167,155 @@ pub async fn prepare_commit<'a>(
...
+    max_retries: usize,
+}
 
-    Err(TransactionError::MaxCommitAttempts(max_retries as i32).into())
+impl<'a> Commit<'a> {
+    /// Commit a transaction, with up configurable number of retries. This is higher-level transaction API.
+    ///
+    /// The function will error early if the a concurrent transaction has already been committed
+    /// and conflicts with this transaction.
+    pub async fn execute(&self) -> DeltaResult<i64> {
```
roeap, line 255: What do you think about implementing `IntoFuture` for `Commit` to execute it?

roeap (commented):
So far I believe this looks great. There are a few things I think we should discuss how we want to model them.

With support for deletion vectors and CDF coming up, we need to perform two checks:
1) that the commit is consistent by itself.
2) conflict resolution.

No. 2) is taken care of by the current commit logic. For 1) I think we could leverage the `build` function on the `CommitBuilder`. These checks should for instance assert these rules from [protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file):

> In the following statements, dvId can refer to either the unique id of a specific Deletion Vector (deletionVector.uniqueId) or to NULL, indicating that no rows are invalidated. Since actions within a given Delta commit are not guaranteed to be applied in order, a valid version is restricted to contain at most one file action of the same type (i.e. add/remove) for any one combination of path and dvId. Moreover, for simplicity it is required that there is at most one file action of the same type for any path (regardless of dvId). That means specifically that for any commit…
> * it is legal for the same path to occur in an add action and a remove action, but with two different dvIds.
> * it is legal for the same path to be added and/or removed and also occur in a cdc action.
> * it is illegal for the same path to be occur twice with different dvIds within each set of add or remove actions.

We should see if we can find more, but generally we should start checking .. :). IF we think adding these checks is beyond the scope of this PR, we should maybe already make `build` fallible and add a TODO / issue?

Another conceptually larger change that I think will make our codebase a bit cleaner is to make the `Snapshot` non-optional. Even if we are creating a table we always require at least the `Protocol` and `Metadata` action, which we can use to create an empty snapshot with no file action / empty `LogSegment` and a version of `-1`.

Beyond the scope of this PR, but mentioning it anyhow :) ... Hopefully we can eventually also model the actions we commit as a `RecordBatch` rather then action types, which I believe will go a long way to more deeply integrate into the datafusion flow. The current sink in DF just emits an `i64` representing rows written. We could eventually roll our own Sinks that emit action batches, which may help us implement the CDC flow etc more cleanly. At that point the commit could also accept an `ExecutionPlan` as an input.

crates/core/src/operations/transaction/mod.rs
```diff
@@ -167,94 +167,155 @@ pub async fn prepare_commit<'a>(
...
-        actions,
-        operation,
-        read_snapshot,
-        app_metadata,
-        15,
-    )
-    .await
+#[derive(Default, Clone)]
+pub struct CommitProperties {
+    pub(crate) app_metadata: HashMap<String, Value>,
```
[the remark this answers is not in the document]
Blajda: The mixture of elements between written and configuration items annoys me too. When creating this struct the primary focus is to support transaction identifiers and eventually domain metadata which are written in the commit. Those two features should be exposed for any transaction performed on the table.

I think app metadata should be kept with the commit builder and we instead can define a merging strategy for how the final `app metadata` is created. The hierarchy can be something like `Operation Metadata` > `DeltaTableConfig Metadata` > `User Metadata`. User metadata should be last to prevent end users from overwriting critical elements like user information, notebook, etc.

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -111,6 +118,150 @@ impl From<DataFusionError> for DeltaTableError {
...
+    /// Iterate over all files in the log matching a predicate
+    fn files_matching_predicate<'a>(
+        &'a self,
+        filters: &'a [Expr],
+    ) -> DeltaResult<impl Iterator<Item = Add>>;
+}
+
+pub(crate) fn files_matching_predicate<'a>(
+    snapshot: &'a EagerSnapshot,
+    filters: &[Expr],
```
hntd187, line 238: Does the missing lifetime matter here?

Blajda: I think the compiler is able to infer the lifetime of filters must be greater than or equal to `'a`. Had to introduce the lifetimes in the early stages to satisfy the compiler but I can try removing them now.

crates/core/src/operations/transaction/mod.rs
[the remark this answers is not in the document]
Blajda: > Another conceptually larger change that I think will make our codebase a bit cleaner is to make the `Snapshot` non-optional. Even if we are creating a table we always require at least the `Protocol` and `Metadata` action, which we can use to create an empty snapshot with no file action / empty `LogSegment` and a version of `-1`.

@roeap I'm try to design the commit interface to prepare for this change. I have one question. Will `Snapshot` always contain the required information required to perform the commit? I.E Will there ever be an instance where commit will need an `EagerSnapshot` or a `DeltaTableState`? Also are there some workflows (e.g blind appends) where having a `log_segment` might not be required.

If only `Snapshot` is required then I'll change the build definition to 
```rust
pub fn build(
        self,
        snapshot: &Snapshot,
        log_store: LogStoreRef,
        operation: DeltaOperation,
    ) -> Result<PreCommit<'a>, CommitBuilderError>
```

But if different levels of information are required then we can maybe use a trait object and progressively enhance depending on what information is available. I tried playing around with this concept in the latest commit.

Interested on your thoughts and how we should proceed here.

crates/core/src/operations/optimize.rs
```diff
@@ -608,7 +606,7 @@ impl MergePlan {
         #[allow(unused_variables)] // used behind a feature flag
         max_spill_size: usize,
         min_commit_interval: Option<Duration>,
-        app_metadata: Option<HashMap<String, serde_json::Value>>,
+        commit_properties: CommitProperties,
```
ion-elgreco, line 609: Why not make it here mutual as well?

ion-elgreco (dismissed):
Looks good! Just left one very small comment

crates/core/src/operations/transaction/mod.rs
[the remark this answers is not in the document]
roeap: hmm .. `Snapshot` will always contain `Metadata` and `Protocol` - for conflict resolution we need to read new actions anyhow. so for blind appends we may not actually need to get any actions beyond that ... So not eagerly fetching all data, but catching things just in time seems like a good way to go.

roeap (commented):
just did a first quick scan before going to work :).

Want to give this a proper review, but excited to see this land.

crates/core/src/operations/optimize.rs
Blajda: Hi Ion, I understand what you mean. Can you expand you what mean by mutual?

crates/core/src/operations/transaction/mod.rs
```diff
@@ -104,157 +175,371 @@ impl From<TransactionError> for DeltaTableError {
...
+    fn config(&self) -> TableConfig;
+
+    /// Get the table protocol of the snapshot
+    fn protocol(&self) -> &Protocol;
+
+    /// Get the table metadata of the snapshot
+    fn metadata(&self) -> &Metadata;
+
+    /// Try to cast this table reference to a `EagerSnapshot`
+    fn eager_snapshot(&self) -> Option<&EagerSnapshot>;
```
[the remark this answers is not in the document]
Blajda: I agree if we want to allow "promoting" a table reference to a particular level then we might need to load data. I'll do it in follow up PR.

crates/core/src/operations/optimize.rs
ion-elgreco: Meant why not make the commit_properties mutable, like you did here and other places: https://github.com/delta-io/delta-rs/pull/2154/files#diff-a51cc3d9c5abab69a21a4d348682a89d656a59501bb3d2be0f24c0dc7b5b4bb6R193

ion-elgreco (approved):

PR #2160, opened 2024-02-01
feat: implement clone for DeltaTable struct

mightyshazam:
# Description
This PR adds the derive(Clone) attribute to the`DeltaTable` struct.

# Related Issue(s)

# Documentation


roeap:
@mightyshazam - thanks, I actually have been waiting for this PR 😆.

For a looong time folks wanted `DeltaTable` to be `Clone`able and we have always rejected it since cloning the state used to be potentially ridiculously expensive ... now its practically for free 🥲.

So I am fine to merge this. However, I'd be curious about the use case. Specifically in most situations I would recommend cloning the state itself, as `DeltaTable` just exposes some higher level functionality on top of it and may not have a valid state yet, which forces you to deal with a lot of `Option`s and/or `Result`s which you avoid by just using the `Snapshot`.

rtyler (approved):

Issue #2169, opened 2024-02-06
Broken filter for newly created delta table


Hanspagh:
# Environment

**Delta-rs version**:
'0.15.2'

**Binding**: 
python

**Environment**:
- **OS**: Mac 

***
# Bug

When creating a new delta table from a pandas dataframe, it appears that the filter predicate is broken for some expression 

**What happened**:
.to_pandas() and .to_pyarrow_dataset() return 0 data

**What you expected to happen**:
The above functions should return the data reflected in the filter predicate

**How to reproduce it**:
This is a large dataset that I cannot share, but please point me in any directions for how to debug this.

This is how I achieved my current results

```python
from pyarrow import dataset as ds
from pyarrow.parquet import ParquetFile, ParquetDataset
import pyarrow.compute as pc
from deltalake import write_deltalake, convert_to_deltalake, DeltaTable
import pandas as pd

df = "some_internal_data"

write_deltalake("broken_nlrot.delta", df)

dt_broken = DeltaTable("broken_nlrot.delta")
dt_broken.to_pyarrow_dataset().to_table(filter=(pc.field("terminal") == "NLROTTM")).shape
# returns (0, 19)

# Trying to read strait with pyarrow
ds.dataset("broken_nlrot.delta").to_table(filter=(pc.field("terminal") == "NLROTTM")).shape
# returns (773536, 19) which is the expected result

#similarly to_pandas also does not seem to work 
dt_broken.to_pandas(filters=[("terminal", "==", "NLROTTM")]).shape 
# returns (0,19)

# again with pyarrow we are fine
ParquetDataset("broken_nlrot.delta", filters=[("terminal", "==", "NLROTTM")]).read().shape
# returns (773536, 19) which is the expected result

#oddly enough it seems to work when I choose a different filter predicate I dont get zero rows, but still wrong resulst
dt_broken.to_pandas(filters=[("terminal", "==", "USSAVGC")]).shape
# (89406, 19) - Wrong

ds.dataset("broken_nlrot.delta").to_table(filter=(pc.field("terminal") == "USSAVGC")).shape
# (420029, 19) - Correct

# pulling the full data in pandas also seems to work
full = dt_broken.to_pandas()
full[full["terminal"] == "NLROTTM"].shape
# (773536, 19)

full[full["terminal"] == "USSAVGC"].shape
# (420029, 19)


# calling optimize seem to fix the problem, but this should not be needed in order for filters to work
dt.optimize.z_order(["terminal"])
[...]

PR #2172, opened 2024-02-06
fix(python): skip empty row groups during stats gathering

ion-elgreco:
# Description
For some odd reason the pyarrow parquet writer will leave empty row groups in the parquet file when it hits the max_row limit that's passed. While grabbing the stats we were checking if all row_groups were having stats added to them but these empty row groups had no stats so it causes the whole file add action to get no stats recorded.

We now skip empty row groups while gathering the stats to prevent this.

In v0.15.2 we now also evaluate files with no stats mentioned as null @roeap @rtyler  not sure if this is entirely correct as well

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2169

### DIFF
--- a/python/tests/test_writer.py
+++ b/python/tests/test_writer.py
@@ -1253,7 +1253,8 @@ def test_with_deltalake_schema(tmp_path: pathlib.Path, sample_data: pa.Table):
     assert delta_table.schema().to_pyarrow() == sample_data.schema
 
 
-def test_write_stats_empty_rowgroups_2169(tmp_path: pathlib.Path):
+def test_write_stats_empty_rowgroups(tmp_path: pathlib.Path):
+    # https://github.com/delta-io/delta-rs/issues/2169
     data = pa.table(
         {
             "data": pa.array(["B"] * 1024 * 33),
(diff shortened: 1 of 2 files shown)


python/tests/test_writer.py
```diff
@@ -1251,3 +1251,24 @@ def test_with_deltalake_schema(tmp_path: pathlib.Path, sample_data: pa.Table):
     )
     delta_table = DeltaTable(tmp_path)
     assert delta_table.schema().to_pyarrow() == sample_data.schema
+
+
+def test_write_stats_empty_rowgroups_2169(tmp_path: pathlib.Path):
+    data = pa.table(
```
wjones127, line 1257: Let's add a link here to the issue, since it might not be obvious what the number in the function name is.
```suggestion
def test_write_stats_empty_rowgroups(tmp_path: pathlib.Path):
    # https://github.com/delta-io/delta-rs/issues/2169
    data = pa.table(
```

wjones127 (dismissed):

python/tests/test_writer.py
ion-elgreco: Good point :) @wjones127 can you approve again?

wjones127 (approved):

PR #2179, opened 2024-02-09
feat: add comment to explain why assert has failed and show state

### LINKED ISSUE #2177: Add error message for parition_by check
Great work on this project!

I triggered the assertion error below. However, as there is no string message after the assertion I get no indication of what went wrong from logging

**Delta-rs version**:
0.15.1

**Environment**:
MacOS

***
# Bug

**What happened**:
I hit this assertion check in write_deltalake:
```python
assert partition_by == table.metadata().partition_columns
```

Produce a string to explain what the error is (and preferably how it might be avoided)

braaannigan:
# Description
Add comment to explain why assert has failed for partition_by columns and show state

# Related Issue(s)
 closes #2177

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -100,7 +100,8 @@ def write_deltalake(
     large_dtypes: bool = ...,
     engine: Literal["pyarrow"] = ...,
     custom_metadata: Optional[Dict[str, str]] = ...,
-) -> None: ...
+) -> None:
+    ...
 
 
 @overload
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@braaannigan I think you ran black while we use ruff for formatting, can you run ruff format over it?

braaannigan:
> @braaannigan I think you ran black while we use ruff for formatting, can you run ruff format over it?

Cheers, I've tried again

ion-elgreco (approved):
Thanks! @braaannigan

PR #2182, opened 2024-02-11
fix: correct map field names

emcake:
# Description
This changes the field names inside map types to be `key` and `value` instead of `keys` and `values`. This matches the parquet spec for encoding map columns: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps 

I found this while trying to load a checkpoint from a table where it was failing with spurious errors. Pyarrow is tolerant of this to a degree but other readers like parquet4s (used in e.g delta standalone) took issue with it.

### DIFF
--- a/crates/core/src/kernel/arrow/mod.rs
+++ b/crates/core/src/kernel/arrow/mod.rs
@@ -14,8 +14,8 @@ pub(crate) mod extract;
 pub(crate) mod json;
 
 const MAP_ROOT_DEFAULT: &str = "entries";
-const MAP_KEY_DEFAULT: &str = "keys";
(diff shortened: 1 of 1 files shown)


ion-elgreco (approved):

roeap:
@emcake @ion-elgreco - sorry for being late to the discussion. Just to create awareness, bit there is no "right" wat for theses names -AFAIK. Different parquet writers and readers behave differently and different arrow implementations also take different positions on this.

So "right" only applies to a given implementation. 

I habe some work queued to try and solve some of this more generally, but the changes in this Pr may not be quite long-lived depending on which implementation we will end up aligning with.

PR #2190, opened 2024-02-18
fix: add data_type and nullable to StructField hash (#2045)

### LINKED ISSUE #2045: Hash for StructField should consider more than the name
Two StructFields with the same name but different data_types are considered to hash equivalently, which I think is incorrect.

I'm working around this for now, but filing for @roeap and I to discuss later

sonhmai:
# Description
Correct hashing of StructField and add tests.
- Current:  Two StructFields with the same name but different `data_type` or different `nullable` are considered to hash equivalently.
- To be: `name`, `data_type` and `nullability` should match for a field to be considered equivalent.

# Related Issue(s)
closes #2045

### DIFF
--- a/crates/core/src/kernel/models/schema.rs
+++ b/crates/core/src/kernel/models/schema.rs
@@ -134,6 +134,8 @@ pub struct StructField {
 impl Hash for StructField {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.name.hash(state);
+        self.data_type.hash(state);
+        self.nullable.hash(state);
     }
 }
 
@@ -215,7 +217,7 @@ impl StructField {
 
 /// A struct is used to represent both the top-level schema of the table
 /// as well as struct columns that contain nested columns.
-#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Eq)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Eq, Hash)]
 pub struct StructType {
     #[serde(rename = "type")]
     /// The type of this struct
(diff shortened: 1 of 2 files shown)


sonhmai:
there seems to be an error unrelated to this PR

```
FAILED tests/test_convert_to_delta.py::test_convert_delta_with_partitioning - _internal.DeltaError: 
Generic error: The schema of partition columns must be provided to convert a Parquet table to a Delta table
```

ion-elgreco:
@sonhmai not so sure about that, these failures weren't there before. Maybe some interaction somewhere else in the code?

sonhmai:
> @sonhmai not so sure about that, these failures weren't there before. Maybe some interaction somewhere else in the code?

yep. will definitely look into details.

sonhmai:
@ion-elgreco  fix pushed with local tests passed. can you plz let workflow run? Thanks.

sonhmai:
@rtyler  @ion-elgreco  all checks passed!

ion-elgreco (approved):
@sonhmai thanks for the fix! 😀

crates/core/src/kernel/models/schema.rs
```diff
@@ -641,6 +643,7 @@ mod tests {
     use super::*;
     use serde_json;
     use serde_json::json;
+    use std::hash::DefaultHasher;
```
emcake, line 646: incase anyone else runs into this - this raises the MSRV of running the tests to 1.76.

sonhmai: thanks for pointing this out!

Issue #2006, opened 2023-12-31
Proposal to create logical plans for operations


Blajda:
# Description
Propose further work that I'd like to perform regarding the creation reusable logical relations. Also helps with identifying relations we would need with substrait. 

**Delta Find Files**
Purpose: Identify files that contain records that satisfy a predicate.

This relation will generate a record batch stream with a single column called `path`. `path` will then map to an `Add` action in the Delta table.
This relation will also maintain a list of files that satisfy the predicate which can be passed sideways to relations downstream.

**Delta Scan**
Purpose: Scan the Delta Table

Update `DeltaScan` to take an optional input stream that contains paths of files to be scanned. This will enable `DeltaScan` to consume output of `DeltaFindFile`.
Currently when using find files, we must wait for the entire operation to complete and then we build the scan. The change enables Delta Scan to start when the first candidate file is identified.
I think this will require some significant work since it will involve refactoring the current DeltaScan implementation.

**Delta Write**
Purpose: Write records to storage, conflict resolution, and commit creation

Takes an single input stream of  data that matches that tables schema and creates `Add` actions for each new file.
Information can be passed sideways to include additional delta actions to add to the commit. E.G `DeltaDelete` can provide a stream of `Remove` actions. 

**Delta Delete**
Purpose: Delete Records from the table.

Given a predicate delete records from the Delta table.
Delta Delete can take an optional stream of records and will output records that do NOT satisfy the predicate.
It will maintain a stream of `Remove` actions can be passed sideways to other operations downstream.

The input stream is optional since there are cases where delete determine which files to remove without a need for a scan. An optimization phase can help determine when this is the case.

## Diagram
High level diagram of how these relation will connect.

```
               ┌───────────────────────┐
               │   Delta Find Files    │
               │                       │
               │  Predicate:           │
           ┌───┤    Version:           │
           │   │                       │
           │   └──────────┬────────────┘
           │              │
           │              ▼
   Files   │   ┌───────────────────────┐
  Matched  │   │     Delta Scan        │
   List    │   │                       │
           │   │   Version:            │
           │   │                       │
           │   │                       │
           │   └──────────┬────────────┘
           │              │
           │              ▼
           │   ┌───────────────────────┐
           └──►│     Delta Delete      │
               │                       │
               │  Predicate:           │
           ┌───┤                       │
           │   └──────────┬────────────┘
  Remove   │              │
  Actions  │              ▼
           │   ┌───────────────────────┐
[...]

PR #2194, opened 2024-02-19
feat: logical Node for find files

hntd187:
# Description
Some of my first workings on David's proposal in #2006, this is also meant to push #2048 and general CDF forward as well by making the logical operations of delta tables more composable than they are today. 

# Related Issue(s)
#2006 
#2048 

I think and @Blajda correct me here, we can build upon this and eventually move towards a `DeltaPlanner` esq enum for operations and their associated logical plan building. 

# Still to do

- [ ] Implement different path for partition columns that don't require scanning the file
- [ ] Plumbing into `DeltaScan` so delta scan can make use of this logical node
- [ ] General polish and cleanup, there are lots of unnecessary fields and way things are built
- [ ] More tests, there is currently a large integration style end to end test, but this can / should be broken down


Blajda (changes requested):
There's quite a bit work that's still required. When implementing this operator my recommendation is to reuse as much of the `DeltaScan` functionality a possible. This will help avoid bugs in the future whenever support for schema evolution / column mapping is introduced.

crates/core/src/delta_datafusion/find_files/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+                .read_parquet(row, ParquetReadOptions::default())
+                .await?
+                .filter(predicate.to_owned())?;
+            if df.count().await? > 0 {
+                batches.push(row);
+            }
+        }
+    }
+    let str_array = Arc::new(StringArray::from(batches));
+    RecordBatch::try_new(only_file_path_schema(), vec![str_array]).map_err(Into::into)
```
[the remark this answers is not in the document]
hntd187: Another thing I have to add is when it's partition columns we don't need to read the file at all since it can be inferred from the path itself.

crates/core/src/delta_datafusion/find_files/physical.rs
```diff
@@ -0,0 +1,124 @@
...
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn schema(&self) -> SchemaRef {
+        self.schema.clone()
+    }
+
+    fn output_partitioning(&self) -> Partitioning {
+        Partitioning::UnknownPartitioning(0)
```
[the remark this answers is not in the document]
hntd187: This was just a lazy initial implementation, I can fix that

crates/core/src/delta_datafusion/find_files/mod.rs
```diff
@@ -1,29 +1,43 @@
...
 
 lazy_static! {
     static ref ONLY_FILES_SCHEMA: Arc<Schema> = {
         let mut builder = SchemaBuilder::new();
-        builder.push(Field::new(PATH_COLUMN, DataType::Utf8, false));
+        builder.push(Field::new(
+            PATH_COLUMN,
+            DataType::Dictionary(Box::new(DataType::UInt16), Box::new(DataType::Utf8)),
+            false,
+        ));
```
Blajda, line 40: Yeah there's definitely a bit of a weird interface mismatch. This operation must only output a file path at most once so there is minimal value in having a dictionary being returned to end user however the dictionary encoding should be used when we perform a scan.
Is it easy to simply have string be returned?

crates/core/src/delta_datafusion/find_files/mod.rs
```diff
@@ -70,81 +85,199 @@ impl QueryPlanner for FindFilesPlanner {
...
+    let limit: Arc<dyn ExecutionPlan> = Arc::new(LocalLimitExec::new(filter, 1));
+    let field_idx = input_schema.index_of(PATH_COLUMN)?;
+    let task_ctx = Arc::new(TaskContext::from(&state));
+    let path_batches: Vec<RecordBatch> = datafusion::physical_plan::collect(limit, task_ctx)
+        .await?
+        .into_iter()
+        .map(|batch| batch.project(&[field_idx]).unwrap())
+        .collect();
+
+    let result_batches = concat_batches(&ONLY_FILES_SCHEMA.clone(), &path_batches)?;
```
Blajda, line 183: I was thinking for this operation to output file paths as soon as they are discovered to allow operations downstream to begin their work ASAP. When performing an memory only scan to makes sense to output as a single batch since IO is minimal.
For this PR I think it's okay since it aligns with the previous behaviour but its something we can change.

Blajda (commented):
Significant improvements. 
How much effort / what would be required to remove the dictionary output schema?

crates/core/src/delta_datafusion/find_files/mod.rs
hntd187: can you elaborate a bit more? I understand what you are suggesting here, but I'm not sure what you are expecting. Like you want the return type to be a vec or record batches or a stream or something?

Blajda: Yes output record batches and these record batches should simply be a string. There is no benefit for using dictionaries in the output.

hntd187: Oh you're talking about the actual schema of the record batch, I thought you were saying the return type of the function was wrong or something of the sort for it to be "immediately available" or something of the sort.

crates/core/src/delta_datafusion/find_files/mod.rs
hntd187: I added to delta scan config the ability to turn off the dictionary encoding, it's on by default everywhere else, but here we use just a simple string as you mentioned

crates/core/src/delta_datafusion/find_files/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+
+async fn scan_table_by_files(
+    snapshot: DeltaTableState,
+    log_store: LogStoreRef,
+    state: SessionState,
+    expression: Expr,
+) -> Result<RecordBatch> {
+    register_store(log_store.clone(), state.runtime_env().clone());
+    let scan_config = DeltaScanConfigBuilder::new()
+        .wrap_partition_values(false)
```
Blajda, line 135: So we don't want to completely disable dictionary values when performing a scan since it can provide significant memory savings benefits. I'd prefer we wrap the partition values during a scan however in the output batches for the operation they should be provided as strings.

Blajda (approved):
LGTM. Thanks for making the changes.

PR #2206, opened 2024-02-24
fix: canonicalize config keys

emcake:
# Description
When merging supplied storage options with environment variables, we need to canonicalize the items in the storage options first. This is because for some keys there is multiple valid definitions (`AWS_ENDPOINT_URL`, `AWS_ENDPOINT`) and also casing (`AWS_ACCESS_KEY_ID`, `aws_access_key_id`). It seems like the 'more canonical' lowercase versions win, which can cause surprise if you have one set of access keys in your environment and one in your storage options. This is exacerbated as the `S3StorageOptions` construction also _sets_ envrionment variables, and it always does so into the canonical lowercase versions. I found this when opening two buckets with different credentials, and the credentials for the first were being used to open the second.

### DIFF
--- a/crates/aws/src/storage.rs
+++ b/crates/aws/src/storage.rs
@@ -669,23 +669,22 @@ mod tests {
     fn when_merging_with_env_unsupplied_options_are_added() {
         ScopedEnv::run(|| {
             let raw_options = hashmap! {};
-            let keys_under_test = vec![
-                AmazonS3ConfigKey::Endpoint,
-                AmazonS3ConfigKey::SecretAccessKey,
-                AmazonS3ConfigKey::Region,
-                AmazonS3ConfigKey::AccessKeyId,
-            ];
 
-            for k in keys_under_test.iter() {
-                std::env::set_var(k.as_ref(), "env_key");
-            }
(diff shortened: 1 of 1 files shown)


rtyler (approved):

rtyler:
Looks like the failure is just in some tests which are not expecting some environment variables to be set. Our option I think is to either have these new tests reset their environment variables, or update the other tests.

emcake:
> Looks like the failure is just in some tests which are not expecting some environment variables to be set. Our option I think is to either have these new tests reset their environment variables, or update the other tests.

I introduced a small utility to all of the tests to restore the environment to how it was when they captured it. Let me know if you think this approach is reasonable?

rtyler (dismissed):

emcake:
☝️ I briefly flirted with changing the tests to be a lot more careful about which environment variables they touched. This was because the integration tests workflow itself actually sets some extra environment vars, which was messing with the test. However on reflection on doing so, by being canoncalizing the keys under test, I managed to partly defeat the purpose of the test...

In the end I went for the simpler solution from the other tests of just unsetting those variables.

rtyler (approved):

PR #2207, opened 2024-02-24
fix(writer): retry storage.put on temporary network errors

qinix:
# Description

Read-only operations are retried by [`ObjectStore`] internally. However, PUT/DELETE operations
are not retried even thought they are technically idempotent. [`ObjectStore`] does not retry
those operations because having preconditions may produce different results for the same
request. PUT/DELETE operations without preconditions are idempotent and can be retried.
Unfortunately, [`ObjectStore`]'s retry mechanism only works on HTTP request level, thus there
is no way to distinguish whether a request has preconditions or not.

This PR provides additional methods for working with [`ObjectStore`] that automatically retry
unconditional operations when they fail, and use these methods in writer to retry on temporary
network errors.

# Related Issue(s)

# See

- https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.2
- https://github.com/apache/arrow-rs/pull/5278

### DIFF
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -13,6 +13,7 @@ use crate::errors::{DeltaResult, DeltaTableError};
 use crate::kernel::{Action, CommitInfo, ReaderFeatures, WriterFeatures};
 use crate::logstore::LogStore;
 use crate::protocol::DeltaOperation;
+use crate::storage::ObjectStoreRetryExt;
 use crate::table::state::DeltaTableState;
 
 pub use self::protocol::INSTANCE as PROTOCOL;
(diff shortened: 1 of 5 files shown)


rtyler:
The code largely looks good to me @qinix. I have reviewed what `object_store::Error` has , and since this is only retrying on the `Generic` variant, I'm not sure what types of failure conditions you're seeing in the real world?

qinix:
I encountered the following error (reformatted for readability):

```
called `Result::unwrap()` on an `Err` value: ObjectStore { 
    source: Generic { 
        store: "S3", 
        source: Reqwest { 
            retries: 0, 
            max_retries: 10, 
            elapsed: 30.000858021s, 
            retry_timeout: 180s, 
            source: reqwest::Error {
                kind: Request,
                url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("REDACTED")), port: None, path: "/table/REDACTED.zstd.parquet", query: None, fragment: None },
                source: TimedOut
            }
        }
    }
}
```

Inside the `Generic` error it was [`object_store::client::retry::Error`](https://github.com/apache/arrow-rs/blob/4325900818dfd61745fc81614b49764b210c95d7/object_store/src/client/retry.rs#L43). I can't `downcast` to this concrete type as it's not publicly exported ([the `mod object_store::client` is private](https://github.com/apache/arrow-rs/blob/4325900818dfd61745fc81614b49764b210c95d7/object_store/src/lib.rs#L482)).

I dislike the way of retrying `Generic` error, however it's impossible to know the concrete error without modifying `object_store` to export the concrete error or doing the `err.to_string().contains(...)` trick.

Besides of `Generic` error, other variants are returned from server side in normal logic, thus those should not be retried.

rtyler (approved):

PR #2215, opened 2024-02-25
chore: remove caches from github actions

rtyler:
These caches cause more problems than they're worth and cause runners to exhaust their disk space


Blajda (approved):
LGTM! Hopefully the integration tests are less flakey after this.

PR #2222, opened 2024-02-26
perf: directly create projection instead of using DataFrame::with_column

emcake:
# Description
`DataFrame::with_column` performs a linear operation in the number of columns to append on an existing column, checking that nothing collides. On top of this once the projection a normalization step (also linear in number of columns) is performed before returning the dataframe.

For a merge where we are performing a `when_matched_update_all` type operation on wide tables (100+ columns), this is in effect a `2*N^2` operation as we were adding the remapped case columns one at a time with `with_column` and then remapping it.

This PR uses `project` directly to construct the logical plan. We don't need any of the special checking for name clashes or windowing that `with_column` provides and we discard it immediately down to an unoptimized logical plan anyway, so this produces no change to schema - just a much more compact logical plan.

This reduces an example merge I had from taking 5+ minutes to just optimize the table, down to about 13 seconds including the merge.

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -1223,7 +1223,6 @@ async fn execute(
             Expr::Column(Column::from_qualified_name_ignore_case(name.clone()))
                 .alias(delta_field.name()),
         );
-        //new_columns = new_columns.with_column(&name, case)?;
         new_columns.push((name, case));
     }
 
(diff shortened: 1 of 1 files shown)


Blajda (approved):
LGTM! Thanks for looking into the performance issues :)

ldacey:
Nice - this will be huge. So this should be in the 0.15.4 release then?

PR #2229, opened 2024-02-28
feat: merge schema

### LINKED ISSUE #1386: Schema evolution `mergeSchema` support

aersam:
# Description
It's a draft currently. This replaces the old "overwrite_schema" parameter with a schema_write_mode parameter that basically allows to distinguish between overwrite/merge/none

# Related Issue(s)
Fixes #1386 

# TODO

- [x] Get Partitioning Test through
- [x] Test Overwrite Schema in Append (not very meaningful, I guess, but why not?)
- [x] Test Python

# Documentation

### DIFF
--- a/crates/core/src/operations/cast.rs
+++ b/crates/core/src/operations/cast.rs
@@ -2,11 +2,29 @@
 //!
 use arrow_array::{new_null_array, Array, ArrayRef, RecordBatch, StructArray};
 use arrow_cast::{cast_with_options, CastOptions};
-use arrow_schema::{DataType, Fields, SchemaRef as ArrowSchemaRef};
+use arrow_schema::{DataType, Fields, SchemaRef as ArrowSchemaRef, Schema as ArrowSchema};
 
 use std::sync::Arc;
 
 use crate::DeltaResult;
+
+fn merge_schema(
+    left: ArrowSchemaRef,
+    right: ArrowSchemaRef,
+) -> DeltaResult<ArrowSchemaRef> {
+    let fields = left
+        .fields()
+        .iter()
+        .map(|field| {
+            let right_field = right.field_with_name(field.name());
+            match right_field {
+                Some(right_field) => field.try_merge(right_field)?,
+                None => Ok(field.clone())
+            }
+        })
+        .collect();
+    Ok(ArrowSchemaRef::new(ArrowSchema::new(fields)))
+}
 
 fn cast_struct(
     struct_array: &StructArray,
--- a/crates/core/src/operations/cast.rs
+++ b/crates/core/src/operations/cast.rs
@@ -1,23 +1,23 @@
 //! Provide common cast functionality for callers
 //!
+use arrow::datatypes::DataType::Dictionary;
 use arrow_array::{new_null_array, Array, ArrayRef, RecordBatch, StructArray};
 use arrow_cast::{cast_with_options, CastOptions};
 use arrow_schema::{
     ArrowError, DataType, Field as ArrowField, Fields, Schema as ArrowSchema,
-    SchemaRef as ArrowSchemaRef
+    SchemaRef as ArrowSchemaRef,
 };
-use arrow::datatypes::DataType::Dictionary;
 use std::sync::Arc;
 
 use crate::DeltaResult;
 
-pub (crate) fn merge_field(left: &ArrowField, right: &ArrowField) -> Result<ArrowField, ArrowError> {
-    if let Dictionary(_, value_type) = right.data_type()  {
+pub(crate) fn merge_field(left: &ArrowField, right: &ArrowField) -> Result<ArrowField, ArrowError> {
+    if let Dictionary(_, value_type) = right.data_type() {
         if value_type.equals_datatype(left.data_type()) {
             return Ok(left.clone());
         }
     }
-    if let Dictionary(_, value_type) = left.data_type()  {
+    if let Dictionary(_, value_type) = left.data_type() {
         if value_type.equals_datatype(right.data_type()) {
             return Ok(right.clone());
         }
@@ -30,7 +30,10 @@ pub (crate) fn merge_field(left: &ArrowField, right: &ArrowField) -> Result<Arro
     Ok(new_field)
 }
 
-pub(crate) fn is_compatible_for_merge(schema: ArrowSchema, other: ArrowSchema) -> Result<(), ArrowError> {
+pub(crate) fn is_compatible_for_merge(
+    schema: ArrowSchema,
+    other: ArrowSchema,
+) -> Result<(), ArrowError> {
     for f in schema.fields() {
         if let Ok(other_field) = other.field_with_name(f.name()) {
             if let Err(e) = merge_field(f.as_ref(), other_field) {
@@ -66,7 +69,6 @@ pub(crate) fn merge_schema(
 
     Ok(ArrowSchema::new(fields))
 }
-
 
 fn cast_struct(
     struct_array: &StructArray,
--- a/crates/core/src/operations/cast.rs
+++ b/crates/core/src/operations/cast.rs
@@ -18,15 +18,15 @@ fn cast_struct(
         .iter()
         .map(|field| {
             let col_or_not = struct_array.column_by_name(field.name());
-            if col_or_not.is_none() {
-                if !add_missing {
-                    return Err(arrow_schema::ArrowError::SchemaError(format!(
-                        "Could not find column {0}",
-                        field.name()
-                    )));
-                }
-            }
-            match col_or_not {
+            match col_or_not {                
+                None => 
+                    match add_missing {
+                        true => Ok(new_null_array(field.data_type(), struct_array.len())),
+                        false => Err(arrow_schema::ArrowError::SchemaError(format!(
+                            "Could not find column {0}",
+                            field.name()
+                        ))),
+                    }
                 Some(col) => {
                     if let (DataType::Struct(_), DataType::Struct(child_fields)) =
                         (col.data_type(), field.data_type())
@@ -45,7 +45,6 @@ fn cast_struct(
                         Ok(col.clone())
                     }
                 }
-                None => Ok(new_null_array(field.data_type(), struct_array.len())),
             }
         })
         .collect::<Result<Vec<_>, _>>()
--- a/crates/core/src/operations/cast.rs
+++ b/crates/core/src/operations/cast.rs
@@ -7,7 +7,6 @@ use arrow_schema::{
     ArrowError, DataType, Field as ArrowField, Fields, Schema as ArrowSchema,
     SchemaRef as ArrowSchemaRef,
 };
-use futures::SinkExt;
 use std::sync::Arc;
 
 use crate::DeltaResult;
--- a/crates/core/src/operations/cast.rs
+++ b/crates/core/src/operations/cast.rs
@@ -4,9 +4,9 @@ use arrow_array::{new_null_array, Array, ArrayRef, RecordBatch, StructArray};
 use arrow_cast::{cast_with_options, CastOptions};
 use arrow_schema::{
     ArrowError, DataType, Field as ArrowField, Fields, Schema as ArrowSchema,
-    SchemaRef as ArrowSchemaRef,
+    SchemaRef as ArrowSchemaRef
 };
-
+use arrow::datatypes::DataType::Dictionary;
 use std::sync::Arc;
 
 use crate::DeltaResult;
@@ -21,8 +21,16 @@ pub(crate) fn merge_schema(
         .map(|field| {
             let right_field = right.field_with_name(field.name());
             if let Ok(right_field) = right_field {
+                if let Dictionary(_, value_type) = right_field.data_type()  {
+                    if value_type.equals_datatype(field.data_type()) {
+                        return Ok(field.as_ref().clone());
+                    }
+                }
                 let mut new_field = field.as_ref().clone();
-                new_field.try_merge(right_field)?;
+                let merge_res = new_field.try_merge(right_field);
+                if let Err(e) = merge_res {
+                    return Err(e);
+                }
                 Ok(new_field)
             } else {
                 Ok(field.as_ref().clone())
--- a/crates/core/src/operations/cast.rs
+++ b/crates/core/src/operations/cast.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 
 use crate::DeltaResult;
 
-fn merge_schema(left: ArrowSchemaRef, right: ArrowSchemaRef) -> DeltaResult<ArrowSchemaRef> {
+pub (crate) fn merge_schema(left: ArrowSchemaRef, right: ArrowSchemaRef) -> DeltaResult<ArrowSchemaRef> {
     let fields = left
         .fields()
         .iter()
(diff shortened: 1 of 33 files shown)


crates/core/src/operations/delete.rs
```diff
@@ -167,9 +169,15 @@ async fn excute_non_empty_expr(
...
         false,
-        false,
+        SchemaWriteMode::None,
     )
-    .await?;
+    .await?
+    .into_iter()
+    .map(|a| match a {
+        Action::Add(a) => a,
+        _ => panic!("Expected Add action"),
```
aersam, line 178: am I allowed to do that? :)

ion-elgreco:
I'll take a look later this evening! And thanks for picking this up🤗

crates/core/src/operations/write.rs
```diff
@@ -87,6 +88,33 @@ impl From<WriteError> for DeltaTableError {
...
+///Specifies how to handle schema drifts
+#[derive(PartialEq)]
+pub enum SchemaWriteMode {
+    /// Use existing schema and fail if it does not match the new schema
+    None,
+    /// Overwrite the schema with the new schema
+    Overwrite,
+    /// Append the new schema to the existing schema
+    Merge,
+}
```
rtyler, line 100: There is already some schema evolution code inside of the root writer module, would you be up for consolidating some of this code? It's been on my todo list to bring this write operation and the RecordBatch/JsonWriters into alignment

aersam: When I first looked at the code I asked myself, why there are so many writer implementations? Actually I think there should be only one writer, which is the record_batch writer. The Json Writer should just be producing record batches and use the record batch writer and the writer operation should also use the record batch writer. But maybe I oversee something here and it's probably a bigger undertaking? And also it seems the RecordBatch Writer does not have Overwrite schema yet, right?

A lot simpler would be to use the existing WriteMode enum to not duplicate there. And maybe make a separate file with the schema merge logic which then also handles the partition columns.

What level of consolidating did you had in mind?

crates/core/src/operations/write.rs
```diff
@@ -87,6 +88,33 @@ impl From<WriteError> for DeltaTableError {
     }
 }
 
+///Specifies how to handle schema drifts
+#[derive(PartialEq)]
+pub enum SchemaWriteMode {
+    /// Use existing schema and fail if it does not match the new schema
+    None,
```
ion-elgreco, line 95: Why not allow SchemaWriteMode to be optional instead

crates/core/src/operations/write.rs
```diff
@@ -391,8 +452,8 @@ pub(crate) async fn write_execution_plan(
     write_batch_size: Option<usize>,
     writer_properties: Option<WriterProperties>,
     safe_cast: bool,
-    overwrite_schema: bool,
-) -> DeltaResult<Vec<Add>> {
+    schema_write_mode: SchemaWriteMode,
```
ion-elgreco, line 455: So we get Option<SchemaWriteMode> here

python/deltalake/_internal.pyi
```diff
@@ -174,7 +174,7 @@ def write_to_deltalake(
     partition_by: Optional[List[str]],
     mode: str,
     max_rows_per_group: int,
-    overwrite_schema: bool,
+    schema_write_mode: Optional[str],
```
ion-elgreco, line 177: Maybe rename it to schema_mode, the mode is also just mod and not write_mode

python/deltalake/writer.py
```diff
@@ -185,6 +185,7 @@ def write_deltalake(
     description: Optional[str] = None,
     configuration: Optional[Mapping[str, Optional[str]]] = None,
     overwrite_schema: bool = False,
+    schema_write_mode: Literal["none", "merge", "overwrite"] = "none",
```
ion-elgreco, line 188: The None should be a python NoneType.

crates/core/src/operations/write.rs
aersam: @rtyler I just tried what happens if I delete writer.rs and try to replace it with RecordBatchWriter in this PR: https://github.com/delta-io/delta-rs/pull/2232 It seems to be possible. Would you generally prefer that approach?

aersam:
@ion-elgreco Thanks for your feedback, I've done those changes. However for me the next step is to elaborate if this approach is the one to take or go for the more radical approach in #2232 where I just completely dropped writer.rs in favor of RecordBatchWriter. What do you think? #2232 is more of an experiment yet, but basically it seems possible

ion-elgreco:
@aersam maybe take the writer.rs as the default template, extend it with the merge functionality and move that then into RecordBatchWriter? It seems to be the newest writer, so I would assume there are some changes in there that are improvments over the RecordBatchWriter

aersam:
> @aersam maybe take the writer.rs as the default template, extend it with the merge functionality and move that then into RecordBatchWriter? It seems to be the newest writer, so I would assume there are some changes in there that are improvments over the RecordBatchWriter

Yes, seems like. Better would be to remove RecordBatchWriter, but this seems to be a public Rust API, so would be breaking. Will continue this here

rtyler:
The RecordBatchWriter being public doesn't mean we have to keep it, but we should keep some form of taking record batches and being able to provider a writer. The RecordBatchWriter is used in kafka-delta-ingest and a number of other projects, but I think it _could_ be unified in one direction or another with the writer in operations/

The most important characteristics is writing RecordBatches with pure Arrow and Parquet, and not relying on datafusion since that comes into play with the write operation

ion-elgreco:
@rtyler should we maybe then not unify? The next improvement for the write operation would be using logical plans with Datafusion, so maybe with that mind it's not needed to unify them

aersam:
I think we should move that discussion to a separate issue/discussion 🙂

Personally I also don't like that the Python Code depends on DataFusion. But probably not easy to do without, as for check constraints and filters there is need for some sql parser. But I'd much prefer something based on substrait there.

For me, in a perfect world, any datafusion stuff would even have it's own crate. Compilation would also be faster, I guess

ion-elgreco:
> I think we should move that discussion to a separate issue/discussion 🙂
> 
> Personally I also don't like that the Python Code depends on DataFusion. But probably not easy to do without, as for check constraints and filters there is need for some sql parser. But I'd much prefer something based on substrait there.
> 
> For me, in a perfect world, any datafusion stuff would even have it's own crate. Compilation would also be faster, I guess

Well you need an execution engine to do most of the operations

ion-elgreco:
@aersam your venv is perhaps not using the latest version of ruff etc

aersam:
> @aersam your venv is perhaps not using the latest version of ruff etc

I think the failures are in Code I did not touch. I could commit these fixes, but they have nothing to do with my changes

ion-elgreco:
@aersam yeah that's fine, I think we don't have a version pinned here, which we probably should. Maybe you could make a small separate PR for that, to pin all the linters, type checker versions?

aersam:
> @aersam yeah that's fine, I think we don't have a version pinned here, which we probably should. Maybe you could make a small separate PR for that, to pin all the linters, type checker versions?

Here we go: https://github.com/delta-io/delta-rs/pull/2240

aersam:
I'll have a look into the test failures

aersam:
@ion-elgreco Tests should be good now, and I also some minimal docs.

aersam:
> @aersam great! I'll do a thorough review later tonight after work

Good! Just did another push that should also make the pyarrow 8 tests go through, sorry :)

crates/core/src/operations/cast.rs
```diff
@@ -1,36 +1,106 @@
...
+    SchemaRef as ArrowSchemaRef,
+};
 use std::sync::Arc;
 
 use crate::DeltaResult;
 
+pub(crate) fn merge_field(left: &ArrowField, right: &ArrowField) -> Result<ArrowField, ArrowError> {
+    if let Dictionary(_, value_type) = right.data_type() {
+        if value_type.equals_datatype(left.data_type()) {
+            return Ok(left.clone());
```
ion-elgreco, line 17: That clone is not necessary I think

crates/core/src/operations/cast.rs
```diff
@@ -1,36 +1,106 @@
...
-
+use arrow_schema::{
+    ArrowError, DataType, Field as ArrowField, Fields, Schema as ArrowSchema,
+    SchemaRef as ArrowSchemaRef,
+};
 use std::sync::Arc;
 
 use crate::DeltaResult;
 
+pub(crate) fn merge_field(left: &ArrowField, right: &ArrowField) -> Result<ArrowField, ArrowError> {
```
ion-elgreco, line 14: Why did you opt for creating your own function instead of the try_merge in arrow directly?

python/tests/test_writer.py
```diff
@@ -136,47 +136,156 @@ def test_update_schema(existing_table: DeltaTable):
...
+    old_table_data = existing_table.to_pyarrow_table()
+    new_data = pa.table(
+        {
+            "utf8": pa.array(["bla", "bli", "blubb"]),
+            "new_x": pa.array([1, 2, 3], pa.int32()),
+            "new_y": pa.array([1, 2, 3], pa.int32()),
+        }
+    )
+
+    write_deltalake(
```
ion-elgreco, line 201: This should not be possible; schema mode overwrite should only be working when you do overwrite.

docs/usage/writing/index.md
```diff
@@ -23,7 +23,9 @@ of Spark's `pyspark.sql.DataFrameWriter.saveAsTable` DataFrame method. To overwr
 
 `write_deltalake` will raise `ValueError` if the schema of the data
 passed to it differs from the existing table's schema. If you wish to
-alter the schema as part of an overwrite pass in `overwrite_schema=True`.
+alter the schema as part of an overwrite pass in `schema_mode="overwrite"` or `schema_mode="merge"`.
```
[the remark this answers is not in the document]
aersam: Well... I changed that :) I allowed overwrite in append as well, which then just drops columns. I think that's a valid use case, but surely a rather exotic one.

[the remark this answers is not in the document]
aersam: True, then I'll remove that "feature". First should then come up with those operations later on

crates/core/src/operations/write.rs
```diff
@@ -594,17 +659,54 @@ impl std::future::IntoFuture for WriteBuilder {
...
+                || (this.schema_mode == Some(SchemaMode::Overwrite)
+                    && this.mode != SaveMode::Overwrite)
+            {
+                if let Some(snapshot) = &this.snapshot {
+                    let table_schema = snapshot
+                        .physical_arrow_schema(this.log_store.object_store().clone())
+                        .await
+                        .or_else(|_| snapshot.arrow_schema())
+                        .unwrap_or(schema.clone());
+                    if !can_cast_batch(schema.fields(), table_schema.fields()) {
```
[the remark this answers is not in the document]
aersam: I'll save the result from the check above so we don't do it twice

crates/core/src/operations/cast.rs
aersam: I'll try getting rid of it, but will require explicit Lifetimes, I'll try

crates/core/src/operations/cast.rs
aersam: Because of the Dictionary Type, which is how I get partitioned values. I need to be able to allow DictionaryType<_, value_type> == value_type

docs/usage/writing/index.md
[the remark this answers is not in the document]
ion-elgreco: We can definitely add that operation soon, shouldn't be much work since it's mostly a metadata driven operation

crates/core/src/operations/cast.rs
aersam: Ok, I gave up. the problem is that try_merge of arrow requires a &mut to be passed, so a clone is required there anyway. and then you create a reference to something owned in the function... the alternative would be passing around Arc's, but that also is not nice since right is a reference anyway. Also we use it to create a schema instance which required the data to be owned anyway, so cloning is required there, too.

crates/core/src/operations/write.rs
```diff
@@ -548,37 +582,68 @@ impl std::future::IntoFuture for WriteBuilder {
...
+                                new_schema = None // we overwrite anyway, so no need to cast
+                            } else if this.schema_mode == Some(SchemaMode::Merge) {
+                                new_schema = Some(Arc::new(merge_schema(
+                                    table_schema.as_ref().clone(),
+                                    schema.as_ref().clone(),
+                                )?));
+                            } else {
+                                // this is a feature! Unless you specify a schema_mode explicity, we want to check the schema!
+                                return Err(DeltaTableError::Generic(
+                                    "Schema of data does not match table schema".to_string(),
```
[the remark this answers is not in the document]
aersam: done :)

python/tests/test_writer.py
```diff
@@ -124,59 +124,159 @@ def test_enforce_schema(existing_table: DeltaTable, mode: str):
...
         write_deltalake(
             existing_table,
-            new_data,
+            pa.table({"x1": pa.array([1, 2, 3])}),
             mode="append",
-            overwrite_schema=False,
+            schema_mode="overwrite",
+            engine="rust",
+        )
+    with pytest.raises(ValueError):
```
ion-elgreco, line 247: Can you make the test a bit more explicit by also catching the message or in a seperate test. 

Because this fails because UTF8 column can't be merged since LHS and RHS have different dtypes

python/tests/test_writer.py
```diff
@@ -124,59 +124,159 @@ def test_enforce_schema(existing_table: DeltaTable, mode: str):
...
         )
+
+    new_data = pa.table(
+        {
+            "utf8": pa.array(["bla", "bli", "blubb"]),
+            "new_x": pa.array([1, 2, 3], pa.int32()),
+            "new_y": pa.array([1, 2, 3], pa.int32()),
+        }
+    )
     with pytest.raises(DeltaError):
```
ion-elgreco, line 215: Maybe its because I just woke up but it's not clear when it will still throw a deltaerror over a value error because of incorrect schema

ion-elgreco (changes requested):
Just some small comments on tests but after that we are good to go

aersam:
I dont think the failure is related to my changes?

ion-elgreco:
> > @aersam it's not, not sure what's going here maybe credentials expired?
> 
> I'd suspect an update of object_store that was breaking in terms of the azure emulator. Suspect commit: [apache/arrow-rs@0fd9b8a](https://github.com/apache/arrow-rs/commit/0fd9b8af8f4dd090680b149e8607474b42effa19) .
> 
> I cannot make that test run locally, but I'll try to fix object_store to use 0.9.0 and retry

Good catch, I was already wondering, how did anything change there

ion-elgreco (approved):
Great addition :) I think many folks will be happy this feature has landed! Approved from my side

Let's have @rtyler also gives his blessing before merging🙏

crates/core/src/operations/write.rs
aersam: After having tried some stuff I came to conclusion that this is a bigger project I'm not sure I'm the right one to do, since I don't really see the use case for RecordBatchWriter and how the connection of writer.rs to DataFusion is an issue. I'd prefer having this merged first and then discuss further code deduplication

rtyler (approved):
I am happy to see this come in. I agree it's a non-trivial effort to consolidate our writer support. The write module here is just growing to be excessive and unmaintainable unfortunately. ` prepare_predicate_actions()` as an example is now almost 400 lines of deeply important logic smashed together which does concern me, but that's not something new :smile:

rtyler:
I am rebasing and will squash this for a merge

PR #2232, opened 2024-02-29
feat: merge schema by using record batch writer

### LINKED ISSUE #1386: Schema evolution `mergeSchema` support

aersam:
# Description
This is a counter proposal to my own PR here: https://github.com/delta-io/delta-rs/pull/2229 

The Idea is mainly to get completely rid of writer.rs and use RecordBatchWriter instead. I can hardly tell much about the side-effects, but so far it compiles :)

# Related Issue(s)
Fixes #1386  (once completed, which it is NOT)

# Documentation


aersam:
Closing, since RecordBatchWriter seems to be older than writer.rs. I guess the better way would be replacing RecordBatchWriter, but this is a public Rust API, so probably not easy

Issue #1777, opened 2023-10-26
We cannot append data to existing Delta Lake tables if the schema of data to write includes timestamp columns with timezone.


sugibuchi:
# Environment
* Python 3.10.12
* PyArrrow: 12.0.0

**Delta-rs version**:
* 0.12.0

**Binding**: 
* Python

**Environment**:
- **Cloud provider**: Microsoft Azure
- **OS**: Ubuntu 20.04.6 LTS (Focal Fossa)
- **Other**: JupyterLab Version 4.0.7

***
# Bug

We cannot append data to existing Delta Lake tables if the schema of data to write includes timestamp columns with timezone.

**What happened**:
The first write succeeds. But subsequent append writes fail.

**What you expected to happen**:

We can append data including timestamp columns with timezone in its schema.

**How to reproduce it**:

```python
from datetime import datetime, timezone

import pyarrow as pa
from deltalake import write_deltalake

schema_with_tz = pa.schema([pa.field("ts", pa.timestamp(unit="us", tz=timezone.utc))]) 

table_with_tz1 = pa.table([[datetime(2023,1,1,1,1,1,tzinfo=timezone.utc)]], schema_with_tz)
table_with_tz2 = pa.table([[datetime(2023,2,2,2,2,2,tzinfo=timezone.utc)]], schema_with_tz)

write_deltalake("with_tz.delta", table_with_tz1, mode="append") # Success
write_deltalake("with_tz.delta", table_with_tz2, mode="append") # Fail!
```

`pa.timestamp(unit="us", tz=timezone.utc)` looks compliant with the timestamp data type in Delta Lake.

> Microsecond precision timestamp elapsed since the Unix epoch, 1970-01-01 00:00:00 UTC.
> https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types

But the second `write_deltalake(..., mode="append")` fails with the following error.

```
File /opt/conda/lib/python3.10/site-packages/deltalake/writer.py:185, in write_deltalake(table_or_uri, data, schema, partition_by, filesystem, mode, file_options, max_partitions, max_open_files, max_rows_per_file, min_rows_per_group, max_rows_per_group, name, description, configuration, overwrite_schema, storage_options, partition_filters, large_dtypes)
    181 if table:  # already exists
    182     if schema != table.schema().to_pyarrow(as_large_types=large_dtypes) and not (
    183         mode == "overwrite" and overwrite_schema
    184     ):
--> 185         raise ValueError(
[...]

Issue #1019, opened 2022-12-15
Cannot create Delta Lake table from DataFusion data frame with timestamp field with timezone


andrei-ionescu:
# Environment

**0.5.0**:

**Binding**: 
None. Rust.

**Environment**:
- **Cloud provider**: N/A
- **OS**: MacOS Ventura
- **Other**: N/A

***
# Bug

**What happened**:

When trying to create a Delta Lake table based on an Arrow Schema that contains a field of type timestamp with timezone the following error occurs:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SchemaError("Invalid data type for Delta Lake: Timestamp(Microsecond, Some(\"UTC\"))")', src/main.rs:220:86
```

**What you expected to happen**:

Being able to create a Delta Lake table with a timezone timestamp type.

**How to reproduce it**:

1. Create an Arrow Schema `arrow_schema` with a field of timestamp type with UTC timezone.
2. Do the following:

```
let delta_schema = deltalake::Schema::try_from(Arc::new(arrow_schema)).unwrap();
```

**More details**:

The issue seems to be found in the converter from `ArrowDataType` into `SchemaDataType` found here: https://github.com/delta-io/delta-rs/blob/main/rust/src/delta_arrow.rs#L218.

There is no case when the Arrow Data Type is `ArrowDataType::Timestamp(TimeUnit::Microsecond, Some(tz))`.

It seems that the Delta Lake format does NOT support timestamp with timezone — it needs to be transformed to UTC first: 

>Notes
>
>Timestamps with local timezone are internally normalized and persisted in UTC. Whenever the value or a portion of it is extracted the local session timezone is applied.

Quote from https://docs.databricks.com/sql/language-manual/data-types/timestamp-type.html

**Possible fixes**:

1. **Easy fix for UTC timestamp types** — include the timestamp type having the **UTC** timezone into the current pattern and transform it into the `timestamp` primitive. This will fix the exact problem described above, specific to UTC.
4. **Permanent fix** — transform the timezoned timestamp into UTC time but this requires extensive changes into other areas.

ion-elgreco:
@rtyler this issue has already been resolved, can you close it?

PR #2236, opened 2024-03-03
feat(python, rust): timestampNtz support

ion-elgreco:
# Description

- This addresses all our timestamp inconsistencies, where we were reading Primitive:timestamp as a datatetime without UTC, and now we can properly write datetimes with no timezone as columns to Primitive::timestampNtz.
- addressing small bug where checkConstraints feature was not set in writerFeatures when you are on table writer version 7.
- bumping default protocol to 3,7
- Made the pyarrow writer and reader more flexible so we can write/read a 3,7 table as long as it has the supported features there.
- Properly parses timestamps with UTC into pyarrow timestamps with UTC
- Added configkey translation to tablefeature inside the Create Operation

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1598
- closes https://github.com/delta-io/delta-rs/issues/1019
- closes https://github.com/delta-io/delta-rs/issues/1777


crates/core/src/operations/transaction/protocol.rs
```diff
@@ -61,11 +61,11 @@ impl ProtocolChecker {
     }
 
     pub fn default_reader_version(&self) -> i32 {
-        1
+        3
     }
 
     pub fn default_writer_version(&self) -> i32 {
-        2
+        7
```
rtyler, line 68: This is a non-trivial change in behavior isn't it? I don't think timezone support gets us 3/7

ion-elgreco: Correct, but it smoothes out our table creation versioning since feature toggling in 3,7 is more flexible

ion-elgreco: Hmm, what's your worry? 

I don't see the issue of automatically enabling a feature during create based on the source data. It's something we can easily infer, if it contains a timezone, then enable the feature and bump the protocol versions.

What I can do, is undo the change of the default protocol version back to 1,2. However I still believe it's more convenient if it's by default on 3,7 especially since we started supporting 3,7 so that we could pick and choose on which table features we added support instead of going linear from 2 to 7.

crates/core/src/kernel/expressions/scalars.rs
```diff
@@ -433,6 +445,16 @@ impl PrimitiveType {
                     .ok_or(self.parse_error(raw))?;
                 Ok(Scalar::Timestamp(micros))
             }
+            TimestampNtz => {
+                let timestamp = NaiveDateTime::parse_from_str(raw, "%Y-%m-%d %H:%M:%S%.f")
```
hntd187, line 449: So the rough part of this is that the format for the timestamp can be either `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456`, I don't believe this format would cover both right?

crates/core/src/kernel/models/schema.rs
```diff
@@ -554,6 +556,7 @@ impl Display for PrimitiveType {
             PrimitiveType::Binary => write!(f, "binary"),
             PrimitiveType::Date => write!(f, "date"),
             PrimitiveType::Timestamp => write!(f, "timestamp"),
+            PrimitiveType::TimestampNtz => write!(f, "timestampNTZ"),
```
hntd187, line 559: nit: NTZ -> Ntz

crates/core/src/operations/create.rs
```diff
@@ -233,8 +237,36 @@ impl CreateBuilder {
...
+            .filter(|v| match v {
+                ReaderFeatures::Other(_) => false,
+                _ => true,
+            })
+            .collect::<HashSet<ReaderFeatures>>();
+
+        let contains_timestampntz = &self
+            .columns
+            .iter()
+            .any(|f| f.data_type() == &DataType::TIMESTAMPNTZ);
```
hntd187, line 263: Should the enum variant be all caps here?

crates/core/src/kernel/expressions/scalars.rs
ion-elgreco: Not sure, I saw the mention in the protocol on both formats. But timestamp was only parsed with the ms precision. Maybe @roeap can chip in here?

crates/core/src/kernel/models/schema.rs
ion-elgreco: I am not sure anymore for what exactly, but I had to do this on purpose

hntd187:
Also delta-scala doesn't universally cast things to UTC, it respects the `isAdjustedToUTC` param for timestamps, where timestampntz has this set to false, while normal timestamps are set to true

crates/core/src/kernel/expressions/scalars.rs
hntd187: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#timestamp-without-timezone-timestampNtz it talks about supporting both formats for this datatype here

crates/core/src/operations/create.rs
ion-elgreco: This points to the pub const on DataType:

`pub const TIMESTAMPNTZ: Self = DataType::Primitive(PrimitiveType::TimestampNtz);`

hntd187: missed this, never mind :-D

crates/core/src/kernel/expressions/scalars.rs
ion-elgreco: I checked, it has no problem parsing both formats with "%Y-%m-%d %H:%M:%S%.f", both can be parsed into a timestamp

crates/core/src/operations/transaction/protocol.rs
hntd187: because doing this breaks all readers/writers in the wild that aren't 3/7, it breaks prior versions of delta-spark as well, even if you never wrote a single row to the table because it requires readers and writers to meet that bar to even read the table. I'm saying, we should only update the protocol if the features are explicitly enabled or explicitly used. For example, databricks enables deletion vectors by default which makes basically the default protocol 3,7, but if you disable that in the table create it sets the protocol to 1,2. The protocol versions should be the absolute minimum required to read/write the table, we should never just update them if there isn't a feature that requires it.

crates/core/src/kernel/models/schema.rs
ion-elgreco: Was not needed after all xD

crates/core/src/operations/transaction/protocol.rs
ion-elgreco: Gotcha, I see your point now, that will definitely break backwards compatibility indeed with older databricks runtimes. 

Then I'll change back the default to 1,2 but bump it only  during a protocol commit when we see a feature that is only on 3,7.

hntd187: Excellent, so when you create a table with timestamp_ntz, it will create a table with 3,7 as the versions and timestampNtz in the table_features

ion-elgreco: I have undone the changes :) So it's now only done while writing a table with no timezone.

crates/core/src/kernel/models/actions.rs
```diff
@@ -281,15 +283,15 @@ impl From<String> for WriterFeatures {
...
     fn from(value: &str) -> Self {
         match value {
-            "appendOnly" => WriterFeatures::AppendOnly,
-            "invariants" => WriterFeatures::Invariants,
-            "checkConstraints" => WriterFeatures::CheckConstraints,
-            "changeDataFeed" => WriterFeatures::ChangeDataFeed,
+            "appendOnly" | "delta.appendOnly" => WriterFeatures::AppendOnly,
+            "invariants" | "delta.invariants" => WriterFeatures::Invariants,
+            "checkConstraints" | "delta.checkConstraints" => WriterFeatures::CheckConstraints,
+            "changeDataFeed" | "delta.enableChangeDataFeed" => WriterFeatures::ChangeDataFeed,
```
rtyler, line 289: @ion-elgreco why are these changes required? Are we not picking up table properties correctly right now? If so, is there a bug this is related to?

ion-elgreco: @rtyler I am doing that so I can go from tableConfigKeys into tableFeatures. This is what spark-delta also does if your table protocol versions are 3,7 and you do set tblproperties "delta.enableChangeDataFeed" = true, it will additionally set the feature ChangeDataFeed into the protocol commit.

crates/core/src/kernel/expressions/scalars.rs
```diff
@@ -222,10 +225,17 @@ impl Scalar {
                 .downcast_ref::<Date32Array>()
                 .map(|v| Self::Date(v.value(index))),
             // TODO handle timezones when implementing timestamp ntz feature.
-            Timestamp(TimeUnit::Microsecond, None) => arr
-                .as_any()
-                .downcast_ref::<TimestampMicrosecondArray>()
-                .map(|v| Self::Timestamp(v.value(index))),
+            Timestamp(TimeUnit::Microsecond, tz) => match tz {
```
guilhem-dvr, line 228: nit: the CL addresses the `TODO` no?

ion-elgreco: Right, forgot to remove that : )

rtyler (approved):

crates/core/src/kernel/models/actions.rs
roeap: hmm, just passing by, but this seems strange. these are the values written in the feature vector list, so this seems like a strange place to do this conversion as well, isnce it describes completely differnet fields, now? one is values in the feature vecotor field, andother one is keys in the table properties, which should be handled in the table config?

crates/core/src/kernel/models/schema.rs
```diff
@@ -468,7 +468,7 @@ fn default_true() -> bool {
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Eq, Hash)]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "snake_case")]
```
roeap, line 471: are we sure this is right?

ion-elgreco: Without this it would fail for the timestampNtz, since spark wrote it as timestamp_ntz in the log. I couldnt find his anywhere in the docs, but it only impacted timestampNtz since the rest of them are one word long

crates/core/src/kernel/models/actions.rs
ion-elgreco: I added the table config keys here as well for quick parsing to table features, see my comment above.

In spark when your protocol is at 3,7 it will auto convert the configs into table features so that's basically what I am doing here as well.

PR #2240, opened 2024-03-04
fix: fix ruff and mypy version and do formatting

aersam:
# Description
As requested by @ion-elgreco in #2229 , we should fix the formatter versions

# Related Issue(s)

# Documentation


ion-elgreco:
Thankss :D

ion-elgreco (approved):

PR #2257, opened 2024-03-06
feat(python): release GIL in the write_deltalake function

### LINKED ISSUE #2234: Release GIL in deltalake.write_deltalake
I'm running `deltalake.write_deltalake` many times in parallel and noticing that my Python process is freezing up a bit.  I suspect that this function has a long period where it doesn't release the GIL.  Is this true?  

If so, I suspect that it's likely accidental, and could be easily changed.  I'm not familiar with the rust-Python tooling, but most Python binding systems make this pretty easy.

franz101:
# Description
Release GIL in deltalake.write_deltalake by wrapping it in py.allow_threads
# Related Issue(s)
- closes #2234
# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -1401,60 +1401,62 @@ fn write_to_deltalake(
     writer_properties: Option<HashMap<String, Option<String>>>,
     custom_metadata: Option<HashMap<String, String>>,
 ) -> PyResult<()> {
-    let batches = data.0.map(|batch| batch.unwrap()).collect::<Vec<_>>();
-    let save_mode = mode.parse().map_err(PythonError::from)?;
+    py.allow_threads(|| {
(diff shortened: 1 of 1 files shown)


mrocklin:
Cool!  If this works well I suspect that the same pattern would also be useful on other common operations (compact, vacuum, etc..)

ion-elgreco:
@franz101 you're missing `py: Python,` as the first variable of the function defintion

franz101:
> Cool! If this works well I suspect that the same pattern would also be useful on other common operations (compact, vacuum, etc..)
Essentially all methods [here](https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.update):
- `create`
- `delete`
- `optimize`
- `vacuum`
- ...

mrocklin:
Looks good to me :)

PR #2260, opened 2024-03-07
fix(#2256): use consistent units of time

### LINKED ISSUE #2256: Inconsistent units of time
# Environment

**Delta-rs version**: latest

**Binding**: rust

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

**What happened**:

Sometimes duration fields suffixed with `_ms` report time as milliseconds (expected), but other times they are reported as microseconds. 

https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/update.rs#L400

https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/delete.rs#L204

All fields using `_ms` should be in milliseconds. 

`_ms` -> milliseconds
`_us` -> microseconds

**How to reproduce it**:

**More details**:

Microsecond granularity is likely not needed for any operations that traverse the network.

cmackenzie1:
# Description

Update timing fields using the suffix `_ms` to be in milliseconds.

Doing a quick grep through the repo I only found a few instances of this in the delete.rs file.

# Related Issue(s)

Closes #2256

# Documentation

### DIFF
--- a/crates/core/src/operations/delete.rs
+++ b/crates/core/src/operations/delete.rs
@@ -201,7 +201,7 @@ async fn execute(
 
     let scan_start = Instant::now();
     let candidates = find_files(snapshot, log_store.clone(), &state, predicate.clone()).await?;
-    metrics.scan_time_ms = Instant::now().duration_since(scan_start).as_micros();
+    metrics.scan_time_ms = Instant::now().duration_since(scan_start).as_millis();
 
     let predicate = predicate.unwrap_or(Expr::Literal(ScalarValue::Boolean(Some(true))));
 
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@cmackenzie1 the metrics are optional anyways but just to get some consistency across other implementations, do you know if spark uses milli or micro precision?

cmackenzie1:
@ion-elgreco looks like Milliseconds https://github.com/delta-io/delta/blob/2fddb8b0b87b08ec57b13e86fcd2feec83ea10f1/connectors/standalone/src/main/java/io/delta/standalone/Operation.java#L178

ion-elgreco (approved):

PR #2263, opened 2024-03-08
docs: use dagster deltalake polars library

avriiil:
use `dagster-deltalake-polars ` instead of `dagster-polars`


avriiil:
@ion-elgreco - polars example updated here. thanks for your input!

ion-elgreco (approved):

Issue #2247, opened 2024-03-05
Property setting in `create` is not handled correctly


ion-elgreco:
# Environment
Currently our property (table configuration) setting during the create action is handled incorrectly. There are some cases where we need to do additional actions such as bumping the protocol version, which we aren't doing.  @rtyler @roeap 

Examples:

- Settting "delta.minWriterVersion = '5'" should bump writer protocol to version 5 during the commit with protocol action.
- Setting "delta.enableChangeDataFeed = true" should bump it to the minim required protocol version to read and write CDF.

There are a couple more cases, my plan is to address this while working on the SET TBLPROPERTIES since the behavior of this should be common. 
The common part will be  `current protocol + TableConfig -> Protocol + Metadata Action`. The current protocol will be defined after things like knowing if table contains TimestampNtz, or if operation is calling SET TBLPROPERTIES grabbing the latest protocol of the existing table.

Issue #1663, opened 2023-09-24
Add `set_table_properties` to DeltaTable class 


ion-elgreco:
# Description

**Use Case**
Currently it's not possible to alter the table_properties after a write (or TableConfig as it's referred to in code.) So now you need Spark-delta to alter these properties.

MrPowers:
@ion-elgreco - can you provide more details on the specific table properties that you're referring to here?

ion-elgreco:
> @ion-elgreco - can you provide more details on the specific table properties that you're referring to here?

These ones:
https://books.japila.pl/delta-lake-internals/DeltaConfigs/#table-properties
And  https://docs.delta.io/table-properties.html

srsharathreddy-92:
it would be great if we can change logRetentionDuration property after first write.

pedrosalgadowork:
Any update on this, so we can also change it from python?

dtheodor:
Does this issue cover adding support for DDL statements in general, such as [CREATE TABLE](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-table-using.html) and [ALTER TABLE ...](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-alter-table.html)? Currently only possible with spark.

PR #2264, opened 2024-03-08
feat(python, rust): add `set table properties` operation

ion-elgreco:
# Description
- Adds `set table properties` operation
- Some logic is shared for the create operation

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1663
- fixes https://github.com/delta-io/delta-rs/issues/2247


echai58:
Hi, does this PR allow for setting of custom metadata too? On table creation we're able to store custom metadata via the configuration parameter, so would be great if we could set custom metadata via this operation as well.

ion-elgreco:
@echai58 that's the idea yes, you would be able to change the table configuration.

pedrosalgadowork:
Any idea when this PR will be merged?

crates/core/src/operations/set_tbl_properties.rs
```diff
@@ -0,0 +1,311 @@
...
+            let commit = CommitBuilder::from(this.commit_properties)
+                .with_actions(actions.clone())
+                .build(
+                    Some(&this.snapshot),
+                    this.log_store.clone(),
+                    operation.clone(),
+                )?
+                .await?;
+
+            this.snapshot.merge(actions, &operation, commit.version)?;
```
rtyler, line 307: @ion-elgreco I had to modify this to compile with the latest `main`. I feel uneasy about merging state rather than just reloading from storage but up to you :shrug:

ion-elgreco: Ah yes this piece is outdated now, with the changes I made in the post commit. We are advancing the state there already so the commit already returns the fully updated snapshot :p

PR #2268, opened 2024-03-08
fix: add .venv to .gitignore

gacharya:
# Description
In `CONTRIBUTING.md`, we recommend that the developer uses a python virtual env. 

```
create a Python virtual environment (required for development builds), e.g. as described [here](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/)
```

However, `.venv` is not ignored in the `.gitignore`. 

# Related Issue(s)

# Documentation


ion-elgreco:
@gacharya you can use the make commands from the python directory to setup your venv

rtyler (approved):

PR #2269, opened 2024-03-09
feat(python): release GIL for common functions

franz101:
# Description
Inspired by the previous request to release the GIL added to most common functions.
# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/2234

If some function are not supported / should not be released or I forgot to add the function, happy to modify the PR

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -288,7 +288,7 @@ impl RawDeltaTable {
         enforce_retention_duration: bool,
         custom_metadata: Option<HashMap<String, String>>,
     ) -> PyResult<Vec<String>> {
-        py.allow_threads(|| {
+        let cmd = py.allow_threads(|| {
             let mut cmd = VacuumBuilder::new(
                 self._table.log_store(),
                 self._table.snapshot().map_err(PythonError::from)?.clone(),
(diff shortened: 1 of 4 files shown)


franz101:
ah something messed up my formatter (`make format`)

mrocklin:
I'm excited for this.  Thanks for working on it @franz101 !  Looking briefly it looks like the linter is sad.

python/src/lib.rs
```diff
@@ -282,75 +282,81 @@ impl RawDeltaTable {
...
 
-        let (table, metrics) = rt()?
-            .block_on(cmd.into_future())
-            .map_err(PythonError::from)?;
-        self._table.state = table.state;
-        Ok(metrics.files_deleted)
+            let (table, metrics) = rt()?
+                .block_on(cmd.into_future())
+                .map_err(PythonError::from)?;
+            self._table.state = table.state;
```
wjones127, line 311: Could we move updating the table state outside of the `allow_threads` call? this requires mutating `self`, which should be done while holding the GIL.

Same thing applies to other methods.

adriangb:
For the cases when you need the GIL, could you also release it and re-acquire it when needed, if the fraction of time it's not needed is large?

Issue #1470, opened 2023-06-16
Can't write data from parquet file to delta table (Rust)


cjingithub:
# Environment

**Delta-rs version**:  0.12

**Binding**: Rust

**Environment**:
- **OS**: macOS Monterey 12.6.2
- **Other**: polars = "0.30", tokio = "1", tokio-stream = "0.1"

***
# Bug

**What happened**: 
Cannot write data from parquet file to delta table 

> Error: Generic("Arrow RecordBatch schema does not match: RecordBatch schema: Field { name: \"id\", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: \"id\", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }")

**What you expected to happen**:
Write data from parquet file to delta table with no issues

**How to reproduce it**:
```rust
use std::collections::HashMap;
use std::error::Error;

use deltalake::writer::{RecordBatchWriter, DeltaWriter};
use deltalake::{SchemaField, SchemaDataType, DeltaOps};
use deltalake::parquet::arrow::ParquetRecordBatchStreamBuilder;
use polars::prelude::*;
use tokio::fs::File;
use tokio::io::AsyncReadExt;
use tokio_stream::StreamExt;

#[derive(Debug)]
struct Test;
impl Test {
    fn columns() -> Vec<SchemaField> {
        vec![
            SchemaField::new(
                "id".to_string(),
                SchemaDataType::primitive("string".to_string()), 
                true,
                HashMap::new(),
            ),
        ]
    }
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    // create parquet file with polars
    let path = "/path/to/parquet/df.parquet";
    create_bad_df(path)?;

    // read parquet file
    let data = read_file(path).await?;
    let cursor = std::io::Cursor::new(data);
    let builder = ParquetRecordBatchStreamBuilder::new(cursor).await?;
    let mut stream = builder.build()?;
    let mut records = Vec::new();
    while let Some(record) = stream.next().await {
        let record = record?;
        records.push(record);
    }

    // create delta table 
    let delta_path = "/path/to/delta/";
    let ops = DeltaOps::try_from_uri(delta_path).await?;
    let mut table = ops
        .create()
        .with_columns(Test::columns())
        .await?;

    // writing to delta
    let mut writer = RecordBatchWriter::for_table(&table)?;
[...]

PR #2274, opened 2024-03-09
fix(python, rust): prevent table scan returning large arrow dtypes

ion-elgreco:
# Description
Such a small change, but fixes many issues where parquets were written with arrow where the source data was in large dtype  format.

By default the parquet::ParquetReader decodes the arrow metadata which in return may give you large dtypes. This would cause issues during DataFusion parquet scan with a filter since the filter wouldn't coerce to the large dtypes. Simply disabling the arrow metadata decoding gives us the parquet schema converted to an arrow schema without large types 👯‍♂️ 

# Related issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1470

### DIFF
--- a/crates/core/src/operations/transaction/state.rs
+++ b/crates/core/src/operations/transaction/state.rs
@@ -11,6 +11,7 @@ use datafusion_common::Column;
 use datafusion_expr::Expr;
 use itertools::Itertools;
 use object_store::ObjectStore;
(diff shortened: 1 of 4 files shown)


Blajda (approved):
LGTM!
Based on the documentation for this function I don't foresee any issue.

Issue #2262, opened 2024-03-07
Generic DeltaTable error: Version mismatch with new schema merge functionality in AWS S3


liamphmurphy:
# Environment

**Delta-rs version**: python v0.16

**Binding**: ^^

**Environment**:
- **Cloud provider**: AWS s3 with dynamo

***
# Bug

**What happened**:

To test the rust engine, we cleared out any existing delta tables in our nonprod environment and switched from pyarrow over to the rust engine with schema merging, with this `write_deltalake` call:
```
 write_deltalake(s3_path, table, schema=pyarrow_schema, mode="append", engine="rust", partition_by=["Uid","date","hour"], schema_mode="merge", configuration={"delta.logRetentionDuration": "interval 7 day"})
```

Despite it being a brand new Delta table and after some successful writes, eventually the lambdas started erroring with `Generic DeltaTable error: Version mismatch`. I believe the error is coming from here: https://github.com/delta-io/delta-rs/blob/3e6a4d61923602d189f559636b3e3e3f61b6a924/crates/core/src/table/state.rs#L192

**What you expected to happen**:

Especially since we are testing with a fresh table, I'd expect all writes to work (and not just some) even with the new schema merge flag set.

**How to reproduce it**:
I was not able to reproduce with a randomly generated dataset locally, so my guess is its something more to do with the dynamo locking on S3 If you have thoughts on how I could test this better, please let me know.

Note that we have roughly 10 concurrent lambdas that could potentially write to Lambda. However, before this change we had 50 writing with pyarrow and all was well.

rtyler:
Does this only manifest with the schema evolution? Or are you able to see errors with append or merge writes as well?

ion-elgreco:
> Does this only manifest with the schema evolution? Or are you able to see errors with append or merge writes as well?

It happens at any operation when there is concurrency and the state gets updated at the end

PR #2280, opened 2024-03-11
fix(rust): update snapshot when having concurrent writers

### LINKED ISSUE #2279: Successful writes return error when using concurrent writers
# Environment

**Delta-rs version**: `deltalake 0.17.1`

**Binding**: `rust`

**Environment**:
- **Cloud provider**: `S3`
- **OS**: Tested locally on `MacOs Sonoma 14.2.1` and using Docker base image `rust:1.76-bookworm`
- **Other**:

***
# Bug

**What happened**:
I have two concurrent writers into the same table. Each writer performs an Append operation to the table inserting one row. I use `S3DynamoDbLogStore` to prevent data loss. One writer returns back an error while the other succeeds. However when querying the table both entries are successfully inserted.

When the writer succeeds, I expect the corresponding data to be part of the resulting table.
When the writer fails, I expect the data corresponding to that write not to be part of the resulting table.

**How to reproduce it**:
First we set up `S3DynamoDBLogStore`. Then we use two parallel threads to insert two rows in parallel.

```rust
async fn insertToTable(uri: &str, value: i32) {
    // Setup access to S3 and Dynamo. Put values for AWS_S3_LOCKING_PROVIDER=dynamodb and DELTA_DYNAMO_TABLE_NAME
    let storage_options: HashMap<String, String> =  HashMap::new();
    let delta_table: deltalake::DeltaOps = 
        deltalake::DeltaOps::try_from_uri_with_storage_options(uri, storage_options).await.unwrap();
    
    // Create one row
    let field1 = Int32Array::from(vec![value]);
    let schema = Arc::new(Schema::new(vec![
        Field::new("foo", DataType::Int32, false),
    ]));
    let record_batch =
        vec![RecordBatch::try_new(schema.clone(), vec![Arc::new(field1)]).unwrap()];
    
    // Write row to delta table
    match delta_table.write(record_batch).await {
        Ok(_) => {
            log::info!("Wrote to delta table at '{}'", uri);
        }
        Err(e) => {log::info!("Error writing to delta table: {}", e.to_string());},
    }
}

#[tokio::main]
async fn main() {
    deltalake::aws::register_handlers(None);
    let task1 = tokio::spawn(insertToTable(uri, 1001));
    let task2 = tokio::spawn(insertToTable(uri, 1002));
    let _ = tokio::try_join!(task1, task2);
}
```

We get back:
```sh
> cargo run
[WARN  deltalake_aws::logstore] LockClientError::VersionAlreadyExists(1)
[INFO my_project] Wrote to delta table at 's3://...'
[INFO my_project] Error writing to delta table: Generic DeltaTable error: Version mismatch
```

When reading the table I get back both values (while the one write returned an error).

**More details**:
By digging a bit deeper into the code, seems that the error originates at the final part of the write operation. At the end of the write operation, the writer attempts to [advance](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/write.rs#L820) the (in-memory) snapshot so to include the commit made by the writer. However the snapshot was already advanced by a different process / writer, that the writer is not aware of. We get back a [DeltaTableError::Generic("Version mismatch")](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/table/state.rs#L192).

When making the change [here](https://github.com/helanto/delta-rs/commit/71342cdbbff47c11b2e76b239fb67bd75976d163) the issue is fixed. Not sure however if this is correct path to follow.

helanto:
# Description
This PR handles the case of updating the snapshot when having concurrent writers. The issue is apparent when the data store supports concurrent writes (e.g. `S3DynamoDbLogStore`). 

Since our log store implementation guarantees no data loss, it can be the case that two (or more) writers starting from the same snapshot (e.g. version 10), both manage to commit their changes. One will write version 11, while the other version 12. However the latter will not be able to update the snapshot from version 10 to version 12 (because it lacks information about version 11) and will fail even when the write succeeds. In such cases we update the snapshot to version = committed_version - 1, so that the writer knows how to merge the snapshot.

# Related Issue(s)

- fixes #2279
- fixes https://github.com/delta-io/delta-rs/issues/2262

# Documentation

### DIFF
--- a/crates/core/src/operations/write.rs
+++ b/crates/core/src/operations/write.rs
@@ -812,6 +812,11 @@ impl std::future::IntoFuture for WriteBuilder {
             // created actions, it may be safe to assume, that we want to include all actions.
             // then again, having only some tombstones may be misleading.
             if let Some(mut snapshot) = this.snapshot {
+                if snapshot.version() != commit.version - 1 {
+                    snapshot
+                        .update(this.log_store.clone(), Some(commit.version - 1))
+                        .await?;
+                }
                 snapshot.merge(commit.data.actions, &commit.data.operation, commit.version)?;
                 Ok(DeltaTable::new_with_state(this.log_store, snapshot))
             } else {


rtyler:
I need to noodle on this a moment, the update does add additional read calls to storage and I _think_ there should be a way in which we get to this point in the write operation and should be able to avoid the reads

JonasDev1:
The conflict detection also reads the concurrent write logs. It would be most efficient to combine both

Issue #2162, opened 2024-02-03
Bug/Question:  arrow's`FixedSizeList` is not roundtrippable


universalmind303:
# Bug

**What happened**:

Currently if you try to write an arrow `FixedSizeList` using delta, the original type is converted to a list. I'm wondering if this is a bug or intentional. Are there any suggested workarounds for working with arrow `FixedSizeList` without losing the fixed size? 

**What you expected to happen**:
I'd expect being able to roundtrip arrow's `FixedSizeList` to/from delta without losing the original type. 

**How to reproduce it**:

**More details**:

ion-elgreco:
Intentional, the delta primitive types doesn't have a fixedsizedlist/array, only list

PR #2283, opened 2024-03-11
feat: support time type

### LINKED ISSUE #2272: Support `Time32` and `Time64` types
# Description

The delta-rs `PrimitiveType` supports timestamp and date, but not time. This means that Arrow `Time32` and `Time64` types are not supported.

**Use Case**

I would like to use `Time32` and `Time64` types in my DataFusion `DeltaTable`s.

**Related Issue(s)**

rebasedming:
# Description
This PR adds a `Time` `PrimitiveType` so that `DeltaTable`s support the Arrow time type. Nanosecond precision was chosen to align with DataFusion's default time precision.

I'm opening this as draft since there was discussion in the related issue about how time is not supported by the delta protocol. I'm curious to understand what the consequences of that would be wrt merging this PR -- please let me know in the comments!

# Related Issue(s)

- closes #2272 

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -890,6 +890,10 @@ pub(crate) fn get_null_of_arrow_type(t: &ArrowDataType) -> DeltaResult<ScalarVal
             precision.to_owned(),
             scale.to_owned(),
         )),
+        ArrowDataType::Time32(TimeUnit::Second) => Ok(ScalarValue::Time32Second(None)),
+        ArrowDataType::Time32(TimeUnit::Millisecond) => Ok(ScalarValue::Time32Millisecond(None)),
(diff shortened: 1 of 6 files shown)


ion-elgreco:
@rebasedming if we would merge this, then it would break Delta protocol compatibility, which would not allow other writers or reader implementations to read delta tables.

You would need to put a feature request in at delta repo, like this one https://github.com/delta-io/delta/issues/2319

hntd187:
Yeah, if it's not in the protocol it's very unlikely this will be accepted.

universalmind303:
I wonder if we should at least write some additional metadata as to what the original arrow type was. This way, it'd be fully compatible with the delta spec, and arrow users could still retain the original information too. Then the user could just parse that metadata and either cast, or decode directly to the original type. This would also fix this issue https://github.com/delta-io/delta-rs/issues/2162

Issue #2242, opened 2024-03-04
Partition column comparison is an assertion rather than if block with raise exception


joe-sharman:
# Environment

**Delta-rs version**:
0.16.0
**Binding**: 
python
**Environment**:
MacOS

***
# Bug

**What happened**:
An assertion is done between partition columns from incoming data and target table. If this fails an AssertionError is raised.
https://github.com/delta-io/delta-rs/blob/15538b9b42ba3d0d20ffd287bb4cc463ff8dfb1d/python/deltalake/writer.py#L340

**What you expected to happen**:
This should be an if block, and raise an exception if the two values aren't the same. `ValueError` would be fine.

**How to reproduce it**:
Try to write a table with the wrong partition columns specified in the incoming data.

**More details**:
This exception should also be changed, maybe something like `FileExistsError`
https://github.com/delta-io/delta-rs/blob/15538b9b42ba3d0d20ffd287bb4cc463ff8dfb1d/python/deltalake/writer.py#L333

Edit:
assert statements are ignored when running in optimise mode, so this could cause issues for a user running in optimised mode

ion-elgreco:
@joe-sharman do you want to make a contribution to address this?

joe-sharman:
I'll put something together :)

joe-sharman:
Hey @ion-elgreco I've had a go making the change. I've tried to follow the commit conventions in the repo and have used built-in exceptions rather than custom as that seems to be the convention here.

PR #2286, opened 2024-03-13
fix: replace assert and AssertionError with appropriate exceptions

joe-sharman:
# Description

- Replaces assert and AssertionError with built-in exceptions
- Amended tests to reference new exception types
- Following conventions in file of using built-in exceptions rather than custom exceptions

# Related Issue(s)
Closes https://github.com/delta-io/delta-rs/issues/2242

# Documentation


ion-elgreco:
@joe-sharman Thanks for the contribution!! 🕺

ion-elgreco (approved):

PR #2289, opened 2024-03-15
feat: no longer load full table into ram in write by using concurrent write

### LINKED ISSUE #2255: Do not load full source into RAM on write_to_deltalake

aersam:
# Description
This is a followup of #2265 

It additionally uses streams/channels to concurrently write at the cost of more memory consumption. Default is keeping one recordbatch in RAM only, so it's opt-in. 

I tested this with a local file and it went from 700s to 200s if I work with 10 concurrent streams. Of course memory consumption goes up, but given that we currently load the whole table in RAM, it's OK :)

This adds a depenency on async-channel as I need a multi-consumer channel.

# Related Issue(s)
Fixes #2255


ion-elgreco:
@aersam your old PR can be closed then? Can you also add a python test where you use this concurrent_stream parameter?

aersam:
> @aersam your old PR can be closed then? Can you also add a python test where you use this concurrent_stream parameter?

Yes, if you're ok to add an extra dependency?

crates/core/src/operations/write.rs
```diff
@@ -114,28 +116,58 @@ impl FromStr for SchemaMode {
...
-    /// The input plan
-    input: Option<Arc<dyn ExecutionPlan>>,
     /// Datafusion session state relevant for executing the input plan
     state: Option<SessionState>,
     /// SaveMode defines how to treat data already written to table location
     mode: SaveMode,
     /// Column names for table partitioning
     partition_columns: Option<Vec<String>>,
+    /// Number of streams to run concurrently on reading
+    nr_concurrent_streams: u32,
```
rtyler, line 164:
```suggestion
    concurrent_streams: u32,
```

crates/core/src/operations/write.rs
```diff
@@ -152,45 +184,61 @@ pub struct WriteBuilder {
...
+                state: None,
+                mode: SaveMode::Append,
+                partition_columns: None,
+                predicate: None,
+                target_file_size: None,
+                write_batch_size: None,
+                safe_cast: false,
+                schema_mode: None,
+                writer_properties: None,
+                nr_concurrent_streams: 1,
```
rtyler, line 211:
```suggestion
                concurrent_streams: 1,
```

We already have the [num_cpus](https://docs.rs/num_cpus/latest/num_cpus/) package as a dependency in core, would it be beneficiail for the default performance case (and validation of this code path) to set concurrent_streams to at least the number of CPUs?

Cargo.toml
```diff
@@ -70,4 +70,5 @@ uuid = { version = "1" }
 async-trait = { version = "0.1" }
 futures = { version = "0.3" }
 tokio = { version = "1" }
+async-channel = { version = "2.2" }
```
rtyler, line 73: As far as I can tell this is only used in one narrow slice within the core crate (datafusion-enabled writes) so I don't believe we need to sprout this dependency at the top-level

crates/core/Cargo.toml
```diff
@@ -66,7 +66,7 @@ tokio = { workspace = true, features = [
     "fs",
     "parking_lot",
 ] }
-
+async-channel = { workspace = true }
```
rtyler, line 69: Since this is only required for the datafusion write, I am comfortable adding the new dependency but would like it to be optional and gated/included with the `datafusion` feature

python/src/lib.rs
```diff
@@ -1407,20 +1408,22 @@ fn write_to_deltalake(
...
     data: PyArrowType<ArrowArrayStreamReader>,
+    data_schema: PyArrowType<ArrowSchema>,
     mode: String,
     max_rows_per_group: i64,
     schema_mode: Option<String>,
     partition_by: Option<Vec<String>>,
     predicate: Option<String>,
     name: Option<String>,
     description: Option<String>,
+    nr_concurrent_streams: Option<u32>,
```
rtyler, line 1419:
```suggestion
    concurrent_streams: Option<u32>,
```

python/deltalake/_internal.pyi
```diff
@@ -171,13 +171,15 @@ def write_new_deltalake(
...
     data: pyarrow.RecordBatchReader,
+    data_schema: pyarrow.Schema,
     partition_by: Optional[List[str]],
     mode: str,
     max_rows_per_group: int,
     schema_mode: Optional[str],
     predicate: Optional[str],
     name: Optional[str],
     description: Optional[str],
+    nr_concurrent_streams: Optional[int],
```
rtyler, line 182:
```suggestion
    concurrent_streams: Optional[int],
```

python/deltalake/_internal.pyi
```diff
@@ -171,13 +171,15 @@ def write_new_deltalake(
 def write_to_deltalake(
     table_uri: str,
     data: pyarrow.RecordBatchReader,
+    data_schema: pyarrow.Schema,
```
rtyler, line 174: Why does this argument need to be added? the `RecordBatchReader` should already have a schema associated with it which can be used

python/deltalake/writer.py
```diff
@@ -129,6 +129,7 @@ def write_deltalake(
     mode: Literal["error", "append", "ignore"] = ...,
     name: Optional[str] = ...,
     description: Optional[str] = ...,
+    nr_concurrent_streams: Optional[int] = ...,
```
rtyler, line 132:
```suggestion
    concurrent_streams: Optional[int] = ...,
```

python/deltalake/writer.py
```diff
@@ -157,6 +158,7 @@ def write_deltalake(
     mode: Literal["overwrite"],
     name: Optional[str] = ...,
     description: Optional[str] = ...,
+    nr_concurrent_streams: Optional[int] = ...,
```
rtyler, line 161:
```suggestion
    concurrent_streams: Optional[int] = ...,
```

python/deltalake/writer.py
```diff
@@ -191,6 +193,7 @@ def write_deltalake(
     max_rows_per_group: int = 128 * 1024,
     name: Optional[str] = None,
     description: Optional[str] = None,
+    nr_concurrent_streams: Optional[int] = None,
```
rtyler, line 196:
```suggestion
    concurrent_streams: Optional[int] = None,
```

python/deltalake/writer.py
```diff
@@ -246,6 +249,7 @@ def write_deltalake(
             If this value is set, then min_rows_per_group should also be set.
         name: User-provided identifier for this table.
         description: User-provided description for this table.
+        nr_concurrent_streams: Number of concurrent streams to use when writing. At least nr_concurrent_streams files will be written.
```
rtyler, line 252:
```suggestion
        concurrent_streams: Number of concurrent streams to use when writing. At least concurrent_streams files will be written.
```

python/deltalake/writer.py
```diff
@@ -316,13 +320,15 @@ def write_deltalake(
...
             data=data,
+            data_schema=schema,
             partition_by=partition_by,
             mode=mode,
             max_rows_per_group=max_rows_per_group,
             schema_mode=schema_mode,
             predicate=predicate,
             name=name,
             description=description,
+            nr_concurrent_streams=nr_concurrent_streams,
```
rtyler, line 331:
```suggestion
            concurrent_streams=concurrent_streams,
```

python/src/lib.rs
```diff
@@ -1440,7 +1446,9 @@ fn write_to_deltalake(
         if let Some(partition_columns) = partition_by {
             builder = builder.with_partition_columns(partition_columns);
         }
-
+        if let Some(nr_concurrent_streams) = nr_concurrent_streams {
+            builder = builder.with_nr_concurrent_streams(nr_concurrent_streams);
+        }
```
rtyler, line 1451:
```suggestion
        if let Some(concurrent_streams) = concurrent_streams {
            builder = builder.with_concurrent_streams(concurrent_streams);
        }
```

crates/core/src/operations/write.rs
```diff
@@ -724,24 +715,72 @@ impl std::future::IntoFuture for WriteBuilder {
...
+                Some(WriteData::Vecs(input_batches)) => {
+                    let stream = futures::stream::iter(input_batches.into_iter().map(Ok));
+                    let rec_batch_stream =
+                        Box::pin(RecordBatchStreamAdapter::new(input_schema.clone(), stream))
+                            as SendableRecordBatchStream;
+                    vec![rec_batch_stream]
+                }
+                Some(WriteData::RecordBatches((input_batches, _))) => {
+                    let (sender, receiver) =
+                        async_channel::bounded(this.nr_concurrent_streams as usize);
```
rtyler, line 733:
```suggestion
                        async_channel::bounded(this.concurrent_streams as usize);
```

crates/core/src/operations/write.rs
```diff
@@ -724,24 +715,72 @@ impl std::future::IntoFuture for WriteBuilder {
...
+                    producer_task = Some(tokio::task::spawn(async move {
+                        for batch in input_batches {
+                            sender.send(batch).await?;
+                        }
+                        sender.close();
+                        Ok(())
+                    }));
+                    (0..this.nr_concurrent_streams)
```
rtyler, line 741:
```suggestion
                    (0..this.concurrent_streams)
```

crates/core/src/operations/write.rs
```diff
@@ -200,69 +248,70 @@ impl WriteBuilder {
...
     pub fn with_description(mut self, description: impl Into<String>) -> Self {
-        self.description = Some(description.into());
+        self.config.description = Some(description.into());
+        self
+    }
+
+    /// Set the number of concurrent streams to use when writing data from a RecordBatch
+    pub fn with_nr_concurrent_streams(mut self, nr_concurrent_streams: u32) -> Self {
+        self.config.nr_concurrent_streams = nr_concurrent_streams;
         self
```
rtyler, line 315:
```suggestion
    pub fn with_concurrent_streams(mut self, concurrent_streams: u32) -> Self {
        self.config.concurrent_streams = concurrent_streams;
        self
```

crates/core/src/operations/write.rs
aersam: I would not. If you write a small file with small batches you'd create lots of very small files otherwise. It would make sense for larger files only

python/deltalake/_internal.pyi
aersam: Hm, true. I'll change that.

aersam:
> Let's get the other one in first

I hope it's ok to prioritize this one from my side to not have to keep both branches up-to-date

ion-elgreco:
> > Let's get the other one in first
> 
> I hope it's ok to prioritize this one from my side to not have to keep both branches up-to-date

That's ok!

ion-elgreco:
@aersam Should we perhaps for python users max out the amount of concurrent streams.

In most use cases we are passing a recordBatchReader where the recordBatches are already in memory before constructing the reader, in that case you won't see any memory difference. And it wouldn't be different than the prior behavior since the reader was always collected.

I also have one suggestion on the python side, I think it's better if we simplify it and just provide a parameter called `parallelize`, which is always set to True. If users want to control the amount of concurrent streams, they should set an env_var which we then can parse in python_lib, if it's not set and parallelize = True, then we take the max possible streams.

aersam:
How about `parallelize:bool|int` on python side? 🙂

ion-elgreco:
@aersam that also works! :)

aersam:
I finally had the time to update this branch with the new parallel parameter in python. Hope it's looking good now!

aersam:
I only did some manual test on my own data, but could probably write some benchmark in python, using duckdb or polars as source. Would it make sense to add this to the code somehow?

ion-elgreco:
@aersam here you could add it, and even maybe reuse some of the benchmarks there: https://github.com/delta-io/delta-rs/tree/main/crates/benchmarks

aersam:
I did some very basic benchmarking, but the results were not as I hoped :) While RAM consumption is significantly lower, the speed is not good enough yet. I think maybe the channel must be bigger, I'll do some more testing

I did my test quick and dirty using python, I can share the code if you want. Basically it's this:

```python
import duckdb
from uuid import uuid4

with duckdb.connect() as con: # get your 42.parquet here: https://duckdb.org/2024/03/26/42-parquet-a-zip-bomb-for-the-big-data-age.html
    con.execute("select b, random() as a from read_parquet('42.parquet') limit 300000000")
    reader = con.fetch_record_batch()
    write_deltalake(f"_test/{uuid4()}", reader, schema=reader.schema, mode="overwrite", engine="rust") 
```

aersam:
Pretty sure the non-async write causes issues. But object_store 0.10 will change a lot there, so maybe better to wait for that

ion-elgreco:
> Pretty sure the non-async write causes issues. But object_store 0.10 will change a lot there, so maybe better to wait for that

Yes let's see how effective these changes are with new upload trait

ion-elgreco:
@aersam fyi, ObjectStore just got bumped in the repo to 0.10

ion-elgreco:
@aersam hey, do you think you have time to resolve the merge conflicts?

aersam:
> @aersam hey, do you think you have time to resolve the merge conflicts?

I'm sorry, I had a bit a shift in priorities, so it will take time to do so. Especially since there were quite some changes in the writer as I see

ion-elgreco:
> > @aersam hey, do you think you have time to resolve the merge conflicts?
> 
> I'm sorry, I had a bit a shift in priorities, so it will take time to do so. Especially since there were quite some changes in the writer as I see

No worries! Just ping me once it's ready for another review round

rtyler:
I'm going to close this, I will smash this memory issue soon enough :smile:

PR #2291, opened 2024-03-15
fix: merge predicate for concurrent writes

### LINKED ISSUE #2084: _internal.DeltaError when merging
# Environment

**Delta-rs version**: 0.15.1

**Binding**: Python

**Environment**:
- **Cloud provider**: Azure (deltatable stored in Azure)
- **OS**: Linux/OpenShift

***
# Bug

**What happened**:
I try to merge data into an existing deltatable like so:

```
deltatable.merge(
                mergeable_table,
                f"target._merge_key = source._merge_key",
                source_alias="source",
                target_alias="target",
            )
            .when_matched_update_all()
            .when_not_matched_insert_all()
            .execute()
```

This sometimes works, and sometimes fails with the following error:
`_internal.DeltaError: Generic DeltaTable error: Schema error: No field named target._merge_key. Valid fields are _merge_key, <other camel-case columns>`

I expect the merge to succeed. The error claims that `target._merge_key` is not a valid field name, but `_merge_key` is, which is a bit strange.

**How to reproduce it**:
I have had trouble reproducing this locally and/or reliably.
The merge operations are run as part of a batch job which begins with 
```
deltatable.optimize.compact()
deltatable.vacuum(dry_run=False, retention_hours=48, enforce_retention_duration=False)
deltatable.cleanup_metadata()
```
followed by a number of merge operations. If the first merge fails, all subsequent merge operations also fail. Sometimes the first merge succeeds, and then all subsequent merges also seem to succeed. So I wonder if any of these optimize/vacuum/cleanup methods could (sometimes) corrupt the table? (Maybe relevant: Are these operations run synchronously or asynchronously?)

### LINKED ISSUE #2227: Merging to a table with multiple distinct partitions in parallel fails
# Environment

**Delta-rs version**:
0.15.3
**Environment**: GKE
- **Cloud provider**: Google

***
# Bug

**What happened**:

I don't have a good MRE, but errors are raised if multiple partitions are merged to in parallel. This does not happen if we use "overwrite" or "append" table writes, just when we `.execute()` a merge. For example, if I pass the unique partitions which were updated based on the transaction log to dynamic Airflow tasks (each task processes a single partition) then only one of the tasks will succeed. Other tasks will fail:

```
  File "/home/airflow/.local/lib/python3.11/site-packages/deltalake/table.py", line 1597, in execute
    metrics = self.table._table.merge_execute(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_internal.DeltaError: Generic DeltaTable error: Schema error: No field named s.unique_row_hash. Valid fields are
```

Example of an Airflow task which attempts to merge each of these mapped tasks which represent distinct partitions which should not have any conflict with each other. Right now I need to limit my schedule to only allow one partition to update at a time which slows down the pipeline a huge amount compared to using "overwrite" instead of merges.

image

Since each partition is unique (for example, each partition is an individual date) they should be able to be written to in parallel. The error message about the column not existing is false as well. Clearing the task makes it succeed as well.

**How to reproduce it**:

Merge to a Delta Table using partition_filters where multiple partitions are written to in parallel.

**More details**:

JonasDev1:
# Description
This merge request will change the commitInfo.operationParameters.predicate of the merge operation. This is required to allow conflict checking on concurrent commits. Before, the predicate was just the simple merge predicate like `source.eventId = target.eventId`, but the conflict checker doesn't know these aliases and doesn't have access to the source df. 
So I now use the early_filter, which is also used before to filter a subset of the target table. Basically, the early_filter only contains static filters and partition filters that are converted to fixed values based on the source data.  The early_filter can be None if no column/partition pre-filtering is possible, or if the merge contains a not_match_source operation. (See the generalize_filter function in the file).
The commitInfo predicate uses exactly this filter, except that the target alias is removed.  

The predicate is used by the conflict checker, for example when there are multiple concurrent merges. If there is a predicate, the conflict checker will check whether the concurrent commit wrote or deleted data within that predicate. If the predicate is None, the conflict checker will treat the commit as a `read_whole_table' and interpret any concurrent updates as a conflict. 

Example:
Target table with partition country
Merge with predicate `source.id = target.id AND target.country='DE'` -> CommitInfo predicate `country='DE'`
Merge with predicate `source.id = target.id AND target.country=source.country` -> CommitInfo predicate `country='DE' OR country='US'`
Merge with predicate `source.id = target.id` -> CommitInfo predicate None (As full target table join is required)

# Related Issue(s)
- closes #2084
- closes #2227

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -866,10 +866,6 @@ async fn try_construct_early_filter(
     let table_metadata = table_snapshot.metadata();
     let partition_columns = &table_metadata.partition_columns;
 
-    if partition_columns.is_empty() {
-        return Ok(None);
-    }
-
     let mut placeholders = HashMap::default();
 
     match generalize_filter(
@@ -2720,6 +2716,32 @@ mod tests {
         })
         .eq(col(Column::new(target.clone().into(), "id")))
         .and(col(Column::new(target.clone().into(), "id")).eq(lit("C")));
+
+        assert_eq!(generalized, expected_filter);
+    }
+
+
+    #[tokio::test]
+    async fn test_generalize_filter_keeps_only_static_target_references() {
+        let source = TableReference::parse_str("source");
+        let target = TableReference::parse_str("target");
+
+        let parsed_filter = col(Column::new(source.clone().into(), "id"))
+            .eq(col(Column::new(target.clone().into(), "id")))
+            .and(col(Column::new(target.clone().into(), "id")).eq(lit("C")));
+
+        let mut placeholders = HashMap::default();
+
+        let generalized = generalize_filter(
+            parsed_filter,
+            &vec!["other".to_owned()],
+            &source,
+            &target,
+            &mut placeholders,
+        )
+        .unwrap();
+
+        let expected_filter = col(Column::new(target.clone().into(), "id")).eq(lit("C"));
 
         assert_eq!(generalized, expected_filter);
     }
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -29,6 +29,7 @@
 //! ````
 
 use std::collections::HashMap;
+use std::ops::Deref;
 use std::sync::Arc;
 use std::time::Instant;
 
@@ -855,6 +856,7 @@ fn replace_placeholders(expr: Expr, placeholders: &HashMap<String, ScalarValue>)
     .unwrap()
 }
 
+
 async fn try_construct_early_filter(
     join_predicate: Expr,
     table_snapshot: &DeltaTableState,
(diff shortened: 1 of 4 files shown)


ion-elgreco:
@Blajda :)

JonasDev1:
I have found another bug: After the successful merge, the merge operation will retry a DeltaTable, for this purpose new commit will be merged in the old snapshot, but the concurrent commit is not included in it, which results in a version mismatch error. I'll try to recreate this in an integration test.

ion-elgreco:
@JonasDev1 yup it's a known issue and there is a PR open for it: https://github.com/delta-io/delta-rs/pull/2280

JonasDev1:
I have added some merge concurrency tests where the version mismatch error also occurs

Blajda (approved):
LGTM!
The Delta protocol doesn't specify anything regarding conflicting resolution so I think this is okay.
Some of the details might change if we need to inter-op with other implementations.

PR #2295, opened 2024-03-16
fix(rust): add missing chrono-tz feature

ion-elgreco:
# Description
Hopefully resolves: `Failed to convert into Arrow schema: Parser error: Invalid timezone "UTC": only offset based timezones supported without chrono-tz feature`


ion-elgreco:
@rtyler Yey 👯

PR #2305, opened 2024-03-20
fix: schema evolution not coercing with large arrow types

### LINKED ISSUE #2298: Schema evolution not coercing with Large arrow types
# Environment

**Delta-rs version**:

**Binding**: 0.16.0

# Bug

**What happened**:
Writing a delta table that has a partition column which is primitive type String and then trying to write with schema_mode="merge" to add new columns results in, when your partition column is in arrow dtype LargeUtf8
```python
SchemaMismatchError: Schema error: Fail to merge schema field 'bar' because the from data_type = LargeUtf8 does not equal Dictionary(UInt16, Utf8)
Schema error: Fail to merge schema field 'bar' because the from data_type = LargeUtf8 does not equal Dictionary(UInt16, Utf8)
```

aersam:
# Description
This fixes schema merge for large arrow types in combination with dictionary types. Basically we just allow merging any arrow data type that is the same delta type

# Related Issue(s)
Fixes #2298

# Documentation

### DIFF
--- a/crates/core/src/kernel/arrow/mod.rs
+++ b/crates/core/src/kernel/arrow/mod.rs
@@ -260,7 +260,8 @@ impl TryFrom<&ArrowDataType> for DataType {
                 } else {
                     panic!("DataType::Map should contain a struct field child");
                 }
-            }
+            },
+            ArrowDataType::Dictionary(_, value_type) => Ok(value_type.as_ref().try_into()?),
             s => Err(ArrowError::SchemaError(format!(
                 "Invalid data type for Delta Lake: {s}"
             ))),
--- a/crates/core/src/kernel/arrow/mod.rs
+++ b/crates/core/src/kernel/arrow/mod.rs
@@ -260,7 +260,7 @@ impl TryFrom<&ArrowDataType> for DataType {
                 } else {
                     panic!("DataType::Map should contain a struct field child");
                 }
-            },
+            }
             ArrowDataType::Dictionary(_, value_type) => Ok(value_type.as_ref().try_into()?),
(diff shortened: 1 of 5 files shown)


ion-elgreco:
@aersam wouldn't it maybe be better to have the schema merge function take in delta fields instead of arrow

aersam:
> @aersam wouldn't it maybe be better to have the schema merge function take in delta fields instead of arrow

Yes! But thats a topic throughout the whole crate, everything is based on arrow schemas instead of delta schemas.

ion-elgreco:
@aersam can be addressing in a follow up if it's to big of a change now.

Btw, forgot to add that in the issue but it is also not working with large lists, and nested versions

aersam:
> @aersam can be addressed in a follow up if it's to big of a change now.
> 
> Btw, forgot to add that in the issue but it is also not working with large lists, and nested versions

To merge nested correctly, there's no other way than to really do the full thing on our own. So that's what I did now. Since we're still merging arrow schemas, I now do first a conversion to delta and then back to arrow again, which is not optimal, but still does work.

aersam:
I'll adjust the tests 🙂

aersam:
Should I move the delta schema merge code to kernel?

ion-elgreco:
@aersam I believe kernel should only for stuff related to delta protocol implementation, and schema evolution is more on the operation side

aersam:
Oh I guess I should merge metadata, too. Will add that

ion-elgreco (approved):
@aersam Great fix!! 😃

Issue #2266, opened 2024-03-08
Failure to read from table in S3 with special characters like spaces in path


alexr-bq:
# Environment

**Delta-rs version**:
0.17.0 (and earlier)

**Binding**: 
Rust

**Environment**:
- **Cloud provider**: AWS
- **OS**: OSX Sonoma 14.1.1

***
# Bug

**What happened**:
When trying to read from a table path with a space in the name, it fails.
Err(NotATable("no log files"))

**What you expected to happen**:
Expected the table to be read as normal.

**How to reproduce it**:

Create a table with a space in the path, ie at "s3a://testbucket/data/test1 /"
Then run:
`
deltalake::aws::register_handlers(None);
    let storage_options: HashMap<String, String> = <your storage options here>

    let table_filepath = "s3a://testbucket/data/test1 /"; 
    let mut table = deltalake::open_table_with_storage_options(
        table_filepath,
        storage_options,
    )
    .await;
    println!("Table result: {:?}", table);
`

**More details**:

Looked into what is happening here a little. It seems like we URL encode the path right away, so the above example becomes s3a://testbucket/data/test1%20/. Then, we pass this to the S3 client (object_store-0.7.1/src/aws/client.rs) which passes "data/test1%20" as part of its API call, but that S3 API does not expect a URL encoded string.

PR #2308, opened 2024-03-20
fix(rust): special characters in s3 paths

alexr-bq:
# Description
Fixes https://github.com/delta-io/delta-rs/issues/2266

We construct a "Prefix handler" for S3 paths. This prefix is passed into an s3 API that does not expect a URL-encoded path, so here we decode the path to support paths with characters such as spaces.

# Related Issue(s)
2266


crates/aws/src/lib.rs
```diff
@@ -46,7 +46,18 @@ impl LogStoreFactory for S3LogStoreFactory {
         location: &Url,
         options: &StorageOptions,
     ) -> DeltaResult<Arc<dyn LogStore>> {
-        let store = url_prefix_handler(store, Path::parse(location.path())?)?;
+        let path_decoded = match urlencoding::decode(location.path()) {
+            Ok(res) => {
+                let result = res.into_owned();
+                result
+            }
```
rtyler, line 53:
```suggestion
            Ok(res) => res.into_owned(),
```

crates/aws/src/lib.rs
```diff
@@ -46,7 +46,18 @@ impl LogStoreFactory for S3LogStoreFactory {
         location: &Url,
         options: &StorageOptions,
     ) -> DeltaResult<Arc<dyn LogStore>> {
-        let store = url_prefix_handler(store, Path::parse(location.path())?)?;
+        let path_decoded = match urlencoding::decode(location.path()) {
```
rtyler, line 49: This line would benefit from a comment as to what this decoding does/why it exists

alexr-bq: Added

rtyler (commented):
I would feel more confident with a test, but I'm willing to roll the dice here :smile: :game_die:

jkylling:
Is it correct to use URL encoding here? I believe Delta Lake uses URI encoded paths everywhere and not URL encoded paths.

Issue #2309, opened 2024-03-20
delta_rs don't seems to respect the row group size


djouallah:
I would like to have delta_rs to write parquet files with bigger row group in Parquet, ideally 8 Millions rows per group, but some far, it does not seems to works,  What I am missing ?

```
write_deltalake("/lakehouse/default/Tables/scada_duckdb",\
   df,mode="append",
    engine='rust',
    partition_by=['YEAR'],\
    max_rows_per_file=4*chunk_size,\
    max_rows_per_group= chunk_size,\
    min_rows_per_group=chunk_size,\
   storage_options={"allow_unsafe_rename":"true"})
    del df
```

ion-elgreco:
Two things, I see the max_rows_per_group is incorrectly passed to the write_batch_size from python to rust, I'll make a fix to remove that.

You should use the WriterProperties class and pass that to write_deltalake, that contains `max_row_group_size`

djouallah:
@ion-elgreco  thanks, But I am more interested in max_row_group_size, I want the minimum to be 8 Millions rows ?

ion-elgreco:
@djouallah I don't see a way to set a minimum in the parquet crate

djouallah:
I am using this and it does not seems to be working ?

`writer_properties=("max_row_group_size"==8000000),`

ion-elgreco:
@djouallah works for me:

```python
from deltalake import WriterProperties
import polars as pl
import pyarrow.parquet as pq
import os
df = pl.DataFrame({
    "foo": list(range(10_000_000))
})

wp = WriterProperties(max_row_group_size=8_000_000)
df.write_delta("test_table",mode="append", delta_write_options={"writer_properties":wp, "engine":"rust"})
```

```python
file =  list(filter(lambda x: '.parquet' in x, os.listdir("test_table")))[0]
metadata = pq.read_metadata(os.path.join("test_table", file))
for i in range(metadata.num_row_groups):
    print(metadata.row_group(i))
```

result:

```
<pyarrow._parquet.RowGroupMetaData object at 0x7f3bb4dbcea0>
  num_columns: 1
  num_rows: 8000000
  total_byte_size: 64282340
  sorting_columns: ()
<pyarrow._parquet.RowGroupMetaData object at 0x7f3bb4dbcea0>
  num_columns: 1
  num_rows: 2000000
  total_byte_size: 16297257
  sorting_columns: ()

```

djouallah:
[...]

PR #2314, opened 2024-03-22
fix(python): wrong batch size

ion-elgreco:
# Description
- closes https://github.com/delta-io/delta-rs/issues/2309

### DIFF
--- a/crates/core/src/operations/writer.rs
+++ b/crates/core/src/operations/writer.rs
@@ -386,7 +386,7 @@ impl PartitionWriter {
                 self.flush_arrow_writer().await?;
             }
         }
-
+        self.arrow_writer.flush();
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


rtyler:
I'm struggling to understand this change and how it relates to the issue, I don't see how WriterProperties are sent in the Rust writer, or how removing this argument fixes things, maybe it just clears up confusion?

The `write_deltalake` function still has this parameter but I'm _also_ not understanding why :laughing:

ion-elgreco:
@rtyler when I built the rust engine on the python side, I misinterpreted the max row group size as the batch size.

The actual max_batch_size can be set with the python WriterProperies class.

It's confusing because we are still supporting both pyarrow and rust engines

ion-elgreco:
@wjones127 There are some tests here:  https://github.com/delta-io/delta-rs/blob/main/python/tests/test_writerproperties.py.

Setting the max_row_group_size inside WriterProperties works fine, see here: https://github.com/delta-io/delta-rs/issues/2309#issuecomment-2016733155.

 I just by accident pass `max_rows_per_group` which is used for the pyarrow writer to parquet::write_batch_size, which is something different : P

wjones127 (approved):

Issue #2087, opened 2024-01-16
After dt.optimize.compact(), commitInfo.operationMetrics.filesAdded is a JSON map when other readers (e.g. Databricks) expect a string


mattslack-db:
# Environment

**Delta-rs version**:
rust-v0.16.5
Python deltalake 0.15.1

**Binding**: 
Python binding, but believe not specific to Python as it is just a wrapper over this method

**Environment**:
- **Cloud provider**:
Azure
- **OS**:
Ubuntu
- **Other**:
Databricks 14.1

***
# Bug

**What happened**:
Running `dt.optimize.compact()` creates an entry in the delta log which is not readable by Databricks Runtime.

Example:
`"commitInfo":{"timestamp":1705439516034,"operation":"OPTIMIZE","operationParameters":{"targetSize":"104857600"},"operationMetrics":{"filesAdded":{"avg":50424.6,"max":50517,"min":50335,"totalFiles":5,"totalSize":252123},"filesRemoved":{"avg":14010.447154471545,"max":15213,"min":13942,"totalFiles":123,"totalSize":1723285},"numBatches":123,"numFilesAdded":5,"numFilesRemoved":123,"partitionsOptimized":5,"preserveInsertionOrder":true,"totalConsideredFiles":123,"totalFilesSkipped":0},"clientVersion":"`

**What you expected to happen**:
Databricks Runtime (and Delta Sharing) expect filesAdded to be a string

**How to reproduce it**:
This can be reproduced in Databricks by running `dt.optimize.compact()` against a Delta table after some appends and then running `DESCRIBE HISTORY` against the same table.

**More details**:

Blajda:
Hi @mattslack-db thanks for reporting this issue.
Does a specification exist for the expected format of operationMetrics? Testing integration with Spark can be bit annoying so we would test against that.

ion-elgreco:
@Blajda this is part of the commitInfo, which is a free format according to the protocol. It's what we discussed last week in the call

@mattslack-db we are actually following the protocol but spark-delta is not since they assume a dtype. We are looking at aligning this and making some of the assumptions of what the expected schema is into the protocol

mattslack-db:
@ion-elgreco agreed, according to the protocol, "Implementations are free to store any valid JSON-formatted data via the commitInfo action.", so although Databricks can make assumptions about the schema of `commitInfo` (in this case, that `operationMetrics` is a `map<string, string>`) it shouldn't fall over if that schema is not followed.

[...]

PR #2317, opened 2024-03-22
fix(rust): serialize MetricDetails from compaction runs to a string

liamphmurphy:
I am by no means a Rust developer and haven't touched it in years; so please let me know if there's a better way to go about this. The Rust z_order and optimize.compact already serializes the metrics before it is passed back to Python, which then deserializes it back, so the Python behavior in terms of expecting this as a Dict has not changed which I think is what we want.

# Description
Adds a custom serialzer and Display implementation for the `MetricDetails` fields, namely `filesAdded` and `filesRemoved` so that those fields are written as strings instead of a struct to the commit log. Query engines expect these fields to be strings on reads.

I had trouble getting the pyspark tests running locally, but here is an example optimize commit log that gets written with these changes: 

```
{"commitInfo":{"timestamp":1711125995487,"operation":"OPTIMIZE","operationParameters":{"targetSize":"104857600","predicate":"[]"},"clientVersion":"delta-rs.0.17.1","readVersion":10,"operationMetrics":{"filesAdded":"{\"avg\":19956.0,\"max\":19956,\"min\":19956,\"totalFiles\":1,\"totalSize\":19956}","filesRemoved":"{\"avg\":4851.833333333333,\"max\":10358,\"min\":3734,\"totalFiles\":6,\"totalSize\":29111}","numBatches":6,"numFilesAdded":1,"numFilesRemoved":6,"partitionsOptimized":1,"preserveInsertionOrder":true,"totalConsideredFiles":6,"totalFilesSkipped":0}}}
```

# Related Issue(s)
- https://github.com/delta-io/delta-rs/issues/2087

# Documentation

N/A


crates/core/src/operations/optimize.rs
```diff
@@ -75,17 +78,25 @@ pub struct Metrics {
...
-    /// Maximum file size of a operation
-    pub max: i64,
     /// Average file size of a operation
     pub avg: f64,
+    /// Maximum file size of a operation
+    pub max: i64,
+    /// Minimum file size of a operation
+    pub min: i64,
     /// Number of files encountered during operation
     pub total_files: usize,
```
liamphmurphy, line 101: Changing the order so that the serialization writes out the fields in alphabetical order.

ion-elgreco:
@liamphmurphy you could add a python test to test that DESCRIBE HISTORY now works in pyspark on a delta-rs compacted  delta table

liamphmurphy:
> @liamphmurphy you could add a python test to test that DESCRIBE HISTORY now works in pyspark on a delta-rs compacted delta table

Good idea 👍 , just need to figure out why the spark tests aren't working locally.

ion-elgreco:
@liamphmurphy are you adding the markers in pytest?

liamphmurphy:
> @liamphmurphy are you adding the markers in pytest?

Was a numpy error, had to reinstall it manually through pip

liamphmurphy:
@ion-elgreco Pushed up some changes for the lint / cargo fmt errors. The pyspark integration tests pass locally so not sure why that failed. The last big one seems to be the benchmark showing a regression, will be slow to respond until the week.

ion-elgreco (approved):
Thanks @liamphmurphy!

In ideal world delta-spark would just be able to parse any commit info value to string, but this also works :D

PR #2325, opened 2024-03-23
fix: remove predicate pushdown in scan when predicate doesn't contain partition column

ion-elgreco:
# Description
For now we can remove the predicate pushdown if it's not a partition column, before it would filter the target data and result in lossed data.

It has some performance impact, but at least it's correct now.


ion-elgreco:
Supersed by this PR: https://github.com/delta-io/delta-rs/pull/2326

Issue #2158, opened 2024-02-01
When_matched_update causes records to be lost with  explicit predicate


sebdiem:
# Environment

**Delta-rs version**: 0.15.0

**Binding**: python

**Environment**:
- **Cloud provider**:
- **OS**: OSX, linux
- **Other**:

***
# Bug

**What happened**:

I have  a table with several columns, among them `instance_id` and `cost`.
I want to update rows for which cost is null based on data contained in a pandas dataframe.
For this purpose I use a merge operation like this:

```
    delta_table.merge(
        pa.Table.from_pandas(costs),
        predicate="t.instance_id = s.instance_id and t.cost is null",
        source_alias="s",
        target_alias="t",
    ).when_matched_update({"cost": "s.cost"}).execute()
```

In 0.14.0 it worked perfectly fine and rows for which cost was already set stayed untouched while other were updated.
In 0.15.0 however the behavior is different: the rows with cost set **were deleted**, the other were updated.

**What you expected to happen**:
Same behavior as in 0.14.0

**How to reproduce it**:
Here is a minimal python script with the following dependencies:
```
pandas==1.4.4
pyarrow==15.0.0
deltalake==0.15.0
```

```python
import os
import shutil

import pandas as pd
import pyarrow as pa
from deltalake import DeltaTable, write_deltalake
from deltalake.schema import (
    Field,
    PrimitiveType,
    Schema,
)

# first create a table with the given schema
path = "/tmp/delta-rs-bug-3"
shutil.rmtree(path)
os.makedirs(path)
schema = Schema(
    [
        Field("instance_id", PrimitiveType("string"), nullable=False),
        Field("month", PrimitiveType("string"), nullable=False),
        Field("cost", PrimitiveType("float"), nullable=True),
    ]
)
write_deltalake(
    path,
    data=[],
    schema=schema,
    name="test_table",
    partition_by=["month"],
    overwrite_schema=False,
)
table = DeltaTable(path)

# then fill this table with some entries
initial_df = pd.DataFrame(
    [
        {"instance_id": "1", "month": "2024-02", "cost": 32.0},
        {"instance_id": "2", "month": "2024-02", "cost": None},
    ]
)
table.merge(
    pa.Table.from_pandas(initial_df),
    predicate="t.instance_id = s.instance_id",
    source_alias="s",
    target_alias="t",
).when_not_matched_insert_all().execute()

# now try to merge new data into this table: boum
new_df = pd.DataFrame(
    [
[...]

PR #2326, opened 2024-03-23
fix: merge pushdown handling

Blajda:
# Description
Fix broken test case with partitions

- fixes https://github.com/delta-io/delta-rs/issues/2158


ion-elgreco (approved):
Awe-some!

PR #2327, opened 2024-03-24
feat: implement transaction identifiers

### LINKED ISSUE #2130: Transaction Identifiers
# Description
Implement [Transaction Identifiers](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers) as outlined in the protocol.

Any user operation that commits to the log will have the ability to specify this attribute at MOST once (TBC).
The delta protocol does not specify any semantics for `version` so a hook should be provided to end users that will be called during conflict resolution. This will allow users to abort the transaction if the application semantics for `version` are violated.

**Related Issue(s)**

Blajda:
# Description
Enable users to read transactions identifiers on table and allows writing of transactions identifiers for all operations.

# Related Issue(s)
- closes #2130


Blajda:
@roeap Please provide some feedback on how I'm interacting with the log replay.
Transaction Identifiers are at the `DeltaTableState` level which encapsulates a `EagerSnapshot` however `EagerSnapshot` is responsible for using `LogReplayStream` for reading checkpoint and commit files. Either I move the application transcription into the eager snapshot, read the same files twice, or implement some `hook` for when the replay is being performed. 

I implemented the hook with a visitor trait called `ReplayVisitor` which can be specified whenever a new `EagerSnapshot` is being created or whenever states are being updated merged. This will allow information from checkpoints & commits to flow upwards depending on what feature are being used.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -337,6 +339,20 @@ impl Snapshot {
...
+/// TODO!
+pub trait ReplayVistor: Send {
+    /// TODO!
+    fn visit_batch(&mut self, batch: &RecordBatch) -> DeltaResult<()>;
+}
+
+struct PrintVistor {}
+impl ReplayVistor for PrintVistor {
+    fn visit_batch(&mut self, _batch: &RecordBatch) -> DeltaResult<()> {
+        println!("Hello world!");
```
rtyler, line 351: :question:

rtyler (commented):
I like exposing Txn :smile:  kafka-delta-ingest is the only application I've seen in real life that really makes heavy use of them :smile:

crates/core/src/kernel/snapshot/mod.rs
Blajda: It's still WIP / draft. Opened this get some feedback on the new mechanism of implementing a visitor pattern for hooking into the log replay.

roeap:
@Blajda - very sorry for the long delay! Will do a review now.

crates/core/src/kernel/snapshot/log_segment.rs
```diff
@@ -30,9 +30,12 @@ lazy_static! {
     pub(super) static ref COMMIT_SCHEMA: StructType = StructType::new(vec![
         ActionType::Add.schema_field().clone(),
         ActionType::Remove.schema_field().clone(),
+        ActionType::Txn.schema_field().clone(),
+    ]);
+    pub(super) static ref CHECKPOINT_SCHEMA: StructType = StructType::new(vec![
+        ActionType::Add.schema_field().clone(),
+        ActionType::Txn.schema_field().clone(),
```
roeap, line 37: would it maybe make sense to make this configurable? The goal is to be as minimal as possible when reading the log and eventuially update the parquet reader to leverage push-down as best we can.

We will always require add / remove but can maybe read txns and others based on config. Another alternative would maybe be to just read the transactions separately and in another PR add a simple caching layer for reading especially the json files?

What do you think?

crates/core/src/table/state.rs
```diff
@@ -3,23 +3,75 @@
...
+
+        for idx in 0..id.len() {
+            let app = ex::read_str(id, idx)?;
+            let version = ex::read_primitive(version, idx)?;
+
+            self.app_transaction_version.insert(app.to_owned(), version);
+        }
+
+        Ok(())
+    }
```
roeap, line 72: Never actually worked with the transaction IDs, but I think we may have to insert transactrion Ids only when its not yet inserted. Reason being that we replay the log from highes commit to lowest.

roeap (commented):
Looking good, and I do like the visitor pattern to extract more actions during replay.

There are some thoughts I had during review that I am keen to ghet your opinion on.

PR #2329, opened 2024-03-24
feat(rust): derive Copy on some public enums

lasantosr:
# Description
This PR derives the `Copy` trait on public enums that had the `Clone` trait derived, when applicable.


crates/core/src/operations/writer.rs
```diff
@@ -386,7 +386,7 @@ impl PartitionWriter {
                 self.flush_arrow_writer().await?;
             }
         }
-        self.arrow_writer.flush();
+        self.arrow_writer.flush()?;
```
lasantosr, line 389: I've also included this non-related change, because the arrow result was not being used, but it should.

crates/core/src/operations/merge/mod.rs
```diff
@@ -997,7 +997,7 @@ async fn execute(
 
     let source_schema = source.schema();
     let target_schema = target.schema();
-    let join_schema_df = build_join_schema(source_schema, &target_schema, &JoinType::Full)?;
+    let join_schema_df = build_join_schema(source_schema, target_schema, &JoinType::Full)?;
```
ion-elgreco, line 1000: Why is this not a reference anymore?

lasantosr: That's a clippy warning, because `target.schema()` already [returns a reference](https://docs.rs/datafusion/latest/datafusion/logical_expr/enum.LogicalPlan.html#method.schema), so that double reference was being automatically dereferenced.

ion-elgreco (approved):
Thanks! @lasantosr

Issue #2342, opened 2024-03-25
Javascript (server side) bindings via n-api


universalmind303:
# Description

I'd like a similar sdk as the python sdk but available in server side js (node, deno, bun, ...).  The sdk should closely resemble the python one, and only deviate when either necessary due to napi limitations, or when it is unidiomatic in JS.

[napi.rs](https://napi.rs/) is very similar in nature to pyo3 and is relatively easy to use. 

**Use Case**
I want to read delta files via a js application.

**Related Issue(s)**

https://github.com/pola-rs/nodejs-polars/issues/176

universalmind303:
@rtyler I'd be happy to contribute to this one. Feel free to assign it to me.

ion-elgreco:
@universalmind303 I tried creating a reader with polars rust directly but the Scan is not flexible enough to read schema evolved table unfortunately : (

universalmind303:
> @universalmind303 I tried creating a reader with polars rust directly but the Scan is not flexible enough to read schema evolved table unfortunately : (
> 
> So curious to see how you are planning to do this with polars 😄

for polars, there is a way to push down some operations to the datasource during planning. `AnonymousScan`. It doesn't support all of the pushdowns, but it can likely be improved to support all of the same pushdowns as parquet. The projection pushdowns should handle schema evolution automatically for you.

ion-elgreco:
@universalmind303 I tried the a LogicalPlan::scan for parquet but polars will error out during collection, since it concats the dfs instead of doing a diagonal concat. Also if you provide a schema to the reader and a column is not in the parquet, it will error out instead of creating a null array

TheKnightCoder:
What is the update on this? Would also love this to use delta with nodejs

universalmind303:
> What is the update on this? Would also love this to use read/write delta with nodejs , huge unlock with the amount of js devs imo.

I had an initial POC for it [here](https://github.com/delta-io/delta-rs/pull/2345), but since then, the python API has changed quite a bit. I made a bit of progress getting it to feature parity with the python one, but not quite there yet. 

I'm hoping to find some time to work on this again soon. If anyone else wants to work on it in the meantime, by all means!

PR #2345, opened 2024-03-26
draft: js bindings for deltalake

universalmind303:
# Description

# Related Issue(s)
closes https://github.com/delta-io/delta-rs/issues/2342

# Documentation


rtyler:
Nice start @universalmind303 ! I would recommend we split this into a separate repository to evolve independently like @mightyshazam has started with [delta-dotnet](https://github.com/delta-incubator/delta-dotnet). That will allow iteration without needing to take on all the CI and other processes we have out of necessity in this project.

My recommendation would be to start a new repo with this work, and then we can suggest it for the inclusion in as an incubating project if you'd like

universalmind303:
> My recommendation would be to start a new repo with this work, and then we can suggest it for the inclusion in as an incubating project if you'd like

sure that sounds good. Ill close this PR, and continue work on the bindings in a separate repo.

PR #2346, opened 2024-03-27
fix: make struct fields nullable in stats schema

qinix:
# Description

Currently only top level fields are mapped to nullable=true in stats schema. However, delta-spark generated stats may have null stats fields even if origin data fields are declared as NOT NULL. So stats for columns nested within a struct should also be mapped to nullable=true as well.


rtyler:
hm interesting, would you be able to add a test case or example stats to the pull request that we can incorporate as a test to ensure that nullable nested fields continue to be ssupported

rtyler (approved):

Issue #2344, opened 2024-03-26
Cannot merge into table partitioned by date type column on 0.16.3


echai58:
# Environment

**Delta-rs version**: 0.16.3

**Binding**: python

***
# Bug

**What happened**:
After upgrading to 0.16.3 from 0.15.3, I'm seeing a `DeltaError: Generic DeltaTable error: Unable to convert expression to string` error when trying to merge into a table that is partitioned by a column of type `pa.date32()`. If I don't partition by date, or partition by a  column of string/int type, the merge succeeds. So it is something to do with date type partitions. 

**What you expected to happen**:
This used to work, so I would expect this behavior to not break in later versions. I'm not sure if this degradation came from 0.16.3 specifically, or some version in between 0.15.3 and now.

**How to reproduce it**:

```
import pyarrow as pa
import pandas as pd
from deltalake import DeltaTable

schema = pa.schema(
    [
        ("date", pa.date32()),
        ("k", pa.string()),
        ("v", pa.string()),
    ]
)

DeltaTable.create("test", schema=schema, partition_by=["date"])

t = pd.DataFrame.from_dict(
    {
        "date": [pd.to_datetime("2022-02-01")],
        "k": ["a"],
        "v": ["b"],
    }
)

dt.merge(
    t,
    predicate='s.date = t.date',
    source_alias="s",
    target_alias="t",
).when_matched_update_all().when_not_matched_insert_all().execute()
```

The above is a minimal reproduction, leading to the following error `DeltaError: Generic DeltaTable error: Unable to convert expression to string`.

**More details**:

ion-elgreco:
@echai58 would you be able to check at what version this stopped working?

echai58:
> @echai58 would you be able to check at what version this stopped working?

@ion-elgreco Yes, just went through versions and it started breaking on 0.16.2, worked on versions before that.

ion-elgreco:
@echai58 great, thanks for checking! I'll take a look in the next couple days, it may have to do with the early filter construction

echai58:
@ion-elgreco thanks for working on a fix for this! Do you have plans to do a new version release with this fix soon? It's blocking me upgrading right now :slightly_smiling_face:

ion-elgreco:
> @ion-elgreco thanks for working on a fix for this! Do you have plans to do a new version release with this fix soon? It's blocking me upgrading right now :slightly_smiling_face: 

Somewhere next week I guess

echai58:
[...]

PR #2357, opened 2024-03-28
fix(python, rust): expr parsing date/timestamp

ion-elgreco:
# Description
We weren't parsing all scalar values yet, parses date32/64 and timestampmicros now as well.

# Related Issue(s)
- fixes https://github.com/delta-io/delta-rs/issues/2344


crates/core/src/delta_datafusion/expr.rs
```diff
@@ -339,6 +343,36 @@ impl<'a> fmt::Display for ScalarValueFormat<'a> {
             ScalarValue::UInt16(e) => format_option!(f, e)?,
             ScalarValue::UInt32(e) => format_option!(f, e)?,
             ScalarValue::UInt64(e) => format_option!(f, e)?,
+            ScalarValue::Date32(e) => match e {
+                Some(e) => {
+                    let dt =
+                        NaiveDate::from_num_days_from_ce_opt((EPOCH_DAYS_FROM_CE + (*e)).into())
+                            .unwrap();
```
hntd187, line 350: Maybe do a `.map_err(...)?` so we can capture errors here?

crates/core/src/delta_datafusion/expr.rs
```diff
@@ -339,6 +343,36 @@ impl<'a> fmt::Display for ScalarValueFormat<'a> {
...
+                    write!(f, "{}", dt)?
+                }
+                None => write!(f, "NULL")?,
+            },
+            ScalarValue::Date64(e) => match e {
+                Some(e) => write!(
+                    f,
+                    "{}",
+                    NaiveDateTime::from_timestamp_millis((*e).into())
+                        .unwrap()
```
hntd187, line 360: and here

crates/core/src/delta_datafusion/expr.rs
```diff
@@ -339,6 +343,36 @@ impl<'a> fmt::Display for ScalarValueFormat<'a> {
...
+                        .date()
+                )?,
+                None => write!(f, "NULL")?,
+            },
+            ScalarValue::TimestampMicrosecond(e, tz) => match e {
+                Some(e) => match tz {
+                    Some(tz) => write!(
+                        f,
+                        "{}",
+                        NaiveDateTime::from_timestamp_micros(*e).unwrap().and_utc()
```
hntd187, line 370: and here :-)

crates/core/src/delta_datafusion/expr.rs
```diff
@@ -339,6 +343,36 @@ impl<'a> fmt::Display for ScalarValueFormat<'a> {
...
+                    )?,
+                    None => write!(f, "{}", NaiveDateTime::from_timestamp_micros(*e).unwrap())?,
```
hntd187, line 372: and finally here :-)

ion-elgreco: @hntd187 it actually is not an error but it returns Option<NaiveDateTime> when the value that is passed is out of range.. This generally shouldn't happen but I can return an error when it's None though.

Blajda (commented):
The existing unit tests should be extended to ensure we can round trip these new types. https://github.com/delta-io/delta-rs/pull/2357/files#diff-8e1a60f799cd1ad29923ee75871099a19c15ddfc00bef4617d82bdc02cbe3198R523-R526

ion-elgreco:
@Blajda that datafusion sql parser is for some reason parsing a date/timestamp as a substraction : S I also tried wrapping it in single quotes but then it gets parsed as a string. So I am not sure what format datafusion expects datetimes to be

```rust
 left: BinaryExpr(BinaryExpr { left: Column(Column { relation: None, name: "_timestamp_ntz" }), op: Gt, right: Literal(TimestampMicrosecond(1262304000000000, None)) })
 right: BinaryExpr(BinaryExpr { left: Column(Column { relation: None, name: "_timestamp_ntz" }), op: Gt, right: BinaryExpr(BinaryExpr { left: BinaryExpr(BinaryExpr { left: Literal(Int64(2010)), op: Minus, right: Literal(Int64(1)) }), op: Minus, right: Literal(Int64(1)) }) })
```

Blajda:
Yeah you can't dump it directly. The query dialect is postgress so you if you are using a string you can try the cast notation i.e '::date'. Alternative approach use [the date functions](https://arrow.apache.org/datafusion/user-guide/sql/scalar_functions.html#to-timestamp-micros) and convert the scalars to their correct representations.

ion-elgreco:
@Blajda yeah but then Im not really testing round tripping anymore. 

Im assuming the string format needs to be in the same format as the serialization format for the conflict resolution across other engine writers

Blajda:
There's no specification on the query format for conflict resolution with other engines. At least for our engine any queries we convert to a string we should be able to read it back without error

Blajda (approved):

Issue #2381, opened 2024-04-04
Allow checkpoint creation when partion column is "timestampNtz "


thomasfrederikhoeck:
# Description
Right now you can create a table partioned on a column of type timestampNtz. This is all on main.:

```
import pandas as pd
dates = pd.date_range(datetime(2021,1,1,3,4,6,3),datetime(2021,1,3,3,4,6))
df = pd.DataFrame({"time":dates, "a":[i for i in range(len(dates))]})
write_deltalake("mytable",df, partition_by="time")
dt = DeltaTable("mytable")
print(dt.schema())
> Schema([Field(time, PrimitiveType("timestampNtz"), nullable=True), Field(a, PrimitiveType("long"), nullable=True)])

```

However when trying to create a checkpoint I'm hit by

```

dt.create_checkpoint()
> PanicException: not implemented: Primitive type timestampNtz is not supported for partition column values.

```
**Use Case**

**Related Issue(s)**

ion-elgreco:
@thomasfrederikhoeck We just need to add an arm for PrimitiveType::TimestampNtz`

https://github.com/delta-io/delta-rs/blob/6f81b8034dbef7a0120e354e178f0c859564465e/crates/core/src/protocol/checkpoints.rs#L444

I'll see if I can find some time tomorrow

thomasfrederikhoeck:
GIven https://github.com/delta-io/delta-rs/issues/2380 it might also be worth looking at the `Timestamp` arm?

thomasfrederikhoeck:
@ion-elgreco I can grap this one :-)

Issue #2380, opened 2024-04-04
ValueError: Partition value cannot be parsed from string.


thomasfrederikhoeck:
# Environment

**Delta-rs version**:
Main including https://github.com/delta-io/delta-rs/commit/6f81b8034dbef7a0120e354e178f0c859564465e
**Binding**: 
python

**Environment**:
- **Cloud provider**: local
- **OS**: Windows
- **Other**:

***
# Bug

**What happened**:
When I try to create a checkpoint on a table partioned by timestamp I'm hit with a `ValueError. Note that I have build from master including` https://github.com/delta-io/delta-rs/pull/2357:

```python
import pandas as pd
from datetime import datetime
import deltalake as dl

dates = pd.date_range(datetime(2021,1,1,3,4,6,3),datetime(2021,1,3,3,4,6))

df = pd.DataFrame({"time":dates, "a":[i for i in range(len(dates))]})

schema = dl.schema.Schema(fields=[
    dl.schema.Field("time",dl._internal.PrimitiveType.from_json('"timestamp"')),
    dl.schema.Field("a",dl._internal.PrimitiveType.from_json('"integer"'))
    ]
) 

write_deltalake("mytable",df, schema=schema,partition_by="time")
dt = DeltaTable("mytable")
dt.create_checkpoint()
```
which gives:
```
ValueError: Partition value 2021-01-02 03:04:06.000003Z cannot be parsed from string.

```
**What you expected to happen**:
That the checkpoint was created.
**How to reproduce it**:
Run code above

**More details**:

thomasfrederikhoeck:
it appears that the `Z` is missing in the parsing:
https://github.com/delta-io/delta-rs/blob/6f81b8034dbef7a0120e354e178f0c859564465e/crates/core/src/protocol/checkpoints.rs#L444

ion-elgreco:
@thomasfrederikhoeck it shouldn't write the partition values with a Z. Also my PR didn't touch the partition value serialization.

thomasfrederikhoeck:
Yes it appears that pyarrow serialize `timestamp` with `Z` while and `timestampNtz` without.

```python
import pyarrow as pa
import pytz

tz = "UTC"

def get_data(with_tz):
    tzinfo = pytz.timezone(tz) if  with_tz else None
    dates = pd.date_range(
        datetime(2021,1,1,3,4,6,3, tzinfo=tzinfo),
[...]

PR #2383, opened 2024-04-04
fix(rust): timestamp deserialization format, missing type

ion-elgreco:
# Description
Our timestamp deserialization format didn't include the %6f to decode this value: 1970-01-01 00:00:00.123456. Additionally during timestampNtz I didn't add deserialization of that primitive type :) 

- fixes https://github.com/delta-io/delta-rs/issues/2380
- fixes https://github.com/delta-io/delta-rs/issues/2381

### DIFF
--- a/crates/core/src/delta_datafusion/expr.rs
+++ b/crates/core/src/delta_datafusion/expr.rs
@@ -27,7 +27,7 @@ use std::{
 };
 
 use arrow_schema::DataType;
-use chrono::{Date, NaiveDate, NaiveDateTime, TimeZone};
+use chrono::{DateTime, NaiveDate};
 use datafusion::execution::context::SessionState;
 use datafusion_common::Result as DFResult;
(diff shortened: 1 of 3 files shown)


crates/core/src/delta_datafusion/expr.rs
```diff
@@ -27,7 +27,7 @@ use std::{
 };
 
 use arrow_schema::DataType;
-use chrono::{Date, NaiveDate, NaiveDateTime, TimeZone};
+use chrono::{DateTime, NaiveDate};
```
cmackenzie1, line 30: Why the switch to `DateTime` from `NaiveDateTime`?

cmackenzie1: Cool, thanks for claryfying @ion-elgreco. Would be great to capture these kinds of changes in the commit message when making them.

ion-elgreco: @cmackenzie1 I'll try to make more meaningful commit messages : P

rtyler (approved):

Issue #913, opened 2022-11-03
Create Checkpoints based on table config when writing data


roeap:
# Description

Delta allows configuring at which commit intervals a new checkpoint should be created (default 10). We should honor this config.

**Use Case**

Checkpoints are an important performance optimization in the delta log.

**Related Issue(s)**

ion-elgreco:
@roeap I guess this requires a small check after each operation to see the if the current commit is greater than then the interval + last checkpoint commit. 

Do we store somewhere this checkpoint version on the table as an attribute? When I do dt.metadata() I always get no results.

ion-elgreco:
Never mind, found how to write the table properties during write :)

roeap:
> Do we store somewhere this checkpoint version on the table as an attribute? When I do dt.metadata() I don't see this info.

This information is not tracked, and I think we also should not :) - i.e. other writers are not adding this metadata, so we could not rely on it.

IMO, we could just do a quick check if `_last_checkpoint` is present, and if not, revert to `version % 10 == 0`.

ion-elgreco:
@ldacey from the table configuration. You can grab it from the metadata of the table

roeap:
@ldacey, as @ion-elgreco points out, there is a configuration for delta-tables, namely `delta.checkpointInterval` which controls at which interval checkpoints should be created. delta-rs does not honour that yet.

In your scenario - assuming accounts means independent writers - I recommend having a dedicated job to create checkpoints. Checkpoint creation in the current spec does suffer from potential race conditions if multiple writers try to create teh same checkpoint. This will be mitigated once we have v2 checkpoints implemented, which is currently ongoing work...

PR #2390, opened 2024-04-06
feat(python,rust): create checkpoint on interval from config

ion-elgreco:
# Description
This creates a checkpoint if it meets the condition of the interval.

Also bumping the default interval to 100 commits. 10 commits can be a bit aggressive

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/913

### DIFF
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -65,17 +65,18 @@
 //!       └───────────────────────────────┘                   
 //!</pre>
 
-use std::collections::HashMap;
-
 use chrono::Utc;
 use conflict_checker::ConflictChecker;
 use futures::future::BoxFuture;
 use object_store::path::Path;
 use object_store::{Error as ObjectStoreError, ObjectStore};
 use serde_json::Value;
+use std::collections::HashMap;
 
 use self::conflict_checker::{CommitConflictError, TransactionInfo, WinningCommitSummary};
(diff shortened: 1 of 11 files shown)


crates/core/src/operations/transaction/mod.rs
```diff
@@ -429,25 +458,96 @@ impl<'a> PreCommit<'a> {
...
+        if self.create_checkpoint {
+            if let Some(table) = table {
+                let last_checkpoint =
+                    read_last_checkpoint(log_store.object_store().as_ref(), log_store.log_path())
+                        .await?;
+
+                let checkpoint_interval = table.config().checkpoint_interval() as i64;
+
+                if let Some(last_checkpoint) = last_checkpoint {
+                    if (version - last_checkpoint.version) >= checkpoint_interval {
```
Blajda, line 497: Suppose you have multiple writers appending to the table each second and the time it takes for the checkpoint to be written is 5 seconds. Because there is a lag for the checkpoint being visible the writers will create checkpoints for their version at the same time. The modulo logic will prevent this for occuring.

ion-elgreco: Right.. didn't think about that one. I'll switch to using modulo then .

ion-elgreco: @Blajda do you have any preference on the design? This PR or the V2 PR?

PR #2391, opened 2024-04-06
feat(rust): post commit hook (v2), create checkpoint hook

ion-elgreco:
# Description
Introduces a post commit, which can do additional actions before returning the FinalizedCommit.

Current commit hook will creates a checkpoint if it meets the condition of the interval.

Also bumping the default interval to 100 commits. 10 commits can be a bit aggressive

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/913

### DIFF
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -76,7 +76,6 @@ use std::collections::HashMap;
 use self::conflict_checker::{CommitConflictError, TransactionInfo, WinningCommitSummary};
 use crate::checkpoints::create_checkpoint_for;
 use crate::errors::DeltaTableError;
-use crate::kernel::log_segment::read_last_checkpoint;
 use crate::kernel::{
     Action, CommitInfo, EagerSnapshot, Metadata, Protocol, ReaderFeatures, WriterFeatures,
 };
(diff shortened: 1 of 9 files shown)


crates/core/src/operations/transaction/mod.rs
```diff
@@ -293,19 +302,27 @@ impl CommitData {
     }
 }
 
+#[derive(Clone, Debug, Copy)]
+/// Properties for post commit hook.
+pub struct PostCommitHookProperties {
+    create_checkpoint: bool,
+}
+
```
Blajda, line 310: I think we can remove this struct and just keep the config in the commit builder and commit properties for now.

ion-elgreco: I was planning to add support for log retention cleanup, auto compact and auto vacuüm afterwards. So perhaps still good to keep this then?

Blajda:
Overall it looks good to me. In the future if we are going to expand on post commit hooks then we should maybe refactor so the commit code is responsible for returning the snapshot version of that delta table that was successfully committed. This would remove the burden of updating the update from each operation too.

ion-elgreco:
@Blajda that's a good idea! 

That would essentially solve the issue for concurrent writers where sometimes it tries to advance the state after a commit, but the delta of the commit version and the table version is greater than 1

crates/core/src/operations/transaction/mod.rs
Blajda: Sound good in that case.

Blajda (approved):

PR #2396, opened 2024-04-07
feat(rust): advance state in post commit

### LINKED ISSUE #2262: Generic DeltaTable error: Version mismatch with new schema merge functionality in AWS S3

### LINKED ISSUE #2279: Successful writes return error when using concurrent writers

ion-elgreco:
# Description
We advance the state in the post commit now, so it's done in a single location as per suggestion from @Blajda here: https://github.com/delta-io/delta-rs/pull/2391#issuecomment-2041500757

This PR also supersedes this one: https://github.com/delta-io/delta-rs/pull/2280

# Related Issue(s)
- fixes #2279
- fixes #2262

### DIFF
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -595,36 +595,45 @@ pub struct PostCommit<'a> {
 
 impl<'a> PostCommit<'a> {
     /// Runs the post commit activities
-    async fn run_post_commit_hook(
-        &self,
-        version: i64,
-        commit_data: &CommitData,
-    ) -> DeltaResult<()> {
-        if self.create_checkpoint {
-            self.create_checkpoint(&self.table_data, &self.log_store, version, commit_data)
-                .await?
-        }
-        Ok(())
-    }
-    async fn create_checkpoint(
-        &self,
-        table: &Option<&'a dyn TableReference>,
-        log_store: &LogStoreRef,
-        version: i64,
-        commit_data: &CommitData,
-    ) -> DeltaResult<()> {
-        if let Some(table) = table {
-            let checkpoint_interval = table.config().checkpoint_interval() as i64;
-            if ((version + 1) % checkpoint_interval) == 0 {
-                // We have to advance the snapshot otherwise we can't create a checkpoint
-                let mut snapshot = table.eager_snapshot().unwrap().clone();
-                snapshot.advance(vec![commit_data])?;
+    async fn run_post_commit_hook(&self) -> DeltaResult<Option<DeltaTableState>> {
+        if let Some(table) = self.table_data {
+            if let Some(mut snapshot) = table.eager_snapshot().cloned() {
+                if self.version - snapshot.version() > 1 {
+                    // This may only occur during concurrent write actions. We need to update the state first to - 1
+                    // then we can advance.
+                    snapshot
+                        .update(self.log_store.clone(), Some(self.version - 1))
+                        .await?;
+                    snapshot.advance(vec![&self.data])?;
+                } else {
+                    snapshot.advance(vec![&self.data])?;
+                }
                 let state = DeltaTableState {
                     app_transaction_version: HashMap::new(),
                     snapshot,
                 };
-                create_checkpoint_for(version, &state, log_store.as_ref()).await?
+                // Execute each hook
+                if self.create_checkpoint {
+                    self.create_checkpoint(&state, &self.log_store, self.version)
+                        .await?;
+                }
+                return Ok(Some(state));
+            } else {
+                return Ok(None);
             }
+        } else {
+            return Ok(None);
+        }
+    }
+    async fn create_checkpoint(
+        &self,
+        table_state: &DeltaTableState,
+        log_store: &LogStoreRef,
+        version: i64,
+    ) -> DeltaResult<()> {
+        let checkpoint_interval = table_state.config().checkpoint_interval() as i64;
+        if ((version + 1) % checkpoint_interval) == 0 {
+            create_checkpoint_for(version, table_state, log_store.as_ref()).await?
         }
         Ok(())
     }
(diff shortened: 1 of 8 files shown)


Blajda:
@ion-elgreco I've looked a this quick a bit and feel like a deep refactor is required for this. IMO if the commit logic is going to be responsible for providing the committed then it must return return a DeltaTableState / Snapshot. What are the expected scenarios where that struct  will be `None`? 
I also think it will be worth having the commit code take ownership of the passed Snapshot so a clone can be avoided.
Let me know what you think.

ion-elgreco:
@Blajda Yeah it felt a bit strange to not be able to take ownership of the snapshot. 

While working on it I didn't understand why the incoming snapshot could be None until I hit the create path, which obviously has no table snapshot.

We should just return always a snapshot. If the incoming snapshot is None, we can just create a snapshot after committing and return that. Do you agree?

crates/core/src/operations/transaction/mod.rs
```diff
@@ -595,59 +589,72 @@ pub struct PostCommit<'a> {
...
-            self.create_checkpoint(&self.table_data, &self.log_store, version, commit_data)
-                .await?
+    async fn run_post_commit_hook(&self) -> DeltaResult<DeltaTableState> {
+        if let Some(table) = self.table_data {
+            let mut snapshot = table.eager_snapshot().clone();
+            if self.version - snapshot.version() > 1 {
+                // This may only occur during concurrent write actions. We need to update the state first to - 1
+                // then we can advance.
+                snapshot
+                    .update(self.log_store.clone(), Some(self.version - 1))
```
JonasDev1, line 599: What if there was more than one concurrent write?

ion-elgreco: It doesn't matter how many there are. This just simply updates the current table state to version - 1, so that if you then advance with the commit data, you will get state.version == commit.version

Blajda (approved):
LGTM! Thanks for making the changes.

PR #2421, opened 2024-04-15
feat: implement repartitioned for DeltaScan

jkylling:
# Description
This implements repartitioned from the ExecutionPlan trait of DeltaScan. Currently, Delta tables without partitions are read with all its files in a single file group of the underlying `ParquetExec`. This seems to mean that Delta tables without partitions are read without concurrency. With repartitioned we can repartition the DeltaScan to get concurrency when reading.


ion-elgreco (approved):
Thanks!

PR #2424, opened 2024-04-16
feat: lazy static runtime in python

ion-elgreco:
# Description
As suggested by @wjones127 to create a lazy static runtime, supersedes this PR: https://github.com/delta-io/delta-rs/pull/1950


rtyler (approved):

ion-elgreco:
@rtyler thanks for pointing that out! :) didn't know that existed, let me make the change

wjones127 (dismissed):

ion-elgreco:
@rtyler @wjones127 I changed it to use OnceLock and refactored the lifetime in other areas where the runtime is used

python/src/filesystem.rs
```diff
@@ -25,7 +24,7 @@ pub(crate) struct FsConfig {
 #[derive(Debug, Clone)]
 pub struct DeltaFileSystemHandler {
     pub(crate) inner: Arc<DynObjectStore>,
-    pub(crate) rt: Arc<Runtime>,
+    pub(crate) rt: Arc<&'static Runtime>,
```
wjones127: Since it's a static, I don't think it even needs to hold this as part of the struct. Can just reference the global static in the methods.

ion-elgreco: @wjones127 yeah this is already changed :P I was committing a bit to fast here

ion-elgreco: Or you mean to not pass it through the struct?

ion-elgreco: @wjones127 I removed it from the structs

wjones127 (approved):
Looks great!

Issue #2427, opened 2024-04-20
File skipping according to documentation


t1g0rz:
# Environment

**Delta-rs version**: 0.16.4

**Binding**:  python

# Description

[Here](https://github.com/delta-io/delta-rs/blob/ebbdd69274e8d22055f47aca03bf56f997304cee/docs/delta-lake-best-practices.md?plain=1#L95) documentation says:  
> Ensure the transaction log stores metadata stats for all the columns that benefit from file skipping.

And [here](https://github.com/delta-io/delta-rs/blob/ebbdd69274e8d22055f47aca03bf56f997304cee/docs/delta-lake-best-practices.md?plain=1#L103) guides us to explicitly specify columns for file skipping:  
> It takes some time to compute column statistics when writing files, and it isn’t worth the effort if you cannot use the column for file skipping.
> Suppose you have a table column containing a long string of arbitrary text.  It’s unlikely that this column would ever provide any data-skipping benefits.  So, you can just avoid the overhead of collecting the statistics for this particular column.

I couldn't find a way to exclude some columns from file skipping purposes. In my case, I have quite wide tables (> 200 columns), and out of these, 195 will never be used for file skipping.  
An unintended consequence of including all these columns in statistics calculation is the explosive growth of the delta log size because it writes very long strings of min/max/nulls. This indirectly creates all conditions for issue #2301 and [this from delta slack](https://delta-users.slack.com/archives/C013LCAEB98/p1711921602281259?thread_ts=1711921242.418959&cid=C013LCAEB98).

I found the `delta.dataSkippingNumIndexedCols` setting [here](https://docs.delta.io/latest/table-properties.html)  and I wonder if it's possible to explicitly specify column names. Or should I reorder the table to have skipping columns first and then set delta.dataSkippingNumIndexedCols?

ion-elgreco:
@t1g0rz I checked, but we don't respect it yet 😄 

Buttt I am working on a fix 🕺

PR #2428, opened 2024-04-21
feat(python, rust): respect column stats collection configurations

ion-elgreco:
# Description
All of the Rust and Python write actions will now properly adhere to the configuration regarding the amount of columns stats have to be collected for. Either by dataSkippingNumIndexedCols or dataSkippingStatsColumns.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2427


mightyshazam (approved):

crates/core/src/writer/stats.rs
```diff
@@ -1,7 +1,3 @@
-use std::sync::Arc;
```
roeap, line 1: while still on nightly i think, `rustfmt` has an [opinion](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports) - i.e. the prior ordering would be the idiomatic one.

crates/core/src/writer/stats.rs
```diff
@@ -76,7 +85,21 @@ fn stats_from_file_metadata(
...
+        if num_indexed_cols == -1 {
+            schema_descriptor.num_columns()
+        } else if num_indexed_cols >= 0 {
+            min(num_indexed_cols as usize, schema_descriptor.num_columns())
+        } else {
+            return Err(DeltaWriterError::DeltaTable(DeltaTableError::Generic(
+                "delta.dataSkippingNumIndexedCols valid values are >=-1".to_string(),
+            )));
+        }
+    };
```
roeap, line 100: would it be a bit clearer if we just create the names or indexes of the columns we want to collect stats for here, rather then introducing additional branching below?

python/deltalake/writer.py
```diff
@@ -264,6 +265,24 @@ def write_deltalake(
...
+                if str_indexed_cols is not None:
+                    num_indexed_cols = int(str_indexed_cols)
+            string_stats_cols = configuration.get("delta.dataSkippingStatsColumns")
+            if isinstance(string_stats_cols, str):
+                stats_cols = string_stats_cols.split(",")
+                if stats_cols and isinstance(stats_cols, list):
+                    if not all(isinstance(inner, str) for inner in stats_cols):
+                        raise ValueError(
+                            "dataSkippingStatsColumns needs to be a comma-separated list of column names in string format. Example: 'foo,bar,baz'"
+                        )
```
roeap, line 285: is there a way for us to isolate reading the configuration in the rust side. as kernel progresses we will likely see a lot of this move there, and this way we have to keep multiple code locations in sync.

crates/core/src/writer/stats.rs
ion-elgreco: Let me check, we need to fetch the stats by idx, so I might be able to construct a vec of indexes :)

python/deltalake/writer.py
ion-elgreco: I have added the exact same code in Rust. What I can do is move that into a pub fn, that I can call from python

roeap (approved):
👍

PR #2429, opened 2024-04-21
fix(python): load_as_version with datetime object with no timezone specified

t1g0rz:
# Description
Upon attempting to retrieve the version with a datetime object, the `load_as_version` method throws a `ValueError: Failed to parse datetime string: premature end of input`.

Datetime objects without a specified timezone will be treated as UTC datetimes.


ion-elgreco:
Thanks :) @t1g0rz

ion-elgreco (approved):

PR #2435, opened 2024-04-22
chore: introduce the Operation trait to enforce consistency between operations

rtyler:
I am spinning this pull request out from some of my other in-flight work to extend behaviors with the operations.

The motivation here is to provide a base trait which can act as a standards-bearer for the operations implementations. Right now this is only enforcing `IntoFuture` but there are a couple of other changes I am actively working on which I hope will dramatically clean up these operations and enable some of the use-cases that @hntd187 and I have been exploring.


hntd187:
To David's point this is why I think a marker trait by itself it maybe not sufficient, we should eventually move towards a trait that implements and expected life cycle of an operation (or multiple smaller ones that implement a part of it) the most ready example I can think of is something similar to Hadoop's commit protocol, but we should do something that makes sense for us, I'm just using as an example of generalization of such a process.

rtyler:
@hntd187 that's certainly the plan here :smile: I just wanted to get something started we could begin hanging functionality off of

ion-elgreco (approved):

PR #2437, opened 2024-04-22
fix(python,rust): missing remove actions during `create_or_replace`

ion-elgreco:
# Description
The overwrite mode never added the remove actions, which causes your table to get in an invalid state.


wjones127 (commented):
Seems like we should add a test in Rust for this case.

ion-elgreco:
> Seems like we should add a test in Rust for this case.

I'm on it!

wjones127 (approved):
Great work!

Issue #2438, opened 2024-04-22
TableMerger - when_matched_delete() fails when Column names contain special characters


mkp-jansen:
# Environment

**Delta-rs version**: 0.16.4

**Binding**: python

**Environment**:
- **OS**: Windows 11 Business

***
# Bug

**What happened**:
At our company we are currently thinking about using delta for our sensor data. The package delta-rs provides pretty much all the functionality we need. However, for reasons I won't be able to change we often have column names with two dashes, e.g. "y--1" ("y-1" works). We need to be able to delete data from the delta lake. When using the TableMerger this fails as shown in the example below.

In the documentation it says the following:

```
Column names with special characters, such as numbers or spaces should be encapsulated in backticks: "target.123column" or "target.my column"
```
However, there is no argument in "when_matched_delete()" to specifiy the columns with special characters.

**What you expected to happen**:

I guess the desired behaviour would be that you can simply delete the matching rows, even when the column names contain special characters.

I would be happy to give a fix a shot (also in rust) - but I would need some guidance along the way.

**How to reproduce it**:

```python
from deltalake import DeltaTable, write_deltalake
import pyarrow as pa

data = pa.table({"x": [1, 2, 3], "y--1": [4, 5, 6]})
write_deltalake("tmp", data)
dt = DeltaTable("tmp")
new_data = pa.table({"x": [2, 3]})

(
    dt.merge(
        source=new_data,
        predicate='target.x = source.x',
        source_alias='source',
        target_alias='target')
    .when_matched_delete()
    .execute()
)
```

mkp-jansen:
I forgot to add the error message:

```python

---------------------------------------------------------------------------
DeltaError                                Traceback (most recent call last)
Cell In[6], line 16
      6 dt = DeltaTable("tmp")
      7 new_data = pa.table({"x": [2, 3]})
      9 (
     10     dt.merge(
     11         source=new_data,
     12         predicate='target.x = source.x',
     13         source_alias='source',
     14         target_alias='target')
     15     .when_matched_delete()
---> 16     .execute()
     17 )

File c:\...\.venv\Lib\site-packages\deltalake\table.py:1778, in TableMerger.execute(self)
   1772 def execute(self) -> Dict[str, Any]:
   1773     """Executes `MERGE` with the previously provided settings in Rust with Apache Datafusion query engine.
   1774 
   1775     Returns:
   1776         Dict: metrics
   1777     """
-> 1778     metrics = self.table._table.merge_execute(
[...]

PR #2441, opened 2024-04-22
fix(python, rust): use from_name during column projection creation

ion-elgreco:
# Description
@Blajda I don't think `from_qualified_name_ignore_case` was needed here since the delta_fields don't have relation information, they are just the column names. 

`from_qualified_name_ignore_case` will try to parse `__delta_rs_c_y--1` and results into  `__delta_rs_c_y`, while `from_name `just keeps the column name as-is, which is preferred. 

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2438


Blajda (approved):
LGTM!

PR #2453, opened 2024-04-25
fix(python): reuse table state in write engine

ion-elgreco:
# Description
Instead of reusing the table state, it was being instantiated every time you call write with the rust engine.

- https://github.com/delta-io/delta-rs/discussions/2448

### DIFF
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-python"
-version = "0.17.1"
+version = "0.17.2"
 authors = ["Qingping Hou <dave2008713@gmail.com>", "Will Jones <willjones127@gmail.com>"]
 homepage = "https://github.com/delta-io/delta-rs"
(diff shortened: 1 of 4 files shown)


adriangb:
Now we're cooking with gas!

```
Time to write tmp commit entry: 62.949541ms
Time to write final commit entry: 154.489917ms
Time taken: 0.5385329723358154
Time to write tmp commit entry: 102.603ms
Time to write final commit entry: 149.216375ms
Time taken: 0.5450408458709717
```

This immediately makes writes go a lot, lot faster 🥇!

Can we get this released asap?

ion-elgreco:
@adriangb good to see :) 

Let me see what I can do on the release.

Btw, one last question, did you see also these performance issues with the pyarrow engine?

adriangb:
I did notice the pyarrow engine being faster in some cases, this seems like a likely culprit.

wjones127 (approved):

PR #2459, opened 2024-04-28
feat(python, rust): cleanup expired logs post-commit hook

ion-elgreco:
# Description
This adds the cleanup expired logs post-commit hook. Tables by default have `delta.enableExpiredLogCleanup = true`, and` delta.logRetentionDuration = interval 30 days`. So moving forward those logs will be automatically cleaned up.

If users set `delta.enableExpiredLogCleanup = false` during table creation, the log clean up will never be executed in the post-commit. When cleanup_expired_logs in the post-commit hook properties is passed, then that take precedence over the table config. 

Additionally to most operations you can now pass in Python the PostCommitHookProperties to control the post commit hook, only really needed for advanced use cases, where you want to disable creating checkpoints across the board or override the log clean up.

@Blajda this is a follow up PR of the hooks, will look into creating an auto_vacuum_hook afterwards

### DIFF
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -364,7 +364,7 @@ impl CommitProperties {
         self
     }
 
-    /// Configure the property to clean up expired transactions
+    /// Specify if it should clean up the logs when the logRetentionDuration interval is met
     pub fn with_cleanup_expired_logs(mut self, cleanup_expired_logs: Option<bool>) -> Self {
         self.cleanup_expired_logs = cleanup_expired_logs;
         self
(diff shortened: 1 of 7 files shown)


rtyler:
I am working on resolving conflicts here for a merge

rtyler:
The copy paste needed for this change really really motivates me to work on the middleware concept more when I'm more fully online :zombie:

ion-elgreco:
> The copy paste needed for this change really really motivates me to work on the middleware concept more when I'm more fully online :zombie:

Which copy paste do you mean?

rtyler:
@ion-elgreco there are a lot of conflicts related to the handling of commit properties and post commit hook behavior that I hope we can consolidate. :crossed_fingers:

rtyler (approved):

PR #2465, opened 2024-04-30
feat(python): add  parameter to DeltaTable.to_pyarrow_dataset()

adriangb:
Otherwise there is no way to union this with another dataset.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -1030,6 +1030,7 @@ class DeltaTable:
         partitions: Optional[List[Tuple[str, str, Any]]] = None,
         filesystem: Optional[Union[str, pa_fs.FileSystem]] = None,
         parquet_read_options: Optional[ParquetReadOptions] = None,
+        schema: Optional[pyarrow.Schema] = None,
     ) -> pyarrow.dataset.Dataset:
         """
         Build a PyArrow Dataset using data from the DeltaTable.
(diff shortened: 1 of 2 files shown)


ion-elgreco:
@adriangb you might just better passthrough `as_large_types` to the `.to_pyarrow()`

adriangb:
They both seem useful right? It seems like the `as_large_types` just blindly makes all types large. I could have a mix of small and large types of which I know the schema, so passing in the schema explicitly (given that it's simple to do so) seems worth having as an option.

ion-elgreco:
@adriangb that's true. If you can fix the tests then we can merge

adriangb:
it looks like the test just fails on older pyarrow versions and only for the map type. How about I split it in two and skip the failing one on pyarrow < 10?

ion-elgreco:
@adriangb can you fix the tests? Then we can merge it :)

python/deltalake/table.py
```diff
@@ -1022,6 +1022,8 @@ def to_pyarrow_dataset(
         partitions: Optional[List[Tuple[str, str, Any]]] = None,
         filesystem: Optional[Union[str, pa_fs.FileSystem]] = None,
         parquet_read_options: Optional[ParquetReadOptions] = None,
+        schema: Optional[pyarrow.Schema] = None,
+        as_large_types: bool = False,
```
ion-elgreco, line 1026: Doc description is missing for this param. I would also mention if the schema is passed that takes precedence over as_large_types

ion-elgreco (changes requested):

python/deltalake/table.py
adriangb: done!

ion-elgreco (approved):

ion-elgreco:
Thankss @adriangb

PR #2485, opened 2024-05-07
fix(python): reuse state in `to_pyarrow_dataset`

ion-elgreco:
# Description
Reuse the state so we save time instead of reconstructing and verifying the state.

### DIFF
--- a/python/src/filesystem.rs
+++ b/python/src/filesystem.rs
@@ -5,7 +5,7 @@ use deltalake::storage::{DynObjectStore, ListResult, MultipartId, ObjectStoreErr
 use deltalake::DeltaTableBuilder;
 use pyo3::exceptions::{PyIOError, PyNotImplementedError, PyValueError};
 use pyo3::prelude::*;
-use pyo3::types::{IntoPyDict, PyBytes};
+use pyo3::types::{IntoPyDict, PyBytes, PyType};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::sync::Arc;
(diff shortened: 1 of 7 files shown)


adriangb:
Yup this fixes the issue 🥳

ion-elgreco:
> @ion-elgreco looks like it needs a rebase. Also just curious if you'll be making a release with this fix since it's such a huge performance win for an essential feature?

@wjones127 could you take a look at the PR?

python/tests/test_fs.py
```diff
@@ -241,16 +240,3 @@ def test_roundtrip_azure_decoded_sas(azurite_sas_creds, sample_data: pa.Table):
...
-    store = DeltaStorageHandler(str(tmp_path.absolute()))
-
-    with (tmp_path / "asd.pkl").open("wb") as handle:
-        pickle.dump(store, handle)
-
-    with (tmp_path / "asd.pkl").open("rb") as handle:
-        store_pkl = pickle.load(handle)
-
-    infos = store_pkl.get_file_info(["asd.pkl"])
-    assert infos[0].size > 0
```
wjones127, line 256: Are we dropping the ability to pickle?

python/src/filesystem.rs
```diff
@@ -41,21 +41,27 @@ impl DeltaFileSystemHandler {
 #[pymethods]
 impl DeltaFileSystemHandler {
     #[new]
-    #[pyo3(signature = (table_uri, options = None, known_sizes = None))]
+    #[pyo3(signature = (table_uri, table = None, options = None, known_sizes = None))]
     fn new(
-        table_uri: &str,
+        table_uri: String,
+        table: Option<&RawDeltaTable>,
```
wjones127, line 47: Maybe instead make a separate classmethod `from_table` that takes the table? Then perhaps we can keep the pickling support.

wjones127 (approved):
Looks better now. Thanks.

PR #2486, opened 2024-05-07
feat: introduce CDC write-side support for the Update operations

### LINKED ISSUE #604: Support CDC

### LINKED ISSUE #2095: Change Data Feed in Delta
# Description

Finally making an issue about this for discussion. I have a draft PR of reading CDF available already in #2048 which still needs work on it's reader side but I wanted to more discuss the writer side of it as it will take much more refactoring. I wanted to also ask how we wanted to approach this. The reader can be mostly encapsulated as it's own thing where the writer will touch all writing operations in delta lake. Do w e want to roll these into the same PR, or make subsequent PRs? I think subsequent would be better, but just my opinion.

So for reader, it's first stages are in-flight with #2048, I just need to figure out how I want to validate the correctness of this. @MrPowers maybe you can help me figure that out?

For writers, well there is a bit more to do here. CDC actions have to be added to the commit log along with the subsequent add/remove actions, generating additional change data files in the `_change_data` directory of a delta table. Currently we encapsulate the builders of these operations in such a way that the builder builds and commits all the actions itself without giving any ability for features to influence the actions of the commit before it's written. So, in order to make CDF work we would be required to update every action and add it's subsequent CDF aware functionality to the operation. I'd argue this would only exacerbate the current issue with builders owning the entire life cycle of the operation and we should not do this. 

I would instead suggest that we refactor the builders to only create and return a list of actions to commit and a snapshot to commit to. Then let a subsequent (maybe global) part of the code do the actual commit. This way you can compose operations in a more maintainable way. I spoke with @r3stl355 about this as well because some of the work he did for replaceWhere would have been another good candidate to benefit from this type of rethinking. So for writers I am proposing we take this approach.

1. Refactor builders internals to build sets of actions to commit and not perform the actual commit.
2. Add a central place to actually perform commit / write / update the table
3. Implement subsequent CDF compositions on top of the new operations

This will benefit us in the sense that CDF's implementation will have no effect on what those other operations do. Only augment them, so generally our implementations of these features will be more resilient to mistakes as we implement more features down the line. Things like row-tracking come to mind when thinking about potential issues down the line as row-tracking has a specific clause for readers regarding CDF files. Additionally checkpoints must specifically go the opposite way and remove CDF from their checkpointing. I linked these under the related issues, but hopefully that makes sense. 

**Use Case**
https://delta.io/blog/2023-07-14-delta-lake-change-data-feed-cdf/
https://docs.delta.io/latest/delta-change-data-feed.html
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-cdc-file

**Related Issue(s)**
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#reader-requirements-for-row-tracking
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#checkpoints-1

rtyler:
This change introduces a `CDCTracker` which helps collect changes during merges and update. This is admittedly rather inefficient, but my hope is that this provides a place to start iterating and improving upon the writer code

There is still additional work which needs to be done to handle table features properly for other code paths (see the middleware discussion we have had in Slack) but this produces CDC files for Update operations

Fixes #604
Fixes #2095


ion-elgreco:
I think it's better to disable it until all operations (delete and merge) are supported, otherwise we cannot push any python releases until those are added

crates/core/src/operations/update.rs
```diff
@@ -413,6 +439,48 @@ async fn execute(
...
+                    "_change_data",
+                ));
+                let mut writer = DeltaWriter::new(store, config);
+                for batch in batches {
+                    writer.write(&batch).await?;
+                }
+                // Add the AddCDCFile actions that exist to the commit
+                actions.extend(writer.close().await?.into_iter().map(|add| {
+                    Action::Cdc(AddCDCFile {
+                        // This is a gnarly hack, but the action needs the nested path, not the
```
ion-elgreco, line 468: Can't we pass the write_action_type to the writer_config and then use some generic return types, so that based on the writer_config we either return Action::Cdc or Action::Add?

crates/core/src/operations/update.rs
```diff
@@ -231,15 +247,23 @@ async fn execute(
         .await?;
     let scan = Arc::new(scan);
 
-    // Create a projection for a new column with the predicate evaluated
-    let input_schema = snapshot.input_schema()?;
+    // Wrap the scan with a CDCObserver if CDC has been abled so that the tracker can
+    // later be used to produce the CDC files
+    let scan: Arc<dyn ExecutionPlan> = match snapshot.table_config().enable_change_data_feed() {
```
ion-elgreco, line 252: If the table is at v7 writer, then we should only look at the writer_feature whether it exists or not.

In spark it's possible to disable change data feed through the config, even though the writer feature exists.

rtyler: I'm confused by this comment. Are you stating that our enable_change_data_feed() function is incorrect?  I understand what portion of the protocol you're referring to here, but I'm not clear what change is being suggested

rtyler:
> I think it's better to disable it until all operations (delete and merge) are supported, otherwise we cannot push any python releases until those are added

How would you disable this? It doesn't make sense to me to include short-term configuration or feature flags to me.  The protocol states that when the enable change data feed table-feature is enabled, that writers can _optionally_ produce CDC files. Our writers just optionally will only create them on updates for now :laughing:

crates/core/src/operations/update.rs
rtyler: `writer_config` is already a mess, I do not want to add more of anything there. Since I started this pull request we added another field there :upside_down_face: 

I wouldn't call this hack a hack though! That makes it sounds so ugly :laughing:   The concern I have with making the DeltaWriter generic with regards to the Action generated is that the Add has some more expansive requirements  with file stats. 

One option I considered was adding an `From<Add>` on `AddCDCFile` to at least relegate this code into a better location.  I'm not happy with our writer structure at the moment anyways, and it's due for some surgery but I don't want to do that surgery here :frowning: `

crates/core/src/operations/update.rs
ion-elgreco: That function only grabs it from the config, which is valid for writer versions 1..6, but for writer version >=7 we need to look at the writer's features in the protocol if ChangeDateFeed is enabled there.

crates/core/src/operations/transaction/protocol.rs
```diff
@@ -228,6 +228,11 @@ pub static INSTANCE: Lazy<ProtocolChecker> = Lazy::new(|| {
     let mut writer_features = HashSet::new();
     writer_features.insert(WriterFeatures::AppendOnly);
     writer_features.insert(WriterFeatures::TimestampWithoutTimezone);
+    #[cfg(feature = "cdf")]
+    {
+        writer_features.insert(WriterFeatures::ChangeDataFeed);
+        writer_features.insert(WriterFeatures::GeneratedColumns);
```
ion-elgreco, line 234: Is this needed? I don't remember generated columns being required

crates/core/src/operations/cdc.rs
```diff
@@ -0,0 +1,454 @@
...
+/// >
+/// > For Writer Version 7, all writers must respect the delta.enableChangeDataFeed configuration flag in
+/// > the metadata of the table only if the feature changeDataFeed exists in the table protocol's
+/// > writerFeatures.
+pub(crate) fn should_write_cdc(snapshot: &DeltaTableState) -> DeltaResult<bool> {
+    if let Some(features) = &snapshot.protocol().writer_features {
+        // Features should only exist at writer version 7 but to avoid cases where
+        // the Option<HashSet<T>> can get filled with an empty set, checking for the value
+        // explicitly
+        if snapshot.protocol().min_writer_version == 7
```
ion-elgreco, line 287: There should be an else_if that returns true if it's V7 and ChangeDateFeed.

It now just goes into the configuration grab at the end, but it's possible that the WriterFeature is enabled and someone tried to disable it through the config, which doesn't have an effect

rtyler: I believe that the `table_config().enable_change_data_feed()` function's value below would be correct for that case wouldn't it?

ion-elgreco: Not entirely, it's possible to have delta.enableChangeDataFeed = False in the config, even though the WriterFeature ChangeDateFeed is set in the protocol features of the table.

That's why we need to do:

If writer >= 7 {
   If writer_features.contains(cdf) {
        True
    } Else {
  False 
}}

Because >=7 we should only look at features

crates/core/src/operations/transaction/protocol.rs
ion-elgreco: But we are not adding the support here, right?

Protocol states this: "The value of delta.generationExpression SHOULD be parsed as a SQL expression.
Writers MUST enforce that any data writing to the table satisfy the condition (<value> <=> <generation expression>) IS TRUE. <=> is the NULL-safe equal operator which performs an equality comparison like the = operator but returns TRUE rather than NULL if both operands are NULL"

We might only support v7 with CDF at this stage

crates/core/src/operations/cdc.rs
ion-elgreco: Hmmmmm

Protocol states it does still need to look at the config: 

"For Writer Version 7, all writers must respect the delta.enableChangeDataFeed configuration flag in the metadata of the table only if the feature changeDataFeed exists in the table protocol's writerFeatures." 

I guess I misunderstood what spark was doing then :) forget everything I said ^^

rtyler:
Doing further acceptance testing I have identified what I believe to be a [bug in DataFusion](https://github.com/apache/datafusion/issues/10749) and will put this into Draft until I can figure out the path forward

rtyler:
In discussion with @ion-elgreco , due to apache/datafusion#10749 which is really an issue with arrow-rs. We decided that we can move forward without struct/list CDC working with the following conditions:

* warn if the schema has a struct or a list in it when attempting to create CDC batches, to try to make it clear to the user that this is not yet supported
* generated column support doesn't exist yet, but as long as `delta.generatedExpression` is not set, it's safe to allow that table feature to exist (code must be modified).

crates/core/src/operations/transaction/protocol.rs
```diff
@@ -148,17 +149,33 @@ impl ProtocolChecker {
...
-            match snapshot.protocol().min_writer_version {
-                0 | 1 => None,
-                2 => Some(&WRITER_V2),
-                3 => Some(&WRITER_V3),
-                4 => Some(&WRITER_V4),
-                5 => Some(&WRITER_V5),
-                6 => Some(&WRITER_V6),
-                _ => snapshot.protocol().writer_features.as_ref(),
-            };
+        if (4..7).contains(&min_writer_version) {
```
ion-elgreco, line 164: Shouldn't we only allow writing to version 4 and 7 with CDC.

Version 5 is column mapping which we don't support and 6 identity columns.

ion-elgreco (commented):
Just one comment, on the range of support, then we are good to go🤗

ion-elgreco (approved):
🎉🎉

PR #2495, opened 2024-05-09
feat: adopt kernel schema types

### LINKED ISSUE #2489: Adopt the delta kernel types
What feels like long ago, @roeap imported some kernel types manually into the `kernel` module from the early work on delta-kernel-rs. Now `delta_kernel` has been released with a 0.0.1 we can move to adopting those symbols and begin incrementally adopting functionality from the kernel.

This issue is _just_ about the types. I'm not expecting to pull any of the log replay or snapshot code in

roeap:
# Description

First pass adopting `delta_kernel` in `delta-rs`.

depends on https://github.com/delta-incubator/delta-kernel-rs/pull/189 being merged and released.

This PR focusses on the schema types. Adopting the action types would be a follow up, but might have a even greater blast radius then this one.

# Related Issue(s)

part of #2489

# Documentation


crates/core/src/kernel/arrow/mod.rs
```diff
@@ -21,258 +16,6 @@ const MAP_KEY_DEFAULT: &str = "key";
...
-            ArrowDataType::Float32 => Ok(DataType::Primitive(PrimitiveType::Float)),
-            ArrowDataType::Float64 => Ok(DataType::Primitive(PrimitiveType::Double)),
-            ArrowDataType::Boolean => Ok(DataType::Primitive(PrimitiveType::Boolean)),
-            ArrowDataType::Binary => Ok(DataType::Primitive(PrimitiveType::Binary)),
-            ArrowDataType::FixedSizeBinary(_) => Ok(DataType::Primitive(PrimitiveType::Binary)),
-            ArrowDataType::LargeBinary => Ok(DataType::Primitive(PrimitiveType::Binary)),
-            ArrowDataType::Decimal128(p, s) => {
-                Ok(DataType::Primitive(PrimitiveType::Decimal(*p, *s)))
-            }
-            ArrowDataType::Decimal256(p, s) => DataType::decimal(*p, *s).map_err(|_| {
```
ion-elgreco, line 218: This arrow type is missing in kernel, I think that should be upstreamed there as well.

roeap: hmm, given that precision/scale are limited to 38, does that make sense? not entirely sure anymore but I think the 256 bit type only makes sense for larger p/s values.

ion-elgreco: It's mostly for convenience. Users might have their source data using decimal256, but still precision/scale below 38.

python/tests/test_writer.py
```diff
@@ -1458,7 +1458,7 @@ def test_invalid_decimals(tmp_path: pathlib.Path, engine):
 
     with pytest.raises(
         SchemaMismatchError,
-        match=re.escape("Invalid data type for Delta Lake: decimal(39,1)"),
+        match=re.escape("Invalid data type for Delta Lake: Decimal256(39, 1)"),
```
roeap, line 1461: this change is a bit more significant then meets the eye - i.e. we are no longer trying to convert 256 bit decimals to a compliant type in delta. main reason being, for precisions that fit into 128 bit decimals, the user should really be using these, if the larger type is required, we cannot store it in the table anyhow.

crates/core/src/kernel/scalars.rs
```diff
@@ -0,0 +1,246 @@
...
+                            })
+                            .collect::<Option<Vec<_>>>()
+                    })?;
+                if struct_fields.len() != values.len() {
+                    return None;
+                }
+                Some(Self::Struct(
+                    StructData::try_new(struct_fields, values).ok()?,
+                ))
+            }
```
roeap, line 211: @scovich - here is an example of creating a struct scalar in the wild.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -315,8 +315,8 @@ impl Snapshot {
         let stats_fields = if let Some(stats_cols) = self.table_config().stats_columns() {
             stats_cols
                 .iter()
-                .map(|col| match schema.field_with_name(col) {
-                    Ok(field) => match field.data_type() {
+                .map(|col| match schema.field(col) {
```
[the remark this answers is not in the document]
roeap: good question, let me investigate.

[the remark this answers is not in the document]
roeap: @alexwilcoxson-rel - stated some asking around if it would be legal to have a nested field defined in that property.

There are a couple of questions that also come to mind, and maybe you have an opinion? So using `.` as a separator is - AFAIK - just a convention (certainly a common one though). In delta one could for instance define a column name that does contain a `.` character under the `columnMapping` feature. In kernel we have some places where we do represent nested coluns as dot separated, but we do already know that we eventually need an array representation.

One reason one might want to specify a nested field, is if we cannot compute stats for some child fields. There we could also just be lenient and simply omit these (actually not sure what we do now :D). Is that your case as well, or do you have others?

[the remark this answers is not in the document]
roeap: btw - the fact that `.fields()` does not any parsing / traversing is by design. i.e. in this case we would need to split here...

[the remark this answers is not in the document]
roeap: So turns out nested fields are perfectly valid, but they do require some more parsing, as it is also legal to escape field names that contain special characters.

https://github.com/delta-io/delta/blob/4b102d34a2ce881b2a851b4c6cfbf2ab3ab5534f/spark/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L549-L561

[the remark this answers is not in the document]
roeap: i opened #2572 to track this.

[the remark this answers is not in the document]
alexwilcoxson-rel: in our case if nested fields our omitted thats fine we only care about the stats on the root fields for query performance

we only included the nested fields cause they are not nullable fields of nullable parent fields and if we didn't collect them chckpoint parsing would break

also at the time, delta-rs did not respect the stats configuration so we just used the parquet writer properties to only collect stats on what we needed

Cargo.toml
```diff
@@ -26,6 +26,9 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
+delta_kernel = { version = "0.1" }
+# delta_kernel = { path = "../delta-kernel-rs/kernel" }
```
rtyler, line 30:
```suggestion
delta_kernel = { version = "0.1", path = "../delta-kernel-rs/kernel" }
```

These can combined for ease of development, since [cargo allows multiple locations](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations)

crates/core/Cargo.toml
```diff
@@ -15,6 +15,8 @@ rust-version.workspace = true
 features = ["datafusion", "json", "unity-experimental"]
 
 [dependencies]
+delta_kernel.workspace = true
+
```
rtyler, line 19:
```suggestion
delta-kernel = { workspace  = true }
```

rtyler (approved):
Suggested some changes, but I would rather see this merged than wait for my nitpicks :laughing: 

(also don't want to block up more testing before Data and AI Summit)

PR #2395, opened 2024-04-07
chore(rust): bump arrow v51 and datafusion v37

lasantosr:
# Description
Update the arrow and datafusion dependencies.

# Related Issue(s)
- closes #2328

# Documentation

Issue #2188, opened 2024-02-16
internal.DeltaError: Generic DeltaTable error: Internal error: Invalid HashJoinExec partition count mismatch 1!=2


t1g0rz:
# Environment

**Delta-rs version**:
rust-v0.17.0

**Binding**: 
python
deltalake-0.15.3

**Environment**:
- **Cloud provider**: AWS
- **OS**:  22.04.1-Ubuntu
- **Other**:  
  - RAM: 949Mi
  - 1 CPU 
  - t2.micro
  - ami-03f4878755434977f

***
# Bug

**What happened**:

When I run a simple script a few times (twice, in my case) to simulate updating and partitioning, I encounter the following error.

```
Traceback (most recent call last):
  File "/home/ubuntu/test.py", line 52, in <module>
    er = dt.merge(source=tmp_df, predicate="target.idx = source.idx", source_alias='source', target_alias='target').when_matched_update_all().when_not_matched_insert_all().execute()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/deltalake/table.py", line 1713, in execute
    metrics = self.table._table.merge_execute(
_internal.DeltaError: Generic DeltaTable error: Internal error: Invalid HashJoinExec, partition count mismatch 1!=2,consider using RepartitionExec.
This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker
```

**What you expected to happen**:
Table updated w/o any issues

**How to reproduce it**:
Here is the script. You can run it on t2.micto AWS

```python
from deltalake import DeltaTable, write_deltalake
from sklearn.datasets import load_iris
import pyarrow as pa
import pandas as pd
from time import time, sleep
import boto3


session = boto3.session.Session()
credentials = session.get_credentials()

S3_BUCKET_RAW_DATA = <S3_BUCKET>
DELTA_TABLE_NAME = 'iris'

storage_options={
    'AWS_S3_LOCKING_PROVIDER': 'dynamodb',
    'DELTA_DYNAMO_TABLE_NAME': 'delta_log',
    'AWS_REGION': 'us-east-1',
    'DELTA_DYNAMO_REGION': 'us-east-1',
    'AWS_ACCESS_KEY_ID': credentials.access_key,
    'AWS_SECRET_ACCESS_KEY': credentials.secret_key,
}

if credentials.token:
    storage_options['AWS_SESSION_TOKEN'] = credentials.token

df = load_iris(as_frame=True)['data']
[...]

PR #2499, opened 2024-05-11
chore: upgrade to Datafusion 38

rtyler:


ion-elgreco:
With v38, I think we can revert this temp fix: https://github.com/delta-io/delta-rs/pull/2395/files#diff-60c94c425cf655602ada7636a35fb8e25391bc92f5b9e35462562604449be324R1055

echai58:
datafusion 38 should also close this issue: https://github.com/delta-io/delta-rs/issues/2188 with https://github.com/apache/datafusion/pull/10095

abhiaagarwal:
It could be nice to skip directly to DataFusion 39, which just dropped an RC — it depends on Arrow 52, which released a few days ago, and pyo3 0.21.1, meaning all three can be bumped together :) working on that PR rn

Issue #2411, opened 2024-04-12
Use predicates to restrict scan in merge operation


Tommel71:
# Description

Use predicates in the merge operation to read only parts of the affected partition into memory.

**Use Case**

I only ever write to one partition at a time.
I have a large table and I want to merge 2000 rows (or only one row - the issue remains) using `when_not_matched_insert_all` into it. The merge operation respects partitions, so merging into a new partition is fast. However, it looks like the merge operation currently reads the entire partition into memory, although in my case, the predicate and metadata could be used to restrict the search to only one file in the partition or even just one or very few row groups in parquet when my data is z-ordered. This would greatly speed up the merge operation in my case.

Currently, my query to merge 2000 rows into a partition of 2GB uncompressed parquet files takes 30 seconds, which forces me to internally keep track of whether or not the data has been written which in turn exposes me to data inconsistencies.

**Related Issue(s)**

ion-elgreco:
@Tommel71 with 0.19.1 is it improved? Because it should potentially skip more files now

We still read the full file though if we think it's relevant

Tommel71:
@ion-elgreco Thank you for coming back to this. Good to know that we can use this in the future. For now I have no good way to replicate what I had before because we now implemented the functionality by appending data instead of merging.

I am pretty sure this would now also be a viable approach.

PR #2513, opened 2024-05-14
feat: improve merge performance by using predicate non-partition columns min/max for prefiltering

JonasDev1:
# Description
This pr improves the merging performance by adding min/max filters to the early filter.
The number of files scanned from the target file table is reduced by using the table statistics. 
I have extended the early filter for this purpose. This filter is responsible for pre-filtering the target table. 
Previously, the early filter only consisted of partition columns by filtering for all unique values from the source. Now the non-partition columns are also used by aggregating the min/max values from the source and adding a between expression to the early filter. 

It is also automatically part of the conflict detection based on the predicate.

I added a property `extended_early_filter` to make this advanced filtering optional. I don't know if this is important, and maybe we can replace the bool with an enum. What do you think about this?  

## Example:
Merge into table t with partition date

Predicate: source.date = target.date and source.timestamp = target.timestamp and source.id = target.id and frob > 42

Early filter before: `date = '2024-‚05-14' and frob > 42`
Early filter now: `date = '2024-05-14' and timestamp BETWEEN '…15:00' AND '…15:05' and id BETWEEN 'A' AND 'B' and frob > 42`

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -49,11 +49,14 @@ use datafusion::{
 use datafusion_common::tree_node::{Transformed, TreeNode};
 use datafusion_common::{Column, DFSchema, ScalarValue, TableReference};
 use datafusion_expr::expr::Placeholder;
-use datafusion_expr::{col, conditional_expressions::CaseBuilder, lit, when, Expr, JoinType};
 use datafusion_expr::{
-    BinaryExpr, Distinct, Extension, LogicalPlan, LogicalPlanBuilder, Operator, Projection,
+    col, conditional_expressions::CaseBuilder, lit, max, min, when, Between, Expr, JoinType,
+};
+use datafusion_expr::{
+    Aggregate, BinaryExpr, Extension, LogicalPlan, LogicalPlanBuilder, Operator,
     UserDefinedLogicalNode, UNNAMED_TABLE,
 };
+use either::{Left, Right};
 use futures::future::BoxFuture;
 use itertools::Itertools;
 use parquet::file::properties::WriterProperties;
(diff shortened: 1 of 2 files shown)


JonasDev1:
#2411

ion-elgreco:
@JonasDev1 why did you make the advanced filtering optional?

If this provides better performance across the board, we should enable it always (so also for python bindings)

JonasDev1:
My concern was if you want to do e.g. merges via columns with null this would not work, but I think that it would not work without the advanced filtering either as `is equal` for null is not defined in sql. 


Spark has an extra null safe operator `<=>` for this, which is not in Datafusion available yet.

JonasDev1:
What about the review? 

I can of course also remove the flag again

ion-elgreco:
> My concern was if you want to do e.g. merges via columns with null this would not work, but I think that it would not work without the advanced filtering either as `is equal` for null is not defined in sql.
> 
> Spark has an extra null safe operator `<=>` for this, which is not in Datafusion available yet.

My main issue is that it might work or not work based on the contents of the data. I think that's a bit tricky because a person needs to be aware of the contents of their data they are trying to write. 

`<=> for this, which is not in Datafusion available yet.` @JonasDev1 can you maybe raise an issue about this upstream in datafusion, so we get this first

thomasfrederikhoeck:
@JonasDev1 do you know if the issue was raised in DataFusion? I would love to see this feature to really enhance large table performance in the cases where partition is not meaningful due to high cardinality so it is very nice that you have initiated it.

JonasDev1:
@ion-elgreco I have tested the merge with null values in merge predicate with extend filtering and without. In both cases the merge doesn't work as expected and will lead to duplicate rows. Therefore this pull request will create no changes and the behaviour is simillar to Spark. 

The null-safe operation would be a nice extension, but is not bounded to this pull request

ion-elgreco:
@JonasDev1 can you resolve the merge conflicts? Then I'll do another round of reviewing

JonasDev1:
@ion-elgreco Done

crates/core/src/operations/merge/mod.rs
```diff
@@ -666,35 +669,40 @@ impl ExtensionPlanner for MergeMetricExtensionPlanner {
...
 ///
-/// 1. for any relations between a source column and a target column, if the target column is a
-/// partition column, then replace source with a placeholder matching the name of the partition
+/// 1. for any relations between a source column and a partition target column,
+/// replace source with a placeholder matching the name of the partition
 /// columns
 ///
-/// 2. for any other relation with a source column, remove them.
+/// 2. for any is equal relations between a source column and a non-partition target column,
+/// replace source with is between expression with min(source_column) and min(source_column) placeholders
```
ion-elgreco, line 685:
```suggestion
/// replace source with is between expression with min(source_column) and max(source_column) placeholders
```

ion-elgreco (approved):
Thankss, and sorry for the delay! @JonasDev1

ion-elgreco:
@rtyler can't a branch be merged if it can't be rebased anymore?

@JonasDev1 are you perhaps able to manually rebase and force push it?

PR #2525, opened 2024-05-18
feat(rust): make PartitionWriter public

adriangb:
I would like to use PartitionWriter within my own crate. Would you mind making this public so I don't have to vendor it and all of the associated code?

### DIFF
--- a/crates/core/src/operations/writer.rs
+++ b/crates/core/src/operations/writer.rs
@@ -226,7 +226,6 @@ impl DeltaWriter {
     }
 }
 
-
 /// Write configuration for partition writers
 #[derive(Debug)]
 pub struct PartitionWriterConfig {
@@ -274,7 +273,6 @@ impl PartitionWriterConfig {
         })
     }
 }
-
 
 /// Partition writer implementation
 /// This writer takes in table data as RecordBatches and writes it out to partitioned parquet files.
--- a/crates/core/src/operations/writer.rs
+++ b/crates/core/src/operations/writer.rs
@@ -226,8 +226,10 @@ impl DeltaWriter {
     }
 }
 
+
+/// Write configuration for partition writers
 #[derive(Debug)]
-pub(crate) struct PartitionWriterConfig {
+pub struct PartitionWriterConfig {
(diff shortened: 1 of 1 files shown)


rtyler:
@adriangb I'm not opposed to the idea, can you share more about how this can be used outside of the delta context, I'm curious

adriangb:
I have a Rust service that does buffering of data before writing to Delta. I control all of the writes going to Delta so I want to avoid all of the overhead of checking that the schemas match, etc which happens if you go through the public APIs. I also want to take control of the write retry loop because I may have to switch to pessimistic concurrency control under high write load (take out a lock on a low latency system like Redis only at the stage of writing the commit) to avoid write contention.

rtyler:
Sounds reasonable to me! I think this API is stable enough to make public in the next release, please run a `cargo fmt` and then this should merge

rtyler (approved):

adriangb:
done @rtyler !

ion-elgreco (approved):

roeap:
@rtyler @adriangb - recently I have been thinking if we should adopt the visibility crate to expose a larger api behind a feature gate and less guarantees. We uses this - so gar successfully I believe- in kernel ...

There also is some writing stuff - e.g. muti-part put from object_store on my todo list, that may affect this.

adriangb:
@roeap not sure I understood fully, are you saying that this right now should not be made public or that there is an opportunity for a more thoughtful approach to a public API within the rust crates in the future?

ion-elgreco:
@roeap best to wait for object store 0.10 before working on the multi part

@aersam also already put some work there

roeap:
@adriangb - i think we can make this public now, but maybe behind a feature gate. The [visibility](https://docs.rs/visibility/latest/visibility/) crate enables this.

@ion-elgreco - yes, still some things to cross off first, and we'll work from where we are, when i get to it 🙂.

adriangb:
> i think we can make this public now, but maybe behind a feature gate. The [visibility](https://docs.rs/visibility/latest/visibility/) crate enables this.

Interesting, TIL!

PR #2533, opened 2024-05-20
chore: expose `files_by_partition` to public api

edmondop:
# Description
This [documentation](https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.schema) describes the usage of `files_by_partition` but that API doesn't seem to be surfaced from the underlying rust implementation

- [ ] Add unit tests once confirmed this makes sense

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -715,7 +715,7 @@ class DeltaTable:
         """
         Get the files for each partition
 
-        ""
+        """
         return self._table.files_by_partitions(partition_filters)
 
     def metadata(self) -> Metadata:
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@edmondop can you fix the linting issues? Then we can merge it

ion-elgreco (dismissed):

edmondop:
I fixed the linting, but I think I should add a test since I have grown the API surface before this get merged. 

I submitted to verify if I was wrong in reading the documentation or if a method was missing from the public api

edmondop:
@ion-elgreco looks also there is a flaky concurrency test

ion-elgreco (approved):

ion-elgreco:
@edmondop let me know when you added the test

Issue #2376, opened 2024-04-02
Support for Arrow PyCapsule interface


kylebarron:
# Description

Looking at the code below, the `write_deltalake` function is hard-coded to support only specific pyarrow and pandas objects. 
https://github.com/delta-io/delta-rs/blob/6f81b8034dbef7a0120e354e178f0c859564465e/python/deltalake/writer.py#L283-L306

With the [Arrow PyCapsule interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html), you can support any Arrow-based Python input, regardless of the library implementation.

**Use Case**

I'm working on a [Python library for Arrow-backed geospatial operations](https://github.com/geoarrow/geoarrow-rs), and support for the PyCapsule interface would allow interfacing with my or any other Python Arrow library without needing hard-coded support.

**Related Issue(s)**

ion-elgreco:
@kylebarron hey Kyle, are you willing to make this contribution here?

kylebarron:
I put up a short initial PR for the write side to get things moving. I think we may be able to wrap some additional classes to export through the pycapsule interface as well (deferring to `to_pyarrow` under the hood)

PR #2534, opened 2024-05-21
feat(python): handle PyCapsule interface objects in write_deltalake

kylebarron:
# Description

Adds support for the [Arrow PyCapsule interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html). 

Since pyarrow is already a required dependency, this takes the minimal route of converting pycapsule interface objects into pyarrow objects. This requires pyarrow 15 or higher for the stream conversion (https://github.com/apache/arrow/issues/39217).

This doesn't modify the existing hard-coded support for pyarrow and pandas

# Related Issue(s)

- closes https://github.com/delta-io/delta-rs/issues/2376

# Documentation

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -77,7 +77,9 @@ class ArrowStreamExportable(Protocol):
     https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
     """
 
-    def __arrow_c_stream__(self, requested_schema: Optional[object] = None) -> object: ...
+    def __arrow_c_stream__(
+        self, requested_schema: Optional[object] = None
+    ) -> object: ...
 
 
 @dataclass
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -77,7 +77,7 @@ class ArrowStreamExportable(Protocol):
     https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
     """
 
-    def __arrow_c_stream__(self, requested_schema: object | None = None) -> object: ...
+    def __arrow_c_stream__(self, requested_schema: Optional[object] = None) -> object: ...
 
 
 @dataclass
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -15,6 +15,7 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Protocol,
     Tuple,
     Union,
     overload,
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@kylebarron can you fix the linting issues?

Also wondering, how we should typehint this now, since an input can have the c_stream attribute or not

kylebarron:
> @kylebarron can you fix the linting issues? Then we can merge it

I'm pretty packed but I can try to find some time soon.

> Also wondering, how we should typehint this now, since an input can have the c_stream attribute or not

You can use these type hints: https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#protocol-typehints

ion-elgreco:
@kylebarron ah nice, do you mind adding those typehints when you find the time

kylebarron:
I believe I fixed the lint and fixed the type hinting.

In the future, a more involved PR could remove pyarrow as a required dependency entirely by passing the C stream pycapsule directly to Rust (arrow-rs has an example of how to do that [here](https://github.com/apache/arrow-rs/blob/9be0eb530de85a68358dc31c146f71ccb97739ca/arrow/src/pyarrow.rs#L416-L425))

python/deltalake/writer.py
```diff
@@ -70,6 +71,15 @@
 DEFAULT_DATA_SKIPPING_NUM_INDEX_COLS = 32
 
 
+class ArrowStreamExportable(Protocol):
+    """Type hint for object exporting Arrow C Stream via Arrow PyCapsule Interface.
+
+    https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
+    """
+
+    def __arrow_c_stream__(self, requested_schema: object | None = None) -> object: ...
```
ion-elgreco, line 80:
```suggestion
    def __arrow_c_stream__(self, requested_schema: Optional[object] = None) -> object: ...
```

python/deltalake/writer.py
```diff
@@ -70,6 +71,15 @@
 DEFAULT_DATA_SKIPPING_NUM_INDEX_COLS = 32
 
 
+class ArrowStreamExportable(Protocol):
+    """Type hint for object exporting Arrow C Stream via Arrow PyCapsule Interface.
+
+    https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html
+    """
+
+    def __arrow_c_stream__(self, requested_schema: Optional[object] = None) -> object: ...
```
ion-elgreco, line 80:
```suggestion
    def __arrow_c_stream__(
        self, requested_schema: Optional[object] = None
    ) -> object: ...
```

ion-elgreco:
@kylebarron that would be nice, we could potentially make it opt-in since it would then only be needed for reading

ion-elgreco (approved):
Thanks for adding this!

PR #2535, opened 2024-05-22
docs: pull delta from conda not pip

avriiil:
This PR updates the conda install .yml file to pull deltalake from conda-forge instead of pip. Mixing pip and conda is not recommended when it can be avoided.


avriiil:
@ion-elgreco - 2 of the tests were failing. anything I can fix at my end to solve for this?

ion-elgreco (approved):

PR #2539, opened 2024-05-25
feat: implement transaction identifiers - continued

### LINKED ISSUE #2130: Transaction Identifiers

roeap:
# Description

This is based on @Blajda's work in #2327 and aims to revive transactions identifiers.

This PR elaborates a bit on the `ReplayVisitor` concept introduced by David. specifically we moved things "one level down" to be tracked on the eager snapshot. This was mainly required to make it work with the commit flow, to properly handle updating the state after commits, without piping the visitors through all the ways.

The nice thing about this, that we can isolate the mechanics as to when or how we track additional actions in the `EagerSnapshot` and expose an interface the looks like what we might get for kernel - i.e. some opaque iterator over the respective actions.

# Related Issue(s)

closes #2130

# Documentation


crates/core/src/operations/transaction/mod.rs
```diff
@@ -273,31 +280,28 @@ impl CommitData {
...
         }
-        Ok(jsons.join("\n"))
-    }
-
-    /// Obtain the byte representation of the commit.
-    pub fn get_bytes(&self) -> Result<bytes::Bytes, TransactionError> {
-        // Data MUST be read from the passed `CommitData`. Don't add data that is not sourced from there.
-        let actions = &self.actions;
-        Ok(bytes::Bytes::from(Self::log_entry_from_actions(actions)?))
+        Ok(bytes::Bytes::from(jsons.join("\n")))
```
roeap, line 304: since i to grok the commit logic a bit, it seemed like this was just seprated for legacy reasons, and the `log_entry_from_actions` was ever only supposed to be used with the data in `CommitData`

crates/core/src/operations/transaction/mod.rs
```diff
@@ -407,15 +427,20 @@ impl<'a> CommitBuilder {
...
+            self.app_transaction,
+        );
+        PreCommit {
             log_store,
             table_data,
             max_retries: self.max_retries,
             data,
             post_commit_hook: self.post_commit_hook,
-        })
+        }
```
roeap, line 443: this did not need to be fallible - a lot of the touched files just updates from this change.

roeap (commented):
After this PR, there is very little left that the `DeltaTableState` is actually doing. Moist methods are just proxies to the methods on `EagerSnapshot`, which makes sense, since that also represents the state 😆.

After this, and getting the first kernel version in (#2495) I hope that we can finally drop the `DeltaTableState` ...

rtyler (approved):

crates/core/src/kernel/snapshot/visitors.rs
```diff
@@ -0,0 +1,192 @@
...
+            if id.is_valid(idx) {
+                let app_id = ex::read_str(id, idx)?;
+                if self.app_transaction_version.contains_key(app_id) {
+                    continue;
+                }
+                self.app_transaction_version.insert(
+                    app_id.to_owned(),
+                    Transaction {
+                        app_id: app_id.into(),
+                        version: ex::read_primitive(version, idx)?,
```
vegarsti, line 85: Sorry for a drive-by comment, but: Does this mean this will only keeps the first version? since we continue if the hash map has the key

vegarsti: Sorry, `test_app_txn_visitor` answered this in the affirmative.

Is that the correct behavior, though? The Delta protocol [says](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers)

> Delta only ensures that the **latest** version for a given appId is available in the table snapshot.

crates/core/src/table/state.rs
```diff
@@ -155,10 +161,9 @@ impl DeltaTableState {
             .map(|add| add.object_store_path())
     }
 
-    /// HashMap containing the last txn version stored for every app id writing txn
-    /// actions.
-    pub fn app_transaction_version(&self) -> &HashMap<String, i64> {
-        &self.app_transaction_version
+    /// HashMap containing the last transaction stored for every application.
```
vegarsti, line 164: Here it says _last_, but given the above I wonder if that's not the case?

crates/core/src/kernel/snapshot/visitors.rs
roeap: The replay visits commits from latest to oldest. By keeping the first one we encounter, we are in fact only showing the latest.

vegarsti: Oh I see! Thanks!

PR #2541, opened 2024-05-27
fix: cast support fields nested in lists and maps

HawaiianSpork:
# Description

The current implementation of cast only works for structs nested in structs.  This PR adds supports for structs contained in other types (lists and maps).  This PR also prevents cast from adding nullable column if the field is not nullable, instead it will throw an error.

Note: This is only a partial solution which would let you merge schema with nested missing columns, it does not allow delta-rs to read the merged schema (though Spark can).  To read the merged schema will require another change where delta-rs defines its own datafusion parquet schemaAdapter.

### DIFF
--- a/crates/core/src/operations/cast.rs
+++ b/crates/core/src/operations/cast.rs
@@ -3,9 +3,13 @@
 use crate::kernel::{
     ArrayType, DataType as DeltaDataType, MapType, MetadataValue, StructField, StructType,
 };
-use arrow_array::{new_null_array, Array, ArrayRef, RecordBatch, StructArray};
+use arrow_array::cast::AsArray;
+use arrow_array::{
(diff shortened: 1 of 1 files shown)


rtyler:
> Please add also a bunch of python tests

@ion-elgreco I'm confused as to what a Python set of tests would do that the tests in Rust don't already do?  :confused:

ion-elgreco:
> > Please add also a bunch of python tests
> 
> @ion-elgreco I'm confused as to what a Python set of tests would do that the tests in Rust don't already do?  :confused: 

We mostly need to check the reader side

ion-elgreco (approved):
Looks good! Thanks

ion-elgreco:
@HawaiianSpork thanks for the PR! can you rebase please so we can merge

HawaiianSpork:
@ion-elgreco thank you.  Sorry, I did not get around to writing python tests.  The code has been rebased.

ion-elgreco (approved):

PR #2548, opened 2024-05-29
fix: clippy warnings

imor:
# Description
This PR fixes many clippy warnings in the Rust code.

# Related Issue(s)

# Documentation


crates/core/src/table/state_arrow.rs
```diff
@@ -572,7 +572,7 @@ impl DeltaTableState {
                 // into StructArrays, until it is consolidated into a single array.
                 columnar_stats = columnar_stats
                     .into_iter()
-                    .group_by(|col_stat| {
```
imor, line 575: `group_by` is deprecated. The library suggests to use `chunk_by` instead.

rtyler (approved):

PR #2560, opened 2024-06-01
docs: dt.delete add context + api docs link

avriiil:
This PR adds more general context about using `dt.delete()` and a link to the API docs


ion-elgreco (approved):
😄

PR #2561, opened 2024-06-01
docs: add drop columns usage page

avriiil:
This PR adds a `usage` page about dropping columns


avriiil:
@MrPowers -- is this the recommended way to drop columns with `python deltalake` atm?

Using the `alter table` functionality as recommended [here](https://delta.io/blog/2022-08-29-delta-lake-drop-column) would be better, but AFAICT there is no column drop functionality in `dt.alter` yet.

ion-elgreco:
@avriiil let's maybe hold off with this until we have a `drop column` operation, I'll probably look into this after `add column` operation is merged

avriiil:
@ion-elgreco - makes sense! Feel free to ping me when the `drop column` API is ready and I'll be happy to update this docs page

PR #2562, opened 2024-06-02
feat(python, rust): `add column` operation

ion-elgreco:
# Description
The schema evolution code is quite well abstracted from the writer, so it seemed straight forward to expose this through an `add column` api. This would make it easier for users to add new columns or fields in structs.

At some point we can add type widening to the schema evolution as a separate path, which also than could be used to create an `alter column` operation.

### DIFF
--- a/crates/core/src/kernel/models/actions.rs
+++ b/crates/core/src/kernel/models/actions.rs
@@ -5,6 +5,8 @@ use std::str::FromStr;
 // use std::sync::Arc;
 
 // use roaring::RoaringTreemap;
+use crate::DeltaConfigKey;
+use maplit::hashset;
 use serde::{Deserialize, Serialize};
 use tracing::warn;
 use url::Url;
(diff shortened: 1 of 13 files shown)


crates/core/src/kernel/models/actions.rs
```diff
@@ -137,10 +137,10 @@ pub struct Protocol {
 
 impl Protocol {
     /// Create a new protocol action
-    pub fn new(min_reader_version: i32, min_wrriter_version: i32) -> Self {
```
rtyler, line 140: :laughing:

crates/core/src/operations/add_column.rs
```diff
@@ -0,0 +1,125 @@
...
+    snapshot: DeltaTableState,
+    /// Fields to add/merge into schema
+    fields: Option<Vec<StructField>>,
+    /// Delta object store for handling data files
+    log_store: LogStoreRef,
+    /// Additional information to add to the commit
+    commit_properties: CommitProperties,
+}
+
+impl super::Operation<()> for AddColumnBuilder {}
```
rtyler, line 29: :+1:

python/deltalake/table.py
```diff
@@ -1828,6 +1829,36 @@ class TableAlterer:
...
+            from deltalake import DeltaTable
+            from deltalake.schema import Field, PrimitiveType, StructType
+            dt = DeltaTable("test_table")
+            new_fields = [
+                Field("baz", StructType([Field("bar", PrimitiveType("integer"))])),
+                Field("bar", PrimitiveType("integer"))
+            ]
+            dt.alter.add_column(
+                new_fields
+            )
```
rtyler, line 1854: Since this is taking a series of fields, I feel that it should be called `add_columns` to be more clear.

python/src/lib.rs
```diff
@@ -528,6 +531,40 @@ impl RawDeltaTable {
...
+                self._table.log_store(),
+                self._table.snapshot().map_err(PythonError::from)?.clone(),
+            );
+
+            let new_fields = fields
+                .iter()
+                .map(|v| v.inner.clone())
+                .collect::<Vec<StructField>>();
+
+            cmd = cmd.with_fields(new_fields);
```
rtyler, line 552: Since `fields` is owned, can this just consume the value? e.g.

```rust
cmd = cmd.with_fields(fields.into_iter().map(|v| v.inner)))
```

(or some such nonsense :shrug: )

crates/core/src/operations/add_column.rs
```diff
@@ -0,0 +1,125 @@
...
+        Self {
+            snapshot,
+            log_store,
+            fields: None,
+            commit_properties: CommitProperties::default(),
+        }
+    }
+
+    /// Specify the fields to be added
+    pub fn with_fields(mut self, fields: Vec<StructField>) -> Self {
```
rtyler, line 43: I recommend making fields `an impl IntoIter` sort of thing for more flexibility by the caller

crates/core/src/operations/add_column.rs
```diff
@@ -0,0 +1,125 @@
...
+                        min_writer_version: 7,
+                        writer_features: Some(hashset! {WriterFeatures::TimestampWithoutTimezone}),
+                        reader_features: Some(hashset! {ReaderFeatures::TimestampWithoutTimezone}),
+                    };
+                    // Convert existing properties to features since we advance the protocol to v3,7
+                    Some(convert_properties_to_features(
+                        new_protocol,
+                        &metadata.configuration,
+                    ))
+                }
```
rtyler, line 97: This code looks really familiar to me, don't we have something similar in update and merge code paths? Part of the middleware thinking I need to return to is to handle this sorts of protocol evolutions.

Would it be possible to at least consolidate this checking code into a shared function in operations/mod.rs?

crates/core/src/operations/add_column.rs
```diff
@@ -0,0 +1,125 @@
...
+            } else {
+                None
+            };
+
+            let operation = DeltaOperation::AddColumn {
+                fields: fields.into_iter().collect_vec(),
+            };
+
+            metadata.schema_string = serde_json::to_string(&new_table_schema)?;
```
rtyler, line 106: Just thinking out loud here, not feedback on this pull request, but implementing TryInto<String> for StructType would probably be handle for us

rtyler (commented):
I think this is overall a good idea to add :clap:

crates/core/src/operations/add_column.rs
ion-elgreco: Yeah, this started to bother me as well, in multiple locations we need to check and evolve the schema. I'll do another round on Thursday and see if I can do some minor refactoring across our API

python/deltalake/table.py
ion-elgreco: Good point!

python/src/lib.rs
ion-elgreco: Hmm I'll have to try, reason I cloned is because StructField didn't have the Copy Trait

nadyr-mg:
Hi, it's really a good idea to add such operation. Is there any update on this?

crates/core/src/operations/add_column.rs
```diff
@@ -0,0 +1,125 @@
...
+            let table_schema = this.snapshot.schema();
+            let new_table_schema =
+                merge_struct(table_schema, &fields.clone().into_iter().collect())?;
+
+            // TODO(ion): Think of a way how we can simply this checking through the API or centralize some checks.
+            let contains_timestampntz = PROTOCOL.contains_timestampntz(&fields);
+            let protocol = this.snapshot.protocol();
+
+            let maybe_new_protocol = if contains_timestampntz {
+                if protocol.min_reader_version == 3 && protocol.min_writer_version == 7 {
```
hntd187, line 78: So if protocol is 3,7 you enable timestamp ntz, otherwise you set the protocol to 3,7 and enable timezone ntz? Am I missing something or does this seem redundant? Wouldn't you just check if it has timestamps and update the protocol/add table features? It seems unnecessary to create a new one.

crates/core/src/operations/add_column.rs
```diff
@@ -0,0 +1,125 @@
...
+            let operation = DeltaOperation::AddColumn {
+                fields: fields.into_iter().collect_vec(),
+            };
+
+            metadata.schema_string = serde_json::to_string(&new_table_schema)?;
+
+            let mut actions = vec![metadata.into()];
+
+            if let Some(new_protocol) = maybe_new_protocol {
+                actions.push(new_protocol.into())
```
hntd187, line 111: From the above new protocol it ever be None in this case? it doesn't seem so.

ion-elgreco:
> Hi, it's really a good idea to add such operation. Is there any update on this?

Unfortunately I am busy for the next 4+ months, after that I'll probably pick the pace up again on delta-rs

crates/core/src/operations/add_column.rs
ion-elgreco: Afaik, if we just add a new value to the HashSet of table features, it would would not automatically bump also reader version 3 and writer version 7, perhaps that should be implemented to make this checks redundant.

So if you add a table feature, just modify itself to bump reader and writer.

crates/core/src/kernel/models/actions.rs
```diff
@@ -137,30 +139,243 @@ pub struct Protocol {
...
-    /// set the reader features in the protocol action
+    /// set the reader features in the protocol action, automatically bumps min_reader_version
     pub fn with_reader_features(
         mut self,
         reader_features: impl IntoIterator<Item = impl Into<ReaderFeatures>>,
     ) -> Self {
-        self.reader_features = Some(reader_features.into_iter().map(|c| c.into()).collect());
+        let all_reader_features = reader_features
+            .into_iter()
+            .map(|c| c.into())
```
hntd187, line 158: for the future you could do `.map(Into::into)` instead

crates/core/src/kernel/models/actions.rs
```diff
@@ -137,30 +139,243 @@ pub struct Protocol {
...
+    /// Converts existing properties into features if the reader_version is >=3 or writer_version >=3
+    /// only converts features that are "true"
+    pub fn move_table_properties_into_features(
+        mut self,
+        configuration: &HashMap<String, Option<String>>,
+    ) -> Protocol {
+        if self.min_writer_version >= 7 {
+            let mut converted_writer_features = configuration
+                .iter()
+                .filter(|(_, value)| {
```
hntd187, line 192: Just understanding here, so this unwraps the value of the key, if it's already none, return false. If it's Some, parse it into a bool which you check is both ok and actually true, but then collect the hashmap back into a string?

Why not at this point just keep it bools because you just drop the values below anyways. Or better yet `collect::<HashMap<_, _>>>()` ?

crates/core/src/kernel/models/actions.rs
```diff
@@ -137,30 +139,243 @@ pub struct Protocol {
...
+                })
+                .collect::<HashMap<&String, &Option<String>>>()
+                .keys()
+                .map(|key| (*key).clone().into())
+                .filter(|v| !matches!(v, WriterFeatures::Other(_)))
+                .collect::<HashSet<WriterFeatures>>();
+
+            if configuration
+                .keys()
+                .any(|v| v.contains("delta.constraints."))
```
hntd187, line 205: `starts_with` instead

crates/core/src/operations/write.rs
```diff
@@ -816,6 +845,34 @@ impl std::future::IntoFuture for WriteBuilder {
                         .or_else(|_| snapshot.arrow_schema())
                         .unwrap_or(schema.clone());
 
+                    let configuration = snapshot.metadata().configuration.clone();
+                    let current_protocol = snapshot.protocol();
+                    let maybe_new_protocol = if PROTOCOL.contains_timestampntz(
+                        TryInto::<StructType>::try_into(schema.clone())?.fields(),
+                    ) && !current_protocol
+                        .reader_features
+                        .clone()
```
hntd187, line 854: Do you need to clone the protocol to just check for a key in the features?

hntd187 (changes requested):
Overall looks okay, I would like the starts_with addressed though

crates/core/src/kernel/models/actions.rs
ion-elgreco:
```suggestion
            .map(Into::into)
```
Cool didn't know you could just pass a fn

crates/core/src/kernel/models/actions.rs
ion-elgreco:
```suggestion
                .any(|v| v.starts_with("delta.constraints."))
```

crates/core/src/operations/write.rs
ion-elgreco: When I didn't clone it would move the features value out of the protocol and then complain that Protocol l doesn't have the copy trait

crates/core/src/kernel/models/actions.rs
ion-elgreco: So I need to first filter the hashmap for enabled configurations, to then try parse each key as a feature

hntd187 (approved):

PR #2565, opened 2024-06-03
fix: add raise_if_key_not_exists to CreateBuilder

### LINKED ISSUE #2564: `raise_if_not_exists` for properties not configurable on CreateBuilder
The `CreateBuilder` calls `apply_properties_to_protocol` with `raise_if_not_exists` set to `true`.
https://github.com/delta-io/delta-rs/blob/f2e30a0bbf58a3b56b2f912a02e10e7af0835301/crates/core/src/operations/create.rs#L282
This was introduced fairly recently, in https://github.com/delta-io/delta-rs/pull/2264.

Could we make `raise_if_not_exists` configurable on the builder? We are using the `with_configuration` to add some opaque configuration, using config keys that delta-rs doesn't know about, so we can't upgrade delta-rs.

(I'm pushing a PR for this now.)

vegarsti:
This PR adds a `raise_if_key_not_exists` flag to the `CreateBuilder` which we propagate (currently always `true`), such that users can choose not to fail if the configuration contains a key that is not a valid DeltaConfigKey. Also expose this in the Python binding.

Closes #2564.

@ion-elgreco

### DIFF
--- a/crates/core/src/operations/create.rs
+++ b/crates/core/src/operations/create.rs
@@ -62,6 +62,7 @@ pub struct CreateBuilder {
     log_store: Option<LogStoreRef>,
     configuration: HashMap<String, Option<String>>,
     metadata: Option<HashMap<String, Value>>,
+    raise_if_key_not_exists: bool,
 }
 
 impl super::Operation<()> for CreateBuilder {}
@@ -87,6 +88,7 @@ impl CreateBuilder {
             log_store: None,
             configuration: Default::default(),
             metadata: Default::default(),
+            raise_if_key_not_exists: true,
         }
     }
 
@@ -196,6 +198,12 @@ impl CreateBuilder {
         self
     }
 
+    /// Specify whether to raise an error if the table properties in the configuration are not DeltaConfigKeys
+    pub fn with_raise_if_not_exists(mut self, raise_if_key_not_exists: bool) -> Self {
+        self.raise_if_key_not_exists = raise_if_key_not_exists;
+        self
+    }
+
     /// Specify additional actions to be added to the commit.
     ///
     /// This method is mainly meant for internal use. Manually adding inconsistent
@@ -279,7 +287,7 @@ impl CreateBuilder {
                 .iter()
                 .map(|(k, v)| (k.clone(), v.clone().unwrap()))
                 .collect::<HashMap<String, String>>(),
-            true,
+            self.raise_if_key_not_exists,
         )?;
 
         let protocol = convert_properties_to_features(protocol, &configuration);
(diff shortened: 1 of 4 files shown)


ion-elgreco:
Can you expose this to python as well

vegarsti:
Yeah! The approach is ok?

vegarsti:
Would you mind pointing me to the relevant Python file? 😄 Is it `deltalake/create.py`?

ion-elgreco:
> Would you mind pointing me to the relevant Python file? 😄 Is it `deltalake/table.py`?

Yes there! And python/src/lib.rs

vegarsti:
Thanks! Added now, let me know if it doesn't make sense. Will add the test you mentioned as well.

python/deltalake/table.py
```diff
@@ -457,6 +457,7 @@ def create(
         configuration: Optional[Mapping[str, Optional[str]]] = None,
         storage_options: Optional[Dict[str, str]] = None,
         custom_metadata: Optional[Dict[str, str]] = None,
+        raise_if_key_not_exists: Optional[bool] = True,
```
ion-elgreco, line 460: I would keep it required

vegarsti: Yeah, fair! Should I move it further up in the signature, then, i.e. before the Optional ones?

vegarsti: or wait, I'll still use the default True, so it should be fine

ion-elgreco: Best to keep it aligned across :)

vegarsti: Not sure what you mean by that 🤔  Had to move it further up in `create_deltalake`: The Rust code was failing to compile because of the ambigious ordering (it wants non-optional before optional ones).

vegarsti:
BTW the order in the `create` classmethod and and `create_deltalake` aren't exactly the same

vegarsti:
Added test to the Rust code now

vegarsti:
Should I squash the Rust and Python commits?

ion-elgreco:
> Should I squash the Rust and Python commits?

No worries, I'll do that with the merge commit

ion-elgreco (approved):
Thankss @vegarsti!

vegarsti:
Thanks for helping so quickly!

crates/core/src/operations/create.rs
```diff
@@ -196,6 +198,12 @@ impl CreateBuilder {
         self
     }
 
+    /// Specify whether to raise an error if the table properties in the configuration are not DeltaConfigKeys
+    pub fn with_raise_if_not_exists(mut self, raise_if_key_not_exists: bool) -> Self {
```
vegarsti, line 202: oops @ion-elgreco I messed this up -- should have been `with_raise_key_if_not_exists` for consistency

ion-elgreco: @vegarsti oops :), you can make a small follow up PR

vegarsti: https://github.com/delta-io/delta-rs/pull/2569

Issue #1778, opened 2023-10-26
Unable to write high-precision decimal values to Delta table using serde_json/JsonWriter


ryanaston:
# Environment

**Delta-rs version**: 
0.15.0 (also tried 0.16.1)

**Binding**: 
Rust

**Environment**:
- **Cloud provider**: any
- **OS**: any
- **Other**: n/a

***
# Bug

**What happened**:
Writes began failing when attempting to insert high-precision decimal values into a Delta table using the JsonWriter with a `Vec<serde_json::Value>`.  Discovered `serde_json` was deserializing these values as strings in scientific notation which could not be parsed into the Arrow DecimalType:

`Generic DeltaTable error: Failed to convert into Arrow schema: Parser error: can't parse the string value 3.9178294781e-6 to decimal`

Some digging uncovered the `serde_json` feature flag `"arbitrary_precision"` which retains the value in its full form stored in string format, however this too cannot be decoded to an Arrow DecimalType:

`Generic DeltaTable error: Failed to convert into Arrow schema: Json error: whilst decoding field 'decimal_col': expected decimal got {"$serde_json::private::Number": "0.0000039178294781"}`

**What you expected to happen**:
High-precision decimal values be written accurately and successfully to a Delta table

**How to reproduce it**:

**Cargo.toml**
```
[package]
name = "decimal_issue"
version = "0.1.0"
edition = "2021"

[dependencies]
deltalake = "0.15.0"
serde = "1"
serde_json = { version = "1", features = ["arbitrary_precision"] } # remove feature to see original behavior
tokio = "1.33.0"
```
**src/main.rs**
```
use deltalake::{operations::create::CreateBuilder, writer::{JsonWriter, DeltaWriter}};

#[tokio::main]
async fn main() {
    let data = serde_json::from_str::<Vec<serde_json::Value>>(r#"[{"decimal_col": 0.0000039178294781}]"#).unwrap();
    let table = CreateBuilder::new().with_location("memory://").with_column("decimal_col", deltalake::SchemaDataType::primitive("decimal(38,16)".to_string()), true, None).await.unwrap();
    let mut writer = JsonWriter::for_table(&table).unwrap();

    match writer.write(data).await {
        Ok(_) => {},
        Err(err) => {
            eprintln!("{}", err);
            std::process::exit(1);
        }
    }
}
```

`cargo run`

**More details**:
Lower precision decimals (5 or less) do not have this issue.

[...]

PR #2581, opened 2024-06-09
chore: bump to datafusion 39, arrow 52, pyo3 0.21

abhiaagarwal:
# Description
Updates the arrow and datafusion dependencies to 52 and 39(-rc1) respectively. This is necessary for updating pyo3.

While most changes with trivial, some required big rewrites. Namely, the logic for the Updates operation had to be rewritten (and simplified) to accommodate some new sanity checks inside datafusion: (https://github.com/apache/datafusion/pull/10088).

Depends on delta-kernel having its arrow and object-store version bumped as well. This PR doesn't include any major changes for pyo3, I'll open a separate PR depending on this PR.

# Related Issue(s)

# Documentation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,33 +26,33 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
-delta_kernel = { version = "0.1" }
+delta_kernel = { git = "https://github.com/abhiaagarwal/delta-kernel-rs", branch = "arrow_52" }
 # delta_kernel = { path = "../delta-kernel-rs/kernel" }
 
 # arrow
-arrow = { version = "51" }
-arrow-arith = { version = "51" }
-arrow-array = { version = "51", features = ["chrono-tz"] }
-arrow-buffer = { version = "51" }
-arrow-cast = { version = "51" }
-arrow-ipc = { version = "51" }
-arrow-json = { version = "51" }
-arrow-ord = { version = "51" }
-arrow-row = { version = "51" }
-arrow-schema = { version = "51" }
-arrow-select = { version = "51" }
-object_store = { version = "0.9" }
-parquet = { version = "51" }
+arrow = { version = "52" }
+arrow-arith = { version = "52" }
+arrow-array = { version = "52", features = ["chrono-tz"] }
+arrow-buffer = { version = "52" }
+arrow-cast = { version = "52" }
+arrow-ipc = { version = "52" }
+arrow-json = { version = "52" }
+arrow-ord = { version = "52" }
+arrow-row = { version = "52" }
+arrow-schema = { version = "52" }
+arrow-select = { version = "52" }
+object_store = { version = "0.10.1" }
+parquet = { version = "52" }
 
 # datafusion
-datafusion = { version = "37.1" }
-datafusion-expr = { version = "37.1" }
-datafusion-common = { version = "37.1" }
-datafusion-proto = { version = "37.1" }
-datafusion-sql = { version = "37.1" }
-datafusion-physical-expr = { version = "37.1" }
-datafusion-functions = { version = "37.1" }
-datafusion-functions-array = { version = "37.1" }
+datafusion = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
+datafusion-expr = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
+datafusion-common = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
+datafusion-proto = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
+datafusion-sql = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
+datafusion-physical-expr = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
+datafusion-functions = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
+datafusion-functions-array = { git = "https://github.com/apache/datafusion", tag = "39.0.0-rc1" }
 
 # serde
 serde = { version = "1.0.194", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
-delta_kernel = { git = "https://github.com/abhiaagarwal/delta-kernel-rs", branch = "arrow_52" }
+delta_kernel = { version = "0.1.1" }
 # delta_kernel = { path = "../delta-kernel-rs/kernel" }
 
 # arrow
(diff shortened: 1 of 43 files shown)


crates/core/src/operations/update.rs
```diff
@@ -307,66 +304,24 @@ async fn execute(
         },
     ));
 
-    // Perform another projection but instead calculate updated values based on
-    // the predicate value.  If the predicate is true then evalute the user
-    // provided expression otherwise return the original column value
-    //
-    // For each update column a new column with a name of __delta_rs_ + `original name` is created
```
abhiaagarwal, line 314: I would really appreciate if someone could take a look at this diff — I think it should replicate the exact logic except heavily simplified and it passes all tests, but there could be an edge case I'm missing.

abhiaagarwal:
This is now ready for review. All python and rust tests are passing locally on my machine (docs seem to have issues building, though). This only enables the gil-refs feature in the python bindings, it does not migrate to the new API (will be done in a follow-up PR)

crates/core/src/delta_datafusion/expr.rs
```diff
@@ -276,33 +272,6 @@ impl<'a> Display for SqlFormat<'a> {
                     write!(f, "{expr} IN ({})", expr_vec_fmt!(list))
                 }
             }
-            Expr::GetIndexedField(GetIndexedField { expr, field }) => match field {
```
ion-elgreco, line 279: Why is this being removed?

crates/core/src/operations/update.rs
ion-elgreco: Looks fine to me the change, before it was creating these __delta_rs columns but seems unnecessary to first add those columns and then select only those. Doing this in one projection seems much better

crates/core/src/delta_datafusion/expr.rs
abhiaagarwal: Those functions don't exist before, and based on my understanding from https://github.com/apache/datafusion/pull/10568, those rewrites happen earlier now so the functionality stays the same, it's just done by DF

crates/core/src/operations/update.rs
abhiaagarwal: Believe me when I say that I stared at this for a couple hours before figuring out what it was doing 🥳 should probably give a nice speed increase too, I didn't benchmark the old code but I did benchmark the projection operation in general and it's very expensive

crates/core/src/delta_datafusion/expr.rs
ion-elgreco: Just wondering if we can catch the new functions and keep the old display formatting

python/src/filesystem.rs
```diff
@@ -590,31 +580,23 @@ impl ObjectOutputStream {
 
     fn write(&mut self, data: &PyBytes) -> PyResult<i64> {
         self.check_closed()?;
-        let len = data.as_bytes().len() as i64;
         let py = data.py();
-        let data = data.as_bytes();
-        let res = py.allow_threads(|| match rt().block_on(self.writer.write_all(data)) {
```
ion-elgreco, line 596: Did you forget re-adding this? Now the GIL will be in place

python/src/filesystem.rs
```diff
@@ -590,31 +580,23 @@ impl ObjectOutputStream {
...
-            }
-        })?;
-        self.buffer_size += len;
-        if self.buffer_size >= self.max_buffer_size {
-            let _ = self.flush(py);
-            self.buffer_size = 0;
+        let bytes = data.as_bytes();
+        self.buffer.extend_from_slice(bytes);
+        let len = bytes.len();
+        if self.max_buffer_size >= self.buffer.content_length() {
```
ion-elgreco, line 587: This should be the other way around, right?

To only flush when the current buffer is greater than the max_buffer_size (this was added a while ago to prevent deadlocks)

crates/core/src/delta_datafusion/expr.rs
abhiaagarwal: Hmmmm, I'm not entirely sure this is possible. Those operations have been converted into ScalarFunctions, so the concept of "taking an index of a field" is now "a scalar function that operates on a field and an index". Had to modify a bunch of tests to accommodate. https://github.com/delta-io/delta-rs/blob/af5997cf7dd8cce97b07b5c8c78e23cb7092549f/crates/core/src/delta_datafusion/expr.rs#L656-L659

It could be possible manually match on scalar functions and convert them into the old style, though.

python/src/filesystem.rs
abhiaagarwal: Yep, but the write function isn't async or multithreaded (it just extends a buffer in memory) so I don't believe releasing the GIL has any impact there? It releases it in `flush` if it gets called.

python/src/filesystem.rs
abhiaagarwal: Haha yep! Explains that one odd test failure, thank you for catching.

python/src/filesystem.rs
ion-elgreco: > Yep, but the write function isn't async or multithreaded (it just extends a buffer in memory) so I don't believe releasing the GIL has any impact there? It releases it in `flush` if it gets called. 

I think it's more that some folks want to run stuff in parallel/threaded and just to release the GIL whenever possible, doesn't hurt to keep it that way with allow_threads

abhiaagarwal: added it back! also figured I'd add a check in `close` to make sure we're only flushing if the buffer has data inside it.

crates/core/src/delta_datafusion/expr.rs
ion-elgreco: Hmm, it would make the roundtrippable format closer to SQL, but I guess with the timestamp parsing I already kind of stopped doing that 😆

ion-elgreco (approved):

ion-elgreco:
Nice work! @abhiaagarwal

abhiaagarwal:
Thank you @ion-elgreco! DF 39 should be released later this week, and the upstream delta-kernel needs to be resolved as well. I'll open a follow-up PR for moving to the new pyo3 lifetimes, and I'm also interested in implementing an async-native API using [`pyo3-asyncio`](https://github.com/awestlake87/pyo3-asyncio) since a good chunk of my work code can benefit from it :)

ion-elgreco:
@abhiaagarwal yeah leaving it unmerged for now. Also checking with rest on slack if they are fine with keeping the git reference of delta-kernel-rs for time being.

Great, you can ping me when you got the new bound APIs working!

Just two things regarding pyo3-asyncio. It seems the original maintainer is not very active anymore, so it's still locked to pyo3 0.20 unfortunately. 

The other thing is, where do you need it for ;) as most of the async stuff is internal

abhiaagarwal:
@ion-elgreco yeah, but it appears that the maintainer of pyo3 is interested in taking stewardship over it, so maybe that'll be resolved soon.

I'm mostly interested to see if there are any perf improvements. Most of the code I use with the python deltalake package in already is async native (ie. async main), having async methods won't "infect" the rest of our codebase. We use it as a pseudo backend with FastAPI, for reference. There appears to be work being done in the main pyo3 repo that'll allow the gil to be released in async methods, so we could have the best of both worlds.

ion-elgreco:
@abhiaagarwal can you bump delta-kernel-rs to 0.1.1, it includes your changes now. And also bump the python cargo toml to 0.18.1?

abhiaagarwal:
@ion-elgreco sure, I'll do it by EOD. Also will bump DF since they released 39

Cargo.toml
```diff
@@ -26,33 +26,33 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
-delta_kernel = { version = "0.1" }
+delta_kernel = { git = "https://github.com/abhiaagarwal/delta-kernel-rs", branch = "arrow_52" }
```
rtyler, line 29: this should now be updated to 0.1.1

abhiaagarwal:
Bumped to published versions, also added a new test that seemingly confirms that the issues related to decimals in #1778 et al on both engines seem to be fixed

ion-elgreco:
> Bumped to published versions, also added a new test that seemingly confirms that the issues related to decimals in #1778 et al on both engines seem to be fixed 

Yes I was waiting for some time for the arrow bump since they added Scientific notation support for decimals :)

Thanks for adding the test! 🙂

ion-elgreco (approved):

PR #2598, opened 2024-06-14
test: reintroduce azurite SAS integration tests

giacomorebecchi:
# Description
Reintroduce integration tests using Azurite and SAS tokens.

These tests were skipped believing that the mistake was due to the upgrade of version in `object-store>=0.9.1". Actually, what `object-store` started doing from that version on, is really using the SAS token for the first time. For more details, see:
- https://github.com/apache/arrow-rs/issues/5475#issuecomment-2084748858

I changed the command to create the SAS token using the `azure-storage-blob` library (adding it to the dev deps), rather than the azure CLI, and removed the pytest markers to skip the tests.

@ion-elgreco pinging you as promised, sorry if this took more time than expected!

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -27,6 +27,7 @@ pandas = [
     "pandas"
 ]
 devel = [
+    "azure-storage-blob==12.20.0",
     "mypy~=1.8.0",
     "ruff~=0.3.0",
     "packaging>=20",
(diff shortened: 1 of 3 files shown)


ion-elgreco:
Thankss! <3

ion-elgreco (approved):

Issue #2602, opened 2024-06-15
panic in `push_down_filter`


ryzhyk:
# Environment

**Delta-rs version**:  v0.18.0

**Binding**:  Rust

**Environment**:
- **Cloud provider**: na
- **OS**: Linux
- **Other**:

***
# Bug

**What happened**:

Running the following query: `select * from snapshot where id > 10000 and id < 20000` against a Delta table panics with:

```
called `Result::unwrap()` on an `Err` value: Context("Optimizer rule 'push_down_filter' failed", Internal("Vec returned length: 1 from supports_filters_pushdown is not the same size as the filters passed, which length is: 2"))
```

If I remove the second condition and just leave `select * from snapshot where id > 10000`, it works. I believe this is a regression in v0.18.0, since this worked before.  

**What you expected to happen**:

**How to reproduce it**:

Here is a complete repro:

```rust
use std::sync::Arc;
use deltalake::{datafusion::prelude::SessionContext, DeltaTableBuilder};

#[tokio::main(flavor = "multi_thread", worker_threads = 10)]
async fn main() {

    let datafusion = SessionContext::new();

    let table_builder = DeltaTableBuilder::from_uri("data");
    let delta_table = Arc::new(table_builder.load().await.unwrap());
    datafusion.register_table("snapshot", delta_table).unwrap();

    let df = datafusion
        .sql("select * from snapshot where id > 10000 and id < 20000")
        .await
        .unwrap();

    df.collect().await.unwrap();

}
```

where `data` contains an empty delta table. 

Here is a complete repro, including Rust code and the empty table: [conjunctive_join.zip](https://github.com/user-attachments/files/15845071/conjunctive_join.zip).  Just run `cargo run` to reproduce.

**More details**:

rtyler:
thanks for the repro code, we'll take a looksee

rtyler:
I have taken your test code and created a failing test in this pull request which we can start to fix #2604

adamfaulkner-at:
Hi, I think you need to apply the same fix as https://github.com/delta-io/delta-rs/pull/2604 for `DeltaTableProvider`.

ion-elgreco:
@adamfaulkner-at you have time to put in a fix?

rtyler:
I have done some regression testing with the provided example and a unit test and was not able to reproduce this with our contemporary version of delta-rs

PR #2604, opened 2024-06-15
chore: implement regression test for push down panic

rtyler:
See #2602

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-core"
-version = "0.18.0"
+version = "0.18.1"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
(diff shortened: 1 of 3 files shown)


abhiaagarwal:
@rtyler I haven't tested this yet, but try:
https://github.com/delta-io/delta-rs/blob/df6e6fd1856286487e04a9b6d916ee14c3951808/crates/core/src/delta_datafusion/mod.rs#L699-L704

```rust
Ok(filter.into_iter().map(|_| TableProviderFilterPushDown::Inexact).collect())
```

or something of this nature.

rtyler:
@abhiaagarwal I gave that a try in a small slot of time I had yesterday with no success, I haven't been able to spend more time with the problem yet :frowning:

abhiaagarwal:
@rtyler there's also the same slip of logic below it, did you change both spots?

ion-elgreco (approved):

PR #2606, opened 2024-06-16
fix(python): constrain multipart upload size to fixed length

### LINKED ISSUE #2605: 0.18.1 reintroduces S3 multipart upload bug
# Environment

**Delta-rs version**: 0.18.1

**Binding**:  Python 

**Environment**: 
- **Cloud provider**: AWS
- **OS**: Linux/Windows
- **Other**:

***
# Bug

**What happened**: Same error as https://github.com/delta-io/delta-rs/issues/890 but on S3 directly instead of non-S3

**How to reproduce it**: Write regular size data to S3 with write_deltalake()

**More details**: 0.18.0 works fine

abhiaagarwal:
# Description

Object stores expected fixed lengths for all multipart upload parts right up until the last part. The original logic just flushed when it exceeded the threshold. Now, it flushes when the threshold is met exclusively with the same fixed buffer, unless we're completing the transaction, in which case the last piece is allowed to be smaller.

Bumps the constant to reflect that the minimum expected size by most object stores is 5MiB. Also adds a UserWarning if a constant is specified to be less. 

Also releases the GIL in more places by moving the flushing logic to a free function. 

# Related Issue(s)

Closes #2605 

# Documentation

See: [MultipartUpload](https://docs.rs/object_store/latest/object_store/trait.MultipartUpload.html) docs

### DIFF
--- a/python/src/filesystem.rs
+++ b/python/src/filesystem.rs
@@ -1,5 +1,5 @@
 use crate::error::PythonError;
-use crate::utils::{delete_dir, rt, walk_tree};
+use crate::utils::{delete_dir, rt, walk_tree, warn};
 use crate::RawDeltaTable;
 use deltalake::storage::object_store::{MultipartUpload, PutPayloadMut};
 use deltalake::storage::{DynObjectStore, ListResult, ObjectStoreError, Path};
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::sync::Arc;
 
-const DEFAULT_MAX_BUFFER_SIZE: usize = 4 * 1024 * 1024;
+const DEFAULT_MAX_BUFFER_SIZE: usize = 5 * 1024 * 1024;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub(crate) struct FsConfig {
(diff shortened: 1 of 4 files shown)


python/src/filesystem.rs
```diff
@@ -306,6 +307,19 @@ impl DeltaFileSystemHandler {
             .map_or(DEFAULT_MAX_BUFFER_SIZE, |v| {
                 v.parse::<usize>().unwrap_or(DEFAULT_MAX_BUFFER_SIZE)
             });
+        if max_buffer_size < DEFAULT_MAX_BUFFER_SIZE {
```
ion-elgreco, line 310: It seems that below 5MB should never work, so shouldn't we just panic/raise here.

ion-elgreco:
@abhiaagarwal looks good, just one remark on the buffer size warning

python/src/filesystem.rs
abhiaagarwal: It *could* work, a 5MiB (note: MiB) limit is mostly an implementation detail that most cloud providers share. It isn't necessarily set in stone by any spec or anything, you could self host your own minio and make the limit whatever you want. I'm just putting that warning there to make sure that end users know what they're getting into if they tweak around with things

abhiaagarwal:
@ion-elgreco don't merge yet — I found about an edge case in my logic that won't work (what happens when remaining = max_buffer_length = len(chunk)?) and I want to fix up the test to actually parameterize it + check for the warning message. I also opened an upstream issue in object_store to pull out the "fixed buffer" logic into its own struct so it can live upstream, for the future :)

ion-elgreco (approved):
Thankss!

Issue #2478, opened 2024-05-03
Cannot merge to a table with a timestamp column after upgrading delta-rs


echai58:
# Environment

**Delta-rs version**: 0.15.3 => 0.17.2

**Binding**: python

***
# Bug

**What happened**:
I have a lot of tables created via delta-rs from an older version, around 0.15.3. I'm looking to upgrade, but I'm running into breaking errors, I think from the 0.16.0 breaking change to how timestamps work.

I have tables that have a `timestamp` column, and due to older versions of delta-rs, they are written without a timezone. When I upgraded to 0.17.2, and I read in the table, the schema says `timestamp[us, tz=UTC]` - okay, that's reasonable, it's attaching a UTC timezone.

But, when I try to write to the table via a merge and the predicate compares the timestamp columns, it fails with 
```
DeltaError: Generic DeltaTable error: External error: Arrow error: Invalid argument error: Invalid comparison operation: Timestamp(Microsecond, Some("UTC")) == Timestamp(Microsecond, None)
```

My theory is that delta-rs is interpreting the schema of the table as having a UTC timezone, whereas the physical data actually reflects the table having no timezone. Thus, when it casts the source table I pass in, it applies the UTC timezone, which Arrow then cannot compare against the physical table, which has no timezone.

**What you expected to happen**:
Is there any way to upgrade delta-rs such that this doesn't break? Would be really unfortunate to have to rewrite all of these tables because of this breaking change.

**How to reproduce it**:
Run this with an older (< 0.16) delta-rs:
```
t = pa.Table.from_pandas(pd.DataFrame({"p": [1],"a": [pd.Timestamp("2022-01-01")], "b": [1], }), schema=pa.schema([("p", pa.int64()), ("a", pa.timestamp("us", tz="UTC")), ("b", pa.int64())]))

write_deltalake(
    "test",
    t,
    mode="overwrite",
    partition_by=["p"],
)
```

Then, run this with a new delta-rs:
```
new_data = pa.Table.from_pandas(pd.DataFrame({"p": [1],"a": [pd.Timestamp("2022-01-01")], "b": [4]}), schema=pa.schema([("p", pa.int64()), ("a", pa.timestamp("us", tz="UTC")), ("b", pa.int64())]))

dt.merge(
    source=new_data,
    predicate="s.p = t.p and s.a = t.a",
    source_alias="s",
    target_alias="t",
).when_matched_update_all().when_not_matched_insert_all().execute()
```

[...]

Issue #2341, opened 2024-03-25
Inconsistent arrow timestamp type breaks datafusion query


qinix:
# Environment

**Delta-rs version**: current main(https://github.com/delta-io/delta-rs/commit/abafd2d0cb8dde32ffa990dc30fb97a5581688ec)

**Binding**: rust, python

***
# Bug

**What happened**:

Prior versions of this library serialize `PrimitiveType::Timestamp` as `ArrowDataType::Timestamp(TimeUnit::Microsecond, None)`, while PR #2236 changed this behavior to serialize as `ArrowDataType::Timestamp(TimeUnit::Microsecond, Some("UTC"))`, leads to different timestamp schema in different parquet file.

When querying delta lake with Arrow Datafusion and filtering by timestamp column, datafusion throws `Invalid comparison operation: Timestamp(Microsecond, Some("UTC")) < Timestamp(Microsecond, None)`. After changing the filter type to `Timestamp(Microsecond, Some("UTC"))`, datafusion throws `Invalid comparison operation: Timestamp(Microsecond, None) < Timestamp(Microsecond, Some("UTC"))`. The left side of the comparison operation is a timestamp column in delta lake.

ion-elgreco:
This actually had breaking effects since the prior behavior was incorrect. I suggest you rewrite your tables so that the parquet timestamp is probably encoded

ion-elgreco:
@qinix I assume you have this issue only on tables written with older versions and then read with the latest main?

qinix:
> @qinix I assume you have this issue only on tables written with older versions and then read with the latest main?

Yes, it is

cmettler:
i have a similar issue when trying to delete records via a UTC timestamp field. I was not able to create a UTC timestamp for the right side of the predicate:

pa = duckdb.sql("select now()::timestamptz at time zone 'CET' at time zone 'UTC' as ts" ).to_arrow_table()
write_deltalake(target_table, pa,engine="rust",storage_options=storage_options,mode='overwrite')
dt = DeltaTable(target_table,storage_options=storage_options)
dt.delete("ts >= to_timestamp_micros('2024-03-27 00:00:00Z','%Y-%m-%d %H:%M:%S%#z')")

fails with: 
ValueError: Invalid comparison operation: Timestamp(Microsecond, Some("UTC")) >= Timestamp(Microsecond, None)

ravid08:
[...]

PR #2615, opened 2024-06-21
feat(rust,python): cast each parquet file to delta schema

HawaiianSpork:
# Description

By casting the read record batch to the delta schema datafusion can read tables where the underlying parquet files can be cast to the desired schema. Fixes:

- Errors querying data where some of the parquet files may not have columns that were added later because of schema migration.  This includes nested columns for structs that are in Maps, Lists, or children of other structs
- maps and lists written with different different element names
- timestamps of different units.
- Any other cast supported by arrow-cast.

This can be done now since data-fusion exposes a SchemaAdapter which can be overwritten.

We should note that this makes all times being read by delta-rs as having microsecond precision to match the Delta protocol.

# Related Issue(s)
- This makes solving #2478 and #2341 just a matter of adding code to delta-rs cast.

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -39,7 +39,8 @@ use arrow_cast::display::array_value_to_string;
 use arrow_schema::Field;
 use async_trait::async_trait;
 use chrono::{DateTime, TimeZone, Utc};
-use datafusion::datasource::file_format::{parquet::ParquetFormat, FileFormat};
+use datafusion::datasource::file_format::FileFormat;
+use datafusion::datasource::physical_plan::parquet::ParquetExecBuilder;
 use datafusion::datasource::physical_plan::{
     wrap_partition_type_in_dict, wrap_partition_value_in_dict, FileScanConfig,
 };
@@ -76,6 +77,7 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 
 use crate::delta_datafusion::expr::parse_predicate_expression;
+use crate::delta_datafusion::schema_adapter::DeltaSchemaAdapterFactory;
 use crate::errors::{DeltaResult, DeltaTableError};
 use crate::kernel::{Add, DataCheck, EagerSnapshot, Invariant, Snapshot, StructTypeExt};
 use crate::logstore::LogStoreRef;
@@ -92,6 +94,7 @@ pub mod logical;
 pub mod physical;
 
 mod find_files;
+mod schema_adapter;
 
 impl From<DeltaTableError> for DataFusionError {
     fn from(err: DeltaTableError) -> Self {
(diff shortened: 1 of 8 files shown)


crates/core/src/delta_datafusion/mod.rs
```diff
@@ -618,35 +615,30 @@ impl<'a> DeltaScanBuilder<'a> {
...
-            None
+        if let Some(predicate) = logical_filter {
+            if config.enable_parquet_pushdown {
+                exec_plan_builder = exec_plan_builder.with_predicate(predicate);
+            }
         };
 
-        let scan = ParquetFormat::new()
-            .create_physical_plan(
-                self.state,
```
HawaiianSpork, line 632: State was previously passed in but was not used by the physical plan.

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -618,35 +615,30 @@ impl<'a> DeltaScanBuilder<'a> {
...
+            object_store_url: self.log_store.object_store_url(),
+            file_schema,
+            file_groups: file_groups.into_values().collect(),
+            statistics: stats,
+            projection: self.projection.cloned(),
+            limit: self.limit,
+            table_partition_cols,
+            output_ordering: vec![],
+        })
+        .with_schema_adapter_factory(Arc::new(DeltaSchemaAdapterFactory {}));
```
ion-elgreco, line 628: Why is the correct schema not passed in the factory?

HawaiianSpork: oops, fixed

ion-elgreco:
@HawaiianSpork can you add a  test where  we have  a delta table that contains parquets with nanosecond timestamps in the files. Maybe just create a parquet table and then use convert to delta?

HawaiianSpork:
> @HawaiianSpork can you add a test where we have a delta table that contains parquets with nanosecond timestamps in the files. Maybe just create a parquet table and then use convert to delta?

@ion-elgreco, I'd be happy to add more tests but want to make sure I create the correct ones. `../test/tests/data/table_with_edge_timestamps` data has parquet files with nanosecond timestamp precision, you can see how this change leads datafusion only seeing microsecond precision.  Do you think this is fine?

FYI, @wjones127 and @roeap for making the original commit that read the schema from the parquet files: #1266.

rtyler:
This looks promising, but I would like to update the title if you don't mind for the changelog in the future. Schema evolution is typically understood in the Delta context as changes to the Delta schema (i.e. a transaction commit occurs). 

I am understanding this correctly it's more about schema adaptation on reads

ion-elgreco (approved):

ion-elgreco:
@HawaiianSpork can you rebase so that we can merge?

HawaiianSpork:
@ion-elgreco I have merged with master as requested.

PR #2618, opened 2024-06-24
fix: enable parquet pushdown for DeltaScan via TableProvider impl for DeltaTable

alexwilcoxson-rel:
# Description
The DeltaScanBuilder used by DeltaTable's TableProvider impl does not specify DeltaScanConfig. The builder uses Default for DeltaScanConfig and thus enable_parquet_pushdown is disabled.

This change makes it so DeltaScanConfig is an Option on the scan builder. If unset during build a default value will be created using the DeltaScanConfigBuilder rather than DeltaScanConfig::Default. The former has enable_parquet_pushdown defaulted to true. I considered changing the latter but it is used in a couple other places where I was not sure of the impact.

# Related Issue(s)

# Documentation

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,13 +43,13 @@ jobs:
           override: true
 
       - name: build and lint with clippy
-        run: cargo clippy --features azure,datafusion,s3,gcs,glue --tests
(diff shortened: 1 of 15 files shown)


HawaiianSpork (approved):

rtyler (dismissed):

rtyler:
moved to #2637

Issue #1998, opened 2023-12-28
Generic DeltaTable error: type_coercion in Struct column in merge operation


gustavodecarlo:
# Environment

**Delta-rs version**: rust-v0.16.5

**Binding**: python-v0.14.0

**Environment**: local machine
- **Cloud provider**:
- **OS**: macOS Sonoma 14.2.1
- **Other**: python version 3.9.13

***
# Bug

**What happened**:
_internal.DeltaError: Generic DeltaTable error: type_coercion
caused by
Error during planning: Failed to coerce then ([Struct([Field { name: "created_datetime", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "version", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "created_datetime", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "version", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "created_datetime", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "version", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "created_datetime", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "version", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "created_datetime", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "version", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }])]) and else (None) to common types in CASE WHEN expression

**What you expected to happen**:

merge executed successfully

**How to reproduce it**:
Need to pip install both polars and deltalake
Same code in this repo: https://github.com/gustavodecarlo/poc-python-deltars-upsert/blob/main/poc_python_deltars_upsert/dataflow.py#L12

```python
from deltalake import DeltaTable, Schema
import polars as pl

df = pl.DataFrame(
        {
            "sales_order_id": ["1000", "1001", "1002", "1003"],
[...]

PR #2644, opened 2024-07-03
chore: add a reproduction case for merge failures with struct<string>

rtyler:
```
E       _internal.DeltaError: Generic DeltaTable error: type_coercion
E       caused by
E       Error during planning: Failed to coerce then ([Struct([Field { name: "city", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "state", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "street", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "city", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "state", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "street", data_type: LargeUtf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "city", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "state", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "street", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "city", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "state", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "street", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), Struct([Field { name: "city", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "state", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "street", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }])]) and else (None) to common types in CASE WHEN expression
```

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -825,7 +825,7 @@ class DeltaTable:
         target_alias: Optional[str] = None,
         error_on_type_mismatch: bool = True,
         writer_properties: Optional[WriterProperties] = None,
-        large_dtypes: bool = True,
+        large_dtypes: bool = False,
         custom_metadata: Optional[Dict[str, str]] = None,
     ) -> "TableMerger":
         """Pass the source data which you want to merge on the target delta table, providing a
--- a/python/tests/test_merge.py
+++ b/python/tests/test_merge.py
@@ -970,17 +970,18 @@ def test_struct_casting(tmp_path: pathlib.Path):
     assert not df.empty
 
     schema = pa.Table.from_pandas(df=df).schema
-
     dt = DeltaTable.create(tmp_path, schema, name="test")
     metadata = dt.metadata()
     assert metadata.name == "test"
 
     result = (
         dt.merge(
-            source=df_merge, predicate="t.id = s.id", source_alias="s", target_alias="t"
+            source=df_merge,
+            predicate="t.id = s.id",
+            source_alias="s",
+            target_alias="t",
         )
         .when_matched_update_all()
-        .when_not_matched_insert_all()
         .execute()
     )
     assert result is not None
--- a/python/tests/test_merge.py
+++ b/python/tests/test_merge.py
@@ -940,3 +940,47 @@ def test_merge_field_special_characters_delete_2438(tmp_path: pathlib.Path):
     expected = pa.table({"x": [1], "y--1": [4]})
 
     assert dt.to_pyarrow_table() == expected
+
+
+@pytest.mark.pandas
+def test_struct_casting(tmp_path: pathlib.Path):
+    import pandas as pd
+
+    cols = ["id", "name", "address", "scores"]
+    data = [
+        (
+            2,
+            "Marry Doe",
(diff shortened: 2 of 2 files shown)


rtyler:
@ion-elgreco this test can pass by passing a non-default `large_dtypes=False` to the `merge()` function. In resolving the original issue (#1998) with #2003 you alluded to better type coercion support in Datafusion.

This test is a simple reproduction case of the [paper cut](https://en.wikipedia.org/wiki/Paper_cut_bug) hit by a `deltalake` user.  It was not clear to the user why `deltalake` enlarges the type of a `pandas.DataFrame` in a way that breaks a simple merge in the default case. I'm inclined to agree, but I'm not sure what the solution should be here.

Changing the default value of ``large_dtypes=False` results in this, and all the other tests passing :laughing:   I believe that the upcast/enlarging should not be the default behavior, since it clearly breaks such a simple example, but I don't have a good enough understanding of why the upcast was done by default in the first place :confused:

ion-elgreco:
@rtyler I don't recall why this was there. But we should ideally drop all this large dtypes logic and improve the schema checking so that it's done on delta schema level and not arrow.

Also the cast kernels need to be improved so that you don't cast large utf8 to utf8 or utf8-view to utf8...

If we do it like that then all issues should be resolved I believe

rtyler:
@ion-elgreco what are your thoughts on changing the default behavior of the function? In this reproduction case the source table is String, the source dataframe is Utf8/String and it is only the default behavior of the `deltalake` package that is turning anything into `Largeutf8` :facepalm:

ion-elgreco:
@rtyler not sure why I left on true, probably by accident. I am fine with changing it to False, but we are just hiding the problem now : P

Perhaps an idea would be to have SOURCE record batch arrow schema used to read the target files with. Let's say we have this table_schema `{"A": "string", "B": "string"}`, in our code it changes the delta schema to arrow schema with non large types always, so `{"A": 'utf8', "B":"utf8"}`, however the incoming data might be in  `{"A": 'large-utf8', "B":"utf8view"}` or any other combination.

So perhaps the solution to all this, would be that delta scans are done with the record batch arrow schema of the incoming data or are casted to that schema, so that source and target is in the same arrow schema. @rtyler what do you think?

ion-elgreco (approved):

PR #2649, opened 2024-07-04
feat(rust): fix size_in_bytes in last_checkpoint_ to i64

### LINKED ISSUE #2646: i32 limit in JSON stats
# Environment

**Delta-rs version**: 0.18.2

**Binding**: Python

**Environment**:
- **Cloud provider**: AWS
- **OS**: Amazon Linux 2

***
# Bug

**What happened**: 

When reading a DeltaLake table from Polars using `pl.read_delta`, I get the following error:

`DeltaProtocolError: Invalid JSON in file stats: invalid value: integer `4051124561`, expected i32 at line 1 column 70`

which ultimately comes from deltalake:

image

**What you expected to happen**: My code in Python can successfully read other tables, it is just this particular table that throws this problem. The particular table where this happens is frequently accessed and I can read it successfully through Spark, so I was expecting to read it as well from deltalake. Not sure if the int32 limitation is part of the protocol or the library should allow for bigger int types.

sherlockbeard:
# Description
see #2646 

# Related Issue(s)
closes #2646 

# Documentation

### DIFF
--- a/crates/core/src/kernel/snapshot/log_segment.rs
+++ b/crates/core/src/kernel/snapshot/log_segment.rs
@@ -377,7 +377,7 @@ struct CheckpointMetadata {
     /// The number of fragments if the last checkpoint was written in multiple parts.
     pub(crate) parts: Option<i32>,
     /// The number of bytes of the checkpoint.
-    pub(crate) size_in_bytes: Option<i32>,
+    pub(crate) size_in_bytes: Option<i64>,
     /// The number of AddFile actions in the checkpoint.
     pub(crate) num_of_add_files: Option<i32>,
     /// The schema of the checkpoint file.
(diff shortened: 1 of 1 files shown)


rtyler:
@sherlockbeard thanks for the pull request! I had been following the issue but haven't had the time to consider it further. Do you know if Spark writes int64s in the JSON stats?64-bit or larger are of unclear validity in JSON to me, since JavaScript loves 32bit numbers 

if Spark is willing to write 64bit numbers to JSON, I sure am too :smile:

sherlockbeard:
@rtyler revised my changes !.

> Do you know if Spark writes int64s in the JSON stats?

Yes

https://github.com/delta-io/delta/blob/97439835a4a667ac2ad86ec6054f0e85e8214760/spark/src/main/scala/org/apache/spark/sql/delta/LastCheckpointInfo.scala#L102-L112

looks like size and num_of_add_files are also long in spark .

rtyler (approved):

PR #2658, opened 2024-07-11
docs: add integration docs for s3 backend

avriiil:
This PR adds a dedicated docs page for using S3 as object storage.


avriiil:
@ion-elgreco - this should be better :)

rtyler (approved):

PR #2695, opened 2024-07-22
chore: refactor `write_deltalake` in `writer.py`

fpgmaas:
# Description

Te body of the `write_deltalake` function is quite long, and thus a bit difficult to read and understand. This PR proposes to reduce the complexity of the `write_deltalake` function a little bit, by extracting some logic to a separate function, and moving some function definitions to outside of the body.

I guess this is a bit opinionated, so if you think this does not add value, feel free to close the PR.

### DIFF
--- a/python/deltalake/writer.py
+++ b/python/deltalake/writer.py
@@ -64,6 +64,9 @@ else:
 PYARROW_MAJOR_VERSION = int(pa.__version__.split(".", maxsplit=1)[0])
 DEFAULT_DATA_SKIPPING_NUM_INDEX_COLS = 32
 
+DTYPE_MAP = {
+        pa.large_string(): pa.string(),
+    }
 
 class ArrowStreamExportable(Protocol):
     """Type hint for object exporting Arrow C Stream via Arrow PyCapsule Interface.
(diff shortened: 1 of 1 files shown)


python/deltalake/writer.py
```diff
@@ -662,6 +610,73 @@ def __enforce_append_only(
...
+        )
+
+    if isinstance(schema, DeltaSchema):
+        schema = schema.to_pyarrow(as_large_types=large_dtypes)
+    elif schema is None:
+        schema = data.schema
+
+    return data, schema
+
+def __sort_arrow_schema(schema: pa.schema) -> pa.schema:
```
ion-elgreco, line 673:
```suggestion
def _sort_arrow_schema(schema: pa.schema) -> pa.schema:
```
Small nit, but a single leading underscore already tells enough that its a private function, same goes for the other ones

fpgmaas: Thanks, changed that now. Mainly did this to be in line with the existing `__enforce_append_only`, so I adjusted that as well.

ion-elgreco (dismissed):

fpgmaas:
@ion-elgreco sorry, could you trigger the pipeline once again? I accidentally used `tuple` to type hint, which is not compatbile with `Python<3.9`

ion-elgreco (approved):

PR #2710, opened 2024-07-27
feat(python, rust): add `drop column` operation [BLOCKED]

ion-elgreco:
# Description
Adds a drop column operation, can be used to drop root fields and nested fields in a struct.

### DIFF
--- a/crates/core/src/operations/drop_column.rs
+++ b/crates/core/src/operations/drop_column.rs
@@ -115,14 +115,32 @@ impl std::future::IntoFuture for DropColumnBuilder {
                     .collect::<Vec<StructField>>(),
             );
 
-            let not_found: Vec<String> = fields_not_found
+            let mut not_found: Vec<String> = fields_not_found
                 .iter()
                 .map(|(key, value)| format!("{}.{}", key, value.join(".")))
                 .collect();
+            
+            // Catch root fields that do not exist
+            not_found.append(
+                &mut fields_map
+                    .values()
+                    .filter_map(|v| {
+                        if v.len() == 1 {
+                            if table_schema.field(&v[0]).is_none() {
+                                Some(v[0].clone())
+                            } else {
+                                None
+                            }
+                        } else {
+                            None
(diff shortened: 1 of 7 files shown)


crates/core/src/operations/drop_column.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub fn with_commit_properties(mut self, commit_properties: CommitProperties) -> Self {
+        self.commit_properties = commit_properties;
+        self
+    }
+
+    /// Specify if you want to raise if the specified column does not exist
+    pub fn with_raise_if_not_exists(mut self, raise: bool) -> Self {
+        self.raise_if_not_exists = raise;
+        self
+    }
```
guilhem-dvr, line 63: Question out of curiosity: What would be the use case or reason to allow drop column commits for columns that are missing from the schema?

ion-elgreco: Let's say I drop 10 columns, and one is a typo, it could still be useful for the commit to pass.

I probably need to add another check though to check if the schema actually changed, otherwise the commits are pointless

hntd187:
Don't drop columns require column mapping otherwise you have to re-write all the datafiles?

ion-elgreco:
> Don't drop columns require column mapping otherwise you have to re-write all the datafiles?

Why would it? Column mapping is when you rename them, but dropping is just a logical change

hntd187:
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-mapping says so right here in the protocol

ion-elgreco:
> https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-mapping says so right here in the protocol

It doesn't describe though why it's needed though?

Is it because you could potentially re-add the same column name but have completely different type or data?

ion-elgreco:
@hntd187 added a check to see if column mapping is enabled. I'll put in DRAFT until we have actual column mapping support

Issue #2662, opened 2024-07-13
Way to check if Delta table exists at specified path


jorritsandbrink:
# Description
It would be nice to have a way to check if a Delta table exists at a given path.

Delta Spark has the `DeltaTable.isDeltaTable()` method: https://docs.delta.io/latest/api/python/spark/index.html#delta.tables.DeltaTable.isDeltaTable

**Use Case**
When merging into a Delta table, you first define a `DeltaTable` object before calling `merge` on it:

```py
dt = DeltaTable("path_to_table_that_may_not_exist_yet")

dt.merge(...)
```

This fails with a `TableNotFoundError` if the table does not exist.

Having the `isDeltaTable` method (or similar) would enable this flow:

```py
if DeltaTable.isDeltaTable("path_to_table_that_may_not_exist_yet"):
    dt = DeltaTable("path_to_table_that_may_not_exist_yet")
    dt.merge(...)
else:
    write_deltalake(...)
```

The above requires two path lookups, which adds unnecessary latency.

Alternatively, something like this could be nice:

```py
dt = DeltaTable("path_to_table_that_may_not_exist_yet", return_none_if_not_exists=True)

if dt is not None:
    dt.merge(...)
else:
    write_deltalake(...)
```

omkar-foss:
Hey @jorritsandbrink, so there's a function called `try_get_deltatable()` in the code [here](https://github.com/delta-io/delta-rs/blob/main/python/deltalake/writer.py#L682) that provides this functionality. 

**`try_get_deltatable()`** returns `None` if the table is not found at the specified path (i.e. `TableNotFoundError` occurs), otherwise it returns the `DeltaTable` instance already existing at the specified path. It can be used with your above mentioned flow as follows:

```python

from deltalake.writer import try_get_deltatable

dt = try_get_deltatable("path_to_table_that_may_not_exist_yet")
if dt is not None:
    dt.merge(...)
else:
    write_deltalake(...)
```

And yes I agree with you, since this is a very common use case, we can abstract it to something more meaningful like `DeltaTable.isDeltaTable`.

Until then, I suppose we should consider adding the `try_get_deltatable` function to the delta-rs documentation, may help other users looking for this functionality.

ion-elgreco:
Feel free to open a PR for a static method is_deltatable()

omkar-foss:
take

omkar-foss:
>Feel free to open a PR for a static method is_deltatable()

[...]

PR #2715, opened 2024-07-29
feat(python): add DeltaTable.is_deltatable static method (#2662)

omkar-foss:
closes https://github.com/delta-io/delta-rs/issues/2662.

# Description
This adds a static method `is_deltatable(path, opts)` to the
`DeltaTable` Python class, which returns `True` if able to
locate a delta table at specified `path` and `False` otherwise.

It does so by reusing the Rust internal `is_delta_table_location()`
via the `DeltaTableBuilder`.

Additionally, this also adds documentation of the usage with
examples for the `DeltaTable.is_deltatable()` method.


ion-elgreco:
@omkar-foss thanks for the PR, I do believe we should do a bit differently. Now we are loading the state of the table, which could be potentially very large. We could also simply look if _delta_log exists or not, and whether that is then also not empty.

omkar-foss:
@ion-elgreco Yes I completely agree. I wasn't aware that the `DeltaTable()` constructor loads the entire Delta table eagerly (thanks for letting know!). If that's the case then indeed, checking the `_delta_log` and it's emptiness will be a lighter and a constant-time operation than checking the table state which might grow large and expensive.

As a matter of fact, the `IsDeltaTable` in Delta Lake OSS (Spark version) also checks for `_delta_log` and it's emptiness (see [here](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/io/delta/tables/DeltaTable.scala#L809-L817)), so I suppose it should be perfectly appropriate to apply the same logic here as well. It'll provide consistency of functionality as well for Delta Lake users.

I'll update the PR accordingly, thanks for your quick review 😁👍🏽

ion-elgreco:
@omkar-foss there might be some internal code on the rust side that does something like this already, I am not certain, but worth it to check

omkar-foss:
Hey @ion-elgreco, I've updated the PR as discussed.

I've attempted to reuse the 
Rust side internal `is_delta_table_location()` (code [here](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/logstore/mod.rs#L217-L231)) which reads the `_delta_log` and returns a boolean result, and it's working well.

Kindly check out the PR when you get some time and let me know if any improvement suggestions or changes required. Thanks!

crates/core/src/table/builder.rs
```diff
@@ -330,6 +330,12 @@ impl DeltaTableBuilder {
         }
         Ok(table)
     }
+
+    /// Check if the [`DeltaTable`] exists
+    pub async fn verify_deltatable_existence(self) -> DeltaResult<bool> {
```
ion-elgreco, line 335: Great work! Just minor feedback, I think this method could be on the DeltaTable struct directly

python/src/lib.rs
```diff
@@ -136,6 +136,26 @@ impl RawDeltaTable {
...
+            builder = builder.with_storage_options(storage_options)
+        }
+        let res = rt()
+            .block_on(builder.verify_deltatable_existence())
+            .map_err(PythonError::from);
+        match res {
+            Ok(true) => Ok(true),
+            Ok(false) => Ok(false),
+            Err(err) => Err(err)?,
+        }
```
ion-elgreco, line 156: Can't we leave the match part out? 
```suggestion
        Ok(rt()
            .block_on(builder.verify_deltatable_existence())
            .map_err(PythonError::from)?)
```

crates/core/src/table/builder.rs
omkar-foss: Thanks! Indeed, having it in the `DeltaTable` struct makes more sense. Updated PR.

python/src/lib.rs
omkar-foss: Done, updated PR.

ion-elgreco:
@omkar-foss Nice work! Thanks for the contribution :)

ion-elgreco (approved):

PR #2722, opened 2024-08-02
feat(python, rust): cdc write-support for `overwrite` and `replacewhere` writes

ion-elgreco:
# Description
A normal write with mode "overwrite" or "overwrite" with predicate, will write CDC actions and CDC data now.

This one should be merged first: https://github.com/delta-io/delta-rs/pull/2721

# Related Issue(s)
- related https://github.com/delta-io/delta-rs/issues/2720

PR #2725, opened 2024-08-04
feat: use logical plan in delete, delta planner refactoring

ion-elgreco:
# Description
Changes delete to use logical plans from the get-go. I had to introduce `with_files` on `DeltaTableProvider` in case you want to constraint the scope of files to be read.

Additionally abstracts a DeltaPlanner struct, which was left as TODO, I used this to create a delete MetricObservation node

This one needs to be merged first: https://github.com/delta-io/delta-rs/pull/2722

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -92,6 +92,7 @@ pub mod cdf;
 pub mod expr;
 pub mod logical;
 pub mod physical;
+pub mod planner;
 
 mod find_files;
 mod schema_adapter;
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -344,7 +344,10 @@ pub struct DeltaScanConfigBuilder {
     file_column_name: Option<String>,
     /// Whether to wrap partition values in a dictionary encoding to potentially save space
     wrap_partition_values: Option<bool>,
+    /// Whether to push down filter in end result or just prune the files
     enable_parquet_pushdown: bool,
+    /// Schema to scan table with
+    schema: Option<SchemaRef>,
 }
 
 impl Default for DeltaScanConfigBuilder {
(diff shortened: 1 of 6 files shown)


crates/core/src/delta_datafusion/planner.rs
```diff
@@ -0,0 +1,59 @@
...
+use datafusion_expr::LogicalPlan;
+
+/// Deltaplanner
+pub struct DeltaPlanner<T: ExtensionPlanner> {
+    /// custom extension planner
+    pub extension_planner: T,
+}
+
+#[async_trait]
+impl<T: ExtensionPlanner + std::marker::Send + Sync + 'static + Clone> QueryPlanner
```
hntd187, line 44: nit: can just be Send instead of the full path

hntd187 (approved):
1 little nit but otherwise good

crates/core/src/delta_datafusion/planner.rs
ion-elgreco, line 44:
```suggestion
impl<T: ExtensionPlanner + Send + Sync + 'static + Clone> QueryPlanner
```

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -505,22 +517,17 @@ impl<'a> DeltaScanBuilder<'a> {
...
         };
 
-        let schema = match self.schema {
-            Some(schema) => schema,
-            None => self.snapshot.arrow_schema()?,
-        };
+        let schema = config
+            .schema
+            .clone()
+            .unwrap_or(self.snapshot.arrow_schema()?);
```
alexwilcoxson-rel, line 529: should be `unwrap_or_else` as to not eagerly call `self.snapshot.arrow_schema()` ?

ion-elgreco: Oops, didn't realize unwrap_or always executes the evaluation call, will change to unwrap_or_else

ion-elgreco: Changed it to match arm instead, unwrap_or_else and then error handling is weird unless I just unwrap the result of snapshot.arrow_schema() which could panic

rtyler (approved):

Issue #2572, opened 2024-06-04
Properly handle nested fields when computing stats / stats-schema


roeap:
# Bug

Right now our logic to compute stats, specifically using the `` property only considers root level fields, but does not traverse into fields. Also we need to parse fields as the field names may be escaped and contain special characters .. 

https://github.com/delta-io/delta/blob/4b102d34a2ce881b2a851b4c6cfbf2ab3ab5534f/spark/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L549-L561

**What you expected to happen**:

Properly parse field names when generating stats and stats schema

**More details**:

ion-elgreco:
@roeap fyi, stats parsing seems also not entirely working for checkpoints 
 https://github.com/delta-io/delta-rs/issues/2571

alexwilcoxson-rel:
My team will need to address this to adopt 0.18.x.

My proposal is when looking up stats column in delta schema
- if stats column is formatted `[col.na](http://col.na/)$me` then use schema.field("[col.na](http://col.na/)$me")
- if stats column does not have `s, treat . as delimiter for nested columns and traverse the schema a la: https://github.com/delta-io/delta-rs/pull/2519/files this was reverted essentially when the kernel code was adopted

rtyler:
I was not able to reproduce this with a more contemporary release so I am closing this

PR #2737, opened 2024-08-06
fix: column parsing to include nested columns and enclosing char

gtrawinski:
# Description
Added column parsing to include nested columns and `` enclosing to adopt 0.18.x

# Addressing issue
#2572  
https://github.com/delta-io/delta/blob/4b102d34a2ce881b2a851b4c6cfbf2ab3ab5534f/spark/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L549-L561

# Changes:
## Added:
_extract_enclosed_string_ - is a utility function to remove enclosing characters from a string.
_get_stats_field_ - is a function to retrieve a field from a schema, handling nested field names and removing enclosing backticks.

### DIFF
--- a/crates/core/src/kernel/snapshot/mod.rs
+++ b/crates/core/src/kernel/snapshot/mod.rs
@@ -315,7 +315,7 @@ impl Snapshot {
         let stats_fields = if let Some(stats_cols) = self.table_config().stats_columns() {
             stats_cols
                 .iter()
-                .map(|col| match schema.field(col) {
+                .map(|col| match get_stats_field(schema, col) {
                     Some(field) => match field.data_type() {
                         DataType::Map(_) | DataType::Array(_) | &DataType::BINARY => {
                             Err(DeltaTableError::Generic(format!(
@@ -763,6 +763,45 @@ mod datafusion {
     }
 }
 
+/// Retrieves a specific field from the schema based on the provided field name.
+/// It handles cases where the field name is nested or enclosed in backticks.
+fn get_stats_field<'a>(schema: &'a StructType, stats_field_name: &str) -> Option<&'a StructField> {
+    let dialect = sqlparser::dialect::GenericDialect {};
+    match sqlparser::parser::Parser::new(&dialect).try_with_sql(stats_field_name) {
+        Ok(mut parser) => match parser.parse_multipart_identifier() {
+            Ok(parts) => find_nested_field(schema, &parts),
+            Err(_) => schema.field(stats_field_name),
+        },
+        Err(_) => schema.field(stats_field_name),
+    }
+}
+
+fn find_nested_field<'a>(
+    schema: &'a StructType,
+    parts: &[sqlparser::ast::Ident],
+) -> Option<&'a StructField> {
+    if parts.is_empty() {
+        return None;
+    }
+    let part_name = &parts[0].value;
+    match schema.field(part_name) {
+        Some(field) => {
+            if parts.len() == 1 {
+                Some(field)
+            } else {
+                match field.data_type() {
+                    DataType::Struct(struct_schema) => {
+                        find_nested_field(struct_schema, &parts[1..])
+                    }
+                    // Any part before the end must be a struct
+                    _ => None,
+                }
+            }
+        }
+        None => None,
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


alexwilcoxson-rel:
Is meant to address https://github.com/delta-io/delta-rs/issues/2572

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -956,4 +986,60 @@ mod tests {
...
+        let field = get_stats_field(&schema, "b.nested_struct").unwrap();
+
+        assert_eq!(
+            *field,
+            StructField::new("nested_struct", DataType::BOOLEAN, true)
+        );
+    }
+
+    #[test]
+    fn test_field_with_last_name_nested_backticks() {
```
alexwilcoxson-rel, line 1021: I'm not sure if we need to support nested enclosed field names vs the whole string being enclosed, but I would defer to someone more familiar with the protocol.

gtrawinski: Can you please mention that person? I don't know who should it be.

hntd187: Yeah, I personally have never seen this in the wild, but it doesn't mean it's not there somewhere. The protocol itself doesn't give any hardline opinion about this type of thing, so I suppose it's fine to have it, but understand it's a edge case we'll probably not see often. If it greatly complicates the code I'd consider removing it, but otherwise it's prolly fine.

alexwilcoxson-rel: I have seen the `s mentioned by @roeap on #2572 referencing the [jvm implementation](https://github.com/delta-io/delta/blob/4b102d34a2ce881b2a851b4c6cfbf2ab3ab5534f/spark/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L549-L561).

Before I knew about that and before the kernel types were introduced in 0.18. I added support for nested fields in the dataskipping table config by using the `.` as the delimiter in #2519.

So we could just split on `.` (or use datafusion parser as suggested) like we were in 0.17 and leave #2572 open to consider the ` enclosure for special chars.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -738,6 +738,36 @@ mod datafusion {
...
+    {
+        &string_to_extract[1..string_to_extract.len() - 1]
+    } else {
+        string_to_extract
+    }
+}
+
+/// Retrieves a specific field from the schema based on the provided field name.
+/// It handles cases where the field name is nested or enclosed in backticks.
+fn get_stats_field<'a>(schema: &'a StructType, stats_field_name: &str) -> Option<&'a StructField> {
```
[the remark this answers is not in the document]
alexwilcoxson-rel: this would require the datafusion flag enabled which idk how everyone would feel for this portion of the code, we could use sqlparser crate directly which datafusion is reexporting there.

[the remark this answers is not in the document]
ion-elgreco: Yeah that's also an option, didn't consider that tbh! : )

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -763,6 +763,37 @@ mod datafusion {
...
+                match field.data_type() {
+                    DataType::Struct(struct_schema) => find_nested_field(struct_schema, parts, part_idx + 1),
+                    // Any part before the end must be a struct
+                    _ => None,
+                }
+            }
+        }
+        None => None,
+    }
+}
```
ion-elgreco, line 795: This a little bit simpler?
```suggestion
fn find_nested_field<'a>(schema: &'a StructType, parts: &[sqlparser::ast::Ident]) -> Option<&'a StructField> {
    let part_name = &parts[0].value;
    match schema.field(part_name) {
        Some(field) => {
            if parts.len() == 1 {
                Some(field)
            } else {
                match field.data_type() {
                    DataType::Struct(struct_schema) => find_nested_field(struct_schema, parts[1..]),
                    // Any part before the end must be a struct
                    _ => None,
                }
            }
        }
        None => None,
    }
}
```

alexwilcoxson-rel: nice! yes thank you

ion-elgreco: Cool, if you can push that change, rest of code looks great! Also nice amount of tests 💪

ion-elgreco (approved):
Thankss guys

PR #2747, opened 2024-08-10
feat: introduce CDC generation for merge operations

rtyler:
it's been a slog y'all


crates/core/src/operations/merge/mod.rs
```diff
@@ -1427,11 +1439,62 @@ async fn execute(
...
+        change_data.push(
+            cdc_projection
+                .clone()
+                .filter(
+                    col(SOURCE_COLUMN)
+                        .is_true()
+                        .and(col(TARGET_COLUMN).is_null()),
+                )?
+                .select(write_projection.clone())?
+                .with_column(CDC_COLUMN_NAME, lit("insert"))?,
```
ion-elgreco, line 1472: Why do we create inserts? Isn't it cheaper to let the cdf_reader calculate what was inserted?

ion-elgreco: @rtyler I think we can add one optimization here, if we see it's insert only. So there is only a when_not_matched_insert_all statement, then we could avoid writing CDC as a whole.

ion-elgreco (approved):
Nice work @rtyler 😄

PR #2749, opened 2024-08-11
feat: improved test fixtures

roeap:
# Description

This PR is mainly preparation for the work we need to do on data skipping.
It intorduces a new set of test utilities, but at least also gets rid of one other "test_utils" file. Over the next few PRs I hope to gradually consolidate our usage on "Localized" test utitlitis and use the shared ones.

Initially I wanted to place this in the `test` crate, but somehow cargo kept on complaining about version conflicts, regarding the action structs.

We also add a function to get the partition values schema on snapshot, which I forgot to remove in rebasing some commits. WE will need it going forward so I was hoping it does not add too much complexity to this PR.


crates/core/src/test_utils/factories/mod.rs
ion-elgreco, line 1: Nice!

crates/core/src/test_utils/factories/actions.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn into_action(self) -> Action {
+        Action::Protocol(self)
+    }
+}
+
+impl Metadata {
+    pub fn into_action(self) -> Action {
+        Action::Metadata(self)
+    }
+}
```
ion-elgreco, line 138: We already have the `From<Add/Remove/Protocol/Metadata> for Action` in src/kernel/models/mod.rs, why is this required?

crates/core/src/operations/transaction/state.rs
```diff
@@ -385,6 +412,6 @@ mod tests {
             .unwrap()
             .collect::<Vec<_>>();
         assert_eq!(files.len(), 2);
-        assert!(files.iter().all(|add| add.path.contains("included")));
+        // assert!(files.iter().all(|add| add.path.contains("included")));
```
ion-elgreco, line 415: Can be removed, right? Since the add.paths are now random UUIDs

crates/core/src/test_utils/factories/actions.rs
roeap: it is not, also, now its gone :).

ion-elgreco (approved):
Great work!

Issue #2750, opened 2024-08-11
Write also insert change types in writer CDC


ion-elgreco:
I need to rectify a small mistake of not writing insert change types, will update the writer before we ship out v0.19.

I'll also add some tests on the rust side @roeap, and update the wrong tests in Python to highlight that also inserts are showing up

PR #2751, opened 2024-08-11
fix(python, rust): cdc in writer not creating inserts

ion-elgreco:
# Description
Improved logic
- If it's partition scan, we don't write CDC actions and data. CDF reader will use remove actions in this case now to know what's deleted or not
- When the predicate doesn't hit a partition, we also write CDC data for the inserts

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2750

### DIFF
--- a/crates/core/src/operations/write.rs
+++ b/crates/core/src/operations/write.rs
@@ -36,8 +36,9 @@ use arrow_schema::{ArrowError, DataType, Fields, SchemaRef as ArrowSchemaRef};
 use datafusion::execution::context::{SessionContext, SessionState, TaskContext};
 use datafusion::physical_plan::filter::FilterExec;
 use datafusion::physical_plan::projection::ProjectionExec;
+use datafusion::physical_plan::union::UnionExec;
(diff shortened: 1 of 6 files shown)


crates/core/src/operations/write.rs
```diff
@@ -548,16 +549,19 @@ async fn execute_non_empty_expr(
...
     // For each identified file perform a parquet scan + filter + limit (1) + count.
     // If returned count is not zero then append the file to be rewritten and removed from the log. Otherwise do nothing to the file.
     let mut actions: Vec<Action> = Vec::new();
 
-    let input_schema = snapshot.input_schema()?;
-    let input_dfschema: DFSchema = input_schema.clone().as_ref().clone().try_into()?;
+    // Take the insert plan schema since it might have been schema evolved, if its not
+    // it is simply the table schema
+    let df_schema = insert_plan.schema();
+    let input_dfschema: DFSchema = df_schema.clone().as_ref().clone().try_into()?;
```
hntd187, line 561: seems redundant to double clone here

rtyler (approved):

rtyler (approved):

PR #2765, opened 2024-08-12
feat: fail fast on forked process

### LINKED ISSUE #2744: Regression in Python multiprocessing support
# Environment

**Delta-rs version**: 0.18.2
**Binding**: Python
**Environment**: 
- **Cloud provider**: Impacts local filesystem
- **OS**: Ubuntu 20.04
- **Other**: Python 3.10

***
# Bug

**What happened**: Just upgraded `deltalake` for the first time in a while and we are now getting a deadlock when initialising a `deltalake.DeltaTable` object when using Python `multiprocessing`. Sometimes it just hangs and sometimes the OS kills it.
```
thread 'tokio-runtime-worker' panicked at library/std/src/sys/pal/unix/thread.rs:274:13:
failed to join thread: Resource deadlock avoided (os error 35)
```
[full_log.txt](https://github.com/user-attachments/files/16534884/full_log.txt)

**What you expected to happen**: `DeltaTable` should initialise successfully regardless of `multiprocessing`. 

**How to reproduce it**:
```
from multiprocessing import Pool, Process


def load():
    DeltaTable("crates/test/tests/data/simple_table_with_checkpoint")


def main():
    print("start normal")
    load()
    print("Start multiprossed")
    p = Process(target=load)
    p.start()
    p.join()
    print("DONE")


if __name__ == "__main__":
    main()
```
Note, to reproduce we must use `deltalake.DeltaTable` once then use it again in a child python process using `multiprocessing`. 

**More details**: I have done a `git bisect` and identified that https://github.com/delta-io/delta-rs/pull/2424 is the cause.

Tom-Newton:
# Description
The tokio runtime does not support forked processes, so fail fast if we detect this scenario

- Avoids difficult to debug deadlocks
- Using the deltalake python bindings with `multiprocessing` can easily run into this problem because the default `multiprocessing` start method is `fork` on linux systems

# Related Issue(s)
- closes #2744

# Documentation

### DIFF
--- a/python/src/utils.rs
+++ b/python/src/utils.rs
@@ -15,9 +15,10 @@ pub fn rt() -> &'static Runtime {
         Some(pid) if pid == &std::process::id() => {} // Reuse the static runtime.
         Some(pid) => {
             panic!(
-                "Forked process detected - current PID is {} but the tokio runtime was by {}. The tokio runtime 
-                does not support forked processes https://github.com/tokio-rs/tokio/issues/4301. If you are seeing this 
-                message while using Python multithreading make sure to use the `spawn` or `forkserver` mode.", 
+                "Forked process detected - current PID is {} but the tokio runtime was created by {}. The tokio \
+                runtime does not support forked processes https://github.com/tokio-rs/tokio/issues/4301. If you are \
+                seeing this message while using Python multithreading make sure to use the `spawn` or `forkserver` \
+                mode.", 
                 pid, std::process::id()
             );
         }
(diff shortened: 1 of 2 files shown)


wjones127 (approved):

westonpace (approved):
I like this idea!

westonpace:
(ignore my review...I thought I was looking at a different repo :facepalm:

ion-elgreco:
@westonpace :P

Issue #2771, opened 2024-08-14
Leverage `partitionValues_parsed` from arrow state to report partition values


roeap:
# Description

Currently we parse partition values ad-hoc from raw string data in teh various state / snapshot APIs. Since we now have `partitionValues_parsed` included in the arrow state, we should directly read from that data to avoid additional work.

**Use Case**

**Related Issue(s)**

omkar-foss:
Please ignore the issue mention [above](https://github.com/delta-io/delta-rs/issues/2771#ref-issue-2398288045), it was a typo while trying something out. Sorry!

ion-elgreco:
@roeap are you still looking into this?

roeap:
@ion-elgreco - sort of, the main avenue to a more optimized treatment would now IMO be via kernel. Hoping to file a larger issue soon to align on the best approach for adoption.

ion-elgreco:
@roeap alright, looking forward to that :)

roeap:
this has been addressed in #3801 (and likely also before). Closing it for now.

PR #2772, opened 2024-08-14
feat: more economic data skipping with datafusion

roeap:
# Description

This PR adresses some inefficiencies when creating stats for file pruning using the `PruningPredicate`. Rather then generating stats from file actions which we generate ad-hoc from teh underlying arrow data, we now read the stats directly from the raw file data - i.e. avoiding expensive roundtrips through file actions.

To make this work, we needed to include parsing partition values (adding the `partitionValues_parsed` field to action data) during log replay. As a follow-up we should make more use of the parsed partiton values #2771

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/kernel/snapshot/log_data.rs
+++ b/crates/core/src/kernel/snapshot/log_data.rs
@@ -2,7 +2,9 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use arrow_array::{Array, Int32Array, Int64Array, MapArray, RecordBatch, StringArray, StructArray};
+use arrow_array::{
+    Array, Int32Array, Int64Array, MapArray, RecordBatch, StringArray, StructArray, UInt64Array,
+};
 use chrono::{DateTime, Utc};
 use delta_kernel::expressions::Scalar;
 use indexmap::IndexMap;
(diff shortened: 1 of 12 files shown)


crates/core/src/kernel/snapshot/log_data.rs
```diff
@@ -693,6 +708,122 @@ mod datafusion {
...
+                let result = result
+                    .as_any()
+                    .downcast_ref::<ArrowEngineData>()
+                    .ok_or(DeltaTableError::generic(
+                        "failed to downcast evaluator result to ArrowEngineData.",
+                    ))
+                    .ok()?;
+                results.push(result.record_batch().clone());
+            }
+            let batch = concat_batches(results[0].schema_ref(), &results).ok()?;
```
ion-elgreco, line 741: Isn't it cheaper to concat first and then run it through the Evaluator?

roeap: Probably, at the very least that would give more opportunity for parallelism. in fact we should already concept then batches on `EagerSnapshot`. However  the schemas of these batches are not yet normalized, so we cannot concatenate them yet.

For this we need to do either some internal casting/filtering in the log replay, or even better do column selection when reading the checkpoints ...

rtyler (dismissed):
manually tested, some rustdocs can come later :smile:

rtyler (approved):
:upside_down_face: i'm sure it's fine :upside_down_face:

PR #2773, opened 2024-08-14
chore: fix a bunch of clippy lints and re-enable tests

rtyler:
Many of these tests were did not get correctly re-enabled after the subcrate changes


roeap (approved):
❤️

Issue #2656, opened 2024-07-09
Trailing slash on AWS_ENDPOINT raises S3 Error


dylanspag-lmco:
# Environment

**Delta-rs version**: 0.18.0

**Binding**: Python

**Environment**:
- **Cloud provider**: Local/On-Prem lakeFS with Minio
- **OS**: Ubuntu 24.04 LTS
- **Other**: Python 3.10

***
# Bug

**What happened**:

I added a trailing slash to the `AWS_ENDPOINT` entry of the `storage_options` dictionary that gets passed to `deltalake.write_deltalake`. This raised the following error and prevented any data from being written to the object store.

```python
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Cell In[14], line 1
----> 1 deltalake.write_deltalake(table_or_uri=f"s3a://{repo_name}/main/userdata/",
      2                           data = subset,
      3                           mode='overwrite',
      4                           storage_options=storage_options)

File /opt/conda/lib/python3.10/site-packages/deltalake/writer.py:258, in write_deltalake(table_or_uri, data, schema, partition_by, mode, file_options, max_partitions, max_open_files, max_rows_per_file, min_rows_per_group, max_rows_per_group, name, description, configuration, schema_mode, storage_options, partition_filters, predicate, large_dtypes, engine, writer_properties, custom_metadata)
    171 def write_deltalake(
    172     table_or_uri: Union[str, Path, DeltaTable],
    173     data: Union[
   (...)
    201     custom_metadata: Optional[Dict[str, str]] = None,
    202 ) -> None:
    203     """Write to a Delta Lake table
    204
    205     If the table does not already exist, it will be created.
   (...)
    256         custom_metadata: Custom metadata to add to the commitInfo.
    257     """
--> 258     table, table_uri = try_get_table_and_table_uri(table_or_uri, storage_options)
    259     if table is not None:
    260         storage_options = table._storage_options or {}

File /opt/conda/lib/python3.10/site-packages/deltalake/writer.py:673, in try_get_table_and_table_uri(table_or_uri, storage_options)
    670     raise ValueError("table_or_uri must be a str, Path or DeltaTable")
    672 if isinstance(table_or_uri, (str, Path)):
--> 673     table = try_get_deltatable(table_or_uri, storage_options)
    674     table_uri = str(table_or_uri)
    675 else:

[...]

PR #2770, opened 2024-08-14
fix: trim trailing slash in url storage options (#2656)

omkar-foss:
# Description
This trims trailing slash if present in a storage option's
value if it's key ends with `_URL` (e.g. `HOST_URL`) or if the value
itself seems to be a url (i.e. starts with `http://` or `https://`).

This also adds supporting test for this fix.

# Related Issue(s)

- closes #2656

# Documentation

N/A

PR #2775, opened 2024-08-15
fix: trim trailing slash in url storage options (#2656)

omkar-foss:
# Description
This trims trailing slash if present in a storage option's value
if it's key ends with `_URL` (e.g. `HOST_URL`) or if the value
itself seems to be a url (i.e. starts with `http://` or `https://`).

This also adds supporting test for this fix.

# Related Issue(s)

- closes https://github.com/delta-io/delta-rs/issues/2656

# Documentation

N/A

### DIFF
--- a/crates/core/src/table/builder.rs
+++ b/crates/core/src/table/builder.rs
@@ -261,7 +261,22 @@ impl DeltaTableBuilder {
     /// - [S3 options](https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html#variants)
     /// - [Google options](https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html#variants)
     pub fn with_storage_options(mut self, storage_options: HashMap<String, String>) -> Self {
-        self.storage_options = Some(storage_options);
+        self.storage_options = Some(
+            storage_options
+                .clone()
+                .into_iter()
+                .map(|(k, v)| {
+                    let needs_trim = v.starts_with("http://")
+                        || v.starts_with("https://")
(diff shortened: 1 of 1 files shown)


omkar-foss:
Tried closing PR #2770 and re-opening the same here in this PR, but the workflows are still stuck. It'll be great if someone can help resolve this, thanks.

ion-elgreco:
> Tried closing PR #2770 and re-opening the same here in this PR, but the workflows are still stuck at `Expected — Waiting for status to be reported`. It'll be great if someone can help resolve this, thanks.

Workflows require approval each time

omkar-foss:
> Workflows require approval each time

Ohhh got it, thanks! 👍🏽

crates/core/src/table/builder.rs
```diff
@@ -261,7 +261,22 @@ impl DeltaTableBuilder {
...
     pub fn with_storage_options(mut self, storage_options: HashMap<String, String>) -> Self {
-        self.storage_options = Some(storage_options);
+        self.storage_options = Some(
+            storage_options
+                .clone()
+                .into_iter()
+                .map(|(k, v)| {
+                    let needs_trim = v.starts_with("http://")
+                        || v.starts_with("https://")
+                        || k.to_lowercase().ends_with("_url");
```
roeap, line 271: what are we checking here? i.e. what does a `_url` suffix signify? 

I'm probably just missing something obvious 😆.

roeap: ah - from the tests i gather this is a convention we are introducing?

Should we document this behavior in a docstring somewhere?

roeap (approved):
Thanks for taking care of this!

Issue #2595, opened 2024-06-13
`to_pyarrow_table()` on a table in S3 kept getting "Generic S3 error: error decoding response body"


k-ye:
# Environment

**Delta-rs version**: `deltalake==0.18.1`

**Binding**:  Python

**Environment**: 
- **Cloud provider**: AWS S3
- **OS**: Linux
- **Other**:
  - Python 3.10.4
  - `pyarrow==16.1.0`
  - `pyarrow-hotfix==0.6`

***
# Bug

**What happened**:

Trying to do a simple table loading from S3, but kept getting this `OSError: Generic S3 error: error decoding response body`

```py
table = DeltaTable(table_uri, storage_options=storage_options)
print(f"version: {table.version()}")
print(f"schema: {table.schema()}")
print(table.files())

ts = time.time()
df = table.to_pyarrow_table()
```

```
version: 0
schema: Schema([Field(id, PrimitiveType("string"), nullable=True), Field(path, PrimitiveType("string"), nullable=True)])
['0-e03dac34-16a0-4b6e-82c8-fd1098d1bf45-0.parquet']
Traceback (most recent call last):
  File "test.py", line 32, in <module>
    df = table.to_pyarrow_table()
  File "***/lib/python3.10/site-packages/deltalake/table.py", line 1161, in to_pyarrow_table
    return self.to_pyarrow_dataset(
  File "pyarrow/_dataset.pyx", line 562, in pyarrow._dataset.Dataset.to_table
  File "pyarrow/_dataset.pyx", line 3804, in pyarrow._dataset.Scanner.to_table
  File "pyarrow/error.pxi", line 154, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 88, in pyarrow.lib.check_status
OSError: Generic S3 error: error decoding response body
```

Stack shows that this is actually in `pyarrow`. Not sure if it possible to tweak `pyarrow`'s behavior with S3 from `deltalake`.

**What you expected to happen**:

I can get the pyarrow table.

**How to reproduce it**:

**More details**:

I have verified the integrity of this table with these methods:

1. Cloning the table locally, then load from there. `to_pyarrow_table()` runs fine.
2. Reading the S3 table with `duckdb` (and its `delta` extension). Worked fine, too.

k-ye:
Seems related to https://github.com/delta-io/delta-rs/issues/2301 and https://github.com/delta-io/delta-rs/issues/2592

[...]

Issue #2592, opened 2024-06-12
AsyncChunkReader::get_bytes error: Generic MicrosoftAzure error: error decoding response body


thomasfrederikhoeck:
# Environment

**Delta-rs version**: 0.18.1

**Binding**:  Python

**Environment**:
- **Cloud provider**: Azure
- **OS**: WIndows
- **Other**:

***
# Bug

**What happened**:
After 0.18.1 was released it fixes the inital issue with #2301 for me but instead I started hitting this. The Z-order operations start and I can see that there is usage of network, CPU and memory but after 30 secs-ish I'm hit with the following. The Rust logs doens't show anything strange:
```
metrics = self.table._table.z_order_optimize(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_internal.DeltaError: Failed to parse parquet: Parquet error: Z-order failed while scanning data: ParquetError(General("AsyncChunkReader::get_bytes error: Generic MicrosoftAzure error: error decoding response body"))
```
**What you expected to happen**:
That the Z-order completes.

**How to reproduce it**:
```
import os
os.environ["RUST_LOG"]="debug"

from deltalake import DeltaTable
 
blob_path = "az://<redacted path>"
storage_options = {"AZURE_STORAGE_ACCOUNT_NAME": "<redacted sa>", "AZURE_CONTAINER_NAME":'<redacted container>', 'use_azure_cli': 'true'}

dt = DeltaTable(blob_path, storage_options=storage_options)
dt.optimize.z_order(["StatusDateTime"])
```
**More details**:

abhiaagarwal:
I'm reasonably confident the error is orignating from here, based on my read of various error messages:

https://github.com/delta-io/delta-rs/blob/f0416921a3814a33ea1b3796a2a1468f8c76ca3d/crates/core/src/operations/optimize.rs#L500-L511

Since it's run in a blocking context in the python side, I'm wondering if that's causing any weirdness (it shouldn't).

thomasfrederikhoeck:
@abhiaagarwal I wish I could assist but my Rust knowledge is very limited. But let me know if I need to test something.

Josh-Hiz:
This same issue is happening occasionally when also reading from a deltatable:

```
  File "pyarrow\\_dataset.pyx", line 562, in pyarrow._dataset.Dataset.to_table
  File "pyarrow\\_dataset.pyx", line 3804, in pyarrow._dataset.Scanner.to_table
[...]

PR #2789, opened 2024-08-17
feat: configurable IO runtime

ion-elgreco:
# Description
A configured Tokio handle can be passed into the TableBuilder nested in the RuntTime enum, conveniently added a default which simply creates a new runtime and uses that handle.

# Related issues:
- https://github.com/delta-io/delta-rs/issues/2595, https://github.com/delta-io/delta-rs/issues/2592

### DIFF
--- a/crates/core/src/storage/mod.rs
+++ b/crates/core/src/storage/mod.rs
@@ -14,7 +14,7 @@ use object_store::memory::InMemory;
 use object_store::prefix::PrefixStore;
 use object_store::{GetOptions, PutOptions, PutPayload, PutResult};
 use serde::{Deserialize, Serialize};
-use tokio::runtime::{Builder as RuntimeBuilder, Runtime};
+use tokio::runtime::{Builder as RuntimeBuilder, Handle, Runtime};
 use url::Url;
 
 use bytes::Bytes;
@@ -39,7 +39,7 @@ lazy_static! {
 }
 
 /// Creates static IO Runtime with optional configuration
-fn io_rt(config: Option<&RuntimeConfig>) -> &'static Runtime {
+fn io_rt(config: Option<&RuntimeConfig>) -> &Runtime {
     static IO_RT: OnceLock<Runtime> = OnceLock::new();
     IO_RT.get_or_init(|| {
         let rt = match config {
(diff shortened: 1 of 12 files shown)


crates/core/src/storage/mod.rs
```diff
@@ -178,6 +397,21 @@ pub fn limit_store_handler<T: ObjectStore>(store: T, options: &StorageOptions) -
     }
 }
 
+/// Simple function to wrap the given [ObjectStore] in a [DeltaIOStorageBackend] if USE_EXPERIMENTAL_IO_RUNTIME is enabled
+pub fn deltars_store_handler(store: ObjectStoreRef, options: &StorageOptions) -> ObjectStoreRef {
+    let use_exerimental_rt = options
+        .0
+        .get(storage_constants::EXPERIMENTAL_RT)
+        .and_then(|v| v.parse::<bool>().ok())
+        .unwrap_or_default();
```
roeap, line 406: One thing i wondered if this should really be done via the storage options, or if this should be configured via the `DeltaTableOptions`. What triggered this most, is thinking about stuff like coordinated commits or external catalogs, where other things may occur, that are io bound ...

roeap (commented):
@ion-elgreco, looking great.

The only comment is around how to enable this... As we are looking into some performance optimizations, the `delta_rs_storage_hadler`, may actually also be the place where we may want to introduce some simple caching for the commit files (specifically the json commits) in order to keep the log replay logic as simple as possible.

The above is just another argument, why i think we should consider passing the config via the `DeltaTableOptions`...

this will also allow us to pass in configuration for the runtime itself. For now it's ok, but if we GA this feature, we probably want to expose some sort of way to pass configuration to the runtime.

ion-elgreco:
@roeap do you mean the DeltaTableLoadOptions which is used in the table builder?

roeap:
i thought maybe this one: https://github.com/delta-io/delta-rs/blob/eea53f46933e6b49f986190da32fb0fbec39ff40/crates/core/src/table/builder.rs#L53-L82

ion-elgreco:
@roeap Hmm does it need to be accessible later though? 

Below could work also I believe, we can still save it in the DeltaTableConfig but wondering where else it will be used
```rust
#[derive(Debug)]
pub struct DeltaTableLoadOptions {
  ....
    pub log_batch_size: usize,
    pub io_runtime_config: Option<IORuntimeConfig>, // If passed enable experimental IO runtime with custom config
}   
```

Then DeltaTableBuilder has `with_io_runtime_config`, during build we can pass the IORuntimeConfig into  `build_storage` which can wrap the objectstore into the delta_rs_handler with the runtime specification

roeap:
Yeah, this triggers some memories - at some point i felt there is a bunch of config distributed in various ways throughout our codebase, and we need to consolidate into a single way of handling config. The duplication of `log_buffer_size` we already see here is a good indication if this... I guess my main question would be - if you were to keep only on of these, which one would it be 😆, and then pick that one.

Naming wise I do find that "..LoadOptions" is the worse name, as it implies that this only relates to the initial table load?

roeap:
@ion-elgreco - after a quick glance at the relevant code, it feels maybe the some options should be flattened into the `DeltaTableBuilder`, and the config be added as a field on the builder instead of the load options?

ion-elgreco:
> @ion-elgreco - after a quick glance at the relevant code, it feels maybe the some options should be flattened into the `DeltaTableBuilder`, and the config be added as a field on the builder instead of the load options?

Yeah that makes sense! Now it's double 😆

I'll do another round on this tomorrow then to refactor the table builder a bit and add the IO runtime config on the table config.

Will have to take a look at what is normally configurable on a Tokio runtime :)

crates/core/src/storage/mod.rs
```diff
@@ -30,6 +38,294 @@ lazy_static! {
...
+    thread_name: Option<String>,
+    enable_io: bool,
+    enable_time: bool,
+}
+
+/// Provide custom Tokio RT or a runtime config
+#[derive(Debug, Clone)]
+pub enum IORuntime {
+    /// Tokio RT handle
+    RT(&'static Runtime),
```
alexwilcoxson-rel, line 91: instead of an instance of borrowed Runtime you could accept an owned Handle to the runtime which one could get with 

`runtime.handle().clone()`

ion-elgreco: Actually that's what I did first but it didn't work. All tasks got canceled

ion-elgreco: Donee now

roeap (approved):
I think this turned out quote nice, great work! 👍

PR #2806, opened 2024-08-20
perf: spawn sync parquet write on blocking runtime

alexwilcoxson-rel:
# Description
In our service we have tried to dance around the fact the underlying Delta/PartitionWriter runs the synchronous ArrowWriter write within an async method, blocking a runtime thread. This PR allows you to opt into to supplying a runtime to the DeltaWriter to spawn blocking tasks on.

I intend to clean up some of the code by implementing some of the methods on WriterState enum and write some tests, but want to get initial feedback.

Using this in our service with its own runtime has simplified our code and kept the main runtime free for io and incoming requests.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/operations/writer.rs
+++ b/crates/core/src/operations/writer.rs
@@ -130,6 +130,8 @@ pub struct DeltaWriter {
     config: WriterConfig,
     /// partition writers for individual partitions
     partition_writers: HashMap<Path, PartitionWriter>,
+    /// Optional runtime for Parquet blocking IO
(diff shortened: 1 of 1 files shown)


ion-elgreco:
Do you still need this now that you can wrap all IO tasks in their own runtime?

rtyler:
In case you had missed it, @ion-elgreco is referring to the work in #2789

alexwilcoxson-rel:
resolved using DeltaIOStorageBackend and a separate runtime for the write and close calls

Issue #2802, opened 2024-08-20
Large Memory Spike on Merge


rob-harrison:
# Environment

**Delta-rs version**: 0.18.2

**Binding**: Python

**Environment**:
- **Cloud provider**: AWS
- **OS**: Debian GNU/Linux 12 (bookworm)
- **Other**: python 3.11

***
# Bug

**What happened**:

1. delta table with ±8 million rows, partitioned with largest partition 800k rows, size ±550GiB according to describe detail.
2. table optimized with .compact() and checkpoint every 100 commits.
3. merge operation with 1k rows and a single partition in predicate: part_col IN ('123456789')
4. memory spikes from ±200MiB to ±14GiB on .execute() - see memory_profiler output:

```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
35    200.7 MiB    200.7 MiB           1       @profile
36                                             def _perform_merge(self, detections: List[Detection], class_blacklist: List[str] = None):
37                                                 # upsert detections to datastore
38    202.2 MiB      1.5 MiB           1           batch = build_pyarrow_batch(detections=detections, class_blacklist=class_blacklist)
39    202.2 MiB      0.0 MiB           1           logging.info(f'upserting batch of [{batch.num_rows}] detections to delta table')
40    202.2 MiB      0.0 MiB           1           if batch.num_rows > 0:
41    209.2 MiB      7.0 MiB           2               dt = get_delta_table(table_path=self.delta_table_path,
42    202.2 MiB      0.0 MiB           1                                    dynamo_table_name=self.dynamo_table_name)
43    209.2 MiB      0.0 MiB           1               h3_indices_column = batch.column('h3_indices')
44    209.2 MiB      0.0 MiB           1               partition_key_values = self._get_partition_keys(h3_indices=h3_indices_column.to_pylist())
45    209.2 MiB      0.0 MiB           1               merge_predicate = f'{TARGET_ALIAS}.{HD_MAP_FRAME_DETECTION_PARTITION_BY_COLUMN} in ({partition_key_values}) and {TARGET_ALIAS}.{HD_MAP_FRAME_DETECTION_KEY_COLUMN} = {SOURCE_ALIAS}.{HD_MAP_FRAME_DETECTION_KEY_COLUMN}'
[...]

Issue #2726, opened 2024-08-04
`IN (...)` clauses appear to be ignored in merge commands with S3 - extra partitions scanned


MuneebBaderoen:
# Environment

**Delta-rs version**: 0.18.2

**Binding**: Python

**Environment**:
- **Cloud provider**: AWS - S3
- **OS**: macOS Sonoma 14.5
- **Other**: Python 3.12

***
# Bug

**What happened**:

It appears that when performing a `merge` operation, specifying `partition_column IN ('value_1')` as the initial predicate, whether it's a single value, or multiple values - the `IN` clause is ignored as if it doesn't exist, and no errors are raised. 

A separate, but potentially related issue, is that when performing a `merge` operation, when specifying `partition_column = 'value_1'`, sometimes I see additional partitions being queried from S3. The exact additional partition retrieved is non-deterministic, but there's always an extra one in the example setup I have. I set up the example to debug the IN clause behaviour described above, and spotted this along the way. Performing the same operation a second time queries only the exact partitions specified by the clauses in the predicate.

**What you expected to happen**:

- When querying with a predicate that uses an IN clause, only the files for the partitions matching all clauses are requested from S3
- When querying with a predicate that uses an = clause, only the files for the partitions matching all clauses are requested from S3

**How to reproduce it**:
MRE can be found here: https://github.com/MuneebBaderoen/delta-rs-in-predicate-mre

**More details**:
Slack thread: https://delta-users.slack.com/archives/C013LCAEB98/p1722382232123799

The impact of this behaviour is that attempting to upsert data across two partitions (for example at the boundary of days, or the boundary of months) dramatically increases the volume of data downloaded from S3. 

On the boundary between days, the implementation I have would upsert data for `partition_day IN ('05', '06')` - but this clause would be ignored, resulting in all data for all partitions in the month being downloaded. This is visible in the localstack logs in the MRE provided.

On the boundary between months, the implementation I have would attempt to upsert data for `partition_month IN ('07', '08') partition_day IN ('31', '01')` - but both `IN` clauses would be ignored, resulting in all data for all partitions in the year being downloaded. This is visible in the localstack logs in the MRE provided.

PR #2807, opened 2024-08-21
fix(python, rust): allow `in` pushdowns in early_filter

ion-elgreco:
# Description
is_in filters can be used now to do partition pruning, I also moved all the filter logic into it's own module.

# Related issues
- closes https://github.com/delta-io/delta-rs/issues/2802
- closes https://github.com/delta-io/delta-rs/issues/2726

### DIFF
--- a/crates/core/src/operations/merge/filter.rs
+++ b/crates/core/src/operations/merge/filter.rs
@@ -252,13 +252,17 @@ pub(crate) fn generalize_filter(
             }
         }
         Expr::InList(in_list) => {
-            let compare_expr = generalize_filter(
+            let compare_expr = match generalize_filter(
                 *in_list.expr,
                 partition_columns,
                 source_name,
                 target_name,
                 placeholders,
-            );
+            ) {
+                Some(expr) => expr,
+                None => return None, // Return early
+            };
+
             let mut list_expr = Vec::new();
             for item in in_list.list.into_iter() {
                 match item {
@@ -277,14 +281,15 @@ pub(crate) fn generalize_filter(
                     }
                 }
             }
-            match (compare_expr, list_expr) {
-                (Some(compare), list) if !list.is_empty() => Expr::InList(InList {
-                    expr: compare.into(),
-                    list: list,
+            if !list_expr.is_empty() {
+                Expr::InList(InList {
+                    expr: compare_expr.into(),
+                    list: list_expr,
                     negated: in_list.negated,
                 })
-                .into(),
-                (_, _) => None,
+                .into()
+            } else {
+                None
             }
         }
         other => match references_table(&other, source_name) {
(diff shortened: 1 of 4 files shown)


rtyler:
@ion-elgreco I pulled this into my CI infrastructure, and also have some bigger tables at `$WORKPLACE` that @roeap and I were doing some performance analysis with. I will take a look and report back with a review.

crates/core/src/operations/merge/filter.rs
```diff
@@ -0,0 +1,382 @@
...
+        }
+        Expr::InList(in_list) => {
+            let compare_expr = generalize_filter(
+                *in_list.expr,
+                partition_columns,
+                source_name,
+                target_name,
+                placeholders,
+            );
+            let mut list_expr = Vec::new();
```
JonasDev1, line 262: Can we directly return None if compare_expr is None, because otherwise we could add unnecessary placeholders in the for loop?

JonasDev1:
In the mod.rs are also tests related to the filters, maybe you can move them also to the filter.rs

ion-elgreco:
> In the mod.rs are also tests related to the filters, maybe you can move them also to the filter.rs
Ah yes, missed those :)

JonasDev1:
A test with a condition like `target.col in (source.col1, source.col2)` would be interesting

crates/core/src/operations/merge/filter.rs
ion-elgreco: Good one,  pushed a small  refactor now

ion-elgreco:
I've put it in draft until I add some more tests in Rust

JonasDev1:
> > A test with a condition like `target.col in (source.col1, source.col2)` would be interesting
> 
> Hmm I will add some more tests, by the way I wonder if we want placeholder exprs if a user already explicitly passed a predicate for that partition column.
> 
> If I do `s.partition and t.partition and t.partition in (3,4)`, I will get multiple placeholders saying `t.partition=3 or t.partition=4 and t.partition in (3,4)`, result will be the same but I wonder if this would make it less efficient in terms of query runtime

The placeholder expr could be more restrictive than the explicit expr. I think performance should not be affected

omkar-foss:
> A test with a condition like `target.col in (source.col1, source.col2)` would be interesting

Hi! I've added a Rust test that covers this condition in [here](https://github.com/ion-elgreco/delta-rs/pull/2/files#diff-12f59fe3c4440b7ae4ee1a5ac810b42c1d7357c246aae7b5770e840e52d3ec52R1997-R2013).

rtyler:
I have done some manual testing with this pull request and I can say that at least it did not adversely impact performance :smile: . I will rebase this branch, merge it, and then we can go from there

rtyler (approved):

Issue #2808, opened 2024-08-21
DeltaTableBuilder flags ignored


dragoscalinescu:
# Environment

**Delta-rs version**: 0.19.0

**Binding**:  rust

***
# Bug

**What happened**:
Delta-rs rust library stopped using the `without_files` and `without_tombstones` flags, which causes high memory usage for applications which do not need to keep track of the files.

**What you expected to happen**:
Files not to be tracked as per flags.

**How to reproduce it**:
```
    DeltaTableBuilder::from_uri("some_table_uri")
        .without_files()
        .without_tombstones()
        .load()
        .await
        .unwrap();
````

Issue #2796, opened 2024-08-19
Require files in config is not anymore used to skip reading add actions


ion-elgreco:
# Environment
We should instantiate this optimization for append only use cases again

Issue #867, opened 2022-10-08
Change file_uris() to error if without_files was set


wjones127:
# Description

You can instantiate a `DeltaTable` struct without loading the files to save memory. If you call `.file_uris()` on that table, it will return an empty iterator, which isn't quite correct. I fear this might lead to issues later, where we can't differentiate between an empty table and one that was loaded without files. Perhaps this should return an error instead?

**Use Case**

**Related Issue(s)**

Found while looking at #866.

PR #2809, opened 2024-08-21
fix(python, rust): use require files

ion-elgreco:
# Description

## Follow up PR
Use tombstones flag as well

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2796
- closes https://github.com/delta-io/delta-rs/issues/2808
- closes https://github.com/delta-io/delta-rs/issues/867

# Documentation


ion-elgreco:
@rtyler this essentially reintroduces the option to skip reading all add actions and keeping that in memory for append only workloads.

It also puts the necessary guard rails in place so people don't do operations where we need to know the add actions :p

PR #2811, opened 2024-08-21
fix: use table config target file size

ion-elgreco:
# Description
If the target_file_size is None but we have a snapshot, then we should take it from the table config


crates/core/src/operations/mod.rs
```diff
@@ -283,6 +283,22 @@ pub fn get_num_idx_cols_and_stats_columns(
     )
 }
 
+/// Get the target_file_size from the table configuration in the sates
+/// If table_config does not exist (only can occur in the first write action) it takes
+/// the configuration that was passed to the writerBuilder.
+pub fn get_target_file_size(
```
rtyler, line 289: I don't think this should become part of our public API, so I'm going to slap a crate level visibility on it.

ion-elgreco: Woops, my bad, should have been internal indeed

rtyler (approved):

rtyler (approved):

Issue #2296, opened 2024-03-17
Option To Use ObjectStore Conditional Put


tustvold:
# Description

Currently delta-rs creates a new transaction log entry, and then moves it into place with copy_if_not_exists.

https://github.com/apache/arrow-rs/issues/4879 added support in ObjectStore for conditional PUT, including PutIfNotExists.

**Use Case**

This would have a couple of potentially compelling advantages over using CopyIfNotExists:

* Two fewer request to object storage (Copy + Delete)
* Natively supported by many stores, including [Minio](https://github.com/minio/minio/blob/c201d8bda903bf951ca4734d502b3f365fa8ba04/cmd/object-handlers-common.go#L139), which doesn't support CopyIfNotExists
* No risk of leaving unreferenced files around from failed transactions

**Related Issue(s)**

PR #2813, opened 2024-08-22
perf: conditional put for default log store (e.g. azure, gcs, minio, cloudflare)

ion-elgreco:
# Description
When it's a default log store, we pass bytes around, for other log stores we keep the default tmp_commit logic in place (for stores like `S3DynamoDBLogStore` or the new `S3LogStore`)

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2296


crates/core/src/logstore/default_logstore.rs
```diff
@@ -1,12 +1,25 @@
...
-use crate::{operations::transaction::TransactionError, storage::ObjectStoreRef, DeltaResult};
+use super::{CommitOrBytes, LogStore, LogStoreConfig};
+use crate::{
+    operations::transaction::TransactionError,
+    storage::{commit_uri_from_version, ObjectStoreRef},
+    DeltaResult,
+};
+
+fn put_options() -> &'static PutOptions {
+    static PUT_OPTS: OnceLock<PutOptions> = OnceLock::new();
```
alexwilcoxson-rel, line 16: why put it in a static lock?

crates/core/src/logstore/default_logstore.rs
```diff
@@ -45,17 +58,45 @@ impl LogStore for DefaultLogStore {
     async fn write_commit_entry(
         &self,
         version: i64,
-        tmp_commit: &Path,
+        commit_or_bytes: CommitOrBytes,
     ) -> Result<(), TransactionError> {
-        super::write_commit_entry(self.storage.as_ref(), version, tmp_commit).await
+        match commit_or_bytes {
+            CommitOrBytes::LogBytes(log_bytes) => {
+                match self
```
alexwilcoxson-rel, line 65: this inner match seems unnecessary, just map the error on put_opts(...).await

ion-elgreco: Ah it was some old code, needed when I also did tmp_commit logic here, will remove it :)

crates/core/src/logstore/default_logstore.rs
ion-elgreco: Was thinking because it needs to be called at every put call, but I guess it doesn't matter since I have to clone anyways

hntd187: I don't have any problems with this code, but it does increase the MRSV, so keep that in mind since OnceLock was added to std in 1.80 I believe

ion-elgreco: @hntd187 we already use OnceLock in a couple other places ^^

rtyler (approved):
This looks reasonable to me, I do have concerns just because we don't have great integration testing here between Rust/Spark (not a problem for this PR).

I think this change should require us to bump to 0.20

ion-elgreco:
@rtyler what are you worried about interaction between Spark and Rust here?

rtyler:
@ion-elgreco we don't have good round-trip tests with Spark writers and Rust writers collaborating (or not) with our log store implementation.

ion-elgreco:
@rtyler I see, but in this case I am not touched the S3DynamoDBLogStore, only stores that use copyIfNotExists or S3 flavours with PutIfAbsent

PR #2816, opened 2024-08-22
feat: public method to get partitions for DeltaTable (#2671)

### LINKED ISSUE #2671: Function behaving similarly to `SHOW PARTITIONS` in the Python API
# Function behaving similarly to `SHOW PARTITIONS` in the Python API

I am wondering if there is something similar to `SHOW PARTITIONS` from the Spark world of interacting with Delta Tables. This is a metadata-only query that returns back all of the partitions live for a specific Delta Table in a tabular format. Functionality such as `DeltaTable.files_by_partitions` is super helpful for querying but not quite the same thing. [get_active_partitions](https://github.com/delta-io/delta-rs/blob/f432c4f8337c2b0d47958645684e5df336c61522/python/src/lib.rs#L933) is almost what I need but (1) it's not exposed in the public API which makes usage a bit clunky (but certainly nothing life ruining) and (2) the struct it returns is not the most ergonomic.

Any openness to a PR that does this?

**No related issues from what I could tell.**

omkar-foss:
# Description
This adds a public method `partitions()` to the `DeltaTable` class to get properly formatted partitions (list of dicts) for the table. Also provides an option to return partitions as a list of tuples, and proxies the partition filters to rust `get_active_partitions()`.

This also adds supporting tests for this feature.

# Related Issue(s)

- closes #2671

# Documentation

Documentation included in method's docstring.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -514,25 +514,19 @@ class DeltaTable:
     def partitions(
         self,
         partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
-        as_tuple_list: bool = False,
-    ) -> List[Dict[str, str]] | List[Tuple[str]]:
+    ) -> List[Dict[str, str]]:
         """
         Returns the partitions as a list of dicts. Example: `[{'month': '1', 'year': '2020', 'day': '1'}, ...]`
 
         Args:
             partition_filters: The partition filters that will be used for getting the matched partitions, defaults to `None` (no filtering).
-            as_tuple_list: If `True`, returns the partitions as a list of tuples. Example: `[(("day", "5"), ("month", "4"), ("year", "2021")), ...]`
         """
 
-        partitions: List[Any] = []
+        partitions: List[Dict[str, str]] = []
         for partition in self._table.get_active_partitions(partition_filters):
             if not partition:
                 continue
-            if as_tuple_list:
-                sorted_partition = sorted(partition, key=lambda x: x[0])
-                partitions.append(tuple(sorted_partition))
-            else:
-                partitions.append({k: v for (k, v) in partition})
+            partitions.append({k: v for (k, v) in partition})
         return partitions
 
     def files(
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -511,6 +511,30 @@ class DeltaTable:
         """
         return self._table.version()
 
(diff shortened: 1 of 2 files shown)


python/deltalake/table.py
```diff
@@ -510,6 +510,30 @@ def version(self) -> int:
         """
         return self._table.version()
 
+    def partitions(
+        self,
+        partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
+        as_tuple_list: bool = False,
+    ) -> List[Dict[str, str]] | List[Tuple[str]]:
```
ion-elgreco, line 517: I would prefer if it's just one output type, list[dict] seems fine to be, people can always construct from that a list(tuple) if they want.

ion-elgreco (dismissed):

python/deltalake/table.py
omkar-foss: Makes sense, updated PR. Thanks.

thomasfrederikhoeck:
How does this work if partition is a date or a datetime? Would be nice with a test for this

omkar-foss:
> How does this work if partition is a date or a datetime? Would be nice with a test for this

Thanks for the suggestion @thomasfrederikhoeck! Have added a test for a date partitioned table in [this commit](https://github.com/delta-io/delta-rs/pull/2816/commits/226c671228e269db9c56ae3d6fd308d3dc364193).

thomasfrederikhoeck:
> > How does this work if partition is a date or a datetime? Would be nice with a test for this
> 
> 
> 
> Thanks for the suggestion @thomasfrederikhoeck! Have added a test for a date partitioned table in [this commit](https://github.com/delta-io/delta-rs/pull/2816/commits/d5863d2ff7f822cc05f73da0b6ba7538acbe5a7b).

Nice! 👍🏻

rtyler (approved):

omkar-foss:
Hi @rtyler thanks for approving this PR. It seems that one of the concurrency tests for Windows is failing [here](https://github.com/delta-io/delta-rs/actions/runs/10532187167/job/29328798180?pr=2816#step:4:2022) with a `PermissionDenied` error.

Please let me know if I shall check this out, or if it's an intermittent issue that'll get resolved on re-running the test. Thanks!

rtyler (approved):

PR #2826, opened 2024-08-25
chore: set max_retries in CommitProperties

helanto:
# Description
Adds a `with_max_retries` in `CommitProperties`. Allow for clients to control the maximum number of times a commit will be retried before failing.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -346,6 +346,15 @@ impl CommitProperties {
         self
     }
 
+    /// Specify maximum number of times to retry the transaction before failing to commit
+    pub fn with_max_retries(
+        mut self,
+        max_retries: usize,
+    ) -> Self {
+        self.max_retries = max_retries;
+        self
+    }
+
     /// Specify if it should create a checkpoint when the commit interval condition is met
     pub fn with_create_checkpoint(mut self, create_checkpoint: bool) -> Self {
         self.create_checkpoint = create_checkpoint;
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -347,10 +347,7 @@ impl CommitProperties {
     }
 
     /// Specify maximum number of times to retry the transaction before failing to commit
-    pub fn with_max_retries(
-        mut self,
-        max_retries: usize,
-    ) -> Self {
+    pub fn with_max_retries(mut self, max_retries: usize) -> Self {
         self.max_retries = max_retries;
         self
     }
--- a/python/deltalake/__init__.py
+++ b/python/deltalake/__init__.py
@@ -10,6 +10,7 @@ from .table import (
 from .table import (
     ColumnProperties as ColumnProperties,
 )
+from .table import CommitProperties as CommitProperties
 from .table import DeltaTable as DeltaTable
 from .table import Metadata as Metadata
 from .table import PostCommitHookProperties as PostCommitHookProperties
--- a/python/deltalake/_internal.pyi
+++ b/python/deltalake/_internal.pyi
@@ -3,7 +3,12 @@ from typing import Any, Dict, List, Literal, Mapping, Optional, Tuple, Union
 import pyarrow
 import pyarrow.fs as fs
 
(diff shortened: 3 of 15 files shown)


ion-elgreco:
This needs to be exposed for python users as well

rtyler (dismissed):

helanto:
> This needs to be exposed for python users as well

Not super familiar with how Python binding works. Made an attempt to expose the `max_commit_retries` in Python API. Let me know what you think.

ion-elgreco:
@helanto Hmm perhaps we can introduce a CommitProperties class, where you have custom_metadata and commit_retries as variables, this would be a breaking change, so we need to warn users that custom_metadata should be passed into CommitProperties

helanto:
> @helanto Hmm perhaps we can introduce a CommitProperties class, where you have custom_metadata and commit_retries as variables, this would be a breaking change, so we need to warn users that custom_metadata should be passed into CommitProperties

I believe I made the change. How do we warn users ?

ion-elgreco:
> > @helanto Hmm perhaps we can introduce a CommitProperties class, where you have custom_metadata and commit_retries as variables, this would be a breaking change, so we need to warn users that custom_metadata should be passed into CommitProperties
> 
> I believe I made the change. How do we warn users @ion-elgreco ?

You can check in other places of the code base, where we show a warn when a parameter is passed that is deprecated

helanto:
> You can check in other places of the code base, where we show a warn when a parameter is passed that is deprecated

This means that we keep both parameters around. Do you have something in mind along the following lines?
```python
def operation(
    ....,
    custom_metadata: Optional[Dict[str, str]] = None,  # We keep this around for compatibility reasons.
    commit_properties: OptionalCommit[Properties] = None,
):
    if custom_metadata is not None:
        warn("Please provide custom_metadata as part of commit_properties").
    ...
```

ion-elgreco:
> > You can check in other places of the code base, where we show a warn when a parameter is passed that is deprecated
> 
> This means that we keep both parameters around. Do you have something in mind along the following lines?
> ```python
> def operation(
>     ....,
>     custom_metadata: Optional[Dict[str, str]] = None,  # We keep this around for compatibility reasons.
>     commit_properties: OptionalCommit[Properties] = None,
> ):
>     if custom_metadata is not None:
>         warn("Please provide custom_metadata as part of commit_properties").
>     ...
> ```

Yess, and then custom_metadata can be dropped in later versions

ion-elgreco:
@helanto do you have time to make that change? Because then we can merge :  )

helanto:
> @helanto do you have time to make that change? Because then we can merge : )

Yes I will work on it today!

helanto:
> > @helanto do you have time to make that change? Because then we can merge : )
> 
> Yes I will work on it today!

I believe I made the changes if you can take a look @ion-elgreco

ion-elgreco:
@helanto just one more thing, commit_metadata in the pyo3 functions can be removed, and in Python if custom metadata is still passed, you can wrap it into CommitPropeti e

helanto:
> @helanto just one more thing, commit_metadata in the pyo3 functions can be removed, and in Python if custom metadata is still passed, you can wrap it into CommitPropeti e

I look at pyo3 function definitions and they don't have any `commit_metadata` argument ? Do you mean `custom_metadata` ? The pyo3 functions look like the following:
```rust
#[pyo3(signature = ...)]
pub fn some_function(
        ...,
        custom_metadata: Option<HashMap<String, String>>,
        max_commit_retries: Option<usize>,
)
```

I can create a pyo3 class that wraps `custom_metadata` and `max_commit_retries`, similar to `PyPostCommitHookProperties` [here](https://github.com/delta-io/delta-rs/blob/main/python/src/lib.rs#L1588). At the moment the `CommitProperties` only exist as a publicly facing Python class, it's not a pyo3 class. @ion-elgreco

ion-elgreco:
Yeah I meant custom_metadata

helanto:
> Yeah I meant custom_metadata. We need a rust struct which has a derive indeed

Now it should be good! Let me know what you think @ion-elgreco

ion-elgreco (approved):
Thankss!!

ion-elgreco (approved):

PR #2841, opened 2024-09-03
fix: stats is optional in add action

jkylling:
# Description
The stats field in add actions is optional, https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file

We currently silently skip reading add actions from snapshots without stats.

# Related Issue(s)

# Documentation


rtyler (dismissed):

rtyler:
@jkylling I am trying to add a test case to validate this scenario and cannot reproduce the issue since my Add actions with no stats are coming through. Lots of our tests have Add actions with no stats too :confused:

jkylling:
@rtyler if we remove stats from the checkpoint schema the checkpoint_writer::simple_checkpoint_simple_checkpoint_test fails because some files are missing https://github.com/delta-io/delta-rs/commit/6da8b6bad3306caf1a29ddad6954691f92ea7059

ion-elgreco:
@jkylling can you add a test for this

rtyler (approved):

Issue #2851, opened 2024-09-06
Merge with LargeList arrow type not working


ericvandever:
# Environment

--------Version info---------
Polars:              1.6.0
Index type:          UInt32
Platform:            macOS-14.5-arm64-arm-64bit
Python:              3.12.3 (main, Jun  6 2024, 07:44:06) [Clang 15.0.0 (clang-1500.3.9.4)]
----Optional dependencies----
deltalake            0.19.2
numpy                2.1.1
pyarrow              17.0.0

***
# Bug

**What happened**:
In version 0.19 large_dtypes flags were deprecated causing my workflow with merging deltatables including list types to break.

**What you expected to happen**:
Continue to be able to merge on tables including list (arrow largelist)  type columns.   Or if that is intentionally unsupported, at least some documentation/migration guide to let me know what I have to change. 

**How to reproduce it**:

works in 0.18.x, fails on 0.19.x

```python
    pl.DataFrame({"a": ["test"], "b": [[1]]}).write_delta(tmp_path)
    pl.DataFrame({"a": ["test1"], "b": [[1]]}).write_delta(
        tmp_path,
        mode="merge",
        delta_merge_options={
            "predicate": "s.a = t.a",
            "source_alias": "s",
            "target_alias": "t",
        },
    ).when_not_matched_insert_all().execute()
```

**More details**:
Using polars implies large_dtypes=True which is included in the underlying call to write_deltalake.  If I bypass polars function and call DeltaTable.create with a schema including only non-large types things work as expected.  Is it intended that the TableMerger does not support large_dtypes?  Broader question (between both polars and delta): what is the implications of forcing non-large types and interacting with polars?  I know the polars maintainers have strong opinions that they should be using the large types, but for my workflows, I don't need anywhere near the # of elements supported by the standard list, so largelist is just using extra space.

**test output**
```
============================== 1 failed in 0.12s ===============================
FAILED                        [100%]
tests/polars/deltalake/test_persistence.py:726 (test_polar_deltatable)
tmp_path = PosixPath('/private/var/folders/xq/fhffzxx95nvgzgl911sjsthh0000gn/T/pytest-of-ericv/pytest-59/test_polar_deltatable0')

    @pytest.mark.write_disk
    def test_polar_deltatable(tmp_path: Path) -> None:
        pl.DataFrame({"a": ["test"], "b": [[1]]}).write_delta(tmp_path)
[...]

PR #2853, opened 2024-09-07
fix: re-enable optional old casting behavior in merge

ion-elgreco:
# Description

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/2851

@ericvandever, this doesn't fix the coercion but allows you to still use large_dtypes=False to downcast so that you can still run most of your queries

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -1000,13 +1000,18 @@ class DeltaTable:
                 commit_properties, custom_metadata
             )
 
-        if large_dtypes:
+        if large_dtypes is not None:
             warnings.warn(
                 "large_dtypes is deprecated",
                 category=DeprecationWarning,
                 stacklevel=2,
             )
-        conversion_mode = ArrowSchemaConversionMode.PASSTHROUGH
+            if large_dtypes:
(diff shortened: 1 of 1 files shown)


rtyler:
@ion-elgreco I'm sure you're getting busier as the autumn is starting, so i'll try to be quick:

* Weren't we deprecating large_dtypes?
* Since there's new logic here, would you have the time to write a test?
* Does this work come back out at some point? Or does this branch need to exist as long as large_dtypes doe?

ion-elgreco:
> @ion-elgreco I'm sure you're getting busier as the autumn is starting, so i'll try to be quick:
> 
> * Weren't we deprecating large_dtypes?
> * Since there's new logic here, would you have the time to write a test?
> * Does this work come back out at some point? Or does this branch need to exist as long as large_dtypes doe?

We are, but some type coercion in Datafusion isn't working properly yet, particularly list <=> large list. So I only re add it temporarily. 

The test will be short lived I guess, and the transformation code is well tested so should be fine.

I made an upstream issue for this type coercion problem 

Last question I don't get? :p

rtyler (dismissed):

rtyler (approved):

Issue #2879, opened 2024-09-04
Assume role support has been broken since 2022 :rofl: 


oleh-motrunych:
kafka-delta-ingest version 0.2.0 fails to assume the specified IAM role for cross-account writes to S3, resulting in an Access Denied (403) error, the previous version of the application worked correctly under the same configuration.

error log
```
2024-09-04T18:23:33 [ERROR] - my.app: Ingest service exited with error Writer { source: Storage { source: Generic { store: "S3", source: Client { status: 403, body: Some("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>CAE9GFWZPREMDXXX</RequestId><HostId>fsdfsdfsdfAaXXXXXqUenLk4JxzSkdGI/msdfsfsfdHyg=</HostId></Error>") } } } }
```
Current setup:
	•	AWS Account A: kafka-delta-ingest running as an ECS task.
	•	AWS Account B: S3 bucket for storage and DynamoDB for lock table.

Steps Taken:

	1.	With AWS_S3_ASSUME_ROLE_ARN set: The error still occurred with the same 403 Access Denied response.
	2.	Without AWS_S3_ASSUME_ROLE_ARN: The error persisted, Same result.
	3.	With a dummy IAM role in AWS_S3_ASSUME_ROLE_ARN: Same result.
	4.	With S3 in the same account (Account A): Everything worked perfectly, confirming the issue is specific to cross-account role assumption.
	5.	Using the old version: Previous versions of kafka-delta-ingest (before v0.2.0) work fine with the exact same IAM role assumption setup.

Expected Behavior:

The application should properly assume the IAM role specified in the AWS_S3_ASSUME_ROLE_ARN environment variable and perform cross-account writes to S3 in Account B.

rtyler:
![image](https://github.com/user-attachments/assets/9a3bc7cf-c8b5-4b84-8752-026f61625369)


Based on my analysis [this change](https://github.com/delta-io/delta-rs/commit/17999d24a58fb4c98c6280b9e57842c346b4603a#diff-87e43fff59e072c5dbd99295f15fbf86d5a0e53f6586689191b2c61e07ace91aL390-L394) was responsible for removing assume role support. In essence `object_store` does not have native assume role support and in order to support this functionality we must first assume the role, get the temporary credentials, and _then_ construct the appropriate `AmazonS3` object store

PR #2887, opened 2024-09-16
feat: improve AWS credential loading between S3 and DynamoDb code paths

rtyler:
This change implements a number of improvements to the code paths for loading credentials.  This is a prerequisite to fixing assume role support #2879 but should also address a number of bugs I noticed:

* Python libraries can pass in keys via `storage_options` which are used for configuration of the AmazonS3 ObjectStore, but those credentials would not be used in the construction of the DynamoDB connection
* Using AWS credentials such as those from ~/.aws/profile or SSO would not be properly dropped into the AmazonS3 object store creation

There is some additional work that needs to come in to clean up how various options overrides are managed still.

Sponsored-by: [Scribd Inc.](https://tech.scribd.com)


rtyler:
Oh goodie, these tests fail in the GitHub Action runner environment and do not in local development. What a fun mystery this will be! :male_detective:

crates/aws/src/storage.rs
```diff
@@ -281,7 +232,9 @@ impl S3StorageOptions {
 
     fn ensure_env_var(map: &HashMap<String, String>, key: &str) {
         if let Some(val) = str_option(map, key) {
-            std::env::set_var(key, val);
+            unsafe {
```
hntd187, line 235: why is setting an environment variable unsafe?

hntd187 (approved):

PR #2896, opened 2024-09-22
fix: check lowercase config keys with/without aws prefix in s3logstorefactory

ion-elgreco:
# Description
We were only checking if the lowercase aws_copy_if_not_exists or aws_conditional_put are set to return the stores. But we should ignore the case and check for with and without aws prefix, since ObjectStore does this as well.


rwhaling:
@ion-elgreco - happy to!  is there a readme anywhere for how to set up for local development?  I've got rust etc. set up on this machine already but totally new to this repo, and not super familiar with rust/python bindings.

rwhaling:
keeping up with this branch, and my local dev environment is kind of a mess, but looks like this resolves my issue - "conditional_put", "CONDITIONAL_PUT", "aws_condititional_put", and "AWS_CONDITIONAL_PUT" all seem to work 👍

crates/aws/src/credentials.rs
```diff
@@ -19,7 +19,7 @@ use deltalake_core::storage::StorageOptions;
 use deltalake_core::DeltaResult;
 use tracing::log::*;
 
-use crate::constants;
+use crate::constants::{self, AWS_ENDPOINT_URL};
```
hntd187, line 22: prolly want to revert this and use consistent references here

hntd187 (approved):
LGMT I had one nit, but not necessary.

PR #2909, opened 2024-09-26
fix: update storage_options when using DeltaTable in write_deltalake

NilsIrl:
# Description

The previous code was erroneous and if a DeltaTable instance was kept for too long its credentials could expire with no way of updating them.


ion-elgreco:
@NilsIrl can you add a test please

NilsIrl:
Sorry I don't have the bandwidth to do that.

ion-elgreco:
This won't work with the recent changes anymore.

Issue #631, opened 2022-06-09
Data retrieval is grindingly slow with DeltaTable().to_pandas


boonware:
# Environment

**Delta-rs version**:

**Binding**: 

**Environment**:
- **Cloud provider**: S3 / IBM COS
- **OS**: MacOS / Linux
- **Other**: `deltalake v0.5.7 `

***
# Bug

**What happened**:
I have a Python Flask application that exposes a REST API to access the contents of a Delta Lake table for use in a web browser application. API requests to return records are grindingly slow, often in excess of 2 minutes. I have used a partitioned table with on the order of a few hundred (less than 1k) files. I have profiled the application and the bottleneck is in the `to_pandas` call. I see similar results if I replace the call with `to_pyarrow_table` or `to_pyarrow_dataset`.

Why is this call so slow?

```
dataset = DeltaTable(url).to_pandas(partitions=partitions, columns=columns).to_dict()
```

Is there a recommended approach that I am missing here?

**What you expected to happen**:

Data could be returned from the table in a reasonable length of time, i.e. that expected by a Reactive Web Application. 

**How to reproduce it**:
The code above using S3 or IBM COS as backend storage. Table with order of 100K records partitioned on a timestamp field that creates on the order of 100 files.

**More details**:
As stated above, application profiling shows the delay is not at the REST controller level.

MrPowers:
@boonware - I wouldn't expect reading multiple files into a pandas DataFrame via delta-rs to provide low latency queries.

Query engines that are optimized to read multiple files in parallel like Spark or Dask should be a lot faster.

Even single node query engines like polars or DataFusion should provide better performance.

Make sure to query the data and leverage partition filters, predicate pushdown filters, and column pruning.  Or just use a database.

someaveragepunter:
hopefully this thread is still being monitored after closed.
Just to clarify I've benchmarked a few comparisons when reading the same parquet files(s) of sizes between 1Mb - 100Mb from S3:
```
- pyarrow over s3fs: 5x faster
- duckdb: 3x faster
- awswrangler: 3x faster
```
than delta-rs. the files were written with delta-rs but read back from the 3 packages above

I like `delta-rs` for all the benefits it provides over raw parquet. however cannot justify trading the performance for the features.
is getting performance parity or btter with `pyarrow `and `duckdb` on the `delta-rs` roadmap

MrPowers:
[...]

PR #2922, opened 2024-10-04
feat: read delta table using datafusion

PeterKeDer:
# Description

This adds a new method `datafusion_read` to `DeltaTable`, which reads the delta table using datafusion and returns pyarrow batches. This supports reading with SQL predicates, e.g. `dt.datafusion_read(predicate="id = 10 and value > 123")`.

The performance of reading delta tables from AWS S3 using datafusion is vastly superior than using pyarrow, while also being much more reliable. For example, in one of our internal tables on S3 with ~20000 partitions of up to 300 MB each (in memory)

```python
dt = DeltaTable("s3://<omitted>")

# this returns one parquet file of size ~15 MB
files = dt.files(partition_filters=[("part_id", "=", "5956"), ("date", "=", "2024-10-03"), ("hour", "=", "18")])

# baseline of reading parquet directly: ~2 seconds
pl.read_parquet(files[0])

# default to_pyarrow_table: idk, I stopped it when it didn't even finish running after 2 minutes
dt.to_pyarrow_table(filters=[("part_id", "=", "5956"), ("date", "=", "2024-10-03"), ("hour", "=", "18")])

# read using datafusion: ~4 seconds
pa.Table.from_batches(dt.datafusion_read(predicate="part_id = 5956 and date = '2024-10-03' and hour = 18"))
```

# Related Issue(s)

Similar issue reported in https://github.com/delta-io/delta-rs/issues/631


python/src/lib.rs
```diff
@@ -1232,6 +1239,65 @@ impl RawDeltaTable {
...
+                df = df.filter(filter).unwrap();
+            }
+
+            if let Some(columns) = columns {
+                df = df
+                    .select_columns(&columns.iter().map(String::as_str).collect::<Vec<_>>())
+                    .unwrap();
+            }
+
+            Ok(rt().block_on(async { df.collect().await }).unwrap())
```
houqp, line 1296: we should propagate the errors properly instead of unwraps here

ion-elgreco (changes requested):
I am on the fence with adding support like this in its current state

There were previous discussions on this, and I believe using an external TableProvider that can be registered in Datafusion python is the best way. At this stage that's not possible yet though

However maybe to make this more practical and have it working as a drop in replacement for to_pyarrow_dataset, it should take in pyarrow expressions which are then mapped to Datafusion expressions

rtyler:
@ion-elgreco why bother mapping into pyarrow? The python library has to have datafusion bundled either way, I think it's valuable to expose it to users  This approach is in line with what I had suggested previously, exposing just a simple way to get at Datafusion and return record batches for the users. 


The datafusion-python discussion has taken months and frankly I don't see much value in Datafusion bindings for Python to begin with, so I don't hang high hopes on that meeting the use-case here.

I would suggest, if @PeterKeDer is open to it, inverting this a little bit. Adding predicate searches on a loaded DeltaTable leads to exposing a subset of Datafusion and is likely to see API bloat in the future as users would want more functionality from Datafusion up in Python.

Perhaps a:

```python
class QueryBuilder:
  def __init__(self):
     # some state, like SessionContext
  def use(self, name, table):
     # for registering table in that context
  def execute(self, df_sql) -> List[RecordBatch]:
     # Do the thing
```

Such that a user would be able to:

```python
dt = DeltaTable('s3://foo')
dt2 = DeltaTable('s3://bar')

qe = QueryBuilder().use('foo', dt).use('bar', dt2)

# Forgive lazy SQL syntax, it's early =_=
results = qe.execute('SELECT foo.id, bar.description FROM foo JOIN foo.id = bar.foreign_id')
```

That way users would be able to use multiple tables in their query if they have them, and we wouldn't need to do much result pruning from Datafusion since they could just use the SQL engine directly to pull the information out that they want.

ion-elgreco:
@rtyler most integrations use the pyarrow dataset interface and the have mapped their own expressions to pyarrow expressions, so was simply thinking it would be a replacement for that.

I like the queryBuilder approach though! Maybe call it `register` over `use`.

PeterKeDer:
@ion-elgreco @rtyler thanks for the suggestions!

I will look into the `QueryBuilder` approach. Just so I understand correctly, the general idea is that we leave functionality like projecting/filtering/joining to datafusion SQL, instead of duplicating them in our python API?

ion-elgreco:
With this PR https://github.com/apache/datafusion-python/pull/921 we don't need this anymore since people can register our tables directly in Datafusion, so let's wait for that

houqp:
The datafusion python change would still require users to install a compatible datafusion python package and learn the datafusion python api right? Given that we already have datafusion embedded as a dependency, it seems like a better out of the box experience for our users if they can get much faster data scan with a single api call without any extra setup.

ion-elgreco:
> The datafusion python change would still require users to install a compatible datafusion python package and learn the datafusion python api right? Given that we already have datafusion embedded as a dependency, it seems like a better out of the box experience for our users if they can get much faster data scan with a single api call without any extra setup.

Any version after Datafusion with ffi support would be compatible. 

I don't think it's worth the maintenance, and API bloat for something that can be solved natively within Datafusion-python.

rtyler:
I share @houqp 's concerns about the datafusion-python code. Considering how early in draft status it is, and points to unreleased datafusion code :see_no_evil:  it's obviously still too early to tell. I'm comfortable keeping this in a draft state for the time being.

PeterKeDer:
Hey @rtyler @ion-elgreco I took a shot implementing the `QueryBuilder` approach. IMO this would still be pretty valuable for people who don't want to manage a separate datafusion-python dependency. Would appreciate another look!

rtyler (dismissed):
I think this is a cute little API addition. @ion-elgreco do you know of a good way we could mark this as experimental or subject to change? I understand your reluctance to add more API surface to maintain but I am curious to see how people interact/use this type of API.

If we find the datafusion-python route eventually to be more valuable, then we could remove the experimental API

ion-elgreco:
@rtyler we can raise a warning when calling QueryBuilder.

Something like this:

class ExperimentalWarning(Warning))

And then call warnings.warn with that and some text

PeterKeDer:
Thanks, I added the experimental warning to `QueryBuilder`

ion-elgreco:
@rtyler do we still want this, now that you can register deltatables directly in Datafusion-python

asura-io:
@ion-elgreco @rtyler 

Bumping this PR, I'm in favor of moving forward with this approach, rather than using `datafusion-python`. 

Some rationale (mostly echoing @houqp's rationale [above](https://github.com/delta-io/delta-rs/pull/2922#issuecomment-2440478259)):
1. Adding `datafusion-python` adds an additional python dependency for the user. 
2. `delta-rs` already contains the datafusion crate. This will work out of the box. 
3. The change @PeterKeDer put together is pretty simple and the `QueryBuilder` API is intuitive and extensible. 
4. `datafusion-python` is in an early stage of development and may present other issues with integration.

rtyler:
@ion-elgreco I still want this :smile:  I am cleaning this pull request up now and preparing it for merge

rtyler (dismissed):

ion-elgreco (approved):

Issue #2925, opened 2024-10-06
createdTime lost in metadata after schema change


crpcrp:
# Environment

**Delta-rs version**: 0.20.1

**Binding**: Python

**Environment**:
- **Cloud provider**:
- **OS**: Windows
- **Other**:

***
# Bug

**What happened**:
Using mode='append' and schema_mode='merge', the createdTime field became null in the metadata if there was a change in the schema in an append.

**What you expected to happen**:
The createdTime should be kept as its original value even after schema changes.

**How to reproduce it**:

```python
import pandas as pd
from deltalake import write_deltalake,


data1 = {
    "a": [1, 2],
    "b": [10, 20]
}

data2 = {
    "a": [3, 4],
    "c": ['X', 'Y']
}

df1 = pd.DataFrame(data1)
df2 = pd.DataFrame(data2)

write_deltalake('my_table',df1,mode='append',schema_mode='merge')
write_deltalake('my_table',df2,mode='append',schema_mode='merge')
```

**More details**:
```
{
    "metaData": {
        "id": "326d7d04-a989-48ea-99e6-590eb8d66946",
        "name": null,
        "description": null,
        "format": {
            "provider": "parquet",
            "options": {}
        },
        "schemaString": "{\"type\":\"struct\",\"fields\":[{\"name\":\"a\",\"type\":\"long\",\"nullable\":true,\"metadata\":{}},{\"name\":\"b\",\"type\":\"long\",\"nullable\":true,\"metadata\":{}},{\"name\":\"c\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}]}",
        "partitionColumns": [],
        "createdTime": null,
        "configuration": {}
    }
}
{
    "add": {
        "path": "part-00001-026f1f36-2361-4a5f-a115-b2f00ec7962d-c000.snappy.parquet",
        "partitionValues": {},
        "size": 1038,
        "modificationTime": 1728224675640,
        "dataChange": true,
        "stats": "{\"numRecords\":2,\"minValues\":{\"a\":3,\"c\":\"X\"},\"maxValues\":{\"a\":4,\"c\":\"Y\"},\"nullCount\":{\"a\":0,\"b\":2,\"c\":0}}",
        "tags": null,
        "deletionVector": null,
        "baseRowId": null,
        "defaultRowCommitVersion": null,
        "clusteringProvider": null
    }
}
{
    "commitInfo": {
        "timestamp": 1728224675640,
        "operation": "WRITE",
[...]

PR #2926, opened 2024-10-06
fix: keep original metadata during schema evolution

ion-elgreco:
# Description

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2925


rtyler (changes requested):
I do not believe this is the right fix for the "bug", _technically_ the `createdTime` is an optional field but it should be from the time when the action itself is being created. 

I will take this and rework it with a test

ion-elgreco:
> I do not believe this is the right fix for the "bug", _technically_ the `createdTime` is an optional field but it should be from the time when the action itself is being created. 
> 
> I will take this and rework it with a test

Ah I thought it was the time of table creation

rtyler (approved):

PR #2942, opened 2024-10-14
feat(python): reading/writing transaction identifiers in python

PeterKeDer:
# Description

We'd like to read/write the delta transaction identifiers for stream processing. It seems that this is currently available to through rust but not python. This PR adds the following to the python API:
- `app_transactions` field in `CommitProperties`, which is a list of the `Transaction` class (`PyTransaction` in rust)
- `DeltaTable.transaction_versions` for accessing the latest transaction of a delta table, which returns a dictionary of app id -> `Transaction`

Would like some feedback on the naming/API. I'm not sure which is the correct terminology to use - "app transaction" vs "transaction version" vs "transaction identifier"? The delta protocol refers to it as [transaction identifiers](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers).


houqp:
transaction identifier would have been a better name if we are to start from scratch, but to keep it consistent with the rust core, it's better to keep using app transaction for now.

houqp (approved):

python/src/lib.rs
```diff
@@ -1232,6 +1232,19 @@ impl RawDeltaTable {
...
     }
+
+    pub fn transaction_versions(&self) -> HashMap<String, String> {
+        self._table
+            .get_app_transaction_version()
+            .iter()
+            .map(|(app_id, transaction)| {
+                (
+                    app_id.to_owned(),
+                    serde_json::to_string(transaction).unwrap(),
```
ion-elgreco, line 1243: Maybe return the PyTransaction here

ion-elgreco (changes requested):

python/src/lib.rs
PeterKeDer: @ion-elgreco updated! I made `PyTransaction` a `pyclass` so we can pass it to/from python more easily

ion-elgreco (approved):
Nice work!

PR #2955, opened 2024-10-22
chore: fix many warnings across the codebase

hntd187:
# Description
Fixed a lot of warnings across the codebase according to clippy and `cargo check` the S3 options one is the only one I am not 100% sure is correct, but @rtyler maybe you can comment on if it's a good idea to remove the deprecated key usages. Otherwise I can revert those.

# Related Issue(s)
Closes my own internal OCD about warnings (despite there still being one)


crates/aws/src/constants.rs
```diff
@@ -99,8 +105,8 @@ pub const S3_OPTS: &[&str] = &[
     AWS_SECRET_ACCESS_KEY,
     AWS_SESSION_TOKEN,
     AWS_S3_LOCKING_PROVIDER,
-    AWS_S3_ASSUME_ROLE_ARN,
-    AWS_S3_ROLE_SESSION_NAME,
```
rtyler, line 103: I mentioned to @hntd187 in Slack that I would like to avoid removing these deprecation warnings until we have a few more releases which is print deprecation warnings for people

rtyler (approved):

PR #2956, opened 2024-10-22
fix: dramatically reduce checkpoint memory consumption

rtyler:
Both commits describe the specific fixes, but basically our checkpoint code was collecting too much into memory when it could iterate! :roller_coaster: 

With a test table:

Before

`Maximum resident set size (kbytes): 19964728`

After

`Maximum resident set size (kbytes): 4017132`

Sponsored-by: [Scribd Inc](https://tech.scribd.com)


ion-elgreco (approved):
Nice catch!

Issue #2882, opened 2024-09-13
Creating delta table with timestampNtz will create spark-incompatible delta protocol


wahani:
# Environment

**Delta-rs version**: python-0.16.0+ (example works until 0.15.3, also tested with 0.19.2)

**Binding**: Python 

**Environment**:
- **Cloud provider**:
- **OS**: MacOS Monterey
- **Other**: 
    - pyspark: 3.5.0
    - pandas: 2.2.2
    - pyarrow: 17.0.0   

***
# Bug

**What happened**:

Beginning with version python-0.16.0+ I receive a DeltaTableFeatureException when reading a delta table with a local spark session. This is related to the newly introduced handling of timezones. In 0.15.3 the write operation will create the following protocol `{"protocol":{"minReaderVersion":1,"minWriterVersion":2}}` with 0.16.0 and onward we will see `{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"readerFeatures":["timestampNtz"],"writerFeatures":["timestampNtz"]}}`. However, when reading the delta table using a local spark session, I will get the following error message:

```
py4j.protocol.Py4JJavaError: An error occurred while calling o33.load.
: org.apache.spark.sql.delta.DeltaTableFeatureException: [DELTA_FEATURES_PROTOCOL_METADATA_MISMATCH] Unable to operate on this table because the following table features are enabled in metadata but not listed in protocol: invariants.
        at org.apache.spark.sql.delta.DeltaErrorsBase.tableFeatureMismatchException(DeltaErrors.scala:2281)
        at org.apache.spark.sql.delta.DeltaErrorsBase.tableFeatureMismatchException$(DeltaErrors.scala:2278)
        at org.apache.spark.sql.delta.DeltaErrors$.tableFeatureMismatchException(DeltaErrors.scala:3382)
        at org.apache.spark.sql.delta.DeltaLog.assertTableFeaturesMatchMetadata(DeltaLog.scala:436)
        at org.apache.spark.sql.delta.Snapshot.init(Snapshot.scala:247)
        at org.apache.spark.sql.delta.Snapshot.<init>(Snapshot.scala:531)
        at org.apache.spark.sql.delta.SnapshotManagement.$anonfun$createSnapshot$2(SnapshotManagement.scala:634)
        at org.apache.spark.sql.delta.SnapshotManagement.createSnapshotFromGivenOrEquivalentLogSegment(SnapshotManagement.scala:796)
[...]

PR #2957, opened 2024-10-23
fix: always include the Invariants writerFeature when upgrading to v7

rtyler:
The presence of a timestampntz column in CreateBuilder results in the reader/writer being set to 3/7. The protocol reaquires that if the table is writer v7, that `invariants` "must exist in the table protocl's writerFeatures.

    [DELTA_FEATURES_PROTOCOL_METADATA_MISMATCH] Unable to operate on this table because the following table features are enabled in metadata but not listed in protocol: invariants. SQLSTATE: 0AKDE

_Sidenote:_ Yes I know that we cannot actually support invariant utilization in delta-rs, but creating a table with `timestampntz` will (rightfully) be invalid and unreadable by Databricks and other compliant Delta readers:

    
Sponsored-by: Scribd Inc

### DIFF
--- a/crates/core/src/kernel/models/actions.rs
+++ b/crates/core/src/kernel/models/actions.rs
@@ -372,7 +372,10 @@ impl Protocol {
     /// Enable timestamp_ntz in the protocol
     pub fn enable_timestamp_ntz(mut self) -> Protocol {
         self = self.with_reader_features(vec![ReaderFeatures::TimestampWithoutTimezone]);
-        self = self.with_writer_features(vec![WriterFeatures::TimestampWithoutTimezone]);
+        self = self.with_writer_features(vec![
+            WriterFeatures::TimestampWithoutTimezone,
+            WriterFeatures::Invariants,
+        ]);
         self
     }
 }
--- a/crates/core/src/operations/create.rs
+++ b/crates/core/src/operations/create.rs
@@ -274,7 +274,9 @@ impl CreateBuilder {
             Protocol {
                 min_reader_version: 3,
                 min_writer_version: 7,
-                writer_features: Some(hashset! {WriterFeatures::TimestampWithoutTimezone}),
+                writer_features: Some(
+                    hashset! {WriterFeatures::TimestampWithoutTimezone, WriterFeatures::Invariants},
+                ),
                 reader_features: Some(hashset! {ReaderFeatures::TimestampWithoutTimezone}),
             }
         } else {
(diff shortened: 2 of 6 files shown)


ion-elgreco:
@rtyler to my understanding it's not required, but it's assumed when a table has NOT NULL columns

ion-elgreco:
See discussion here: https://github.com/delta-io/delta-rs/issues/2882#issuecomment-2401988070

rtyler:
@ion-elgreco I should also mention that this came from a table created by delta-rs and does not contain any actual invariants. I believe newer DBRs and Delta/Spark releases are more strictly adhering to the protocol and expect invariants to be defined on writerVersion 7

ion-elgreco:
> I added this because of the error and [this statement in the protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-invariants)
> 
> > If the table is on Writer Version 7, the feature invariants must exist in the table protocol's writerFeatures.

It's a bit poorly written, but the sentence starts with "To support this feature". I also  misread it before and assumed it was required but it's just to support it in v7 we need to add the feature.

Did you check when you create a v3,7 table with nullable columns in spark what the protocol log states?

rtyler:
Okay so I did some testing with PySpark and I am going to pull this into  draft again to modify the changes a bit. Essentially if a non-nullable column exists and things are upgraded to writer7, the `invariants` writer feature is added, but _no invariants are set_ :confused: 

This behavior is not specified anywhere in the protocol which is fun :skull: 

```python
DeltaTable.create(spark).tableName('nonulliz').addColumn('ts', 'TIMESTAMP_NTZ', nullable=False).location('./nonulliz').execute()
```


```json
{
  "commitInfo": {
    "timestamp": 1729706895482,
    "operation": "CREATE TABLE",
    "operationParameters": {
      "partitionBy": "[]",
      "clusterBy": "[]",
      "description": null,
      "isManaged": "false",
      "properties": "{}"
    },
    "isolationLevel": "Serializable",
    "isBlindAppend": true,
    "operationMetrics": {},
    "engineInfo": "Apache-Spark/3.5.3 Delta-Lake/3.2.0",
    "txnId": "1e0e57c1-d1a2-41d9-9402-53516b8f4c77"
  }
}
{
  "metaData": {
    "id": "721d696f-8396-4a69-851c-23702b8b61bb",
    "format": {
      "provider": "parquet",
      "options": {}
    },
    "schemaString": "{\"type\":\"struct\",\"fields\":[{\"name\":\"ts\",\"type\":\"timestamp_ntz\",\"nullable\":false,\"metadata\":{}}]}",
    "partitionColumns": [],
    "configuration": {},
    "createdTime": 1729706895473
  }
}
{
  "protocol": {
    "minReaderVersion": 3,
    "minWriterVersion": 7,
    "readerFeatures": [
      "timestampNtz"
    ],
    "writerFeatures": [
      "timestampNtz",
      "invariants"
    ]
  }
}
```

PR #2983, opened 2024-11-10
perf: batch json decode checkpoint actions when writing to parquet

alexwilcoxson-rel:
# Description
This change pushes more serialized json actions into the decoder before flushing. For a log with 10s of thousands of actions, the current implementation took ~18 seconds, this change dropped it to 3.

# Related Issue(s)
n/a

# Documentation
https://docs.rs/arrow-json/53.2.0/arrow_json/reader/struct.Decoder.html#method.decode

### DIFF
--- a/crates/core/src/protocol/checkpoints.rs
+++ b/crates/core/src/protocol/checkpoints.rs
@@ -353,18 +353,22 @@ fn parquet_bytes_from_state(
     // Count of actions
     let mut total_actions = 0;
 
-    for j in jsons {
-        let buf = serde_json::to_string(&j?).unwrap();
-        let _ = decoder.decode(buf.as_bytes())?;
-
-        total_actions += 1;
+    let span = tracing::debug_span!("serialize_checkpoint").entered();
+    for chunk in &jsons.chunks(CHECKPOINT_RECORD_BATCH_SIZE) {
+        let mut buf = Vec::new();
+        for j in chunk {
+            serde_json::to_writer(&mut buf, &j?)?;
+            total_actions += 1;
+        }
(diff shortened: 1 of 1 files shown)


hntd187:
Does it make sense to tie it to record batch size? Could we instead pull that into it's own configuration or at least constant?

alexwilcoxson-rel:
> Does it make sense to tie it to record batch size? Could we instead pull that into it's own configuration or at least constant?

I think if its less than the checkpoint batch size, then the `decode` and `flush` will yield a batch < the checkpoint batch size constant. For example

```rust
    const CHECKPOINT_RECORD_BATCH_SIZE: usize = 5000;

    let mut decoder = ReaderBuilder::new(arrow_schema)
        .with_batch_size(CHECKPOINT_RECORD_BATCH_SIZE)
        .build_decoder()?;

    // Count of actions
    let mut total_actions = 0;

    for chunk in &jsons.chunks(2500) {
        let mut buf = Vec::new();
        // write 2500 serialized json objects to buffer
        for j in chunk {
            serde_json::to_writer(&mut buf, &j?)?;
            total_actions += 1;
        }
        // internally buffers 2500 objects and returns because
        // buf is exhausted even though 2500 < 5000
        let _ = decoder.decode(&buf)?;
        // flush yields 2500 row batch and writes it
        while let Some(batch) = decoder.flush()? {
            writer.write(&batch)?;
        }
    }
```

hntd187:
I understand, I guess I'm asking is there any reason a user might want to configure this independently from the batch size like above? If we don't think it's ever gonna happen then we can forgo configuration.

rtyler (approved):

PR #2987, opened 2024-11-12
fix: cache credential resolution with the AWS credential provider

rtyler:
`object_store` invokes `get_credential` on _every_ invocation of a get/list/put/etc. The provider invocation for environment based credentials is practically zero-cost, so this has no/low overhead.

In the case of the AssumeRoleProvider or any provider which has _some_ cost, such as an invocation of the AWS STS APIs, this can result in rate-limiting or service quota exhaustion.

In order to prevent this, the credentials are attempted to be cached only so long as they have no expired, which is defined in the `aws_credential_types::Credential` struct

Sponsored-by: Scribd Inc


mightyshazam (approved):
LGTM

hntd187 (approved):

kylebarron:
By my reading, `aws_config` includes an identity cache by default: https://docs.rs/aws-config/latest/aws_config/identity/struct.IdentityCache.html

Were you having issues with that cache?

PR #2997, opened 2024-11-17
feat: concurrent commit changes to `PartitionWriter`

Dan-J-D:
# Description
Added `drain_files_written()` to `PartitionWriter` to drain the not yet committed changes.

# Related Issue(s)
None

# Documentation
None

### DIFF
--- a/crates/core/src/operations/writer.rs
+++ b/crates/core/src/operations/writer.rs
@@ -419,6 +419,7 @@ impl PartitionWriter {
         Ok(())
     }
 
+    /// Retrieves the list of [Add] actions associated with the files written.
     pub async fn drain_files_written(&self) -> Vec<Add> {
         self.files_written.write().await.drain(..).collect()
     }
--- a/crates/core/src/operations/writer.rs
+++ b/crates/core/src/operations/writer.rs
@@ -287,7 +287,7 @@ pub struct PartitionWriter {
     buffer: ShareableBuffer,
     arrow_writer: ArrowWriter<ShareableBuffer>,
     part_counter: usize,
-    files_written: RwLock<Vec<Add>>,
+    files_written: Vec<Add>,
     /// Num index cols to collect stats for
     num_indexed_cols: i32,
     /// Stats columns, specific columns to collect stats from, takes precedence over num_indexed_cols
@@ -316,7 +316,7 @@ impl PartitionWriter {
             buffer,
             arrow_writer,
             part_counter: 0,
-            files_written: RwLock::new(Vec::new()),
+            files_written: Vec::new(),
             num_indexed_cols,
             stats_columns,
         })
@@ -382,7 +382,7 @@ impl PartitionWriter {
         .map_err(|err| WriteError::CreateAdd {
             source: Box::new(err),
         })?;
-        self.files_written.write().await.push(add_metadata);
+        self.files_written.push(add_metadata);
 
         Ok(())
     }
@@ -420,14 +420,14 @@ impl PartitionWriter {
     }
 
     /// Retrieves the list of [Add] actions associated with the files written.
-    pub async fn drain_files_written(&self) -> Vec<Add> {
-        self.files_written.write().await.drain(..).collect()
+    pub async fn drain_files_written(&mut self) -> Vec<Add> {
+        self.files_written.drain(..).collect()
     }
 
     /// Close the writer and get the new [Add] actions.
     pub async fn close(mut self) -> DeltaResult<Vec<Add>> {
         self.flush_arrow_writer().await?;
-        Ok(self.files_written.read().await.clone())
+        Ok(self.files_written.clone())
     }
 }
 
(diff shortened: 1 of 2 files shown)


hntd187:
What is the motivation behind this change? Can you give more details behind what you are trying to accomplish here?

hntd187:
You have to open a new one every time because you shouldn't be committing partitions concurrently with a locking service or mechanism. I do not think this is correct in that case because if you are sharing a partition writer between commits then you could end up putting add actions in the wrong commits. They are meant to be ordered and conflict free, which I don't know if this would accomplish.

Dan-J-D:
> You have to open a new one every time because you shouldn't be committing partitions concurrently with a locking service or mechanism. I do not think this is correct in that case because if you are sharing a partition writer between commits then you could end up putting add actions in the wrong commits. They are meant to be ordered and conflict free, which I don't know if this would accomplish.

When I talk about concurrently I more am talking about multi system, I am not in need for the locking mechanism. If I remove the locking mechanism, will this solve your issue and PR be merged?

hntd187:
There wouldn't be anything left in your PR if you removed the locking mechanism here, right?

Dan-J-D:
> There wouldn't be anything left in your PR if you removed the locking mechanism here, right?

This would still be left
```rs
pub async fn drain_files_written(&self) -> Vec<Add> {
    self.files_written.write().await.drain(..).collect()
}
```

Dan-J-D:
@hntd187 I have removed the lock implementation.

crates/core/src/operations/writer.rs
```diff
@@ -218,18 +218,11 @@ impl DeltaWriter {
...
-                acc.extend(actions);
-                futures::future::ready(Ok(acc))
-            })
-            .await?;
-
+        let mut actions = Vec::new();
+        for (_, writer) in writers {
+            let writer_actions = writer.close().await?;
+            actions.extend(writer_actions);
+        }
```
rtyler, line 225: @Dan-J-D I'm not sure I understand the motivation of this change. I believe @alexwilcoxson-rel or @wjones127 introduced the original code above to ensure that the process which was running the writer would be able to concurrently produce the `Vec<Action>`

Based on my understanding of this change, it walks back that performance improvement but I cannot figure out for what gain :thinking:

crates/core/src/operations/writer.rs
```diff
@@ -427,10 +419,15 @@ impl PartitionWriter {
         Ok(())
     }
 
+    /// Retrieves the list of [Add] actions associated with the files written.
+    pub async fn drain_files_written(&mut self) -> Vec<Add> {
```
rtyler, line 423: Is the goal for this function to allow for data files (`.parquet`) to be written without the process committing to the Delta table?  I have read the discourse with @hntd187 but I'm still struggling to understand how this public API would be used

crates/core/src/operations/writer.rs
Dan-J-D: I was on a older version of the library and I just copy & pasted my local copy into github, but there were newer changes already made, so I added them back

crates/core/src/operations/writer.rs
Dan-J-D: So let's say in an event where a server has written a lot of data into the table but hasn't committed the additional files yet and there was a power outage. The files will be in storage but not in the metadata. So this can be used to prevent data loss (While writing data, also creating commits of the newly added files).

alexwilcoxson-rel: Would something be calling drain periodically while writes are still happening then? What would be the difference vs closing the writer periodically and creating a new one?

crates/core/src/operations/writer.rs
```diff
@@ -427,10 +419,15 @@ impl PartitionWriter {
...
+        self.files_written.drain(..).collect()
+    }
+
     /// Close the writer and get the new [Add] actions.
     pub async fn close(mut self) -> DeltaResult<Vec<Add>> {
         self.flush_arrow_writer().await?;
-        Ok(self.files_written)
+        Ok(self.files_written.clone())
```
alexwilcoxson-rel, line 430: self is moved here, so after calling close its not usable. This clone should be unnecessary.

crates/core/src/operations/writer.rs
Dan-J-D: Saved cost of reallocation, re-initialization and if you are closing after a certain time frame it is possible to create just a lot of small files which aren't optimized. Then will need to OPTIMIZE.

Dan-J-D: Yes, I made an alternative to delta-kafka-ingest for my specific use case and it will be receiving data all day.

Dan-J-D:
Now that I understand the correct usage for `PartitionWriter`, i will be closing this PR. Although this has benefits, its very niche. Thanks for reviewing this PR

PR #3010, opened 2024-11-21
feat: override dynamodb config

### LINKED ISSUE #2287: Accommodate separate set of credentials to access DynamoDB
# Description
When using an S3 compatible storage other than AWS S3, one has a set of access and secret key for the storage and another set for the DynamoDB. In this case, we need an option to provide separate credentials for DynamoDB access.

**Use Case**
Concurrent writes to an S3 compatible storage that does not share the credentials as DynamoDB.

**Related Issue(s)**

### LINKED ISSUE #2497: [deltalake_aws] Allow passing AWS_SESSION_TOKEN and other credentials more directly to DynamoDB client.
# Description

My understanding of the code [here](https://github.com/delta-io/delta-rs/blob/81593e919497221a1a08bf8db9d20e8e4a39a8a6/crates/aws/src/storage.rs#L146-L153) is that AWS_SESSION_TOKEN, AWS_SECRET_ACCESS_KEY, and AWS_ACCESS_KEY_ID are assigned to environment variables, and later inferred by the S3 client and the DynamoDB client. It would be nicer if they could be passed directly to these clients without being part of the process's environment.

I think an ideal solution would allow me to configure my own object store and dynamo client, and provide those when registering a table.

**Use Case**

My process is a big monolith with a bunch of stuff in it, completely unrelated to me. Many different identities are used in this process for many different purposes. Environment variables are a single global namespace, so it would be dangerous to plug my credentials in it.

**Related Issue(s)**

thomas-chauvet:
# Description

This PR add the possibility to override dynamoDB configuration via `storage_options`:

- `AWS_ENDPOINT_URL_DYNAMODB` overrides `AWS_ENDPOINT_URL`,
- `AWS_REGION_DYNAMODB` overrides `AWS_REGION`,
- `AWS_ACCESS_KEY_ID_DYNAMODB` overrides `AWS_ACCESS_KEY_ID`,
- `AWS_SECRET_ACCESS_KEY_DYNAMODB` overrides `AWS_SECRET_ACCESS_KEY`.

It solves cases where we use an S3-compatible storage without mutual exclusion but we want to use DynamoDB anyway as a locking provider.
It also solves the case where we want to pass different credentials or region for S3 and dynamoDB.

# Related Issue(s)
- closes #2497 
- closes #2287 

# Documentation

Documentation has been updated to explain how to override configuration.

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -31,7 +31,7 @@ jobs:
         run: make check-rust
 
   test-minimal:
-    name: Python Build (Python 3.8 PyArrow 16.0.0)
+    name: Python Build (Python 3.9 PyArrow 16.0.0)
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: "-C debuginfo=line-tables-only"
@@ -43,7 +43,7 @@ jobs:
       - name: Setup Environment
         uses: ./.github/actions/setup-env
         with:
-          python-version: 3.8
+          python-version: 3.9
 
       - name: Build and install deltalake
         run: |
@@ -135,7 +135,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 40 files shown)


thomas-chauvet:
Issue with DCO, I'll re-open a PR ASAP.

PR #3012, opened 2024-11-21
feat: provide direct TableProvider integration in datafusion-python

### LINKED ISSUE #2536: Provide a direct integration with Datafusion
I recently discovered that both [polars](https://github.com/pola-rs/polars/issues/16248) and [datafusion-python](https://github.com/apache/datafusion-python/issues/703) do not push down timestamp predicates correctly for pyarrow datasets. This is problematic: using a timestamp filter is a very, very common use case. I suspect both those libraries can implement / fix it but for datafusion going from datafusion (inside of delta-rs) -> pyarrow -> back to datafusion (in datafusion-python) seems like unecessary overhead and evidently somewhat brittle. And since the failure is silent it took me weeks to discover and I only found it because I noticed that using deltalake was _a lot_ slower than just accessing the raw parquet data for certain queries.

Would this crate be opposed to offering a direct integration with datafusion, since datafusion is already used internally? This would be ideal as some sort of extra or plugin but sadly that's not really possible with the current state of PyO3 extensions.

Some ideas:
1. datafusion-python allows implementing a PythonTableProvider that while still requiring going from rust -> python -> rust would at least remove a brittle conversion to pyarrow and back and would be more akin to serializing to python and deserializing back to rust (maybe even using serde or https://github.com/davidhewitt/pythonize?). Basically this crate would have to make a python wrapper for the existing TableProvider that could then be passed into datafusion-python.
2. This crate depends on datafusion-python and re-exports the entire Python API.
3. We add datafusion-python as an optional cargo dependency and then re-export the API under `detlake-datafusion` (which you wouldn't be able to mix and match with plain `deltalake` e.g. passing a `deltalake.DetlaTable` between them, you'd have to make a `deltalake_datafusion.DeltaTable` although the latter maybe could extract information from the former or call it via the Python APIs).

timsaucer:
# Description

This change adds in a single method on the Delta Table object to expose a PyCapsule that provides a DataFusion FFI Table Provider. With this change, you can register a delta table in datafusion-python without the need to export a pyarrow_dataset. This enables full push down filter support, greatly improving performance in some cases.

# Related Issue(s)

Closes #2536

# Documentation

This is a follow on from https://github.com/apache/datafusion/pull/12920 and https://github.com/apache/datafusion-python/issues/823


timsaucer:
FYI @ion-elgreco

ion-elgreco:
@timsaucer super excited with this change!

rtyler (dismissed):

timsaucer:
I'll try to address that ci failure tomorrow morning

ion-elgreco (approved):

adriangb:
Amazing!

Issue #2991, opened 2024-11-13
Faulty behavior on nullable columns when using `DeltaTable.merge`


joepatol:
# Environment

**Delta-rs version**:  0.21.0

**Binding**: Python

**Environment**:
- **Cloud provider**: local & Azure
- **OS**: Ubuntu
- **Other**:

***
# Bug

There seems to be faulty behavior in `DeltaTable.merge` when columns are non-nullable. (i.e. `nullable=False`).

**What happened**:
if I merge an empty table with `None` values the first time, this operation runs ok. The second time the operation fails.

**What you expected to happen**:
I expect if a column is non-nullable, trying to write a `null` / `None` value to it, would always immediately raise an error.

**How to reproduce it**:
Run the following script, make sure the table **does not exist**:

```python
from deltalake.schema import Schema, Field, PrimitiveType
from deltalake import DeltaTable
import pandas as pd
import os
from pathlib import Path

TABLE_PATH = str(Path(os.path.dirname(__file__)) / "my_table")

schema = Schema([
    Field("id", PrimitiveType("integer"), nullable=False),
    Field("bool", PrimitiveType("boolean"), nullable=False),
])


df1 = pd.DataFrame(
    columns=["id", "bool"],
    data=[
        [1, True],
        [2, None],
        [3, False],
    ],
)


df2 = pd.DataFrame(
    columns=["id", "bool"],
    data=[
        [4, False],
        [5, None],
        [6, True],
    ],
)


if not DeltaTable.is_deltatable(TABLE_PATH):
    DeltaTable.create(TABLE_PATH, schema=schema)

for df in [df1, df2]:
    dt = DeltaTable(TABLE_PATH)

    upsert_result = dt.merge(
        source=df,
        source_alias="s",
        target_alias="t",
        predicate="s.id = t.id",
    ) \
        .when_matched_update_all() \
        .when_not_matched_insert_all() \
        .execute()

    print(upsert_result)

    print(dt.to_pandas())
```

Running this the first time yields:

```
[...]

PR #3026, opened 2024-11-24
fix: check top level nullability during write

ion-elgreco:
# Description
Adds a check for top level column nullability whether the arrays in recordbatch contain nulls when they shouldn't.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2991

## Notes
This is probably required in nested fields as well, so something for a follow-up


hntd187 (dismissed):
LGTM

rtyler (approved):
I think a good future refactor would be to store a reference to the snapshot in the DeltaDataChecker and then just use that instead of copying the constraints, invariants, etc for doing checks

PR #3031, opened 2024-11-25
fix: workaround for Add actions being read slightly differently out of parquet files

### LINKED ISSUE #3030: delta.checkpointInterval breaks writes on 0.22.0
# Environment

**Delta-rs version**: 0.22.0

**Binding**: python

***
# Bug

**What happened**: I am trying to upgrade to the latest release of delta-rs and it seems to introduce a breaking bug when using `delta.checkpointInterval`.

**What you expected to happen**: checkpointInterval continues to work as before, and I can write safely.

**How to reproduce it**: This introduces a breaking bug in both pyarrow and rust writer engines. In pyarrow, it does not overwrite successfully (two rows in output), and in rust, it panics.

```
import pandas as pd 

write_deltalake(
    "test",
    pd.DataFrame(
        {
            "a": ["a"],
            "b": [1],
        }
    ),
    configuration={
        "delta.checkpointInterval": "5"
    }
)

for i in range(4):
    write_deltalake(
        "test",
            pd.DataFrame(
                {
                    "a": ["a"],
                    "b": [i],
                }
            ),
        mode="overwrite",
    )
```

At this point, the delta table looks correct, e.g.:
image

But, on the next write, when using `rust` writer engine, we get the following exception:
```
PanicException: called `Result::unwrap()` on an `Err` value: DeletionVector("Unknown storage format: ''.")
```

and on pyarrow, it manifests itself with an incorrect overwrite:
image
with two rows showing up.

rtyler:
This  workaround fixes #3030 but I'm not quite happy about how. I believe there is likely an upstream issue in arrow that I have not been able to reproduce where the nullable struct (deletionVector) which has non-nullable members (e.g. storageType) that are being read as empty strings rather than `null`.

This issue started to appear with the v0.22 release which included an upgrade to datafusion 43 and arrow 53. I believe the latter is causing the issue here since it affects non-datafusion code paths.

This workaround at least allows us to move forward and release a v0.22.1 while continuing to hunt down the issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ debug = "line-tables-only"
 
 [workspace.dependencies]
 delta_kernel = { version = "0.4.1", features = ["sync-engine"] }
-# delta_kernel = { path = "../delta-kernel-rs/kernel", version = "0.3.0" }
+#delta_kernel = { path = "../delta-kernel-rs/kernel", features = ["sync-engine"]  }
 
 # arrow
 arrow = { version = "53" }
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-core"
-version = "0.22.0"
+version = "0.22.1"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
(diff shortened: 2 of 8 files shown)


ion-elgreco (approved):

ion-elgreco:
@rtyler did you file an issue in arrow-rs?

rtyler:
@ion-elgreco my suspicions are that it lies upstream, but I don't have any confirmation

Issue #2613, opened 2024-06-20
Support string and binary view types.


ritchie46:
Arrow has adopted the (IMO) much better binary and string view types. Supporting these would mean we could move data zero-copy to delta-rs. Currently it fails:

```python
df = pl.DataFrame({"a": ["foo", "bar"]})
df.write_delta("/tmp/bar.bar", mode="overwrite")
```

```
---------------------------------------------------------------------------
ArrowNotImplementedError                  Traceback (most recent call last)
Cell In[5], line 2
      1 df = pl.DataFrame({"a": ["foo", "bar"]})
----> 2 df.write_delta("/tmp/bar.bar", mode="overwrite")

File ~/code/polars/py-polars/polars/dataframe/frame.py:3968, in DataFrame.write_delta(self, target, mode, overwrite_schema, storage_options, delta_write_options, delta_merge_options)
   3965     delta_write_options["schema_mode"] = "overwrite"
   3967 schema = delta_write_options.pop("schema", None)
-> 3968 write_deltalake(
   3969     table_or_uri=target,
   3970     data=data,
   3971     schema=schema,
   3972     mode=mode,
   3973     storage_options=storage_options,
   3974     large_dtypes=True,
   3975     **delta_write_options,
   3976 )
   3977 return None

File ~/miniconda3/lib/python3.10/site-packages/deltalake/writer.py:519, in write_deltalake(table_or_uri, data, schema, partition_by, mode, file_options, max_partitions, max_open_files, max_rows_per_file, min_rows_per_group, max_rows_per_group, name, description, configuration, overwrite_schema, schema_mode, storage_options, partition_filters, predicate, large_dtypes, engine, writer_properties, custom_metadata)
    514 else:
    515     file_options = ds.ParquetFileFormat().make_write_options(
    516         use_compliant_nested_type=False
    517     )
--> 519 ds.write_dataset(
    520     data,
    521     base_dir="/",
    522     basename_template=f"{current_version + 1}-{uuid.uuid4()}-{{i}}.parquet",
    523     format="parquet",
    524     partitioning=partitioning,
    525     # It will not accept a schema if using a RBR
    526     schema=schema if not isinstance(data, RecordBatchReader) else None,
    527     file_visitor=visitor,
    528     existing_data_behavior="overwrite_or_ignore",
    529     file_options=file_options,
    530     max_open_files=max_open_files,
    531     max_rows_per_file=max_rows_per_file,
[...]

PR #3036, opened 2024-11-29
chore: bump kernel

ion-elgreco:
# Description
With this change: https://github.com/delta-io/delta-kernel-rs/pull/533 and this change:  https://github.com/delta-io/delta-rs/pull/2738, we fully support view types.

I'll make a change also upstream in polars to change the arrow compat level to latest

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/2613

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -542,6 +542,7 @@ impl<'a> DeltaScanBuilder<'a> {
             .filter
             .map(|expr| context.create_physical_expr(expr, &df_schema).unwrap());
 
+        // dbg!(logical_filter.clone());
         // Perform Pruning of files to scan
         let (files, files_scanned, files_pruned) = match self.files {
             Some(files) => {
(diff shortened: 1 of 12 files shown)


rtyler (approved):

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -542,6 +542,7 @@ impl<'a> DeltaScanBuilder<'a> {
             .filter
             .map(|expr| context.create_physical_expr(expr, &df_schema).unwrap());
 
+        // dbg!(logical_filter.clone());
```
hntd187, line 545: We prolly don't need these now ?

ion-elgreco: I was debugging a regression caused by the kernel bump. There is something wrong with the statistics being passed to Datafusion

crates/core/src/kernel/snapshot/log_data.rs
```diff
@@ -803,16 +810,16 @@ mod datafusion {
             lazy_static::lazy_static! {
                 static ref ROW_COUNTS_EVAL: Arc<dyn ExpressionEvaluator> =  ARROW_HANDLER.get_evaluator(
                     crate::kernel::models::fields::log_schema_ref().clone(),
-                    Expression::column("add.stats_parsed.numRecords"),
+                    Expression::column(vec!["add.stats_parsed.numRecords"].into_iter()),
```
roeap, line 813: I think this may be the cause for the failing DF tests.

```suggestion
                    Expression::column(["add", "stats_parsed", "numRecords"]),
```

roeap (commented):
i think there is just a small oversight, that causes the DF failures.

PR #3040, opened 2024-12-01
feat: add out_of_range flag to load_cdf

### LINKED ISSUE #3035: load_cdf return an empty dataframe when a version is out of range
# Description

In spark delta table you can enable an option to manage out of range versions or timestamps.  https://docs.delta.io/latest/delta-change-data-feed.html#read-changes-in-streaming-queries
image

Right now the behaviour of load_cdf is inconsistent, if you provide an out of range version you get an error:
image

But with a timestamp out of range, you get an empty dataset:
image

It would be useful for incremental pipelines to have a way to manage this behaviour and make it consistent.

pblocz:
# Description

This PR changes load_cdf so:

By default
- load_cdf with starting_version out of range returns an InvalidVersion
- load_cdf with a starting_timestamp out of range returns a ChangeDataTimestampGreaterThanCommit

And you can use `.with_allow_out_of_range()` in rust or `allow_out_of_range=True` in python to make it so that both cases return an empty dataframe.

This aligns it closer with how reading the change feed works in [spark](https://docs.delta.io/latest/delta-change-data-feed.html#read-changes-in-streaming-queries) and makes the behaviour consistent.

This in contrast to the  current behavior:
- load_cdf with starting_version out of range returns an InvalidVersion
- load_cdf with a starting_timestamp out of range returns an empty dataframe

# Related Issue(s)
- closes #3035


python/tests/test_cdf.py
```diff
@@ -677,3 +677,40 @@ def test_write_overwrite_partitioned_cdf(tmp_path, sample_data: pa.Table):
...
     ).sort_by(sort_values)
+
+
+def test_read_cdf_version_out_of_range():
+    dt = DeltaTable("../crates/test/tests/data/cdf-table/")
+
+    try:
+        b = dt.load_cdf(4).read_all().to_pydict()
+        assert False, "Should not get here"
+    except _internal.DeltaError as e:
```
ion-elgreco, line 688: You can do deltalake.exceptions actually

python/tests/test_cdf.py
```diff
@@ -677,3 +677,40 @@ def test_write_overwrite_partitioned_cdf(tmp_path, sample_data: pa.Table):
...
+    b = dt.load_cdf(4, enable_out_of_range=True).read_all()
+
+    assert len(b) == 0
+
+
+def test_read_timestamp_cdf_out_of_range():
+    dt = DeltaTable("../crates/test/tests/data/cdf-table/")
+    start = "2033-12-22T17:10:21.675Z"
+
+    try:
```
ion-elgreco, line 703: You should use pytest.raises

ion-elgreco (commented):
Looks good! Just couple comments

crates/core/src/operations/load_cdf.rs
```diff
@@ -139,8 +151,13 @@ impl CdfLoadBuilder {
                 .log_store
                 .read_commit_entry(version)
                 .await?
-                .ok_or(DeltaTableError::InvalidVersion(version))?;
-            let version_actions = get_actions(version, snapshot_bytes).await?;
+                .ok_or(DeltaTableError::InvalidVersion(version));
+
+            if snapshot_bytes.is_err() && version >= end && self.enable_out_of_range {
```
hntd187, line 156: Do you need this? You adjusted the end version above, so it should never exceed the last version right?

crates/core/src/operations/load_cdf.rs
```diff
@@ -236,6 +253,11 @@ impl CdfLoadBuilder {
             }
         }
 
+        // All versions were skipped due to date our of range
```
hntd187, line 256: Typo in comment: "out of range"

crates/core/src/operations/load_cdf.rs
```diff
@@ -236,6 +253,11 @@ impl CdfLoadBuilder {
...
+        if !self.enable_out_of_range && change_files.is_empty() && add_files.is_empty() && remove_files.is_empty() {
+            return Err(DeltaTableError::ChangeDataTimestampGreaterThanCommit { ending_timestamp: ending_timestamp });
```
hntd187, line 258: So according to the link above too, this is only thrown if the timestamp exceeds the last commit timestamp and out of range is not enabled. You should be able to move this up outside the loop to check immediately, I don't think it matters if the actions are empty in this case.

crates/core/src/operations/load_cdf.rs
pblocz: From my checks if the start that you provide to `self.log_store.get_latest_version(start)` is greater than the last commit, then it returns that same start, so version could exceed the end of log.

I think I could adjust the call to do  `self.log_store.get_latest_version(0)` so it always start at the beginning and make sure I get the real end, but did not want to change the code too much.

python/tests/test_cdf.py
pblocz: I have changed it to use pytest.raises

python/tests/test_cdf.py
pblocz: I have changed it to use the exceptions module

crates/core/src/operations/load_cdf.rs
pblocz: I have made this change, use ` self.log_store.get_latest_version(0)` to read the actual last version and use it in the check

hntd187 (approved):
This looks good to me now, just rerunning CI right now local concurrency test is notoriously flakey.

pblocz:
Awesome, seems that all the tests passed now

hntd187:
@pblocz if you could sign off on your commits we can get this merged soon

pblocz:
@hntd187 Done. I don't know why, but thought the sign off was something else 😅

PR #3044, opened 2024-12-02
feat: move unity catalog integration to it's own crate

hntd187:
# Description
This moves the catalog integration of unity catalog to it's own crate. It also replaces the mock server with the httpmock crate and the custom retry logic with request_middleware's retry middleware. I think this will be easier to manage as a client and I do not believe many people were using this integration anyways.

# Related Issue(s)

# Documentation


crates/catalog-unity/src/client/retry.rs
```diff
@@ -0,0 +1,124 @@
...
+            _ => Self::new(ErrorKind::Other, err),
+        }
+    }
+}
+
+// impl From<RetryError> for reqwest::Error {
+//     fn from(value: RetryError) -> Self {
+//         Into::into(value)
+//     }
+// }
```
rtyler, line 64: :question:

crates/catalog-unity/src/credential.rs
```diff
@@ -415,39 +410,32 @@ impl TokenCredential for ImdsManagedIdentityOAuthProvider {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::data_catalog::client::mock_server::MockServer;
-    use futures::executor::block_on;
-    use hyper::body::to_bytes;
-    use hyper::{Body, Response};
-    use reqwest::{Client, Method};
+
+    // use hyper::{ Response};
```
rtyler, line 414:
```suggestion
```

crates/catalog-unity/src/credential.rs
```diff
@@ -415,39 +410,32 @@ impl TokenCredential for ImdsManagedIdentityOAuthProvider {
...
-                .to_str()
-                .unwrap();
-            assert_eq!(t, "env-secret");
-            let t = req.headers().get("metadata").unwrap().to_str().unwrap();
-            assert_eq!(t, "true");
-            Response::new(Body::from(
-                r#"
+        let client = reqwest_middleware::ClientBuilder::new(Client::new()).build();
+
+        let _retry_config = RetryConfig::default();
```
rtyler, line 428: What's this for?

crates/catalog-unity/src/credential.rs
```diff
@@ -458,45 +446,41 @@ mod tests {
...
-        server.push_fn(move |req| {
-            assert_eq!(req.uri().path(), format!("/{tenant}/oauth2/v2.0/token"));
-            assert_eq!(req.method(), &Method::POST);
-            let body = block_on(to_bytes(req.into_body())).unwrap();
-            let body = String::from_utf8(body.to_vec()).unwrap();
-            assert!(body.contains("federated-token"));
-            Response::new(Body::from(
-                r#"
+        let client = reqwest_middleware::ClientBuilder::new(Client::new()).build();
+        let _retry_config = RetryConfig::default();
```
rtyler, line 474: Another unused variable that should be removed? :thinking:

crates/core/Cargo.toml
```diff
@@ -81,7 +81,7 @@ dashmap = "6"
 errno = "0.3"
 either = "1.8"
 fix-hidden-lifetime-bug = "0.2"
-hyper = { version = "0.14", optional = true }
+#hyper = { version = "0.14", optional = true }
```
rtyler, line 84:
```suggestion
```

crates/core/Cargo.toml
```diff
@@ -97,21 +97,22 @@ tracing = { workspace = true }
...
+reqwest-middleware = { version = "0.4.0", optional = true }
 
 [dev-dependencies]
 criterion = "0.5"
 ctor = "0"
 deltalake-test = { path = "../test", features = ["datafusion"] }
 dotenvy = "0"
 fs_extra = "1.2.0"
-hyper = { version = "0.14", features = ["server"] }
+#hyper = { version = "0.14", features = ["server"] }
```
rtyler, line 115:
```suggestion
```

crates/core/Cargo.toml
```diff
@@ -137,4 +138,4 @@ datafusion = [
 datafusion-ext = ["datafusion"]
 json = ["parquet/json"]
 python = ["arrow/pyarrow"]
-unity-experimental = ["reqwest", "hyper"]
+unity-experimental = ["reqwest", "reqwest-middleware"]
```
rtyler, line 141: Shouldn't this feature come off the core crate now that this code is moved into its own crate?

crates/catalog-unity/src/credential.rs
hntd187: I wasn't sure if I was going to remove it at that point from the testing, so it's a side effect left over from that.

crates/core/Cargo.toml
hntd187: So this is a discussion point I wanted to talk about. The main trait for the catalog and the errors remained in the core crate. So core can build I assume generic catalog support and catalog crates just depend on core and implement. At first glance I kinda didn't have a cohesive error handling thing here, but I think thinking about it. There should be a generic catalog error in core, but have no conditionals features on it. And individual crates implement a conversion to that error type. Sound good?

crates/catalog-unity/src/lib.rs
```diff
@@ -616,10 +631,11 @@ impl UnityCatalog {
...
             ))
             .header(AUTHORIZATION, token)
-            .send_retry(&self.retry_config)
-            .await?;
+            .send()
+            .await
+            .map_err(UnityCatalogError::from)?;
 
-        Ok(resp.json().await?)
+        Ok(resp.json().await.map_err(UnityCatalogError::from)?)
```
roeap, line 638: if we have an `impl From<...>` for the error, why do we need the map? does the function return a different kind of err.

crates/core/src/operations/merge/filter.rs
```diff
@@ -252,16 +252,13 @@ pub(crate) fn generalize_filter(
             }
         }
         Expr::InList(in_list) => {
-            let compare_expr = match generalize_filter(
+            let compare_expr = generalize_filter(
```
ion-elgreco, line 255: Why is this changed?

crates/gcp/tests/context.rs
```diff
@@ -76,10 +76,12 @@ impl StorageIntegration for GcpIntegration {
         let account_path = self.temp_dir.path().join("gcs.json");
         info!("account_path: {account_path:?}");
         std::fs::write(&account_path, serde_json::to_vec(&token).unwrap()).unwrap();
-        std::env::set_var(
-            "GOOGLE_SERVICE_ACCOUNT",
-            account_path.as_path().to_str().unwrap(),
-        );
+        unsafe {
```
roeap, line 79: Am I missing something, or did we remove `unsafe` around `set_var` further above, specifically in `crates/aws/src/storage.rs`, and here we are adding it?

roeap (commented):
Overall looks good, just some minor questions.

More gene4rally speaking though, I think we need to make some more deliberate choices around how we want to integrate with catalogs.

AS I have been out for a while, I am not sure where we are in terms of eagerness to push the next release. I would hope that we can do some more cleaning and most importantly testing before we release get "more serious" with unity ...

w.r.t. testing I hope that we can even do some integration tests using OSS UC. Happy the help with that.

crates/gcp/tests/context.rs
hntd187: I have to add it back to the other one, thanks for pointing it out

crates/core/src/operations/merge/filter.rs
hntd187: I fixed a few warnings along the way, this statement could be just written with an `?` operator instead.

crates/catalog-unity/src/lib.rs
hntd187: This is a bit of jank, the base error type in core no longer has a from for rewqest::Error so we convert to our own error type that does have a From to DataCatalogError

roeap (dismissed):
LGTM!

One ore thing though, and not sure if this would actually happen, but it seems GH wants to re-apply a bunch of commits already on main. If all of these were duplicated, this would give us (IMO) a fairly dirty history. Could we squash these to make it a bit cleaner?

hntd187:
@roeap yes, I'll squash prior to merge

rtyler (approved):

rtyler:
@hntd187 unfortunately it looks like this needs to be rebased again :facepalm:

rtyler:
I have resolved conflicts, I will push and land this before I cut another release today

Issue #3049, opened 2024-12-09
Write operations with without_files lead to data loss due to checkpointing


JonasDev1:
# Environment

**Delta-rs version**: 21.

**Binding**: Rust

***
# Bug

**What happened**:
I am appending to a delta table that is read with without_files to reduce memory usage. The append process works fine, but during automatic checkpoint creation, older commited files are removed because they are not in the table snapshot and are therefore not part of the checkpoint.

**What you expected to happen**:
Either an error should be displayed or the automatic checkpoint creation should be deactivated

ion-elgreco:
Yeahhh, forgot about that interaction with the code base

PR #3059, opened 2024-12-15
fix: raise during checkpoint creating with without_files

ion-elgreco:
# Description
- If explicitly creating checkpoint, we raise
- If checkpoint creation called in post_commit_hook, we warn

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/3049

# Documentation


rtyler (approved):
this is silly

PR #3061, opened 2024-12-16
chore: bump versions for next release(s)

rtyler:
This change also removes `uv.lock`. From the [documentation](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile) 

> Unlike the pyproject.toml, which is used to specify the broad requirements of your project, the lockfile contains the exact resolved versions that are installed in the project environment. This file should be checked into version control, allowing for consistent and reproducible installations across machines.
> 
> A lockfile ensures that developers working on the project are using a consistent set of package versions. Additionally, it ensures when deploying the project as an application that the exact set of used package versions is known.

The problem with the presence of this `uv.lock` for purposes of development of a library like `deltalake` is that `pip` installations of the released package will _not_ have the contents of that `uv.lock` file. 

Therefore I want it to be removed so that our local installations, and CI, are always using `pyproject.toml` to resolve their dependencies in a manner which is consistent with the users of the installed package.


ion-elgreco:
I am fine with removing it, but we should remove the checkout of the lockfile in the CI now, so that it can pass

PR #3067, opened 2024-12-18
fix: allow more than 15 concurrent transactions to have been committed

### LINKED ISSUE #3066: TransactionError::VersionAlreadyExists needs exponential backoff with commit peeking
# Environment

**Delta-rs version**: irrelephant :elephant: 
**Binding**: both

***
# Bug

**What happened**:

For systems with high write throughput, and therefore version contention, conflicting writes can result in `VersionAlreadyExists` being raised. The retry logic in the `PreparedCommit` operation immediately and blindly retries the operation with state_version +1 rather than peeking at the next commit version from the table.

 This can result in logs such as:

```
LockClientError::VersionAlreadyExists(47804)
LockClientError::VersionAlreadyExists(47805)
LockClientError::VersionAlreadyExists(47806)
LockClientError::VersionAlreadyExists(47807)
LockClientError::VersionAlreadyExists(47808)
LockClientError::VersionAlreadyExists(47809)
LockClientError::VersionAlreadyExists(47810)
LockClientError::VersionAlreadyExists(47811)
LockClientError::VersionAlreadyExists(47812)
LockClientError::VersionAlreadyExists(47813)
LockClientError::VersionAlreadyExists(47814)
LockClientError::VersionAlreadyExists(47815)
Successfully flushed v47816 to Delta table 
```

Basically there are two problems:

* The retry logic is a `while true {}` without any pause/sleep in between retries, so it's very easy for the retry to hit the max retry limit (hardcoded to 15 for AWS)
* The retry logic foolishly assumes that the version of the currently loaded state is close enough to the latest version and attempts to create a version number of ` loaded + 1` which is almost 100% guaranteed to be wrong if there is more than one concurrent writing process. This causes additional unnecessary retries, since the aforementioned `while true` loop will move the attempted version number up by 1 from the _loaded state_ rather than peeking at the latest commit in the table and attempting a commit with `peeked_version + 1`

I would expect the retry to take `peeked_version + 1` at a bare minimum!

**How to reproduce it**:

Use more than one concurrent writer :laughing: 

**More details**:

rtyler:
There are high-concurrency scenarios where multiple writers may have committed versions to the log since the current process' state was loaded. Because of the `snapshot.version() + 1` and retry logic in the transaction commit operation, this would result in failures if more than 15 versions had been committed since the table state had been loaded.

This approach fetches the latest known version on the table prior to committing the transaction, and if the latest version has shifted, it will perform a conflict check with the versions that have since been committed.

Fixes #3066


rtyler:
**Note:** This pull request is rebased on #3069 so that the CI would pass. I would expect the other pull request to be merged first

ion-elgreco (approved):

ion-elgreco:
Looks good! 

Just a question, the issue also mentions to also introduce exponential backoff. But I don't see that being implemented here, are you addressing that in another PR?

PR #3068, opened 2024-12-18
Use projected_table_schema for projection in DeltaSchemaAdapter

jkylling:
# Description

After upgrading from deltalake 0.20.1 to 0.22.3 it looks like Parquet column projection is  broken when using DeltaTable::scan. Instead of scanning only the a single column, it looks like all columns are fetched from storage.

Inspection with a debugger reveals that the adapted_projections are wrong here:
https://github.com/apache/datafusion/blob/88f58bf929167c5c5e2250ad87caa88d4dff11e5/datafusion/core/src/datasource/physical_plan/parquet/opener.rs#L153-L159 The adapted_projections are obtained in
https://github.com/delta-io/delta-rs/blob/5b2f46b06e0eb508f932a8b39feb11b568a78a32/crates/core/src/delta_datafusion/schema_adapter.rs#L46-L60 Changing line 49 to use the projected_table_schema seems to solve the problem.

Opening this to see if we have test coverage.


jkylling:
For reference, the corresponding code in Datafusion 42 is https://github.com/apache/datafusion/blob/a43ce8bd67e8d649e0e4f5260f8a5e6e10d62dbc/datafusion/core/src/datasource/physical_plan/parquet/opener.rs#L80-L82 When upgrading from 42 to 43 the meaning of `table_schema` was changed from being the projected schema, to be the full table schema, and a new argument `projected_schema` was added. This was mapped to `projected_table_schema` and `table_schema` in our code, but we forgot to update all uses of `table_schema` in the `DeltaSchemaAdapter` to use the `projected_table_schema`.

The changes to the SchemaAdapter API were introduced in https://github.com/apache/datafusion/pull/12135

rtyler (approved):
I will have this merged one `main` goes green again

PR #3073, opened 2024-12-21
chore: api changes for Datafusion 44

rtyler:
This is currently a work in progress to help pre-test Datafusion 44 before it's released

PR #3074, opened 2024-12-21
Fix compilation after update to DataFusion 44

alamb:
Proposed PR that targets @rtyler 's PR https://github.com/delta-io/delta-rs/pull/3073 to update to latest DataFusion and fixes compilation


alamb:
Actually, I think the issue may be the wrong pin. Closing this change

PR #3077, opened 2024-12-22
chore: Reduce compiler warnings by updating to use non deprecated DataFusion APIs

alamb:
- targets https://github.com/delta-io/delta-rs/pull/3073 from @rtyler 

This PR makes some small changes to reduce the build warnings by using non deprecated APIs

### DIFF
--- a/crates/catalog-unity/src/error.rs
+++ b/crates/catalog-unity/src/error.rs
@@ -10,7 +10,6 @@ pub enum UnityCatalogError {
     },
 
     /// A generic error qualified in the message
-
     #[error("{source}")]
     Retry {
         /// Error message
@@ -19,7 +18,6 @@ pub enum UnityCatalogError {
     },
 
     #[error("Request error: {source}")]
-
     /// Error from reqwest library
     RequestError {
         /// The underlying reqwest_middleware::Error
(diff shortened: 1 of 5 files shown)


crates/test/src/datafusion.rs
```diff
@@ -1,18 +1,10 @@
...
 
 pub fn context_with_delta_table_factory() -> SessionContext {
-    let cfg = RuntimeConfig::new();
-    let env = RuntimeEnv::try_new(cfg).unwrap();
-    let ses = SessionConfig::new();
-    let mut state = SessionStateBuilder::new()
-        .with_config(ses)
-        .with_runtime_env(Arc::new(env))
-        .build();
+    let mut state = SessionStateBuilder::new().build();
```
alamb, line 7: This was just using the default config and runtime env which the builder already does

crates/catalog-unity/src/error.rs
```diff
@@ -10,7 +10,6 @@ pub enum UnityCatalogError {
     },
 
     /// A generic error qualified in the message
-
```
alamb, line 13: Drive by clean up -- clippy was complaining about blank lines between comments

PR #3080, opened 2024-12-22
fix: use of deprecated DeltaTable::peek_next_commit

jkylling:
Maybe we could run CI with
```
cargo clippy ... -- -D warnings
```
to catch this early?


ion-elgreco (approved):

roeap:
Good idea! Feel free to update this :).

Right now I am looking into how we can more deeply integrate with delta-kernel-rs, and APIs like these make this somewhat hard.

So interested in your use case for using this API at all :).

For context, over simplified, there are two kinds of users - the ones who only want the right data and do not care about the log at all, and the ones who care almost exclusively about the log...

In this case, would the CDF APIs also work for you?

PR #3085, opened 2024-12-25
Add steps to install uv in venv

FranArenas:
# Description
uv is needed to use the make file for development, but uv is not installed in the venv and there is not a command to activate the venv. 

This PR adds the sh command to activate the venv and the command to install uv before executing the Makefile.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ If you want to claim an issue to work on, you can write the word `take` as a com
     python -m venv .venv
     ```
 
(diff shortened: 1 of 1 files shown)


ion-elgreco:
UV should actually be installed globally, I would just put a link to the UV installation for this

FranArenas:
> UV should actually be installed globally, I would just put a link to the UV installation for this

Makes sense, I just added the link to the uv repo and warned about the need of the tool.

ion-elgreco (approved):

PR #3087, opened 2024-12-28
chore: bump kernel 0.6 and datafusion 44

ion-elgreco:


rtyler (approved):
Looking good to me, I'm pleasantly surprised that this Datafusion upgrade doesn't take an arrow upgrade with it :people_hugging:

alamb:
> Looking good to me, I'm pleasantly surprised that this Datafusion upgrade doesn't take an arrow upgrade with it 🫂

The next one will, however.

We expect to release another minor (non API breaking) arrow release  later this month if you want to propose anything for inclusion:
- https://github.com/apache/arrow-rs/issues/6887

Issue #2843, opened 2024-09-03
Utilize Amazon S3 condition write to support concurrent write


Cpaulyz:
# Description

Hi, I noticed that Amazon S3 supports condition write now (https://aws.amazon.com/about-aws/whats-new/2024/08/amazon-s3-conditional-writes/). Is it possible to utilize this feature to replace dynamodb-lock to support concurrent write natively?

thomasfrederikhoeck:
I guess the following upstream needs to be closed first: https://github.com/apache/arrow-rs/issues/6285

danielgafni:
Just to confirm, this **won't** allow _parallel_ writing, but only remove the dependency on DynamoDB as locking mechanism, right?

Cpaulyz:
> Just to confirm, this **won't** allow _parallel_ writing (for example, writing 100 partitions at once), but will remove the dependency on DynamoDB as locking mechanism, right?

Yes. I think so.

rtyler:
If you're up for doing some experimentation (on non-production workloads), I _believe_ that our [conditional put support](https://delta-io.github.io/delta-rs/usage/writing/writing-to-s3-with-locking-provider/#enabling-concurrent-writes-for-alternative-clients) will "just work" for S3.

satyanmandavilli:
[https://github.com/apache/arrow-rs/issues/6285](url) is closed now. Do we have any update on this issue? Looking forward to remove DynamoDB dependency.

gianghq:
Keen on this as well.

ion-elgreco:
> [apache/arrow-rs#6285](https://github.com/apache/arrow-rs/issues/6285) is closed now. Do we have any update on this issue? Looking forward to remove DynamoDB dependency.

You will have to wait on the next arrow release that aligns with datafusion and object store. Then there are still bugs in datafusion that stops us from actually upgrading for some time.

PR #3091, opened 2024-12-29
refactor: make put-if-absent default for S3 stores

ion-elgreco:
# Description
- closes https://github.com/delta-io/delta-rs/issues/2843

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ thiserror = { version = "2" }
 url = { version = "2" }
 urlencoding = "2.1.3"
 uuid = { version = "1" }
+path-tree = { version = "0.8.1"} # pin to 0.8.1 due to nightly features
 
 # runtime / async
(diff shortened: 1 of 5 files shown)


rtyler:
This is a good change overall and I don't _think_ it should be a probably to slyly slip this in as a default when the AWS_LOCKING_PROVIDER environment variable is not set.

In my setups for example I require the S3DynamoDbLogStore for concurrent coordination with Spark writers, so I will want to do a little more testing with this to make sure that the right log store is chosen in the right circumstances

rtyler (approved):
I've reviewed and updated this one :+1:

ion-elgreco:
> @ion-elgreco hey, the relevant [page](https://delta-io.github.io/delta-rs/usage/writing/writing-to-s3-with-locking-provider/) in the docs is simply out-of-date, right? seems like there should be no additional locking mechanism necessary with S3, just checking...

Correct!

PR #3092, opened 2024-12-29
chore: pin path tree

ion-elgreco:
# Description
Should resolve the build failures we see atm @roeap

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,3 +75,6 @@ async-trait = { version = "0.1" }
 futures = { version = "0.3" }
 tokio = { version = "1" }
 num_cpus = { version = "1" }
+
+[patch.crates-io]
+path-tree = { git = "https://github.com/viz-rs/path-tree.git", rev = "03f85c289cea8a3ae9ee7bcadba5633fb465b8da"} # pin to 0.8.1 due to nightly features


Cargo.toml
```diff
@@ -75,3 +75,6 @@ async-trait = { version = "0.1" }
 futures = { version = "0.3" }
 tokio = { version = "1" }
 num_cpus = { version = "1" }
+
+[patch.crates-io]
+path-tree = { git = "https://github.com/viz-rs/path-tree.git", rev = "03f85c289cea8a3ae9ee7bcadba5633fb465b8da"} # pin to 0.8.1 due to nightly features
```
rtyler, line 80: Is this a dev only dependency? We cannot release with non-crates.io released packages in the manifests

ion-elgreco: I have no clue, I couldn't see it in the dependency tree. Might only appear during tests

fundon: You can upgrade MSRV.

rtyler: @fundon that may be correct however I have been quite resistant to such rapid MSRV version upgrades for us in the past. For example we just had to bump to 1.81, and there's some rumblings about perhaps 1.82 which was only released about 2.5 months ago. An MSRV upgrade does not come with zero cost for our own build infrastructure nor for our users, so I try to make sure we're upgrading for "pretty good reasons".

In this case, i actually haven't seen the failure that @ion-elgreco was seeing when he looked at this. :see_no_evil:

fundon: Or lock version: `=0.8.1`, see https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility

ion-elgreco: > Or lock version: `=0.8.1`, see https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility

I did that actually but it still pulled in 0.8.2

fundon: Sorry, I caused you some trouble. `0.8.2` has been yanked. No changes required.

rtyler: Hah, I just saw the failure in a GitHub action, and also just noticed that @fundon you are the author of path-tree, and I _also_ noticed that 0.8.2 is yanked :laughing: 

grafik


@fundon I'm happy to help test out changes on older rusts if you want to give 0.8.3 a try, I can be found at rtyler on the [delta-users slack](https://delta.io) or as `rtyler` on IRC (libera.chat)

ion-elgreco: > Sorry, I caused you some trouble. `0.8.2` has been yanked. No changes required.

No worries! Thanks for your help!

PR #3101, opened 2025-01-04
feat: allow multiple Python threads to work with a single DeltaTable instance

### LINKED ISSUE #2958: Python DeltaTable does not support writes in multiple threads
# Environment

**Delta-rs version**: 0.20.1

**Binding**: python

**Environment**:
- **Cloud provider**:
- **OS**: macOS
- **Other**:

***
# Bug

**What happened**:

Using the same `DeltaTable` across multiple threads in `write_deltalake` causes the error
```
RuntimeError: Already borrowed
```

This is not an issue if we use the table URI directly.

Writes should work in a multithreaded environment.

**How to reproduce it**:

```python
from concurrent.futures import ThreadPoolExecutor
import polars as pl

path = 'some local path'
table = pl.DataFrame({'a': [1,2,3]}).to_arrow()
write_deltalake(path, table)

dt = DeltaTable(path)

with ThreadPoolExecutor() as exe:
    list(exe.map(lambda _: write_deltalake(dt, table, mode='append'), range(2)))
```
produces
```
  File "<stdin>", line 2, in <module>
  File "/Users/peter.ke/.pyenv/versions/3.10.13/lib/python3.10/concurrent/futures/_base.py", line 621, in result_iterator
    yield _result_or_cancel(fs.pop())
  File "/Users/peter.ke/.pyenv/versions/3.10.13/lib/python3.10/concurrent/futures/_base.py", line 319, in _result_or_cancel
    return fut.result(timeout)
  File "/Users/peter.ke/.pyenv/versions/3.10.13/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/Users/peter.ke/.pyenv/versions/3.10.13/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/Users/peter.ke/.pyenv/versions/3.10.13/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "<stdin>", line 2, in <lambda>
  File "/Users/peter.ke/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deltalake/writer.py", line 302, in write_deltalake
    table.update_incremental()
  File "/Users/peter.ke/.pyenv/versions/3.10.13/lib/python3.10/site-packages/deltalake/table.py", line 1258, in update_incremental
RuntimeError: Already borrowed
```

**More details**:

Cause of error:
- `write_deltalake` calls `write_to_deltalake` from rust to write, which borrows the `RawDeltaTable` immutably while releasing the GIL. The immutable borrow means theoretically writes in multiple threads should be fine
- However, `write_deltalake` also calls `table.update_incremental()`, which borrows the `RawDeltaTable` mutably
- Since GIL was released in `write_to_deltalake`, another thread running `write_deltalake` will try to mutably borrow the previously immutably borrowed `RawDeltaTable` and fails

From my brief investigation, I see two potential solutions:
- The Rust API exposed to python could benefit from being refactored to be more immutable so it fits better in a multi-threaded environment. e.g. on the rust side, `update_incremental` can return a new `DeltaTable` rather than mutating the existing one- similarly for other internals like `DeltaTableState` or `Snapshot`.
 
  This should be similar performance/memory wise, since cloning a `Snapshot` seems relatively cheap given that the bulk `files` uses arrow. However, this looks like a large-ish refactor because a lot of the internals expose mutable APIs.

- Remove `update_incremental` from `write_deltalake` (or add a flag to disable it). We'd just need to ensure `write_to_deltalake` returns an updated `RawDeltaTable` so `table._table` can kept updated.

rtyler:
This change introduces an internal Mutex inside of RawDeltaTable which allows the PyO3 bindings to share the Python object between threads at the Python layer.

PyO3 will raise a `RuntimeError: Already borrowed` for any function call which takes a mutable reference to `self`. Introducing the internal Mutex ensures that all function signatures can operate with just self-references safely.

The Rust-level Mutex is a simple passthrough for most operations which do not need to modify the underlying state. The critical sections which typically need to acquire and mutate with a lock are after I/O bound operations are completed as far as I can tell, so I don't anticipate deadlock or performance issues.

There is still some cleanup of errors that needs to happen to make the code here more ergonomic when blending DeltaError with PoisonError from the lock, as such right now there's a lot of ugly error mapping.

Fixes #2958

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -10,7 +10,7 @@ use std::collections::{HashMap, HashSet};
 use std::ffi::CString;
 use std::future::IntoFuture;
 use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time;
 use std::time::{SystemTime, UNIX_EPOCH};
 
@@ -31,6 +31,7 @@ use deltalake::errors::DeltaTableError;
 use deltalake::kernel::{
     scalars::ScalarExt, Action, Add, Invariant, LogicalFile, Remove, StructType, Transaction,
 };
+use deltalake::logstore::LogStoreRef;
 use deltalake::operations::add_column::AddColumnBuilder;
 use deltalake::operations::add_feature::AddTableFeatureBuilder;
 use deltalake::operations::collect_sendable_stream;
@@ -53,11 +54,13 @@ use deltalake::parquet::errors::ParquetError;
 use deltalake::parquet::file::properties::WriterProperties;
 use deltalake::partitions::PartitionFilter;
 use deltalake::protocol::{DeltaOperation, SaveMode};
-use deltalake::storage::IORuntime;
+use deltalake::storage::{IORuntime, ObjectStoreRef};
+use deltalake::table::state::DeltaTableState;
 use deltalake::DeltaTableBuilder;
 use deltalake::{DeltaOps, DeltaResult};
 use error::DeltaError;
 use futures::future::join_all;
+use tracing::log::*;
 
 use pyo3::exceptions::{PyRuntimeError, PyValueError};
 use pyo3::prelude::*;
@@ -82,7 +85,9 @@ enum PartitionFilterValue {
 
 #[pyclass(module = "deltalake._internal")]
 struct RawDeltaTable {
-    _table: deltalake::DeltaTable,
+    /// The internal reference to the table is guarded by a Mutex to allow for re-using the same
+    /// [DeltaTable] instance across multiple Python threads
+    _table: Arc<Mutex<deltalake::DeltaTable>>,
     // storing the config additionally on the table helps us make pickling work.
     _config: FsConfig,
 }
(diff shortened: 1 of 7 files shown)


python/src/lib.rs
```diff
@@ -362,12 +472,15 @@ impl RawDeltaTable {
...
-                self._table.snapshot().map_err(PythonError::from)?.clone(),
-            )
-            .with_enforce_retention_duration(enforce_retention_duration)
-            .with_dry_run(dry_run);
+            let snapshot = match self._table.lock() {
+                Ok(table) => Ok(table.snapshot().cloned().map_err(PythonError::from)?),
+                Err(e) => Err(e),
+            }
+            .expect("SHOOT!");
+            // TODO: XXX
```
roeap, line 480: artifact?

python/src/lib.rs
```diff
@@ -894,17 +1007,22 @@ impl RawDeltaTable {
...
+                .map_err(PythonError::from)
+                .map_err(PyErr::from)
+        })?;
+        let metadata = self.with_table(|t| {
+            t.metadata()
+                .cloned()
+                .map_err(PythonError::from)
+                .map_err(PyErr::from)
+        })?;
+        //.map_err(|_| DeltaProtocolError::new_err("table does not yet have a schema"))?
```
roeap, line 1022: can we remove this?

roeap (commented):
looking good!

just some very minor cleanup that you may want to consider?

python/src/lib.rs
rtyler: d'oh! good catch, I did miss some of this junk in my diff review before poooshing

roeap (approved):
👍

PR #3103, opened 2025-01-06
feat: lakefs integration

ion-elgreco:
# Description
Makes LakeFS first-class citizen in delta-rs. Operations are executed in a transaction branch, where we write into and then try to commit and merge if there is no conflict. If there is a conflict it means there was a concurrent writer, so we remove the log, and retry with the conflict checker.

Main concepts that this PR introduces through out the core crate:
- LogStore makes distinction between reading and writing objectstore, the writing store can receive an optional `operation_id`
- Every operation has customExecuteHandler which implements a pre_execute and post_execute.
    - pre-execute is executed usually right away, or after sanity checks are done in the operation
    - post-execute is executed after commitbuilder has finished
- CommitBuilder also uses operation_id to pass it to commit_entry functions

### DIFF
--- a/crates/aws/src/logstore/default_logstore.rs
+++ b/crates/aws/src/logstore/default_logstore.rs
@@ -11,6 +11,7 @@ use deltalake_core::{
 };
 use object_store::{Error as ObjectStoreError, ObjectStore};
 use url::Url;
+use uuid::Uuid;
 
 /// Return the [S3LogStore] implementation with the provided configuration options
 pub fn default_s3_logstore(
@@ -30,7 +31,7 @@ pub fn default_s3_logstore(
 /// Default [`LogStore`] implementation
 #[derive(Debug, Clone)]
 pub struct S3LogStore {
-    pub(crate) storage: Arc<dyn ObjectStore>,
+    pub(crate) storage: ObjectStoreRef,
     config: LogStoreConfig,
 }
 
@@ -53,7 +54,7 @@ impl LogStore for S3LogStore {
     }
 
     async fn read_commit_entry(&self, version: i64) -> DeltaResult<Option<Bytes>> {
-        read_commit_entry(self.storage.as_ref(), version).await
+        read_commit_entry(self.object_store(None).as_ref(), version).await
     }
 
     /// Tries to commit a prepared commit file. Returns [`TransactionError`]
@@ -65,10 +66,14 @@ impl LogStore for S3LogStore {
         &self,
         version: i64,
         commit_or_bytes: CommitOrBytes,
+        _operation_id: Uuid,
     ) -> Result<(), TransactionError> {
         match commit_or_bytes {
             CommitOrBytes::TmpCommit(tmp_commit) => {
-                Ok(write_commit_entry(&self.object_store(), version, &tmp_commit).await?)
+                Ok(
+                    write_commit_entry(self.object_store(None).as_ref(), version, &tmp_commit)
+                        .await?,
+                )
             }
             _ => unreachable!(), // S3 Log Store should never receive bytes
         }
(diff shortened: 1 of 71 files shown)


crates/aws/src/logstore/default_logstore.rs
```diff
@@ -104,10 +110,14 @@ impl LogStore for S3LogStore {
         get_earliest_version(self, current_version).await
     }
 
-    fn object_store(&self) -> Arc<dyn ObjectStore> {
+    fn object_store(&self, _operation_id: Option<Uuid>) -> Arc<dyn ObjectStore> {
         self.storage.clone()
     }
 
+    fn reading_object_store(&self) -> Arc<dyn ObjectStore> {
```
rtyler, line 117: I don't understand the purpose of this reading_object_store() function. My understanding based on reading the code is that this is to provide a "read-only" or "reader" `ObjectStoreRef` but reviewing the rest of the PR it looks like this is _always_ functionally equivalent to `object_store(None)`.

Assuming my understanding is correct, I think it's best to use `object_store(None)` and keep the LogStore trait slim. I also worry that users might misinterpret the return value of this function to mean read-only in some way, which we have no way of guaranteeing since the returned `ObjectStoreRef` technically can still have puts/etc executed with it.

crates/core/src/logstore/mod.rs
```diff
@@ -201,6 +203,23 @@ pub trait LogStore: Sync + Send {
         Ok(())
     }
 
+    /// Build a new object store for an URL using the existing storage options
+    fn build_new_store(&self, url: &Url) -> DeltaResult<ObjectStoreRef> {
```
rtyler, line 207: Could you expand more in the rustdoc comment on the intended usage of this function,. My understanding is that it's to create a new `ObjectStoreRef` using the existing logstore's options but with a different URL?

It _feels_ a little sticky to me but I don't understand why this is useful enough to suggest alternatives

crates/core/src/operations/mod.rs
```diff
@@ -62,10 +66,61 @@ pub mod update;
...
+        operation_id: Uuid,
+    ) -> DeltaResult<()>;
+}
+
 #[allow(unused)]
 /// The [Operation] trait defines common behaviors that all operations builders
 /// should have consistent
-pub(crate) trait Operation<State>: std::future::IntoFuture {}
+pub(crate) trait Operation<State>: std::future::IntoFuture {
+    fn get_log_store(&self) -> &LogStoreRef;
```
rtyler, line 98: I'm going to make a suggestion and I realize that it will require a lot of copypasta :laughing: 

```suggestion
    fn log_store(&self) -> LogStoreRef;
```

This would be symmetrical with the function on the DeltaTable itself :smile: 

I would recommend that the implementers also return self.log_store.clone() since `LogStoreRef` is an`Arc<dyn LogStore>`

crates/core/src/operations/mod.rs
```diff
@@ -62,10 +66,61 @@ pub mod update;
...
+    fn get_custom_execute_handler(&self) -> Option<&Arc<dyn CustomExecuteHandler>>;
```
rtyler, line 99: Since this is returning an Arc, the `&` should be removed and implementers should return a cloned reference

crates/core/src/operations/transaction/mod.rs
```diff
@@ -680,31 +721,61 @@ impl PostCommit<'_> {
...
-            }
+
             let cleanup_logs = if let Some(cleanup_logs) = self.cleanup_expired_logs {
                 cleanup_logs
             } else {
                 state.table_config().enable_expired_log_cleanup()
             };
 
+            // Run arbitrary before_post_commit_hook code
+            if let Some(custom_execute_handler) = &self.customer_execute_handler {
```
rtyler, line 732: In some places this is called "customer" versus "custom", I think that's likely just a typo, but it would be good to make sure everything is consistently calling this a `custom_execute_handler :smile:

ion-elgreco: That's a brain fart, It should been custom indeed :p

crates/aws/src/logstore/default_logstore.rs
ion-elgreco: Yeah you're right, especially since the lakefslogstore returns a reading store by default when you don't pass an operation id.

Will remove the reading_object_store!

crates/core/src/logstore/mod.rs
ion-elgreco: Yes that's exactly what it does. I was hoping initially to modify the path of a store in place but it's fixed unfortunately, so I added this trait function to make it simpler to create a new store

ion-elgreco: I will actually move this to lakefs logstore only, it's probably to usecase specific

rtyler (dismissed):

rtyler (approved):

PR #3110, opened 2025-01-09
fix: reading cdf from a checkpointed table

### LINKED ISSUE #3097: load_cdf errors out on starting-timestamp only request
# Environment

**Delta-rs version**:
pyrhon-0.23.1

**Binding**: 

**Environment**: 
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

DeltaRS load_cdf  starting_timestamp still requires a starting version values passed into the method.

**What happened**:

I'm running following code:
```
    timestamp = "2025-01-02T00:00:00Z"
    table = DeltaTable(path)
    cdf = table.load_cdf(starting_timestamp=timestamp).read_all()
```

This code errors out with following message:
```
self = DeltaTable(), starting_version = 0, ending_version = None
starting_timestamp = '2025-01-02T00:00:00Z', ending_timestamp = None
columns = None, allow_out_of_range = True

    def load_cdf(
        self,
        starting_version: int = 0,
        ending_version: Optional[int] = None,
        starting_timestamp: Optional[str] = None,
        ending_timestamp: Optional[str] = None,
        columns: Optional[List[str]] = None,
        allow_out_of_range: bool = False,
    ) -> pyarrow.RecordBatchReader:
>       return self._table.load_cdf(
            columns=columns,
            starting_version=starting_version,
            ending_version=ending_version,
            starting_timestamp=starting_timestamp,
            ending_timestamp=ending_timestamp,
            allow_out_of_range=allow_out_of_range,
        )
E       _internal.DeltaError: Invalid table version: 0
```

The table should return the rows from given timestamp.

**How to reproduce it**:

**More details**:

My guess is that starting version is still being evaluated even if it was not passed into the method. It defaults to 0, so if the table being used is vaccumed (like in this example) and the 0 version is missing, the starting_timestamp only cdf errors out

hntd187:
# Description
Fixes a bug where you cannot read CDF from a table that has been checkpointed or vacuumed

# Related Issue(s)
- closes #3097


crates/core/src/operations/load_cdf.rs
```diff
@@ -107,6 +107,25 @@ impl CdfLoadBuilder {
         self
     }
 
+    async fn calculate_earliest_version(&self) -> DeltaResult<i64> {
+        let ts = self.starting_timestamp.unwrap_or(DateTime::UNIX_EPOCH);
+        for v in 0..self.snapshot.version() {
```
ion-elgreco, line 112: This could be potentially expensive, can't we do a binary search between 0 and the last checkpoint version?

The log store has the get_earliest_version function, which is also not efficient. I thought redirection would be configurable there but it's not unfortunately. But we could replace that with a binary search there and it would help for other use cases as well

rtyler (approved):
I am comfortable with an inefficient bug fix now and optimizations later. I'm going to pull this in ahead of the LakeFS work @ion-elgreco so I can release this as a patch before we start the 0.24 work

ion-elgreco:
> I am comfortable with an inefficient bug fix now and optimizations later. I'm going to pull this in ahead of the LakeFS work @ion-elgreco so I can release this as a patch before we start the 0.24 work

Alrighty, I can do a follow up on this then after lakefs work is in :)

hntd187:
Yeah, I can touch up the perf on this a bit after all the other stuff is settled down.

Issue #3046, opened 2024-12-03
Add Python-side support for Unity Catalog


omkar-foss:
# Description

**Use Case**

Add Python-side support for Unity Catalog using https://pypi.org/project/unitycatalog/

**Related Issue(s)**

https://github.com/delta-io/delta-rs/issues/2973

**Related PR(s)**

https://github.com/delta-io/delta-rs/pull/3044

omkar-foss:
take

PR #3078, opened 2024-12-22
feat: support reading tables via Unity Catalog provided credentials

hntd187:
# Description
You can read tables via UC provided creds.

# Related Issue(s)

# Documentation
[Temp Creds API](https://docs.databricks.com/api/workspace/temporarytablecredentials/generatetemporarytablecredentials)

PR #3113, opened 2025-01-10
feat(python): add capability to read unity catalog (uc://) uris

omkar-foss:
# Description
This adds capability to read directly from uc:// uris using the local catalog-unity crate. This also exposes the UC temporary credentials in storage_options of the `DeltaTable` instance so polars or similar readers can use it.

# Related Issue(s)

- closes https://github.com/delta-io/delta-rs/issues/3046

# Documentation

N/A


python/src/lib.rs
```diff
@@ -169,12 +170,24 @@ impl RawDeltaTable {
...
+                    Ok(tup) => tup,
+                    Err(err) => return Err(PyRuntimeError::new_err(err.to_string())),
+                }
+            } else {
+                (table_uri.to_string(), "".to_string())
+            };
+
+            let mut options = storage_options.clone().unwrap_or_default();
+            if !uc_token.is_empty() {
+                options.insert("UNITY_CATALOG_TEMPORARY_TOKEN".to_string(), uc_token);
```
omkar-foss, line 184: Have added this key `UNITY_CATALOG_TEMPORARY_TOKEN` for Polars or similar readers to be able reuse the credentials. cc: @ion-elgreco

crates/catalog-unity/src/lib.rs
```diff
@@ -181,6 +181,11 @@ pub enum UnityCatalogConfigKey {
     /// - `azure_use_azure_cli`
     /// - `use_azure_cli`
     UseAzureCli,
+
+    /// Allow http url (e.g. http://localhost:8080/api/2.1/...)
+    /// Supported keys:
+    /// - `unity_allow_http_url`
+    AllowHttpUrl,
```
omkar-foss, line 188: This allows users to work with a local (non-https) Unity Catalog REST API with `delta-rs`.

python/src/lib.rs
ion-elgreco, line 184: I think this would be a better fit for code inside an object store factory

ion-elgreco: Als you might have to wait before @hntd187's code gets merged (https://github.com/delta-io/delta-rs/pull/3078)

omkar-foss: >Als you might have to wait before @hntd187's code gets merged (https://github.com/delta-io/delta-rs/pull/3078)

Yep will have to wait for https://github.com/delta-io/delta-rs/pull/3078 to get merged before this one, the temp credentials part is in there :)

> I think this would be a better fit for code inside an object store factory

Yes, this code is currently duplicated in `new()` and `is_deltatable()`. Shall I move it to a private function? Or object store factory works too, let me know what you have in mind about that.

omkar-foss: Done, I've moved UC specific code to `UnityCatalogFactory` and added an internal helper method to `RawDeltaTable` for the rest of the boilerplate. Let me know if this looks fine.

omkar-foss:
@ion-elgreco I've updated this PR to include the temp credentials functionality from PR https://github.com/delta-io/delta-rs/pull/3078. Cheers.

python/src/lib.rs
```diff
@@ -185,12 +209,8 @@ impl RawDeltaTable {
         log_buffer_size: Option<usize>,
     ) -> PyResult<Self> {
         py.allow_threads(|| {
-            let mut builder = deltalake::DeltaTableBuilder::from_uri(table_uri)
-                .with_io_runtime(IORuntime::default());
-            let options = storage_options.clone().unwrap_or_default();
-            if let Some(storage_options) = storage_options {
-                builder = builder.with_storage_options(storage_options)
-            }
+            let (mut builder, table_path, options) =
```
ion-elgreco, line 212: I was thinking more in the lines of this, then you don't have to modify any of the python/lib code

https://github.com/delta-io/delta-rs/blob/9dc8f328859b4fe66fa82a636aae662cee22db30/crates/azure/src/lib.rs#L83-L86

omkar-foss: Done thanks, I've moved the UC part to it's own factory. No more modifying of `RawDeltaTable` python/lib code :)

ion-elgreco:
@omkar-foss looks good from my side

omkar-foss:
I'll need to add a few python tests for this flow, will add tomorrow and then if all good, I suppose we can merge it.

Also please note, the current UC REST integration that we have is more fine-tuned for Databricks Unity Catalog. The Unity Catalog OSS has slightly different response payloads for some APIs. e.g. Generate Temp Credentials API returns `r2_temp_credentials` and `url` in the JSON response for [Databricks UC](https://docs.databricks.com/api/workspace/temporarytablecredentials/generatetemporarytablecredentials) but not for [OSS UC](https://docs.unitycatalog.io/swagger-docs/#/TemporaryCredentials/generateTemporaryTableCredentials). So this causes json decoding errors if you try out the current UC client implementation with an OSS UC instance.

I'll look into supporting OSS UC smoothly in a separate PR after this one. Cheers.

ion-elgreco:
@omkar-foss we could add an integration test with OSS version to catch these things essier

omkar-foss:
> > @omkar-foss we could add an integration test with OSS version to catch these things essier
> 
> Yes good idea! I'll add a failing test for Temp Creds API with the UC OSS JSON payload along with some other python tests.

Hi, I've added python integration tests for both UC Databricks and UC OSS with their respective mock payloads. Using [mockoon](https://github.com/mockoon/mockoon) as a mock HTTP server in GitHub Actions for the integration tests.

omkar-foss:
>Why not with the actual UC server?

Since I had to anyway add a mock server for the UC Databricks integration test, thought might as well use it for UC OSS test to keep it simple, since we're testing only 2 UC REST APIs (Temp Credentials and Get Table Details).

Let me know if you foresee a strong use case for adding the actual UC OSS server, if so, I'll add it here :)

crates/catalog-unity/src/lib.rs
```diff
@@ -431,6 +450,45 @@ impl UnityCatalogBuilder {
...
+        };
+        let temp_creds_res = unity_catalog
+            .get_temp_table_credentials(catalog_id, database_name, table_name)
+            .await?;
+        let credentials = match temp_creds_res {
+            TableTempCredentialsResponse::Success(temp_creds) => {
+                temp_creds.get_credentials().unwrap()
+            }
+            TableTempCredentialsResponse::Error(_error) => {
+                panic!("Unable to get temporary credentials from Unity Catalog.")
```
hntd187, line 486: I don't think being unable to get temp creds should result in a panic, perhaps a nice error result?

crates/catalog-unity/src/lib.rs
```diff
@@ -431,6 +450,45 @@ impl UnityCatalogBuilder {
...
     }
 
+    /// Returns the storage location and temporary token to be used with the
+    /// Unity Catalog table.
+    pub async fn get_uc_location_and_token(
+        table_uri: &str,
+    ) -> Result<(String, HashMap<String, String>), UnityCatalogError> {
+        let uri_parts: Vec<&str> = table_uri[5..].split('.').collect();
+        if uri_parts.len() != 3 {
+            panic!("Invalid Unity Catalog URI: {}", table_uri);
```
hntd187, line 460: Return an error instead of panic

crates/catalog-unity/src/lib.rs
```diff
@@ -431,6 +450,45 @@ impl UnityCatalogBuilder {
...
+        }
+
+        let catalog_id = uri_parts[0];
+        let database_name = uri_parts[1];
+        let table_name = uri_parts[2];
+
+        let unity_catalog = match UnityCatalogBuilder::from_env().build() {
```
hntd187, line 467: just use `?` operator instead

crates/catalog-unity/src/lib.rs
```diff
@@ -431,6 +450,45 @@ impl UnityCatalogBuilder {
...
+            Ok(uc) => uc,
+            Err(_e) => panic!("Unable to build Unity Catalog."),
+        };
+        let storage_location = match unity_catalog
+            .get_table_storage_location(Some(catalog_id.to_string()), database_name, table_name)
+            .await
+        {
+            Ok(s) => s,
```
hntd187, line 475: just use `?`

crates/catalog-unity/src/lib.rs
```diff
@@ -692,6 +755,67 @@ impl UnityCatalog {
...
+        options: &StorageOptions,
+    ) -> DeltaResult<(ObjectStoreRef, Path)> {
+        use futures::executor::block_on;
+
+        let result = block_on(UnityCatalogBuilder::get_uc_location_and_token(
+            table_uri.as_str(),
+        ));
+
+        let (table_path, temp_creds) = match result {
+            Ok(tup) => tup,
```
hntd187, line 776: use `?`

crates/catalog-unity/src/lib.rs
```diff
@@ -692,6 +755,67 @@ impl UnityCatalog {
...
+            Err(_err) => panic!("Unable to get UC location and token."),
+        };
+
+        let mut storage_options = options.0.clone();
+
+        if !temp_creds.is_empty() {
```
hntd187, line 782: Does it matter if it's empty? `extend` should work on an empty collection

hntd187 (changes requested):
Mostly good, I'd like this to not panic, instead returning errors

crates/catalog-unity/src/lib.rs
omkar-foss: Done

crates/catalog-unity/src/lib.rs
omkar-foss: Yep, done

crates/catalog-unity/src/lib.rs
omkar-foss: Done

crates/catalog-unity/src/lib.rs
omkar-foss: Done

crates/catalog-unity/src/lib.rs
omkar-foss: Yes, silly me, thanks for this. Done.

omkar-foss:
> Mostly good, I'd like this to not panic, instead returning errors

Thanks for your review, Steve. I've updated the PR as per your comments.

ion-elgreco:
@omkar-foss can we add an integration page to the docs? :)

crates/catalog-unity/src/lib.rs
```diff
@@ -431,6 +484,42 @@ impl UnityCatalogBuilder {
...
+            Some(catalog_id.to_string()),
+            database_name,
+            table_name
+        ).await?;
+        let temp_creds_res = unity_catalog
+            .get_temp_table_credentials(catalog_id, database_name, table_name)
+            .await?;
+        let credentials = match temp_creds_res {
+            TableTempCredentialsResponse::Success(temp_creds) => {
+                temp_creds.get_credentials().unwrap()
```
hntd187, line 514: missed an error case here

hntd187 (commented):
Looks good, just one more error case nit :)

omkar-foss:
> Looks good, just one more error case nit :)

Thanks, done :)

omkar-foss:
> @omkar-foss can you add an integration page to the docs? :)

Yes sure, is it okay if I add it in the next PR?

omkar-foss:
> > @omkar-foss can you add an integration page to the docs? :)
> 
> Yes sure, is it okay if I add it in the next PR?

@ion-elgreco On a second thought, I've added basic UC docs to this PR itself. It looks as follows:

Screenshot at 2025-02-16 16-30-56

In the next PR, I'll also add some code examples and all, along with the UC OSS payload support.

ion-elgreco (dismissed):

ion-elgreco:
@omkar-foss the UC integration test is failing

omkar-foss:
> @omkar-foss the UC integration test is failing

Yes I checked, it's passing on my local. Seems like it's an intermittent failure. Would it be possible for you to retrigger that particular failing job, let's see if it passes?

ion-elgreco:
@omkar-foss can you try to make it not flaky?

omkar-foss:
> @omkar-foss can you try to make it not flaky?

Seems like it gets stuck intermittently due to deadlock, checking on it now.

hntd187 (dismissed):
Nice, I appreciate you doing all the doc work with this as well, everything looks great to me.

omkar-foss:
> > @omkar-foss can you try to make it not flaky?
> 
> Seems like it gets stuck intermittently due to a deadlock, checking on it now.

I've added a dedicated function to execute futures pertaining to UC, based on [the function](https://github.com/delta-io/delta-rs/blob/main/crates/aws/src/storage.rs#L262-L295) we have for AWS S3 storage. This is helping to prevent the deadlocks by using an appropriate `block_on()`. A deadlock can still occur, but very low chances (less than 2% probability as per my sampling of 100 runs, each 5 seconds apart).

@ion-elgreco @hntd187 guys, please check it out and let me know if this makes sense. Thank you :)

ion-elgreco (approved):

ion-elgreco (approved):

PR #3117, opened 2025-01-12
refactor: use kernel table features

ion-elgreco:
# Description
Swapped out table features code from internal to delta-kernel-rs.


roeap (approved):
LGTM!

PR #3121, opened 2025-01-12
chore: pin to delta-kernel 0.6.0 for now

rtyler:


roeap (dismissed):

rtyler:
pin already went in elsewhere

PR #3122, opened 2025-01-12
fix: ignore crc files when checking if provided path correspond to a valid delta table

### LINKED ISSUE #3115: `is_deltatable()` returns False for tables in S3 created by delta-spark
# Environment

**Delta-rs version**:
deltalake==0.23.2
**Binding**: 
python

**Environment**:
- **Cloud provider**:
AWS (running on ECS)
- **OS**:
Mac OS for local development, Linux for AWS (bug appears in both)
- **Other**:

***
# Bug

**What happened**:
`deltalake.DeltaTable.is_deltatable("s3://path/to/table.delta")` always returns False when reading tables created previously by Spark, while `deltalake.DeltaTable("s3://path/to/table.delta")` returns a usable DeltaTable object that I can call `.history()` on to confirm it's reading a table that exists. I further confirmed that I shouldn't get a DeltaTable object back if it doesn't exist by doing `deltalake.DeltaTable("s3://path/to/non-existent-table.delta")`, which raises a `deltalake._internal.TableNotFoundError`.

I expected `DeltaTable.is_deltatable(...)` to return True when given a S3 path that contains a Delta table, regardless of whether the table was written by Spark, polars, duckdb, etc.

**How to reproduce it**:
```
from pyspark.conf import SparkConf

spark = SparkSession.builder.config(conf=SparkConf({"spark.jars.packages": "io.delta:delta-spark_2.12:3.2.0,org.apache.hadoop:hadoop-aws:3.3.4,com.amazonaws:aws-java-sdk-bundle:1.12.262,org.apache.hadoop:hadoop-common:3.3.4", "spark.sql.extensions": "io.delta.sql.DeltaSparkSessionExtension", "spark.sql.catalog.spark_catalog": "org.apache.spark.sql.delta.catalog.DeltaCatalog"})).getOrCreate()

spark.createDataFrame([{"a": 1]).write.format("delta").save("s3://some/path.delta")

import deltalake
deltalake.DeltaTable.is_deltatable("s3://some/path.delta") # returns False
deltalake.DeltaTable("s3://some/path.delta") # succeeds

```

**More details**:
This seems specific to detecting whether a table exists when the table in question was written by Spark. I was able to determine that `is_deltatable` succeeds for tables in S3 written by polars, and directly from deltalake via `write_deltalake()`. I suspect there is some table metadata that `is_deltatable` is checking for which is only written by `deltalake`, and not when a table is written using `delta-spark`. The Spark-written tables were written using pyspark==3.5.1 and delta-spark==3.2.0.

I have not checked whether the issue appears for local delta tables written by spark, so the S3 aspect of this issue may be unrelated. I suspect it's more specific to detecting delta tables written by delta-spark.

guillotjulien:
# Description

The current implementation of `is_delta_table_location` only check the first file in the `_delta_log` folder to check for the presence of a commit or checkpoint file. This creates issue when checking tables created using the Spark driver since `.crc` files are created and appear before commit files in the directory.

This MR modifies the behavior of `is_delta_table_location` so that files are checked in a loop. `.crc` files are ignored and the loop execution stops as soon as a commit, checkpoint or invalid file is found.

# Related Issue(s)

- closes #3115

### DIFF
--- a/crates/core/src/kernel/snapshot/log_segment.rs
+++ b/crates/core/src/kernel/snapshot/log_segment.rs
@@ -28,6 +28,7 @@ lazy_static! {
     static ref CHECKPOINT_FILE_PATTERN: Regex =
         Regex::new(r"\d+\.checkpoint(\.\d+\.\d+)?\.parquet").unwrap();
     static ref DELTA_FILE_PATTERN: Regex = Regex::new(r"^\d+\.json$").unwrap();
+    static ref CRC_FILE_PATTERN: Regex = Regex::new(r"^(\.)?\d+(\.crc|\.json)?\.crc$").unwrap();
     pub(super) static ref TOMBSTONE_SCHEMA: StructType =
         StructType::new(vec![ActionType::Remove.schema_field().clone(),]);
 }
@@ -60,6 +61,12 @@ pub(crate) trait PathExt {
         self.filename()
             .map(|name| DELTA_FILE_PATTERN.captures(name).is_some())
             .unwrap_or(false)
+    }
+
+    fn is_crc_file(&self) -> bool {
+        self.filename()
+            .map(|name| CRC_FILE_PATTERN.captures(name).is_some())
+            .unwrap()
     }
 }
 
--- a/crates/core/src/kernel/snapshot/log_segment.rs
+++ b/crates/core/src/kernel/snapshot/log_segment.rs
@@ -28,7 +28,7 @@ lazy_static! {
     static ref CHECKPOINT_FILE_PATTERN: Regex =
         Regex::new(r"\d+\.checkpoint(\.\d+\.\d+)?\.parquet").unwrap();
     static ref DELTA_FILE_PATTERN: Regex = Regex::new(r"^\d+\.json$").unwrap();
-    static ref CRC_FILE_PATTERN: Regex = Regex::new(r"^(\.)?\d+(\.crc|\.json)?\.crc$").unwrap();
+    static ref CRC_FILE_PATTERN: Regex = Regex::new(r"^(\.\d+(\.crc|\.json)|\d+)\.crc$").unwrap();
     pub(super) static ref TOMBSTONE_SCHEMA: StructType =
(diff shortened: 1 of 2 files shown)


crates/core/src/kernel/snapshot/log_segment.rs
```diff
@@ -28,6 +28,7 @@ lazy_static! {
     static ref CHECKPOINT_FILE_PATTERN: Regex =
         Regex::new(r"\d+\.checkpoint(\.\d+\.\d+)?\.parquet").unwrap();
     static ref DELTA_FILE_PATTERN: Regex = Regex::new(r"^\d+\.json$").unwrap();
+    static ref CRC_FILE_PATTERN: Regex = Regex::new(r"^(\.)?\d+(\.crc|\.json)?\.crc$").unwrap();
```
ion-elgreco, line 31: These are valid files?

00001.crc.crc 
00001.json.crc

guillotjulien: Not from what I observed. The only `.crc` file generated without `.` prefix follow the `^\d+\.crt$` pattern. We can further restrict the regex like this: `^(\.\d+(\.crc|\.json)|\d+)\.crc$`.

It should cover every case:
image

ion-elgreco: Why isn't it just` r"^\d+\.crc$"`?

I have not seen .json.crc or .crc.crc before tbh

guillotjulien: Those get generated when I write a delta table locally using Spark 3.5.4 and delta-spark 3.3.0. iirc I also saw them with Spark 3.2 / delta-spark 2.3.0, so I assume they're legit?

ion-elgreco: I guess :P I just wanted some confirmation on this, because I hadn't seen them before

ion-elgreco (approved):

ion-elgreco:
Thanks for the fix!

Issue #3116, opened 2025-01-11
Generated Columns


ion-elgreco:
# Description

**Use Case**

**Related Issue(s)**

PR #3123, opened 2025-01-12
feat: generated columns

ion-elgreco:
# Description
Adds generated columns.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/3116

### DIFF
--- a/crates/core/src/operations/write.rs
+++ b/crates/core/src/operations/write.rs
@@ -36,7 +36,7 @@ use arrow_cast::can_cast_types;
 use arrow_schema::{ArrowError, DataType, Fields, SchemaRef as ArrowSchemaRef};
 use datafusion::execution::context::{SessionContext, SessionState, TaskContext};
 use datafusion_common::DFSchema;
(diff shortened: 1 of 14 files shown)


crates/core/src/operations/write.rs
```diff
@@ -439,7 +438,11 @@ async fn write_execution_plan_with_predicate(
     let checker = if let Some(snapshot) = snapshot {
         DeltaDataChecker::new(snapshot)
     } else {
-        DeltaDataChecker::empty()
+        debug!("Using plan schema to derive generated columns, since no shapshot was provided. Implies first write.");
```
ion-elgreco, line 441: A cleaner way maybe, is to check upstream whether an metadata action existed, then we convert that schema into StructType and grab the generated columns from there

rtyler (approved):
This looks good to me, it does touch a lot of areas of the write path my performance work has been touching :grimacing: so I have some brutal conflict resolutions in my future.

I will release this in the Python 0.24 and let it marinate a couple more days before I cut the :crab: crates

Issue #2965, opened 2024-10-29
[python] File size too large - maybe stats related


convoi:
# Environment

**Delta-rs version**: deltalake 0.20.2

**Binding**: python

**Environment**:
- **Cloud provider**:
- **OS**: Mac OS Sonoma (Apple Silicon)
- **Other**:

***
# Bug

**What happened**:
I'm storing highly compressible but large strings in a delta table. If I write the data frame using pandas to parquet directly, the resulting parquet file is very small (2kb for a 1MB input string containing just the letter "a").
If I write the same data frame to a delta table, the resulting file is 2.0 MB.

**What you expected to happen**:
I expect the delta parquet files to have a similar size as the normal parquet files.
I explicitly stated to create the files without large statistics (truncate it to 16 chars) and it seems this is is true on the row group statistics level, as the output of parquet meta suggests.
However, when inspecting the file with a hex editor I still see the uncompressed strings. Is there a column index written, and if so, how do I turn it of?

**How to reproduce it**:

```python
import deltalake as dt
import pandas as pd
import pyarrow as pa
one_mb = "a" * 1024 * 1024
df = pd.DataFrame({
        "name": ["vin1"] * 1,
        "date": ["2022-01-01"] * 1,
        "large_data": [one_mb] * 1,
    })
dt.write_deltalake("test_delta", df, engine="rust", mode="overwrite",
                     writer_properties=dt.WriterProperties(
                         compression="ZSTD",
                         statistics_truncate_length=16,
                         default_column_properties=ColumnProperties(dictionary_enabled=False, max_statistics_size=1),
                         column_properties={
                             "large_data": ColumnProperties(dictionary_enabled=False,
                                                           max_statistics_size=1,
                                                           bloom_filter_properties=BloomFilterProperties(
                                                               set_bloom_filter_enabled=False)),
                         }
                     ),
                     )
```

**More details**:
parquet meta output on the two files is also different:
on the small one:
```
Row group 0:  count: 1  299.00 B records  start: 4  total(compressed): 299 B total(uncompressed):1.000 MB 
--------------------------------------------------------------------------------
            type      encodings count     avg size   nulls   min / max
name        BINARY    Z _ R     1         82.00 B    0       "vin1" / "vin1"
date        BINARY    Z _ R     1         100.00 B   0       "2022-01-01" / "2022-01-01"
large_data  BINARY    Z _ R     1         117.00 B   0       

```
on the delta table one:
```
[...]

PR #3126, opened 2025-01-14
feat(python, rust): add statistics_enabled to ColumnProperties

maxitg:
# Description

- Add support for [`statistics_enabled`](https://arrow.apache.org/rust/parquet/file/properties/struct.ColumnProperties.html#structfield.statistics_enabled) in `ColumnProperties`.
- Remove deprecated and unused `max_statistics_size` from `ColumnProperties`.

# Related Issue(s)

- #2965

# Documentation

The only available way currently to avoid writing statistics to the transaction log for a column is via `max_statistics_size`. However, that has been [deprecated](https://arrow.apache.org/rust/parquet/file/properties/struct.ColumnProperties.html#structfield.max_statistics_size) in `arrow-rs`, furthermore it appears to be ignored: https://github.com/apache/arrow-rs/issues/2033. Being able to skip statistics is necessary for large data columns, as otherwise transaction log explodes after a few tens of thousands of transactions and becomes non-scalable.

# Examples

```python
from deltalake import DeltaTable, write_deltalake, WriterProperties, ColumnProperties
import pandas as pd
import pyarrow as pa
from pathlib import Path
import shutil

data = {'int': [0, 1, 2], 'data': [b'0' * 100000, b'1' * 100000, b'2' * 100000]}

df = pd.DataFrame(data)

def check_size(data_column_properties: ColumnProperties):
    TMP_PATH = "/tmp/delta-test/delta-test-0"

    shutil.rmtree(TMP_PATH, ignore_errors=True)
    write_deltalake(
        TMP_PATH,
        df.sort_values(["int"]),
        schema=pa.schema([pa.field("int", pa.int64()), pa.field("data", pa.binary())]),
        mode="append",
        writer_properties=WriterProperties(
            compression="zstd",
            column_properties={"data": data_column_properties},
        ),
        engine="rust",
    )

    print(DeltaTable(TMP_PATH).to_pandas())

    return (Path(TMP_PATH) / "_delta_log/00000000000000000000.json").stat().st_size

[check_size(c)
 for c
 in [ColumnProperties(max_statistics_size=0), ColumnProperties(statistics_enabled="NONE")]]

#    int                                               data
# 0    0  b'00000000000000000000000000000000000000000000...
# 1    1  b'11111111111111111111111111111111111111111111...
# 2    2  b'22222222222222222222222222222222222222222222...
#    int                                               data
# 0    0  b'00000000000000000000000000000000000000000000...
# 1    1  b'11111111111111111111111111111111111111111111...
# 2    2  b'22222222222222222222222222222222222222222222...
# [201143, 1100]
```

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -218,7 +218,6 @@ class ColumnProperties:
         self,
         dictionary_enabled: Optional[bool] = None,
         statistics_enabled: Optional[Literal["NONE", "CHUNK", "PAGE"]] = None,
-        max_statistics_size: Optional[int] = None,
         bloom_filter_properties: Optional[BloomFilterProperties] = None,
     ):
         """Create a Column Properties instance for the Rust parquet writer:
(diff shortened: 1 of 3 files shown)


maxitg:
> @maxitg looks good, could you remove the max_statistics_size as well?

@ion-elgreco I’ve added a second commit to remove it. I initially avoided it to prevent a breaking change.

rtyler (approved):
Thanks for the improvement! This is a situation where I'm definitely okay with breaking changes since the code wasn't working anyways, we're also on the precipice of a breaking set of changes anyways :smile:

Issue #3128, opened 2025-01-14
Set max_retry, retry_timeout from python


dlandriscina:
Would it be possible to allow the insertion of these variables from Python? 
I'm using polars and saving deltalake on S3 bucket I've received several times "OSError: Generic S3 error: Error after 10 retries in 1.74678935s, max_retries:10, retry_timeout:180s,". 
In addition, I think it would be useful to also have a variable that expresses how many seconds must pass between one attempt and another

Thanks in advance

ion-elgreco:
At the moment we are not parsing them from what I can see, I'll put in fix for all the stores ;  )

PR #3129, opened 2025-01-14
feat: retryable config enablement through storage_options

ion-elgreco:
# Description
- closes https://github.com/delta-io/delta-rs/issues/3128

### DIFF
--- a/crates/azure/src/lib.rs
+++ b/crates/azure/src/lib.rs
@@ -46,13 +46,11 @@ impl ObjectStoreFactory for AzureFactory {
     ) -> DeltaResult<(ObjectStoreRef, Path)> {
         let config = config::AzureConfigHelper::try_new(options.as_azure_options())?.build()?;
 
-        let (_scheme, path) =
+        let (_, path) =
             ObjectStoreScheme::parse(&url).map_err(|e| DeltaTableError::GenericError {
                 source: Box::new(e),
             })?;
         let prefix = Path::parse(path)?;
-
-        // let (inner, prefix) = parse_url_opts(url, config)?;
 
         let mut builder = MicrosoftAzureBuilder::new().with_url(url.to_string());
 
--- a/crates/azure/src/lib.rs
+++ b/crates/azure/src/lib.rs
@@ -5,11 +5,11 @@ use std::sync::Arc;
 use deltalake_core::logstore::{default_logstore, logstores, LogStore, LogStoreFactory};
 use deltalake_core::storage::{
     factories, limit_store_handler, url_prefix_handler, ObjectStoreFactory, ObjectStoreRef,
-    StorageOptions,
+    RetryConfigParse, StorageOptions,
 };
-use deltalake_core::{DeltaResult, Path};
-use object_store::azure::AzureConfigKey;
-use object_store::parse_url_opts;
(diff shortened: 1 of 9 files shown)


crates/azure/src/lib.rs
```diff
@@ -36,14 +36,34 @@ impl AzureOptions for StorageOptions {
...
 impl ObjectStoreFactory for AzureFactory {
     fn parse_url_opts(
         &self,
         url: &Url,
         options: &StorageOptions,
     ) -> DeltaResult<(ObjectStoreRef, Path)> {
         let config = config::AzureConfigHelper::try_new(options.as_azure_options())?.build()?;
-        let (inner, prefix) = parse_url_opts(url, config)?;
+
+        let (_scheme, path) =
```
roeap, line 49: Nit: IIRC, there is a slight difference between `_scheme` and `_`, the latter never getting allocated, so maybe use that?

crates/azure/src/lib.rs
```diff
@@ -36,14 +36,34 @@ impl AzureOptions for StorageOptions {
...
+            ObjectStoreScheme::parse(&url).map_err(|e| DeltaTableError::GenericError {
+                source: Box::new(e),
+            })?;
+        let prefix = Path::parse(path)?;
+
+        // let (inner, prefix) = parse_url_opts(url, config)?;
```
roeap, line 55: Remove?

crates/azure/src/lib.rs
```diff
@@ -36,14 +36,34 @@ impl AzureOptions for StorageOptions {
...
+
+        for (key, value) in config.iter() {
+            builder = builder.with_config(key.clone(), value.clone());
+        }
+
+        let inner = builder
+            .with_retry(self.parse_retry_config(&options)?)
+            .build()?;
+
         let store = limit_store_handler(url_prefix_handler(inner, prefix.clone()), options);
```
roeap, line 67: Not related to this PR, but why do we wrap this in the limiting store again?

crates/core/Cargo.toml
```diff
@@ -98,6 +98,7 @@ rand = "0.8"
 z85 = "3.0.5"
 maplit = "1"
 sqlparser = { version = "0.53.0" }
+humantime = { version = "2.1.0"}
```
roeap, line 101:
```suggestion
humantime = { version = "2.1.0" }
```

roeap (commented):
Just some small nits to look at.

crates/azure/src/lib.rs
ion-elgreco: Interesting! I did not know this!

crates/azure/src/lib.rs
ion-elgreco: Its due to this PR: https://github.com/delta-io/delta-rs/pull/2458

I will put a small PR to actually document this storage option

ion-elgreco: @roeap I've added docs in this PR, and also included the the other undocumented options :)

rtyler (approved):
Thank you for documenting some of these magic hidden variables a bit better :smile: 

I see you're really loving traits lately :laughing:

PR #3138, opened 2025-01-16
chore: fix typos

roeap:
# Description

Thanks to my recently installed spellchecker, I was made aware of some types scattered across the codebase.


ion-elgreco:
@roeap I think the latest pyarrow version 19 broke some things in our CI. Maybe we should pin it for now?

roeap:
@ion-elgreco - I'll let you be the judge on that since I have no good handle on how big the effort would be to just fix it?

ion-elgreco:
> @ion-elgreco - I'll let you be the judge on that since I have no good handle on how big the effort would be to just fix it?

Let's pin pyarrow in the CI for now to < 19. It likely is an issue in arrow I think because it's complained about malformed things even though with lower pyarrow version it can read parquet fine

ion-elgreco (approved):

Issue #3139, opened 2025-01-17
ValueError: Invalid comparison operation: Utf8 == LargeUtf8 on write_deltalake in overwrite-mode


gilgeorges:
# Environment

**Delta-rs version**: 0.24.0

**Binding**: python

**Environment**:
- **Cloud provider**: none - local
- **OS**: Docker (Debian image) on WSL2 on Windows 10
- **Other**: Python 3.13

***
# Bug

**What happened**:
When attempting a `write_deltalake` in overwrite-`mode` with a simple `predicate`  ("<col> = <value>"), if value is an integer, everything works as expected - any existing records under the same label are overwritten. If however value is a string (see code below), `write_deltalake` (through polars' `write_delta`) complains of an "Invalid comparison operation: Utf8 == LargeUtf8"
```
Traceback (most recent call last):
  File "<python-input-11>", line 1, in <module>
    test()
    ~~~~^^
  File "<python-input-3>", line 5, in test
    d.write_delta(tempdir / 'test.delta', mode='overwrite', delta_write_options=dict(
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        engine='rust',
        ^^^^^^^^^^^^^^
        predicate='C = \'a\''
        ^^^^^^^^^^^^^^^^^^^^^
    ))
    ^^
  File "/workspaces/debian-2/.venv/lib/python3.13/site-packages/polars/dataframe/frame.py", line 4305, in write_delta
    write_deltalake(
    ~~~~~~~~~~~~~~~^
        table_or_uri=target,
        ^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
        **delta_write_options,
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/workspaces/debian-2/.venv/lib/python3.13/site-packages/deltalake/writer.py", line 323, in write_deltalake
    write_deltalake_rust(
    ~~~~~~~~~~~~~~~~~~~~^
        table_uri=table_uri,
        ^^^^^^^^^^^^^^^^^^^^
    ...<13 lines>...
        post_commithook_properties=post_commithook_properties,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #3141, opened 2025-01-17
refactor: logical plans in writer

ion-elgreco:
# Description
We were still using physical plans as starting point, we now only take in logical plans and pass them around. This solves also issues with physical type coercion, since logical type coercion is more flexible.

This is a breaking change though since folks who use .`with_execution_plan `now how to provide a logical plan instead of a physical plan. 

# Related Issue(s)
- close https://github.com/delta-io/delta-rs/issues/3139

### DIFF
--- a/crates/core/src/operations/write.rs
+++ b/crates/core/src/operations/write.rs
@@ -33,15 +33,13 @@ use std::vec;
 use arrow_array::RecordBatch;
 use arrow_cast::can_cast_types;
 use arrow_schema::{ArrowError, DataType, Fields, SchemaRef as ArrowSchemaRef};
+use datafusion::catalog::TableProvider;
+use datafusion::datasource::{provider_as_source, MemTable};
 use datafusion::execution::context::{SessionContext, SessionState, TaskContext};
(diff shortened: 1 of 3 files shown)


roeap (commented):
Love us adopting logical plans more and more!!

Will give this a more thorough read later on, just looking at the changes alone w/o context - do we want to pass around a `DataFrame` or could we just use the `LogicalPlan` directly without the attached session?

ion-elgreco:
@roeap the way I was thinking about it, internally we can use dataframes since we use them directly with DF API, and the input for the builder would just be a plan.

I was thinking yesterday to also make the schema evolution into a custom logical plan node. Any thoughts on this?

hntd187:
> @roeap the way I was thinking about it, internally we can use dataframes since we use them directly with DF API, and the input for the builder would just be a plan.
> 
> I was thinking yesterday to also make the schema evolution into a custom logical plan node. Any thoughts on this?

Most everything _should_ become a logical plan node at some point. I know that's much easier said than done, but still.

rtyler (approved):

PR #3142, opened 2025-01-18
refactor: combine normal and cdf plan until write for merge

ion-elgreco:
# Description
During some exploration to make merge streamable, the complication is an ArrowArrayStreamReader can only be consumed once, so when we materialize (execute) a physical plan which has this LazyTableProvider twice, the second run will have no data.

This PR makes the MERGE+ MERGE_CDF plan a combined plan, and splits out the data during the write. A side benefit is that we now just have one function to do the writing and return all actions.

@JonasDev1 your work in MERGE to use min_max pruning from the source for the target scan also complicates things a bit since we consume the stream as a whole. We can solve this by caching the df beforehand, but then everything will stay in memory defeating the streamed execution. I'm curious if you have any idea's on how we could do this without full materialization? I couldn't find you on Slack, but feel free to ping me there to discuss it a bit more

## Other thoughts
I moved this writer temporarily under the merge module, but the idea is to use this later also for the normal write operation. But I need the Logical plan refactor to be merged first

### DIFF
--- a/crates/core/src/operations/merge/writer.rs
+++ b/crates/core/src/operations/merge/writer.rs
@@ -9,7 +9,7 @@ use arrow_schema::{Schema, SchemaRef as ArrowSchemaRef};
 use datafusion::catalog::TableProvider;
 use datafusion::datasource::MemTable;
 use datafusion::execution::context::{SessionContext, SessionState, TaskContext};
-use datafusion_expr::col;
+use datafusion_expr::{col, lit};
 use datafusion_physical_plan::ExecutionPlan;
 use futures::StreamExt;
 use object_store::prefix::PrefixStore;
@@ -170,9 +170,21 @@ pub(crate) async fn write_execution_plan_v2(
 
                         let normal_df = batch_df
                             .clone()
-                            .filter(col(CDC_COLUMN_NAME).is_null())?
+                            .filter(col(CDC_COLUMN_NAME).in_list(
+                                vec![lit("delete"), lit("source_delete"), lit("update_preimage")],
+                                true,
+                            ))?
                             .drop_columns(&[CDC_COLUMN_NAME])?;
-                        let cdf_df = batch_df.filter(col(CDC_COLUMN_NAME).is_not_null())?;
+
+                        let cdf_df = batch_df.filter(col(CDC_COLUMN_NAME).in_list(
+                            vec![
+                                lit("delete"),
+                                lit("insert"),
+                                lit("update_preimage"),
+                                lit("update_postimage"),
+                            ],
+                            false,
+                        ))?;
 
                         let normal_batch =
                             concat_batches(&write_schema, &normal_df.collect().await?)?;
--- a/crates/core/src/operations/merge/writer.rs
+++ b/crates/core/src/operations/merge/writer.rs
@@ -1,0 +1,236 @@
+//! Writer for MERGE operation, can write normal and CDF data at same time
+
+use std::sync::Arc;
+use std::vec;
+
+use arrow::compute::concat_batches;
+use arrow_array::RecordBatch;
+use arrow_schema::{Schema, SchemaRef as ArrowSchemaRef};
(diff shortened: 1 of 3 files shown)


JonasDev1:
One possible optimization: 
You could also manage the two writes without the union and the data split durring the write, if you add the cdc column to the original df with all possible states copy, delete, insert, update and then split them in the write plan. This should be faster and save memory.

JonasDev1:
Regarding min max: I haven't understood it yet, but each stream should actually only get one part of the data, then only this part would be cached, or do you mean that the problem occurs because the cache is not cleared between the calls?

ion-elgreco:
> One possible optimization: You could also manage the two writes without the union and the data split durring the write, if you add the cdc column to the original df with all possible states copy, delete, insert, update and then split them in the write plan. This should be faster and save memory.

What do you mean split them in the write plan?

ion-elgreco:
> Regarding min max: I haven't understood it yet, but each stream should actually only get one part of the data, then only this part would be cached, or do you mean that the problem occurs because the cache is not cleared between the calls?

The issue is, that the early filter uses min_max stats, which requires an aggregation execution plan. This execution plan will consume the source stream. Once consumed, the RecordBatchGenerator returns 0 batches.

The only way to resolve that is caching the result in a memtable prior to executing, but than all of your data is in memory and not streamed anymore

JonasDev1:
> > One possible optimization: You could also manage the two writes without the union and the data split durring the write, if you add the cdc column to the original df with all possible states copy, delete, insert, update and then split them in the write plan. This should be faster and save memory.
> 
> What do you mean with split them in the write plan?

Instead of doing the split into normal_df and cdf_df by null filtering you can filter by change value. Then you need to pass the df only once with all changes. 

Something like:
```
normal_df = batch_df.clone().filter(col(CDC_COLUMN_NAME).isin("copy", "insert", "copy")).drop_columns(&[CDC_COLUMN_NAME])?;
cdf_df = batch_df.filter(col(CDC_COLUMN_NAME).isin("delete", "insert", "copy"))
```

ion-elgreco:
> > > One possible optimization: You could also manage the two writes without the union and the data split durring the write, if you add the cdc column to the original df with all possible states copy, delete, insert, update and then split them in the write plan. This should be faster and save memory.
> > 
> > What do you mean with split them in the write plan?
> 
> Instead of doing the split into normal_df and cdf_df by null filtering you can filter by change value. Then you need to pass the df only once with all changes. 
> 
> Something like:
> ```
> normal_df = batch_df.clone().filter(col(CDC_COLUMN_NAME).isin("copy", "insert", "copy")).drop_columns(&[CDC_COLUMN_NAME])?;
> cdf_df = batch_df.filter(col(CDC_COLUMN_NAME).isin("delete", "insert", "copy"))
> ```
> 
> 

Ah right, see what you mean now! Let me check!

rtyler (approved):
very interesting, let's see what happens  :smile:

crates/core/src/operations/merge/writer.rs
```diff
@@ -0,0 +1,248 @@
...
+                        let table_provider: Arc<dyn TableProvider> = Arc::new(MemTable::try_new(
+                            batch.schema(),
+                            vec![vec![batch.clone()]],
+                        )?);
+                        let batch_df = session_context.read_table(table_provider).unwrap();
+
+                        let normal_df = batch_df
+                            .clone()
+                            .filter(col(CDC_COLUMN_NAME).in_list(
+                                vec![lit("delete"), lit("source_delete"), lit("update_preimage")],
```
JonasDev1, line 174: Is the copy missing here?

ion-elgreco: I could have probably just filtered on that now that I think about it haha

JonasDev1: Ahhh the true means negation :D 
Then it makes sense, thank you

PR #3145, opened 2025-01-19
feat: streamed execution in MERGE

ion-elgreco:
# Description
This adds support for streamed execution in MERGE, with two caveats:

- If CDF is enabled we have to materialize since a stream can only be consumed once. I could make it do this only in streaming mode, but I think it might make sense in general, but I can also only enable this for streaming
- Initially MERGE derives statistics about the source to prune the target a bit better and allow for more concurrent writers, the problem here is that deriving statistics about source means we have to consume source. Caching that early one will defeat the purpose of the streamed execution. So I've disabled this part of the early filter construction when streaming mode is on.

In the future we can look into the work of [influxdb](https://github.com/influxdata/influxdb3_core/blob/main/iox_query/src/exec/split.rs) or [lancedb](https://github.com/lancedb/lance/blob/c4def70444977ca2a235f6524069b092afba5071/rust/lance/src/io/exec/utils.rs#L29). Especially the work of lancedb is interesting since it can allow for replay of the same stream, which would less stressful on memory when we have cdf enabled in merge. This should work well I think with this PR: https://github.com/delta-io/delta-rs/pull/3142

I've actually based some of the work in here on top of this already: https://github.com/delta-io/delta-rs/pull/3142, so that has to be merged first. Will also rewrite the commits then since they are messy now (:


ion-elgreco:
I broke CDF by projecting to early and removing the before state, but need to cache when everything is still there, but we can't cache a dataframe when there are table references

roeap (commented):
took me a bit to read through all of this - very exciting stuff!

Do I read that correctly, that this right now mainly affects tables that are coming from the python side, or if someone were to specify a table externally in rs thta is not a delta table (or at lest not ours? :)

So the flow where this makes most sense would be where someone has a `RecordBatchReader` in python and wants to merge that with our table? From the test side it looks like we are mostly consuming pyarrow tables, which are already materialized?

Would it make sense to add an example o.a. to showcase where using this functionality makes most sense?

Alos excited about this b/c once we have kernelized things, this fits very well into the execution model (i think).

roeap:
This reminds me - at some point we were discussing that we may no longer be able to use the `Dataset` abstraction from pyarrow, when deletion vectors land... which are one of the main things I hope to finally unblock with the kernel migration :)

Is that correct, or am I remembering something wrong?

In any case we should be able to provider a recordbatch reader implementation of we can no longer use that api.

ion-elgreco:
> took me a bit to read through all of this - very exciting stuff!
> 
> Do I read that correctly, that this right now mainly affects tables that are coming from the python side, or if someone were to specify a table externally in rs that is not a delta table (or at lest not ours? :)
> 

Yes at least in an automatic fashion. But rust users can use these lines to get the same result with these lines and then passing the DF into our operation. But they do have to enable the flag, streaming=True otherwise the early_filter_construction will consume the stream before the writer receives it:
```
            let arrow_stream: Arc<Mutex<ArrowArrayStreamReader>> = Arc::new(Mutex::new(source)); // Should be a send+sync stream
            let arrow_stream_batch_generator: Arc<RwLock<dyn LazyBatchGenerator>> =
                Arc::new(RwLock::new(ArrowStreamBatchGenerator::new(arrow_stream)));

            let table_provider: Arc<dyn TableProvider> = Arc::new(LazyTableProvider::try_new(
                schema.clone(),
                vec![arrow_stream_batch_generator],
            )?);
            ctx.read_table(table_provider).unwrap()
```
I am not sure if we can automatically enable streaming when see a certain TableProvider, I believe we can do downcast_ref on tableProviders but not sure how to do this once it's in a logical plan. Especially when people have their own implementation of a lazyTableProvider it will become tricky, but maybe you have some ideas? 

> So the flow where this makes most sense would be where someone has a `RecordBatchReader` or `Dataset` in python and wants to merge that with our table? Our tables being datasets right now?
> 
> Also excited about this b/c once we have kernelized things, this fits very well into the execution model (i think).

Exactly, anything that will provide RecordBatchStream will be consumed and marked as streaming mode. I am currently also refactoring the schema evolution into a logical node for write (https://github.com/ion-elgreco/delta-rs/tree/feat/lazy_schema_evolution), so that it brings us closer to streamed execution as well. 

> This reminds me - at some point we were discussing that we may no longer be able to use the `Dataset` abstraction from pyarrow, when deletion vectors land... which are one of the main things I hope to finally unblock with the kernel migration :)
> 
> Is that correct, or am I remembering something wrong?
> 
> In any case we should be able to provider a recordbatch reader implementation of we can no longer use that api.
Yeah that is a tricky one, because Dataset construct doesn't provide you an interface to filter down dataset fragments. So you are saying with kernel we can use the dataset abstraction? :O

Issue #2854, opened 2024-09-07
Binary col in `delta.dataSkippingStatsColumns` not working


ion-elgreco:
A somewhat related bug - if you specify a column in `delta.dataSkippingStatsColumns` that is binary you'll get this error, even though the same column works without DSSC config option specified

```
_internal.DeltaError: Generic DeltaTable error: Stats column binary_column has unsupported type binary
```

It seems odd that by using the DSSC config option, a lot of things that work normally suddenly break?

_Originally posted by @mattfysh in https://github.com/delta-io/delta-rs/issues/2849#issuecomment-2333511648_

omkar-foss:
take

Issue #2373, opened 2024-04-02
Write struct with binary column failed with Json error: Binary is not supported by JSON


flxliu:
# Environment

**Delta-rs version**:
'0.17.1'

**Binding**: 
Python

**Environment**:
- **Cloud provider**:
- **OS**: Linux 20.04.1-Ubuntu
- **Other**:

***
# Bug

**What happened**:
I have a struct column, which has a nest binary column. writing this strut failed. 

It's easy to repro by using duckdb

```
import duckdb
import deltalake

result = duckdb.sql("select {'a': NULL::BLOB}").arrow()
deltalake.write_deltalake("a-rs", result)
```

output:
`Exception: Json error: Binary is not supported by JSON`

The file is generated though, it looks the parquet file itself is correct when querying it from duckdb.

**What you expected to happen**:
create the delta table correctly. 

**How to reproduce it**:

run the code above

**More details**:

I tried to read the generated delta table using `a = deltalake.DeltaTable('a-rs')`. It throws error "Exception: Json error: Binary is not supported by JSON". Which seems to be https://github.com/delta-io/delta-rs/issues/2302

omkar-foss:
This issue is related to https://github.com/delta-io/delta-rs/issues/2854, so I should be able resolve both these in PR https://github.com/delta-io/delta-rs/pull/3146.

omkar-foss:
take

ion-elgreco:
This actually runs fine nowadays.

PR #3146, opened 2025-01-20
fix: stats column binary_column has unsupported type binary

omkar-foss:
# Description
Handle `BINARY` column appropriately when specified in `delta.dataSkippingStatsColumns` while writing to a delta table.

# Related Issue(s)

- closes https://github.com/delta-io/delta-rs/issues/2854
- related https://github.com/delta-io/delta-rs/issues/2373

# Related PR(s)

- https://github.com/delta-io/delta-rs/pull/1766
- https://github.com/delta-io/delta-rs/pull/1498 (reverted in PR https://github.com/delta-io/delta-rs/pull/1544)
- 

# Documentation

N/A

### DIFF
--- a/crates/core/src/kernel/snapshot/mod.rs
+++ b/crates/core/src/kernel/snapshot/mod.rs
@@ -25,6 +25,7 @@ use futures::stream::BoxStream;
 use futures::{StreamExt, TryStreamExt};
 use object_store::path::Path;
 use object_store::ObjectStore;
+use tracing::warn;
 
 use self::log_segment::{LogSegment, PathExt};
 use self::parse::{read_adds, read_removes};
@@ -652,23 +653,32 @@ fn stats_schema(schema: &StructType, config: TableConfig<'_>) -> DeltaResult<Str
     let stats_fields = if let Some(stats_cols) = config.stats_columns() {
         stats_cols
             .iter()
-            .map(|col| match get_stats_field(schema, col) {
-                Some(field) => match field.data_type() {
-                    DataType::Map(_) | DataType::Array(_) | &DataType::BINARY => {
-                        Err(DeltaTableError::Generic(format!(
-                            "Stats column {col} has unsupported type {}",
-                            field.data_type()
-                        )))
+            .filter_map(|col| match get_stats_field(schema, col) {
+                Some(field) => {
+                    let is_binary_type = matches!(field.data_type(), &DataType::BINARY);
+                    if is_binary_type {
+                        warn!("Column {} is of binary type and excluded from loading in stats columns.", col);
+                        return None
                     }
-                    _ => Ok(StructField::new(
-                        field.name(),
-                        field.data_type().clone(),
-                        true,
-                    )),
-                },
-                _ => Err(DeltaTableError::Generic(format!(
-                    "Stats column {col} not found in schema"
-                ))),
+                    Some(match field.data_type() {
+                        DataType::Map(_) | DataType::Array(_) => {
+                            Err(DeltaTableError::Generic(format!(
+                                "Stats column {col} has unsupported type {}",
+                                field.data_type()
+                            )))
+                        }
+                        _ => Ok(StructField::new(
+                            field.name(),
+                            field.data_type().clone(),
+                            true,
+                        )),
+                    })
+                }
+                _ => {
+                    Some(Err(DeltaTableError::Generic(format!(
+                        "Stats column {col} not found in schema"
+                    ))))
+                }
             })
             .collect::<Result<Vec<_>, _>>()?
     } else {
--- a/crates/core/src/operations/write/mod.rs
+++ b/crates/core/src/operations/write/mod.rs
@@ -31,7 +31,7 @@ pub(crate) mod metrics;
 pub(crate) mod schema_evolution;
 pub mod writer;
 
-use arrow_schema::Schema;
+use arrow_schema::{DataType, Schema};
 pub use configs::WriterStatsConfig;
 use datafusion::execution::SessionStateBuilder;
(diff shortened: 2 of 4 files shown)


ion-elgreco:
@omkar-foss are you still working on a fix :)?

omkar-foss:
I checked the delta lake protocol, and it looks like the stats are processed only for numeric columns types and stored in this structure: https://github.com/delta-io/delta/blob/master/PROTOCOL.md?plain=1#L1942-L1955

Since minValues and maxValues cannot be computed for BINARY columns, one option would be that we keep BINARY columns excluded from column statistics altogether.

Another option could be that we support BINARY columns and only compute numRecords and nullCount for them. minValues and maxValues will be unavailable.

@ion-elgreco any thoughts on which way we should go? My inclination is towards keeping them excluded.

ion-elgreco:
@omkar-foss let's exclude it then yeah, maybe add a warm trace when such a column was set

omkar-foss:
> @omkar-foss let's exclude it then yeah, maybe add a warm trace when such a column was set

Yes sure, I'll update this PR accordingly. Thank you :)

omkar-foss:
> > @omkar-foss let's exclude it then yeah, maybe add a warm trace when such a column was set
> 
> Yes sure, I'll update this PR accordingly. Thank you :)

@ion-elgreco I've updated the PR, also added a few tests on python and rust side to cover the fix. Let me know if any further changes or improvements, thank you :)

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -652,9 +653,20 @@ fn stats_schema(schema: &StructType, config: TableConfig<'_>) -> DeltaResult<Str
     let stats_fields = if let Some(stats_cols) = config.stats_columns() {
         stats_cols
             .iter()
+            .filter(|col| {
```
ion-elgreco, line 656: Maybe use filter_map here

python/src/lib.rs
```diff
@@ -390,7 +391,22 @@ impl RawDeltaTable {
                 .map_err(PythonError::from)?
                 .config()
                 .stats_columns()
-                .map(|v| v.iter().map(|s| s.to_string()).collect::<Vec<String>>()))
+                .map(|v| v.iter().map(|s| s.to_string())
+                .filter(| s | {
```
ion-elgreco, line 395: We shouldnt filter here actually

crates/core/src/operations/write/mod.rs
```diff
@@ -599,9 +599,33 @@ impl std::future::IntoFuture for WriteBuilder {
             let target_file_size = this.target_file_size.or_else(|| {
                 Some(super::get_target_file_size(&config, &this.configuration) as usize)
             });
-            let (num_indexed_cols, stats_columns) =
+            let (num_indexed_cols, mut stats_columns) =
                 super::get_num_idx_cols_and_stats_columns(config, this.configuration);
 
+            stats_columns = match stats_columns {
+                Some(columns) => columns.iter().filter(| s | {
+                    let (_index, field) = match schema.column_with_name(s) {
```
ion-elgreco, line 607: Can we do the filtering perhaps inside the stats generation function? I think that will simplify it and would require the check only there

omkar-foss: Sure, can you share the name of the function you're suggesting? I'll have a look.

python/src/lib.rs
omkar-foss: It looks like if we don't filter here, then the result of `get_stats_columns()` returns the binary column as well. Without a filter here, [this test](https://github.com/omkar-foss/delta-rs/blob/fix-dssc-binary-col/python/tests/test_writer.py#L1952-L1974) fails with the following error:
```bash
>       assert dt._table.get_stats_columns() == ["y"]
E       AssertionError: assert ['x', 'y'] == ['y']
E         
E         At index 0 diff: 'x' != 'y'
E         Left contains one more item: 'y'
E         
E         Full diff:
E           [
E         +     'x',
E               'y',
E           ]

tests/test_writer.py:1974: AssertionError
```

crates/core/src/kernel/snapshot/mod.rs
omkar-foss: Done, thanks!

crates/core/src/writer/record_batch.rs
```diff
@@ -1080,6 +1118,54 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_write_dssc_with_binary_col() {
```
omkar-foss, line 1122: @ion-elgreco this is the test associated with noop for binary cols while writing stats to json.

omkar-foss:
@ion-elgreco I've made the necessary changes to this PR as per our discussion. Please have a look when you get a chance, let me know any further changes required. Thanks! :)

ion-elgreco (approved):
Thanks for sticking with it 😄

omkar-foss:
@ion-elgreco looks like we're getting the same intermittent `BufError` failures again ([here](https://github.com/delta-io/delta-rs/actions/runs/13775527214/job/38540888622?pr=3146#step:5:116)), can you restart the failed ones?

omkar-foss:
The `BufError` [here](https://github.com/delta-io/delta-rs/actions/runs/13775527214/job/38540888622?pr=3146#step:5:116) seems related to this uv issue: https://github.com/astral-sh/uv/issues/8692

Probably we need to update the uv version in GitHub CI checks.

omkar-foss:
@ion-elgreco the [PyPI incident](https://status.python.org/incidents/2zf08fkp0nd1) shows as resolved now, shall we attempt to the re-run failing CI checks and try it out?

ion-elgreco (approved):

PR #3148, opened 2025-01-20
chore: remove lazy_static and once_cell dependencies

nyurik:
# Description
[LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) has been available since 1.80 which is less than MSRV


roeap (approved):
Thanks for taking care of this!! Always great to get rid of dependencies ❤️.

roeap:
@nyurik - looks like we need one more `cargo fmt` :).

rtyler (approved):

PR #3150, opened 2025-01-21
feat: expand ListingSchemaProvider to support register and deregister table

Nordalf:
# Description
The `ListingSchemaProvider` was missing the implementations of the register and deregister traits from the `SchemaProvider`. These are now implemented, and to meet requirements from remote object stores, the `DeltaTableBuilder` is now used to pass along the storage options.

The `tables` property has been modified to instead of having `DashMap<String, String>`, it is now using `DashMap<String, Arc<dyn TableProvider>>` to support `downcast_ref` and mutability elsewhere.

Inspiration from: [Datafusion ListingSchemaProvider](https://github.com/apache/datafusion/blob/main/datafusion/core/src/catalog_common/listing_schema.rs)

# Related Issue(s)
There is no related issues. I have a clone of this with extra modifications in my own project. I find it useful to have when adding a new schema and wanted to share this with others 👍 

# Documentation
None

### DIFF
--- a/crates/core/src/data_catalog/storage/mod.rs
+++ b/crates/core/src/data_catalog/storage/mod.rs
@@ -13,9 +13,9 @@ use futures::TryStreamExt;
 use object_store::ObjectStore;
 
 use crate::errors::DeltaResult;
-use crate::open_table_with_storage_options;
-use crate::storage::*;
 use crate::table::builder::ensure_table_uri;
+use crate::DeltaTableBuilder;
+use crate::{storage::*, DeltaTable};
 
 const DELTA_LOG_FOLDER: &str = "_delta_log";
 
@@ -36,7 +36,7 @@ pub struct ListingSchemaProvider {
     /// Underlying object store
     store: Arc<dyn ObjectStore>,
     /// A map of table names to a fully quilfied storage location
-    tables: DashMap<String, String>,
+    tables: DashMap<String, Arc<dyn TableProvider>>,
     /// Options used to create underlying object stores
     storage_options: StorageOptions,
 }
@@ -73,6 +73,7 @@ impl ListingSchemaProvider {
                 parent = p;
             }
         }
+
         for table in tables.into_iter() {
             let table_name = normalize_table_name(table)?;
             let table_path = table
@@ -80,10 +81,34 @@ impl ListingSchemaProvider {
                 .ok_or_else(|| DataFusionError::Internal("Cannot parse file name!".to_string()))?
                 .to_string();
             if !self.table_exist(&table_name) {
-                let table_url = format!("{}/{table_path}", self.authority);
-                self.tables.insert(table_name.to_string(), table_url);
+                let table_url = format!("{}/{}", self.authority, table_path);
+                let Ok(delta_table) = DeltaTableBuilder::from_uri(table_url)
+                    .with_storage_options(self.storage_options.0.clone())
+                    .build()
+                else {
+                    continue;
+                };
+                let _ = self.register_table(table_name, Arc::new(delta_table));
             }
         }
+        Ok(())
+    }
+
+    /// Tables are not initialized but have a reference setup. To initialize the delta
+    /// table, the `load()` function must be called on the delta table. This function helps with
+    /// that and ensures the DashMap is updated
+    pub async fn load_table(&self, table_name: &str) -> datafusion::common::Result<()> {
+        if let Some(mut table) = self.tables.get_mut(&table_name.to_string()) {
+            if let Some(delta_table) = table.value().as_any().downcast_ref::<DeltaTable>() {
+                // If table has not yet been loaded, we remove it from the tables map and add it again
+                if delta_table.state.is_none() {
+                    let mut delta_table = delta_table.clone();
+                    delta_table.load().await?;
+                    *table = Arc::from(delta_table);
+                }
+            }
+        }
+
         Ok(())
     }
 }
@@ -112,31 +137,31 @@ impl SchemaProvider for ListingSchemaProvider {
     }
 
     async fn table(&self, name: &str) -> datafusion_common::Result<Option<Arc<dyn TableProvider>>> {
-        let Some(location) = self.tables.get(name).map(|t| t.clone()) else {
+        let Some(provider) = self.tables.get(name).map(|t| t.clone()) else {
             return Ok(None);
         };
-        let provider =
-            open_table_with_storage_options(location, self.storage_options.0.clone()).await?;
-        Ok(Some(Arc::new(provider) as Arc<dyn TableProvider>))
+        Ok(Some(provider))
     }
 
     fn register_table(
         &self,
-        _name: String,
-        _table: Arc<dyn TableProvider>,
+        name: String,
+        table: Arc<dyn TableProvider>,
     ) -> datafusion_common::Result<Option<Arc<dyn TableProvider>>> {
-        Err(DataFusionError::Execution(
-            "schema provider does not support registering tables".to_owned(),
-        ))
+        if !self.table_exist(name.as_str()) {
+            self.tables.insert(name, table.clone());
+        }
+        Ok(Some(table))
     }
 
     fn deregister_table(
         &self,
-        _name: &str,
+        name: &str,
     ) -> datafusion_common::Result<Option<Arc<dyn TableProvider>>> {
-        Err(DataFusionError::Execution(
-            "schema provider does not support deregistering tables".to_owned(),
-        ))
+        if let Some(table) = self.tables.remove(name) {
+            return Ok(Some(table.1));
+        }
+        Ok(None)
     }
 
     fn table_exist(&self, name: &str) -> bool {
@@ -177,6 +202,7 @@ mod tests {
     async fn test_query_table() {
         let schema = Arc::new(ListingSchemaProvider::try_new("../test/tests/data/", None).unwrap());
         schema.refresh().await.unwrap();
+        schema.load_table("simple_table").await.unwrap();
 
         let ctx = SessionContext::new();
         let catalog = Arc::new(MemoryCatalogProvider::default());


ion-elgreco:
@hntd187 can you take a look? :)

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -54,7 +54,7 @@ impl ListingSchemaProvider {
         Ok(Self {
             authority: uri.to_string(),
             store,
-            tables: DashMap::new(),
+            tables: Mutex::new(DashMap::new()),
```
hntd187, line 57: What is the necessity of the lock? Dashmap shouldn't need a locking mechanism, it's a direct replacement for things such as this.

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -81,7 +82,14 @@ impl ListingSchemaProvider {
                 .to_string();
             if !self.table_exist(&table_name) {
                 let table_url = format!("{}/{}", self.authority, table_path);
-                self.tables.insert(table_name.to_string(), table_url);
+                let Ok(delta_table) = DeltaTableBuilder::from_uri(table_url)
+                    .with_storage_options(self.storage_options.0.clone())
+                    .load()
```
hntd187, line 87: I don't think we need to load every table on listing. You only need a list of tables, loading them gives you much more than that. Additionally, this might be a performance issue or cause a lot of unnecessary API calls if you have a very large catalog and never read 99% of the tables in it.

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -108,39 +116,54 @@ impl SchemaProvider for ListingSchemaProvider {
...
+        name: String,
+        table: Arc<dyn TableProvider>,
     ) -> datafusion_common::Result<Option<Arc<dyn TableProvider>>> {
-        Err(DataFusionError::Execution(
-            "schema provider does not support registering tables".to_owned(),
-        ))
+        self.tables
+            .lock()
+            .expect("Can't lock tables")
+            .insert(name, table.clone());
```
hntd187, line 148: This will replace the table provider if it currently exists, are you sure this is what you want?

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -108,39 +116,54 @@ impl SchemaProvider for ListingSchemaProvider {
...
     }
 
     async fn table(&self, name: &str) -> datafusion_common::Result<Option<Arc<dyn TableProvider>>> {
-        let Some(location) = self.tables.get(name).map(|t| t.clone()) else {
+        let Some(provider) = self
+            .tables
+            .lock()
+            .expect("Can't lock tables")
+            .get(name)
+            .map(|t| t.clone())
```
hntd187, line 133: would `.cloned()` suffice here instead?

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -108,39 +116,54 @@ impl SchemaProvider for ListingSchemaProvider {
     }
 
     fn table_names(&self) -> Vec<String> {
-        self.tables.iter().map(|t| t.key().clone()).collect()
+        self.tables
+            .lock()
+            .expect("Can't lock tables")
```
hntd187, line 121: If there is a reason to keep the locking mechanism I think we should handle these errors instead of an expect panic when we can't acquire a lock.

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -36,7 +36,7 @@ pub struct ListingSchemaProvider {
     /// Underlying object store
     store: Arc<dyn ObjectStore>,
     /// A map of table names to a fully quilfied storage location
-    tables: DashMap<String, String>,
+    tables: Mutex<DashMap<String, Arc<dyn TableProvider>>>,
```
roeap, line 39: do we need to wrap the `DashMap` in a `Mutex`? I thought that `DashMap` handles mutability / locking internally and can just be put in an `Arc` if it needs to be shared across threads?

Then again, I might be remembering that wrong.

hntd187 (changes requested):
I'm supportive of this change, but I would like some changes to the internals.

Nordalf:
Thanks for the great comments @roeap and @hntd187. I will take a look and handle them tomorrow 🚀

Nordalf:
> Thanks for the great comments @roeap and @hntd187. I will take a look and handle them tomorrow 🚀

This did not age well but hopefully I get some time this week to check out. I did figure out that `DashMap` already handles locking.

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Mutex not needed. Will be removed

crates/core/src/data_catalog/storage/mod.rs
Nordalf: @hntd187  - Nice comment. I did not think about that. How would you approach this instead? Use a string for the `table_uri` and then only load the table when necessary? So all you have is a reference to the location?

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Will test

crates/core/src/data_catalog/storage/mod.rs
Nordalf: The problem with this approach is that the value of the `DashMap` is an `Arc<dyn TableProvider>`, meaning when getting the value, we are not allowed to mutate it. This is required by the `.load()` function. My idea to get around this is to remove the table from the map, take ownership, execute the `.load()`, and register it again

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Locking will be removed

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Adding check for table existence

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Mutex has been removed, since DashMap handles it

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Mutex has been removed, since DashMap handles it

crates/core/src/data_catalog/storage/mod.rs
Nordalf: @hntd187  - check out the latest commit. I have removed the `load`, only building, and added a method to load the table from the provider when downcasting

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Mutex has been removed, since DashMap handles it

crates/core/src/data_catalog/storage/mod.rs
Nordalf: Check added

ion-elgreco:
@Nordalf is this ready?

Nordalf:
@ion-elgreco - to me it feels ready but I need to squash the commits first 👍 

The latest changes resolves the comments but those changes may require a quick look 😄

Nordalf:
@ion-elgreco and @rtyler - This PR should be ready

ion-elgreco:
@Nordalf there are some failed tests

Nordalf:
> @Nordalf there are some failed tests

Oh damn... taking a look at it when I get home later

Nordalf:
Looking into fixing the tests now

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -112,31 +137,31 @@ impl SchemaProvider for ListingSchemaProvider {
...
+        name: String,
+        table: Arc<dyn TableProvider>,
     ) -> datafusion_common::Result<Option<Arc<dyn TableProvider>>> {
-        Err(DataFusionError::Execution(
-            "schema provider does not support registering tables".to_owned(),
-        ))
+        if !self.table_exist(name.as_str()) {
+            self.tables.insert(name, table.clone());
+        }
+        Ok(Some(table))
```
roeap, line 154: I think the sematics here are a little bit different, i.e. we should raise if the table already exists. 

https://github.com/apache/datafusion/blob/784df33f8930f91eada0d67aa5acc25a4c25cea2/datafusion/catalog/src/schema.rs#L57-L69

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -112,31 +137,31 @@ impl SchemaProvider for ListingSchemaProvider {
...
 
     async fn table(&self, name: &str) -> datafusion_common::Result<Option<Arc<dyn TableProvider>>> {
-        let Some(location) = self.tables.get(name).map(|t| t.clone()) else {
+        let Some(provider) = self.tables.get(name).map(|t| t.clone()) else {
             return Ok(None);
         };
-        let provider =
-            open_table_with_storage_options(location, self.storage_options.0.clone()).await?;
-        Ok(Some(Arc::new(provider) as Arc<dyn TableProvider>))
+        Ok(Some(provider))
```
roeap, line 143: We need to make sure we are returning an up to date table here. Datafusion internally will not make any calls to update the log data o.a. so if any operations happend (including the ones we may have done in the same DF session) we would serve a stale snapshot.

crates/core/src/data_catalog/storage/mod.rs
```diff
@@ -73,17 +73,42 @@ impl ListingSchemaProvider {
...
+                if delta_table.state.is_none() {
+                    let mut delta_table = delta_table.clone();
+                    delta_table.load().await?;
+                    *table = Arc::from(delta_table);
+                }
             }
         }
+
         Ok(())
     }
```
roeap, line 113: I think the `table` function is the right place to create an up to date snapshot. see comment below.

roeap (changes requested):
Left a few comment.

The general Issue we are facing is that we right now eagerly load the state of the table. THings will get easier once #3137 and we have done some follow up work ...

crates/core/src/data_catalog/storage/mod.rs
Nordalf: So an idea could be to move some of the logic down to the `table` function and then ensure it is updated?

crates/core/src/data_catalog/storage/mod.rs
Nordalf: You are right on this one. I will raise the error but should the table be returned if registered then?

crates/core/src/data_catalog/storage/mod.rs
Nordalf: I like that idea - if this is what you meant, @roeap

crates/core/src/data_catalog/storage/mod.rs
Nordalf: I will change the code later to:

```
        if self.table_exist(name.as_str()) {
            return exec_err!("The table {name} already exists");
        };
        Ok(self.tables.insert(name, table))

```

crates/core/src/data_catalog/storage/mod.rs
roeap: This should work ... eventually we may just want to do that in the `scan` method of the table provider, since right now we are kind-of locking in the version for every provider.

But that would go a bit to far for this PR ...

rtyler:
Based on the work @roeap has done lately in `main` and associated pull  requests, is this a PR that we should resurrect @Nordalf ? I'm not sure how much you two have worked on this in recent changes

Issue #2844, opened 2024-09-04
Z-order is no-op for single column of strings with length >= 15


cjolowicz:
# Environment

**Delta-rs version**:

0.19.1

**Binding**: 
Python and Rust

**Environment**:
- **Cloud provider**: local filesystem and R2
- **OS**: Linux
- **Other**:

***
# Bug

**What happened**:

Apply z-order to a Delta Table on a column that contains strings with a length of 15 characters or more, for example a column with ISO 8601 timestamps. The records in the new Parquet files retain their original order.

I initially witnessed when z-ordering a large partition on timestamps with delta-rs in Rust. I've reproduced this with Python bindings and a small data frame using strings containing zero-padded integers (see repro below).

**What you expected to happen**:

The resulting Parquet files are ordered by the column specified for z-ordering. 

**How to reproduce it**:

```python
# test_zorder.py
import shutil
import pandas
from deltalake import write_deltalake, DeltaTable

def test_zorder() -> None:
    table = "a"
    field = "b"
    items = [f"{item:015}" for item in [2, 3, 1]]

    shutil.rmtree(table, ignore_errors=True)

    write_deltalake(table, pandas.DataFrame({field: items}))

    DeltaTable(table).optimize.z_order([field])

    sorted_items = DeltaTable(table).to_pyarrow_table().to_pydict()[field]

    assert sorted(items) == sorted_items
```

```
uv run --with deltalake --with pandas --with pytest pytest -vv
```

Output:

```
========================= test session starts ==========================
platform linux -- Python 3.12.5, pytest-8.3.2, pluggy-1.5.0 -- /home/claudio/.cache/uv/archive-v0/A-uQ68p-4BWRUFltJ5Mv2/bin/python
cachedir: .pytest_cache
rootdir: ...
collected 1 item                                                       

test_zorder.py::test_zorder FAILED                               [100%]

=============================== FAILURES ===============================
_____________________________ test_zorder ______________________________

...
    
>       assert sorted(items) == sorted_items
E       AssertionError: assert ['000000000000001', '000000000000002', '000000000000003'] == ['000000000000002', '000000000000003', '000000000000001']
E         
E         At index 0 diff: '000000000000001' != '000000000000002'
E         
E         Full diff:
E           [
E         +     '000000000000001',
E               '000000000000002',
E               '000000000000003',
E         -     '000000000000001',
E           ]

test_zorder.py:20: AssertionError
======================= short test summary info ========================
[...]

PR #3157, opened 2025-01-24
feat: enable file merging by last modification time using preserve-insertion-order

esarili:
This change leverages the previously unused `preserve-insertion-order` configuration to enable merging files sorted by their last modification time during compaction. This is particularly beneficial for append-only workloads, improving data locality after optimize runs by merging files that were created around similar times.


hntd187:
@esarili can you sign off the commits? There should be directions in the failing CI check below. https://github.com/delta-io/delta-rs/pull/3157/checks?check_run_id=36132013214

crates/core/src/operations/optimize.rs
```diff
@@ -877,12 +878,15 @@ pub fn create_merge_plan(
     filters: &[PartitionFilter],
     target_size: Option<i64>,
     writer_properties: WriterProperties,
+    preserve_insertion_order: bool,
```
rtyler, line 881: I'm curious what you think about this going into WriterProperties rather than all these functions growing an additional argument.

Does this only benefit call paths for the create_merge_plan flow?

roeap:
as it so happens, @hntd187 and I just had a quick discussion around using the file updated times more generally. Turns out there are at least some scenarios where this may not do what one would expect it to.

For the use case at hand, I wonder if this could also be done using "single value z-order" which should just degrade to sorting for a single value while also targeting a specific file size.

This of course assumes that there is some value in the data that somehow correlates when the data was inserted. IF that were not the case, there would also not be much value though in preserving the order as query engines could not leverage this information.

crates/core/src/operations/optimize.rs
ion-elgreco: The writer properties are used in other operations, so we should only add it if it makes senso for those ops as well

esarili:
> as it so happens, @hntd187 and I just had a quick discussion around using the file updated times more generally. Turns out there are at least some scenarios where this may not do what one would expect it to.
> 
> For the use case at hand, I wonder if this could also be done using "single value z-order" which should just degrade to sorting for a single value while also targeting a specific file size.
> 
> This of course assumes that there is some value in the data that somehow correlates when the data was inserted. IF that were not the case, there would also not be much value though in preserving the order as query engines could not leverage this information.

We run optimize command in regular intervals (every two-hours, on last two partitions where data is partitioned by day). AFAIU z-order runs on entire partition and sorts data on record level which might is resource intensive and might not be necessary for the append only workflows. With using file update times, we are hoping to avoid sorting entire partition record by record but still have some amount of locality after optimize runs.

crates/core/src/operations/optimize.rs
esarili: That was my thinking as well

esarili: Do you mind if I mark this resolved?

ion-elgreco:
@cjolowicz honestly, I am having a tough time understand the use of this. Why do you need to collocate data that got inserted around the same time?

cjolowicz:
Hi @ion-elgreco 

Let me add some context here. Our use case is a multi-tenant service for storing and querying logs of various schemas. Every log has a timestamp field used for partitioning (by day) and querying (by arbitrary time frames). We ran into the following problem: Compaction shuffles log batches randomly across the compacted files, because it sorts the input files in order of file size and fills bins greedily. As a result, queries for small time frames still need to fetch a large number of files.

Our first approach was to use z-ordering on the timestamp column. Unfortunately, this approach hit several limitations:

- it's resource-intensive to sort row-by-row; typically, input batches already have good-enough data locality
- DataFusion's sorting implementation with spill-to-disk has a [bug](https://github.com/apache/datafusion/issues/12136#issuecomment-2656449956) that causes it to crash on large datasets
- the upload phase of z-ordering is sequential so it takes very long
- timestamps are long ISO 8601 strings but z-ordering [only looks at the first 16 bytes](https://github.com/delta-io/delta-rs/issues/2844#issuecomment-2660854600) of each column

The approach in this PR intends to use ingestion time (the modification time of input files) as a proxy for the timestamp column. The unimplemented `preserve_insertion_order` flag seemed a perfect place for this algorithm. Unfortunately, this doesn't take into account that compaction also merges already compacted files. The modification timestamps of these files are essentially random, since compacted files are written concurrently. So the approach in this PR doesn't work (at least, not without some way of excluding compacted files from the input).

As a third approach, we're currently looking at a compaction algorithm that uses min/max stats to merge files in order of one or more designated columns (in our case, just the timestamp column of the dataset).

OT: Would it make sense to fail `preserve_insertion_order` as unimplemented, or even remove it altogether? Currently the flag is silently ignored, and metrics always report that the flag was enabled even if it wasn't.

ion-elgreco:
@cjolowicz but as you are pointing out, preserve_insertion_order will fail to meet it's purpose after you have already compacted before. To me it seems like a duct tape solution for z-order, I would rather see a fix upstream for the diskspilling not working or an improved z-order algorithm.

Regarding sequential uploads, I would like to see your feedback whether this helps: https://github.com/delta-io/delta-rs/pull/3255

I think we should just remove this preserve_insertion_order flag all together

PR #3160, opened 2025-01-25
fix: removing openssl from the dependency tree

### LINKED ISSUE #2857: Transitive dependencies are still vendoring openssl even when you don't want them to.
# Environment

**Delta-rs version**: 0.19.2

**Binding**:  python

***
# Bug

**What happened**:

Basically it seems that something in our transitive graph is still pulling in a full openssl build even when the `native-tls` feature is not specified.

Related to #2856 

**How to reproduce it**:

**More details**:

rtyler:
This is still marked as draft for the time being because there is not a released version of delta-kernel-rs which has the `default-engine-rustls` flag _and_ a compatible set of arrow dependencies :see_no_evil: 

Fixes #2857

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
-delta_kernel = { version = "=0.6.1", features = ["default-engine"] }
+delta_kernel = { version = "=0.6.1", features = ["default-engine-rustls"] }
 #delta_kernel = { path = "../delta-kernel-rs/kernel", features = ["sync-engine"]  }
 
 # arrow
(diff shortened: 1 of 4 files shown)


gotocoding-DB:
@rtyler Hi, Tyler, could you please explain little more about: "which has the default-engine-rustls flag and a compatible set of arrow dependencies".
I can't get which dependency exactly is the problem.
E.g. this works for me:
```
[dependencies]
delta_kernel = { version = "0.6.1", features = [
    "cloud",
    "default-engine-rustls",
    "developer-visibility",
    "sync-engine",
] }
arrow = { version = "54.0.0", features = ["prettyprint", "chrono-tz"] }
arrow-schema = { version = "54.0.0" }
arrow-array = { version = "54.0.0", features = ["chrono-tz"] }
```

ion-elgreco (approved):

PR #3162, opened 2025-01-25
fix: wrap future streams also in io_runtime

ion-elgreco:
@alamb pointed out that the get stream would be not executed on the IO runtime unless it was wrapped. So thanks to his example PR I've added this now for the other streams :) Thanks @alamb! 

Also addressed some other missing areas where I should have enabled the IO runtime:

- DeltaFileSystemHandler
- Creating a table
- Is Delta Table check
- Writing a new table (used by pyarrow engine)
- Writing a new table (rust engine)

I've also made it possible to pass an optional runtime in DeltaOps::try_from_uri, since this was missing but it seems a common used entry point to open a delta table


ion-elgreco:
@Tom-Newton can you give this a spin?

alamb:
I plan to give this a look later today

ion-elgreco:
@Tom-Newton what is your timeout set to?

Tom-Newton:
> @Tom-Newton what is your timeout set to?

`deltalake.DeltaTable(path, storage_options={"timeout": "120s"})`

ion-elgreco:
@Tom-Newton can you set it to a very large interval, 30 minutes or something?

Tom-Newton:
> @Tom-Newton can you set it to a very large interval, 30 minutes or something?

I gave it a try with 30 minutes. The success rate is not noticeably better but we do get a different error message. 
```
      File "/redacted/pip-svc_deltalake/site-packages/deltalake/table.py", line 1270, in update_incremental
    OSError: Generic MicrosoftAzure error: Error after 0 retries in 993.500001046s, max_retries:10, retry_timeout:180s, source:error sending request for url (https://redacted.blob.core.windows.net/redacted/_delta_log/00000000000000146433.json)
```
Potentially this is useful information, but I have not yet had a chance to investigate further. It sounds like in this case Azure is returning and error response, which would likely be informative.

ion-elgreco:
@Tom-Newton this is during reading I presume?

docs/usage/optimize/small-file-compaction-with-optimize.md
```diff
@@ -124,7 +124,7 @@ Let’s write 100 hours worth of data to the Delta table.
 
 === "Rust"
     ```rust
-    let mut table = DeltaOps::try_from_uri("observation_data")
+    let mut table = DeltaOps::try_from_uri("observation_data", IORuntime::default())
```
rtyler, line 127: I believe this and the other documentation change are supposed to be wrapped in a `Some`

rtyler (changes requested):
I would like to hold off on this change being merged for the time being. I need to understand the consequences of the changes to the core storage module here.

The :spaghetti: we have floating around to accommodate asynchronous behaviors with Python is creeping further into core here, and that gives me a lot of concern so I need to spend a fair bit more time evaluating this change

alamb:
> > I would like to hold off on this change being merged for the time being. I need to understand the consequences of the changes to the core storage module here.
> > The 🍝 we have floating around to accommodate asynchronous behaviors with Python is creeping further into core here, and that gives me a lot of concern so I need to spend a fair bit more time evaluating this change
> 
> Yeah I'm fine with that, the discussion where this started from is also still ongoing: [apache/datafusion#14286 (comment)](https://github.com/apache/datafusion/pull/14286#issuecomment-2626852759) so I don't think there is a full consensus on how to solve this topic yet, right @alamb?

Yes I agree there is not yet consensus on the topic. We also don't fully understand the consequences / implications of doing these cross runtime shenanigans -- as we do that with other projects hopefully we'll come up with a good pattern that we can then share here as well

Issue #2968, opened 2024-10-31
rust engine consume a lot of memory compared to pyarrow


djouallah:
# Environment

**Delta-rs version**:
0.21.0

**Binding**: 

**Environment**:
- **OS**:
Linux

***
# Bug

switching from pyarrow engine to rust increase memory usage by nearly 3X, the job used to works fine, but now, getting OOM errors.

I added a reproducible example with only 60 input files to demo the issue

https://colab.research.google.com/drive/1fahlV0FgKSAS8sQvRMu47s3bDP1ekLbb#scrollTo=333a177b-f075-412e-8ca1-32d44f8c07eb

rtyler:
@djouallah :wave: in the attached notebook, which write_deltalake call is resulting in memory pressure?  I'm less familiar with duckdb, but I assume the `df`  objects that it is producing are pyarrow.DataSet? or are they another type?

djouallah:
it is an Arrow RecordBatchReader I think

ion-elgreco:
Rust engine materalizes everything to memory prior to starting the whole process. Pyarrow probably writes batch by batch

djouallah:
just for my own understanding, is this something that can be fixed by datafusion ?

ion-elgreco:
@djouallah there is a PR to address this but the contributor didn't have time to finish it yet: https://github.com/delta-io/delta-rs/pull/2289

djouallah:
using 0.22, I just use the pyarrow engine, yes, processing 2300 csv files and it works great !!!
<img width="388" alt="image" src="https://github.com/user-attachments/assets/b7ccfabf-035a-4778-b607-ea84ac137ac8">

ldacey:
Yeah, I saw the "pyarrow" deprecation message and swapped to the rust engine for one of my tables. It is only about 2.6 million rows per partition but it is very wide and all text data (200+ columns). The pods were failing due to OOM errors after doing this so I need to use the pyarrow engine and overwrite the partition still.

rtyler:
I'm returning to this and I have some exciting developments! Datafusion 44 introduced [LazyMemoryExec](https://docs.rs/datafusion/latest/datafusion/physical_plan/memory/struct.LazyMemoryExec.html) which is the missing piece to make this a _lot_ simpler.

My prior attempts at solving this problem were effectively in a similar solution domain as `LazyMemoryExec`, except I was blurring the lines between Python and Rust more than was probably safe or sane :laughing: 


[...]

Issue #2522, opened 2024-05-17
Memory leak on writes/merges


echai58:
# Environment

**Binding**: python

***
# Bug

**What happened**:
We're noticing constantly rising memory in our processes that write to deltalake. I wrote a minimal reproduction that loops and writes to deltalake, and the memory usage seems to indicate a memory leak.

**What you expected to happen**:
Memory to be reclaimed after writes.

**How to reproduce it**:
This is my script I tested with:
```
import datetime
from deltalake import write_deltalake, DeltaTable
import numpy as np
import pandas as pd 
import pyarrow as pa

schema = pa.schema(
    [
        ("date", pa.date32()),
        ("c1", pa.string()),
        ("c2", pa.int32()),
        ("v1", pa.float64()),
        ("v2", pa.float64()),
        ("v3", pa.int32()),
    ]
)

def write_table(dt: DeltaTable, date: datetime.date):
    num_rows = 150
    data = pd.DataFrame(
        {
            "date": np.repeat(date, num_rows),
            "c1": [f"str_{i}" for i in np.random.randint(0, 100, num_rows)],
            "c2": np.random.randint(1, 100, num_rows),
            "v1": np.random.rand(num_rows) * 100,
            "v2": np.random.rand(num_rows) * 100,
            "v3": np.random.randint(0, 100, num_rows),
        }
    )
    dt.merge(
        pa.Table.from_pandas(data, schema=schema),
        "s.date = t.date and s.c1 = t.c1 and s.c2 = t.c2",
        source_alias="s",
        target_alias="t",
    ).when_matched_update_all().when_not_matched_insert_all().execute()
    if dt.version() % 10 == 0:
        dt.create_checkpoint()

def main():
    DeltaTable.create("test", schema)

    for date in pd.date_range("2022-01-01", periods=25):
        for _ in range(50):
            write_table(DeltaTable("test"), date.date())

if __name__ == "__main__":
    main()
```
**More details**:
Here's the memray graph:
![image](https://github.com/delta-io/delta-rs/assets/56415623/7c25fc24-3a6d-4448-a94f-bf29225c68cb)

I also tested this with just `write_deltalake(mode="append")`, and the issue seems to also persist:
![image](https://github.com/delta-io/delta-rs/assets/56415623/ffaa5d00-8f14-479f-925a-53e83a4a5c46)

I saw https://github.com/delta-io/delta-rs/issues/2068 and tried setting that env var, and got the following (doesn't seem to help):
[...]

PR #3196, opened 2025-02-08
feat: use LazyTableProvider by default for write_to_deltalake for memory efficiency

rtyler:
This defaults write_to_deltalake in Python to attempt to use the LazytableProvider for a more stream-like execution. It's currently opted out for schewma evolution since that's not supported by default.

Some improvements in schema mismatch detection inside of the operations::write module are required as well

- closes https://github.com/delta-io/delta-rs/issues/2968
- closes https://github.com/delta-io/delta-rs/issues/2522

### DIFF
--- a/crates/core/src/operations/write.rs
+++ b/crates/core/src/operations/write.rs
@@ -834,11 +834,22 @@ impl std::future::IntoFuture for WriteBuilder {
                 .unwrap_or_default();
             let mut schema_drift = false;
             let mut df = if let Some(plan) = this.input {
-                if this.schema_mode == Some(SchemaMode::Merge) {
-                    return Err(DeltaTableError::Generic(
-                        "Schema merge not supported yet for Datafusion".to_string(),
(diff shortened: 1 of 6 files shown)


crates/core/src/operations/write.rs
```diff
@@ -834,11 +834,29 @@ impl std::future::IntoFuture for WriteBuilder {
...
+                            "Schema merge not supported yet for Datafusion".to_string(),
+                        ));
+                    }
+                    Some(SchemaMode::Overwrite) => {}
+                    None => {
+                        if let Some(snapshot) = &this.snapshot {
+                            let table_schema = snapshot.input_schema()?;
+                            let plan_schema = plan.schema().as_arrow();
+
+                            if table_schema.fields.len() != plan_schema.fields.len() {
```
ion-elgreco, line 849: I think an not_eq comparison on the StructTypes might be better here?

PR #3202, opened 2025-02-10
feat: return metrics about the commit

HawaiianSpork:
# Description
`PostCommit` now contains the number of retries it took to do the commit and `FinalizeCommit` now contains number of retries, whether there was a checkpoint created and number of log files cleaned-up.

This can be useful for consumers to track if the retries are approaching the maximum retry count.  It also allows consumers to correlate commit times with the amount of work being done for that commit.


ion-elgreco:
@HawaiianSpork can you fix the fmt and add a small test?

rtyler (approved):

PR #3204, opened 2025-02-10
Add option to disable conflict checker in CommitProperties

### LINKED ISSUE #3186: Delta Version 0.24 no longer allows disabling the conflict checker
# Description

**Use Case**
I would like to disable use of the ConflictChecker and simply throw an error if the user tries to update the table with a stale version. I used to do this by setting `max_retries=1` in `CommitProperties`. In version `0.20.1` this would work because a conflict would be detected but we would never retry and apply the conflict checker's resolution. However, in more recent versions, the conflict checker is called BEFORE the attempt for the first commit which makes disabling the functionality impossible. It would be very much appreciated if a `CommitProperty` or something similar could allow us to specify whether we want to enable the Conflict Checker's behavior.

**Related Issue(s)**

daphnenhuch-at:
# Description
Add option to disable conflict checker in CommitProperties

# Related Issue(s)
Closes #3186 3186

# Documentation


ion-elgreco (changes requested):

daphnenhuch-at:
Question: When do you expect the next official release of deltalake to be released?

ion-elgreco:
Looks good, can you add a test for this new behaviour please?

daphnenhuch-at:
Added a test!

ion-elgreco:
@daphnenhuch-at looks good, can you squash the commits into one please. Then we can merge it

hntd187:
It's not clear to me why you would want disable the conflict checker. I don't have any issues with this PR, I'm more curious what the use case is.

daphnenhuch-at:
> It's not clear to me why you would want disable the conflict checker. I don't have any issues with this PR, I'm more curious what the use case is.

Our usage wants to disable any concurrent writes and that requires disabling the conflict checker and just erroring any time a conflicting/concurrent write occurs

crates/core/src/operations/transaction/mod.rs
```diff
@@ -604,14 +604,22 @@ impl<'a> std::future::IntoFuture for PreparedCommit<'a> {
             // unwrap() is safe here due to the above check
             let read_snapshot = this.table_data.unwrap().eager_snapshot();
 
-            let mut attempt_number = 1;
-            while attempt_number <= this.max_retries {
+            let mut attempt_number = 0;
```
ion-elgreco, line 607:
```suggestion
            let mut attempt_number = 1;
```
Should start at one btw

ion-elgreco (changes requested):

daphnenhuch-at:
This is good to go, could someone stamp it so that it's merge ready?

crates/core/src/operations/transaction/mod.rs
ion-elgreco: @daphnenhuch-at you resolved this, but this was still required so I committed it manually.

Since we do attempt_number <= max.retries + 1, we should counting at 1, since max_retries would be attempt number 2 that will still pass (1 retry) and attempt number 3 (retry 2) it will fail.

ion-elgreco (approved):
Thanks!

daphnenhuch-at:
@ion-elgreco When is the next official delta lake release scheduled for?

ion-elgreco:
> @ion-elgreco When is the next official delta lake release scheduled for?

I hope to get the next python release out once my open PRs are merged

daphnenhuch-at:
> > @ion-elgreco When is the next official delta lake release scheduled for?
> 
> I hope to get the next python release out once my open PRs are merged

What about a rust release?

ion-elgreco:
> > > @ion-elgreco When is the next official delta lake release scheduled for?
> > 
> > 
> > I hope to get the next python release out once my open PRs are merged
> 
> What about a rust release?

Not sure, @rtyler thoughts?

Issue #3207, opened 2025-02-11
Adhere to `delta.checkpoint.writeStatsAsStruct`


ion-elgreco:
We currently always write stats_parsed without checking the delta.checkpoint.writeStatsAsStruct flag on the config. We should check this before writing the stats_parsed.

Matthieusalor:
Seems like it would fix part of https://github.com/delta-io/delta-rs/issues/3201 - in the case where I would put `delta.checkpoint.writeStatsAsStruct` to false, is that assumption correct ? Also, shouldn't stat_parsed be limited to list columns in the skipping config ?

ion-elgreco:
> Seems like it would fix part of https://github.com/delta-io/delta-rs/issues/3201 - in the case where I would put `delta.checkpoint.writeStatsAsStruct` to false, is that assumption correct ? Also, shouldn't stat_parsed be limited to list columns in the skipping config ? 
> 

Yes.

Regarding the second question indeed, but I will address that in another fix

PR #3209, opened 2025-02-11
fix: adhere write stats configuration

ion-elgreco:
# Description
@roeap according to the protocol we should only write struct stats when enabled. Currently we were ignoring this config and always writing struct stats. The table config is now adhered when either of two are disabled.

- closes https://github.com/delta-io/delta-rs/issues/3207

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,14 @@ API that lets you query, inspect, and operate your Delta Lake with ease.
 
 ## Table of contents
 
+- [Table of contents](#table-of-contents)
 - [Quick Start](#quick-start)
 - [Get Involved](#get-involved)
 - [Integrations](#integrations)
 - [Features](#features)
+  - [Cloud Integrations](#cloud-integrations)
+  - [Supported Operations](#supported-operations)
+  - [Protocol Support Level](#protocol-support-level)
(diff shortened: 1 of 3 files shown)


roeap:
@ion-elgreco one question :).

We can certainly merge this fix the way it is implemented now, but ...

delta-kernel-rs contains some nice facilities to [transform schemas](https://github.com/delta-io/delta-kernel-rs/blob/eedfd473f334ac282d23847110f282118b2203eb/kernel/src/schema/mod.rs#L645-L661) in a more structured manner. In some cases implementations can be very simple.

https://github.com/delta-io/delta-kernel-rs/blob/eedfd473f334ac282d23847110f282118b2203eb/kernel/src/scan/data_skipping.rs#L79-L90

While a little more work, this could also brings us forward in adopting kernel etc and simplifying our codebase :).

Using this can of course also be deferred to a later PR.

ion-elgreco:
@roeap I can look into that later for sure, but let's fix some inconsistencies first in the current implementation

ion-elgreco:
@roeap can we merge this?

rtyler (approved):

Issue #3201, opened 2025-02-10
Checkpointing makes filtering on columns not in dataSkippingStatsColumns completely erroneous


Matthieusalor:
# Environment

**Delta-rs version**: pyton 0.24

- **OS**: linux
- **Other**:

***
# Bug

**What happened**: Creating a checkpoint and filtering on a column that has not been added to dateSkippingStatsColumns will consider everything before the checkpoint as invalid. 

**What you expected to happen**: This should not happen.

**How to reproduce it**:

```python3
import deltalake
import pandas as pd
from pandas.tseries.offsets import Day


df = pd.DataFrame({
    "date": [pd.Timestamp("2020-01-01"),
             pd.Timestamp("2020-01-02")],
    "ref_date": [pd.Timestamp("2020-01-01"),
             pd.Timestamp("2020-01-02")],
    "values": [1, 2]
})

deltalake.write_deltalake(
    data=df,
    table_or_uri="mytabletest",
    configuration={"delta.dataSkippingStatsColumns": "ref_date"}
)

dtable = deltalake.DeltaTable("mytabletest")
dtable.create_checkpoint()

df["date"] += Day(1)
df["ref_date"] += Day(1)

deltalake.write_deltalake(
    data=df,
    table_or_uri="mytabletest",
    mode="append",
    configuration={"delta.dataSkippingStatsColumns": "ref_date"}
)

# Return 2 rows (Missing 2)
deltalake.DeltaTable("mytabletest").to_pandas(filters=[("date", ">=", pd.Timestamp("2020-01-01"))])

# Return 4 rows (Correct)
deltalake.DeltaTable("mytabletest").to_pandas(filters=[("ref_date", ">=", pd.Timestamp("2020-01-01"))])

# Return 2 rows (Missing 2)
deltalake.DeltaTable("mytabletest").to_pandas(filters=[("values", ">=", 0)])
```

**More details**:

When inspecting the partition_expression created brefore and after checkpointing we see this :
((is_valid(ref_date) and (ref_date >= 2020-01-03 00:00:00.000000)) and (ref_date <= 2020-01-04 00:00:00.000000))

((((((is_valid(ref_date) and (date >= null[timestamp[us]])) and (ref_date >= 2020-01-01 00:00:00.000000)) and (values >= null[int64])) and (date <= null[timestamp[us]])) and (ref_date <= 2020-01-02 00:00:00.000000)) and (values <= null[int64]))

[...]

Issue #3173, opened 2025-01-30
File pruning issue when filtering on columns without stats and presence of struct fields


mrjsj:
# Environment

**Delta-rs version**:
0.24.0

**Binding**: 
Python

**Environment**:
- **Cloud provider**: N/A
- **OS**: MacOS 15.2
- **Other**:

***
# Bug

**What happened**:
Filtering a delta table converted to a pyarrow dataset on a column without calculated stats AND if there is a `struct` field in the columns, it seems to wrongly prune files.

**What you expected to happen**:
If it can't find stats for the column used for filtering, it shouldn't prune the file.

**How to reproduce it**:
First force delta table to write only stats for the first 32 columns. There won't be calculated stats on year, month and day.
Then we convert the delta table to a pyarrow dataset, and try filtering on the `year` column. We expect to show one record, but none are shown

When there is a struct field in the first 32 columns, and stats are not calculated for year, the filter **does not** work properly.
```python
from deltalake import write_deltalake, DeltaTable
from polars import Schema, List, Struct, Int64, String
import polars as pl

schema = Schema(
    {
        "1": String,
        "2": String,
        "3": String,
        "4": String,
        "5": String,
        "6": String,
        "7": String,
        "8": String,
        "9": String,
        "10": String,
        "11": String,
        "12": String,
        "13": String,
        "14": String,
        "15": String,
        "16": String,
        "17": String,
        "18": String,
        "19": String,
        "20": String,
        "21": String,
        "22": String,
        "23": String,
        "24": String,
        "25": String,
        "26": String,
        "27": String,
        "28": String,
        "29": String,
        "30": String,
        "nested": Struct(
            {
                "31": String,
                "32": String,
            }
        ),
        "year": Int64,
        "month": Int64,
        "day": Int64,
    }
)

df = pl.DataFrame(
    {
        "1": ["foo"],
        "2": ["foo"],
        "3": ["foo"],
        "4": ["foo"],
        "5": ["foo"],
        "6": ["foo"],
        "7": ["foo"],
        "8": ["foo"],
        "9": ["foo"],
        "10": ["foo"],
        "11": ["foo"],
        "12": ["foo"],
        "13": ["foo"],
        "14": ["foo"],
        "15": ["foo"],
        "16": ["foo"],
        "17": ["foo"],
        "18": ["foo"],
        "19": ["foo"],
        "20": ["foo"],
        "21": ["foo"],
        "22": ["foo"],
        "23": ["foo"],
[...]

PR #3210, opened 2025-02-11
fix: only use stats for required cols

ion-elgreco:
# Description

# Related Issue(s)
- fixes https://github.com/delta-io/delta-rs/issues/3173
- fixes https://github.com/delta-io/delta-rs/issues/3201

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -6,6 +6,8 @@ mod query;
 mod schema;
 mod utils;
 
+use core::num;
+use std::cmp::min;
 use std::collections::{HashMap, HashSet};
 use std::ffi::CString;
 use std::future::IntoFuture;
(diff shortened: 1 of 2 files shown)


python/src/lib.rs
```diff
@@ -1058,6 +1060,31 @@ impl RawDeltaTable {
             )),
             None => None,
         };
+        let stats_cols = self.get_stats_columns()?;
+        let num_index_cols = self.get_num_index_cols()?;
+
+        let inclusion_stats_cols = if let Some(stats_cols) = stats_cols {
+            stats_cols
+        } else if num_index_cols == -1 {
```
hntd187, line 1068: So if no columns are set collect stats for every column? This might be expensive and users might not realize this, might be better to bound it to like 32 (like databricks does) and let the num index cols go higher.

hntd187:
Sorry @ion-elgreco I removed this from the merge queue just to get your opinion about the change I proposed above.

python/src/lib.rs
ion-elgreco: If no columns are set, we use num_index_cols, where the default is indeed 32. But if it's -1 then it's use all column stats

ion-elgreco: https://github.com/delta-io/delta-rs/blob/cf5f38af409c8d5e76e1fc7ad6c8eee8601220fc/crates/core/src/table/config.rs#L43-L49

Issue #3165, opened 2025-01-27
`load_cdf` can't read last version


ion-elgreco:
# Environment

**Delta-rs version**: latest on main

**Binding**: python

# Bug

**What happened**:
When you try to read latest version with load_cdf it fails.

**What you expected to happen**:
Be able to read the latest version, if you read latest -1 only it works fine.

**How to reproduce it**:
```python
from deltalake import DeltaTable
import polars as pl

pl.DataFrame({"foo":[1,2,3]}).write_delta(
    "foo_table", 
    mode='append', 
    delta_write_options = {"configuration": {"delta.enableChangeDataFeed": "true"}} )

DeltaTable("foo_table").load_cdf(starting_version=0, ending_version=0, allow_out_of_range=False).read_all()
```
```DeltaError: Invalid table version: 0```

@hntd187 can you look into this?

PR #3218, opened 2025-02-14
fix: load cdf latest version

ion-elgreco:
# Description
The eq check was incorrect, should be only greater than.

Also found small bug in get_latest_version. It would return a version that might be invalid if the start version is not existent. Add a robustness check in there and one recursion loop, to start from -1. This will either result in the correct version or stay -1 an throw not a table location :)

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/3165


hntd187 (approved):
Looks good

hntd187 (dismissed):

hntd187 (approved):

PR #3219, opened 2025-02-14
fix(rust, python): treat FSCK files_removed as strings

### LINKED ISSUE #3140: Spark history read cannot deserialize metrics after FSCK Repair Run
# Environment
Local, S3

**Delta-rs version**: 0.24.0

**Binding**: Python

**Environment**:
- **Cloud provider**: Happens locally and on S3
- **OS**: NA
- **Other**: NA

***
# Bug

**What happened**:
Following an FSCK repair run from this library, a follow up lookup on the table history via Spark fails with the following error:
```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.lang.String` from Array value (token `JsonToken.START_ARRAY`)
```

I fixed a similar bug with optimize runs early last year: https://github.com/delta-io/delta-rs/pull/2317

So I believe the fix here is similar (looking into it, but no luck so far). 

I believe the specific issue is due to the repair run outputting `files_removed` as an array and not a string: 
```
"files_removed":["part-00001-aa65f683-d845-408a-aa3d-2a3bd04e0f8a-c000.snappy.parquet"]
```

**What you expected to happen**: Following a repair run, delta-rs will write out the `files_removed` metric in a way Spark understands.

**How to reproduce it**:

Here is an MRE:

```
import pandas as pd
import time

# write some data into a delta table
df = pd.DataFrame({"id": [1, 2], "value": ["foo", "boo"]})
write_deltalake("./data/delta", df, mode="append")
# second write so there's a file to delete manually
write_deltalake("./data/delta", df, mode="append")

# Load data from the delta table
dt = DeltaTable("./data/delta")

print("Take a moment to delete one of the parquet files written out")
time.sleep(10)

dt.repair()

### SPARK SECTION ###
from delta.tables import DeltaTable

# Initialize Spark session with Delta support
spark = SparkSession.builder \
    .appName("DeltaTableHistory") \
    .config("spark.jars.packages", "io.delta:delta-spark_2.12:3.2.1,org.apache.hadoop:hadoop-aws:3.2.1") \
    .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension") \
    .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog") \
    .getOrCreate()

delta_table_path = "./data/delta"
delta_table = DeltaTable.forPath(spark, delta_table_path)

# Get the history of the Delta table, fails here
history_df = delta_table.history()
```

**More details**:

From what I recall, I understand that delta-rs is technically following protocol spec. However from a potentially selfish perspective, this bug caused a breaking change in our Spark streaming jobs (which use table history under the hood) and I believe the best solution for now is to write this out in a way Spark understands.

liamphmurphy:
# Description
This changes how the files_removed array is serialized, making it a stringified representation of the list of files. This is done to ensure compatibility with Spark.

To maintain the current behavior with the python bindings, the repair method wrapper in python will deserialize back to a list.

# Related Issue(s)
 closes #3140

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -69,6 +69,8 @@ MAX_SUPPORTED_READER_VERSION = 3
 NOT_SUPPORTED_READER_VERSION = 2
 SUPPORTED_READER_FEATURES = {"timestampNtz"}
 
+FSCK_METRICS_FILES_REMOVED_LABEL = "files_removed"
+
 FilterLiteralType = Tuple[str, str, Any]
 FilterConjunctionType = List[FilterLiteralType]
 FilterDNFType = List[FilterConjunctionType]
@@ -1430,7 +1432,11 @@ class DeltaTable:
             commit_properties,
             post_commithook_properties,
         )
-        return json.loads(metrics)
+        deserialized_metrics = json.loads(metrics)
+        deserialized_metrics[FSCK_METRICS_FILES_REMOVED_LABEL] = json.loads(
+            deserialized_metrics[FSCK_METRICS_FILES_REMOVED_LABEL]
+        )
+        return deserialized_metrics
 
     def transaction_versions(self) -> Dict[str, Transaction]:
         return self._table.transaction_versions()
--- a/crates/core/src/operations/filesystem_check.rs
+++ b/crates/core/src/operations/filesystem_check.rs
@@ -22,7 +22,7 @@ use futures::future::BoxFuture;
 use futures::StreamExt;
 pub use object_store::path::Path;
 use object_store::ObjectStore;
-use serde::Serialize;
+use serde::{de::Error as DeError, Deserialize, Deserializer, Serialize, Serializer};
 use url::{ParseError, Url};
 use uuid::Uuid;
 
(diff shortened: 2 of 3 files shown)


python/deltalake/table.py
```diff
@@ -1430,7 +1432,9 @@ def repair(
             commit_properties,
             post_commithook_properties,
         )
-        return json.loads(metrics)
+        deserialized = json.loads(metrics)
+        deserialized[FSCK_METRICS_FILES_REMOVED_LABEL] = json.loads(deserialized["files_removed"])
+        return deserialized
```
liamphmurphy, line 1437: if we don't care about maintaining the current response then we can omit this step, but for sake of breaking the interface, added a deserializing step here.

ion-elgreco:
@liamphmurphy can you squash the commits please

liamphmurphy:
> @liamphmurphy can you squash the commits please

Done! Let me know if something isn't looking right.

rtyler (approved):

rtyler (approved):

PR #3222, opened 2025-02-16
fix: handle v2 uuid named json/parquet checkpoints

### LINKED ISSUE #3215: Error reading V2 UUID-named checkpoint
The log segment code for looking up checkpoint files defines in [log_segment.rs](https://github.com/delta-io/delta-rs/blob/2f1a5ac58126582440b8cba0f29eef13ac13f288/crates/core/src/kernel/snapshot/log_segment.rs#L26)
``
static CHECKPOINT_FILE_PATTERN: LazyLock<Regex> =
    LazyLock::new(|| Regex::new(r"\d+\.checkpoint(\.\d+\.\d+)?\.parquet").unwrap());
``

It doesn't properly parse a UUID-named Checkpoint like: 00000000000000000010.checkpoint.80a083e8-7026-4e79-81be-64bd76c43a11.json

So these checkpoints aren't recognized as such and when trying to open a table we get something like:
``
Error: MetadataError("Number of checkpoint files '0' is not equal to number of checkpoint metadata parts 'None'")
``

zeevm:
# Description
Match V2 UUID named checkpoints and read JSON checkpoints

# Related Issue(s)
- closes #3215

### DIFF
--- a/crates/core/src/kernel/snapshot/log_segment.rs
+++ b/crates/core/src/kernel/snapshot/log_segment.rs
@@ -25,6 +25,9 @@ const LAST_CHECKPOINT_FILE_NAME: &str = "_last_checkpoint";
 
 static CHECKPOINT_FILE_PATTERN: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"\d+\.checkpoint(\.\d+\.\d+)?\.parquet").unwrap());
+static UUID_CHECKPOINT_FILE_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"\d+\.checkpoint([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})\.(parquet|json)").unwrap()
+});
 static DELTA_FILE_PATTERN: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^\d+\.json$").unwrap());
 static CRC_FILE_PATTERN: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^(\.\d+(\.crc|\.json)|\d+)\.crc$").unwrap());
@@ -50,7 +53,10 @@ pub(crate) trait PathExt {
     /// Returns true if the file is a checkpoint parquet file
     fn is_checkpoint_file(&self) -> bool {
         self.filename()
-            .map(|name| CHECKPOINT_FILE_PATTERN.captures(name).is_some())
+            .map(|name| {
+                CHECKPOINT_FILE_PATTERN.captures(name).is_some()
+                    || UUID_CHECKPOINT_FILE_PATTERN.captures(name).is_some()
+            })
             .unwrap_or(false)
     }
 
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@zeevm could you add some tests?

zeevm:
@ion-elgreco 
I don't have a V2 UUID checkpoint delta table sample I can use for a UT. I have one such table I got from a customer to debug an issue they're having but obviously I can't share that as a sample table.
I've experimented with databricks but couldn't get it to generate such checkpoint file, instead I got something like "00000000000000000001.00000000000000000006.compacted.json" as a checkpoint.

If anyone can tell me exactly how to generate such table in databricks I'll do it, otherwise I don't see how to add a UT, all I can do is verify the fix locally with my customers' table.

rtyler:
@zeevm I took a look at this pull request this morning, and I''m not sure if we can safely merge it. I believe these uuid checkpoints are [v2 checkpoints](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#v2-spec) which [can be enabled via a table feature](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#v2-spec).

They're structurally different than v1 checkpoints and may include sidecars which contain additional checkpoint data that I _believe_ delta-rs will currently ignore. 

Without sharing the table data (obviously :laughing:) can you share the error that was run into? I can imagine a scenario where older versions of the transaction log were cleaned up and the historical information on a table would be in v2 checkpoints causing us problems reading the table :thinking:

zeevm:
@rtyler  I see all the "sidecar" code is already implemented, why would the library have a problem reading them?
The crate documentation says reader level 3 is supported, this includes V2 checkpoints, if the lib can't read V2 checkpoints why does it say in the doc it does?

rtyler:
>  I see all the "sidecar" code is already implemented, why would the library have a problem reading them?
The crate documentation says reader level 3 is supported, this includes V2 checkpoints, if the lib can't read V2 checkpoints why does it say in the doc it does?

There is an ability to parse a `sidecar` action in the transaction log, which doesn't mean much other than the transaction log wouldn't be unreadable to delta-rs if these actions are present in the table.  "v2 checkpoints" are a _table feature_  which require writer/reader version 3/7, which can definitely be confusing. Supporting reader v3, for example, does not mean that a reader supports all the table features which exist.

Thanks for the sample table, I'll take a looksee at it shortly.

zeevm:
@rtyler  any insights with the table I provided?

ion-elgreco:
Delta-kernel-rs supports v2 checkpoints reading. @roeap when do you expect your PR to be ready move over to the new log replay relying more on kernel, I assume v2 checkpoints will be auto supported by that change?

rtyler:
As @ion-elgreco alluded to there is work that @roeap has started to incorporate more log-replay support from delta-kernel-rs into delta-rs in #3137. He and I have chatted a bit about how it is still depending on some unmerged delta-kernel-rs pull requests such as this one https://github.com/delta-io/delta-kernel-rs/pull/549

Helping @roeap get these changed merged are at the top of my list at the moment for this PR and others like it :smile:

ryzhyk:
@ion-elgreco , @rtyler , @roeap , I understand that this feature is now part of the kernelization work. Do you have a timeline for when it will become available by any chance?

mansenfranzen:
I'm also very curious to know when V2 checkpoints will be supported. Any updates on this? Thanks a lot for your work!

ryzhyk:
@mansenfranzen , I talked to @roeap on slack. If I understand correctly, while delta-rs doesn't yet support v2 checkpoints, it should in theory still be able to open tables with v2 checkpoints enabled, since Spark periodically creates v1 checkpoints alongside v2. However, this doesn't work due to a bug, which @roeap intends to fix. I'm not sure what the ETA for the fix or v2 support is.

mansenfranzen:
Thanks for the update - much appreciated!

rtyler:
I am going to close this, I believe we have this support through the delta-kernel-rs integration. I don't have real v2 checkpoint tables to test with but there are a [number of test cases in kernel that look like they cover v2 checkpoints](https://github.com/delta-io/delta-kernel-rs/tree/main/kernel/tests/data)

Issue #3185, opened 2025-02-03
Zorder after schema evolution error - "Cannot cast file schema"


liamphmurphy:
# Environment

**Delta-rs version**: 0.24.0 (goes back at least as far as v0.16.4)

**Binding**: python

**Environment**: Local

***
# Bug

**What happened**:
When a partition has data written that is the result of a `schema_mode=merge` call with a schema change, a z_order optimization fails with this error:
```
_internal.DeltaError: Failed to parse parquet: Parquet error: Z-order failed while scanning data: Plan("Cannot cast file schema field details of type Struct([Field { name: \"email\", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]) to table schema field of type Struct([Field { name: \"email\", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: \"address\", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }])")
```

**NOTE**: MRE is below. This seems to only happen when a field is added WITHIN a Pyarrow schema struct.

**What you expected to happen**:
Z_order optimization would work as expected despite a schema evolution.

**How to reproduce it**:

Here is an MRE:

```
import pandas as pd
import pyarrow as pa
from deltalake import write_deltalake, DeltaTable

# Define the path for the Delta table
delta_table_path = "./test-table"

# Define the data for the first write
data_first_write = [
    {"name": "Alice", "age": 30, "details": {"email": "alice@example.com"}},
    {"name": "Bob", "age": 25, "details": {"email": "bob@example.com"}}
]

# Define the data for the second write with an additional 'address' field
data_second_write = [
    {"name": "Charlie", "age": 35, "details": {"email": "charlie@example.com", "address": "123 Main St"}},
    {"name": "Diana", "age": 28, "details": {"email": "diana@example.com", "address": "456 Elm St"}}
]

# Define Pyarrow schemas for the first and second writes
schema_first_write = pa.schema([
    ("name", pa.string()),
    ("age", pa.int32()),
    ("details", pa.struct([
        ("email", pa.string())
    ]))
])

schema_second_write = pa.schema([
    ("name", pa.string()),
    ("age", pa.int32()),
    ("details", pa.struct([
        ("email", pa.string()),
        ("address", pa.string())  # New field in the second schema
    ]))
])

# Convert data to DataFrames
[...]

PR #3238, opened 2025-02-21
fix: use tableprovider scan in z-order

ion-elgreco:
# Description
Our DeltaTableProvider is better to use than reading directly with ctx.read_parquet, since the DeltaTableProvider has a schemaAdapterFactory which maps each batch to the input_schema. This creates consistency while reading. A good side effect is we don't need to that path serialization fix anymore @rtyler 

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/3185

### DIFF
--- a/crates/core/src/kernel/scalars.rs
+++ b/crates/core/src/kernel/scalars.rs
@@ -72,8 +72,8 @@ impl ScalarExt for Scalar {
             },
             Self::Binary(val) => create_escaped_binary_string(val.as_slice()),
             Self::Null(_) => "null".to_string(),
-            Self::Struct(_) => unimplemented!(),
-            Self::Array(_) => unimplemented!(),
+            Self::Struct(_) => self.to_string(),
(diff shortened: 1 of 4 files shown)


ion-elgreco:
@roeap do you perhaps know how to convert the logical file stats back into ADD {stats}?

roeap:
Hmm ... maybe we can leverage arrow_json mechanics for this? https://arrow.apache.org/rust/arrow_json/trait.JsonSerializable.html

ion-elgreco:
> Hmm ... maybe we can leverage arrow_json mechanics for this? https://arrow.apache.org/rust/arrow_json/trait.JsonSerializable.html

Actually kernel already implements `fmt` for Struct(StructData) so it was a matter of updating our .seralize paths

roeap (approved):
LGTM!

More consistent use of Add and table provider will also make the adoption of kernel a bit smoother!!

Issue #3057, opened 2024-12-13
Snapshot references old logs still after `cleanup_expired_logs_for`


valentinmk:
# Environment

**Delta-rs version**: 0.22.3

**Binding**: Python

**Environment**:
- **Cloud provider**: custom S3 (have a prod problems)
- **OS**: reproduce on Mac OS with Python 3.12

***
# Bug

**What happened**: Function `.is_deltatable` return `False` for existing Delta table after some iteration of an ingestion and clean up via compact + vacuum + checkpoint 

**What you expected to happen**: Expects that a `write_delta` function with `mode="append"` or `mode="overwrite"` not broke delta table. A `checkpoint` function do not broke Delta tabel. A vacuum function cleanup all outdated files.

**How to reproduce it**: I've prepared demo script to show my finding. Please guide me if some thing doing wrong, I start to figure out (intuition) that some sort of a 'state' stored with in delta-rs and ie before vacuum it is good idea to reinitiate dt with `DeltaTable`

**More details**:

Initiate a python stuff. It is writing to a folder `lake_delta`.

```python
❯ ipython
Python 3.12.4 | packaged by conda-forge | (main, Jun 17 2024, 10:13:44) [Clang 16.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import polars as pl
   ...: from deltalake import DeltaTable
   ...: from datetime import datetime

In [2]: filename = "lake_delta"
   ...: configuration = {
   ...:     "delta.deletedFileRetentionDuration": "interval 0 days",
   ...:     "delta.logRetentionDuration": "interval 0 days",
   ...:     "delta.targetFileSize": str(128 * 1024 * 1024),
   ...: }

In [3]: for i in range(2):
   ...:     filename = "lake_delta"
   ...:     now = int(datetime.now().timestamp()*1000)
   ...:     data = [[now], [1], [2], [3], [4], [5]]
   ...:     df = pl.DataFrame(data, schema=["dt", "a", "b", "c", "d", "e"])
   ...:     df.write_delta(filename, delta_write_options={"configuration": configuration},mode="append")
   ...:
```
Define helper function to see some information about our table.

```python
In [4]: def print_debug():
   ...:     dt = DeltaTable(filename)
   ...:     print(
   ...:         f"{DeltaTable.is_deltatable(filename)}, Version: {dt.version()}, "
[...]

PR #3252, opened 2025-02-22
fix: refresh snapshot after vacuuming logs

ion-elgreco:
# Description
As per @roeap  suggestion to simply reload the table state after cleaning up logs so that we get a proper snapshot.

Introduces a write function on the DeltaTable which allows us to update the state after writing, which we couldn't do properly before.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/3057
- closes https://github.com/delta-io/delta-rs/issues/3062

### DIFF
--- a/crates/core/src/operations/transaction/mod.rs
+++ b/crates/core/src/operations/transaction/mod.rs
@@ -74,7 +74,6 @@
 //!       └───────────────────────────────┘
 //!</pre>
 use std::collections::HashMap;
-use std::future::Future;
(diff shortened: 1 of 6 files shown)


roeap (commented):
LGTM! agreeing with @rtyler thogh that we should see if we can cosolidate.

Reviewing our APIs is something to keep in ming for the 1.0 plan - no better time to break things :D.

roeap (approved):

Issue #1984, opened 2023-12-19
Use async writer + multipart + explore Datafusion sink


ion-elgreco:
# Description
The rust writer in it current states keeps a buffer instead of steaming to disk which causes the writer use quite some extra memory.

We need to address this performance issue. 

@wjones127 mentioned what needs to happen here: https://delta-users.slack.com/archives/C013LCAEB98/p1700330750311529?thread_ts=1700325888.484149&cid=C013LCAEB98

"
What we want to do instead is stream out to disk. Right now the writer is ArrowWriter<ShareableBuffer> : https://github.com/delta-io/delta-rs/blob/daa700eadaa2a6cc968d0b63cf4c5e7cfd65fc55/crates/deltalake-core/src/writer/record_batch.rs#L252
We should change it so it uses put_multipart and AsyncArrowWriter. That would make the type AsyncArrowWriter<Box<dyn AsyncWrite + Unpin + Send>>
What we want to do instead is combine
"

Another thing is to explore the Datafusion sink functionality as per suggestion of @roeap

wjones127:
Originally discussed in: https://github.com/delta-io/delta-rs/issues/1225

roeap:
btw. had a quick look into the datafusion sinks, and I believe they may not be the best fit for us, considering the work delta needs to do on write. More specifically I had look if it would make sense to implement a `FileFormat` for Delta... 

The `TableProvider` does however have more methods available now, that integrate into the framkework - they also did some really great work integrating with @wjones127's multi-part writer....

tustvold:
FYI I plan to add some first-party writer support to the parquet crate as part of https://github.com/apache/arrow-rs/issues/5524

ion-elgreco:
@tustvold awesome!

PR #3255, opened 2025-02-23
refactor: async writer + multi-part

ion-elgreco:
# Description
Changed to use arrows async writer and use multi-part uploads.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/1984


crates/core/src/operations/write/writer.rs
```diff
@@ -376,7 +402,33 @@ impl PartitionWriter {
         let file_size = buffer.len() as i64;
 
         // write file to object store
-        self.object_store.put(&path, buffer.into()).await?;
+        let mut multi_part_upload = self.object_store.put_multipart(&path).await?;
+        let part_size = upload_part_size();
+        let mut tasks = JoinSet::new();
+        let max_concurrent_tasks = 10; // TODO: make configurable
```
rtyler, line 408: This is really such a major behavior change. I am not terribly familiar with the maturity level of multipart uploads in `object_store`   I don't think this is necessarily a bad change, but I am doubtful of this addressing the originally linked issue.

As best as  I can tell the buffers are still going to fill up memory until the flush, and then the flush is going to fan out to have parallel uploads

```mermaid
flowchart LR
    write --> buffer_batch;
    write --> buffer_batch;
    write --> buffer_batch;
    buffer_batch --> flush;
    flush --> p1;
    flush --> p2;
    flush --> p3;
    flush --> p4;
    p1 --> close;
    p2 --> close;
    p3 --> close;
    p4 --> close;
```

ion-elgreco: Good point, in its current form we indeed still buffer until the flush, but the buffering and writing has more parallelism now, so it should be faster.

We could do two things here btw:

- release this as is for people to experiment with (maybe after 1.0)
- iterate on this to also flush after the min - part size is available, also create some benchmark tests

cjolowicz:
@ion-elgreco In reply to your comment on #3157 , yes I think this would likely speed up the upload phase of z-ordering.  Additionally, we've been watching this with interest because it would unblock compaction to target sizes > 5GB, which is the upper limit for single part uploads in R2.

ion-elgreco:
@cjolowicz maybe give it a spin, I don't have any environments or datasets to actually test the performance myself

rtyler (approved):
I remain skeptical we will see substantive performance improvements here, but willing to try in the next release :smiling_imp:

PR #3257, opened 2025-02-24
docs: update CONTRIBUTING.md and Makefile to account for switch to uv

adamreeve:
# Description

I wanted to start developing delta-rs and found CONTRIBUTING.md confusing as it asks you to create a Python virtualenv that is never used. This updates the documentation to account for the switch to uv.

This was partly addressed previously in #3085.


ion-elgreco (approved):
Thanks for catching this :)

PR #3260, opened 2025-02-24
fix: enable missing cloud features for uc crate

### LINKED ISSUE #3236: Datafusion error: External error: Failed to get a credential from UnityCatalog client configuration.
# Environment

**Delta-rs version**: 0.25

**Binding**: Python

**Environment**:
- **OS**: Windows 10

***
# Bug

trying to read a table from UC using azure databricks 

```
import os
from  deltalake import DeltaTable
os.environ["DATABRICKS_WORKSPACE_URL"] = endpoint
os.environ["DATABRICKS_ACCESS_TOKEN"]  = token
os.environ["UNITY_ALLOW_HTTP_URL"]     = "true"
dt = DeltaTable("uc://databricks_ne.aemo.xx")
```
got this error :  

DeltaError: Generic error: Datafusion error: External error: Failed to get a credential from UnityCatalog client configuration.

omkar-foss:
# Description
This PR enables the `azure` feature for UC crate to make sure the `deltalake_azure` registered handlers are called [here](https://github.com/delta-io/delta-rs/blob/main/crates/catalog-unity/src/models.rs#L426-L430). Without this feature being enabled, it throws a warning from [here](https://github.com/delta-io/delta-rs/blob/main/crates/catalog-unity/src/models.rs#L432-L436) resulting in a `MissingCredentials` error - "Failed to get a credential from UnityCatalog client configuration".

Update: Also enables `gcp` and `r2` for UC crate.

# Related Issue(s)

- closes #3236

# Documentation

N/A

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -137,7 +137,7 @@ jobs:
         run: make setup-dat
 
       - name: Run tests
-        run: uv run --no-sync pytest -m '(unitycatalog_databricks and integration)' --doctest-modules
+        run: ../../delta-rs/.github/scripts/retry_integration_test.sh unitycatalog_databricks 5 10
 
   test-unitycatalog-oss:
     name: Python Build (Python 3.10 Unity Catalog Integration tests)
@@ -166,7 +166,7 @@ jobs:
         run: make setup-dat
 
       - name: Run tests
-        run: uv run --no-sync pytest -m '(unitycatalog_oss and integration)' --doctest-modules
+        run: ../../delta-rs/.github/scripts/retry_integration_test.sh unitycatalog_oss 5 10
 
   test-pyspark:
     name: PySpark Integration Tests
@@ -191,7 +191,6 @@ jobs:
 
       - name: Run tests
         run: make test-pyspark
-
 
   multi-python-running:
     name: Running with Python ${{ matrix.python-version }}
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -43,11 +43,14 @@ tokio = { workspace = true, features = ["rt-multi-thread"] }
 
 deltalake-mount = { path = "../crates/mount" }
 
+# catalog-unity
+deltalake-catalog-unity = { path = "../crates/catalog-unity", features = ["aws", "azure", "gcp", "r2"] }
+
 # Non-unix or emscripten os
 [target.'cfg(any(not(target_family = "unix"), target_os = "emscripten"))'.dependencies]
 mimalloc = { version = "0.1", default-features = false }
 
-# Unix (excluding macOS & emscripten) → jemalloc 
+# Unix (excluding macOS & emscripten) → jemalloc
 [target.'cfg(all(target_family = "unix", not(target_os = "macos"), not(target_os = "emscripten")))'.dependencies]
 jemallocator = { version = "0.5", features = ["disable_initial_exec_tls", "background_threads"] }
 
--- a/.github/scripts/retry_integration_test.sh
+++ b/.github/scripts/retry_integration_test.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+set +e
+TEST_NAME=$1
+MAX_RETRIES=$2
+RETRY_DELAY=$3
+ATTEMPT=1
+run_command() {
+  uv run --no-sync pytest -m "($TEST_NAME and integration)" --doctest-modules 2>&1
+}
+until [ $ATTEMPT -gt $MAX_RETRIES ]
+do
+  echo "Attempt $ATTEMPT"
+  OUTPUT=$(run_command)
+  EXIT_CODE=$?
+  echo "$OUTPUT"
+  if [ $EXIT_CODE -eq 0 ]; then
+    echo "Tests passed."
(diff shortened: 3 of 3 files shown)


ion-elgreco:
@omkar-foss we should also enable R2 and GCS features : )

omkar-foss:
> @omkar-foss we should also enable R2 and GCS features : )

Thought to do that, but haven't tested those out :)

Should we enable it and let the delta-rs community users help with the testing?

ion-elgreco:
@omkar-foss let's just enable them without testing, it should work

ion-elgreco:
Actually, perhaps we should enable the features only in the python crate.

@hntd187 any reason why AWS is only the default feature and not the rest as well?

omkar-foss:
> @omkar-foss Actually, perhaps we should enable the features only in the python crate for now
> 
> @hntd187 any reason why AWS is only the default feature and not the rest as well?

Sure, wait I'll enable these in the python crate instead.

omkar-foss:
> Sure, wait I'll enable these in the python crate instead.

Done @ion-elgreco

ion-elgreco:
> > Sure, wait I'll enable these in the python crate instead.
> 
> Done @ion-elgreco 

That's actually in the published deltalake crate, I think it's should be done in python only just to not pre enable things for others

omkar-foss:
> > > Sure, wait I'll enable these in the python crate instead.
> > 
> > 
> > Done @ion-elgreco
> 
> That's actually in the published deltalake crate, I think it's should be done in python only just to not pre enable things for others

Oh right sorry, I've kept it in deltalake crate instead. Moving it just a minute.

omkar-foss:
@ion-elgreco regarding the flaky test, I'm adding a re-run option so it doesn't bother us when it fails.

python/Cargo.toml
```diff
@@ -43,11 +43,14 @@ tokio = { workspace = true, features = ["rt-multi-thread"] }
 
 deltalake-mount = { path = "../crates/mount" }
 
+# catalog-unity
+deltalake-catalog-unity = { version = "0.8.0", path = "../crates/catalog-unity", features = ["aws", "azure", "gcp", "r2"], optional = true }
```
ion-elgreco, line 47: I think the version needs to be left out here

omkar-foss: Done. Thanks!

omkar-foss:
> @ion-elgreco regarding the flaky test, I'm adding a re-run option so it doesn't bother us when it fails. Will add to this PR.

@ion-elgreco I've added this, can you approve the workflow so let's see if it's working fine in CI?

omkar-foss:
> > @ion-elgreco regarding the flaky test, I'm adding a re-run option so it doesn't bother us when it fails. Will add to this PR.
> 
> @ion-elgreco I've added this, can you approve the workflow so let's see if it's working fine in CI?

I need some time to get this retry part in place, so have removed it from this PR. Will raise a separate PR for this soon.

ion-elgreco (approved):
Cheers!

ion-elgreco (dismissed):

omkar-foss:
> Cheers!

Changes are now down to just adding the dependency. Less is more indeed! Cheers 😁

omkar-foss:
Added some wait time between the two consecutive UC tests, this should prevent the failure in the second one. I'll try and add retry soon but until then this should hopefully do the job.

.github/workflows/python_build.yml
```diff
@@ -137,7 +137,32 @@ jobs:
         run: make setup-dat
 
       - name: Run tests
-        run: uv run --no-sync pytest -m '(unitycatalog_databricks and integration)' --doctest-modules
+        run: |
```
rtyler, line 140: This should be either turned into a script or put into the Makefile. Having big scripts in GitHub Actions is quite a maintenance challenge

omkar-foss: Thanks @rtyler! Done, I've moved the inline parts to a script with an input parameter for integration test name. So if required someone can use it for other integration tests as well :)

omkar-foss:
Test failure looks like some intermittent failure: https://github.com/delta-io/delta-rs/actions/runs/13504095451/job/37735049349?pr=3260#step:5:30

GitHub has been facing some [service disruptions](https://www.githubstatus.com/incidents/135x076mn3tf) today, although those are resolved, this could be related.

omkar-foss:
Github has given up on my PR looks like 😅 😆 
```bash
Failed to download and build `pyspark==3.5.4`
  ├─▶ Failed to extract archive
  ├─▶ failed to unpack
  │   `/home/runner/work/_temp/setup-uv-cache/sdists-v8/.tmppzC2V5/pyspark-3.5.4/deps/jars/hadoop-shaded-guava-1.1.1.jar`
  ├─▶ failed to unpack
  │   `pyspark-3.5.4/deps/jars/hadoop-shaded-guava-1.1.1.jar` into
  │   `/home/runner/work/_temp/setup-uv-cache/sdists-v8/.tmppzC2V5/pyspark-3.5.4/deps/jars/hadoop-shaded-guava-1.1.1.jar`
```

ion-elgreco (approved):

PR #3261, opened 2025-02-24
chore: upgrade to DataFusion 46.0.0

alamb:
# Description
chore: upgrade to DataFusion 46.0.0
# Related Issue(s)
- related to https://github.com/apache/datafusion/issues/14123

PR #3262, opened 2025-02-24
chore: fix some compiler warnings

alamb:
# Description
chore: fix compiler warnings

While working on https://github.com/delta-io/delta-rs/pull/3261 I found several items that generated warnings locally for me:

For example:

```shell
/Users/andrewlamb/.cargo/bin/cargo test --color=always --workspace --profile test --no-fail-fast --config env.RUSTC_BOOTSTRAP=\"1\" -- --format=json -Z unstable-options --show-output
Testing started at 2:41 PM ...
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/andrewlamb/Software/delta-rs/python/Cargo.toml
workspace: /Users/andrewlamb/Software/delta-rs/Cargo.toml
warning: unused import: `std::future::Future`
  --> crates/core/src/operations/transaction/mod.rs:77:5
   |
77 | use std::future::Future;
   |     ^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

I looked at the CI tests and they seem to be passing, so I am not sure why this fails locally for me

# Related Issue(s)
- related to https://github.com/apache/datafusion/issues/14123

# Documentation

### DIFF
--- a/crates/aws/src/credentials.rs
+++ b/crates/aws/src/credentials.rs
@@ -41,6 +41,7 @@ impl AWSForObjectStore {
     }
 
     /// Return true if a credential has been cached
+    #[cfg(test)]
     async fn has_cached_credentials(&self) -> bool {
         let guard = self.cache.lock().await;
         (*guard).is_some()
--- a/crates/core/src/kernel/models/fields.rs
+++ b/crates/core/src/kernel/models/fields.rs
@@ -145,17 +145,6 @@ static REMOVE_FIELD: LazyLock<StructField> = LazyLock::new(|| {
         true,
     )
 });
-static REMOVE_FIELD_CHECKPOINT: LazyLock<StructField> = LazyLock::new(|| {
-    StructField::new(
-        "remove",
-        StructType::new(vec![
-            StructField::new("path", DataType::STRING, false),
-            StructField::new("deletionTimestamp", DataType::LONG, true),
-            StructField::new("dataChange", DataType::BOOLEAN, false),
-        ]),
-        true,
-    )
-});
 // https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-cdc-file
 static CDC_FIELD: LazyLock<StructField> = LazyLock::new(|| {
     StructField::new(
--- a/crates/core/src/kernel/models/fields.rs
+++ b/crates/core/src/kernel/models/fields.rs
@@ -145,6 +145,19 @@ static REMOVE_FIELD: LazyLock<StructField> = LazyLock::new(|| {
         true,
     )
 });
+// TODO implement support for this checkpoint
+#[expect(dead_code)]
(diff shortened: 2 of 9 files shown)


crates/aws/src/credentials.rs
```diff
@@ -41,6 +41,7 @@ impl AWSForObjectStore {
     }
 
     /// Return true if a credential has been cached
+    #[cfg(test)]
     async fn has_cached_credentials(&self) -> bool {
```
alamb, line 45: this is only used in tests it seems

crates/lakefs/src/storage.rs
```diff
@@ -121,14 +121,11 @@ mod tests {
     use maplit::hashmap;
     use serial_test::serial;
 
-    struct ScopedEnv {
-        vars: HashMap<std::ffi::OsString, std::ffi::OsString>,
-    }
+    struct ScopedEnv {}
```
alamb, line 124: the compiler claimed this field was never read -- if the CI passes we could remove the entire structure I think (but I wanted to make the diff smaller to see initially)

alamb:
I will make another follow on PR for clippy errors I am seeing locally if this one passes

crates/core/src/kernel/models/fields.rs
```diff
@@ -145,17 +145,6 @@ static REMOVE_FIELD: LazyLock<StructField> = LazyLock::new(|| {
         true,
     )
 });
-static REMOVE_FIELD_CHECKPOINT: LazyLock<StructField> = LazyLock::new(|| {
```
ion-elgreco, line 148: I think we want to keep this around for now

crates/core/src/operations/filesystem_check.rs
```diff
@@ -79,15 +79,6 @@ where
     serializer.serialize_str(&json_string)
 }
 
-// Custom deserialization that parses a JSON string into MetricDetails
-fn deserialize_vec_string<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
```
ion-elgreco, line 83: Here as well

alamb: The compiler says it is unused (and the function isn't pub). I will put it back with an `#allow(unused)`

Any chance you can give me the rationale to add to the comments for future readers?

crates/core/src/kernel/models/fields.rs
ion-elgreco: I think this is part of the checkpoint we still need to support at some point

alamb:
Fixed

ion-elgreco (approved):

alamb:
BTW the reason I would like to fix these warnings is so that it is easier to find the deprecation warnings during the DataFusion upgrade

ion-elgreco:
@alamb understood! :)

alamb:
🎉  thank you @ion-elgreco  ❤️

PR #3264, opened 2025-02-25
chore: fix clippy warnings on main

alamb:
# Description
Similarly to https://github.com/delta-io/delta-rs/pull/3262 I would like to use compiler warnings and clippy locally to find places that need to be changed after DataFusion upgrade. This means I would to have main have a clean clippy run

I just did what clippy told me to mostly

# Related Issue(s)
- Related to https://github.com/delta-io/delta-rs/pull/3261
- Related to https://github.com/apache/datafusion/issues/14123

# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -327,6 +327,7 @@ impl RawDeltaTable {
     /// This will acquire the internal lock since it is a mutating operation!
     pub fn load_version(&self, py: Python, version: i64) -> PyResult<()> {
         py.allow_threads(|| {
+            #[allow(clippy::await_holding_lock)]
             rt().block_on(async {
                 let mut table = self
                     ._table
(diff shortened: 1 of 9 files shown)


python/src/lib.rs
```diff
@@ -327,6 +327,7 @@ impl RawDeltaTable {
     /// This will acquire the internal lock since it is a mutating operation!
     pub fn load_version(&self, py: Python, version: i64) -> PyResult<()> {
         py.allow_threads(|| {
+            #[allow(clippy::await_holding_lock)]
```
alamb, line 330: Clippy is complaining about this -- it might be a real issue, but I didn't want to mess with it is as part of this PR. This PR makes the situation better in my opinion, as now the clippy warning is explicit in the code

roeap (approved):
Thanks for taking care of this @alamb!

PR #3265, opened 2025-02-25
chore: update tests to use `Column::new` and other expr_fn functions

alamb:
# Description
**Note this PR should have no functional change**

This is part of the DataFusion upgrade in https://github.com/delta-io/delta-rs/pull/3262, which I am trying to split into multiple parts to reduce the size of the diff (so to make it easier to review and debug)

DataFusion 46.0.0 adds a new field, `Column::spans` which requires changing all locations that create `Column` either to have a `span` in it or to use the constructor. So let's change the code to use the constructor here

# Related Issue(s)

- Related to https://github.com/delta-io/delta-rs/pull/3261
- Related to https://github.com/apache/datafusion/issues/14123

# Documentation

### DIFF
--- a/crates/core/src/operations/merge/filter.rs
+++ b/crates/core/src/operations/merge/filter.rs
@@ -453,14 +453,7 @@ mod tests {
         .build()
         .unwrap();
 
-        let join_predicate = col(Column {
-            relation: Some(source_name.clone()),
-            name: "id".to_owned(),
-        })
-        .eq(col(Column {
-            relation: Some(target_name.clone()),
-            name: "id".to_owned(),
-        }));
+        let join_predicate = make_join_predicate(&source_name, &target_name);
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/merge/filter.rs
```diff
@@ -453,14 +453,7 @@ mod tests {
         .build()
         .unwrap();
 
-        let join_predicate = col(Column {
```
alamb, line 456: the key rationale for this change is that `Column` has a new field in DataFUsion 46, so if it is constructed explicitly like this I need to change all the callsites

If I switch to using `Column::new()` then nothing needs to be changed as part of the upgrade (as Column::new handles initializing the newly added field)

ion-elgreco (approved):

PR #3273, opened 2025-02-26
chore: pin the chrono crate to avoid compilation issues with arrow 54.2.0

rtyler:
See arrow/arrow-rs#7202


rtyler:
Didn't see that #3272 was here first, closing this so I can merge that one.

PR #3275, opened 2025-02-27
fix(rust): on write, have a schema evolution maintain metadata table id

### LINKED ISSUE #3274: Schema evolution causing table ID to be regenerated, breaks Spark streaming jobs
# Environment

**Delta-rs version**: 0.25.2

**Binding**: Python

**Environment**:
AWS, Local

***
# Bug

**What happened**:

When a schema evolution write is triggered, we generate a whole new metadata line to represent the state of the new schema, which makes sense. But we're also regenerating the table's ID, which has negative consequences on Spark streaming jobs that use the metadata ID to tell which table is being read. When this changes, the Spark streaming read thinks that this table is a whole new one.

This net new metadata struct and ID is generated here:
https://github.com/delta-io/delta-rs/blob/94a2009c2b31d4509f9d425551bfde8731516fbe/crates/core/src/writer/record_batch.rs#L286

A new metadata struct is generated, but the metadata ID from the table's existing state is utilized instead.

**How to reproduce it**:

Perform any schema evolution write on a table and compare the original `metadata.id` to the new one.

**More details**:

I'm tinkering locally and think I have a fix made, but am also not a Rust expert, so if anyone has strong opinions on how to implement feel free to override me.

liamphmurphy:
# Description
When a schema evolution write occurs, if a metadata ID from the table state exists, use that in the new metadata transaction entry. 

This PR has a unit test and a new integration test that runs a Pyspark stream between evolution runs. I confirmed that the integration test failed with the error I've seen in production when I removed the new logic, and passes with the new logic.

# Related Issue(s)
closes #3274 

# Documentation

### DIFF
--- a/crates/core/src/kernel/models/actions.rs
+++ b/crates/core/src/kernel/models/actions.rs
@@ -111,6 +111,12 @@ impl Metadata {
         self
     }
 
+    /// set the table ID in the metadata action
+    pub fn with_table_id(mut self, id: String) -> Self {
+        self.id = id;
+        self
+    }
+
     /// get the table schema
     pub fn schema(&self) -> DeltaResult<StructType> {
         Ok(serde_json::from_str(&self.schema_string)?)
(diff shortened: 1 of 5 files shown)


rtyler (dismissed):

ion-elgreco:
@liamphmurphy can you squash the commits and fix the CI?

rtyler (dismissed):

Issue #3278, opened 2025-02-28
Partition pruning does not work when values in predicate do not match column datatype


ldacey:
# Environment

**Delta-rs version**:
0.25.2

**Binding**: 

**Environment**:
- **Cloud provider**: Google/Azure
- **OS**: Mac
- **Other**:

***
# Bug

**What happened**:
When I upgraded to 0.25 a lot of my delta tables began to fail due to pod OOM errors despite providing more and more memory to them. Disabling `streamed_exec=True` fixed this for me.

It seems like partition pruning is not happening - tables with hundreds or thousands are partitions are being fully scanned even when the merge predicates are specific and point directly to a single partition. For example, one of my tables is just a single date of data (76k rows) which is merged into a table partitioned by month_id and date_id. This should (and used to be) a quick merge since the predicate we use is explicit. This fails with `streamed_exec=True` though and debugging shows all 2600+ files are scanned.

**What you expected to happen**:
Predicates which filter out partitions should be respected.

**How to reproduce it**:
```python
import os
import shutil
import pyarrow as pa
from deltalake import DeltaTable

table_path1 = "test_delta_not_streamed"
table_path2 = "test_delta_streamed"

for path in [table_path1, table_path2]:
    if os.path.exists(path):
        shutil.rmtree(path)
    os.makedirs(path)

schema = pa.schema([
    pa.field("month_id", pa.int32()),
    pa.field("date_id", pa.int32()),
    pa.field("unique_row_hash", pa.string()),
    pa.field("col1", pa.string())
])

dt1 = DeltaTable.create(
    table_uri=table_path1,
    schema=schema,
    partition_by=["month_id", "date_id"]
)

dt2 = DeltaTable.create(
    table_uri=table_path2,
    schema=schema,
    partition_by=["month_id", "date_id"]
)

partitions = [
    (202501, 20250101),
    (202502, 20250201),
    (202502, 20250226),
    (202503, 20250301),
]

rows = []
for month_id, date_id in partitions:
    rows.append([month_id, date_id, f"hash_{month_id}_{date_id}", "value"])

source_table = pa.Table.from_arrays(
    [
        pa.array([row[0] for row in rows], type=pa.int32()),
        pa.array([row[1] for row in rows], type=pa.int32()),
        pa.array([row[2] for row in rows], type=pa.string()),
        pa.array([row[3] for row in rows], type=pa.string()),
    ],
    names=["month_id", "date_id", "unique_row_hash", "col1"]
)

[...]

PR #3281, opened 2025-03-01
refactor: simplify expressions

ion-elgreco:
# Description
During merge we should simplify the expression before creating the commit predicate. Additionally, inside the DeltaScan we also need to simplify the expression to improve the pruning.

- related https://github.com/delta-io/delta-rs/issues/3278

This doesn't fully fix it but discussing the issue here: https://github.com/apache/datafusion/issues/14944

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -41,6 +41,7 @@ use datafusion::error::Result as DataFusionResult;
 use datafusion::execution::context::SessionConfig;
 use datafusion::execution::session_state::SessionStateBuilder;
 use datafusion::logical_expr::build_join_schema;
+use datafusion::optimizer::simplify_expressions::ExprSimplifier;
 use datafusion::physical_plan::metrics::MetricBuilder;
 use datafusion::physical_planner::{ExtensionPlanner, PhysicalPlanner};
 use datafusion::{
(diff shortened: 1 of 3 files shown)


crates/core/src/operations/merge/mod.rs
```diff
@@ -60,6 +63,7 @@ use filter::try_construct_early_filter;
 use futures::future::BoxFuture;
 use parquet::file::properties::WriterProperties;
 use serde::Serialize;
+use tracing::field::debug;
 use tracing::log::*;
```
rtyler, line 67: `debug!` comes in on the second line, I'm guessing this was an IDE trying to be clever. I'll remove this in a later change so as not to block this.

crates/core/src/operations/merge/mod.rs
```diff
@@ -844,11 +848,32 @@ async fn execute(
             streaming,
         )
         .await?
+    }
+    .map(|e| {
+        // simplify the expression so we have
```
rtyler, line 853: So we have what? Quite a cliff-hanger of a comment! :laughing:   I hope you'll be able to expand on the commentary here in a later change

rtyler (approved):

crates/core/src/operations/merge/mod.rs
ion-elgreco: 😆 I guess I finished the comment in my mind only haha

I'll push a follow-up later

Issue #3271, opened 2025-02-26
Runtime panic in new streaming writer: Cannot start a runtime from within a runtime


Zan-L:
# Environment

**Delta-rs version**: 0.25.2

**Binding**: Python

**Environment**:
- **Cloud provider**: S3
- **OS**: Linux
- **Other**:

***
# Bug

**What happened**:
```python
write_deltalake(data_path, change_dt.to_pyarrow_dataset(), mode='overwrite', schema_mode='overwrite')
```

This call writes a table using PyArrow dataset from another table. Works >90% of the time in the new streaming writer but could triggers the following error for the rest:

```python
_internal.DeltaError: Generic DeltaTable error: External error: Arrow error: C Data interface error: Unknown error: Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.. Detail: Python exception: Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/deltalake/writer.py", line 322, in <genexpr>
    data = RecordBatchReader.from_batches(schema, (batch for batch in data))
                                                                      ^^^^
  File "pyarrow/ipc.pxi", line 671, in pyarrow.lib.RecordBatchReader.__next__
  File "pyarrow/ipc.pxi", line 705, in pyarrow.lib.RecordBatchReader.read_next_batch
  File "pyarrow/error.pxi", line 89, in pyarrow.lib.check_status
  File "/usr/local/lib/python3.12/site-packages/deltalake/schema.py", line 148, in _cast_schema_to_recordbatchreader
    for batch in reader:
                 ^^^^^^
  File "pyarrow/ipc.pxi", line 671, in pyarrow.lib.RecordBatchReader.__next__
  File "pyarrow/ipc.pxi", line 705, in pyarrow.lib.RecordBatchReader.read_next_batch
  File "pyarrow/error.pxi", line 89, in pyarrow.lib.check_status
  File "pyarrow/_fs.pyx", line 1559, in pyarrow._fs._cb_open_input_file
  File "/usr/local/lib/python3.12/site-packages/deltalake/fs.py", line 110, in open_input_file
    return pa.PythonFile(self._handler.open_input_file(path))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3283, opened 2025-03-01
fix: separate runtime for python IO

ion-elgreco:
# Description
Probably solves the panic with long running tasks

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/3271#top


rtyler (approved):

ion-elgreco:
@rtyler i actually need to use the with_io_runtime constructs here, I have this in some commits laying around, will address that once Im back

PR #3289, opened 2025-03-02
chore: upgrade the kernel version and bump our majorish versions too

rtyler:
There are some subsequent failures and changes which need to be made here to go green, but those will come in follow up commits


rtyler:
There was a change in kernel for the `PartialEq` of `Scalar` that I'm looking at in relation to some of these test failures :thinking: :male_detective:

roeap (approved):
LGTM!

PR #3318, opened 2025-03-11
fix: check for all known valid delta files in is_deltatable

### LINKED ISSUE #3314: DeltaTable.is_deltatable sometimes return false for valid deltatable
# Environment

**Delta-rs version**:

0.24.0

**Binding**: 

Python

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

**What happened**:

DeltaTable.is_deltatable sometimes returns false for a valid deltatable. Other DeltaTable operations like showing metadata and partitions work fine.

The check in is_deltatable doesn't account for other types of files than crc, commit and checkpoint files. Other files in the _deltalog directory like "_last_checkpoint", "_last_vacuum_info" and deletion vector files, generated by Spark, can result in is_deltatable returning false if they are the first files listed.

https://github.com/delta-io/delta-rs/blob/rust-v0.24.0/crates/core/src/logstore/mod.rs#L274

**How to reproduce it**:

**More details**:

umartin:
# Description
This PR adds some missing delta file types to the check in is_deltatable().
# Related Issue(s)

- closes #3314

# Documentation

### DIFF
--- a/crates/core/src/kernel/snapshot/log_segment.rs
+++ b/crates/core/src/kernel/snapshot/log_segment.rs
@@ -28,6 +28,12 @@ static CHECKPOINT_FILE_PATTERN: LazyLock<Regex> =
 static DELTA_FILE_PATTERN: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^\d+\.json$").unwrap());
 static CRC_FILE_PATTERN: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^(\.\d+(\.crc|\.json)|\d+)\.crc$").unwrap());
+static LAST_CHECKPOINT_FILE_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^_last_checkpoint$").unwrap());
+static LAST_VACUUM_INFO_FILE_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^_last_vacuum_info$").unwrap());
+static DELETION_VECTOR_FILE_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r".*\.bin$").unwrap());
 pub(super) static TOMBSTONE_SCHEMA: LazyLock<StructType> =
     LazyLock::new(|| StructType::new(vec![ActionType::Remove.schema_field().clone()]));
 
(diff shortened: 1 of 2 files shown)


umartin:
The current implementation makes the assumption that there are allowed file patterns and that all other patterns are disallowed. I don't know if that assumption is valid. I find no reference to the spark generated "_last_vacuum_info" in the protocol. Future versions might add new file types. Are delta engines allowed to add other files like indexes? That sounds like a big risk compared to the poor performance of calling is_deltatable on a non delta table directory with many files. And only files in the _delta_log directory are checked.

Another option would be something like
```python
try:
  DeltaTable(path).version()
  return True
except:
  return False
```
I think that is what the spark implementation does but instead of exceptions they use version < 0 to signal that there is no log to read.

I'm happy to change to code either way you think is best.

ion-elgreco:
> The current implementation makes the assumption that there are allowed file patterns and that all other patterns are disallowed. I don't know if that assumption is valid. I find no reference to the spark generated "_last_vacuum_info" in the protocol. Future versions might add new file types. Are delta engines allowed to add other files like indexes? That sounds like a big risk compared to the poor performance of calling is_deltatable on a non delta table directory with many files. And only files in the _delta_log directory are checked.
> 

> I think that is what the spark implementation does but instead of exceptions they use version < 0 to signal that there is no log to read.
> 
> I'm happy to change to code either way you think is best.

Unfortunately the folks at delta-spark are doing things that are not always following protocol which makes it hard to get proper interoperability.

> Another option would be something like
> 
> ```python
> try:
>   DeltaTable(path).version()
>   return True
> except:
>   return False
> ```
> 
This implies loading the whole table state, which is also not very optimal.

umartin:
@ion-elgreco What do you say? Should I keep the existing semantics and add the missing file types or is this an acceptable solution. I agree with you that reading the whole table state is to expesive.

ion-elgreco:
> @ion-elgreco What do you say? Should I keep the existing semantics and add the missing file types or is this an acceptable solution. I agree with you that reading the whole table state is to expensive.

I prefer we keep the existing semantics but with extended file types. I suggest we add an extra line to the error message, to say why it was an invalid table because we saw file "*.<extension" and valid delta files are *.json , *.parquet etc

ion-elgreco:
> Can we add an error message without breaking the api? It would require changing Ok(false) to Err(msg) when invalid files are found. The python api would return false if the directory doesn't exists or it is empty. For non delta directories it would throw an exception instead of returning false.

Maybe a warn trace would be better then, so we return False still but throw a warning in the logs why it was False

ion-elgreco (approved):
Thanks for the fix :)

Issue #3243, opened 2025-02-22
Onelake incompatible with rustls


djouallah:
# Environment

**Delta-rs version**: 0.25.1

**Binding**: Python

**Environment**: Linux 
- **Cloud provider**: Microsoft Fabric onelake 
- **OS**:
- **Other**:

***
running
```
pip install deltalake --upgrade
dt = DeltaTable(result_table)

```
get 

`OSError: Generic MicrosoftAzure error: Error after 10 retries in 2.475340161s, max_retries:10, retry_timeout:180s, source:error sending request for url (https://onelake.blob.fabric.microsoft.com/sqlengines/ETL.Lakehouse/Tables/dbo/results/_delta_log/_last_checkpoint)`

cc @RobinLin666

ion-elgreco:
@djouallah have you tried another os to read with?

ion-elgreco:
@djouallah I suggest you try to compile of each of these branches and report back if it resolves the connection issue:

- https://github.com/ion-elgreco/delta-rs/tree/fix/rustls
- https://github.com/ion-elgreco/delta-rs/tree/fix/use-native-tls-vendored

djouallah:
windows works fine with onelake

ion-elgreco:
Ok then it's related to the SSL handshakes. Please compile both of these branches and let me know the results of each, then I can narrow down the fix

ion-elgreco:
@djouallah can you check the latest release out please

djouallah:
same errors 

`OSError: Generic MicrosoftAzure error: Error after 10 retries in 3.247428205s, max_retries:10, retry_timeout:180s, source:error sending request for url (https://onelake.blob.fabric.microsoft.com/sqlengines/ETL.Lakehouse/Tables/T10/streaming/_delta_log/_last_checkpoint)`

ion-elgreco:
@djouallah then it's a seperate issue for you, can you try these wheels:

You can download the artifacts from here:

Manylinux_2_17: use release mode · ion-elgreco/delta-rs@274becd (as it is from main) (this should fail I HOPE)
Manylinux_2_28: release mode · ion-elgreco/delta-rs@fcd9d6c (this maybe fixes it?)

Manylinux_2_17 with more rustls features: https://github.com/ion-elgreco/delta-rs/actions/runs/13506844973 (main + more features enabled essentially) (this maybe fixes it?)
[...]

PR #3335, opened 2025-03-17
feat(api): add rustls and native-tls features

### LINKED ISSUE #3334: Allow choosing kernel built-in engine
# Description
Currently deltalake uses the rustls kernel engine, some users prefer the tls engine, need to add a feature to make this user selectable.

zeevm:
# Description
This adds two features to deltalake: "kernel-rustls" and "kernel-tls".
These features are propagated to "core" to choose the kernel engine variant, default is "kernel-rustls" and users can disable default and enable "kernel-tls" instead.

# Related Issue(s)
- closes #3334


zeevm:
@rtyler I see you've self-assigned this, any thoughts on the change?

crates/core/Cargo.toml
```diff
@@ -113,7 +113,7 @@ tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 
 [features]
 cdf = []
-default = ["cdf"]
+default = ["cdf", "rustls"]
```
zeevm, line 116: @rtyler the idea is to not force rustls on users who don't need it for the kernel engine, why make it default here?

zeevm:
@ion-elgreco Any idea why integration tests failed on no disk space? Can you trigger this check again?

crates/core/Cargo.toml
ion-elgreco: Having rustls as the default is not an issue here, if you compile it you can disable default features and only select the ones you needed.

Runner is simply has no disk space because of all the compilations

zeevm: @ion-elgreco @rtyler so how do we push this change forward? Ideally there shouldn't be a dependency on 'rustls' that can't be opted-out, like my original commit.

WDYT?

zeevm: @rtyler any thoughts about my comment? Can we remove "rustls" from "core" defaults?

djouallah:
does this fix this issue : https://github.com/delta-io/delta-rs/issues/3243

zeevm:
> does this fix this issue : #3243

Wasn't designed to fix this, but if the issue is rustls and tls can be used instead with lafefs then this PR might also fix this.

However, at the moment the PR seems to be stuck, I'm trying to push it forward with @rtyler and @ion-elgreco with little success thus far.

ion-elgreco:
> > does this fix this issue : #3243
> 
> Wasn't designed to fix this, but if the issue is rustls and tls can be used instead with lafefs then this PR might also fix this.
> 
> However, at the moment the PR seems to be stuck, I'm trying to push it forward with @rtyler and @ion-elgreco with little success thus far.

Tyler assigned himself so would just have to be patient here

> does this fix this issue : https://github.com/delta-io/delta-rs/issues/3243

It won't because the python binding will keep using rustls. You can compile your own wheels with openssl vendored instead once this gets merged

crates/core/Cargo.toml
rtyler: @zeevm deltalake-core cannot be compiled without at least one delta kernel engine added in. Users that do not want to use rustls should be adding the dependency with

```toml
deltalake-core = { version "0.26", no-default-features, features = ["cdf", "etc"] }
```

The space exhaustion in GitHub CI is annoying and I'll have to juggle some splits of jobs to avoid it :weary:

ion-elgreco (approved):

PR #3338, opened 2025-03-20
perf: use lazy sync reader

ion-elgreco:
# Description
Returns a lazy recordbatchreader

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -3,6 +3,7 @@ mod features;
 mod filesystem;
 mod merge;
 mod query;
+mod reader;
 mod schema;
 mod utils;
 mod writer;
@@ -23,7 +24,7 @@ use delta_kernel::expressions::Scalar;
 use delta_kernel::schema::{MetadataValue, StructField};
 use deltalake::arrow::compute::concat_batches;
 use deltalake::arrow::ffi_stream::{ArrowArrayStreamReader, FFI_ArrowArrayStream};
-use deltalake::arrow::record_batch::{RecordBatch, RecordBatchIterator};
+use deltalake::arrow::record_batch::RecordBatch;
 use deltalake::arrow::{self, datatypes::Schema as ArrowSchema};
 use deltalake::checkpoints::{cleanup_metadata, create_checkpoint};
 use deltalake::datafusion::catalog::TableProvider;
@@ -52,7 +53,7 @@ use deltalake::operations::transaction::{
 use deltalake::operations::update::UpdateBuilder;
 use deltalake::operations::vacuum::VacuumBuilder;
 use deltalake::operations::write::WriteBuilder;
-use deltalake::operations::{collect_sendable_stream, CustomExecuteHandler};
+use deltalake::operations::CustomExecuteHandler;
 use deltalake::parquet::basic::Compression;
 use deltalake::parquet::errors::ParquetError;
 use deltalake::parquet::file::properties::{EnabledStatistics, WriterProperties};
@@ -63,7 +64,7 @@ use deltalake::table::state::DeltaTableState;
 use deltalake::{init_client_version, DeltaTableBuilder};
 use deltalake::{DeltaOps, DeltaResult};
 use error::DeltaError;
-use futures::future::join_all;
+use reader::convert_stream_to_reader;
 use tracing::log::*;
 
 use crate::writer::to_lazy_table;
@@ -899,31 +900,12 @@ impl RawDeltaTable {
             format!("SELECT {select_string} FROM `{table_name}`")
         };
 
-        let plan = rt()
-            .block_on(async { ctx.sql(&sql).await?.create_physical_plan().await })
+        let stream = rt()
+            .block_on(async { ctx.sql(&sql).await?.execute_stream().await })
             .map_err(PythonError::from)?;
 
         py.allow_threads(|| {
-            let mut tasks = vec![];
-            for p in 0..plan.properties().output_partitioning().partition_count() {
-                let inner_plan = plan.clone();
-                let partition_batch = inner_plan.execute(p, ctx.task_ctx()).unwrap();
-                let handle = rt().spawn(collect_sendable_stream(partition_batch));
-                tasks.push(handle);
-            }
-
-            // This is unfortunate.
-            let batches = rt()
-                .block_on(join_all(tasks))
-                .into_iter()
-                .flatten()
-                .collect::<Result<Vec<Vec<_>>, _>>()
-                .unwrap()
-                .into_iter()
-                .flatten()
-                .map(Ok);
-            let batch_iter = RecordBatchIterator::new(batches, plan.schema());
-            let ffi_stream = FFI_ArrowArrayStream::new(Box::new(batch_iter));
+            let ffi_stream = FFI_ArrowArrayStream::new(convert_stream_to_reader(stream));
             let reader = ArrowArrayStreamReader::try_new(ffi_stream).unwrap();
             Ok(PyArrowType(reader))
         })
--- a/python/src/reader.rs
+++ b/python/src/reader.rs
@@ -1,0 +1,39 @@
+use arrow_schema::{ArrowError, SchemaRef};
+use deltalake::arrow::array::RecordBatchReader;
+use deltalake::arrow::record_batch::RecordBatch;
+use deltalake::datafusion::execution::RecordBatchStream;
+use futures::StreamExt;
+use std::pin::Pin;
+
+use crate::utils::rt;
+
+/// A lazy adapter to convert an async RecordBatchStream into a sync RecordBatchReader
+struct StreamToReaderAdapter {
+    schema: SchemaRef,
+    stream: Pin<Box<dyn RecordBatchStream + Send>>,
+}
+
+impl Iterator for StreamToReaderAdapter {
+    type Item = Result<RecordBatch, ArrowError>;
+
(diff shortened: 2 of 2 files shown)


roeap (commented):
Not entirely sure, but I think the mode idiomatic way of doing this might be to use channels rather then blocking on each poll?

In kernel we have to do this quite heavily to bridge between streams and iterators. For instance in the [`list_from`](https://github.com/delta-io/delta-kernel-rs/blob/ec47584f183441f031287f26424fd7dd1ab02baa/kernel/src/engine/default/filesystem.rs#L67-L99) function.

I could not tell you where or if that still exists, but I have a vague memory of datafusion having a generic helper struct somewhere to achieve this as well.

ion-elgreco:
> Not entirely sure, but I think the mode idiomatic way of doing this might be to use channels rather then blocking on each poll?
> 
> In kernel we have to do this quite heavily to bridge between streams and iterators. For instance in the [`list_from`](https://github.com/delta-io/delta-kernel-rs/blob/ec47584f183441f031287f26424fd7dd1ab02baa/kernel/src/engine/default/filesystem.rs#L67-L99) function.
> 
> I could not tell you where or if that still exists, but I have a vague memory of datafusion having a generic helper struct somewhere to achieve this as well.

I don't see how that would work? The RecordBatchReader trait is a sync iterator

roeap:
> I don't see how that would work? The RecordBatchReader trait is a sync iterator

In the example I linked, the `receiver.into_iter()` returns an iterator, while the data is produced on a stream. I think this matches this scenario?

ion-elgreco:
@roeap I still having hard time seeing the benefit, take this working example. Recv() also blocks the current thread until it gets the values from the channel


```rust
use arrow_schema::{ArrowError, SchemaRef};
use deltalake::arrow::array::RecordBatchReader;
use deltalake::arrow::record_batch::RecordBatch;
use deltalake::datafusion::execution::RecordBatchStream;
use futures::StreamExt;
use std::pin::Pin;
use std::sync::mpsc::{sync_channel, Receiver};

use crate::utils::rt;

/// A lazy adapter to convert an async RecordBatchStream into a sync RecordBatchReader
struct StreamToReaderAdapter {
    schema: SchemaRef,
    receiver: Receiver<Result<RecordBatch, ArrowError>>,
}

impl Iterator for StreamToReaderAdapter {
    type Item = Result<RecordBatch, ArrowError>;

    fn next(&mut self) -> Option<Self::Item> {
        self.receiver.recv().ok()
    }
}

impl RecordBatchReader for StreamToReaderAdapter {
    fn schema(&self) -> SchemaRef {
        self.schema.clone()
    }
}

/// Converts a RecordBatchStream into a lazy RecordBatchReader
pub(crate) fn convert_stream_to_reader(
    stream: Pin<Box<dyn RecordBatchStream + Send>>,
) -> Box<dyn RecordBatchReader + Send> {
    let (sender, receiver) = sync_channel(0);
    let schema = stream.schema();

    rt().spawn_blocking(move || {
        rt().block_on(async move {
            let mut stream = stream;
            while let Some(batch) = stream.next().await {
                if sender
                    .send(batch.map_err(|e| ArrowError::ExternalError(Box::new(e))))
                    .is_err()
                {
                    break;
                }
            }
        });
    });

    Box::new(StreamToReaderAdapter { schema, receiver })
}

```

roeap (dismissed):
maybe add a test, otherwise LGTM.

ion-elgreco:
@roeap I didn't check in that code above, could you give some comments on that?

roeap:
@ion-elgreco - i think something like this could also work. (at least the compiler seems happy, didn't actually run this though :))

```rs
fn stream_to_iterator(
    mut stream: Pin<Box<dyn RecordBatchStream + Send>>,
) -> impl Iterator<Item = RecordBatch> {
    let (sender, receiver) = std::sync::mpsc::sync_channel(4_000);

    rt().spawn(async move {
        while let Some(batch) = stream.next().await {
            if sender.send(batch.unwrap()).is_err() {
                break;
            }
        }
    });

    receiver.into_iter()
}
```

While I am also not an expert in this stuff (so take the following with a grain of salt :)), there might however an argument to me made for the polling approach. The channel has a capacity indicating that we can create back-pressure (i.e. if we cannot consume the iterator fast enough, the channel might be at capacity). With the current implementation we don't run that risk. 

In practice I suspect there will not be any noticeable performance difference unless users run under very high load..

ion-elgreco:
> @ion-elgreco - i think something like this could also work. (at least the compiler seems happy, didn't actually run this though :))
> 
> ```rust
> fn stream_to_iterator(
>     mut stream: Pin<Box<dyn RecordBatchStream + Send>>,
> ) -> impl Iterator<Item = RecordBatch> {
>     let (sender, receiver) = std::sync::mpsc::sync_channel(4_000);
> 
>     rt().spawn(async move {
>         while let Some(batch) = stream.next().await {
>             if sender.send(batch.unwrap()).is_err() {
>                 break;
>             }
>         }
>     });
> 
>     receiver.into_iter()
> }
> ```
> 
> While I am also not an expert in this stuff (so take the following with a grain of salt :)), there might however an argument to me made for the polling approach. The channel has a capacity indicating that we can create back-pressure (i.e. if we cannot consume the iterator fast enough, the channel might be at capacity). With the current implementation we don't run that risk.
> 
> In practice I suspect there will not be any noticeable performance difference unless users run under very high load..

That also works, but I think we should use spawn_blocking instead since it can be cpu bound to generate the batches, also I want to have the channel not use a buffer. 4000 batches could be extremely high in terms of memory, and we have no way to tell in advance how large each batch is or how wide. So having it be purely lazy is easier

roeap:
Its perfectly fine IMO to go the polling route... 

Just for me to understand this better as I have not yet look deep enough into the nuances of runtime handling, but to me it seems we do not have that kind of control here. What I mean by that, depending on how the stream was set up the runtime might already schedule next futures (e.g. buffered stream) and it might also be purely IO bound (e.g. consuming a Flight stream). so hard to optimize for one case.

If the work is indeed CPU bound would we just not hit any `.awaits` in the code, so we never relinquish control back to the driver until the work is finished in any case? I always saw the `*_blocking` methods as a means to bridge into the sync world at the cost of blocking a thread no matter what?

ion-elgreco:
> Its perfectly fine IMO to go the polling route...
> 
> Just for me to understand this better as I have not yet look deep enough into the nuances of runtime handling, but to me it seems we do not have that kind of control here. What I mean by that, depending on how the stream was set up the runtime might already schedule next futures (e.g. buffered stream) and it might also be purely IO bound (e.g. consuming a Flight stream). so hard to optimize for one case.
> 
> If the work is indeed CPU bound would we just not hit any `.awaits` in the code, so we never relinquish control back to the driver until the work is finished in any case? I always saw the `*_blocking` methods as a means to bridge into the sync world at the cost of blocking a thread no matter what?

The stream in this case is always a DataFusion RecordbatchStream reading parquet data from storage.

roeap:
> Not sure what you mean with this? DataFusion uses tokio for CPU bounded tasks

Simply the question why `spawn_blocking` helps us with that?

roeap:
@ion-elgreco - if we go the blocking route which probably makes sense fi we want a more pull-based model, I think your initial implementation was better then using channels :).

ion-elgreco:
> > Not sure what you mean with this? DataFusion uses tokio for CPU bounded tasks
> 
> Simply the question why `spawn_blocking` helps us with that? IN either case the task would run to completion until the work is done ...

I am also no expert, but to my understanding spawn_blocking creates a separate blocking thread where we can run some cpu bound stuff, we then run an async task until completion with block_on on this thread. The other threads in the async runtime then won't starve.

Runtime::spawn would just run it on the same async runtime threads as other tasks

ion-elgreco:
> @ion-elgreco - if we go the blocking route which probably makes sense fi we want a more pull-based model, I think your initial implementation was better then using channels :).

^^ Alright, will switch it back haha

roeap:
> Runtime::spawn would just run it on the same async runtime threads as other tasks

That is my understanding as well. I do need to think thought a bit more i believe. I guess `spawn_blocking` would not change the number of threads available overall? So maybe tokio can then better reason about the nature of the task and make sure not all threads have blocking tasks? But if so, why is there a motivation to use separate runtimes for io and cpu bound tasks?

Do you happen to know some good reference (implementation) that uses blocking to schedule cpu work in a pure asynchronous context (as in not bridging to sync)?

ion-elgreco:
> > Runtime::spawn would just run it on the same async runtime threads as other tasks
> 
> That is my understanding as well. I do need to think thought a bit more i believe. I guess `spawn_blocking` would not change the number of threads available overall? So maybe tokio can then better reason about the nature of the task and make sure not all threads have blocking tasks? But if so, why is there a motivation to use separate runtimes for io and cpu bound tasks?
> 

I think spawn_blocking can create extra threads, but those blocking threads have a limit defined in the runtime configuration. 

Why separate IO and CPU runtime, I guess biggest reason because it's way easier to separate io and cpu bound tasks, and I guess in situations you have less control over how tasks are spawned? Thinking about it, actually that might make it safe to just use spawn, since we already have IO on a different runtime. With a big caveat that the current DeltaStorageBackend doesn't spawn list and get streams on the IO runtime hehe 😄 , will have to wait until `SpawnService` PR is merged into object-store.

> Do you happen to know some good reference (implementation) that uses blocking to schedule cpu work in a pure asynchronous context (as in not bridging to sync)?

No sorry :)

roeap (approved):

PR #3346, opened 2025-03-25
feat: during LakeFS file operations, skip merge when 0 changes

### LINKED ISSUE #3190: lakefs: during file operations, skip merge when 0 changes
# Description

**Use Case**
If the branch has no commited files/zero changes, don't merge branch, so that we don't get an empty commit.

Should happen before: https://github.com/delta-io/delta-rs/blob/9dc8f328859b4fe66fa82a636aae662cee22db30/crates/lakefs/src/logstore.rs#L161-L163

**Related Issue(s)**

smeyerre:
# Description
If a branch does not have any file changes, we no longer merge the branch to avoid empty commits.

# Related Issue(s)
- closes #3190 

# Documentation
N/A

### DIFF
--- a/crates/lakefs/src/client.rs
+++ b/crates/lakefs/src/client.rs
@@ -195,7 +195,7 @@ impl LakeFSClient {
             "squash_merge": true,
         });
 
-        debug!("Merging LakeFS, source `{transaction_branch}` into target `{transaction_branch}` in repo: {repo}");
+        debug!("Merging LakeFS, source `{transaction_branch}` into target `{target_branch}` in repo: {repo}");
         let response = self
             .http_client
             .post(&request_url)
@@ -226,16 +226,16 @@ impl LakeFSClient {
     pub async fn has_changes(
         &self,
         repo: &str,
-        source_branch: &str,
-        target_branch: &str,
+        base_branch: &str,
+        compare_branch: &str,
     ) -> Result<bool, TransactionError> {
         let request_url = format!(
-            "{}/api/v1/repositories/{repo}/refs/{source_branch}/diff/{target_branch}",
+            "{}/api/v1/repositories/{repo}/refs/{base_branch}/diff/{compare_branch}",
             self.config.host
         );
 
         debug!(
-            "Checking for changes between `{source_branch}` and `{target_branch}` in repo: {repo}"
+            "Checking for changes from `{base_branch}` to `{compare_branch}` in repo: {repo}"
         );
         let response = self
             .http_client
@@ -501,7 +501,7 @@ mod tests {
             let mock = server
                 .mock(
                     "GET",
-                    "/api/v1/repositories/test_repo/refs/source_branch/diff/target_branch",
+                    "/api/v1/repositories/test_repo/refs/base_branch/diff/compare_branch",
                 )
                 .with_status(StatusCode::OK.as_u16().into())
                 .with_body(response_body)
@@ -516,7 +516,7 @@ mod tests {
 
             let result = rt().block_on(async {
                 client
-                    .has_changes("test_repo", "source_branch", "target_branch")
+                    .has_changes("test_repo", "base_branch", "compare_branch")
                     .await
             });
 
--- a/crates/lakefs/src/client.rs
+++ b/crates/lakefs/src/client.rs
@@ -234,9 +234,7 @@ impl LakeFSClient {
             self.config.host
         );
 
-        debug!(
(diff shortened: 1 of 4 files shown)


smeyerre:
First time contributing here so I'll include my manual testing process here in case I missed anything!

### Manual Testing
Testing was performed using a real LakeFS instance:

1. **Environment Setup**
   - Started LakeFS using Docker Compose
   - Built delta-rs with the implemented changes

2. **Test Steps**
   - Created a Delta table with initial data (verified first commit in LakeFS UI)
   - Added more data to the table (verified second commit in LakeFS UI)
   - Performed a no-op operation (repair with dry_run=True)
   - Verified that no empty commit was created (commit count remained at 2)

3. **Results**
   - Confirmed that real operations (table creation, data addition) still create commits
   - Confirmed that no-op operations no longer create empty commits
   - The fix works correctly without affecting normal Delta Lake operations

ion-elgreco:
I had a quick glance and it looks good! You should change the test thought, a repair dry run doesn trigger pre execute and post execute.

But an easier way to trigger a file operation lakefs commit is vacuum

smeyerre:
> > Ok I'm realizing I missed a failing test earlier. It seems that when creating checkpoints, the checkpoint files aren't reflected by the LakeFS diff API.
> > I can think of a couple ways to go about handling this:
> > 
> > 1. Add something like a `force_merge` parameter to `commit_merge` that gets passed when doing checkpoint operations
> > 2. Add a more rigorous check for changes in `has_changes` by doing something like looking for checkpoint files as well
> > 
> > Option 1 seems a little bit weird to me so I'd be more inclined to do 2 but open to feedback on this (also I could have easily missed a better option).
> 
> What does the diff API return here? In theory it should show any change

Apologies, I've been out of town the last week and a half.

So in my manual testing it seems like the LakeFS diff API actually returns empty results when a checkpoint has been made. Is that not expected?

ion-elgreco:
> > > Ok I'm realizing I missed a failing test earlier. It seems that when creating checkpoints, the checkpoint files aren't reflected by the LakeFS diff API.
> > > I can think of a couple ways to go about handling this:
> > > 
> > > 1. Add something like a `force_merge` parameter to `commit_merge` that gets passed when doing checkpoint operations
> > > 2. Add a more rigorous check for changes in `has_changes` by doing something like looking for checkpoint files as well
> > > 
> > > Option 1 seems a little bit weird to me so I'd be more inclined to do 2 but open to feedback on this (also I could have easily missed a better option).
> > 
> > 
> > What does the diff API return here? In theory it should show any change
> 
> Apologies, I've been out of town the last week and a half.
> 
> So in my manual testing it seems like the LakeFS diff API actually returns empty results when a checkpoint has been made. Is that not expected?

Are you sure the test is correct? I remember we used the diff api before doing a merge on the python side and that worked

smeyerre:
> > > > Ok I'm realizing I missed a failing test earlier. It seems that when creating checkpoints, the checkpoint files aren't reflected by the LakeFS diff API.
> > > > I can think of a couple ways to go about handling this:
> > > > 
> > > > 1. Add something like a `force_merge` parameter to `commit_merge` that gets passed when doing checkpoint operations
> > > > 2. Add a more rigorous check for changes in `has_changes` by doing something like looking for checkpoint files as well
> > > > 
> > > > Option 1 seems a little bit weird to me so I'd be more inclined to do 2 but open to feedback on this (also I could have easily missed a better option).
> > > 
> > > 
> > > What does the diff API return here? In theory it should show any change
> > 
> > 
> > Apologies, I've been out of town the last week and a half.
> > So in my manual testing it seems like the LakeFS diff API actually returns empty results when a checkpoint has been made. Is that not expected?
> 
> Are you sure the test is correct? I remember we used the diff api before doing a merge on the python side and that worked

Ah ok I think I fixed the issue. It seems like I had the branch parameters to the diff API call reversed which messed with the checkpoint diffs specifically. The tests are all passing now, and my manual tests worked as well.

smeyerre:
> @smeyerre great! Can you fix the formatting? Then we're good to go 💯

Oh oops, ok that should be fixed now, unit/formatting tests are passing locally.

ion-elgreco:
@smeyerre don't worry about the other test failures, there is a breaking change in the arrow release, I will merge it once I've addressed it in main

smeyerre:
> @smeyerre don't worry about the other test failures, there is a breaking change in the arrow release, I will merge it once I've addressed it in main

Awesome thanks!

ion-elgreco (approved):
Thanks for adding this optimization!! <3

PR #3362, opened 2025-04-01
refactor: remove 'rustls' from default features of 'core'

zeevm:
# Description
The 'rustls' features on by default on the 'core' crate pulls in a lot of optional dependencies even if not used, and it doesn't seem it's required to be on by default.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -113,7 +113,7 @@ tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 
 [features]
 cdf = []
-default = ["cdf", "rustls"]
+default = ["cdf"]
 datafusion = [
     "dep:datafusion",
     "datafusion-expr",


zeevm:
@rtyler Would appreciate your reviewing this. It doesn't seem like we need 'rustls' on by default on 'core', it pulls in a lot of optional dependencies into a project Cargo.lock even if those aren't used for the final build.

rtyler:
I realize my error in the previous pull request where I tried to add a check for the default build, the default build I added was for `deltalake` not for `deltalake-core`. I was quite confused at CI going green on this change, since it fails quite clearly when I'm running this change.

`deltalake-core` does require rustls _or_ native-tls because it uses the default engine from delta-kernel

```
error[E0433]: failed to resolve: could not find `engine` in `delta_kernel`
   --> crates/core/src/kernel/mod.rs:5:19
    |
5   | use delta_kernel::engine::arrow_expression::ArrowExpressionHandler;
    |                   ^^^^^^ could not find `engine` in `delta_kernel`
    |
note: found an item that was configured out
   --> /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/delta_kernel-0.8.0/src/lib.rs:116:9
    |
116 | pub mod engine;
    |         ^^^^^^
note: the item is gated here
   --> /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/delta_kernel-0.8.0/src/lib.rs:111:1
    |
111 | / #[cfg(any(
112 | |     feature = "default-engine",
113 | |     feature = "sync-engine",
114 | |     feature = "arrow-conversion"
115 | | ))]
    | |___^

For more information about this error, try `rustc --explain E0433`.
warning: `deltalake-core` (lib) generated 1 warning
```

While I don't understand the desire to purge the dependency, I don't see a way to avoid it while still relying on the delta_kernel default engine

zeevm:
@rtyler 
The motivation for this change is to avoid optional crates that don't end up in the build from appearing in Cargo.lock, specifically "ring".
I work at Microsoft and our ci/cd flags this crate as a security issue and won't accept any commit that has it referenced in Cargo.lock
If the default will be tls and not rustls - while letting users change that - i think we'll be good.
Does that seem achievable?

ion-elgreco:
I am against changing the default to TLS just for a single user, since you can change it yourself in your own builds

zeevm:
> I am against changing the default to TLS just for a single user, since you can change it yourself in your own builds

Understood. 

@rtyler will using the engine-base feature resolve the build failure? Then the choice between tls and rustls is done by the user when referencing deltalake with the appropriate feature?

djouallah:
@rtyler can we use  native-tls then,  the package was working fine and now end users when they do pip install upgrade get a broken experience

ion-elgreco:
> @rtyler can we use native-tls then,  the package was working fine and now end users when they do pip install upgrade get a broken experience when querying onelake,  although I am a microsoft employee, I am talking as a user who used this package for years. 

I understand the frustration but the default will stay rustls, I have already said this in the previous pr. The benefits of that simply outweigh it the downside of this one lake incompatibility, which is simply Microsofts responsibility imho.

The fabric team can build custom wheels for their fabric runtime where openssl is vendored since you can now configure the features in core or put an engineer on it and look deeper why TLS resolving doesn't work and seek collaboration with the rustls devs. Microsoft has plenty of engineers so this shouldn't be hard. 

Unfortunately Onelake is the odd duck here, because aws S3, adls, GCS, minio, lakefs, Cloudflare all work fine with rustls.

ion-elgreco:
I have asked to test out a branch a while ago which has rustls native certs feature enabled in the reqwest crate, which should use the OS certs as well. Never got a response on this though..

https://github.com/delta-io/delta-rs/issues/3243#issuecomment-2676116111

PR #3365, opened 2025-04-05
chore: remove cdf feature

ion-elgreco:
# Description
@rtyler the cdf feature actually won't be used without datafusion, so just to simplify it we can gate it behind the datafusion feature


rtyler (approved):
Makes sense to me :+1:

PR #3368, opened 2025-04-07
feat: introduce VacuumMode::Full for cleaning up orphaned files

### LINKED ISSUE #2349: Vacuum command should support "full" mode too!
# Environment

**Delta-rs version**: 0.15.3

**Binding**: Python

**Environment**: 
Local Machine - MacBook M1
We are using Azure Blob Storage.

***
# Bug

**What happened**:
We are seeing an issue where the vacuum command is not removing all the old files. 

Here you can see we ran the vacuum command in Databricks and there are 137 files to remove. 
image
 
Using Delta-RS there was only 4 to remove. 
```
       dt = await get_delta_table(storage_account_path)
        result = dt.vacuum(retention_hours=720, dry_run=True)
        print(f"Number of files to vacuum: {len(result)}")
        for file in result:
            print(file)
```
Number of files to vacuum: 4
345-02f7d21b-7097-4dda-ae68-8e1c11703e53-0.parquet
346-9a154da0-b3f1-4904-be42-4ef4fdf610bc-0.parquet
347-37b41632-8304-4764-ae4d-82effe20371e-0.parquet
350-b0a9041a-13a5-4bfd-91f6-bf629d1191d9-0.parquet

We except it to remove all files as stated in the return of the function. 
"the list of files no longer referenced by the Delta Table and are older than the retention threshold."

Would there be a specific reason for this? 
I am available for further clarification or to provide additional details if needed.

rtyler:
This allows an optional but not-on-by-default mode of removing untracked
files in the delta table directory. Delta/Spark supports a "lite" and
"full" mode for [vacuum]. This change is intentionally not making "full"
the default as it is for Delta/Spark since that may have unintended
consequences for our users who have become accustomed to "lite" being
the default.

Fixes #2349

[vacuum]: https://docs.delta.io/latest/delta-utility.html#remove-files-no-longer-referenced-by-a-delta-table

### DIFF
--- a/crates/core/src/operations/vacuum.rs
+++ b/crates/core/src/operations/vacuum.rs
@@ -31,6 +31,7 @@ use futures::{StreamExt, TryStreamExt};
 use object_store::Error;
 use object_store::{path::Path, ObjectStore};
 use serde::Serialize;
+use tracing::log::*;
 
 use super::{CustomExecuteHandler, Operation};
 use crate::errors::{DeltaResult, DeltaTableError};
@@ -76,6 +77,20 @@ pub trait Clock: Debug + Send + Sync {
     fn current_timestamp_millis(&self) -> i64;
 }
 
+/// Type of Vacuum operation to perform
+#[derive(Debug, Default, Clone, PartialEq)]
+pub enum VacuumMode {
+    /// The `lite` mode will only remove files which are referenced in the `_delta_log` associated
+    /// with `remove` action
+    #[default]
+    Lite,
+    /// A `full` mode vacuum will remove _all_ data files no longer actively referenced in the
+    /// `_delta_log` table. For example, if parquet files exist in the table directory but are no
+    /// longer mentioned as `add` actions in the transaction log, then this mode will scan storage
+    /// and remove those files.
+    Full,
+}
+
 /// Vacuum a Delta table with the given options
 /// See this module's documentation for more information
 pub struct VacuumBuilder {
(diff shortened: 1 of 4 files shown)


crates/core/src/operations/vacuum.rs
```diff
@@ -76,6 +77,20 @@ pub trait Clock: Debug + Send + Sync {
     fn current_timestamp_millis(&self) -> i64;
 }
 
+/// Type of Vacuum operation to perform
+#[derive(Debug, Default, Clone, PartialEq)]
+pub enum VacuumMode {
+    /// The `lite` mode will only remove files which are referenced in the `_delta_log` associagted
+    /// with `remove` action
```
ion-elgreco, line 84: The databricks docs on this says something different and I still can't decipher  what they truly mean, is there some spec on this?

crates/core/src/operations/vacuum.rs
```diff
@@ -76,6 +77,20 @@ pub trait Clock: Debug + Send + Sync {
...
+    #[default]
+    Lite,
```
ion-elgreco, line 86: Let's also expose the mode to python 😊

rtyler: Would you be up for adding this exposure after we merge? :smiling_imp:

ion-elgreco: Sure!

ion-elgreco (dismissed):

ion-elgreco (approved):

crates/core/src/operations/vacuum.rs
```diff
@@ -76,6 +77,20 @@ pub trait Clock: Debug + Send + Sync {
     fn current_timestamp_millis(&self) -> i64;
 }
 
+/// Type of Vacuum operation to perform
+#[derive(Debug, Default, Clone, PartialEq)]
+pub enum VacuumMode {
+    /// The `lite` mode will only remove files which are referenced in the `_delta_log` associagted
```
mschmo, line 83: Obviously not a big deal, but "associagted" -> "associated"

ion-elgreco (approved):

PR #3373, opened 2025-04-08
fix: parse unconventional logs

roeap:
# Description

This PR fixes some errors where in some edge cases we would operate on tables we should not be operating on. Luckily this does not apply to current tables, but only to tables that may include features not yet defined.

These cases were uncovered when testing reads against some unusual logs that are either corrupted or represent edge cases in log layouts.

depends on: https://github.com/delta-io/delta-kernel-rs/pull/684


roeap:
@rtyler @ion-elgreco - this includes the failures on metadata reads that we discussed. we should look at some of the test failures to maybe see what the desired bahvior is. Since we were doing that check later so far we were reading some tables we now error on open ...

roeap:
All required updates to existing tests are contained in the last commit - tests where we change behavior that is.

ion-elgreco (approved):
LGTM! It is a breaking change though, so should be fine since we go 1.0 for py

PR #3374, opened 2025-04-09
fix: if field contains space in constraint expression, checks will fail

Nordalf:
# Description
When adding a constraint to a table, the expression is formatted as: `<FIELD> <CONDITION> <VALUE>`. If - god forbid - a space occurs in a field name, the condition will not evaluate due to the field name not being escaped. During the `enforce_check` we are unable to get the field name, and to avoid breaking backward, I am suggesting adding some logic to add backticks to conditions.

1. Get the table schema
2. Loop through schema and check if field exists
3. If field exists, escape with backticks

I am not fan of this solution myself, but I do not have a better solution at the moment.

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -1418,7 +1418,7 @@ impl DeltaDataChecker {
             return Ok(());
         }
         let table = MemTable::try_new(record_batch.schema(), vec![vec![record_batch.clone()]])?;
-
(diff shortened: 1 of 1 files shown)


ion-elgreco (commented):
Can you add a small test please ^^

ion-elgreco:
> I am not fan of this solution myself, but I do not have a better solution at the moment.

Maybe we can make it work with just a configuration in datafusion

rtyler (approved):

PR #3377, opened 2025-04-11
fix(datafusion): optimize partition pruning and predicate pushdown

aditanase:
# Description

This PR adds 3 optimizations for the Datafusion integration, that go well together.
I am open to split in 2 PRs if you think it's necessary, but the 2nd one would depend on the first one anyway

In the first commit:
- we are treating filters on partition columns as EXACT - meaning that Datafusion will do 2 things:
  - delegate filtering to the Delta table provider and will not add an additional filter node on top
  - trigger the push down limit optimization, that we take advantage of in the 2nd commit
- we are now also pruning files based on the partition column predicate, by implementing the `contained` method which used to be a no-op
  - we can now do a 1st phase of pruning if LiteralGuarantees are present (e.g. column = or in list)

In the 2nd commit:
- we are now taking advantage of the push down limit optimization and will prune files based on row counts, up to the limit requested.
  - of course, this is only triggered if we are ONLY filtering on partition columns
  - this is something that already happens in the spark delta source, which served as inspiration for this patch

# Related Issue(s)
- https://github.com/delta-io/delta/issues/2421
- https://github.com/delta-io/delta/pull/1495

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -27,7 +27,9 @@ use std::fmt::{self, Debug};
 use std::sync::Arc;
 
 use arrow_array::types::UInt16Type;
-use arrow_array::{Array, DictionaryArray, RecordBatch, StringArray, TypedDictionaryArray};
+use arrow_array::{
+    Array, BooleanArray, DictionaryArray, RecordBatch, StringArray, TypedDictionaryArray,
+};
 use arrow_cast::display::array_value_to_string;
 use arrow_cast::{cast_with_options, CastOptions};
 use arrow_schema::{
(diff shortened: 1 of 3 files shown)


aditanase:
Please note this only deals with `Utf8` partitioning columns right now, as that is how the `partitionValues` are stored in the stats. I am in the process of adding support for Dictionary encoded Utf8, but for the rest of the types I am not 100% sure if I should be bailing at the top or attempt a conversion to string before comparing with the `partitionValues`

aditanase:
Added some tests and fixes for the decision around Exact/Inexact predicate pushdown. I was being overly optimistic initially, I took some inspiration from the datafusion ListingTable provider, but simplified it to be better aligned with the PruningPredicate.

I am still planning to add a few tests similar to `test_files_scanned` that confirm the number of files scanned, especially in combination with push-down limits.

Would ar

aditanase:
Rebased and squashed, as nobody started a review yet.

aditanase:
Realized over the weekend that I was breaking a DF optimization that can resolve queries such as
`SELECT COUNT(*) FROM table WHERE partition_col = 'value';`

Added another commit that adds a temporary fix for pruning stats in addition to files, waiting for feedback on implementation.

rtyler:
The code generally looks good to me. I would recommend get this into whatever state you feel comfortable merging in

aditanase:
@rtyler rebased and squashed. LMK what other areas I should look at

PR #3382, opened 2025-04-12
refactor!: move storage module into logstore

roeap:
# Description

TLDR;
* `storage` -> `logstore::storage`
* separate `runtime` mod for io runtime
* code movement

Another pre-factor PR, this time we move the `storage` module into `logstore`.

While working on adding a caching layer for log reads I realised that we are exposing quite a bit of functionality via the storage module and interactions with the logstore module require quite a bit of code parsing. With kernel coming in we will have to deal with yet another storage abstraction.

The good news, in practice we actually have a quite narrow waist when creating the tables, so we can just hook into the approach we take for wrapping the io runtime. Still I believe, that logically the `ObjectStore` sits behind teh log store, or at least the log store is responsible for providing it.

In an immediate follow up I would like to add the cached store (@rtyler - in case you have some time to run bench marks? 😄) and with that make do a pass over our config for storage to make sure it stays manageable as we add yet another set of options.


ion-elgreco:
Just an fyi, the IO runtime wrapping will hopefully replaced by a kind of Spawservice, the initial PR of Tustvold became stale but I'll try to push it forward from there with Andrew, https://github.com/apache/arrow-rs/pull/7253#issuecomment-2798881313

ion-elgreco (approved):
Lgtm! :) Im curious to see your caching layer and on what abstraction level it will be because there are some quite through implementations out there at least on an ObjectStore level

roeap:
> Just an fyi, the IO runtime wrapping will hopefully replaced by a kind of Spawservice,

Awesome, looking forward to that :)

>  there are some quite through implementations out there at least on an ObjectStore level

@ion-elgreco - do you have some links maybe? 😄. The main idea is though to keep it quite simple on that layer - our most effective caching is home we managed the parsed actions :).

Another optimisation, that is not part of this (but I'm hoping to get to one day) is to chace parquet footers - this is somethign we can implement on the DF level ... That said, the PR should be coming up shortly, and we can continue there :).

ion-elgreco:
@roeap the overview is here https://github.com/apache/arrow-rs-object-store/issues/14

> Caches results of requests locally using memory / disk (see [ObjectStoreMemCache](https://github.com/influxdata/influxdb3_core/tree/main/object_store_mem_cache) in influxdb3_core), and [this one](https://github.com/slatedb/slatedb/blob/main/src%2Fcached_object_store%2Fobject_store.rs) in slatedb @criccomini (thanks @ion-elgreco for the pointer)

PR #3383, opened 2025-04-13
feat!: update storage configuration system

roeap:
# Description

Along with the addition of a caching layer, this PR also attempts to formalise how we handle configuration a bit more. With the caching, we can now configure four layers that can be applied to object stores.

* limiting
* io runtime
* retry
* cache

Additionally, we currently need to wrap stores in a `PrefixStore` pointing at the table root.

Each of these layers has its own configuration that is parsed from the same global map of configuration keys. The configuration for all of these is now aggregated in a new struct `StorageConfig`. 

Our previous abstraction `StorageOptions` had over time lost its utility since no implementations lived in the core crate anymore and we were just wrapping a `HashMap`. 

The same configuration is also passed again to `LogStore` and `ObjectStore` factories where again some parsing happens. With this PR we now present the already parsed configuration (`StorageConfig`) which exposes utility methods for integrations to decorate / augment the storages the produce.

The new caching store is fairly contained, and I'd be happy to carve that out if it helps reviews.


crates/core/src/logstore/storage/cache.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+impl std::fmt::Display for CommitCacheObjectStore {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "CommitCacheObjectStore({})", self.inner)
+    }
+}
+
+fn cache_delta(path: &Url) -> bool {
```
ion-elgreco, line 184: Small nit to improve readability
```suggestion
fn is_delta_commit(path: &Url) -> bool {
```

crates/core/src/logstore/storage/cache.rs
```diff
@@ -0,0 +1,327 @@
...
+            _ => return Ok(None),
+        }
+        Ok(Some(()))
+    }
+}
+
+async fn get_default_cache(
+    config: LogCacheConfig,
+) -> DeltaResult<(DeltaLogCache, Option<TempDir>), CachedStoreError> {
+    let dir = tempfile::tempdir()?;
```
ion-elgreco, line 103: I believe this crate will cause issues, I even saw it in tests, tempdir gets destroyed before you actually hit the correct code path when you have tempdir constructed in another function scope

crates/core/src/logstore/storage/cache.rs
```diff
@@ -0,0 +1,327 @@
...
+        if options.range.is_some() || !cache_delta(&cache_key) || options.head {
+            return self.inner.get_opts(location, options).await;
+        }
+
+        let store = self.inner.clone();
+        let loc = location.clone();
+
+        let entry = self
+            .cache
+            .fetch(cache_key, || async move {
```
ion-elgreco, line 243: Since this can be IO bound when we read from tmp directory, we should take care that this runs on the IO runtime. From what I can see Foyer supports configuring a runtime. Taking into account that with this change: https://github.com/apache/arrow-rs-object-store/pull/332, each store will be configured where requests are spawn on a different runtime

crates/core/src/logstore/storage/cache.rs
```diff
@@ -0,0 +1,327 @@
...
+            max_size_disk_mb: 256,
+            cache_directory: None,
+        }
+    }
+}
+
+impl config::TryUpdateKey for LogCacheConfig {
+    fn try_update_key(&mut self, key: &str, v: &str) -> DeltaResult<Option<()>> {
+        match key {
+            "max_size_memory_mb" => self.max_size_memory_mb = config::parse_usize(v)?,
```
ion-elgreco, line 91: I like these changes a lot, but we should think of how to document these well. Especially in the python docs. Maybe we should introduce a DeltaConfig typedDict in python that handles all these special configs:

- limit
- retry
- cache
- ...

crates/core/src/logstore/config.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    /// Runtime configuration.
+    ///
+    /// Configuration to set up a dedicated IO runtime to execute IO related operations.
+    pub runtime: Option<RuntimeConfig>,
+
+    #[cfg(feature = "cloud")]
+    pub retry: ::object_store::RetryConfig,
+
+    pub limit: Option<LimitConfig>,
```
ion-elgreco, line 110: Should also add some nice descriptions here :)

ion-elgreco (commented):
Looks very good, happy to see the simplification of StorageConfig, makes it also easier in the future to extend it! 

Just left some minor comments

crates/core/src/logstore/config.rs
roeap: absolutely, will add more docs overall ..

crates/core/src/logstore/storage/cache.rs
roeap: YES! this one is getting out of hand quite a bit already, but in a follow up we should ideally do that!

crates/core/src/logstore/storage/cache.rs
roeap: Good point, I'll see how we can wire that through.

crates/core/src/logstore/storage/cache.rs
roeap: i think you are right ... now that you mention it i remember some issues as well.

crates/core/src/logstore/config.rs
```diff
@@ -104,16 +105,56 @@ pub struct StorageConfig {
...
+    ///
+    /// Depending on the configuration, the following layers may be added:
+    /// - Retry layer: Adds retry logic to the object store.
+    /// - Limit layer: Limits the number of concurrent requests to the object store.
+    /// - Runtime layer: Executes IO related operations on a dedicated runtime.
+    pub fn decorate_store<T: ObjectStore + Clone>(
+        &self,
+        store: T,
+        table_root: &url::Url,
+        handle: Option<Handle>,
```
ion-elgreco, line 131: The IO handling will later be declared within the initial object store creation, it can't be done afterwards, maybe take that into consideration, it might be short lived

crates/core/src/logstore/config.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    pub(crate) fn decorate_prefix<T: ObjectStore>(
+        store: T,
+        table_root: &url::Url,
+    ) -> DeltaResult<Box<dyn ObjectStore>> {
+        let prefix = match ObjectStoreScheme::parse(table_root) {
+            Ok((ObjectStoreScheme::AmazonS3, _)) => Path::parse(table_root.path())?,
+            Ok((_, path)) => path,
+            _ => Path::parse(table_root.path())?,
+        };
```
roeap, line 152: This part is concerning me a bit - why does AWS need special treatment? However as we move to working with URLs, I hope we can create a very narrow code path and convert urls to paths in exactly one way ...

crates/core/src/logstore/factories.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// The path segment is returned as second element of the tuple. It must point at the path
+    /// corresponding to the path segment of the URL.
+    ///
+    /// The store should __NOT__ apply the decorations via the passed `StorageConfig`
+    fn parse_url_opts(
+        &self,
+        url: &Url,
+        options: &HashMap<String, String>,
+        retry: &RetryConfig,
+    ) -> DeltaResult<(ObjectStoreRef, Path)>;
```
roeap, line 31: Here we document the new expected behavior of this method.

crates/core/src/logstore/mod.rs
```diff
@@ -25,37 +73,65 @@ use crate::kernel::Action;
...
+        &self,
+        store: ObjectStoreRef,
+        location: &Url,
+        options: &StorageConfig,
+        io_runtime: Option<IORuntime>,
+    ) -> DeltaResult<Arc<dyn LogStore>> {
+        let store = options.decorate_store(store, location, io_runtime.map(|r| r.get_handle()))?;
+        self.with_options(Arc::new(store), location, options)
+    }
+}
```
roeap, line 116: Right now we use the internal trait to make sure we are injection ourselves into the flow to wrap our layers around the object store. I feel this can be improved, bit hoping to do that in a follou up where we can have a more targeted discussion around this.

crates/core/src/logstore/mod.rs
```diff
@@ -179,8 +240,24 @@ pub enum PeekCommit {
...
+        handle: Option<Handle>,
+    ) -> DeltaResult<Box<dyn ObjectStore>> {
+        let table_url = table_root.unwrap_or(&self.location);
+        self.options.decorate_store(store, table_url, handle)
+    }
+
+    pub fn object_store_factory(&self) -> ObjectStoreFactoryRegistry {
+        self::factories::object_store_factories()
+    }
 }
```
roeap, line 261: the logstore config is a good way for us to inject functionality into the logstore, since the API dictates log stores also expose it again, but we control the implementation... this should alos gove us a way to just inject a registry taken from - let's say datafusion session - and wire it through.

crates/core/src/logstore/storage/mod.rs
```diff
@@ -143,121 +100,22 @@ impl ObjectStoreRegistry for DefaultObjectStoreRegistry {
...
-    } else {
-        Arc::new(store)
-    }
-}
-
-/// Storage option keys to use when creating [ObjectStore].
-///
-/// The same key should be used whether passing a key in the hashmap or setting it as an environment variable.
-/// Must be implemented for a given storage provider
-pub mod storage_constants {
```
roeap, line 223: i could not find any way we actually inspect the environment to pichk up this variable. that said, wit the updated configuration parsing we should now have a clear way how we can enable much more config via then environment in a manageable way.

crates/core/src/table/builder.rs
```diff
@@ -10,33 +10,9 @@ use serde::{Deserialize, Serialize};
...
-    MissingFeature { feature: &'static str, url: String },
-    #[error("Failed to parse table uri")]
-    TableUri(#[from] url::ParseError),
-}
-
-impl From<BuilderError> for DeltaTableError {
-    fn from(err: BuilderError) -> Self {
-        DeltaTableError::Generic(err.to_string())
-    }
-}
```
roeap, line 39: we never used any of these 😄

crates/core/src/table/builder.rs
```diff
@@ -54,20 +30,13 @@ pub enum DeltaVersion {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct DeltaTableConfig {
-    /// Indicates whether our use case requires tracking tombstones.
-    /// This defaults to `true`
-    ///
-    /// Read-only applications never require tombstones. Tombstones
-    /// are only required when writing checkpoints, so even many writers
-    /// may want to skip them.
-    pub require_tombstones: bool,
```
roeap, line 63: we never reference this config ... reading tomstones has been lazy for a while now.

python/src/lib.rs
```diff
@@ -1334,14 +1324,6 @@ impl RawDeltaTable {
         })
     }
 
-    pub fn get_py_storage_backend(&self) -> PyResult<filesystem::DeltaFileSystemHandler> {
-        Ok(filesystem::DeltaFileSystemHandler {
-            inner: self.object_store()?,
-            config: self._config.clone(),
-            known_sizes: None,
-        })
-    }
```
roeap, line 1343: also a relic ...

crates/core/src/logstore/config.rs
ion-elgreco: Seems not necessary both return the same

roeap: agreed! but if we don't do this, tests start to fail .. i was also wondering if maybe the aws cli behaves differently encoding paths when uploading files?

roeap:
@ion-elgreco - your comments should be addressed now.

crates/core/src/logstore/storage/cache.rs
```diff
@@ -212,7 +208,14 @@ impl CommitCacheObjectStore {
         root_url: url::Url,
         config: LogCacheConfig,
     ) -> DeltaResult<Self> {
-        let (cache, cache_dir) = get_default_cache(config).await?;
+        let (path, cache_dir) = if let Some(dir) = &config.cache_directory {
+            let path = std::fs::canonicalize(dir)?;
+            (path, None)
+        } else {
+            let tmp_dir = tempfile::tempdir()?;
```
ion-elgreco, line 215: Have you tested this? I feel like this will still get destroyed before actually access those tmp files

roeap: this is not hooked up yet, the plan is to actually wire this up in the next PR, then I'll also add tests.

roeap: happy to also just remove the cache impl from this PR, and put it all in hte next one.a follow up.

roeap: went with just removing it ... better then introducing dead code. should hopefully make for a cleaner review in the next one 😄.

ion-elgreco: Yeah because the docs say, once TempDir goes out of scope, it's deconstructed

ion-elgreco (approved):

rtyler (approved):

crates/core/src/logstore/config.rs
zeevm: @roeap @ion-elgreco Why do we wrap with a PrefixStore() always? This totally breaks the DeltaTableBuilder:
    /// Set the storage backend.
    ///
    /// If a backend is not provided then it is derived from `table_uri`.
    ///
    /// # Arguments
    ///
    /// * `storage` - A shared reference to an [`ObjectStore`](object_store::ObjectStore) with
    ///   "/" pointing at delta table root (i.e. where `_delta_log` is located).
    /// * `location` - A url corresponding to the storagle location of `storage`.
    pub fn with_storage_backend(mut self, storage: Arc<DynObjectStore>, location: Url) -> Self {
        self.storage_backend = Some((storage, location));
        self
    }

this api is now broken, the url paths will be all wrong since the passed location is now prefixed and not relative to the assumed root of the storage backend

ion-elgreco: This is to prepare for support for absolute paths, but if you are seeing something wrong. Please create a repro :)

PR #3389, opened 2025-04-16
feat: derive macro for config implementations

roeap:
# Description

Managing our configuration has become more and more complex over time. Not only are we exposing more and more options, but we also must allow integrations to receive configuration values when we invoke factory methods. This means maintaining different pieces of config in various places and making sure we have consistent behaviours all around. 

In this PR we propose to use derive macros to lessen the burden of maintenance and provide a more robust experience.

When defining a new piece of config we derive `DeltaConfig` and optionally apply additional config to individual fields.

```rs
#[derive(Debug, Clone, Default, DeltaConfig)]
pub struct LimitConfig {
    #[delta(alias = "concurrency_limit", env = "OBJECT_STORE_CONCURRENCY_LIMIT")]
    pub max_concurrency: Option<usize>,
}
```

This will implement the `TryUpdateKey` trait, which is now extended by `TryUpdateKey::load_from_environment` as well as `FromIterator` of string like tuples.

With this PR we would establish a pattern, but there are a few things we may want to consider in the future like adding namespaces `a.b.c` etc ... which would of course only mean updating the macro.

in a follow up PR I would look into fully leveraging the macro throughout all crates.

### DIFF
--- a/crates/core/src/logstore/storage/mod.rs
+++ b/crates/core/src/logstore/storage/mod.rs
@@ -6,8 +6,8 @@ use object_store::path::Path;
 use object_store::{DynObjectStore, ObjectStore};
 use url::Url;
 
-use super::config;
 use crate::{DeltaResult, DeltaTableError};
+use deltalake_derive::DeltaConfig;
 
 pub use retry_ext::ObjectStoreRetryExt;
 pub use runtime::{DeltaIOStorageBackend, IORuntime};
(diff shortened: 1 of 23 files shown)


crates/core/src/logstore/storage/mod.rs
```diff
@@ -98,22 +98,88 @@ impl ObjectStoreRegistry for DefaultObjectStoreRegistry {
...
-            // for more information
-            "OBJECT_STORE_CONCURRENCY_LIMIT" | "concurrency_limit" => {
-                self.max_concurrency = Some(config::parse_usize(v)?);
+#[cfg(test)]
+#[allow(drop_bounds, unused)]
+mod tests {
+    use std::collections::HashMap;
+    use std::env;
+
+    use rstest::*;
```
roeap, line 113: Seems like its been too long that I looked into test frameworks for rust. Learned about `rstest` with this PR and I think using fixtures and other mechanics it offers, we can greatly reduce some of our boilerplate.

The kernel PR also adds dat tests for rust, which I want to carve out and use as an opportunity to maybe refactor some of our test scaffolding.

crates/core/src/logstore/storage/mod.rs
```diff
@@ -98,22 +98,88 @@ impl ObjectStoreRegistry for DefaultObjectStoreRegistry {
...
+
+    use super::*;
+    use crate::logstore::config::TryUpdateKey;
+
+    #[fixture]
+    pub fn with_env(#[default(vec![])] vars: Vec<(&str, &str)>) -> impl Drop {
```
ion-elgreco, line 119: Oh this is cool!

crates/derive/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+/// This macro automatically implements TryUpdateKey for a struct,
+/// mapping field names to configuration keys and using appropriate parsers
+/// based on the field type.
+///
+/// Additional key aliases can be specified with the `#[delta(alias = "alias.name")]` attribute.
+/// Multiple aliases can be added by `#[delta(alias = "foo", alias = "bar")]`.
+///
+/// Reading configuration can be achieved by assigning environmene keys to a field
+/// `#[delta(env = "MY_ENV_KEY")]`.
+#[proc_macro_derive(DeltaConfig, attributes(delta))]
```
ion-elgreco, line 24: I fully trust your dark magic here, but maybe it makes sense to move those derive tests to here

ion-elgreco: Oh that's strange, I didn't know that xd

ion-elgreco (approved):

PR #3378, opened 2025-04-11
feat: upgrade to DataFusion 47.0.0

alamb:
# Description
upgrade to DataFusion 47.0.0

# Related Issue(s)

- Related to https://github.com/apache/datafusion/issues/15072

As in the past, I am making a PR to upgrade to DataFusion 47 to both:
1. help delta.rs
2. to test DataFusion 47 before we release it

# Documentation

PR #3390, opened 2025-04-16
fix: clippy warnings

alamb:
# Description
I would like to get a clean clippy run, to make it easier to find what deprecated APIs I need to update when upgrading to DataFusion 47:
- https://github.com/delta-io/delta-rs/pull/3378

I ran this command:
```shell
cargo clippy --all-features --all-targets
```

And basically did what clippy told me. I am happy to add a check for clippy to the CI as well, but figured that might be better as a follow on PR if you would like it.

# Related Issue(s)
- https://github.com/delta-io/delta-rs/pull/3378

# Documentation

### DIFF
--- a/crates/core/src/operations/write/async_utils.rs
+++ b/crates/core/src/operations/write/async_utils.rs
@@ -28,6 +28,7 @@ impl AsyncShareableBuffer {
     }
 
     /// Returns a clone of the underlying buffer as a `Vec`.
+    #[allow(dead_code)]
     pub async fn to_vec(&self) -> Vec<u8> {
         let inner = self.buffer.read().await;
         inner.clone()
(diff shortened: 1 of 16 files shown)


crates/core/src/operations/write/async_utils.rs
```diff
@@ -28,6 +28,7 @@ impl AsyncShareableBuffer {
     }
 
     /// Returns a clone of the underlying buffer as a `Vec`.
+    #[allow(dead_code)]
```
alamb, line 31: some of these things that clippy claims were dead seemed important so I didn't actually removed them, I just added an allow. 

I am happy to remove them too if you prefer

roeap (approved):
LGTM!

ion-elgreco:
> LGTM!

Ship it:p

PR #3419, opened 2025-05-07
feat: added a check for gc code to run

### LINKED ISSUE #3180: Generated columns feature needs to be checked
# Environment
I never added a validation whether generated columns feature is enabled in the tablefeatures or writer version 5 is enabled, we should add a check before running GC code just to be sure.

JustinRush80:
# Description
Add a validation check that only runs if generated columns feature is enabled in the tablefeatures or writer version 5 is enabled

# Related Issue(s)
- closes #3180

# Documentation

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -84,7 +84,7 @@ use crate::operations::cdc::*;
 use crate::operations::merge::barrier::find_node;
 use crate::operations::write::execution::write_execution_plan_v2;
 use crate::operations::write::generated_columns::{
-    add_generated_columns, add_missing_generated_columns,
+    able_to_gc, add_generated_columns, add_missing_generated_columns,
 };
 use crate::operations::write::WriterStatsConfig;
 use crate::protocol::{DeltaOperation, MergePredicate};
@@ -722,7 +722,7 @@ impl ExtensionPlanner for MergeMetricExtensionPlanner {
 #[allow(clippy::too_many_arguments)]
 async fn execute(
     predicate: Expression,
-    source: DataFrame,
+    mut source: DataFrame,
     log_store: LogStoreRef,
     snapshot: DeltaTableState,
     _state: SessionState,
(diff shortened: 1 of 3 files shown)


crates/core/src/operations/write/mod.rs
```diff
@@ -445,18 +445,26 @@ impl std::future::IntoFuture for WriteBuilder {
...
+            let mut generated_col_exp = None;
+            let mut missing_gen_col = None;
+            let mut source = DataFrame::new(state.clone(), this.input.unwrap().as_ref().clone());
+            if let Some(snapshot) = &this.snapshot {
+                if able_to_gc(snapshot)? {
+                    let generated_col_expressions = this
+                        .snapshot
+                        .as_ref()
+                        .map(|v| v.schema().get_generated_columns().unwrap_or_default())
+                        .unwrap_or_default();
```
ion-elgreco, line 458: You can do snapshot.schema() now since you already checked existence :)

crates/core/src/operations/merge/mod.rs
```diff
@@ -779,17 +779,27 @@ async fn execute(
...
+    if able_to_gc(&snapshot)? {
+        let generated_col_expressions = snapshot
+            .schema()
+            .get_generated_columns()
+            .unwrap_or_default();
+
+        let (source, missing_generated_columns) =
+            add_missing_generated_columns(source.clone(), &generated_col_expressions)?;
+
+        source_with_gc = Some(source);
```
ion-elgreco, line 795: small nit, but we can do same as you did in the write mod, just use a mut source

ion-elgreco (commented):
Just two nits, otherwise we are good to go! :)

Thanks for picking this up

crates/core/src/operations/write/mod.rs
JustinRush80: Ahh thank you for the catch!

ion-elgreco:
@JustinRush80 is it ready?

JustinRush80:
> @JustinRush80 is it ready?

Yes I believe so should I be concerned about the failed coverage check?

hntd187 (approved):
Think Ion got most of it

ion-elgreco:
> > @JustinRush80 is it ready?
> 
> Yes I believe so should I be concerned about the failed coverage check?

No it got broken since we added the `dat` tests, so no worries. Can you rebase, so that it can auto merge?

Issue #3422, opened 2025-05-08
delta-rs includes pending versions written by spark


corwinjoy:
# Environment

**Delta-rs version**:
0.25.5

**Binding**: 
Rust

**Environment**:
**OS**: Network file store

***
# Bug
Issue reported by Callum Edwards:

> The spark [delta.io](http://delta.io/) writer seems to write failed (+pending?) versions to _delta_log/.tmp. However, the LocalFileSystem implementation of the ObjectStore in delta-rs does a recursive walk of the _delta_log directory for a list operation.
> This in turn means that failed (and maybe uncommitted) files are returned on the list operation used in the load_with_datetime method of the DeltaTable. This causes the library to crash if the failed version (in the .tmp dir) has since been pruned from the parent _delta_log dir.

Doing some investigation I see that according to the delta protocol:
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#delta-log-entries
"Delta files are stored as JSON in a directory at the **root** of the table named _delta_log, and together with checkpoints make up the log of all changes that have occurred to a table."

In addition, looking at the Java code for Hadoop (and other variations), it seems that pending changes are first written to a temp directory and then later renamed.

https://github.com/delta-io/delta/blob/c7a71d784cfe627decefba50b4ebc54769d13011/docs/source/delta-storage.md?plain=1#L260

> In this `LogStore` implementation, a temp file is created containing a copy of the metadata to be committed into the Delta log. Once that commit to the Delta log is complete, and after the corresponding DynamoDB entry has been removed, it is safe to delete this temp file. In practice, only the latest temp file will ever be used during recovery of a failed commit.

Java code:
https://github.com/delta-io/delta/blob/c7a71d784cfe627decefba50b4ebc54769d13011/storage/src/main/java/io/delta/storage/HadoopFileSystemLogStore.java#L104

So, it looks like it makes sense to change `load_with_datetime` to only look at the root level of the _delta_log path when listing files in the ObjectStore. I will investigate this change but would appreciate any feedback.

corwinjoy:
@adamreeve

PR #3423, opened 2025-05-10
Ignore temp log entries

corwinjoy:
# Description
Fix delta-rs methods so that delta_log entries in subdirectories, representing pending but uncommited transactions, are ignored.

# Related Issue(s)
Closes [load_with_datetime](https://github.com/delta-io/delta-rs/issues/3422)

# Documentation
As per the bug report only root level delta files should be included:

According to the delta protocol:
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#delta-log-entries
"Delta files are stored as JSON in a directory at the **root** of the table named _delta_log, and together with checkpoints make up the log of all changes that have occurred to a table."


crates/core/src/kernel/snapshot/log_segment.rs
```diff
@@ -480,6 +480,7 @@ async fn list_log_files_with_checkpoint(
     let version_prefix = format!("{:020}", cp.version);
     let start_from = log_root.child(version_prefix.as_str());
 
+    // QUESTION: DOES THIS NEED TO BE FILTERED TO ELIMINATE temp subdirs?
```
corwinjoy, line 483: I think this function, plus the one below may need to be fixed as well to ignore temp subdirectories but I could use some feedback here + pointers to the associated unit test.

crates/test/tests/data/simple_table/_delta_log/.tmp/00000000000000000005.json
```diff
@@ -0,0 +1,2 @@
+{"commitInfo":{"timestamp":1587968626637,"operation":"WRITE","operationParameters":{"mode":"Overwrite","partitionBy":"[]"},"readVersion":4,"isBlindAppend":true}}
+{"add":{"path":"part-00000-c1777d7d-89d9-4790-b38a-6ee7e24456b1-c001.snappy.parquet","partitionValues":{},"size":262,"modificationTime":1587968626600,"dataChange":true}}
```
corwinjoy, line 2: This is the pending operation in the .tmp subfolder + parquet file below.
Essentially just copied from existing transactions.

EnricoMi:
What about ignoring any `.` file (not only `.tmp`) unless the file is specifically supported (e.g. `.*.crc`)?

Prefixing a file with a `.` expresses "please ignore".

adamreeve:
> What about ignoring any `.` file (not only `.tmp`) unless the file is specifically supported (e.g. `.*.crc`)?

This fix will already ignore any files or directories beginning with `.`, as it excludes files in any subdirectories regardless of the subdirectory name, and the existing code skips any files where the name doesn't match the [log file name format](https://github.com/delta-io/delta-rs/blob/e47bba56944b0b3ea742328113a65176c6f63a99/crates/core/src/logstore/mod.rs#L509-L510).

I wonder if it would make sense for `object-store` to support a non-recursive list operation as a longer-term fix? This seems like it would be a pretty common operation, and that way this operation could be optimised in object store implementations where possible, rather than needing to filter on the client side. Although for this scenario in Delta Lake I guess there won't often be many nested results to filter out.

crates/core/tests/time_travel.rs
```diff
@@ -13,6 +15,11 @@ async fn time_travel_by_ds() {
         ("00000000000000000002.json", "2020-05-03T22:47:31-07:00"),
         ("00000000000000000003.json", "2020-05-04T22:47:31-07:00"),
         ("00000000000000000004.json", "2020-05-05T22:47:31-07:00"),
+        // Final file is uncommitted by Spark and is in a .tmp subdir
```
EnricoMi, line 18: Since this fix ignores _any_ sub-directory, the test should also contain a testcase of a sub-directory not starting with a dot.

crates/core/src/logstore/mod.rs
```diff
@@ -534,14 +534,25 @@ pub async fn get_latest_version(
...
         let object_store = log_store.object_store(None);
-        let mut files = object_store.list_with_offset(prefix, &offset_path);
+        let mut files = object_store.list_with_offset(Some(prefix), &offset_path);
         let mut empty_stream = true;
 
         while let Some(obj_meta) = files.next().await {
             let obj_meta = obj_meta?;
+            let location_path: Path = obj_meta.location.clone();
+            let parts: Vec<_> = location_path.prefix_match(prefix).unwrap().collect();
+            if parts.len() > 1 {
```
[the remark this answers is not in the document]
corwinjoy: This makes sense except for the error part. list_with_offset only lists objects with the given prefix so this should be impossible, and if it somehow still does then a panic is appropriate.

crates/core/tests/time_travel.rs
corwinjoy: I don't understand why we need a second test case? The name of the subdirectory does not matter, any and all files in subdirectories are ignored. I only used the name `.tmp` to clarify what is going on / better match what spark does.

corwinjoy:
@rtyler Thanks for reviewing this and merging!
Just to confirm, I guess the conclusion was that the functions `list_log_files` and `list_log_files_with_checkpoint` in `log_segment.rs` do not need to be updated to filter out temp files?
If they do, I am happy to add a followup.
Anyway, thanks again!

PR #3426, opened 2025-05-13
feat: spawn io with spawn service

ion-elgreco:
# Description
Using the superior spawn request connector over the deltaIOStoragebackend: https://github.com/apache/arrow-rs-object-store/pull/332/files#diff-3afe0961b5df6dc75c903371865daa9b80c8da3912c9b96df5b51613a5695335

Had to slightly refactor logstore_with and logstore_for to pass the StorageConfig around, which now can hold on the IORuntime directly, so that the `lakefs` integration and others can rebuild a new store with the same io_runtime

- closes https://github.com/delta-io/delta-rs/issues/3427

### DIFF
--- a/crates/aws/src/storage.rs
+++ b/crates/aws/src/storage.rs
@@ -18,7 +18,9 @@ use deltalake_core::{DeltaResult, DeltaTableError, ObjectStoreError, Path};
 use futures::stream::BoxStream;
 use futures::Future;
 use object_store::aws::AmazonS3;
+use object_store::client::SpawnedReqwestConnector;
 use object_store::RetryConfig;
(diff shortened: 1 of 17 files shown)


rtyler:
I have pulled out the arrow pin to land separately to unblock other PRs  :+1: 

I need to cut a release before this change lands because this is a breaking change and will require a bump to 0.27.x. I also need to read up a little bit more on the work you did with object_store to understand the purpose of the spawn service :detective:

rtyler (approved):
I have done some acceptance testing of this change and it's ready to roll. This will require a :crab: minor version change before it's released to crates.io which I can handle.  :crossed_fingers:

PR #3437, opened 2025-05-16
chore: experiment with using sccache in GitHub Actions

rtyler:


rtyler:
With sccache: 
image

Without sccache:
image

Looking promising!

ion-elgreco:
> 
> Looking promising!
Wow huge difference!

ion-elgreco:
@rtyler don't you want it to be added to the other jobs in python_build as well?

rtyler:
> @rtyler don't you want it to be added to the other jobs in python_build as well?

There aren't repeated invocations of `cargo` where it would be beneficial IMHO. If maturin changes RUSTFLAGS underneath, it would bust the cache anyways with Python, so I doubt it's worth the effort

ion-elgreco (approved):

PR #3438, opened 2025-05-16
feat: replace delta-rs cdf reading with kernel based cdf reading

hntd187:
# Description
This replaces the impl of cdf reading with kernel cdf reading. This changes the api a bit to match what kernel uses so `.load_cdf()` becomes `.table_changes()`

This is in a rough state right now, but I'll continue to iterate on this. Looking for early feedback.

Oh and I need to fix tests and port over some of the tests from load_cdf that was removed.


roeap:
@hntd187 - LogStore now exposes an engine, maybe worth looking into?

PR #3439, opened 2025-05-17
feat: implement object store that caches to disk

PeterKeDer:
# Description

This adds a custom object store that caches files to disk. I've seen some discussions around but no implementation of it yet, so I'm opening this quick PR to gather opinions. Very open on changing the API or implementation of it.

This solution is *very simple* (possibly overly-naive) and specific to our use-case of delta-rs, so there might be some edge cases from newer features that I did not consider. To enable it, you specify a `file_cache_path`, and it downloads all files requested by get requests and saves them to this path. For range requests, this still downloads the entire file which means it will be worse for performance in many cases on the first run- I think there's a lot of potential optimizations to be done here.

There is also an option to specify a cache duration for `_last_checkpoint` which to my knowledge is the only file that is mutable. We found that this increases performance by using the local (and potentially outdated) checkpoint parquet because loading incremental json files are way faster than a new checkpoint parquet.

We've been using this at our company for the last couple weeks. It has been very helpful for loading large datasets, so we think a caching feature like this would be quite valuable.

Usage example:
```python
dt = DeltaTable(
  "s3://bucket/table",
  storage_options={
    "file_cache_path": "/Users/peter/delta-cache",
    "file_cache_last_checkpoint_valid_duration": "30m",
})
# do stuff with dt
```

### DIFF
--- a/crates/core/src/logstore/config.rs
+++ b/crates/core/src/logstore/config.rs
@@ -6,11 +6,12 @@
 //!
 //! Specific pieces of configuration must implement the `TryUpdateKey` trait which
 //! defines how to update internal fields based on key-value pairs.
-#[cfg(feature = "cloud")]
-use ::object_store::RetryConfig;
+
 use object_store::{path::Path, prefix::PrefixStore, ObjectStore};
 use std::collections::HashMap;
+use std::sync::Arc;
 
+use super::storage::file_cache::{self, FileCacheConfig};
 use super::storage::LimitConfig;
 use super::{storage::runtime::RuntimeConfig, IORuntime};
 use crate::{DeltaResult, DeltaTableError};
@@ -103,6 +104,11 @@ pub struct StorageConfig {
     /// Configuration to limit the number of concurrent requests to the object store.
     pub limit: Option<LimitConfig>,
 
+    /// File cache configuration.
+    ///
+    /// Configuration to enable file cache for the Delta Log Store.
+    pub file_cache: Option<FileCacheConfig>,
+
     /// Properties that are not recognized by the storage configuration.
     ///
     /// These properties are ignored by the storage configuration and can be used for custom purposes.
@@ -121,12 +127,27 @@ impl StorageConfig {
     /// Depending on the configuration, the following layers may be added:
     /// - Retry layer: Adds retry logic to the object store.
     /// - Limit layer: Limits the number of concurrent requests to the object store.
+    /// - File cache layer: Adds a file cache to the object store.
     pub fn decorate_store<T: ObjectStore + Clone>(
         &self,
         store: T,
         table_root: &url::Url,
     ) -> DeltaResult<Box<dyn ObjectStore>> {
-        let inner = Self::decorate_prefix(store, table_root)?;
+        let inner = self.decorate_root_store(store)?;
+        let inner = Self::decorate_prefix(inner, table_root)?;
+        Ok(inner)
+    }
+
+    /// Wraps an object store with additional layers of functionality without prefix.
+    pub(crate) fn decorate_root_store<T: ObjectStore>(
+        &self,
+        store: T,
+    ) -> DeltaResult<Box<dyn ObjectStore>> {
+        let inner = if let Some(file_cache) = self.file_cache.as_ref() {
+            file_cache::decorate_store(Arc::new(store), file_cache)?
+        } else {
+            Box::new(store)
+        };
         Ok(inner)
     }
 
@@ -169,7 +190,7 @@ where
 
         #[cfg(feature = "cloud")]
         let remainder = {
-            let result = ParseResult::<RetryConfig>::from_iter(remainder);
+            let result = ParseResult::<object_store::RetryConfig>::from_iter(remainder);
             config.retry = result.config;
             result.unparsed
         };
@@ -224,6 +245,11 @@ impl StorageConfig {
             props.retry = retry;
             remainder
         };
+
+        let result = ParseResult::<FileCacheConfig>::from_iter(remainder);
+        result.raise_errors()?;
+        props.file_cache = (!result.is_default).then_some(result.config);
+        let remainder = result.unparsed;
 
         props.unknown_properties = remainder;
         Ok(props)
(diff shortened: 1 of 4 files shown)


ion-elgreco:
@roeap was actually going to add a caching object store based on `foyer` this weekend, which supports hybrid caching :)

roeap:
Just opened a draft PR - still need to wire things through, but the cache implementation should be close to finished.

@PeterKeDer - would love your feedback and happy to discuss which route we should take.

The most obvious difference would be the I used an external implementation of a cache. The policy is also more selective in that we right now only cache specific log files (the json ones) and omit the others. We do however try to make the cache sharable, such that it can be used for multiple table instances - i.e. we use a fully qualified URL as cache key.

To enable additional caching there have been various updates to delta kernel, which allow us to incrementally update a snapshot and scan state.

These would however only get activated once we have kernel for log handling ...

PeterKeDer:
@roeap looks cool! Glad you're already working on it 😄 

I'm curious about the decision to only cache json log files. What's fundamentally blocking your implementation from also caching checkpoints and data files? I'm not too familiar with the delta kernel so there may be a knowledge gap there.

From my experience, downloading the data files generally consumes a lot more time than updating the transaction log. Having a cache work for the data files too would be very valuable for our use cases.

roeap:
@PeterKeDer - there may be a few things to unpack before that choice makes sense 😆.

1. we have 2 phases to look at a) metadata b) data ... in the metadata phase we can be very opinionated about caching as we will always be reading files from the log. In the data phase, its "just abput reading the data files.

2. delta-rs more or less is a cache, since we keep all file actions for a snapshot in memory. in kernel we already added methods to take advantage of that an replay from a given state.

3. In a kernel log replay, we discard some information from the json files we may later need.

With that, the idea is to use the filesystem / byte level caching only for the json files, so we can re-read them to extract other actions. 

Kernel also introduces file format level abstractions - i.e. `read_parquet` - which allows us to also avoid additional decoding of parquet data by e.g. caching the data in feather format (basically flat arrow).

All of this mainly applies to the metadata pahse, in the data phase we then can use general purpose caching - here I hope that we can leverage mechanics build into datafusion ...

All that said, if we make the caching here configurable - just enable / disable - we can see that we get this PR merged, as all the above is obviously a multi-step and longer duration effort 😆.

ion-elgreco:
> @PeterKeDer - there may be a few things to unpack before that choice makes sense 😆.
> 
> 1. we have 2 phases to look at a) metadata b) data ... in the metadata phase we can be very opinionated about caching as we will always be reading files from the log. In the data phase, its "just abput reading the data files.
> 
> 2. delta-rs more or less is a cache, since we keep all file actions for a snapshot in memory. in kernel we already added methods to take advantage of that an replay from a given state.
> 
> 3. In a kernel log replay, we discard some information from the json files we may later need.
> 
> With that, the idea is to use the filesystem / byte level caching only for the json files, so we can re-read them to extract other actions. 
> 
> Kernel also introduces file format level abstractions - i.e. `read_parquet` - which allows us to also avoid additional decoding of parquet data by e.g. caching the data in feather format (basically flat arrow).
> 
> All of this mainly applies to the metadata pahse, in the data phase we then can use general purpose caching - here I hope that we can leverage mechanics build into datafusion ...
> 
> All that said, if we make the caching here configurable - just enable / disable - we can see that we get this PR merged, as all the above is obviously a multi-step and longer duration effort 😆. I would then maybe de-construct this over time?

Regarding Datafusion and parquet caching, there is another oss database that does this, with some clever caching on bytes level instead of file level

roeap:
> Regarding Datafusion and parquet caching, there is another oss database that does this, with some clever caching on bytes level instead of file level

Which one is that? When the time comes I wanted to do some research... I also think that `datafusion-comet` has some clever caching build in I wanted to look at.

ion-elgreco:
> > Regarding Datafusion and parquet caching, there is another oss database that does this, with some clever caching on bytes level instead of file level
> 
> Which one is that? When the time comes I wanted to do some research... I also think that `datafusion-comet` has some clever caching build in I wanted to look at.

This one from seafowl: https://github.com/splitgraph/seafowl/blob/main/src/object_store/cache.rs

PeterKeDer:
> @PeterKeDer - there may be a few things to unpack before that choice makes sense 😆.
> 
> 1. we have 2 phases to look at a) metadata b) data ... in the metadata phase we can be very opinionated about caching as we will always be reading files from the log. In the data phase, its "just abput reading the data files.
> 2. delta-rs more or less is a cache, since we keep all file actions for a snapshot in memory. in kernel we already added methods to take advantage of that an replay from a given state.
> 3. In a kernel log replay, we discard some information from the json files we may later need.
> 
> With that, the idea is to use the filesystem / byte level caching only for the json files, so we can re-read them to extract other actions.
> 
> Kernel also introduces file format level abstractions - i.e. `read_parquet` - which allows us to also avoid additional decoding of parquet data by e.g. caching the data in feather format (basically flat arrow).
> 
> All of this mainly applies to the metadata pahse, in the data phase we then can use general purpose caching - here I hope that we can leverage mechanics build into datafusion ...
> 
> All that said, if we make the caching here configurable - just enable / disable - we can see that we get this PR merged, as all the above is obviously a multi-step and longer duration effort 😆. I would then maybe de-construct this over time?

Makes sense, thanks for the explanation!

Yep, the caching here is entirely configurable, it is disabled by default but you can just pass in a storage option to enable it. If you think this PR is fine to merge, I'm happy to clean this up a bit and add some few test cases. We should also make clear of the potential downsides, i.e. slower initial reads if we rely on range requests, and expanding cache size as we load more files.

PeterKeDer:
@roeap @ion-elgreco I cleaned this up & added simple tests, appreciate if you can take a look and let me know your thoughts

roeap (commented):
Overall looking great @PeterKeDer. 

There is one thing that I hope we can maybe change is how we treat configuration and setup.

Specifically, since recently we can use `deltalake_derive::DeltaConfig` to derive extracting config from string maps, some details [here](https://github.com/delta-io/delta-rs/pull/3389). basic idea would be to include a new config, and add it to `StorageConfig`.

I also tried starting establishing a pattern for adding layers to object stores via `decorate_store` methods on the config.

IF we like this pattern, maybe we can update to use that apporach? I could then in a follow up update this config to alow allow, for the more minimal cache we discussed above?

PeterKeDer:
@roeap I took a shot at moving the config over to `DeltaConfig`, let me know if I'm using it correctly. Also, I noticed `decorate_store` is not being applied to the root store, but it probably should if we add functionality other than prefix. I added a separate function to handle that, let me know if that makes sense.

Separately, do you happen to know why this test is failing? e.g. [here](https://github.com/delta-io/delta-rs/actions/runs/15122822732/job/42552729250?pr=3439). It passes for me locally when running `cargo test --verbose --features azure,datafusion,s3,gcs,glue,hdfs`

PeterKeDer:
@roeap were you able to take a look at the test failures? I wonder if there are differences writing files when the test is running in the pipeline vs locally that causes the cache to not work as intended

PeterKeDer:
@roeap updated per your changes, let me know what you think

abhi-airspace-intelligence:
> > Regarding Datafusion and parquet caching, there is another oss database that does this, with some clever caching on bytes level instead of file level
> 
> Which one is that? When the time comes I wanted to do some research... I also think that `datafusion-comet` has some clever caching build in I wanted to look at.

Regarding this, in DF50 there's now a parquet file cache that can be instantiated. Since the parquet handling is done with the delta kernel and not Datafusion there's probably no way to directly benefit from it, but this epic probably has nice snippets for what it could look like. https://github.com/apache/datafusion/issues/17000

PR #3446, opened 2025-05-18
feat: expose kernel Engine on LogStore

roeap:
# Description

One last pre-factor before starting to use kernel on more actively in our codebase. In this PR we expose an `Engine` on `LogStore` as well as introduce a distinction between `prefixed_store` and `root_store` exposed on log store. This structure is far from ideal, but allowsd us to begin migration. As we progress there, we should see many opportunities to simplify the API surface again.

@ion-elgreco - there was a little bit more handlin needed on the lake fs integration, but I "think" this should work for now.

### DIFF
--- a/crates/core/src/logstore/default_logstore.rs
+++ b/crates/core/src/logstore/default_logstore.rs
@@ -24,7 +24,8 @@ fn put_options() -> &'static PutOptions {
 /// Default [`LogStore`] implementation
 #[derive(Debug, Clone)]
 pub struct DefaultLogStore {
-    pub(crate) storage: ObjectStoreRef,
+    prefixed_store: ObjectStoreRef,
+    root_store: ObjectStoreRef,
     config: LogStoreConfig,
 }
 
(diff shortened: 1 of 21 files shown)


crates/core/src/logstore/default_logstore.rs
```diff
@@ -33,10 +34,21 @@ impl DefaultLogStore {
     ///
     /// # Arguments
     ///
-    /// * `storage` - A shared reference to an [`object_store::ObjectStore`] with "/" pointing at delta table root (i.e. where `_delta_log` is located).
+    /// * `prefixed_store` - A shared reference to an [`object_store::ObjectStore`] with "/"
+    ///   pointing at delta table root (i.e. where `_delta_log` is located).
```
ion-elgreco, line 38: Docs should be updated as well

crates/core/tests/integration_datafusion.rs
```diff
@@ -1250,13 +1250,14 @@ mod local {
     #[tokio::test]
     async fn test_issue_2105() -> Result<()> {
         use datafusion::arrow::datatypes::{DataType as ArrowDataType, Field, Schema};
-        let path = tempfile::tempdir().unwrap();
-        let path = path.into_path();
+        let tmp_dir = tempfile::tempdir().unwrap();
+        let path = tmp_dir.path();
```
roeap, line 1254: this is a small drive by - it seems that `into_path` would persist things to disk and no longer clean up the storage once done.

crates/lakefs/src/logstore.rs
```diff
@@ -66,11 +69,19 @@ impl LakeFSLogStore {
     ///
     /// * `storage` - A shared reference to an [`object_store::ObjectStore`] with "/" pointing at delta table root (i.e. where `_delta_log` is located).
```
ion-elgreco, line 70: We should be consistent here as well and call it prefixed_store

crates/core/src/logstore/default_logstore.rs
roeap: Can you point to to the place in the docs that needs updating? Tried to find something, but found no direct reference to default log store

crates/lakefs/src/logstore.rs
roeap: absolutely!

crates/core/src/logstore/default_logstore.rs
roeap: or did you you mean the also incomplete method docs? 😆

ion-elgreco (approved):

PR #3447, opened 2025-05-18
feat: move some core path handling to use URLs

roeap:
# Description

Delta Kernel uses `Url`s as means of handling file paths. In delta-rs we so par have focussed on object stores `Path`s which also comes with limitiations.

In this PR (this time really the last one before checkpoints :D) we move a significant portion of the path handling to use Urls or at paths relative to the object store root. This also allows us to leverage url parsing into typed log files from kernel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
-delta_kernel = { version = "0.10.0", features = [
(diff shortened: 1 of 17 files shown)


rtyler:
@roeap does this need to be closed or rebased after the LogStore change landed?

roeap:
@rtyler - I'll just re-do it ... rebasing would likely be too painful 😆

PR #3451, opened 2025-05-20
fix: handle edge in schema adapter for single missing field

alexwilcoxson-rel:
will create upstream ticket for fix and tests

### DIFF
--- a/crates/core/src/operations/cast/mod.rs
+++ b/crates/core/src/operations/cast/mod.rs
@@ -178,11 +178,17 @@ pub fn cast_record_batch(
         ..Default::default()
     };
 
-    let s = StructArray::new(
-        batch.schema().as_ref().to_owned().fields,
-        batch.columns().to_owned(),
-        None,
-    );
+    let col_arrays = batch.columns().to_owned();
+    let s = if col_arrays.is_empty() {
+        StructArray::new_empty_fields(batch.num_rows(), None)
+    } else {
+        StructArray::new(
+            batch.schema().as_ref().to_owned().fields,
+            col_arrays,
+            None,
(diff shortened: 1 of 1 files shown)


alexwilcoxson-rel:
> @alexwilcoxson-rel, do you have a link to the upstream issue? I recently created a more fine granular schema adapter, that may also allow pushing down leaf column selections.
> 
> Was thinking about upstreaming that and would refer to that issue.

sorry @roeap I originally intended to open this PR against the fork in our GitHub org and delta-io/delta-rs _is_ the upstream in that case :smile:

I will open a PR to main in this repo with a repro test case here shortly.

alexwilcoxson-rel:
@roeap I haven't gotten a PR up yet as I'm having an issue getting some python reproduce tests written. I'm going to be off for a couple days but wanted to provide you some more info if relevant to your ongoing work.

The existing schema adapter's main feature that we rely on is after a nullable column is added to a delta table, still being able to scan existing files in that table that don't physically have the column. This works via the add_missing parameter to `cast_struct`.

The issue we ran into though is when the mapper is given a batch and the projected schema only contains the new nullable field. The incoming batch is an empty schema but with N rows. The existing code result in a batch with the correct schema but length 0. The below snippet from the change fixes this to create an empty struct array with N rows as input to `cast_struct`

```rust
    let col_arrays = batch.columns().to_owned();
    let s = if col_arrays.is_empty() {
        StructArray::new_empty_fields(batch.num_rows(), None)
    } else {
        StructArray::new(
            batch.schema().as_ref().to_owned().fields,
            col_arrays,
            None,
        )
    };
```

PR #3453, opened 2025-05-21
chore: ensuring default builds work without datafusion

rtyler:
turns out our CI shouldn't have let some things through since it wasn't running a default build of tests too


roeap (approved):
👍

Issue #865, opened 2022-10-05
Delta Log entities with absolute path not supported


kosinsky:
# Environment

**Delta-rs version**:

**Binding**:  Python (however, I think it's true for all versions

**Environment**:
- **Cloud provider**: Azure, however the same with local filesystems
- **OS**:
- **Other**:

***
# Bug

**What happened**:
I'm trying to read Delta table where absolute path is used for `add` and `remove` actions. If path looks like `abfss://<container>@<account_name>.blob.core.windows.net/full/part-00000-a72b1fb3-f2df-41fe-a8f0-e65b746382dd-c000.snappy.parquet` or any other absolute file (including local filesystem) `file_uris` method concats it with table uri.

**What you expected to happen**:
Just use provided absolute uri. https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file says
> A relative path to a file from the root of the table or **an absolute path to a file** that should be added to the table. 

**How to reproduce it**:
Delta Log entry looks like:
```JSON
{"commitInfo":{"timestamp":1587968586154,"operation":"WRITE","operationParameters":{"mode":"ErrorIfExists","partitionBy":"[]"},"isBlindAppend":true}}
{"protocol":{"minReaderVersion":1,"minWriterVersion":2}}
{"metaData":{"id":"5fba94ed-9794-4965-ba6e-6ee3c0d22af9","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"id\",\"type\":\"long\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{},"createdTime":1587968585495}}
{"add":{"path":"abfss://<container>@<account_name>.blob.core.windows.net/full/part-00000-a72b1fb3-f2df-41fe-a8f0-e65b746382dd-c000.snappy.parquet","partitionValues":{},"size":262,"modificationTime":1587968586000,"dataChange":true}}
```
Python code:
```python
import os
from deltalake import DeltaTable

storage_options={'AZURE_STORAGE_ACCOUNT_KEY': <access key>}
table_path = "abfss://<container>@<account_name>.blob.core.windows.net/full"
dt = DeltaTable(table_path,  storage_options=storage_options)
print(dt.file_uris())
print(dt.to_pyarrow_dataset().to_table().to_pydict())
```
I'm getting following output:
```
[...]

PR #3456, opened 2025-05-22
refactor: use full paths in log processing

roeap:
# Description

Delta kernel handles all paths as fully qualified URLs. As another step towards adopting kernel, we move the internal log processing to handle full paths, rather than paths relative to the table root.

### DIFF
--- a/crates/core/src/kernel/snapshot/log_segment.rs
+++ b/crates/core/src/kernel/snapshot/log_segment.rs
@@ -18,7 +18,7 @@ use tracing::debug;
 use super::parse;
 use crate::kernel::transaction::CommitData;
 use crate::kernel::{arrow::json, ActionType, Metadata, Protocol, Schema, StructType};
-use crate::logstore::LogStore;
+use crate::logstore::{LogStore, LogStoreExt};
 use crate::{DeltaResult, DeltaTableConfig, DeltaTableError};
 
 const LAST_CHECKPOINT_FILE_NAME: &str = "_last_checkpoint";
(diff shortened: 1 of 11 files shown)


rtyler:
Will this finally resolve https://github.com/delta-io/delta-rs/issues/865 ?

crates/core/src/kernel/snapshot/log_segment.rs
```diff
@@ -110,18 +110,27 @@ impl LogSegment {
...
+        let root_url = log_store.table_root_url();
+
+        let log_url = root_url
+            .join("_delta_log")
+            .map_err(|e| DeltaTableError::Generic(e.to_string()))?;
+        let log_path = if log_url.scheme() == "s3" {
+            Path::parse(log_url.path())?
+        } else {
+            Path::from_url_path(log_url.path())?
+        };
```
roeap, line 123: @rtyler - i tried to find a reason why we need to commit such atrocities. This is the second time we need to special-case path parsing for S3. Somewhere we must be doing something different we are just reverting in these cases. Any idea what that could be?

ion-elgreco: What are the outputs @roeap if you don't special case it vs you do special case it

roeap: One is url-encoded, the other is not.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -58,9 +59,8 @@ pub struct Snapshot {
     protocol: Protocol,
     metadata: Metadata,
     schema: StructType,
-    // TODO make this an URL
     /// path of the table root within the object store
```
rtyler, line 62: This comment may be out of date, since it refers to the path **within** the object store. I assume `table_url` will now capture the full URL to the table?

crates/core/src/logstore/mod.rs
```diff
@@ -338,6 +338,14 @@ pub trait LogStore: Send + Sync + AsAny {
         self.to_uri(&Path::from(""))
     }
 
+    fn table_root_url(&self) -> Url {
```
rtyler, line 341: I'm still digesting this whole pull request, but please add some rustdoc for why this function exists. Bonus points in adding a unit test :kissing_smiling_eyes:

crates/core/src/kernel/snapshot/mod.rs
roeap: updated!

ion-elgreco:
@roeap is this ready?

roeap:
@ion-elgreco - now it should be, pending any additional feedback of course :).

PR #3461, opened 2025-05-24
fix: use more accurate log path parsing

roeap:
# Description

stacked on: #3456

With more features we see more and more different file types in the delta log. Our parsing logic to identify specific types files may not have been fine granular enough. In this PR we use the log file parsing from delta kernel to handle more recent addition to the spec.


ion-elgreco (approved):
LGTM!

PR #3464, opened 2025-05-24
feat: add a table description and name to the Delta Table from Python

### LINKED ISSUE #3401: Alter table description
# Description
I use the Python lib to define the metadata of my table columns as following:
`table.alter.set_column_metadata("id", {"comment": "the unique id"})` 

But I don't find how to define the description/comment of the table itself.  
I tried 
`table.alter.set_table_properties({"description": "my wonderful table"})`
or
`table.alter.set_table_properties({"comment": "my wonderful table"})`
then it raised 
`DeltaError: Kernel: Generic delta kernel error: Error parsing property 'comment':'my wonderful table'`
It comes from the property value must be in `TableProperty` enum.  

It could be great to add a function to define description table as 
`table.alter.set_table_description("my wonderful table")`
**Use Case**

**Related Issue(s)**

fvaleye:
Hey 👋,

Let's start with a simple one.

# Description
Adds convenient methods to set the table description and name through the Python API:
- Add `table.alter.set_table_description(desc) `method in Python
- Add `table.alter.set_table_name(name) `method  in Python
- Properly handles metadata storage in the Delta table metadata

## Related issue
- closes #3401 

## Implementation Details
- Leverage the `validator` crate
- Limit the number of characters for update operation to a Delta Table: name (255 characters, not empty) and description (4000 characters)
- Added Python bindings in lib.rs for `set_table_name` and `set_table_description`
- Extended TableAlterer class with convenient API methods

## Use Case
Previously, users had to use `set_table_properties` with specific property keys and faced issues with property validation. Now they can simply call:
```
  table.alter.set_table_description('my wonderful table')
  table.alter.set_table_name('my_table')
```

This resolves the `DeltaError` about property values needing to be in TableProperty enum by directly handling description/name metadata rather than as generic properties.

This update fixes the table description setting, which previously required manual property management, and provides a clean, intuitive API for standard table metadata operations.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -94,6 +94,7 @@ rand = "0.8"
 maplit = "1"
 sqlparser = { version = "0.56.0" }
 humantime = { version = "2.1.0" }
+validator = { version = "0.19", features = ["derive"] }
 
 [dev-dependencies]
 criterion = "0.5"
--- a/crates/core/src/operations/update_table_metadata.rs
+++ b/crates/core/src/operations/update_table_metadata.rs
@@ -3,6 +3,7 @@
 use std::sync::Arc;
 
 use futures::future::BoxFuture;
+use validator::Validate;
 
 use super::{CustomExecuteHandler, Operation};
 use crate::kernel::transaction::{CommitBuilder, CommitProperties};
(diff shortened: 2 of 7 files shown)


ion-elgreco:
👋 @fvaleye! This looks useful for sure, but I think it makes more sense to have a new operation that updates a subset of the table_metadata and maybe takes an Enum::TableName/TableDescription as input instead of going through the set_table_properties

ion-elgreco:
@fvaleye great work! This looks ready to me or are you still adding things?

fvaleye:
> @fvaleye great work! This looks ready to me or are you still adding things?

I included your suggestion, LGTM!

ion-elgreco:
@fvaleye Maybe we can use https://github.com/Keats/validator to validate the enum properties directly? Then you don't have to check downstream again in case users directly construct the enum

fvaleye:
> @fvaleye Maybe we can use https://github.com/Keats/validator to validate the enum properties directly? Then you don't have to check downstream again in case users directly construct the enum

Thanks for the idea, loved it.
Since it's a new dependency, I would rather use it everywhere by letting contributors agree on its usage, WDYT?

ion-elgreco:
> > @fvaleye Maybe we can use https://github.com/Keats/validator to validate the enum properties directly? Then you don't have to check downstream again in case users directly construct the enum
> 
> Thanks for the idea, loved it. Since it's a new dependency, I would rather use it everywhere by letting contributors agree on its usage, WDYT?

Yes, let's see what others have to say, @roeap, @rtyler any thoughts on this?

ion-elgreco:
There is also https://docs.rs/garde/latest/garde/ btw, which is a drop-in replacement/rewrite of validator, has some more functions. Both seem well maintained in terms of updates

roeap:
generally speaking I am a great fan of validating data and failing early if we encounter invalid data.

I have experience with neither crate, but it seems `validator` has a smaller size footprint and its dependencies are already covered by the existing ones.

fvaleye:
Let's add `validator` then! 👍
I will include the lib and ping you when it's done @ion-elgreco

fvaleye:
I added the `validator` crate and some validation rules.
Ready for review @ion-elgreco

ion-elgreco (approved):
Great stuff!

fvaleye:
FYI: I didn't add more Python tests in lib.rs because it is already tested (not covered by the coverage tool).

PR #3468, opened 2025-05-25
fix: schema conversion, add conversion test cases

ion-elgreco:
# Description
With the refactor I introduced a minor bug, but it's only fixable with an upstream change in arro3, https://github.com/kylebarron/arro3/pull/334.  cc @kylebarron 

I already have the new code stashed locally, so once `arro3` is released I will push those changes and it should work ✨ 

The conversion test cases will highlight the issue for lists that lose the name, nullability and metadata ^^


rtyler (approved):

python/tests/test_conversion.py
```diff
@@ -0,0 +1,387 @@
...
+        ),
+        (
+            Schema(fields=[Field("foo", DataType.int32())]),
+            Schema(fields=[Field("foo", DataType.int32())]),
+        ),
+        (
+            Schema(fields=[Field("foo", DataType.int16())]),
+            Schema(fields=[Field("foo", DataType.int16())]),
+        ),
+        # Unsigned integers to signed
```
roeap, line 23: casting uint to int can overflow, did we have a `unsafe` parameter o.a.? If not, is this something for us to add in 1.0?

roeap (approved):
just some general comments we may want to look at for 1.0.

python/tests/test_conversion.py
ion-elgreco: It should error, but I see I inverted the boolean, so let me put in a fix for that!

python/src/lib.rs
```diff
@@ -1445,7 +1445,7 @@ impl RawDeltaTable {
         Ok(())
     }
 
-    pub fn get_add_actions(&self, flatten: bool) -> PyResult<PyRecordBatch> {
+    pub fn get_add_actions(&self, flatten: bool) -> PyResult<Arro3RecordBatch> {
```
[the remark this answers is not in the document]
ion-elgreco: But we haven't for all the previous releases, wouldn't that be strange to mark it experimental now. I think we can still deprecate changes beyond 1.0.0

PR #3474, opened 2025-05-26
[RFC] feat!: kernel based log replay - take 2

roeap:
# Description

This is a redo of #3137.

Since the first attempt, kernel has evolved quite a bit, as has our codebase. We also learned a lot, particularly around some of the complexities of "nested" async evaluation that comes with kernels engine concepts.

A few things that could guide designs: 

* Use `Engine`: Anything we can do using kernels `Engine` abstractions, we can do without the need to have datafusion enabled. Potentially significantly extending what we can offer to arrow-only users.
* Expose idiomatic `async`: kernel exposes blocking iterators that only "appear to be sync". We need to figure out how to best handle these. Kernel may one day also expose apis or utilities to more seamlessly integrate with rust `async` so we should hide these complexities from end users.

Since this will (and should) have a significant impact on large parts of the codebase, we should not even try to do a once-off switch. Rather we propose the following migration strategy.

1. Define a new trait `Snapshot` that exposes a minimal API we require in operations planning etc.
2. Implement this trait for the current snapshot implementations as well as the newly introduced kernel-backed snapshots.
3. Refactor individual operations / code-paths to accept this new trait

This should allow us to incrementally "kernelize" our codebase and give us the opportunity to properly test the new snapshots. Opt-in to using new snapshots should for now be done via either feature flags or (maybe even better) runtime configuration.

Still figuring out some basics, but putting it up anyway for feedback.

## Major breaking changes

A lot of APIs will be broken eventually, but listing some of the more fundamental changes.

### No more tombstones

Currently we expose tombstones as part of our state / snapshots. Since Checkpoint writing uses kernel now, we really only need that to plan vacuums, which is a "special" or at least table format specific maintenance operation. As such, we internalise the logic how a vacuum is planned.

### DIFF
--- a/crates/core/src/kernel/snapshot_next/arrow_ext.rs
+++ b/crates/core/src/kernel/snapshot_next/arrow_ext.rs
@@ -1,0 +1,141 @@
+//! Utilities for interacting with Kernel APIs using Arrow data structures.
+//!
+use delta_kernel::arrow::array::BooleanArray;
+use delta_kernel::arrow::compute::filter_record_batch;
+use delta_kernel::arrow::record_batch::RecordBatch;
+use delta_kernel::engine::arrow_data::ArrowEngineData;
+use delta_kernel::scan::{Scan, ScanMetadata};
+use delta_kernel::{
+    DeltaResult, Engine, EngineData, ExpressionEvaluator, ExpressionRef, PredicateRef, Version,
+};
+use itertools::Itertools;
+
+/// [`ScanMetadata`] contains (1) a [`RecordBatch`] specifying data files to be scanned
+/// and (2) a vector of transforms (one transform per scan file) that must be applied to the data read
+/// from those files.
+pub struct ScanMetadataArrow {
+    /// Record batch with one row per file to scan
+    pub scan_files: RecordBatch,
+
+    /// Row-level transformations to apply to data read from files.
+    ///
+    /// Each entry in this vector corresponds to a row in the `scan_files` data. The entry is an
+    /// expression that must be applied to convert the file's data into the logical schema
+    /// expected by the scan:
+    ///
+    /// - `Some(expr)`: Apply this expression to transform the data to match [`Scan::schema()`].
+    /// - `None`: No transformation is needed; the data is already in the correct logical form.
+    ///
+    /// Note: This vector can be indexed by row number.
+    pub scan_file_transforms: Vec<Option<ExpressionRef>>,
+}
+
+pub trait ScanExt {
+    /// Get the metadata for a table scan.
+    ///
+    /// This method handles translation between `EngineData` and `RecordBatch`
+    /// and will already apply any selection vectors to the data.
+    /// See [`Scan::scan_metadata`] for details.
+    fn scan_metadata_arrow(
+        &self,
+        engine: &dyn Engine,
+    ) -> DeltaResult<impl Iterator<Item = DeltaResult<ScanMetadataArrow>>>;
+
+    fn scan_metadata_from_arrow(
+        &self,
+        engine: &dyn Engine,
+        existing_version: Version,
+        existing_data: Box<dyn Iterator<Item = RecordBatch>>,
+        existing_predicate: Option<PredicateRef>,
+    ) -> DeltaResult<impl Iterator<Item = DeltaResult<ScanMetadataArrow>>>;
+}
+
+impl ScanExt for Scan {
+    fn scan_metadata_arrow(
+        &self,
+        engine: &dyn Engine,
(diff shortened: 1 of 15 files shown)


crates/core/src/kernel/snapshot_next/arrow_ext.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Row-level transformations to apply to data read from files.
+    ///
+    /// Each entry in this vector corresponds to a row in the `scan_files` data. The entry is an
+    /// expression that must be applied to convert the file's data into the logical schema
+    /// expected by the scan:
+    ///
+    /// - `Some(expr)`: Apply this expression to transform the data to match [`Scan::schema()`].
+    /// - `None`: No transformation is needed; the data is already in the correct logical form.
+    ///
+    /// Note: This vector can be indexed by row number.
```
ion-elgreco, line 27: Is the order of this vec guaranteed?

crates/core/src/kernel/snapshot_next/arrow_ext.rs
```diff
@@ -0,0 +1,134 @@
...
+    let batch = ArrowEngineData::try_from_engine_data(metadata.scan_files.data)?.into();
+    let scan_files = filter_record_batch(
+        &batch,
+        &BooleanArray::from(metadata.scan_files.selection_vector),
+    )?;
+    Ok(ScanMetadataArrow {
+        scan_files,
+        scan_file_transforms,
+    })
+}
```
ion-elgreco, line 89: This kernel to arrow seems to filter to active add actions, right?

crates/core/src/kernel/snapshot_next/eager.rs
```diff
@@ -0,0 +1,135 @@
...
+            .scan_metadata_from_arrow(
+                engine.as_ref(),
+                current,
+                Box::new(std::iter::once(self.files.clone())),
+                self.predicate.clone(),
+            )?
+            .map_ok(|s| s.scan_files)
+            .try_collect()?;
+
+        self.files = concat_batches(&files[0].schema(), &files)?;
```
ion-elgreco, line 130: Does scan_builder re-use the existing state it has?

crates/core/src/kernel/snapshot_next/iterators.rs
```diff
@@ -0,0 +1,67 @@
...
+impl LogicalFileView {
+    /// Path of the file.
+    pub fn path(&self) -> &str {
+        self.files.column(0).as_string::<i32>().value(self.index)
+    }
+
+    /// Size of the file in bytes.
+    pub fn size(&self) -> i64 {
+        self.files
+            .column(1)
```
ion-elgreco, line 25: My preference would be to do `column_by_name`, I find column ordering more risky then incorrect naming : P

crates/core/src/kernel/snapshot_next/eager.rs
roeap: `scan_metadata_from` does exactly that. It will treat the existing data as if it were a snapshot as of the current version. Depending on what it finds internally it will re-use that data. (if there is a new snapshot, it will currently do a new log repay, but this can be improved in the future.

crates/core/src/kernel/snapshot_next/arrow_ext.rs
roeap: Exactly.

ion-elgreco: Small note would help, I had to look up the delta-kernel-rs docs to understand it ^^

crates/core/src/kernel/snapshot_next/lazy.rs
```diff
@@ -0,0 +1,335 @@
...
+            Err(err) => {
+                return Box::pin(futures::stream::once(async {
+                    Err(DeltaTableError::KernelError(err))
+                }))
+            }
+        };
+
+        // TODO: which capacity to choose?
+        let mut builder = RecordBatchReceiverStreamBuilder::new(100);
+        let tx = builder.tx();
```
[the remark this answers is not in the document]
rtyler: Would a stream by any other name be as fast? Streams are basically channels which are basically async Iterators., Functionally it is something which has `poll_next` on it, channels are typically what we call something which allows cross task/thread streams, which is usually why these things have to be bounded.

My recommendation would be to define an environment value and default, e.g. 

```suggestion
        let mut builder = RecordBatchReceiverStreamBuilder::new(std::env::var("RECORD_BATCH_STREAM").unwrap_or(1024));
        let tx = builder.tx();
```

PR #3479, opened 2025-05-28
chore: update migration docs

ion-elgreco:


python/deltalake/table.py
```diff
@@ -1072,13 +1090,20 @@ def repair(
         return deserialized_metrics
 
     def transaction_versions(self) -> dict[str, Transaction]:
+        """
+        Retrieve the latest transaction versions for each application ID.
+
+        Returns:
+            dict[str, Transaction]: A mapping from application ID (str) to the corresponding Transaction object,
+            representing the most recent transaction for each app.
+        """
```
roeap, line 1099: This is one method I'd like to break ... i.e. do `transaction_version(self. app_id: str) -> int`. This is how spark etc handle transaction versions and also fits the use case for the - idempotent writes - which really only make sense for a single app.

I think I can make this change today.

docs/upgrade-guides/guide-1.0.0.md
```diff
@@ -2,8 +2,72 @@
...
 
+In addition, DeltaLake no longer has a hard dependency on PyArrow, making it more lightweight and suitable for use in minimal or size-constrained environments. Arrow data handling between Python and Rust is now powered by `arro3`, a lightweight Arrow implementation backed by `arrow-rs`.
+
+For users who still require PyArrow-based read functionality, it remains available as an optional dependency. To continue using these features, install DeltaLake with the `pyarrow` extra, `deltalake[pyarrow]`.
+
 ## Breaking changes
 
+### `arro3` adoption
+
+DeltaLake 1.0.0 introduces support for `arro3`. As a result, the `write_deltalake` and `DeltaTable.merge` functions now only accept inputs that implement one of the following interfaces:
```
roeap, line 13: I think what we introduce is support for the `PyCapsule` protocol and we use the minimal implementation of that protocol `arrow3`?

python/deltalake/table.py
ion-elgreco: Yeah makes sense!

docs/upgrade-guides/guide-1.0.0.md
ion-elgreco: Yeah bad wording, let me fix that

roeap (approved):

PR #3484, opened 2025-05-28
refactor!: get transaction versions for specific applications

roeap:
# Description

Update how we expose transaction versions. Prior we would return a hash map of all transaction, now we allow getting the version for a single transaction. This aligns more with the use case for transactions - idempotent writes - and makes the kernel integration easier.


ion-elgreco (approved):
Awesome!

PR #3488, opened 2025-05-28
refactor!: remove and deprecate some python methods

roeap:
# Description

As discussed in slack channel ...


ion-elgreco (dismissed):
Looks good, just needs a fmt though

PR #3497, opened 2025-05-29
ci: improve coverage collection

roeap:
# Description

Trying to improve both the test coverage report and the test runtime.

* collect coverage in:
  * integration-tests
  * integration-tests-lakefs
  
* run python benchmark only on main
* separate build checks from integration tests
* run native tls tests in separate job


rtyler (approved):
let's give it a whirl

PR #3501, opened 2025-05-30
fix: make DeltaTable pickleable

HiromuHota:
# Description
This PR adds `__reduce__` method to DeltaTable to make an object of this class pickleable.

# Related Issue(s)
This will fix https://github.com/dask-contrib/dask-deltatable/issues/87 and potentially other distributed system that have to serialize a DeltaTable object and send it over to workers.

# Documentation
N/A

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -158,12 +158,29 @@ class DeltaTable:
 
         """
         self._storage_options = storage_options
+        self._without_files = without_files
+        self._log_buffer_size = log_buffer_size
         self._table = RawDeltaTable(
             str(table_uri),
             version=version,
             storage_options=storage_options,
             without_files=without_files,
             log_buffer_size=log_buffer_size,
+        )
+
+    def __reduce__(self) -> tuple:
+        """
+        This allows DeltaTable to be pickled.
+        """
+        return (
+            self.__class__,
+            (
+                self.table_uri,
+                self.version(),
+                self._storage_options,
+                self._without_files,
+                self._log_buffer_size,
+            ),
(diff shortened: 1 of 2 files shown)


python/deltalake/table.py
```diff
@@ -158,6 +158,8 @@ def __init__(
 
         """
         self._storage_options = storage_options
```
jacobtomlinson, line 160:
```suggestion
        self._storage_options = storage_options
        self._pickle_storage_options = storage_options is not None
```

python/deltalake/table.py
```diff
@@ -166,6 +168,21 @@ def __init__(
...
+    def __reduce__(self) -> tuple[type, tuple[Any, ...]]:
+        """
+        This allows DeltaTable to be pickled.
+        """
+        return (
+            self.__class__,
+            (
+                self.table_uri,
+                self.version(),
+                self._storage_options,
```
jacobtomlinson, line 180:
```suggestion
                self._storage_options if self._pickle_storage_options else None,
```

jacobtomlinson (commented):
Perhaps using a flag to track if storage options were explicitly passed, and only pickling them if they were would be a more security conscious implementation here?

It avoids credentials being unexpectedly dumped from the environment, but allows credentials to be portable if users pass them explicitly.

ion-elgreco:
I suggest you create a wrapper (https://github.com/dask-contrib/dask-deltatable/issues/87#issuecomment-2938951961) after https://github.com/delta-io/delta-rs/pull/3518 lands :)

PR #3503, opened 2025-05-30
fix: constraint parsing, roundtripping

ion-elgreco:
# Description
This reverts https://github.com/delta-io/delta-rs/pull/3374

The main issue was in the Display for SqlFormat, so when you would add constraints in delta-rs with backticks or quotes for columns with whitespaces, it wouldn't properly persist when it hits the metadata. The default display for `Expr::col` is to use the flat name, but we should always use `quoted_flat_name`

### DIFF
--- a/crates/core/src/delta_datafusion/expr.rs
+++ b/crates/core/src/delta_datafusion/expr.rs
@@ -333,7 +333,10 @@ impl Display for BinaryExprFormat<'_> {
 impl Display for SqlFormat<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.expr {
-            Expr::Column(c) => write!(f, "{c}"),
+            Expr::Column(c) => {
+                let c = c.quoted_flat_name();
+                write!(f, "{c}")
+            }
             Expr::Literal(v) => write!(f, "{}", ScalarValueFormat { scalar: v }),
             Expr::Case(case) => {
                 write!(f, "CASE ")?;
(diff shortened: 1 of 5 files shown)


crates/core/src/delta_datafusion/expr.rs
```diff
@@ -333,7 +333,10 @@ impl Display for BinaryExprFormat<'_> {
 impl Display for SqlFormat<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.expr {
-            Expr::Column(c) => write!(f, "{c}"),
+            Expr::Column(c) => {
+                let c = c.quoted_flat_name();
+                write!(f, "{c}")
+            }
```
roeap, line 339:
```suggestion
            Expr::Column(c) => write!(f, "{}", c.quoted_flat_name()),
```

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -1591,16 +1591,15 @@ impl DeltaDataChecker {
...
         self.ctx.register_table(&table_name, Arc::new(table))?;
 
         let mut violations: Vec<String> = Vec::new();
 
         for check in checks {
-            let check_name = check.get_name();
-            if check_name.contains('.') {
+            if check.get_name().contains('.') {
                 return Err(DeltaTableError::Generic(
                     "Support for nested columns is not supported.".to_string(),
```
roeap, line 1604: While not part of the PR "Support [...] is not supported" does sound somewhat strange :).

roeap (commented):
Just some minor comment to maybe look at really quick?

PR #3504, opened 2025-05-31
fix: creating new DeltaTable with invalid table name path no longer creates empty directory

### LINKED ISSUE #3060: Opening an incorrect table path causes an empty directory to be made
# Environment

**Delta-rs version**: 0.22.3

**Binding**:  python

**Environment**:
- **Cloud provider**: :upside_down_face: 
- **OS**: Linux/amd64
- **Other**:

***
# Bug

**What happened**:

```
>>> dt = DeltaTable('i_just_made_this_up_lol')
[2024-12-16T14:15:26Z DEBUG deltalake_core::table::builder] creating table builder with file:///home/tyler/source/github/delta-io/delta-rs/python/i_just_made_this_up_lol
[2024-12-16T14:15:26Z DEBUG deltalake_core::table::builder] build_storage() with file:///home/tyler/source/github/delta-io/delta-rs/python/i_just_made_this_up_lol
[2024-12-16T14:15:26Z DEBUG deltalake_core::table::builder] Loading a logstore based off the location: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/home/tyler/source/github/delta-io/delta-rs/python/i_just_made_this_up_lol", query: None, fragment: None }
[2024-12-16T14:15:26Z DEBUG deltalake_core::logstore] Found a storage provider for file:/// (file:///home/tyler/source/github/delta-io/delta-rs/python/i_just_made_this_up_lol)
[2024-12-16T14:15:26Z DEBUG deltalake_core::logstore] Found a logstore provider for file:///
  File "/home/tyler/source/github/delta-io/delta-rs/python/deltalake/table.py", line 415, in __init__
    self._table = RawDeltaTable(
_internal.TableNotFoundError: no log files
>>>
(venv) plantain% ls i_just_made_this_up_lol
(venv) plantain%
```

No directory should be made :smile: 

**How to reproduce it**:

**More details**:

smeyerre:
# Description
Previously when using the Python bindings, creating a DeltaTable with an incorrect table path caused an empty directory to be made. Now no directory is created and an appropriate error message is served:
```
>>> dt = DeltaTable('nonexistent_table')
  File "<python-input-1>", line 1, in <module>
    dt = DeltaTable('nonexistent_table')
  File "/home/sam/github/delta-rs/python/deltalake/table.py", line 174, in __init__
    self._table = RawDeltaTable(
                  ~~~~~~~~~~~~~^
        str(table_uri),
        ^^^^^^^^^^^^^^^
    ...<3 lines>...
        log_buffer_size=log_buffer_size,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
_internal.TableNotFoundError: Local path "nonexistent_table" does not exist or you don't have access!
```

# Related Issue(s)

- closes #3060


smeyerre:
This is likely out of scope for this PR, but I'm seeing other places (eg. in `is_deltatable`) where we're currently using `from_url` where it would probably make more sense to use `from_valid_url` as well. In fact I'm curious, when would we want to ever use `from_url` instead?

roeap:
I think we do need to be a little bit careful here. Valid url not only means that the directory exists, but we also ensure e.g. that a path ends in '/' which is quite important since e.g. `Url::join` behaves quite differently if it does not.

We also need to validate how the create command behaves in this case. We still use a prefixed store in many places and while object store will create any directory within its scope, it might not create any parent directory that is part of the prefix. Since we can no longer guarantee that a directory exists we now need to ensure that at the very least in the `CreateOperation` and likely also in some other places (maybe convert_to_delta)?

All that said, I do agree that we should not just eagerly create the directory, before we know its what the user wants.

smeyerre:
> I think we do need to be a little bit careful here. Valid url not only means that the directory exists, but we also ensure e.g. that a path ends in '/' which is quite important since e.g. `Url::join` behaves quite differently if it does not.
> 
> We also need to validate how the create command behaves in this case. We still use a prefixed store in many places and while object store will create any directory within its scope, it might not create any parent directory that is part of the prefix. Since we can no longer guarantee that a directory exists we now need to ensure that at the very least in the `CreateOperation` and likely also in some other places (maybe convert_to_delta)?
> 
> All that said, I do agree that we should not just eagerly create the directory, before we know its what the user wants.

Just want to make sure I'm understanding fully since I'm not super experienced with DeltaTable. You're saying we want to be extra sure that things like `CreateBuilder`, `ConvertToDeltaBuilder`, etc still create the directory properly? And that table paths are formatted with trailing slashes after passing through `from_valid_url`?

Could you give me an example of the expected behavior you're describing with object store directory creation as well?

Thanks!

rtyler (approved):

Issue #3507, opened 2025-06-01
bug: timezone not preserved & raise exc on merge operation


Itayazolay:
# Environment

**Delta-rs version**: 1.0.1

**Binding**: python

**Environment**:
- **Cloud provider**: local
- **OS**: mac os
- **Other**:

***
# Bug

**What happened**:
timezone not preserved and crash on merge (upsert)
**What you expected to happen**:
keep timezone and work

**How to reproduce it**:
```python
df = pd.DataFrame({'a': 1, 'ts': pd.date_range('2021-01-01', '2021-01-02', freq='h', tz='America/Chicago')})
path = '/tmp/delta_tmp/'
write_deltalake(path, df, mode='overwrite')
dt = DeltaTable(path)
print(dt.to_pandas().head(3))
df2=pd.DataFrame({'a': 2, 'ts': pd.date_range('2021-01-01', '2021-01-03', freq='h', tz='America/Chicago')})
dt.merge(df2, predicate='source.ts = target.ts', source_alias='source', target_alias='target').when_matched_update_all().when_not_matched_insert_all().execute()
```
Output: 
```
   a                        ts
0  1 2021-01-01 06:00:00+00:00
1  1 2021-01-01 07:00:00+00:00
2  1 2021-01-01 08:00:00+00:00
Traceback (most recent call last):
  File "/Users/itay/miniconda3/envs/jether_dev/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3550, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-56-3be7592b3257>", line 7, in <module>
    dt.merge(df2, predicate='source.ts = target.ts', source_alias='source', target_alias='target').when_matched_update_all().when_not_matched_insert_all().execute()
  File "/Users/itay/miniconda3/envs/jether_dev/lib/python3.10/site-packages/deltalake/table.py", line 1618, in execute
    metrics = self._table.merge_execute(self._builder)
_internal.DeltaError: Generic DeltaTable error: Arrow error: Invalid argument error: column types must match schema types, expected Timestamp(Nanosecond, Some("America/Chicago")) but found Timestamp(Microsecond, Some("UTC")) at column index 1
```

**More details**:

roeap:
This is interesting, an I am not entirely sure what the cleanest way is to get this to work.

[...]

PR #3508, opened 2025-06-01
fix: setting wrong schema in table provider for `merge`

ion-elgreco:
# Description
The pre-casted schema was set in the table provider, but should be post casting. Als updating py for quick python release.

- closes https://github.com/delta-io/delta-rs/issues/3507

### DIFF
--- a/python/tests/test_conversion.py
+++ b/python/tests/test_conversion.py
@@ -1,6 +1,7 @@
 import pytest
 from arro3.core import DataType, Field, Schema
 
+from deltalake import DeltaTable, write_deltalake
 from deltalake.writer._conversion import _convert_arro3_schema_to_delta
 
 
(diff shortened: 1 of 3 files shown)


python/tests/test_conversion.py
```diff
@@ -385,3 +386,35 @@
...
+        }
+    )
+
+    dt = DeltaTable(tmp_path)
+    dt.merge(
+        df2,
+        predicate="source.ts = target.ts",
+        source_alias="source",
+        target_alias="target",
+    ).when_matched_update_all().when_not_matched_insert_all().execute()
```
roeap, line 420: Could we add a check what the schema looks like when we read the table returned from the merge?

I suspect it may have the timezone in UTC, or will it also have the specified tz of source?

ion-elgreco: It will have the timezone in UTC, but that's already covered by the other schema conversion tests

PR #3510, opened 2025-06-01
feat: add convenience extension for kernel engine types

roeap:
# Description

Adding some extension traits, that make working with delta kernel abstractions a little less verbose. mainly exposing apis in therms of `RecordBatch`es rather than `EngineData`.


roeap:
We'll see test coverage improve as we start using more abstractions in subsequent PRs ...

PR #3513, opened 2025-06-02
:bug: retry advancement of `PreparedCommit` into `PostCommit` in case version 0 has been created (by another writer)

### LINKED ISSUE #3505: Initial table creation may fail with multiple writers
# Environment

**Delta-rs version**: 0.25.5

**Binding**: Python

**Environment**:
- **Cloud provider**: Nebius
- **OS**: Linux
- **Other**:

***
# Bug

**What happened**:

When attempting to write multiple partitions for a new table that doesn't exist yet (using Cloudflare's R2), some of the writers may fail with:

```py
_internal.DeltaError: Delta transaction failed, version 0 already exists.
  File "/usr/local/lib/python3.11/site-packages/dagster/_core/execution/plan/utils.py", line 56, in op_execution_error_boundary
    yield
  File "/usr/local/lib/python3.11/site-packages/dagster/_utils/__init__.py", line 391, in iterate_with_context
    next_output = next(iterator)
                  ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dagster/_core/execution/plan/execute_step.py", line 767, in _gen_fn
    gen_output = output_manager.handle_output(output_context, output.value)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dagster/_core/storage/upath_io_manager.py", line 448, in handle_output
    self.dump_to_path(context=context, obj=obj, path=path)
  File "/usr/local/lib/python3.11/site-packages/dagster_polars/io_managers/base.py", line 144, in dump_to_path
    type_router.dump(obj, path, dump_fn)
  File "/usr/local/lib/python3.11/site-packages/dagster_polars/io_managers/type_routers.py", line 77, in dump
    dump_fn(cast(OutputContext, self.context), obj, path)
  File "/usr/local/lib/python3.11/site-packages/dagster_polars/io_managers/delta.py", line 230, in write_df_to_path
    df.write_delta(
  File "/usr/local/lib/python3.11/site-packages/polars/dataframe/frame.py", line 4616, in write_delta
    write_deltalake(
  File "/usr/local/lib/python3.11/site-packages/deltalake/writer.py", line 339, in write_deltalake
    write_deltalake_rust(
```

This doesn't happen with subsequent table updates, only with the initial (0) version. 

`delta-rs` should not produce this error

**How to reproduce it**:

Start multiple writers for a new partitioned DeltaLake table (not sure if using R2 makes a difference)

**More details**:

danielgafni:
# Description
This PR fixes the `PreparedCommit` to fallback to the retry mechanism in case version 0 has been created by another writer. 

# Related Issue(s)


danielgafni:
What is the DCO check complaining about? My commit has been signed (even confirmed by GitHub). Does it expect a specific format? 

image

ion-elgreco:
@danielgafni do git commit -m "msg" -s

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -618,19 +618,26 @@ impl<'a> std::future::IntoFuture for PreparedCommit<'a> {
...
+                            custom_execute_handler: this.post_commit_hook_handler,
+                            metrics: CommitMetrics { num_retries: 0 },
+                        })
+                    }
+                    Err(TransactionError::VersionAlreadyExists(0)) => Ok(()), // this can happen if the table has been created by another writer since the `this.table_data.is_none()` check above
+                    Err(e) => Err(e),
+                }?;
             }
 
             // unwrap() is safe here due to the above check
```
ion-elgreco: You need to also load the snapshot now when VersionAlreadyExists was thrown the first time in it's current state this will panic.

danielgafni:
@ion-elgreco I do have commit signing enabled by default, and GitHub recognizes them as well (see the green `Verified` label next to my commit). This makes me think there is something specific about this DCO check configuration that my signing doesn't do. 

Edit: actually this is the case --- the `-s` option introduces the commit authorship text which was not present previously. Looks like it has to be [configured in the commit message template](https://www.reddit.com/r/git/comments/1ffax9z/comment/lmuezl3/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) for automatic signing.

crates/core/src/kernel/transaction/mod.rs
danielgafni: Uhh right. Done. 
Can we test this? Seems tricky since we'd have to perfectly time writing the 0 version from another writer.

danielgafni: Should we also initialize `attempt_number` to 2 in this case?

ion-elgreco: Would be easier to test in Rust

ion-elgreco: Can you add a test for this please? There are some concurrency tests, somewhere in the rust codebase which you can use as template

danielgafni: I'll look into that. A more direct pointer would be appreciated

ion-elgreco: `delta-rs/crates/test/src/concurrent.rs` here you can find some related stuff

danielgafni:
@ion-elgreco could you please take another look?

Issue #2580, opened 2024-06-07
Don't commit if commit has transaction identifier that has been committed already


vegarsti:
I'm not sure if this is a bug or a feature request. It might be that I'm using the library incorrectly or misunderstanding how this should work. But, what I'm seeing is that if I make multiple commits with the same app_transaction (same ID and version), they will all be written. Is that the correct behavior?
This commit seems to me to indicate that delta-rs should be able to only keep the latest version per txn app ID, so maybe I'm doing something wrong? 54cfa0638e6861d56efe3aa57680a961ef571381

In case I am onto something, here is a branch I've pushed with a test that fails on main, but which passes with my modification: It writes two commits where both have the same app ID + version. We verify that the second write did not cause a new active file.

There's some nuance here, though, which is that it seems to be allowed for a commit to have multiple app transactions. In which case I have no idea what the correct behavior should be...

Branch https://github.com/vegarsti/delta-rs/commit/4e52f4f8f52a1e336a4697b83e0606e3bd7d34aa

simulified:
sup
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 5; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

Blajda:
If the commits are performed in series then the behavior you are experiencing is expected. See the below quote from the [Delta protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers). 

> The semantics of the application-specific version are left up to the external system. Delta only ensures that the latest version for a given appId is available in the table snapshot. The Delta transaction protocol does not, for example, assume monotonicity of the version and it would be valid for the version to decrease, possibly representing a "rollback" of an earlier transaction.

I don't think the current changes in your branch should be merged into delta-rs since it's application specific.
The enforcement of monotonicity should be either performed in application code. E.G in this case you write a wrapper around operations for your application needs. 

An alternative is to generalize your changes in `transaction/mod.rs`  to instead use some hook interface and then have a specific monotonicity hook implementation that others can use. If the hook approach is decided then maybe a new custom delta-rs  config can be added e.g `delta-rs.transaction-id` that would then be used. Just need to be careful here since there might be a few foot traps.  (e.g `rollbacks`)

vegarsti:
[...]

PR #3515, opened 2025-06-03
fix: remove forced table update from python writer

ohanf:
# Description
Performing (concurrent) idempotent blind-appends by comparing transaction identifiers from python is not possible. The reason for this is the internal table state is updated automatically before the actual write is performed. To perform idempotent blind appends (adapted from Spark's strategy for the same) one must load a table snapshot and compare their App ID + version to what they find in the snapshot, if the "local" version is > the one in the snapshot then the write can proceed, and any race conditions for that specific app+version are handled by the conflict checker and atomic commits. However when the table state is updated, the snapshot the app logic used for the comparison is replaced so the [conflict check for updated transactions](https://github.com/delta-io/delta-rs/blob/2ffaacc9e1d892aa2a9a0d773efb416d9adb8bf5/crates/core/src/kernel/transaction/conflict_checker.rs#L575-L594) will always pass and the write will always be attempted.

This PR simply removes the automatic snapshot update from the python writer.

Running the following example script at the same time in two different windows (but same directory) should illustrate the issue:

```python
import sys
import time
from datetime import datetime
from pathlib import Path

import pyarrow as pa

from deltalake import CommitProperties, DeltaTable, Transaction, write_deltalake

TABLE = "concurrency"
data = pa.Table.from_pylist([{"example": 1}])

if not Path(TABLE).exists():
    write_deltalake(TABLE, data)

def write(table: DeltaTable, version: int, wait: int, skip: bool) -> None:
    txn = Transaction(app_id="demo", version=version)
    props = CommitProperties(app_transactions=[txn], max_commit_retries=0)
    table.update_incremental()
    print(f"{datetime.now()}> table updated to {table.version()=}, sleeping {wait}")
    time.sleep(wait)
    rtx = table.transaction_version("demo")
    if skip and rtx and rtx >= version:
        print(f"{datetime.now()}> skipping since version already written")
        return

    write_deltalake(table, data, mode="append", commit_properties=props)
    
    print(f"{datetime.now()}> wrote {version=}, {table.version()=}")


if __name__ == "__main__":
    wait = int(sys.argv[1])
    skip = False
    if len(sys.argv) > 2:
        skip = True
    table = DeltaTable(TABLE)
    for v in range(5):
        write(table, v, wait, skip)
        print("loop", v)
```

screenshot from running against main
concurrent writers main demo

and against this branch
concurrent writers fixed demo

This should probably be considered a bug fix, but for some it may be "breaking" as writes were previously always attempted and now they may be aborted. However the scope is relatively small as the impact is for writes that included `Transaction` properties, if we remove the commit properties from the demo script we will see warnings on write that there is a mismatch in expected and written version, but no hard failures:
> WARN  deltalake_core::kernel::transaction] Attempting to write a transaction 3 but the underlying table has been updated to 4

# Related Issue(s)
- #2580 

# Documentation

### DIFF
--- a/python/deltalake/writer/writer.py
+++ b/python/deltalake/writer/writer.py
@@ -107,7 +107,6 @@ def write_deltalake(
     if table is not None:
         storage_options = table._storage_options or {}
         storage_options.update(storage_options or {})
-        table.update_incremental()
 
     if isinstance(partition_by, str):
         partition_by = [partition_by]
--- a/python/tests/test_writer.py
+++ b/python/tests/test_writer.py
@@ -1,3 +1,4 @@
+import contextlib
 import json
 import os
 import pathlib
@@ -10,7 +11,13 @@ from arro3.core import Array, ChunkedArray, DataType, RecordBatchReader, Table
 from arro3.core import Field as ArrowField
 
 from deltalake import CommitProperties, DeltaTable, Transaction, write_deltalake
-from deltalake._internal import Field, PrimitiveType, Schema, StructType
+from deltalake._internal import (
+    CommitFailedError,
+    Field,
+    PrimitiveType,
+    Schema,
+    StructType,
+)
 from deltalake.exceptions import (
     DeltaError,
     DeltaProtocolError,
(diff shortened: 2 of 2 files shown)


ion-elgreco:
@ohanf is it possible for you to add a test for this new behaviour?

@roeap  can you also take a look, I am not an expert on Spark delta transactions

ohanf:
> @ohanf is it possible for you to add a test for this new behaviour?

@ion-elgreco done! Let me know if that covers what you wanted to see :)

FWIW running the new test cases against main has two of them fail (as expected, the reason I opened this PR):
```
FAILED tests/test_writer.py::test_write_concurrent_blind_appends[writer1_txn0-writer2_txn0-False-True] - Failed: DID NOT RAISE <class '_internal.CommitFailedError'>
FAILED tests/test_writer.py::test_write_concurrent_blind_appends[writer1_txn2-writer2_txn2-False-True] - Failed: DID NOT RAISE <class '_internal.CommitFailedError'>
```

ion-elgreco (approved):
It's breaking in behavior, but new behavior is more correct

ohanf:
Hi @roeap, anything I can do to help move this along?

roeap:
@ohanf - currently I am traveling and therefore unfortunately have no bandwidth to do proper reviews. I'll be back next week and have capacity again.

Sorry for the delay!

ohanf:
> @ohanf - currently I am traveling and therefore unfortunately have no bandwidth to do proper reviews. I'll be back next week and have capacity again.
> 
> Sorry for the delay!

no worries, safe travels!

rtyler (approved):
The behavior change requires a version change which I have added. Thank you for adding some tests!

PR #3518, opened 2025-06-04
feat: make TableConfig accessible

ion-elgreco:
# Description
Make it easier for other libraries to wrap DeltaTable and reinstantiate the state from the provided input.


rtyler (approved):
I don't fully understand the usecase here but don't see a downside for this code presently

jacobtomlinson:
> I don't fully understand the usecase here

For future readers of this PR here's a little more context. A while ago `delta-rs` intentionally broke the ability to pickle `DeltaTable` objects. Pickle can be a footgun for users from a security perspective when used incorrectly.

However, this broke downstream projects that use pickle. Specifically, Dask is a distributed framework for Python that uses `pickle` (via `cloudpickle`) to pass objects between workers. This is a valid and secure way to use pickle, but Dask's support for Delta Lake was broken due to the pickle changes here.

We petitioned @ion-elgreco to reinstate the ability to pickle but that request was rejected. This PR was a compromise which allows us to subclass `DeltaTable` in Dask and fix the pickle support so that instances of our subclass can be passed between distributed workers. We appreciate being able to support Delta Lake in Dask again.

PR #3520, opened 2025-06-05
chore: update to DataFusion `48.0.0` / arrow to 55.2.0

alamb:
# Description

# Related Issue(s)
- Part of https://github.com/apache/datafusion/issues/15771

This tests upgrading delta-rs to DataFusion 48. It seems to have gone fine, but
I hit a few rough edges that I will list and file fixes for in DataFusion:

# Documentation


crates/core/src/delta_datafusion/mod.rs
```diff
@@ -45,14 +45,15 @@ use datafusion::catalog::{Session, TableProviderFactory};
...
-    ParquetSource,
+    FileSource, ParquetSource,
 };
 use datafusion::datasource::{listing::PartitionedFile, MemTable, TableProvider, TableType};
 use datafusion::execution::context::{SessionConfig, SessionContext, SessionState, TaskContext};
 use datafusion::execution::runtime_env::RuntimeEnv;
 use datafusion::execution::FunctionRegistry;
 use datafusion::optimizer::simplify_expressions::ExprSimplifier;
-use datafusion::physical_optimizer::pruning::{PruningPredicate, PruningStatistics};
+use datafusion::physical_optimizer::pruning::PruningPredicate;
```
alamb: PruningStatistics is no longer accessible via `datafuson::physical_optimizer` as it was moved. We should fix this with a `pub use` I think

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -734,39 +735,37 @@ impl<'a> DeltaScanBuilder<'a> {
...
-            } else {
-                file_groups.into_values().map(FileGroup::from).collect()
-            },
-        )
-        .with_statistics(stats)
-        .with_projection(self.projection.cloned())
-        .with_limit(self.limit)
-        .with_table_partition_cols(table_partition_cols)
-        .build();
+        let file_source =
```
alamb, line 747: this is all reformatting as now `with_schema_adapter_factory` returns an Arc directly

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -1114,8 +1113,11 @@ impl ExecutionPlan for DeltaScan {
         Some(self.metrics.clone_inner())
     }
 
-    fn statistics(&self) -> DataFusionResult<Statistics> {
-        self.parquet_scan.statistics()
+    fn partition_statistics(
```
alamb: statistics was deprecated

crates/sql/src/parser.rs
```diff
@@ -136,31 +136,36 @@ impl<'a> DeltaParser<'a> {
...
-            }
+                _ => self.parse_with_native(),
+            },
+            _ => self.parse_with_native(),
         }
     }
 
+    /// Use the native parser to parse the statement
+    // TODO fix for multiple statememnts and keeping parsers in sync
+    // TODO AAL why did this signature change?
```
alamb: it isn't clear why this signature changed to return a different error type. I think we should restore the previous API

alamb:
@ion-elgreco or @rtyler is there any chance you can kick off the CI tests for this PR (so I can verify that DataFusion 48.0.0 passes the delta.rs tests?)

crates/core/src/delta_datafusion/mod.rs
alamb: PR with fix:
- https://github.com/apache/datafusion/pull/16264

crates/sql/src/parser.rs
alamb: - It seems it was changed in https://github.com/apache/datafusion/pull/15680

alamb: It is because now DataFuson potentially returns additional context about what was expected (Diagnostics) so I think this is more or less expected. I will see if I can update the whole parser to use DataFusion error instead

alamb:
CI failure is due to my unpinning arrow-rs
https://github.com/delta-io/delta-rs/actions/runs/15471319016/job/43562193454?pr=3520 

I will try and unpin it in datafusion as well

alamb:
Ok, I think this is ready for another round of tests

crates/core/src/kernel/snapshot/replay.rs
```diff
@@ -311,7 +312,7 @@ fn parse_partitions(batch: RecordBatch, partition_schema: &StructType) -> DeltaR
 
     insert_field(
         batch,
-        StructArray::try_new(
+        StructArray::try_new_with_length(
```
alamb: I updated a few calls to use `StructArray::try_new_with_length(` as without them several of the unit tests were failing

crates/sql/src/parser.rs
```diff
@@ -136,29 +137,45 @@ impl<'a> DeltaParser<'a> {
...
+                _ => self.parse_with_native(),
+            },
+            _ => self.parse_with_native(),
+        }
+    }
+
+    /// Use the native parser to parse the statement
+    // TODO fix for multiple statememnts and keeping parsers in sync
+    fn parse_with_native(&mut self) -> Result<Statement, ParserError> {
+        // DataFusion's parser returns DataFusionError as it may also add
```
alamb, line 154: It is unfortunate to have to translate back from DataFusion error to a ParserError, but I think it is unavoidable as now DataFusion will potentially augment ParserErrors with Span and context information

To retain the same API we need to strip off any additional information and return the underlying `ParserError`

alamb:
Ok, I resolved some other issues and updated to DataFusion 48 rc2 -- things are looking good for me locally. Will wait for the CI results

alamb:
The remaining failures don't look to me like they are related to this PR so I am claiming success

rtyler:
@alamb This cannot be merged until [this fix](https://github.com/apache/arrow-rs/pull/7539) is released,  The patched crates was hiding the test failure which was still hunting behind a `filter_record_batch` release

alamb:
> @alamb This cannot be merged until [this fix](https://github.com/apache/arrow-rs/pull/7539) is released, The patched crates was hiding the test failure which was still hunting behind a `filter_record_batch` release

Makes sense -- I plan to make an arrow/parquet release later this week so hopefully we can upgrade shortly after that

alamb:
I broke out just the arrow upgrade in a separate PR for your consideration
- https://github.com/delta-io/delta-rs/pull/3558

Then I will rebase this one on top of the arrow update and it should be good to go

rtyler (approved):

alamb:
woohoo!

PR #3522, opened 2025-06-06
fix: using state provided in args in merge op

gtrawinski:
# Description
This change allows to control SessionState in the merge function (similarly as in [update](https://github.com/delta-io/delta-rs/blob/2ffaacc9e1d892aa2a9a0d773efb416d9adb8bf5/crates/core/src/operations/update.rs#L276) or [delete](https://github.com/delta-io/delta-rs/blob/2ffaacc9e1d892aa2a9a0d773efb416d9adb8bf5/crates/core/src/operations/delete.rs#L208) ops). 
We've tested this change on our fork and with memory limit set in SessionContext the usage dropped by a half with no visible impact on the jobs. That might help services potentially running out of memory and panicking. That might result in killing all connections with no warning/tracing. 

# Related Issue(s)

# Documentation


crates/core/src/operations/merge/mod.rs
```diff
@@ -756,8 +756,7 @@ async fn execute(
         extension_planner: MergeMetricExtensionPlanner {},
     };
 
-    let state = SessionStateBuilder::new()
-        .with_default_features()
+    let state = SessionStateBuilder::new_from_existing(_state)
```
alexwilcoxson-rel: change from `_state` to `state` and update parameter as it is used now

Issue #3530, opened 2025-06-13
Feature: Vacuum with version retention


corwinjoy:
# Description
Allow specifying specific versions to retain when performing a vacuum operation.

**Use Case**
We have specific data history that we want to keep. Generally, we want to read either:
1. A recentish data version (within the last 7 days).
2. A collection of old fixed versions going back 250 days.

We cannot run a general vacuum operation, as that will remove files referenced in the old fixed versions (2).
What we would like, is more control of the vacuum operation to specify what versions to keep along with the associated files.
So, our old delta files directory might look like
```
{version 0-files}
{version 1-files}
{version 2-files}
...
{version 178-files}
{version 179-files}
{version 180-files}
```

And then we want to perform a vacuum where we keep versions {2, 179, 180} to get a new directory like:
```
{version 2-files}
{version 179-files}
{version 180-files}
```

We propose enhancing the vacuum operation to add an option to keep specific versions as specified in the VacuumBuilder. The idea would be to provide a vector of versions to keep. Then, after VACUUM it would still be possible to time travel back to the specified versions.
https://docs.rs/deltalake/latest/deltalake/operations/vacuum/struct.VacuumBuilder.html

**Related Issue(s)**
[[Delete older files based on VERSION](https://github.com/delta-io/delta-rs/issues/3143#top)](https://github.com/delta-io/delta-rs/issues/3143)

**Background Reading**
Vacuuming: https://delta.io/blog/remove-files-delta-lake-vacuum-command/
Time Travel: https://delta.io/blog/2023-02-01-delta-lake-time-travel/

corwinjoy:
@adamreeve @CallumREdwards @starzar

PR #3537, opened 2025-06-17
feat: vacuum with version retention

corwinjoy:
# Description
Upgrade the vacuum operation to specify versions to retain when performing a vacuum operation.

# Related Issue(s)
Closes https://github.com/delta-io/delta-rs/issues/3530

### DIFF
--- a/crates/core/src/operations/vacuum.rs
+++ b/crates/core/src/operations/vacuum.rs
@@ -21,7 +21,7 @@
 //! let (table, metrics) = VacuumBuilder::new(table.object_store(). table.state).await?;
 //! ````
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use std::fmt::Debug;
 use std::sync::Arc;
 
@@ -176,8 +176,9 @@ impl VacuumBuilder {
 
     /// Specify table versions that we want to keep for time travel.
     /// This will prevent deletion of files required by these versions.
-    pub fn with_keep_versions(mut self, versions: Vec<i64>) -> Self {
-        self.keep_versions = Some(versions);
+    pub fn with_keep_versions(mut self, versions: &[i64]) -> Self {
+        warn!("Using experimental API VacuumBuilder::with_keep_versions");
+        self.keep_versions = Some(versions.to_vec());
         self
     }
 
@@ -519,9 +520,11 @@ async fn get_stale_files(
 
 #[cfg(test)]
 mod tests {
+    use object_store::{local::LocalFileSystem, memory::InMemory, GetResult, PutPayload};
+
     use super::*;
-    use crate::open_table;
-    use std::time::SystemTime;
+    use crate::{checkpoints::create_checkpoint, open_table};
+    use std::{io::Read, time::SystemTime};
 
     #[tokio::test]
     async fn test_vacuum_full() -> DeltaResult<()> {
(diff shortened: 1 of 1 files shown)


corwinjoy:
@adamreeve @CallumREdwards

crates/core/src/operations/vacuum.rs
```diff
@@ -235,6 +245,43 @@ impl VacuumBuilder {
             None => Utc::now().timestamp_millis(),
         };
 
+        let keep_files = match &self.keep_versions {
+            Some(versions) => {
+                let max_version = self.snapshot.version();
```
adamreeve: Rather than iterating over all versions from 0 to the max version and getting the files for every version, can you instead iterate over just the versions specified in `keep_versions`? (Maybe after sorting them first to simplify the state updates). Then you should be able to remove the `files_by_version` variable and just add files directly to `keep_files`.

corwinjoy: I can do that. The reason I wrote it this way was twofold:
1. Debugging the file dependency is tricky. By building an intermediate breakout, we can more clearly see the required files for each version. This helps diagnose future problems.
2. To get the file dependency for each version, the underlying code has to replay all log files up through max(keep_versions). So, the performance gain by switching the code is not all that great.

crates/core/src/operations/vacuum.rs
```diff
@@ -235,6 +245,31 @@ impl VacuumBuilder {
...
+                let mut keep_files: HashSet<String> = HashSet::new();
+                let mut state =
+                    DeltaTableState::try_new(&self.log_store, DeltaTableConfig::default(), Some(0))
+                        .await?;
+                let mut sorted_versions = versions.clone();
+                sorted_versions.sort();
+                for version in sorted_versions {
+                    if version == 0 {
+                        continue;
+                    }
```
adamreeve: I think this continue isn't needed, we might need to keep some files from version 0, and it looks like `state.update` handles updating to the version it's already on and will return early.

crates/core/src/operations/vacuum.rs
```diff
@@ -235,6 +245,31 @@ impl VacuumBuilder {
             None => Utc::now().timestamp_millis(),
         };
 
+        let keep_files = match &self.keep_versions {
+            Some(versions) => {
+                let mut keep_files: HashSet<String> = HashSet::new();
+                let mut state =
+                    DeltaTableState::try_new(&self.log_store, DeltaTableConfig::default(), Some(0))
```
adamreeve: Rather than start from version 0, should we start from `sorted_versions[0]` (with a length check somewhere)? That might avoid the work needed in the first update call if a snapshot can be used.

crates/core/src/operations/vacuum.rs
adamreeve: That's fair, but I worry that for very large tables with lots of updates the hash map of files per version could get very large. On your comment about having to replay all log files, I think we can avoid some of that work by starting from the first version to keep rather than version 0, assuming snapshots can be used to avoid replaying all log files (I left an inline comment about this too).

crates/core/src/operations/vacuum.rs
corwinjoy: Actually I think it is needed. I am basing this logic on how load_version is written which uses update_incremental. See: https://github.com/delta-io/delta-rs/blob/eff0b582a7e31242494b9ba77a0441174c6c2227/crates/core/src/table/mod.rs#L210
I want to follow the logic they use there. From my reading, it seems version 0 cannot have files. Anyway, in the code, the constructed state starts from version 0.

crates/core/src/operations/vacuum.rs
corwinjoy: See above, this comes from tracing the logic in load_version.

corwinjoy:
Hey @rtyler , any thoughts on this PR? I just wanted to check in.

rtyler:
@corwinjoy I did some poking and refactored the tests a bit. I still have some unease about this change so added a `warn!()` for users who use it.

corwinjoy:
@rtyler Thanks for looking this over and the improved tests! As you say, it's good to be cautious, and I like your improved code that verifies table integrity further. A `warn! `probably makes sense since this is a new feature. Still, I believe it will significantly help many users to have further control over what is being vacuumed. As mentioned above, there have been previous feature requests for something like this, so we hope it will help the community! Thanks again!

rtyler (approved):

PR #3539, opened 2025-06-17
feat(optimize): guarantee global sorting in OPTIMIZE output

### LINKED ISSUE #3538: Allow compacting and maintaining a monotonic ordering
## Problem Statement

After running `OPTIMIZE` (compact) on a Delta table, the file/row layout is not guaranteed to be globally sorted by `(objectId, dateTime)`
clients must perform an extra sort on read, which adds latency, complexity, and resource pressure.

## Background & Motivation

- Delta Lake’s current compaction merges files but does not enforce a global sort order.
- Existing Z-order / clustering features are orthogonal; we need deterministic, monotonic ordering.

## Proposed Feature

1. Add a **global sort** phase to the compaction pipeline:
   - After bin-packing files, feed all pages/records through a DataFusion `df.sort(["objectId","dateTime"])` before writing.
   - Preserve the existing unsorted path when sorting is disabled for backward compatibility.

2. Expose new options on the `OptimizeBuilder` (Rust) and `DeltaTable.optimize` (Python):
   - `sort_enabled: bool` (default = true)
   - `sort_columns: Vec<String>` / `List[str]` (default = `["objectId","dateTime"]`)
   - Builder methods:
     - Rust: `.with_sort_columns(&[...])`, `.disable_sort()`
     - Python: `dt.optimize.compact(sort_enabled=False, sort_columns=["foo"])`

  ## Requirements & Constraints

  - **Strict Ordering**: Global sort by `(objectId, dateTime)` across all partitions and pages.
  - **Performance**: Sorting should not unduly impact compaction throughput; use DataFusion’s spillable memory pools.
  - **Configurable**: Users can disable sorting or choose a different sort key.
  - **Backward Compatible**: Unsorted compaction remains available; default behavior may change only when opted in.

tolleybot:
feat(optimize): guarantee global sorting in OPTIMIZE output

After running `OPTIMIZE` on a Delta table, files were not guaranteed to be
globally sorted by `(objectId, dateTime)`, forcing downstream consumers to
perform an extra sort. This change adds a global sort phase in the compaction
pipeline and exposes new flags to toggle or customize sorting.
 
Closes #3538


python/src/lib.rs
```diff
@@ -572,6 +574,8 @@ impl RawDeltaTable {
         writer_properties: Option<PyWriterProperties>,
         commit_properties: Option<PyCommitProperties>,
         post_commithook_properties: Option<PyPostCommitHookProperties>,
+        sort_enabled: bool,
```
ion-elgreco: This flag is unnecessary, passing cols should be enough to have it be disabled or not, same applies for the rust builder. Just let some() is enough to trigger a different match arm

python/deltalake/table.py
```diff
@@ -158,8 +159,12 @@ def __init__(
 
         """
         self._storage_options = storage_options
+        # Resolve local paths to absolute, leave remote URLs intact
+        uri_str = str(table_uri)
+        if "://" not in uri_str:
+            uri_str = os.path.abspath(uri_str)
```
ion-elgreco: Why are you adding this?

python/tests/conftest.py
```diff
@@ -8,7 +8,10 @@
 from typing import TYPE_CHECKING
 
 import pytest
-from arro3.core import Array, DataType, Field, Schema, Table
+try:
+    from arro3.core import Array, DataType, Field, Schema, Table
+except ImportError:
+    pytest.skip("arro3 not installed; skipping Python deltarlake tests", allow_module_level=True)
```
ion-elgreco: Arro3 is required so this try except should exist

ion-elgreco (changes requested):
Fine by me to add this, I can see the value. Just left a couple comments

crates/core/src/operations/optimize.rs
```diff
@@ -238,6 +400,8 @@ impl<'a> OptimizeBuilder<'a> {
             max_concurrent_tasks: num_cpus::get(),
             max_spill_size: 20 * 1024 * 1024 * 1024, // 20 GB.
             optimize_type: OptimizeType::Compact,
+            sort_enabled: true,
```
hntd187: This shouldn't be enabled by default.

crates/core/src/operations/optimize.rs
```diff
@@ -238,6 +400,8 @@ impl<'a> OptimizeBuilder<'a> {
...
+            sort_columns: vec!["objectId".to_string(), "dateTime".to_string()],
```
hntd187: These seem like they are specific to your data? I don't think they should be the default.

crates/core/src/operations/optimize.rs
```diff
@@ -606,44 +789,123 @@ impl MergePlan {
...
+
+                        let rewrite_result = tokio::task::spawn(Self::rewrite_files(
+                            task_params.clone(),
+                            partition,
+                            files,
+                            store.object_store(Some(operation_id)),
+                            futures::future::ready(Ok(batch_stream)),
+                        ));
+                        util::flatten_join_error(rewrite_result)
+                    } else {
```
hntd187, line 954: This else branch looks like it's a complicated way to do a z-order as below. I might be misunderstanding, could you explain what is happening here?

python/src/lib.rs
```diff
@@ -559,7 +559,9 @@ impl RawDeltaTable {
         min_commit_interval = None,
         writer_properties=None,
         commit_properties=None,
-        post_commithook_properties=None
+        post_commithook_properties=None,
+        sort_enabled = true,
```
hntd187: enabling sort with no columns?

python/src/lib.rs
hntd187: I agree, I think this flag is unnecessary.

python/tests/conftest.py
hntd187: *nit* also deltalake is spelled incorrectly

hntd187 (changes requested):
I'm not entirely convinced this PR necessary. Why is doing a z-order not appropriate with the optimize? I don't really agree it should be "just done" for the user, even Databricks doesn't automatically order the output, it requires a ZORDER clause in a SQL statement.

python/src/lib.rs
tolleybot: Im setting sort_enabled to false by default

crates/core/src/operations/optimize.rs
tolleybot: agree

crates/core/src/operations/optimize.rs
tolleybot: I agrree.  I made the changes

crates/core/src/operations/optimize.rs
tolleybot: I made the changes

crates/core/src/operations/optimize.rs
tolleybot: That else block under if !self.sort_enabled is not doing Z-ordering at all, but I added new global sort path for
compaction:

  1. We spin up a DataFusion runtime (with a spillable memory pool).
  2. Register the Delta object store under delta-rs://
  3. Build a DataFrame over exactly the same files in this merge‐bin via DeltaTableProvider
  4. Call df.sort(sort_columns…) on the user-supplied keys
  5. Turn the sorted plan back into a Stream<RecordBatch>
  6. Hand that stream into our existing rewrite_files() writer, so all output files are globally ordered.

The actual Z-order logic lives separately in the OptimizeOperations::ZOrder branch (around line 917) where we invoke a custom ZOrderUDF. I’ve expanded the comment above the else-block in optimize.rs to spell this out. Hope that clears it up—please let me know if you’d like any further tweaks!

python/tests/conftest.py
tolleybot: @ion-elgreco could you clarify your comment concerning the try except.  Do you want me to revert it or are you ok with it the way I have it? (besides the spelling)

crates/core/src/operations/optimize.rs
```diff
@@ -606,44 +810,129 @@ impl MergePlan {
...
+                        // === Compaction with global sort ===
+                        // Instead of simply merging files, we build a DataFusion plan
+                        // over the same set of files, apply a full global sort on the
+                        // user-specified `sort_columns`, and then rewrite pages in that
+                        // sorted order. This guarantees monotonic ordering across all
+                        // output files. Note: this is a standard DataFusion `sort(...)`
+                        // operation, not Z-ordering (see the ZOrder branch below).
```
hntd187: So judging by the comment which I might be misunderstanding, this rewrites the entire table?

crates/core/src/operations/optimize.rs
```diff
@@ -606,44 +810,129 @@ impl MergePlan {
...
+                        // Use cloned store for DataFusion operations
+                        let log_store_ref = store.clone();
+                        let snapshot = snapshot.clone();
+                        let sort_cols = self.sort_columns.clone();
+                        // Extract underlying file list from MergeBin
+                        let files_for_provider = files.files.clone();
+                        // Asynchronous DataFrame-based sorted reader
+                        let read_sorted = async move {
+                            let memory_pool = FairSpillPool::new(max_spill_size);
+                            let runtime = RuntimeEnvBuilder::new()
```
hntd187: You shouldn't need this machinery shouldn't session ctx and proper object stores already be registered?

crates/core/src/operations/optimize.rs
```diff
@@ -87,6 +87,184 @@ pub struct Metrics {
...
+    use crate::operations::DeltaOps;
+    use crate::writer::test_utils::create_initialized_table;
+    use crate::writer::test_utils::datafusion::get_data;
+    use arrow_array::{RecordBatch, StringArray};
+    use arrow_schema::{DataType as ArrowDataType, Field, Schema as ArrowSchema};
+    use std::sync::Arc;
+    use tempfile::TempDir;
+
+    #[tokio::test]
+    async fn test_optimize_global_sort_enabled() {
```
hntd187, line 257: I don't know if this test does what we need here. This shows a single partition is sorted, which is fine, but that's the same as a Z-order. I think you'd need to partition the table and add a bit more data in order to validate it sorts correctly across partitions.

python/deltalake/table.py
tolleybot: Ive reverted this change. it was an error on my side.

crates/core/src/operations/optimize.rs
tolleybot: Thanks for flagging that, your reading is correct that the original wording made it sound like we were rewriting the entire table. I’ve updated the comment to make it more clear that we only build a DataFusion plan and rewrite the files in the current compaction batch in sorted order (not the full table)

crates/core/src/operations/optimize.rs
tolleybot: Ok, thanks for the feedback.  Ill put something together

tolleybot: Thanks for catching that. I’ve updated test_optimize_global_sort_enabled so it creates the table partitioned on objectId, now when we call
.optimize().with_sort_columns(&["objectId","dateTime"]), it will scan multiple partitions, globally sort across them, and rewrite each partition in order.
This ensures we’re truly testing cross-partition sorting

crates/core/src/operations/optimize.rs
tolleybot: Your correct. The DeltaTableProvider already carries the correct object store, so we don’t need to spin up a custom RuntimeEnv or manually register anything.

crates/core/src/operations/optimize.rs
```diff
@@ -606,44 +810,114 @@ impl MergePlan {
...
+                                log_store_ref.clone(),
+                                scan_config.clone(),
+                            )?
+                            .with_files(files_for_provider);
+                            let df = ctx
+                                .read_table(Arc::new(provider))
+                                .map_err(|e| DeltaTableError::Generic(e.to_string()))?;
+                            let sort_exprs = sort_cols
+                                .iter()
+                                .map(|c| col(c).sort(true, true))
```
hntd187: What if I don't want the order to be ascending and nulls first?

hntd187 (changes requested):
Circling back on this, sorry, but it's not clear to me, could you explain what problem you are trying to solve here in more detail? I'm not convinced this PR is necessary and there seem to be a few places where this could produce unintended results and take lots of extra compute time. If you have a specific use case or problem it'd be helpful to know. I ask because your PR seems very tailored towards your data and there maybe a way to accomplish what you are looking for without this PR.

python/src/lib.rs
tolleybot: agreed, I made that change

crates/core/src/operations/optimize.rs
tolleybot: Ok, Ive updated the code so you can decide if you want ascending or not and how nulls are handled

adamreeve:
> Circling back on this, sorry, but it's not clear to me, could you explain what problem you are trying to solve here in more detail? 

Hi, I work with @tolleybot and can hopefully provide some more context on why we want this feature.

Our use case is that we use Delta Lake to store time-series data, and have consumers that want to read time series and store them in memory with a specific order (by object id then timestamp). We would like to minimise the amount of work these consumers need to do to reorder data.

We can't guarantee a whole table or partition will always be globally ordered like this as data can be split across multiple Parquet files with no inherent order, and any append operations for recent data will break the ordering. But if batches of data that are read from Delta Lake are locally ordered it greatly reduces the work our consumers need to do. Currently we always write ordered data, but the compaction operation arbitrarily reorders data.

For what it's worth, Iceberg allows setting a table order that's used by compaction (https://www.tabular.io/apache-iceberg-cookbook/data-engineering-table-write-order/), so I don't think this requirement is unique to us.

ohanf:
FWIW I found this issue after reading https://blog.cloudflare.com/logexplorer-ga/ which specifically called out the need for this feature in upstream delta-rs, quoting the relevant section 

> While the open-source implementation of Delta Lake provides compaction out of the box, we soon encountered an issue when using it for our workloads. Stock compaction merges data files to a desired target size S by sorting the files in reverse order of their size and greedily filling bins of size S with them. By merging logs irrespective of their timestamps, this process distributed ingested batches randomly across merged files, destroying data locality. Despite compaction, a user querying for a specific time frame would still end up fetching hundreds or thousands of files from R2.

> For this reason, we wrote a custom compaction algorithm that merges ingested batches in order of their minimum log timestamp, leveraging the min/max statistics mentioned previously. This algorithm reduced the number of overlaps between merged files by two orders of magnitude. As a result, we saw a significant improvement in query performance, with some large queries that had previously taken over a minute completing in just a few seconds.

Bringing it up as another data point for something like this feature being useful

hntd187:
I appreciate the feedback, but it's still not clear to me what this is trying to solve that z-order does not solve? I noticed in the tests you are using dates, but they are string typed and not date types, are you sure that those types and z-order does not already solve this? I created several test tables with z-order to try they appear to also do what you are trying to do.

hntd187:
I checked your test out and slightly modified it to write 2 separate tables. They are physically the same and checksums show the methods produce a byte for byte identical output. I think whatever you are doing to re-read the tables back in is altering the ordering of the rows somehow, but on a physical file level they are identical. Is there something I'm missing here?

jeromegn:
I'm very interested in this, but as far as I can tell, z ordering should achieve the same?

hntd187:
> > I checked your test out and slightly modified it to write 2 separate tables. They are physically the same and checksums show the methods produce a byte for byte identical output. I think whatever you are doing to re-read the tables back in is altering the ordering of the rows somehow, but on a physical file level they are identical. Is there something I'm missing here?
> 
> Can I get a copy of your test code. It would be helpful to try myself

It's the same as your code, I just write 2 tables out instead of using the same table

hntd187:
The example above does what you suggest, which makes sense, so I went back and looked at my original test that made identical output and I realized that depending on the type of the associated columns here change the sorting output. If you change the first column to be a string then the output is identical. So sometimes when z-order happens to match up with a global sort sometimes it doesn't. 

With this in mind, I'd say that this is very niche and specific to your use case. There is nothing stopping you from employing this strategy yourself, but I think adding this will confuse users on what they need and anyone who doesn't use delta-rs might also be confused by why this is here or might expect it to exist in other places. Additionally, this assumes you use delta-rs and only delta-rs (likely Databricks's own optimize or z-order will wipe this all out) so while I appreciate your effort to contribute this, I think it would be best to decline this PR.

hntd187:
Closing this, as declined.

PR #3548, opened 2025-06-20
fix: serde checkpoint_files error

aditanase:
# Description

Fix serde error

### DIFF
--- a/crates/core/src/kernel/snapshot/serde.rs
+++ b/crates/core/src/kernel/snapshot/serde.rs
@@ -2,7 +2,7 @@ use std::fmt;
 
 use arrow_ipc::reader::FileReader;
 use arrow_ipc::writer::FileWriter;
-use chrono::{DateTime, TimeZone, Utc};
+use chrono::{TimeZone, Utc};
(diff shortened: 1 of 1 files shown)


rtyler:
@aditanase what error was actually cropping up that led you to this change? I would like to add a regression test to make sure I understand the problem, and I'm happy to write it, but some steps would be helpful

aditanase:
@rtyler @ion-elgreco sorry for the late reply, I was out on vacation for most of July.
It's a very obvious error when trying to deserialize a table that also includes checkpoint files:
>Message:  called `Option::unwrap()` on a `None` value
Location: .../delta-rs/crates/core/src/kernel/snapshot/serde.rs:103

I went in trying to deal with the `Option` and realized that the code was just updated for the `commits` branch above, but not for checkpoints, so I just assumed that people don't test with more complex tables, and proceeded to make it consistent. On a second look, serde probably never worked well since it was serializing nanos and deserializing millis.

aditanase:
Closing as code is not there anymore, this happens in delta-kernel now, and the serde operates at the snapshot level now.

PR #3549, opened 2025-06-20
fix: version binary search

aditanase:
# Description

Version search does not work if there are temporary parquet checkpoint files in the `_delta_log` folder.

The 2nd commit fixes a regression introduced in a recent commit that uses the kernel to look for max_version, which fails if the version doesn't start at 0 (rarely the case on an older table).

### DIFF
--- a/crates/core/src/logstore/mod.rs
+++ b/crates/core/src/logstore/mod.rs
@@ -644,7 +644,7 @@ impl<'de> Deserialize<'de> for LogStoreConfig {
 }
 
 static DELTA_LOG_REGEX: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"(\d{20})\.(json|checkpoint).*$").unwrap());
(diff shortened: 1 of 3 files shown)


ion-elgreco:
A test would be nice

aditanase:
@ion-elgreco thanks! Added 1 tests to `time_travel.rs`, without it it crashes with:
  > called `Result::unwrap()` on an `Err` value: NotATable("../delta-rs/crates/test/tests/data/checkpoints_vacuumed/")

This is covering both changes I made, using the `checkpoints_vacuumed` test table

ion-elgreco (approved):

aditanase:
@ion-elgreco FYI this seems to be the root cause of the issue introduced by migrating to kernel v11 for these operations:
https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/log_segment.rs#L181

It does not handled compacted tables yet.

PR #3550, opened 2025-06-21
fix: use proper DeltaTableState for vacuum commits

### LINKED ISSUE #3534: Attempting to write a transaction 3 but the underlying table has been updated to 3
# Environment

**Delta-rs version**: 0.25.4

**Binding**: Python

**Environment**:
- **Cloud provider**: -
- **OS**: Mac
- **Other**: -

***
# Bug

**What happened**: The underlying table is unexpectedly updated during vacuum.

**What you expected to happen**: No warning message is shown, the table has the expected version.

**How to reproduce it**: Run vacuum right after compact;

```py
from shutil import rmtree

from pyarrow import Table

DELTA_FOLDER = '___/lol'

table = Table.from_pylist([{'x': 123}])

# Two write operations creating optimize/vacuum material
for _ in range(2):
    write_deltalake(
        DELTA_FOLDER, table, mode='append', schema_mode='merge', engine='rust'
    )

# Optimize immediately followed by vacuum triggers warning
delta = DeltaTable(DELTA_FOLDER)
delta.optimize.compact()
delta.vacuum(
    retention_hours=0,
    dry_run=False,
    enforce_retention_duration=False,
)

print(delta.to_pyarrow_table().to_pylist())

if True:
    rmtree(DELTA_FOLDER)
```

**More details**: -

jeromegn:
# Description
The vacuum operation was not using the correct snapshot for the VacuumEnd operation. It was also returning the original snapshot instead of the latest snapshot.

Before this fix, the following was happening:
- vacuum
  - if there was nothing to vacuum, everything was fine, no changes to the DB
  - if there were files to delete:
    - it did a VacuumStart based on `this.snapshot` (correct)
    - it did a VacuumEnd based on `this.snapshot`, which is stale
      - e.g. triggered `Attempting to write a transaction 64 but the underlying table has been updated to 64`
    - it returned `this.snapshot` which had not been updated
    - the next commit was 2 versions behind (the returned table was from before vacuum)
      - e.g triggered `Attempting to write a transaction 64 but the underlying table has been updated to 65`

# Related Issue(s)
- closes #3534

### DIFF
--- a/crates/core/src/operations/vacuum.rs
+++ b/crates/core/src/operations/vacuum.rs
@@ -120,7 +120,7 @@ impl super::Operation<()> for VacuumBuilder {
 }
 
 /// Details for the Vacuum operation including which files were
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct VacuumMetrics {
     /// Was this a dry run
     pub dry_run: bool,
@@ -310,7 +310,7 @@ impl std::future::IntoFuture for VacuumBuilder {
             let operation_id = this.get_operation_id();
             this.pre_execute(operation_id).await?;
 
-            let metrics = plan
+            let result = plan
                 .execute(
                     this.log_store.clone(),
                     &this.snapshot,
@@ -321,10 +321,17 @@ impl std::future::IntoFuture for VacuumBuilder {
                 .await?;
 
             this.post_execute(operation_id).await?;
-            Ok((
-                DeltaTable::new_with_state(this.log_store, this.snapshot),
-                metrics,
-            ))
+
+            Ok(match result {
+                Some((snapshot, metrics)) => (
+                    DeltaTable::new_with_state(this.log_store, snapshot),
+                    metrics,
+                ),
+                None => (
+                    DeltaTable::new_with_state(this.log_store, this.snapshot),
+                    Default::default(),
+                ),
+            })
         })
     }
 }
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/vacuum.rs
```diff
@@ -355,12 +362,9 @@ impl VacuumPlan {
...
         operation_id: uuid::Uuid,
         handle: Option<Arc<dyn CustomExecuteHandler>>,
-    ) -> Result<VacuumMetrics, DeltaTableError> {
+    ) -> Result<Option<(DeltaTableState, VacuumMetrics)>, DeltaTableError> {
         if self.files_to_delete.is_empty() {
-            return Ok(VacuumMetrics {
-                dry_run: false,
-                files_deleted: Vec::new(),
-            });
+            return Ok(None);
```
ion-elgreco, line 364: I would just return the current snapshot here, so that you can remove the match arm above in intofuture

jeromegn: I'd have to clone the snapshot since it's a reference to a snapshot that's passed to this function. Otherwise I can return `(Option<DeltaTableState>, VacuumMetrics)`.

I think cloning _could_ be expensive. Not that this operation happens a lot, but still...

What would you prefer?

jeromegn:
@ion-elgreco any more things I should fix?

ion-elgreco:
> @ion-elgreco any more things I should fix?
> 

Can you fix the PR title?

ion-elgreco (approved):
Thanks!

PR #3570, opened 2025-07-06
feat: convert partition filters to kernel predicates

roeap:
# Description

Another kernel log replay pre-factor. Kernel contains an advanced implementation to handle file skipping. To leverage this with our current partition filter API, we need to convert `PartitionFilter`s into kernel `Predicate`s, which is what we do in this PR.

### DIFF
--- a/crates/core/src/kernel/schema/partitions.rs
+++ b/crates/core/src/kernel/schema/partitions.rs
@@ -3,11 +3,12 @@ use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::convert::TryFrom;
 
-use delta_kernel::expressions::Scalar;
+use delta_kernel::expressions::{Expression, JunctionPredicateOp, Predicate, Scalar};
+use delta_kernel::schema::StructType;
 use serde::{Serialize, Serializer};
 
 use super::{DataType, PrimitiveType};
-use crate::errors::DeltaTableError;
+use crate::errors::{DeltaResult, DeltaTableError};
 use crate::kernel::scalars::ScalarExt;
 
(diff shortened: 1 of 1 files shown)


crates/core/src/kernel/schema/partitions.rs
```diff
@@ -308,9 +313,65 @@ impl TryFrom<&str> for DeltaTablePartition {
...
+                .iter()
+                .map(|v| dt.parse_scalar(v))
+                .collect::<Result<Vec<_>, _>>()?;
+            let (expr, operator): (Box<dyn Fn(Scalar) -> Predicate>, _) = match op {
+                PartitionValue::In(_) => {
+                    (Box::new(|v| column.clone().eq(v)), JunctionPredicateOp::Or)
+                }
+                PartitionValue::NotIn(_) => {
+                    (Box::new(|v| column.clone().ne(v)), JunctionPredicateOp::And)
+                }
```
ion-elgreco, line 362: Doesn't kernel have is_in construct? Wondering about the perf when you chain like a very large is in

ion-elgreco (approved):

crates/core/src/kernel/schema/partitions.rs
roeap: Somewhat ... engines can implement and use it via opaque expression (soon to be merged). Thus I did not want to use the current version of it.

The current default engine implementation to a large degree does that same thing under to hood though. My focus would be on using datafusion expressions once we fully migrated if we want something more optimised.

That said, efficient in-list can be very complex. My guess would be up to tens or hundreds of list elements, this impl should actually perform reasonably well, since kernel is very efficient in evaluating junction predicates.

PR #3580, opened 2025-07-07
feat: if the user asks not to load files, make EagerSnapshot _never_ load files

### LINKED ISSUE #3528: `without_files=True` doesn't apply as soon as version conflicts happen, leading to performance hit
# Environment

**Delta-rs version**: Git, June 1, 2025

***
## Bug

Let's say you're using `without_files=True` / `require_files=false` with the goal of not loading files into memory.

In particular, you have two processes appending at same time. If they don't stomp on each other, all is well, the files don't get loaded into memory, as `EagerSnapshot::try_new_with_visitor()` will see that `require_files` is false, and not load them.

However, if they conflict in version numbers, `EagerSnapshot.update()` is called... and this loads files unconditionally, ignoring the `require_files=false` the user originally passed in which is respected by `EagerSnapshot::try_new_with_visitor()`: 

Compare:

https://github.com/delta-io/delta-rs/blob/9453b2ca644c24eb95d5b974065c8dc7636f2607/crates/core/src/kernel/snapshot/mod.rs#L372

to 

https://github.com/delta-io/delta-rs/blob/9453b2ca644c24eb95d5b974065c8dc7636f2607/crates/core/src/kernel/snapshot/mod.rs#L481

So now you have much larger memory usage, and also more CPU impact.

At scale, if you have many processes appending, this can have a significant impact on performance. With `make develop` here's the performance I got with git, compared to a prototype of options 2/3 mentioned below:

| Version of code | # of processes | Ops / sec |
|------------------------| -------------------|---------------|
| Git                        | 8                     |       10        |
| Git                        | 16                     |       7        |
| Git w/prototype fix | 8                     |        14    |
| Git w/prototype fix | 16                     |        10    |

(My computer has 12 CPU cores).

## How to fix?

Options:

1. It seems a little weird to have `EagerSnapshot` that isn't eager. And a bunch of its methods just lie if `require_files` is false, essentially claiming no files. So one option is to ... have a snapshot trait, and use the appropriate snapshot type depending on whether it's append operation not? This may be major refactoring though.
2. Change `EagerSnapshot.update()` to respect `require_files: false`. I am not sure if this is semantically correct, though.
3. Change `EagerSnapshot.update()` to respect `require_files: false`, but only when appending.

As you can see in table above, options 2/3 definitely speed things up.

itamarst:
Previously the files would be loaded on a conflict even if the user requested they not be loaded, now they're never loaded.

This is a optimization, reducing the performance for loading unneeded files when in append mode.

For my reproducer benchmark (see #3528), this speeds things up by about 30%.

**IMPORTANT:** It is not clear to me if this is semantically correct! On the one hand, if you're doing append only, I don't see why you'd need to load the files, this shouldn't be any different than e.g. restarting from scratch by reopening the database. On the other hand, maybe this code path is used in other situations beyond appends?

Are there other tests I should write?

# Related Issue(s)
Fixes #3528

### DIFF
--- a/crates/core/src/kernel/snapshot/mod.rs
+++ b/crates/core/src/kernel/snapshot/mod.rs
@@ -21,6 +21,7 @@ use std::io::{BufRead, BufReader, Cursor};
 use ::serde::{Deserialize, Serialize};
 use arrow_array::RecordBatch;
 use delta_kernel::path::{LogPathFileType, ParsedLogPath};
+use either::Either;
 use futures::stream::BoxStream;
 use futures::{StreamExt, TryStreamExt};
 use object_store::path::Path;
@@ -342,8 +343,8 @@ pub struct EagerSnapshot {
     pub(crate) transactions: Option<HashMap<String, Transaction>>,
 
     // NOTE: this is a Vec of RecordBatch instead of a single RecordBatch because
-    //       we do not yet enforce a consistent schema across all batches we read from the log.
-    pub(crate) files: Vec<RecordBatch>,
+    //       we do not yet enforce a consistent schema across all batches we read from the log.  If `None`, that indicates this was created with `config.require_files` set to `false`.
+    pub(crate) files: Option<Vec<RecordBatch>>,
 }
 
 impl EagerSnapshot {
@@ -370,13 +371,13 @@ impl EagerSnapshot {
         let snapshot = Snapshot::try_new(log_store, config.clone(), version).await?;
 
         let files = match config.require_files {
-            true => {
+            true => Some(
                 snapshot
                     .files(log_store, &mut visitors)?
                     .try_collect()
-                    .await?
-            }
-            false => vec![],
+                    .await?,
+            ),
+            false => None,
         };
 
         let mut sn = Self {
@@ -424,7 +425,7 @@ impl EagerSnapshot {
             .collect::<DeltaResult<Vec<_>>>()?;
         Ok(Self {
             snapshot,
-            files,
+            files: Some(files),
             tracked_actions: Default::default(),
             transactions: None,
         })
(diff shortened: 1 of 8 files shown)


ion-elgreco:
@itamarst we can put this to the test ofcourse whether this is semantically is correct. Add some tests that does concurrent overwrites with appends and then concurrent schema evolutions. It should reject the append write

@roeap feel free to correct me if my above statement is wrong :)

itamarst:
It's difficult for concurrent tests to be a thorough test of correctness, since there are many possible interleaving execution paths.

rtyler:
@itamarst I think I should be able to come up with a test which doesn't necessarily require the concurrency that your linked issue has. If you're okay with this pull request sitting open for a few days, I will summon the brainpower and come up with something :smile:

itamarst:
Of course, appreciate your feedback!

roeap:
@itamarst @rtyler - Currently I am working on take 3 of getting log replay via kernel in, which touches a lot of the same fields as this PR.

More generally speaking, the `require_files` flag as it is used today should not exist 😆. Or maybe the flqg should exist, but not the way we treat it. I.e. a snapshot without files is the `LazySnapshot`, so rather than emulating the lazy snapshot via optional files on the eager one, we would ideally just use the lazy one ...

However `DeltaTableState` is not quite ready yet to handle both snapshots. And while there is a plan, I do not have a timeline yet. All I can say is that the third time is teh most promising attempt yet 😆.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -456,14 +462,21 @@ impl EagerSnapshot {
             .flat_map(get_visitor)
             .collect::<Vec<_>>();
 
+        // If files is `None` then this can exit early because the snapshot has intentionally been
+        // loaded _without_ files
+        if self.files.is_none() {
+            self.process_visitors(visitors)?;
+            return Ok(read_data);
+        }
```
roeap, line 470: I think this may actually break things in unexpected ways. Specifically `process_visitors` does nothing without doing a replay. Its main use case right now is idempotent writes (`Txn` actions ...).

However also not sure how this would work with `require_files` right now, we may just have an existing bug ...

rtyler: @roeap anything is possible so sure it might break things but we have no tests to prove otherwise :shrug:

itamarst:
Using `LazySnapshot` when you actually want laziness does sound better, yeah :grin: 

So the main goal here from my perspective is just "allow scaling appends". So a different approach is to maybe... reopen the database on a conflict in "append" mode, so the code sticks to the fast path? Which current abstraction layers may not allow or may just not make sense. I'll take a look later. And I guess worst case there's the hopefully successful refactor.

itamarst:
So looking at the code, the alternative solution of "reopen the database on conflict" doesn't seem to feasible without a bunch of potentially difficult refactoring. The place you'd want to do it is `PreparedCommit` but that just has a dyn reference, and not even necessarily to the table. So there's a bunch of abstraction boundaries in the way.

Issue #3298, opened 2025-03-04
[tracking] Kernelize!


roeap:
> [!IMPORTANT]
> This is a living issue that we'll update with new issues / comments as we get more clarity on the concrete implementations.

# Description

This is a tracking issue to align on and coordinate the integration of `delta-kernel-rs` into `delta-rs`.

## Motivation

While tremendous strides have been made by the community to support more and more delta features in `delta-rs`, we are still lagging behind with more features on the way that user will want to leverage. This is exactly the use case the kernel libraries aim to address - a correct and complete implementation of the Delta protocol.

Kernel explicitly does not take an opinion on all io / execution related aspects that are needed to actually consume / work with delta tables. This is what `delta-rs` provides, leaving the current (high level) user facing APIs conceptually as is. 

## Execution

In simplified terms adopting kernel mean carving out the functionality that currently resides in

- `core/src/kernel` (named so in preparation for being replaced by kernel)
- `core/src/protocol` (mainly our snapshot code, that I wanted to update for quite a while now)
- `core/src/schema` (only partition pruning remained in this module after previous updates)

At the heart of the migration is creating a new snapshot implementation (RFC in  #3137) which provides all required machinery (the engine) to kernel and exposes methods tailored to the needs of `delta-rs`.

One potential avenue forward is to get the RFC merge-ready and merge it without being "hooked-up" to the rest of the crate. This PR also exposes a `Snapshot` trait (we already have something similar, but not quite fitting - I think) the we can hopefully leverage to refactor all the operations that require access to the snapshot - i.e. implement that trait for current snapshots ... This should hopefully surface any missing APIs in kernel that we may yet require for full adoption.

## Challenges

- kernel currently only has very limited write paths support, so we'll have to keep maintaining that for now. However we can motivate the API designs based on our needs.
- In terms of feature support there is no full overlap as of now. e.g. kernel supports deletion vectors, which delta-rs does not, but delta-rs supports generated columns, which are not yet part of kernel and still require some designs (i.e. how to handle arbitrary SQL that an engine will need to parse).
- While kernel offers great opportunities for performance enhancements, there are several areas that might take an initial hit until we can implement performance optimizations that work well with kernel. These mainly relate to less frequently requests actions such as `Txn`, `CommtInfo` ...

Any feedback / concers around proceeding with this is highly appreciated.

ion-elgreco:
@roeap Based on the challenges you wrote, I think we should actually move forward with the current API as 1.0 for us.

[...]

PR #3581, opened 2025-07-08
refactor!: use delta-kernel Protocol and Metadata actions

roeap:
# Description

This PR adopts the first Actions from the delta-kernel crate. Specifically `Protocol` and `Metadata`. These actions being used quite pervasively throughout the codebase, there is quite some amount of code to adopt...

Functionally the main aspect to look out for is that we need to use serde when updating these actions. Given the validation required it is not desirable to directly expose the fields in kernel. So while there certainly is some not so nice code in this PR to work around that, we will over time upstream some of the validation logic and end up with much more streamline code here.

Taking this hit now, as having a singlw representation for P&M greatly simplifies subsequent migrations towards kernel ...

This PR is split in 2 commits - one for each action - that can be reviewed separately.

# Related Issue(s)

related: #3298

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,10 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
-delta_kernel = { version = "0.12.1", features = [
+delta_kernel = { version = "0.13.0", features = [
     "arrow-55",
     "internal-api",
-    # https://github.com/delta-io/delta-kernel-rs/pull/1051
-], git = "https://github.com/roeap/delta-kernel-rs", rev = "687e65920e2520524b8a99dc0ebfab971ba7f619" }
+]}
 
 # arrow
 arrow = { version = "55.2.0" }
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -264,8 +264,8 @@ impl RawDeltaTable {
         })?;
         Ok(RawDeltaTableMetaData {
             id: metadata.id().to_string(),
-            name: metadata.name().cloned(),
-            description: metadata.description().cloned(),
+            name: metadata.name().map(String::from),
+            description: metadata.description().map(String::from),
             partition_columns: metadata.partition_columns().clone(),
             created_time: metadata.created_time(),
             configuration: metadata.configuration().clone(),


ion-elgreco (dismissed):

Cargo.toml
```diff
@@ -26,7 +26,10 @@ debug = true
 debug = "line-tables-only"
 
 [workspace.dependencies]
-delta_kernel = { version = "0.12.1", features = ["arrow-55", "internal-api"] }
+delta_kernel = { version = "0.12.1", features = [
+    "arrow-55",
+    "internal-api",
+], git = "https://github.com/roeap/delta-kernel-rs", rev = "687e65920e2520524b8a99dc0ebfab971ba7f619" }
```
rtyler: please leave a comment to the pr this corresponds to. won't this also make us unreleasable until next kernel release with it?

crates/core/src/kernel/models/actions.rs
```diff
@@ -6,118 +6,132 @@ use delta_kernel::schema::{DataType, StructField};
...
+    let value = serde_json::json!({
+        "id": uuid::Uuid::new_v4().to_string(),
+        "name": None::<String>,
+        "description": None::<String>,
+        "format": { "provider": "parquet", "options": {} },
+        "schemaString": serde_json::to_string(schema)?,
+        "partitionColumns": partition_columns.into_iter().map(|c| c.to_string()).collect::<Vec<_>>(),
+        "configuration": configuration.into_iter().map(|(k, v)| (k.to_string(), v.to_string())).collect::<HashMap<_, _>>(),
+        "createdTime": chrono::Utc::now().timestamp_millis(),
+    });
```
rtyler, line 31: eeeeep! you weren't kidding about some messiness! As far as my sleep-deprived brain can tell this is because the `Matadata` inside of kernel is not exposing any constructors outward? 

Assuming that's the case, would it be viable to add a builder API into kernel for these actions? I imagine this is a useful capability for transaction log utilities and other applications that might want to use kernel

crates/core/src/kernel/models/actions.rs
```diff
@@ -176,6 +252,14 @@ impl Protocol {
         }
     }
 
+    pub(crate) fn from_kernel(value: &Protocol) -> ProtocolInner {
+        serde_json::from_value(serde_json::to_value(value).unwrap()).unwrap()
```
rtyler, line 283: :grimacing: we gotta find a better approach

crates/core/src/kernel/snapshot/parse.rs
```diff
@@ -28,19 +32,20 @@ pub(super) fn read_metadata(batch: &dyn ProvidesColumnByName) -> DeltaResult<Opt
...
-                    format: Default::default(),
-                    schema_string: ex::read_str(schema_string, idx)?.to_string(),
-                    partition_columns: collect_string_list(&partition_columns, idx)
+                let value = json!({
+                    "id": ex::read_str(id, idx)?.to_string(),
+                    "name": ex::read_str_opt(name, idx).map(|s| s.to_string()),
+                    "description": ex::read_str_opt(description, idx).map(|s| s.to_string()),
+                    "format": { "provider": "parquet", "options": {} },
+                    "schemaString": ex::read_str(schema_string, idx)?.to_string(),
+                    "partitionColumns": collect_string_list(&partition_columns, idx)
```
rtyler, line 43: As I'm reading this, let's say we _have_ to keep this `json!` hack in place, could you perhaps leave a comment at each site with a link to a github issue to remove it? That way we can come back along later and find all the invocations to clean up. I worry about missing a spot in the future and us having to chase pretty tricky bugs

crates/core/src/kernel/models/actions.rs
```diff
@@ -134,10 +148,72 @@ pub fn contains_timestampntz<'a>(mut fields: impl Iterator<Item = &'a StructFiel
     fields.any(|f| _check_type(f.data_type()))
 }
 
+pub(crate) trait ProtocolExt {
```
rtyler, line 163: Please please please add some rustdoc here. This is a clever way to shove more functionality on the kernel `Protocol`, but important to document

crates/core/src/kernel/models/actions.rs
```diff
@@ -134,10 +148,72 @@ pub fn contains_timestampntz<'a>(mut fields: impl Iterator<Item = &'a StructFiel
...
+        inner = inner.apply_properties_to_protocol(new_properties, raise_if_not_exists)?;
+        Ok(inner.as_kernel())
+    }
+}
+
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 /// Defines a protocol action
-pub struct Protocol {
+pub(crate) struct ProtocolInner {
```
rtyler, line 242: Will you please add some rustdoc here on the purpose of `ProtocolInner`, I presume it's a temporary shim struct that goes away in the future.

crates/core/src/test_utils/factories/actions.rs
```diff
@@ -103,31 +105,33 @@ impl ActionFactory {
...
-            id: uuid::Uuid::new_v4().hyphenated().to_string(),
-            format: Default::default(),
-            schema_string: serde_json::to_string(schema).unwrap(),
-            partition_columns: partition_columns
-                .map(|i| i.into_iter().map(|c| c.to_string()).collect())
+        let value = json!({
+            "id": uuid::Uuid::new_v4().hyphenated().to_string(),
+            "format": { "provider": "parquet", "options": {} },
+            "schemaString": serde_json::to_string(schema).unwrap(),
+            "partitionColumns": partition_columns
```
rtyler, line 126: :male_detective:

crates/core/src/kernel/models/actions.rs
roeap: > Assuming that's the case, would it be viable to add a builder API into kernel for these actions?

I would go as far as saying this is a requirement :). So why not just contribute them now, and block on this here? These constructors will require quite a bit of validation code as evidenced by the amount of code we have to update Protocol. So my thought was to move forward here and then start contributing the required code upstream.

crates/core/src/kernel/snapshot/parse.rs
roeap: opened https://github.com/delta-io/delta-kernel-rs/issues/1055 to track this and sprkinkled comments in the codebase.

crates/core/src/kernel/models/actions.rs
roeap: added some docs and also directly deprecated the struct (and similar ones). 

Getting rid of all deprecated call sites will be fun after kernel migration :) - there is quite a few by now 😆.

PR #3582, opened 2025-07-08
chore: remove the deltalake-sql crate

rtyler:
This crate has not been published in over a year and is not dependend
upon in any other part of our tree.

I would like to remove it since @alamb and I have had to do non-trivial
mainteannce work to keep this crate up to date with the latest API
changes in Datafusion. This represents unproductive churn IMHO.

I'm doing this now because there's an API break in datafusion main which
broke my internal integration tests.


alamb:
🧹

PR #3584, opened 2025-07-09
feat: write `engineInfo` with delta-rs version

zachschuermann:
# Description
Write `"engineInfo": "delta-rs:0.26.2"` (or whatever current version is) in `commitInfo` action during writes. This will generally assist in debugging to understand what version of delta-rs wrote to tables at different points in time.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/protocol/mod.rs
+++ b/crates/core/src/protocol/mod.rs
@@ -11,6 +11,7 @@ use std::str::FromStr;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
+use crate::crate_version;
 use crate::errors::{DeltaResult, DeltaTableError};
 use crate::kernel::{Add, CommitInfo, Metadata, Protocol, Remove, StructField, TableFeatures};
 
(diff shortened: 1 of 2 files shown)


ion-elgreco:
Hmm we already set it actually here

            app_metadata.insert(
                "clientVersion".to_string(),
                Value::String(format!("delta-rs.{}", crate_version())),

roeap:
@ion-elgreco - i was sure we had it, but somehow missed it. I'll consolidate in a follow-up.

zachschuermann:
hey sorry for missing that! thanks @ion-elgreco!

I was still hoping for this being specifically in `commitInfo.engineInfo` since some existing implementations already parse the string in that field and expose it in e.g. DESCRIBE HISTORY

ion-elgreco:
> hey sorry for missing that! thanks @ion-elgreco!
> 
> I was still hoping for this being specifically in `commitInfo.engineInfo` since some existing implementations already parse the string in that field and expose it in e.g. DESCRIBE HISTORY

We can also keep it there and remove it from appmetadata, but then I hope people are not relying on this info there for other reasons

PR #3590, opened 2025-07-13
refactor: make "cloud" feature in object_store optional

zeevm:
Make "cloud" feature of object_store be optional, if not required then parse_url_opts() doesn't need to take a "RetryConfig"

This removes a lot of dependencies for users who don't need "cloud": ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand", "ring", "http-body-util", "form_urlencoded", "serde_urlencoded"]

This code was already checked in in a previous [PR](https://github.com/delta-io/delta-rs/pull/3398) but for some reason was later modified.


rtyler:
Thanks for the pull request @zeevm , do any ideas come to mind for how we can automate this check? To the best of my understanding `cargo` cannot be told explicitly to _disable_ this cloud feature :thinking:

zeevm:
> Thanks for the pull request @zeevm , do any ideas come to mind for how we can automate this check? To the best of my understanding `cargo` cannot be told explicitly to _disable_ this cloud feature 🤔

@rtyler well it's off by default, so as long as there are tests running with the default set of features on "deltalake-core" then this is already being tested.

rtyler:
I think I remember how this got pulled out. We have some APIs that get exposed to downstream crates which include `RetryConfig` and that will require some wrapping since it causes some problems for multiple API signatures to be floating around. 

I think an enum with field variants is going to be the more sustainable way to ensure this feature can be disabled :thinking:

crates/core/src/logstore/factories.rs
```diff
@@ -25,13 +26,22 @@ pub trait ObjectStoreFactory: Send + Sync {
...
     fn parse_url_opts(
         &self,
         url: &Url,
         options: &HashMap<String, String>,
         retry: &RetryConfig,
         handle: Option<Handle>,
     ) -> DeltaResult<(ObjectStoreRef, Path)>;
+
+    #[cfg(not(feature = "cloud"))]
+    fn parse_url_opts(
```
rtyler: Yeah, this is the part that causes trouble. The trait having multiple implementations of the same method gets hairy.

Once you're done tinkering with this PR, I'll extend it to put this behind a field variant to avoid issues in the future

zeevm: How about something like:

`#[cfg(feature = "cloud")]
pub struct RetryCapsule(RetryConfig);
impl Deref for RetryCapsule {
    type Target = RetryConfig;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
#[cfg(not(feature = "cloud"))]
pub struct RetryCapsule();`

And then:

`pub trait ObjectStoreFactory: Send + Sync {
    fn parse_url_opts(
        &self,
        url: &Url,
        options: &HashMap<String, String>,
        retry: &RetryCapsule,
        handle: Option<Handle>,
    ) -> DeltaResult<(ObjectStoreRef, Path)>;
}`

rtyler: What I am thinking is putting a more generic/extendable enum in place for passing other configurations through in the future too

zeevm: > What I am thinking is putting a more generic/extendable enum in place for passing other configurations through in the future too

Got it.
So how do we handle this, should I complete this PR now and you'll take over for the extendable implementation?

rtyler:
I'll take over the PR, thanks! 
-------- Original Message --------On 7/13/25 12:08, zeevm  wrote:
@zeevm commented on this pull request.



In crates/core/src/logstore/factories.rs:
> +    #[cfg(feature = "cloud")]
     fn parse_url_opts(
         &self,
         url: &Url,
         options: &HashMap<String, String>,
         retry: &RetryConfig,
         handle: Option<Handle>,
     ) -> DeltaResult<(ObjectStoreRef, Path)>;
+
+    #[cfg(not(feature = "cloud"))]
+    fn parse_url_opts(


What I am thinking is putting a more generic/extendable enum in place for passing other configurations through in the future too

Got it.
So how do we handle this, should I complete this PR now and you'll take over for the extendable implementation?

—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>
[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/delta-io/delta-rs/pull/3590#discussion_r2203496527",
"url": "https://github.com/delta-io/delta-rs/pull/3590#discussion_r2203496527",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

ion-elgreco:
@rtyler I agree, we can just remove the `Option<Handle>` from parse_url_opts. If the handle exists it already will be in StorageConfig

Nice refactor :)

ion-elgreco (approved):
Sweeet

PR #3557, opened 2025-06-25
fix: scan time was always 0 for merge metrics

alexwilcoxson-rel:
# Description
Calculates scanning time and write time separately in write_execution_plan_v2 and make available to MergeMetrics

# Related Issue(s)
n/a

# Documentation
n/a

PR #3596, opened 2025-07-15
fix: scan time was always 0 for merge metrics

rtyler:
Since I can't push updates to #3557, this pull request adds a little test but is otherwise identical

Closes #3557


rtyler:
I am going to merge this, I reviewed the original change and this is just a slight addition with some tests

PR #3598, opened 2025-07-15
chore: minor API changes after integration testing

rtyler:
I did some integration testing with the latest `HEAD` of `main` on [oxbow](https://github.com/buoyant-data/oxbow) and found some things lacking.

For reference, oxbow is a useful application for validating the "transaction log application" use-case.


crates/core/src/kernel/mod.rs
```diff
@@ -13,6 +13,7 @@ pub mod schema;
 mod snapshot;
 pub mod transaction;
 
+pub use delta_kernel::engine;
```
roeap, line 16: do we need to expose `engine` to the outside world? Ultimately my hope wpuld be that we can hide the fact that an engine exists from end users. Particularly now, since there are still some footguns with "hidden async" when using engine ...

.github/CODEOWNERS
```diff
@@ -1,13 +1,13 @@
...
+crates/core @roeap @rtyler @hntd187
 crates/aws @rtyler
 crates/lakefs @ion-elgreco
+crates/catalog-glue @rtyler
 crates/catalog-unity @roeap @hntd187
 
 delta-inspect/ @rtyler
 
 proofs/ @houqp
-python/ @wjones127 @roeap @ion-elgreco
```
roeap, line 10: I'm happy to keep receiving pings for python to share the load a bit with @ion-elgreco, even though he knows the python code much better than me by now 😆.

crates/core/src/kernel/mod.rs
rtyler: I don't understand what we gain by _not_ exposing this. I can just as well pull in the engine through Cargo.toml directly. There are a lot of stateless public APIs from kernel that are useful for transaction log applications like oxbow. 

Any of those footguns should not be part of kernel's public API, and that is IMHO a kernel problem

.github/CODEOWNERS
rtyler: With everybody's limited time, I would rather have your attention on the core pull requests only so we can have a tighter feedback cycle :kissing_smiling_eyes: 

This really just controls the auto-defaulted code reviewers, we can still add more reviewers as is necessary :smile:

roeap (approved):

PR #3601, opened 2025-07-18
feat!: kernel log replay

roeap:
# Description

This is the third instalment of trying to move our log replay to leverage delta-kernel. Hoping that this time the pill is just small enough to swallow :).

This introduces a regression in handling checkpoints where only `stats_parsed` is written and the `stats` field is omitted. We can still read these tables, but there will be no file stats-based file skipping during a scan. This needs to be handled [upstream](https://github.com/delta-io/delta-kernel-rs/issues/1075).

There are a few more behaviour changes:
* removed `cdc_files` method, as this one was not used anywhere and did not properly handle replay for cdc files.
* delta-kernel does currently not expose the `dataChange` or `tags` field in add action data, so it is no longer part of the `RecordBatch` returned for `add_action_table`.

There is a lot more room for improvement but the aim for this PR was to try and keep the diff as "small" as possible (not claiming that this is a small diff though 😆). My hope is that we can land this PR and then quickly follow up with some more PRs that do some more focussed refactors.

Of course there are not only the aforementioned regressions, but also new features that are enabled via this PR. In theory v2 checkpoints should just work now, which is something to follow up with quickly and properly test ...

Also, one of the most awaited features - deletion vectors (#1094) - is now in reach, and something I want to address ASAP! 

cc @zachschuermann


ion-elgreco:
@roeap third time is a charm ^^

crates/core/src/kernel/snapshot/stream.rs
```diff
@@ -0,0 +1,178 @@
...
+use tokio::task::JoinSet;
+
+use crate::errors::DeltaResult;
+use crate::kernel::Add;
+use crate::DeltaTableError;
+
+/// Trait for types that stream [RecordBatch]
+///
+/// See [`SendableRecordBatchStream`] for more details.
+pub trait RecordBatchStream: Stream<Item = DeltaResult<RecordBatch>> {
```
ion-elgreco, line 18: I guess same remark as last time, why aren't you using existing traits for this that sit in Datafusion or arrow-rs? :)

roeap: i think these are only in datafusion, however we'll always need them also when there is no datafusion dependency.

Tried looking in arrow, but did not find a stream one ...

ion-elgreco: It does exist in arrow: https://docs.rs/arrow/latest/arrow/array/trait.RecordBatchReader.html but it's an iterator, not sure how stream is much different?

roeap: but thats an iterator, not a stream though?

ion-elgreco: That's true yeah

ion-elgreco:
I'm fine with merging it once the test pass! But it will prevent us releasing the python bindings until those missing features are added

roeap:
@ion-elgreco - just making sure you mean the `stats_parsed` one?

The other changes are likely permanent :). `dataChange` is a property only required in conflict resolution (I think), and the cdc files will (likely) just be replaced by a something `scan_metadata` like, just for CDF ...

That said, I might give upstreaming `stats_parsed` a try to move things along.

ion-elgreco:
> @ion-elgreco - just making sure you mean the `stats_parsed` one?
> 
> The other changes are likely permanent :). `dataChange` is a property only required in conflict resolution (I think), and the cdc files will (likely) just be replaced by a something `scan_metadata` like, just for CDF ...
> 
> That said, I might give upstreaming `stats_parsed` a try to move things along.

Yes the stats parsed mostly. The CDC files indeed was never used, and I understood from Stephen, that kernel has better ways of handling this

DataChange I am not sure tbh what the impact would be, didn't other writers rely on this?

roeap:
> DataChange I am not sure tbh what the impact would be, didn't other writers rely on this?

The only real use case I see for this field is for conflict resolution. In that case though we cannot use the same methods as we do for the replay since we need more actions (txn's, ...) and also the data change field. So we would use the raw action stream from kernels log segment for getting actions in conflict resolution, where ew need the property ...

ion-elgreco:
Now that the other stuff has been merged and released, @roeap @rtyler should we get those final tests passing before we merge or we just merge it as-is?

roeap:
@ion-elgreco - I believe @rtyler investigated a bit and it may be that the failing S3 tests are maybe due to our test setup and not necessarily a bug in our code.

One idea from @rtyler (that I like :)) was to update our APIs to just take URLs (and leave the string parsing to urls up to the caller). Will experiment a bit how much of an impact that would be. 

That said, I feel we somehow need to at some point just get it over with and then focus on improving test coverage for a while as we build out some features.

rtyler:
I am landing this to a new topic branch `next` for us to collaborate in

PR #3605, opened 2025-07-20
refactor: make match_partitions and new_metadata public

zeevm:
# Description
make match_partitions and new_metadata public.

We're writing Metadata "actions" and use match_partitions to do column mapping partition filtering (at least until delta-rs fully supports column mapping)

### DIFF
--- a/crates/core/src/kernel/schema/partitions.rs
+++ b/crates/core/src/kernel/schema/partitions.rs
@@ -158,6 +158,7 @@ impl PartitionFilter {
 
     /// Indicates if one of the DeltaTable partition among the list
     /// matches with the partition filter.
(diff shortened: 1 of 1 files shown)


roeap:
@zeevm - both of these methods are to be removed hopefully soon. As a stop-gap for you, we can make them `pub`, but could you also add `#[deprecated]` to the functions to make it clear that they can get removed in any release ...

roeap (dismissed):
👍

rtyler:
Some of this I had stumbled into for #3598 too, great minds :smiling_imp:

zeevm:
> Some of this I had stumbled into for #3598 too, great minds 😈

@rtyler can we do match_partitions as well?

rtyler (approved):

PR #3611, opened 2025-07-21
feat: add column-mapping and deletion-vectors features

zeevm:
Working on adding support for column-mapping and deletion-vectors

This change adds 2 features "column-mapping" and "deletion-vectors" to enable those in the protocol.

Enabling these features also requires enabling the "experimental" feature for now.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -116,6 +116,10 @@ python = ["arrow/pyarrow"]
 native-tls = ["delta_kernel/default-engine"]
 rustls = ["delta_kernel/default-engine-rustls"]
 cloud = ["object_store/cloud"]
+# marker for enabling experimental features
+experimental = []
+column-mapping = []
+deletion-vectors = []
 
 # enable caching some file I/O operations when scanning delta logs
 delta-cache = ["foyer", "tempfile", "url/serde"]
--- a/crates/core/src/kernel/transaction/protocol.rs
+++ b/crates/core/src/kernel/transaction/protocol.rs
@@ -221,7 +221,10 @@ impl ProtocolChecker {
 pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     let mut reader_features = HashSet::new();
     reader_features.insert(ReaderFeature::TimestampWithoutTimezone);
-    // reader_features.insert(ReaderFeature::ColumnMapping);
+    #[cfg(all(feature="experimental", feature="column-mapping"))]
+    reader_features.insert(ReaderFeature::ColumnMapping);
+    #[cfg(all(feature="experimental", feature="deletion-vectors"))]
+    reader_features.insert(ReaderFeature::DeletionVectors);
 
     let mut writer_features = HashSet::new();
     writer_features.insert(WriterFeature::AppendOnly);
--- a/crates/deltalake/Cargo.toml
+++ b/crates/deltalake/Cargo.toml
@@ -54,6 +54,10 @@ unity-experimental = ["deltalake-catalog-unity"]
 lakefs = ["deltalake-lakefs"]
 native-tls = ["deltalake-core/native-tls"]
 rustls = ["deltalake-core/rustls"]
+# marker for enabling experimental features
+experimental = ["deltalake-core/experimental"]
+column-mapping = ["deltalake-core/column-mapping"]
+deletion-vectors = ["deltalake-core/deletion-vectors"]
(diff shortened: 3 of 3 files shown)


crates/core/Cargo.toml
```diff
@@ -116,6 +116,10 @@ python = ["arrow/pyarrow"]
 native-tls = ["delta_kernel/default-engine"]
 rustls = ["delta_kernel/default-engine-rustls"]
 cloud = ["object_store/cloud"]
+# marker for enabling experimental features
+experimental = []
```
hntd187, line 120: I don't agree with adding features in cargo tied to table features like this. (We don't do it for any other delta features)

crates/core/src/kernel/transaction/protocol.rs
```diff
@@ -221,7 +221,10 @@ impl ProtocolChecker {
 pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     let mut reader_features = HashSet::new();
     reader_features.insert(ReaderFeature::TimestampWithoutTimezone);
-    // reader_features.insert(ReaderFeature::ColumnMapping);
+    #[cfg(all(feature="experimental", feature="column-mapping"))]
```
hntd187, line 224: We also do not support these features even from a reading perspective

hntd187 (changes requested):
Is there a specific problem you are trying to solve with this PR? I don't see a reason to be adding these without the proper support for them.

crates/core/src/kernel/transaction/protocol.rs
zeevm: @hntd187 @ion-elgreco  column mapping is partially supported (missing partition-based filtering based on a mapped column, which I'm currently working on adding). It was public in the past, I have code in production that uses it, at some point someone commented out the feature here.

This breaks my production code. I understand people reluctant to rely on reading column mapping tables with the current incomplete implementation, but that doesn't mean we should break pre-existing code.

The way to go I think is to re-enable it, predicated on a feature flag that's clearly marked as "experimental"

WDYT?

hntd187: Well to be honest, there isn't anything here, so there is nothing to experiment with, it's probably just very very dangerous. I understand the frustration with having the update break your code, but understand that column mapping and deletion vectors are two features that if not properly implemented or implemented at all will lead to table reads being entirely wrong. We took a stance to not allow the reading of table features we don't understand as a measure to prevent users from running into these issues. That having been said kernel does support DVs and column mapping so while the process of that integration is ongoing, when it's further along those features will gain support.

So I think it's better to not do this right now and let the support come through kernel. I'm sorry to suggest stranding you on a prior version, but for the sake of other users not accidentally foot gunning themselves I think this is the better option right now.

zeevm: @hntd187 how would users be foot gunning themselves? They'd have to explicitly enable the "column-mapping" feature, and to make it even more explicit that this is experimental they also have to enable the "experimental" feature.

They'd have to be very very deliberate and cognizant of using this experimental feature, this isn't just some pitfall they can stumble upon.

We can also add a compile time warning when enabling this if you deem it necessary.

The delta-kernel integration is a very lengthy process with no clear end in sight unfortunately, and while I'll be more than happy to move to this implementation when it makes it to production, I really need to maintain the previously publicly released column mapping capability.

Without it, our generally available enterprise cloud service will have to drop delta-rs and move to another implementation, I'd really hate for that to happen.

rtyler:
@zeevm these features being present but not actively utilized by delta-rs is not an experimental flag I am willing to incorporate for general use. This will effectively guarantee data inconsistencies or corruption. 

Allowing these features to be used while not _actually_ supporting the functionality would also mean that delta-rs would _not_ be complying with the protocol. That is not an acceptable situation for the project. While I agree it is frustrating that we still don't have support for these features, delta-rs is properly adherent to the protocol for table features and writer/reader versions.

For your customers' use-cases, if you're willing to take these risks then I would recommend building against an internal fork with this change _or_ I can help produce proprietary LTS builds or forks via [Buoyant Data](https://buoyantdata.com) which I have done for some other enterprises.

zeevm:
Other than supporting partitioned tables using a mapped column (which I'll be implementing in delta-rs), what else do you think is missing for _read only_ support of column mapping?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #3613, opened 2025-07-22
fix: Allow non-string primitive types for partition filters when converting to pyarrow dataset

### LINKED ISSUE #3597: TypeError in `DeltaTable.to_pyarrow_dataset` when using non-string partition filter values (e.g., int)
# Environment

**Delta-rs version**: Delta-rs version: >= 0.25.2

**Binding**:  Python

**Environment**: 
- **Cloud provider**:
- **OS**: macOs
- **Other**:

***
# Bug

**What happened**: 
When calling `DeltaTable.to_pyarrow_dataset()` with partition filters that include integers (e.g., `("year", "=", 2024)`), the following error is raised:

    TypeError: argument 'partition_filters': failed to extract enum PartitionFilterValue ('Single | Multiple')
    - variant Single (Single): TypeError: failed to extract field PartitionFilterValue::Single.0, caused by TypeError: 'int' object cannot be converted to 'PyString'
    - variant Multiple (Multiple): TypeError: failed to extract field PartitionFilterValue::Multiple.0, caused by TypeError: 'int' object cannot be converted to 'Sequence'

while it clearly says that 

Partition filters should accept primitive types (`int`, `str`, `bool`, `date`) as the value and internally convert them to strings if necessary.

**How to reproduce it**:

    dt = DeltaTable("s3://my-bucket/my-table")

    dt.to_pyarrow_dataset(partitions=[
        ("year", "=", 2024),         # int triggers error
        ("week", "=", 11),           # int triggers error
        ("date", "=", "2024-03-10"), # str works fine
    ])

**More details**:

- deltalake: 1.1.0  
- Python: 3.10+  
- OS: macOS

smeyerre:
# Description
Previously when calling `DeltaTable.to_pyarrow_dataset()` with partition filters that included non-string types an error was raised, eg:
```
TypeError: argument 'partition_filters': failed to extract enum PartitionFilterValue ('Single | Multiple')
- variant Single (Single): TypeError: failed to extract field PartitionFilterValue::Single.0, caused by TypeError: 'int' object cannot be converted to 'PyString'
- variant Multiple (Multiple): TypeError: failed to extract field PartitionFilterValue::Multiple.0, caused by TypeError: 'int' object cannot be converted to 'Sequence'
```
Now, acceptable primitive types get converted to strings internally and no longer raise errors:
```
>>> dt = DeltaTable("../crates/test/tests/data/partition-type-primitives")
>>> result_int = dt.to_pyarrow_dataset(partitions=[("year", "=", 2020)])
>>> print(result_int.to_table().to_pydict())
{'id': [8, 2, 1, 5, 7, 4, 3, 6], 'value': ['record_8', 'record_2', 'record_1', 'record_5', 'record_7', 'record_4', 'record_3', 'record_6'], 'year': ['2020', '2020', '2020', '2020', '2020', '2020', '2020', '2020'], 'is_active': ['false', 'true', 'true', 'false', 'false', 'true', 'true', 'false'], 'event_date': ['2023-01-02', '2023-01-01', '2023-01-01', '2023-01-01', '2023-01-02', '2023-01-02', '2023-01-02', '2023-01-01'], 'category': ['B', 'B', 'A', 'A', 'A', 'B', 'A', 'B']}
```

# Related Issue(s)
- closes #3597 

# Documentation
- https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.to_pyarrow_dataset


rtyler (approved):

FrankPortman:
Hey - I am still investigating a regression from 1.1.3 on my end, but is there a chance there is a bug here? I noticed the test cases do not cover "in" at all (only "=").

I am hitting this:

```python
val = ['false']

    def encode_partition_value(val: Any) -> str:
        # Rules based on: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization
        if isinstance(val, bool):
            return str(val).lower()
        if isinstance(val, str):
            return val
        elif isinstance(val, (int, float)):
            return str(val)
        elif isinstance(val, datetime):
            return val.isoformat(sep=" ")
        elif isinstance(val, date):
            return val.isoformat()
        elif isinstance(val, bytes):
            return val.decode("unicode_escape", "backslashreplace")
        else:
>           raise ValueError(f"Could not encode partition value for type: {val}")
E           ValueError: Could not encode partition value for type: ['false']
```

ion-elgreco:
@smeyerre Could you follow up on this to fix this regression once you have time? Because we had support before for `PartitionFilterValue(Vec<PyBackedStr>>)`

smeyerre:
> @smeyerre Could you follow up on this to fix this regression once you have time? Because we had support before for `PartitionFilterValue(Vec<PyBackedStr>>)`

Yep I'll look👍

PR #3616, opened 2025-07-23
fix: ensure new checkpoints can be written after old checkpoints

rtyler:
This should also make delta-kernel-rs#1018 redundant

I have ongoing discussions with @zachschuermann on how to remove this sharp edge in kernel nonetheless


zachschuermann:
LGTM - will follow up kernel side, thanks @rtyler!

ion-elgreco (approved):

PR #3618, opened 2025-07-24
fix: allow writing to DeltaTable objects across Python threads

### LINKED ISSUE #3594: Python DeltaTable does not support writes in multiple threads (again?)
# Environment

**Delta-rs version**: >= 0.25.2

**Binding**: Python

**Environment**:
- **Cloud provider**:
- **OS**: macOS
- **Other**:

***
# Bug

**What happened**: 
I am facing the same issue that was found in #2958. With the versions after 0.25.1, the error still occurs.

See #2958.

**How to reproduce it**:
See #2958.

**More details**:
See #2958.

rtyler:
FREEZE!

Looks like we were _really_ close to full multi-threaded support before,
and just needed to sprinkle a little macro magic on the pyclass
definition.

See [pyo3 docs](https://pyo3.rs/v0.23.0/class/thread-safety)

Fixes #3594

### DIFF
--- a/python/deltalake/writer/_utils.py
+++ b/python/deltalake/writer/_utils.py
@@ -33,7 +33,7 @@ def try_get_table_and_table_uri(
         table_uri = str(table_or_uri)
     else:
         table = table_or_uri
-        table_uri = table._table.table_uri()
+        table_uri = table.table_uri
 
     return (table, table_uri)
 
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -100,7 +100,7 @@ enum PartitionFilterValue {
     Multiple(Vec<PyBackedStr>),
 }
 
-#[pyclass(module = "deltalake._internal")]
+#[pyclass(module = "deltalake._internal", frozen)]
 struct RawDeltaTable {
     /// The internal reference to the table is guarded by a Mutex to allow for re-using the same
     /// [DeltaTable] instance across multiple Python threads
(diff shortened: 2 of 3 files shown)


ion-elgreco:
@rtyler nice stuff, didn't know about Frozen but makes total sense. The struct properties are only updated within rust and not python

I don't think Polars allows the DataFrame to be references across a threadpoolexecutor, only processpoolexecutor with `spawn`

rtyler:
> I don't think Polars allows the DataFrame to be references across a threadpoolexecutor, only processpoolexecutor with `spawn`

There is no cross thread dataframe access here, this is entirely about our DeltaTable

:😕

ion-elgreco:
> > I don't think Polars allows the DataFrame to be references across a threadpoolexecutor, only processpoolexecutor with `spawn`
> 
> There is no cross thread dataframe access here, this is entirely about our DeltaTable
> 
> :😕

The multi thread test is using a Polars Dataframe, but it causes this error:

```
       return self._df.__arrow_c_stream__(requested_schema)
E       RuntimeError: Already borrowed

.venv/lib/python3.9/site-packages/polars/dataframe/frame.py:1466: RuntimeError
```

rtyler:
I don't know why that was failing in CI, it didn't fail for me locally. A new run also hasn't failed :thinking: :shrug:

ion-elgreco (approved):

PR #3619, opened 2025-07-24
fix: ensure openssl-sys doesn't creep into the dependency via the kernel default engine

rtyler:


bendu:
I have been trying to use native-tls instead of rustls with deltalake crate. It seems like this change basically flips the default must-have dependency from one tls implementation to the other. Would it make more sense to remove the default-engine out of the workspace Cargo.toml altogether?

I think something like this would help reduce unnecessary crates while keeping the rustls default. https://github.com/bendu/delta-rs/commit/3c7e30d032a304f72cd3880707aba4fea3ae1d29

PR #3623, opened 2025-07-28
fix: coerce polars.Array into a suitable Arrow list type

### LINKED ISSUE #3566: Python deltalake 1.0.2 is not compatible with polars.Array
# Environment

**Delta-rs version**: 1.0.2

**Binding**: python

**Environment**:
- **Cloud provider**: None
- **OS**: macOS
- **Other**:

***
# Bug

**What happened**:
When using polars dataframe with `pl.Array` column, `write_deltalake` raises an error.

> Exception has occurred: ValueError
column types must match schema types, expected List(Field { name: "element", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) but found List(Field { name: "item", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) at column index 0

`write_deltalake` implicitly cast to deltalake schema and successfully write data to table with no errors.

**How to reproduce it**:

```python
import polars as pl

df = pl.DataFrame(
    {"array": [[-5, -4], [0, 0], [5, 9]]}, schema={"array": pl.Array(pl.Int32, 2)}
)
DeltaTable.create(
    "example_table",
    df.to_arrow().schema,
    mode="overwrite",
)
write_deltalake(
    "example_table",
    df,
    mode="overwrite",
)
```

**More details**:
The code works in deltalake 0.25.5

rtyler:
Ideally polars wouldn't be giving us `item` as the list field name, but
it's more important to Just Work :tm: than be pedantic about these
things

Fixes #3566


ion-elgreco (approved):

crates/core/src/kernel/schema/cast/mod.rs
```diff
@@ -104,6 +104,24 @@ fn cast_field(
                 add_missing,
             )?) as ArrayRef)
         }
+        (DataType::FixedSizeList(_, _), DataType::FixedSizeList(child_fields, _)) => {
```
alamb, line 107: 👍  this totally makes sense -- it would also be great to port this stuff upstream to DataFusion

I think @kosiew has been working towards improving DataFusion's built in cast support for these cases

PR #3642, opened 2025-07-29
chore: use pytest-xdist for speeding up python tests

rtyler:
The additional benefit of running these tests in parallel is that more
racey/timing related test failures are cropping up for me.

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -70,17 +70,17 @@ jobs:
         run: make develop
 
       - name: Run tests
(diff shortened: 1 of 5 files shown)


rtyler:
these fail sporadically during repeated testing :thinking: 


```
FAILED tests/test_threaded.py::test_multithreaded_write_using_path - RuntimeError: Already borrowed
FAILED tests/test_table_read.py::test_read_simple_in_threads_and_processes[ProcessPoolExecutor-spawn-False] - tests.test_table_read._SerializableException: Seraializatble exception: assert [7, 5, 9] == [5, 7, 9]
```

PR #3646, opened 2025-08-02
chore!: api pruning

roeap:
# Description

stacked on #3601

This PR updates/deprecates/prunes a bund of APIs on our core Table/State/Snapshot abstractions. This is follow up from the main log-replay PR where we tried to keep the changes limited to what was strictly necessary to make kernel log replay work.

This is also a first step in being more explicit about what our internal representations actually represent. Currently `DeltaTable` may or may not have an initialised snapshot internally. As a result all methods that effectively inspect the snapshot (metadata, schema,..) are fallible or optional. This PR removes/deprecates most of these APIs, which should be accessed by accessing the respective method on the snapshot.


rtyler:
no-op

PR #3647, opened 2025-08-02
refactor!: use kernel TableProperties

roeap:
# Description

This PR is stacked on #3646

Replaces usage of our internal `TableConfig` with kernels `TableProperties`.


rtyler:
@roeap what version of the previous pull request did you use here? it looks like a squashed commit which is why there are so many conflicts :thinking:

roeap:
@rtyler - i squashed the original kernel one a bit to make DCO etc. work, then all based off of that.

PR #3654, opened 2025-08-04
feat: support converting parquet with non-microsecond timestamps to d…

### LINKED ISSUE #3535: `convert_to_deltalake` cannot convert parquet dataset if it has millisecond-precision timestamps
# Environment

**Delta-rs version**: 1.0.2

**Binding**: python

***
# Bug

**What happened**: parquet dataset created with millisecond-precision timestamps cannot be converted to a delta table. Same error message as in https://github.com/delta-io/delta-rs/issues/1019: `SchemaMismatchError: Invalid data type for Delta Lake: Timestamp(Millisecond, Some("UTC"))`

**How to reproduce it**:

```python
import pandas as pd
import deltalake

df = pd.DataFrame({"column1": [0]*2, "column2": [pd.to_datetime("2020-01-01T00:00:00Z"), pd.to_datetime("2020-01-02T00:00:00Z")]})
df["column2"] = df["column2"].dt.as_unit("ms")
df.to_parquet("test", partition_cols=["column1"], index=False, engine="pyarrow", compression="zstd", compression_level=3)
deltalake.convert_to_deltalake(uri="test", partition_by=deltalake.Schema(fields=[deltalake.Field("column1", "integer")]), partition_strategy="hive")
```

however creating the dataset as a delta table from the beginning works:
```python
deltalake.write_deltalake("test1", df, partition_by=["column1"], writer_properties=deltalake.WriterProperties(compression="ZSTD", compression_level=3), mode="append")
```

smeyerre:
…eltalake

# Description
Delta Lake only supports microsecond timestamp precision. Previously when attempting to convert a parquet table with non-microsecond precision to a delta table `SchemaMismatchError: Invalid data type for Delta Lake: Timestamp(...)` would be raised.

Now this can be done successfully and the timestamp is converted to microseconds:
```
>>> import pandas as pd
>>> import deltalake
>>> df = pd.DataFrame({"column1": [0]*2, "column2": [pd.to_datetime("2020-01-01T00:00:00Z"), pd.to_datetime("2020-01-02T00:00:00Z")]})
>>> df["column2"] = df["column2"].dt.as_unit("ms")
>>> df.to_parquet("test", partition_cols=["column1"], index=False, engine="pyarrow", compression="zstd", compression_level=3)
>>> deltalake.convert_to_deltalake(uri="test", partition_by=deltalake.Schema(fields=[deltalake.Field("column1", "integer")]), partition_strategy="hive")
>>> dt = deltalake.DeltaTable("test")
>>> df_converted = dt.to_pandas()
>>> print(df_converted)
                    column2  column1
0 2020-01-01 00:00:00+00:00        0
1 2020-01-02 00:00:00+00:00        0
>>> print(dt.schema())
Schema([Field(column2, PrimitiveType("timestamp"), nullable=True), Field(column1, PrimitiveType("integer"), nullable=True)])
```

# Related Issue(s)
- closes #3535 

# Documentation
- N/A

### DIFF
--- a/crates/core/src/operations/convert_to_delta.rs
+++ b/crates/core/src/operations/convert_to_delta.rs
@@ -71,8 +71,10 @@ fn convert_field_timestamps(field: &ArrowField) -> Result<ArrowField, ArrowError
     };
 
     if let Some(data_type) = converted_data_type {
-        Ok(ArrowField::new(field.name(), data_type, field.is_nullable())
-            .with_metadata(field.metadata().clone()))
+        Ok(
+            ArrowField::new(field.name(), data_type, field.is_nullable())
+                .with_metadata(field.metadata().clone()),
+        )
     } else {
         Ok(field.clone())
     }
--- a/crates/core/src/operations/convert_to_delta.rs
+++ b/crates/core/src/operations/convert_to_delta.rs
@@ -5,7 +5,10 @@ use std::num::TryFromIntError;
 use std::str::{FromStr, Utf8Error};
 use std::sync::Arc;
 
-use arrow_schema::{ArrowError, Schema as ArrowSchema};
+use arrow_schema::{
+    ArrowError, DataType as ArrowDataType, Field as ArrowField, Fields, Schema as ArrowSchema,
+    TimeUnit,
+};
 use delta_kernel::engine::arrow_conversion::TryIntoKernel as _;
 use delta_kernel::schema::StructType;
 use futures::future::{self, BoxFuture};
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/convert_to_delta.rs
```diff
@@ -33,6 +36,75 @@ use crate::{
     DeltaResult, DeltaTable, DeltaTableError, ObjectStoreError, NULL_PARTITION_VALUE_DATA_PATH,
 };
 
+fn convert_timestamps_to_microseconds(schema: ArrowSchema) -> Result<ArrowSchema, ArrowError> {
```
smeyerre, line 39: I wasn't 100% sure where to put these conversion functions. So happy to move them if anyone has a sense they should be elsewhere.

detoxifiedplant:
@smeyerre does it only change the type to `TimeUnit::Microsecond` or it convert the value itself to equivalent Microsecond values?

smeyerre:
> @smeyerre does it only change the type to `TimeUnit::Microsecond` or it convert the value itself to equivalent Microsecond values?

Good catch, this only changes the type. To update the values I think we'll need to rewrite the parquet files if necessary. I'll fix that.

ion-elgreco:
> > @smeyerre does it only change the type to `TimeUnit::Microsecond` or it convert the value itself to equivalent Microsecond values?
> 
> Good catch, this only changes the type. To update the values I think we'll need to rewrite the parquet files if necessary. I'll fix that.

I also wondered this, especially whether it's actually required. A lot of readers they cast on read with the actual table schema

smeyerre:
> > > @smeyerre does it only change the type to `TimeUnit::Microsecond` or it convert the value itself to equivalent Microsecond values?
> > 
> > 
> > Good catch, this only changes the type. To update the values I think we'll need to rewrite the parquet files if necessary. I'll fix that.
> 
> I also wondered this, especially whether it's actually required. A lot of readers they cast on read with the actual table schema

Ok interesting. So I originally went with this implementation because I was trying to mirror what https://github.com/delta-io/delta does. I have an implementation of the value conversion as well rewriting the parquet, but I can think of a couple questions about whether to add it or not:
- Do we want to prioritize mirroring the Scala project or divert?
- Is the performance impact of rewriting the files worth it?

I'm not sure I have enough context to confidently know the answers to those

ion-elgreco:
@smeyerre I am guessing we can get a way without casting :) so lets just try it

crates/core/src/operations/convert_to_delta.rs
ion-elgreco: We have a casting module, but this seems very specific to parquets moving into delta, so fine to keep it here

crates/core/src/operations/convert_to_delta.rs
```diff
@@ -33,6 +36,75 @@ use crate::{
...
+    for field in schema.fields() {
+        if let Some(converted_field) = convert_field_timestamps(field)? {
+            any_changed = true;
+            converted_fields.push(converted_field);
+        } else {
+            converted_fields.push(field.as_ref().clone());
+        }
+    }
+
+    if any_changed {
```
ion-elgreco: keeping track of any_changed seems a bit overcomplicated, we can just always return the convert_fields schema

crates/core/src/operations/convert_to_delta.rs
```diff
@@ -33,6 +36,75 @@ use crate::{
...
+        Ok(ArrowSchema::new(converted_fields))
+    } else {
+        Ok(schema)
+    }
+}
+
+/// Recursively convert timestamp fields to microseconds
+fn convert_field_timestamps(field: &ArrowField) -> Result<Option<ArrowField>, ArrowError> {
```
ion-elgreco: Why don't we just return the original field here?

crates/core/src/operations/convert_to_delta.rs
```diff
@@ -33,6 +36,48 @@ use crate::{
...
+fn convert_field_timestamps(field: &ArrowField) -> Result<ArrowField, ArrowError> {
+    let converted_data_type = match field.data_type() {
+        ArrowDataType::Timestamp(TimeUnit::Nanosecond, tz)
+        | ArrowDataType::Timestamp(TimeUnit::Millisecond, tz)
+        | ArrowDataType::Timestamp(TimeUnit::Second, tz) => {
+            Some(ArrowDataType::Timestamp(TimeUnit::Microsecond, tz.clone()))
+        }
+        // Recursively handle nested structures
+        ArrowDataType::Struct(fields) => {
+            let converted_fields = fields
```
smeyerre, line 59: We're missing maybe a slight optimization here now as we're always creating a new struct/list but I'd think it should be negligible for the schema fields

ion-elgreco: Which optimization?

ion-elgreco (approved):
You are on a roll! @smeyerre 

Keep it up 🤘🚀

detoxifiedplant:
> You are on a roll! @smeyerre 
> 
> Keep it up 🤘🚀

@ion-elgreco @rtyler can you answer the following questions
- does the PR only change the data type to Micros or convert the value to Micros while casting it(i.e. if the source is Millis, it should be converted to equivalent value to Micros before storing).
- Reader of the delta table should be able to read by its data type represented in the schema (Micros).

crates/core/src/operations/convert_to_delta.rs
smeyerre: Oh previously I had it so that we only created a new Struct when there were converted fields, and now we're always recreating the whole struct/list no matter what

smeyerre:
> > You are on a roll! @smeyerre
> > Keep it up 🤘🚀
> 
> @ion-elgreco @rtyler can you answer the following questions
> 
> * does the PR only change the data type to Micros or convert the value to Micros while casting it(i.e. if the source is Millis, it should be converted to equivalent value to Micros before storing).
> * Reader of the delta table should be able to read by its data type represented in the schema (Micros).

This PR means that we convert non-microsecond columns to micros in the **delta lake schema** but not the underlying parquet column values.

@ion-elgreco can correct me if I'm wrong, but I think a reader of the table would generally then see the parquet metadata as well as the delta table schema and then appropriately cast the parquet column value to microseconds on read.

ion-elgreco:
@smeyerre correct! In our case we cast the read recordbatches, and other engines do this as well if they can coerce. We could probably have added a python test just in case

PR #3655, opened 2025-08-04
feat: Encode stats for Binary and FixedLengthBinary

HawaiianSpork:
# Description

This now writes binary stats as escaped string sequences according the delta lake protocol instead of logging a warning. The encoding logic was all there, just needed to call it add add tests.

# Related Issue(s)


rtyler:
@HawaiianSpork when you refer to this being defined in the protocol, are you referring to the [partition value serialization](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization) specification? I'm not sure what min/max values make sense for binary values :confused:

HawaiianSpork:
@rtyler maybe it isn't explicit but I was referring to the protocol says that at least for binary partition identification should be identified by an escaped string https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization

`Encoded as a string of escaped binary values. For example, "\u0001\u0002\u0003"`

We would find it useful for filter on hashes.

rtyler (dismissed):

PR #3660, opened 2025-08-05
feat!: kernel log replay

roeap:
# Description

Kernel based log replay.


ion-elgreco (dismissed):

rtyler:
I want us to make more improvements, but we can do that on `main`. Before the next Python release I will want to make sure we include some performance improvements like [this](https://github.com/delta-io/delta-kernel-rs/pull/1171) from delta-kernel-rs, even if only referring to a merged `main` rather than formal release

Issue #3577, opened 2025-07-07
Partition value strings containing reserved ASCII and non-ASCII are double-encoded.


kdn36:
# Environment

**Delta-rs version**: 1.0.2

**Binding**: use of delta-rs via polars `write_delta`

**Environment**:
- **Cloud provider**: n/a
- **OS**: Ubuntu 22.04.5 LTS
- **Other**: 

***
# Bug

**What happened**:
When writing deltalake files with partition columns containing strings with reserved ASCII or non-ASCII characters, some characters get double-encoded, while a leading or trailing `/` will get dropped. Dependent services may use `k=v` convention to recover the partition values, expecting single-encoded values.

The root cause is here:
https://github.com/delta-io/delta-rs/blob/3a5d18008893e4875a5e5c2ad1155270390f4324/crates/core/src/kernel/scalars.rs#L89
where both `encode` and `Path::from` perform encoding, albeit on different character sets. In addition, `Path::from` will drop leading or trailing `/`.

Note, this may or may not be a bug, as the deltalake standard states that "Actual partition values for a file must be read from the transaction log." But it is unexpected and undesirable.

See related issue in polars:
https://github.com/pola-rs/polars/issues/23390
which has some additional details.

**What you expected to happen**:
Single encoding based on the superset of invalid characters.

**How to reproduce it**:
```python
import polars as pl
from pathlib import Path
from datetime import datetime

from deltalake import DeltaTable
from polars.testing import assert_frame_equal



# df = pl.DataFrame({"a": [10,20,30], "strings": ["aé", "b?", "c!", "d/"]})

(start, stop) = (32, 127)
df = pl.DataFrame(
    {
        "a": list(i for i in range(start, stop)),
        "strings": list('a' + chr(i) for i in range(start, stop)),
    }
)
print(f"df:\n{df}")

# local folder, make sure it's empty
partitioned_tbl_uri = Path("./tmp_mre/partitioned_table").resolve()

# write:
df.write_delta(partitioned_tbl_uri, delta_write_options={"partition_by": "strings"})

# read:
partitioned_tbl = DeltaTable(partitioned_tbl_uri)
pl_df_partitioned = (
    pl.read_delta(str(partitioned_tbl_uri))
    .sort("a")
)
print(f"pl_df_partitioned:\n{pl_df_partitioned}")

assert_frame_equal(df, pl_df_partitioned)
```

gives:
```
df:
shape: (95, 2)
┌─────┬─────────┐
[...]

PR #3661, opened 2025-08-07
fix: use RFC3896 percent encoding with delta protocol correctness

ion-elgreco:
# Description
Before when I addressed partition encoding I was using the URL encoding, where I naively assumed this was URI encoding compatible, but protocol states to use RFC3896 with additional stricter rules. I have used the percent-encoding-rfc3896 crate for this which forks the percent-encoding crate but adds some extra things to make it it RFC3896 compatible.

The ASCII set contains all reserved characters that need to be encoded according to the protocol

- fixes: #3577

@kdn36

### DIFF
--- a/crates/core/src/kernel/scalars.rs
+++ b/crates/core/src/kernel/scalars.rs
@@ -9,10 +9,46 @@ use delta_kernel::{
     expressions::{Scalar, StructData},
     schema::StructField,
 };
-use object_store::path::Path;
+
+use percent_encoding_rfc3986::{utf8_percent_encode, AsciiSet, CONTROLS};
 #[cfg(any(test, feature = "integration_test"))]
 use serde_json::Value;
-use urlencoding::encode;
+
+// ASCII set that needs to be encoded
+const RFC3986_PART: &AsciiSet = &CONTROLS
+    .add(b' ') // space
+    .add(b'!')
+    .add(b'"')
+    .add(b'#')
+    .add(b'$')
+    .add(b'%')
+    .add(b'&')
+    .add(b'\'')
+    .add(b'(')
+    .add(b')')
+    .add(b'*')
+    .add(b'+')
+    .add(b',')
+    .add(b'/')
+    .add(b':')
+    .add(b';')
+    .add(b'<')
+    .add(b'=')
+    .add(b'>')
+    .add(b'?')
+    .add(b'@')
+    .add(b'[')
+    .add(b'\\')
+    .add(b']')
+    .add(b'^')
+    .add(b'`')
+    .add(b'{')
+    .add(b'|')
+    .add(b'}');
+
+fn encode_partition_value(value: &str) -> String {
+    utf8_percent_encode(value, RFC3986_PART).to_string()
+}
 
 use crate::NULL_PARTITION_VALUE_DATA_PATH;
 
@@ -86,7 +122,7 @@ impl ScalarExt for Scalar {
         if self.is_null() {
             return NULL_PARTITION_VALUE_DATA_PATH.to_string();
         }
-        encode(Path::from(self.serialize()).as_ref()).to_string()
+        encode_partition_value(self.serialize().as_str()).to_string()
     }
 
     /// Create a [`Scalar`] from a row in an arrow array.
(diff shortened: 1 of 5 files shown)


python/tests/test_partition_encoding_rfc3896.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    print(f"df:\n{df}")
+
+    # write:
+    df.write_delta(tmp_path, delta_write_options={"partition_by": "strings"})
+
+    # read:
+    partitioned_tbl = DeltaTable(tmp_path)
+    pl_df_partitioned = pl.read_delta(partitioned_tbl).sort("a")
+    print(f"pl_df_partitioned:\n{pl_df_partitioned}")
```
roeap: still needed?

python/tests/test_partition_encoding_rfc3896.py
```diff
@@ -0,0 +1,31 @@
...
+
+from deltalake import DeltaTable
+
+
+@pytest.mark.polars
+def test_partition_encoding_rfc3896(tmp_path):
+    import polars as pl
+    from polars.testing import assert_frame_equal
+
+    (start, stop) = (32, 127)
```
roeap: does this range correspond to the characters we need to encode?

crates/core/src/kernel/scalars.rs
```diff
@@ -86,7 +122,7 @@ impl ScalarExt for Scalar {
         if self.is_null() {
             return NULL_PARTITION_VALUE_DATA_PATH.to_string();
         }
-        encode(Path::from(self.serialize()).as_ref()).to_string()
+        encode_partition_value(self.serialize().as_str()).to_string()
```
roeap: hmm, doesn't `encode_partition_value` already return a `String`?

crates/core/src/kernel/scalars.rs
```diff
@@ -9,10 +9,46 @@ use delta_kernel::{
...
 };
-use object_store::path::Path;
+
+use percent_encoding_rfc3986::{utf8_percent_encode, AsciiSet, CONTROLS};
 #[cfg(any(test, feature = "integration_test"))]
 use serde_json::Value;
-use urlencoding::encode;
+
+// ASCII set that needs to be encoded
+const RFC3986_PART: &AsciiSet = &CONTROLS
```
roeap, line 18: nit: would it be more legible to put all characters in an array and then do something like `fold(...)`?

crates/core/src/kernel/scalars.rs
```diff
@@ -9,10 +9,46 @@ use delta_kernel::{
     expressions::{Scalar, StructData},
     schema::StructField,
 };
-use object_store::path::Path;
+
```
roeap: should there be a blank line in this import block?

roeap (commented):
Just some minor comments.

Would be great to add a link somewhere to the protocol for future reference.

python/tests/test_partition_encoding_rfc3896.py
ion-elgreco: It was only the printable characters of ASCII, but I've extended the test range with the extendable ASCII characters

python/tests/test_partition_encoding_rfc3896.py
ion-elgreco: Nope :) was just copy pasta from the issue

crates/core/src/kernel/scalars.rs
ion-elgreco: I tried, but const or lazylock doesn't like folding over a slice of bytes, maybe you can get some working code for this?

ion-elgreco: @roeap shall we merge it now, we can always optimize later, I want to land this before we merge the log replay stuff

roeap (approved):
👍

crates/core/src/kernel/scalars.rs
roeap: @ion-elgreco - yes :).

kdn36:
Great to see this, thanks a lot. I will test it into polars soon.

One clarifying question on an edge case. Given that `Path` removes leading and trailing slashes, will the encoding be such that any hive partition strings containing a trailing or leading `/` encode correctly and they will not be removed?

ion-elgreco:
@kdn36 trailing slashes will also be encoded yes, you should use the same ASCII set as we use to decode the partitions, since delta protocol partition serialization is more strict than RFC3896

kdn36:
I can confirm that the local tests on polars now pass (using release 1.1.4).

Thanks a lot for your support.

PR #3669, opened 2025-08-10
feat!: use kernel predicates on file streams

roeap:
# Description

Another bigger refactor moving us closer to a world where we process everything as data.

The three commits can (almost :)) be reviewed separately and do the following.

1. Our file view iterators were overly complex and had ownership semantics hard to manage on streams (i.e. borrowed record batch). We simplify the implementation (now called `LogicalFileView`) with an owned record batch (cheap to clone)
2. Kernel has quite powerful file skipping build in. Thus far we allowed for iterating over files based on partition filters along with custom logic for filtering files. While we retain (for now) our own internal partition predicate definition where is is currently used, we do now translate this directly to a kernel predicate and leverage kernels file skipping.
3. To harmonise between snapshots we now expose all file-like methods via async/streams. Doing this I realised that going forward, we may be able to leverage `LogDataHandler` more consequently and avoid having extra traits that cover `DeltaTableState` and `EagerSnapshot`.

Sorry that this one also got so large, but since we are still updating core APIs, the surface area is quite large :(.


crates/core/src/kernel/transaction/conflict_checker.rs
```diff
@@ -203,21 +198,22 @@ impl<'a> TransactionInfo<'a> {
...
                         source: Box::new(err),
-                    },
-                )?,
+                    })?,
             ))
         } else {
-            Ok(Either::Right(std::iter::empty()))
+            Ok(Either::Right(
+                self.read_snapshot.iter().map(|f| f.add_action()),
+            ))
```
roeap, line 209: This may have been a bug in our conflict resolution. in the non-datafusion case, we always return all files to have been read from the table.

The previous logic would make sense if we guarantee that an appropriate predicate is always set when we read from the table (i.e. const eval to `true`), however I could find no place where we would be able to do so. As such it is at least more conservative to always return all files in the df case when there is no read predicate.

As we adopt more kernel tools also for conflict resolution, I plan to revisit all our logic, since we can do file skipping also in the non-df case.

crates/core/src/kernel/schema/partitions.rs
```diff
@@ -369,10 +255,20 @@ fn filter_to_kernel_predicate(
     })
 }
 
+fn parse_scalar(dt: &PrimitiveType, raw: &str) -> DeltaResult<Scalar> {
+    match dt {
+        // NOTE: in partition encoding, empty strings are interpreted as NULL,
+        // which is the lesser of two evils ...
```
ion-elgreco: Why? Empty string is still a string

crates/core/src/kernel/snapshot/iterators.rs
```diff
@@ -0,0 +1,413 @@
...
+                arr.is_valid(self.index)
+                    .then(|| match Scalar::from_array(arr, self.index) {
+                        Some(Scalar::Struct(s)) => Some(s),
+                        _ => None,
+                    })
+                    .flatten()
+            })
+    }
+
+    fn partition_values_map(&self) -> HashMap<String, Option<String>> {
```
ion-elgreco, line 183: Why do we have the StructData version as well?

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -79,9 +82,15 @@ impl Snapshot {
     ) -> DeltaResult<Self> {
         // TODO: bundle operation_id with logstore ...
         let engine = log_store.engine(None);
-        let table_root = log_store.table_root_url();
+        let mut table_root = log_store.table_root_url();
         let version = version.map(|v| v as u64);
 
+        // NB: kernel engine uses Url::join to construct paths,
+        // if the path does not end with a slash, the would override the ebtire path.
```
ion-elgreco:
```suggestion
        // if the path does not end with a slash, the would override the entire path.
```
:)

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -529,24 +559,100 @@ impl EagerSnapshot {
...
+        // a stream just like on the Snapshot. So we swallow the awkward error handling for now
+        // knowing that we will be able to clean this up soon (TM).
+        let data = if let Some(predicate) = predicate {
+            let scan = match self
+                .snapshot
+                .inner
+                .clone()
+                .scan_builder()
+                .with_predicate(predicate)
+                .build()
```
ion-elgreco, line 591: Is this where you mentioned to use Datafusion instead?

crates/core/src/kernel/arrow/engine_ext.rs
```diff
@@ -504,6 +504,28 @@ impl<T: ExpressionEvaluator + ?Sized> ExpressionEvaluatorExt for T {
     }
 }
 
+pub trait StructDataExt {
```
rtyler, line 532: You're killin me here with the lack of rustdocs :grimacing: 

I understand what this trait is but I don't understand why it must exist and when it can go away

 I'm going to add a CI check :smiling_imp:

rtyler (changes requested):
I have rebased this branch to make the review easier.

I will help add more into this pull request, but `iterators.rs` has a _lot_ of new code very little rustdoc or test coverage.

crates/core/src/kernel/schema/partitions.rs
rtyler: I don't disagree :smile: but the [protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization) specifically asks for: 

>   An empty string for any type translates to a null partition value.

I will update the comment to link to the documentation

rtyler: While writing a test case for this, I discovered that [this is already handled in kernel](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/expressions/scalars.rs#L568-L570) and so this function is not necessary

roeap: i think we still need this since on this code path kernel is not yet involved. Specifically we are translating our `PartionFilter` into a kernel predicate and must decide when to choose a NULL scalar.

roeap: i was wrong - we do use kernels parsing after all :D

crates/core/src/kernel/snapshot/iterators.rs
roeap: Right now we mainly need this when we create add / remove actions from the data. THis is a provate function though, only the sruct one is in the public API.

crates/core/src/kernel/arrow/engine_ext.rs
roeap: I provosioned a little help and added more RustDocs ... I'll keep working on myself 😆

crates/core/src/kernel/snapshot/mod.rs
roeap: sort of :D

PR #3675, opened 2025-08-12
feat: provide delta writer option to flush buffer after every batch

### LINKED ISSUE #3578: Refactor delta writer to not use buffer
# Environment

**Delta-rs version**: `1.0.2`

**Binding**: Python (`deltalake`)

**Environment**:
- **Cloud provider**: AWS (S3 tested)
- **OS**: Ubuntu 22.04

---

# Bug

**What happened**:

When using `write_deltalake()` with a generator, memory usage grows continuously during the write process. This happens both locally and when writing to S3, suggesting batches are not being released or flushed as expected.

I expected memory usage to remain stable or plateau over time, as only one batch is yielded at a time by the generator. Each batch is 100,000 rows, and there are 1,000 batches in total. The generator should allow batch-wise memory usage, not cumulative growth.

**How to reproduce it**:

```python
import os
import psutil
import pyarrow as pa

def log_memory():
    print(f"Memory usage: {process.memory_info().rss / 1024**2:.2f} MB")

schema = pa.schema([("id", pa.int64()), ("value", pa.string())])
nbatch = 1_000
batch_size = 100_000

def generate_batches():
    for i in range(nbatch):
        log_memory()
        batch = pa.record_batch(
            {"id": [i * 10 + j for j in range(batch_size)],
             "value": [f"value_{j}" for j in range(batch_size)]},
            schema=schema,
        )
        yield batch

reader = pa.RecordBatchReader.from_batches(schema, generate_batches())

write_deltalake(
    "./test",
    reader,
    mode="overwrite",
)
```
**More details**:
- The issue is not tied to local storage; we see the same behavior when writing to S3 using the AWS bindings.
- Memory usage (via psutil and CloudWatch) steadily increases batch-by-batch, even though the generator should not accumulate state.
-  It appears that write_deltalake() is buffering or retaining batches in memory instead of processing and releasing them incrementally.

smeyerre:
# Description
This provides the option to the delta writer to flush the in-memory buffer to disk after each record batch as opposed to waiting for the targeted file size by passing the `flush_per_batch` parameter. This prevents accumulating lots of memory in some cases.

Note: I believe there is still some accumulated memory usage through things like transaction metadata.

# Related Issue(s)
- closes #3578 

# Documentation
- N/A

### DIFF
--- a/crates/core/src/operations/write/mod.rs
+++ b/crates/core/src/operations/write/mod.rs
@@ -266,7 +266,7 @@ impl WriteBuilder {
         self.write_batch_size = Some(write_batch_size);
         self
     }
- 
+
     /// If true, write in-memory buffer to disk after each batch
     pub fn with_flush_per_batch(mut self, flush_per_batch: bool) -> Self {
         self.flush_per_batch = flush_per_batch;
@@ -2008,19 +2008,17 @@ mod tests {
         fn test_write_builder_flush_per_batch() {
             let ops = DeltaOps::new_in_memory();
             let log_store = ops.write(vec![]).log_store().clone();
-            
+
             // Default
             let builder = WriteBuilder::new(log_store.clone(), None);
             assert!(!builder.flush_per_batch);
-            
+
             // True
-            let builder = WriteBuilder::new(log_store.clone(), None)
-                .with_flush_per_batch(true);
+            let builder = WriteBuilder::new(log_store.clone(), None).with_flush_per_batch(true);
             assert!(builder.flush_per_batch);
-            
+
             // False
-            let builder = WriteBuilder::new(log_store.clone(), None)
-                .with_flush_per_batch(false);
+            let builder = WriteBuilder::new(log_store.clone(), None).with_flush_per_batch(false);
             assert!(!builder.flush_per_batch);
         }
     }
--- a/crates/core/src/operations/write/writer.rs
+++ b/crates/core/src/operations/write/writer.rs
@@ -561,7 +561,8 @@ mod tests {
         let batch = get_record_batch(None, false);
 
         // write single un-partitioned batch
-        let mut writer = get_partition_writer(object_store.clone(), &batch, None, None, None, false);
+        let mut writer =
+            get_partition_writer(object_store.clone(), &batch, None, None, None, false);
         writer.write(&batch).await.unwrap();
         let files = list(object_store.as_ref(), None).await.unwrap();
         assert_eq!(files.len(), 0);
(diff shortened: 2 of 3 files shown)


python/src/lib.rs
```diff
@@ -1699,6 +1702,11 @@ impl RawDeltaTable {
             }
 
             if let Some(writer_props) = writer_properties {
+                // Extract write_batch_size and set it on WriteBuilder because WriterConfig::new()
+                // does not extract write_batch_size from WriterProperties
+                if let Some(write_batch_size) = writer_props.write_batch_size {
+                    builder = builder.with_write_batch_size(write_batch_size);
+                }
```
smeyerre, line 1709: This feels weird to me, but I wasn't sure how much I wanted to change the `write_deltalake` function signature here since we provide this `write_batch_size` setting in the WriterProperties as opposed to a standalone parameter like `target_file_size`.

smeyerre:
Oops also realized I had pyarrow in my local uv env so that failing test passed for me locally but not here

ion-elgreco (commented):
This has the side effect of creating a lot of small files, right?

I was thinking, couldn't we do the flushing more tied together to file-writing, so the flush arrow writer becomes aware of the amount of flushed data.


Typing this on a phone so hope it's clear:

Write_batch --> bytes --> flush to disk --> check if flushed data is gt file_size_limit, if so close multipart and start new multipart

The only thing is the last part can only be smaller if you close the multipart write, so you have to make sure the writer only flushes full parts, and somehow becomes aware to check if the last part was the final part.

smeyerre:
> This has the side effect of creating a lot of small files, right?
> 
> I was thinking, couldn't we do the flushing more tied together to file-writing, so the flush arrow writer becomes aware of the amount of flushed data.
> 
> Typing this on a phone so hope it's clear:
> 
> Write_batch --> bytes --> flush to disk --> check if flushed data is gt file_size_limit, if so close multipart and start new multipart
> 
> The only thing is the last part can only be smaller if you close the multipart write, so you have to make sure the writer only flushes full parts, and somehow becomes aware to check if the last part was the final part.

Yes this does currently write many more small files.

Ahhh ok I see, yeah that makes a lot of sense. Ok I'll work on that, thanks!

smeyerre:
@ion-elgreco Ok follow up question about this:

Currently `create_add()` (https://github.com/delta-io/delta-rs/blob/2920177ac5215e192e0182bed93c42c0b4a98b6f/crates/core/src/operations/write/writer.rs#L433-L444) expects complete metadata from `writer.close()`, but now I think we have to call `writer.close()` multiple times per file to get buffer data. Currently the way I'm handling this is I'm just using the metadata from the first flush for a file and accumulating row counts, but we're losing out on column statistics for the whole set of batches (just using the stats from the first batch) which doesn't seem great.

Another option I can think of is merging the stats from each flush and rebuilding the whole metadata at the end but that seems pretty intense for this.

I'm also seeing this which seems potentially helpful: https://github.com/delta-io/delta-rs/blob/2920177ac5215e192e0182bed93c42c0b4a98b6f/crates/core/src/writer/stats.rs#L436C1-L489C2 

Just wanted to run this past you and see if you had an opinion on the best course of action, thanks!

ion-elgreco:
I think we should avoid calling AsyncArrowWriter.close() before reaching the max file size.

Reset_writer currently returns the old buffer and writet and creates new buffer and writer but that should perhaps change 🤔

PR #3678, opened 2025-08-17
feat: domain metadata read support 

### LINKED ISSUE #3360: delta-rs fails to open a table with domainMetadata action in checkpoint
# Environment

**Delta-rs version**: 0.25

**Binding**: 

**Environment**:
- **Cloud provider**:
- **OS**:
- **Other**:

***
# Bug

**What happened**: I am unable to open a Delta table with a domainMetadata (with liquid clustering) action in its latest checkpoint. The exact error message:
Delta protocol violation: Invalid action field: Unexpected action from checkpoint: domainMetadata

**What you expected to happen**: I expect the table to open successfully

**How to reproduce it**: I've included a zip file containing a delta_log directory with just the single checkpoint file. I can include also the data files, but they are not necessary to reproduce the error I'm facing

**More details**:

[_delta_log.zip](https://github.com/user-attachments/files/19531177/_delta_log.zip)

roeap:
# Description

closes #3360

This PR is stacked on #3669 - to review just look at the last commit.

We are adding read support for domain metadata. Domain metadata writes are not yet supported.


crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -655,6 +674,14 @@ impl EagerSnapshot {
             .application_transaction_version(log_store, app_id.to_string())
             .await
     }
+
+    pub async fn domain_metadata(
```
rtyler, line 678: Because `EagerSnapshot` exposes `snapshot()` I think this convenience method should probably just be removed. It's a layer of indirection that I don't think is worth the extra API surface area

rtyler (approved):

rtyler:
Once this has merged, I will release 0.28.1

crates/core/src/kernel/snapshot/mod.rs
roeap: In principle I do agree ... but the hope is that we can get rid of one of the snapshots once we have proper caching setup.. i.e. we would keep the snapshot data in a configurable cache rather than on the struct. This is why I try to keep the Snapshot APis more or less in-sync.

That said, its probably not too different of a refactor if we remove it.

PR #3695, opened 2025-08-22
fix(pandas): implement-automatic-conversion-for-pandas-null-types

### LINKED ISSUE #3691: Python: Automatically convert Pandas null types to valid Delta Lake types in write_deltalake()
# Description

When writing to an existing DeltaTable, `deltalake` could automatically convert `null` column types in Pandas dataframes. These column types arise in Pandas when a column in a dataframe is completely null, and a more specific schema has not been provided. `write_deltalake()` already converts column types upon write when it is able to, but skips the `null` case.

**Related Issue(s)**

https://github.com/delta-io/delta-rs/issues/686 which has implemented some more general functionality, but missed this case. In particular see this discussion onward: https://github.com/delta-io/delta-rs/issues/686#issuecomment-3144593598

fvaleye:
# Description
This PR fixes type inconsistencies in Pandas DataFrames by implementing automatic conversion of `object` dtype columns containing null values into their correct schema-defined types (when available).

Previously, `object` dtype columns with `null` values could remain untyped, leading to mismatches with schema-defined types and causing validation or serialization errors.

Now, these columns are automatically converted, ensuring consistency between DataFrame values and the schema while reducing type-related issues.

# Related Issue(s)
- closes #3691

### DIFF
--- a/python/deltalake/writer/_conversion.py
+++ b/python/deltalake/writer/_conversion.py
@@ -23,14 +23,15 @@ def _convert_arro3_schema_to_delta(
     def dtype_to_delta_dtype(
         dtype: DataType, field_name: str | None = None
     ) -> DataType:
-        if DataType.is_null(dtype) and existing_schema and field_name:
-            for existing_field in existing_schema:  # type: ignore[attr-defined]
-                if existing_field.name == field_name:
-                    # Prevent infinite recursion: if existing field is also null, keep as null
-                    if DataType.is_null(existing_field.type):
-                        return dtype
-                    return dtype_to_delta_dtype(existing_field.type, None)
-            return dtype
+        if DataType.is_null(dtype) and existing_schema is not None and field_name is not None:
+            try:
+                existing_field = existing_schema.field(field_name)
+                # Prevent infinite recursion: if existing field is also null, keep as null
+                if DataType.is_null(existing_field.type):
+                    return dtype
+                return dtype_to_delta_dtype(existing_field.type, None)
+            except (KeyError, IndexError):
+                return dtype
 
         # Handle nested types
         if (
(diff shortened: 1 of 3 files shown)


FrankPortman:
Thanks for jumping on this so quickly!

python/deltalake/writer/_conversion.py
```diff
@@ -6,14 +6,32 @@
...
+    Returns:
+        Arro3Schema: Delta-compatible schema with converted types.
     """
 
-    def dtype_to_delta_dtype(dtype: DataType) -> DataType:
+    def dtype_to_delta_dtype(
+        dtype: DataType, field_name: str | None = None
+    ) -> DataType:
+        if DataType.is_null(dtype) and existing_schema and field_name:
+            for existing_field in existing_schema:  # type: ignore[attr-defined]
```
ion-elgreco: We can use the field getter instead here: https://kylebarron.dev/arro3/v0.6.1/api/core/schema/#arro3.core.Schema.field

fvaleye: Oh, good catch, thanks!

python/deltalake/writer/_conversion.py
```diff
@@ -6,14 +6,33 @@
...
+
+    Returns:
+        Arro3Schema: Delta-compatible schema with converted types.
     """
 
-    def dtype_to_delta_dtype(dtype: DataType) -> DataType:
+    def dtype_to_delta_dtype(
+        dtype: DataType, field_name: str | None = None
+    ) -> DataType:
+        if DataType.is_null(dtype) and existing_schema and field_name:
```
ion-elgreco: I prefer we also check here if is not None, it's s bit more clear. Rest of code lgtm!

fvaleye: Do you mean:
`        if dtype and DataType.is_null(dtype) and existing_schema and field_name:
`

ion-elgreco: No, rather field_,name is not None and schema is not none

fvaleye: I see!
Thanks for your review @ion-elgreco 🙏 
Let's merge this after this change!

PR #3710, opened 2025-08-26
docs: add Sail integration

shehabgamin:
# Description
This adds an integration page for using Delta Lake with Sail 🚀

### DIFF
--- a/docs/integrations/delta-lake-sail.md
+++ b/docs/integrations/delta-lake-sail.md
@@ -1,8 +1,14 @@
 # Using Delta Lake with Sail
 
-[Sail](https://github.com/lakehq/sail) is an open-source multimodal distributed compute framework, built in Rust, unifying batch, streaming, and AI workloads. For seamless adoption, Sail offers a drop-in replacement for the Spark SQL and DataFrame APIs in both single-host and distributed settings. You can use Sail for both ad-hoc queries on your laptop and large-scale data processing jobs in a cluster.
+[Sail](https://github.com/lakehq/sail) is an open-source multimodal distributed compute framework, built in Rust,
+unifying batch, streaming, and AI workloads. For seamless adoption, Sail offers a drop-in replacement for the Spark SQL
+and DataFrame APIs in both single-host and distributed settings. You can use Sail for both ad-hoc queries on your laptop
+and large-scale data processing jobs in a cluster.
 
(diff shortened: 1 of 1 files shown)


shehabgamin:
@rtyler Just checking in here, let me know if anything else is needed :)

rtyler:
I'm sorry for taking so long to respond on this, I had been waiting to see how the attribution on copying code would work out.

I have taken another look at the Lake  Sail code which copies non-trivial portions of the `deltalake-core` crate. Fortunately now it has been at least attributed properly.

Sail [does not integrate with delta-rs](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/Cargo.toml), but does build on top of kernel with some subslices of our code.

I wish y'all the best of luck.

shehabgamin:
> I'm sorry for taking so long to respond on this, I had been waiting to see how the attribution on copying code would work out.
> 
> 
> 
> I have taken another look at the Lake  Sail code which copies non-trivial portions of the `deltalake-core` crate. Fortunately now it has been at least attributed properly.
> 
> 
> 
> Sail [does not integrate with delta-rs](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/Cargo.toml), but does build on top of kernel with some subslices of our code.
> 
> 
> 
> I wish y'all the best of luck.

@rtyler 

Over four months ago, when I opened this PR, Sail did use delta-rs. That's no longer the case.

On attribution: Sail has always credited upstream work. See [this PR](https://github.com/lakehq/sail/pull/1054), which documents the Linux Foundation outreach. A screenshot was provided to the LF as “evidence” of missing attribution, but that same screenshot clearly shows our attribution in the code.

What’s disappointing here is that this concern was raised at all, especially given that LakeSail employees (myself included) have contributed to delta-kernel-rs, and we’ve also tried to contribute to delta-rs directly (e.g., https://github.com/delta-io/delta-rs/pull/3933).

So while I appreciate you taking another look and acknowledging that the attribution has been present, I want to be very clear: we were attributing from day one. The suggestion that this only "fortunately” worked out after the fact doesn’t reflect what actually happened.


We’ve also done substantial independent work in Sail and later seen similar ideas show up in delta-rs as well. We’ve never viewed that as a problem, because it’s normal for open-source projects to learn from each other. Against that backdrop, it’s hard not to find the attribution complaint a bit ironic.

I wish you the best of luck.

PR #3713, opened 2025-08-28
chore(cargo): add cargo-machete to detect and remove unused dependencies

fvaleye:
# Description
Adding `cargo machete` to find and remove unused dependencies quickly in the CI.
The cargo machete has been run to identify and remove some unused dependencies.

A review should be done to ensure that flagged dependencies are truly unused and to avoid false positives.

# Documentation
- https://github.com/bnjbvr/cargo-machete

### DIFF
--- a/crates/azure/Cargo.toml
+++ b/crates/azure/Cargo.toml
@@ -15,14 +15,10 @@ rust-version.workspace = true
 deltalake-core = { version = "0.28.0", path = "../core", features = ["cloud"] }
 
 # workspace depenndecies
-async-trait = { workspace = true }
 bytes = { workspace = true }
-futures = { workspace = true }
-tracing = { workspace = true }
 object_store = { workspace = true, features = ["azure"] }
 thiserror = { workspace = true }
 tokio = { workspace = true }
-regex = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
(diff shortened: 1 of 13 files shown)


crates/azure/Cargo.toml
```diff
@@ -15,14 +15,10 @@ rust-version.workspace = true
 deltalake-core = { version = "0.28.0", path = "../core", features = ["cloud"] }
 
 # workspace depenndecies
-async-trait = { workspace = true }
 bytes = { workspace = true }
-futures = { workspace = true }
-tracing = { workspace = true }
```
roeap, line 21: I think here @rtyler might say that adding logging to the crates is the way to go and every crate should have tracing as a required dependency :D.

That said, I am fairly sure we already have an issue to track this.

crates/core/Cargo.toml
```diff
@@ -166,3 +166,6 @@ required-features = ["datafusion"]
 [[test]]
 name = "read_delta_partitions_test"
 required-features = ["datafusion"]
+
+[package.metadata.cargo-machete]
+ignored = ["foyer"]
```
roeap, line 171: we can remove this for now I think.

I believe this may have slipped though while i was working on #3440. It will get added back once we merge that PR, but then potentially also as optional.

PR #3717, opened 2025-08-29
feat(storage): expand user with tilde in local path

### LINKED ISSUE #1948: Support ~ for write_deltalake and read operations
# Description

It would be nice if this library supported `~` for I/O operations.  This code errors out:

```python
df = pd.DataFrame({"num": [1, 2, 3], "letter": ["a", "b", "c"]})
write_deltalake("~/tmp/cool/some-table", df)
```

Here is the error message: OSError: Encountered object with invalid path: Error parsing Path "/Users/matthew.powers/Documents/code/delta/delta-examples/notebooks/python-deltalake/~/tmp/cool/some-table": Encountered illegal character sequence "~" whilst parsing path segment "~"

Polars provides a better user experience IMO:

```python
import polars as pl

df = pl.DataFrame({"num": [1, 2, 3], "letter": ["a", "b", "c"]})
df.write_delta("~/tmp/cool/some-table")
```

I often have to write code like this when working with deltalake which is bulky:

```python
from pathlib import Path

path = f"{Path.home()}/data/deltalake_baseline_G1_1e9_1e2_0_0"
dt = DeltaTable(path)
```

fvaleye:
# Description
Support expanding `~` in local paths for DeltaTable storage.

# Related Issue(s)
- closes #1948

# Documentation
- Added [dirs](https://crates.io/crates/dirs) crate


python/tests/test_writer.py
```diff
@@ -2503,3 +2503,26 @@ def test_polars_write_array(tmp_path: pathlib.Path):
...
+
+    unique_id = str(uuid.uuid4())[:8]
+    tilde_path = f"~/tmp_delta_test_{unique_id}"
+
+    try:
+        df.write_delta(tilde_path)
+        dt = DeltaTable(tilde_path)
+        assert dt.to_pandas().equals(df.to_pandas())
+    finally:
+        import shutil
```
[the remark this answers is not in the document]
fvaleye: Thanks for your feedback, dear Copilot, but it's a test - it should be fine to import only when it fails.

hntd187:
I was worried us poor windows users would have been left out here, but turns out the dirs crate handles it! :)

python/tests/test_writer.py
```diff
@@ -2503,3 +2503,27 @@ def test_polars_write_array(tmp_path: pathlib.Path):
...
+
+    df = pl.DataFrame({"num": [1, 2, 3], "letter": ["a", "b", "c"]})
+
+    unique_id = str(uuid.uuid4())[:8]
+    tilde_path = f"~/tmp_delta_test_{unique_id}"
+
+    try:
+        df.write_delta(tilde_path)
+        dt = DeltaTable(tilde_path)
+        assert dt.to_pandas().equals(df.to_pandas())
```
ion-elgreco: You can use the Polars.testing module here

fvaleye: Oh, yes, it would be better only to use `@pytest.mark.polars`. Thanks!

ion-elgreco (approved):

PR #3728, opened 2025-09-02
chore(ci): cache rust dependencies in the CI

fvaleye:
# Description
Integrate `rust-cache` to persist Rust build artifacts and dependencies across CI runs, removing the need for repeated downloads and recompilations of unchanged dependencies.

# Documentation
- https://github.com/Swatinem/rust-cache

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,13 @@ jobs:
           toolchain: "1.85"
           override: true
 
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: "rust-cache-${{ matrix.os }}"
+          env-vars: "CARGO CC CFLAGS CXX CMAKE RUST"
+          cache-all-crates: true
+          cache-targets: false
+
       - name: Format
         run: cargo fmt -- --check
 
@@ -87,6 +94,13 @@ jobs:
           toolchain: "1.85"
           override: true
 
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: "rust-cache-${{ matrix.os }}"
+          env-vars: "CARGO CC CFLAGS CXX CMAKE RUST"
+          cache-all-crates: true
+          cache-targets: false
+
       - name: Run tests
         run: |
           make setup-dat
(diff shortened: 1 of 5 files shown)


rtyler:
@fvaleye How does this work with disk space in GitHub Actions? Part of the
reason I ripped out a bunch of caching last year (I think?) was that we have a
finite amount of space in GitHub Actions' cache and we would periodically get
failed builds because the disk space was exceeded.

I am happy to bring caching back, but we have to make sure it can be
"auto-managed" and not require admin privileges to go delete caches when too much space is being used.

fvaleye:
> @fvaleye How does this work with disk space in GitHub Actions? Part of the reason I ripped out a bunch of caching last year (I think?) was that we have a finite amount of space in GitHub Actions' cache and we would periodically get failed builds because the disk space was exceeded.
> 
> I am happy to bring caching back, but we have to make sure it can be "auto-managed" and not require admin privileges to go delete caches when too much space is being used.

Oh, I didn't know that it was considered as failing jobs, thanks @rtyler 
Normally, caches are currently limited to 10 GB in total.
And as described [here](https://github.com/Swatinem/rust-cache?tab=readme-ov-file#cache-limits-and-control), " exceeding that limit will cause eviction of older caches.".

I would expect to have "auto-evictions" rather than build failures... 
I could try adapting to apply cache only to the `main` branch as a first attempt?

**[edited]**
In hindsight, since our Rust crate dependencies are heavy, they might fail... even with a simple build. Let me check how we could leverage this caching.

.github/workflows/build.yml
```diff
@@ -42,6 +42,13 @@ jobs:
           toolchain: "1.85"
           override: true
 
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: "rust-cache-${{ matrix.os }}"
+          env-vars: "CARGO CC CFLAGS CXX CMAKE RUST"
+          cache-all-crates: true
+          cache-targets: false
```
rtyler, line 50: This being set to `false` I think makes this worth trying. The `target/` directory would bust the cache regularly. A single build of `core` with datafusion produces over 3GB of `.target/` data on my machine :sob:

rtyler (approved):
I'm curious how this will or will not play well with sccache :shrug: 

But worth trying IMO

PR #3732, opened 2025-09-03
feat: get the delta table row count based on the table history

### LINKED ISSUE #3731: Get the table row count based on the table history
# Description
Currently, delta-rs does not have an option to get the number of rows for a specific table.
I want to get the number of rows from the delta log without reading the whole table to the RAM of my application.
The idea is to handle a row counter and iterate over the log in order to get the number.

**Use Case**
number_of_rows = DeltaTable().count()

**Related Issue(s)**

ohadmata:
# Description
Currently, delta-rs does not have an option to get the number of rows for a specific table.
I want to get the number of rows from the delta log without reading the whole table to the RAM of my application.
The idea is to handle a row counter and iterate over the log in order to get the number.

# Related Issue(s)
[- closes #3731](https://github.com/delta-io/delta-rs/issues/3731)

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -528,6 +528,31 @@ class DeltaTable:
             history.append(commit)
         return history
 
+    def count(self) -> int:
+        """
+        Get the row count based on the history of the DeltaTable.
+
+        Returns:
+            The number of rows for this specific table
+        """
+        total_rows = 0
+
+        for entry in sorted(self.history(), key=lambda x: x["version"]):
+            operation, metrics, parameters = (
+                entry.get("operation", "").upper(), entry.get("operationMetrics", {}),
+                entry.get("operationParameters", {})
+            )
+
+            if operation in ("WRITE", "COPY INTO", "CREATE TABLE AS SELECT", "REPLACE TABLE AS SELECT"):
+                mode = parameters.get("mode", "").upper()
+                affected_rows = int(metrics.get("numOutputRows") or metrics.get("num_added_rows"))
+                total_rows = total_rows + affected_rows if mode == "APPEND" else affected_rows
+
+            elif operation in ("DELETE", "TRUNCATE"):
+                total_rows -= int(metrics.get("numDeletedRows"))
+
+        return total_rows
+
     def vacuum(
         self,
         retention_hours: int | None = None,
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -530,15 +530,24 @@ class DeltaTable:
 
     def count(self) -> int:
         """
-        Get the row count based on the history of the DeltaTable.
+        Get the approximate row count based on file statistics added to the Delta table.
+
+        This requires that add actions have been added to the Delta table with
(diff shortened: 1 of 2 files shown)


ion-elgreco:
Thanks for your PR @ohadmata, unfortunately this wouldn't be a stable interface to get the row count since operations and metrics are optional and not agreed upon in delta spec.

You can however sum all the numRecords of each file add action.

rtyler:
@ion-elgreco :shrug: this seems like a kind of useful thing to have right? If I put an interface for counting file stats down in the :crab: layer, would that be cool to expose upwards?

ion-elgreco:
> @ion-elgreco :shrug: this seems like a kind of useful thing to have right? If I put an interface for counting file stats down in the :crab: layer, would that be cool to expose upwards?

Definitely! As long as its done based on the file stats

ohadmata:
@ion-elgreco @rtyler I have change the function to get the count from the parquet files metadata. WDYT?

ion-elgreco:
> @ion-elgreco @rtyler I have change the function to get the count from the parquet files metadata. WDYT?

This won't work on a remote object storage. The most efficient and correct way is to do it base on the delta log stats field "numRecords"

ohadmata:
@ion-elgreco I updated it again to retrieve the number of records from the add action in the Delta log. This approach is much more efficient and avoids opening all the Parquet files.

python/deltalake/table.py
```diff
@@ -528,6 +528,20 @@ def _backwards_enumerate(
...
+    def count(self) -> int:
+        """
+        Get the row count based on the history of the DeltaTable.
+
+        Returns:
+            The number of rows for this specific table
+        """
+        total_rows = 0
+
+        for value in self.get_add_actions().column("num_records").to_pylist():
```
fvaleye, line 545: I recommend leveraging PyArrow’s compute module to avoid Python for loops when working with large tables.

You could exit early if there is no data.
```
    import pyarrow.compute as pc
    import pyarrow as pa
    
    add_actions = self.get_add_actions()
    if add_actions.num_rows == 0:
        return 0 # early exit
    
    # Conversion
    num_records = pa.array(add_actions.column("num_records").to_pylist())
    
    total = pc.sum(pc.fill_null(num_records, 0)) # Use PyArrow
    
    return int(total.as_py()) if total.is_valid else 0
```

For the sake of codebase quality, could you please create a unit test to cover null/None data and a basic count scenario?

ion-elgreco: Pyarrow is not a required dependency anymore :)

I think this is fine for now, but this should be changed once we can have DataFusion directly on the transaction log

fvaleye: Ok 👍 
Thanks @ion-elgreco

ion-elgreco:
@ohadmata can you add a small test? Then we are good to go!

rtyler:
@ion-elgreco I have added a regression test and some improved docstrings, I think this is ready to go!

ohadmata:
Hi, I just got back from vacation and saw that you added a test and merged it. Thanks!

PR #3747, opened 2025-09-09
fix: `write_deltalake` with `mode="overwrite"` mode and `schema_mode=None` does not overwrite schema metadata

### LINKED ISSUE #3744: `write_deltalake` with `schema_mode=None` will overwrite `nullable` properties of columns
# Environment

**Delta-rs version**: 1.1.4

**Binding**:  Python

**Environment**: 
- **Cloud provider**: Azure/Local
- **OS**: Mac/Linux
- **Other**:

***
# Bug

**What happened**: Instantiating a Delta table with a specific schema, including `nullable=False` columns, and later writing to it via `write_deltalake` with `mode="overwrite"`, `schema_mode=None`, and `predicate=.<some predicate to emulate replaceWhere>`, leads to overwriting the nullability properties of the columns.

**What you expected to happen**: Coercing all column safely to the types that match the existing Delta table, including adhering to existing nullability properties. In the case where all columns and dtypes from the to-be-written data already match the Delta table on disk, the nullability of the to-be-written data shouldn't even be considered. It should only validate data via the Delta protocol based on the existing schema if the existing schema says some columns are not nullable.

**How to reproduce it**:  https://gist.github.com/FrankPortman/4a967f5bcb0bf5e7136eff5087ed6880

**More details**:

I went back and forth on whether this should be filed as a bug or not, but the behavior IMO is unexpected, and it behaves differently from Spark (setting no schema mode, attempts all safe column coercions to existing types but doesn't override nullability). Also this is in the Rust code `/// whether to overwrite the schema or to merge it. None means to fail on schmema drift`, and IMO failing on schema drift does not implicitly mean to overwrite column nullability.

FrankPortman:
# Description

`write_deltalake` with `mode="overwrite"` and `schema_mode=None` was overwriting the nullability constraints of various columns in the schema, instead of accepting the existing schema fully (or failing on drift, if there was actually schema drift besides column constraints).

Upon reflection of the `merge` operation, it turned out that worked as intended, but this PR still introduces a test to ensure the schema behavior is consistent between the two.

# Related Issue(s)

Closes #3744

# Documentation
 
N/A

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -1811,6 +1811,108 @@ mod tests {
         assert_merge(table, metrics).await;
     }
     #[tokio::test]
+    async fn test_merge_preserves_nullability_without_schema_merge() {
+        // Test that nullability constraints are preserved when merge_schema is false (default)
+        let delta_schema = vec![
+            StructField::new(
+                "id".to_string(),
+                DataType::Primitive(PrimitiveType::String),
+                false, // non-nullable
+            ),
+            StructField::new(
+                "value".to_string(),
+                DataType::Primitive(PrimitiveType::Integer),
+                false, // non-nullable
+            ),
+            StructField::new(
+                "modified".to_string(),
+                DataType::Primitive(PrimitiveType::String),
+                true, // nullable
+            ),
+        ];
+
+        let table = DeltaOps::new_in_memory()
+            .create()
+            .with_save_mode(SaveMode::ErrorIfExists)
+            .with_columns(delta_schema)
+            .await
+            .unwrap();
+
+        // Verify initial schema nullability
+        let initial_fields: Vec<_> = table
+            .snapshot()
+            .unwrap()
+            .schema()
+            .fields()
+            .cloned()
+            .collect();
+        assert!(
+            !initial_fields[0].is_nullable(),
+            "id should be non-nullable"
+        );
+        assert!(
+            !initial_fields[1].is_nullable(),
+            "value should be non-nullable"
+        );
+        assert!(
+            initial_fields[2].is_nullable(),
+            "modified should be nullable"
+        );
+
+        // Source data with all nullable fields (typical from external sources)
+        let source_schema = Arc::new(ArrowSchema::new(vec![
+            Field::new("id", ArrowDataType::Utf8, true), // nullable in source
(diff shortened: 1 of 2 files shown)


FrankPortman:
Do we feel covered by existing tests for the following cases:

* `schema_mode=None` and schema drift detected ---> should fail with informative error
* `schema_mode=None` and `mode="append"` ---> should *enforce* constraints that exist, not simply accept them
* `schema_mode=None` and `mode="overwrite"` ---> should *enforce* constraints that exist, not simply accept them
  * This test case I am worried about, since this PR fixes a bug with this codepath to begin with, but perhaps the fix also hits the constraint validation test cases after the fact?

ion-elgreco:
> Do we feel covered by existing tests for the following cases:
> 
> * `schema_mode=None` and schema drift detected ---> should fail with informative error

This should be already catched by existing tests I believe, and also raise at the beginning of schema checking

> * `schema_mode=None` and `mode="append"` ---> should _enforce_ constraints that exist, not simply accept them

This should be already catched by existing tests I believe, and also raise at the beginning of schema checking


> * `schema_mode=None` and `mode="overwrite"` ---> should _enforce_ constraints that exist, not simply accept them
>   * This test case I am worried about, since this PR fixes a bug with this codepath to begin with, but perhaps the fix also hits the constraint validation test cases after the fact?

I guess it depends on the type of change if we go from not nullable to nullable with schema mode = None, then it's incorrect. If our data however gets casted from nullable to not nullable, but the table_schema remains the same it's logically correct, but the data should ideally not be changed because it affects the parquet schema

FrankPortman:
Can you expand more on what that means and any implications for this PR? I can't tell if what you said is just regarding testing or regarding the behavior this PR should change.

ion-elgreco:
> Can you expand more on what that means and any implications for this PR? I can't tell if what you said is just regarding testing or regarding the behavior this PR should change.

The first two points is rather, that it shouldn't require changing existing tests since that's already covered, the last remark I feel we need to add a test for to be sure

FrankPortman:
Added 2 test cases - please take a look.

FrankPortman:
Also added a more specific test for `merge` even though that worked as intended already.

FrankPortman:
@rtyler you may want to squash my commits and give it the message from the PR title. That was my assumption as to what would happen when I gave relatively uninformative intra-PR commits, but it seems like they all landed as is 😄 .

Thanks for taking a look!

PR #3748, opened 2025-09-10
Use a safe checkpoint when cleaning up metadata

### LINKED ISSUE #3692: bug: table.cleanup_metadata does not correctly preserve files
# Environment

**Delta-rs version**:
deltalake-1.1.4

**Binding**: 
python

***
# Bug

When calling cleanup_metadata it incorrectly only looks for the latest checkpoint. As a result the `until_version` is not honored and old versions of the table can get incorrectly deleted from the log. Below is a program to reproduce the problem, here we want to delete log entries before version 5.

[test_cleanup.py](https://github.com/user-attachments/files/21927817/test_cleanup.py)

Output:
```
################################################
Failed cleanup test:
log contents before cleanup:
Delta log contents:
  00000000000000000000.json
  00000000000000000001.json
  00000000000000000002.json
  00000000000000000003.json
  00000000000000000004.json
  00000000000000000005.json
  00000000000000000006.json
  00000000000000000007.json
  00000000000000000008.json
  00000000000000000009.checkpoint.parquet
  00000000000000000009.json
  _last_checkpoint

Expected exception occurred: Generic DeltaTable error: Cannot read metadata from log segment
  File "/home/cjoy/src/delta_rs_cleanup/test_cleanup.py", line 67, in <module>
    failed_cleanup_test()
  File "/home/cjoy/src/delta_rs_cleanup/test_cleanup.py", line 46, in failed_cleanup_test
    table.cleanup_metadata()
  File "/home/cjoy/src/delta_rs_cleanup/.venv/lib/python3.12/site-packages/deltalake/table.py", line 981, in cleanup_metadata
    self._table.cleanup_metadata()
_internal.DeltaError: Generic DeltaTable error: Cannot read metadata from log segment

```

With a manual checkpoint at version 5 (See line 45 of the example) we get
```
################################################
Failed cleanup test:
log contents before cleanup:
Delta log contents:
  00000000000000000000.json
  00000000000000000001.json
  00000000000000000002.json
  00000000000000000003.json
  00000000000000000004.json
  00000000000000000005.json
  00000000000000000006.json
  00000000000000000007.json
  00000000000000000008.json
  00000000000000000009.checkpoint.parquet
  00000000000000000009.json
  _last_checkpoint

################################################
log contents after cleanup:
Delta log contents:
  00000000000000000005.checkpoint.parquet
  00000000000000000005.json
  00000000000000000006.json
  00000000000000000007.json
  00000000000000000008.json
  00000000000000000009.checkpoint.parquet
  00000000000000000009.json
  _last_checkpoint

################################################
Version 5 of the data:
   id  value
0  50  str_0
1  51  str_1
2  52  str_2
3  53  str_3
4  54  str_4
5  55  str_5
6  56  str_6
7  57  str_7
8  58  str_8
9  59  str_9
```

I believe that this is a bug and that there is a logic error in `cleanup_expired_logs_for`.
Specifically, this line:
https://github.com/delta-io/delta-rs/blob/ef9e077402013bedf44fccf872f7d7d4fd9157ab/crates/core/src/protocol/checkpoints.rs#L193
The value of `maybe_last_checkpoint` should be the last checkpoint with a version <= `until_version` and created on or before `cutoff_timestamp`.  Otherwise, while the files in the log may be kept, their dependencies will not, and these versions will no longer be loadable / usable.

So, I propose that either:
1. This logic be fixed since it seems like a hidden bug to me and it bit some of our users.
2. Or, if this is really working as intended, then the docs should be updated with a warning since the versions below the latest checkpoint may be rendered unusable.

corwinjoy:
# Description
Inside the function `table.cleanup_metadata `look for the last checkpoint before the requested minimum cutoff time and version. Only delete files before that checkpoint. This is to fix the bug below and make sure that versions inside the retention period are still loadable.

# Related Issue(s)
Fixes #3692


corwinjoy:
Overview via copilot.
## Pull Request Overview

This PR upgrades the `cleanup_expired_logs_for` function to find a safe checkpoint boundary before deleting expired logs, preventing issues where required log files might be deleted that are still needed to reconstruct table states.

- Enhanced the log cleanup logic to identify the most recent checkpoint before the cutoff version and timestamp
- Added comprehensive test coverage for the new safe checkpoint behavior
- Updated existing test configuration to use a shorter retention duration for more precise testing

### Changes

| File | Description |
| ---- | ----------- |
| crates/core/src/protocol/checkpoints.rs | Implemented safe checkpoint boundary logic in `cleanup_expired_logs_for` function with enhanced documentation and debugging |
| crates/core/tests/checkpoint_writer.rs | Added new test case and updated existing test to validate safe checkpoint cleanup behavior |

corwinjoy:
@adamreeve

python/tests/test_table_read.py
```diff
@@ -184,7 +184,8 @@ def test_load_as_version_datetime_with_logs_removed(
     dt.cleanup_metadata()
 
     file = log_path / f"0000000000000000000{expected_version - 1}.json"
-    assert not file.exists()
+    if expected_version > 0:
+        assert file.exists()
```
corwinjoy: @adamreeve This could maybe use improvement.

adamreeve: So this is nonsensical for expected_version = 0. But for expected_version = 1, log files for version 0 and 1 have timestamps outside the retention period. A checkpoint is created for version 1, so I would expect that the file for version 0 is deleted and only version 1 and above are retained. Which is what happened before but this has changed.

Debugging this, I can see that this is due to using the timestamp of the checkpoint file rather than the log file in the cleanup logic. Although the log file for version 1 is outside the retention period, the new checkpoint was only just created so it's inside the retention period, and we end up aborting the cleanup here: https://github.com/corwinjoy/delta-rs/blob/437df15dc4a047f9bc5065e881498a8591b61bcd/crates/core/src/protocol/checkpoints.rs#L269

Maybe the cleanup logic should use the timestamp on the log file corresponding to the checkpoint, rather than the timestamp on the checkpoint itself? As it seems like this scenario could happen in real usage, not just tests, where users might create a checkpoint long after a version is created and just before running cleanup.

corwinjoy: OK, I updated the logic as we discussed. See https://github.com/corwinjoy/delta-rs/pull/15
The new logic for the function is:

1. Find the minimum version for the log files in the retention period. That is, min(version) for files matching DELTA_LOG_REGEX and ts >= cutoff_timestamp. Call this min_retention_version. If empty, set to keep_version.
2. Let keep_version = min(keep_version, min_retention_version).
3. Find the safe checkpoint, where checkpoint_version <= keep_version. (No timestamp restriction). Call this version safe_checkpoint_version.
4. Delete files matching DELTA_LOG_REGEX where log_ver < safe_checkpoint_version && ts <= cutoff_timestamp.

rtyler (dismissed):

rtyler (approved):

corwinjoy:
@rtyler To fix the failing CI test, (https://github.com/delta-io/delta-rs/actions/runs/18060977418/job/51397176816),  I went ahead and updated `tests/test_cleanup.py` to not use pandas.

rtyler (approved):

corwinjoy:
Thanks @rtyler !

PR #3762, opened 2025-09-16
feat(datafusion): add insert_into operation with DataFusion

### LINKED ISSUE #3084: Support DataFusion Insert Into
# Description
Implement `TableProvider::insert_into` of Datafusion.

**Use Case**
We can execute sql `insert into demo(a,b) values(1,2)` to insert data into the delta lake.

fvaleye:
# Description
Implement `TableProvider::insert_into` to support programmatic API calls, and extend `DataSink` to enable SQL insert operations in DataFusion.

# Related Issue(s)
- closes #3084

# Documentation
- [insert_into](https://docs.rs/datafusion/latest/datafusion/datasource/trait.TableProvider.html#tymethod.insert_into) from `TableProvider`
- [DataSink](https://docs.rs/datafusion/latest/datafusion/datasource/sink/trait.DataSink.html)

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider.rs
+++ b/crates/core/src/delta_datafusion/table_provider.rs
@@ -3,6 +3,9 @@ use std::borrow::Cow;
 use std::fmt;
 use std::sync::Arc;
 
+use async_trait::async_trait;
+use futures::StreamExt;
+
 use arrow::compute::filter_record_batch;
 use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
 use arrow::error::ArrowError;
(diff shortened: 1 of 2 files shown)


crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -41,19 +48,226 @@ use datafusion::{
...
+        Some(self.metrics.clone_inner())
+    }
+
+    fn schema(&self) -> &SchemaRef {
+        &self.schema
+    }
+
+    /// Write the data to the delta table
+    /// This is used for insert into operation
+    async fn write_all(
```
ion-elgreco, line 169: @rtyler @fvaleye I believe this should have been piped through the WriteBuilder because now we have another surface layer where we write differently

fvaleye:
@rtyler @ion-elgreco here to help with refactoring 👋 . 
I have added this feature to understand the integration better, but the DataFusion integration requires further refactoring.

crates/core/src/delta_datafusion/table_provider.rs
ahirner: As I'm looking into partioned overwrites, `WriteBuilder` is the preferred interface now?

If so, does it sound OK to map `WriteBuilder::with_replace_where` et al. as `TableProviderFactory::create` options also? In that way, people can `CREATE EXTERNAL TABLE` with protected `insert_into` implementations, thus have a more integrated datafusion experience.

PS: So far I tried `RecordBatchWriter::for_table` and `DeltaWriter` but wasn't able to make partitioned overwrite work (well). I just stumbled upon `WriteBuilder`.

PR #3763, opened 2025-09-16
feature: OptimizeBuilder accepts SessionConfig for finer-grained control of execution

### LINKED ISSUE #3751: Allow configuring SessionConfig in optimize
We would like to read Datafusion SessionConfig settings from our env to be able to control compute based on available resources and our desired concurrency in optimize operations

https://github.com/delta-io/delta-rs/blob/538a90cd438ade2c0f11ee468d43a93f2aa6fcfc/crates/core/src/operations/optimize.rs#L1107

rtyler:
After some debate decided that SessionConfig is sufficient to pull
through rather than SessionContext. Due to the wacky nature of some of
our code-paths in this operation, the signatures of a number of internal
functions had to change to allow SessionConfig to be pulled all the way
down into the ZOrderExecContext

Fixes #3751


fvaleye:
Hey @rtyler 👋,

If you would like minimal changes:
You could adapt`optimize.rs`:

`let ctx = SessionContext::new_with_config_rt(SessionConfig::default(), runtime);`
to
```
// Create SessionConfig from environment variables
let config_options = ConfigOptions::from_env().unwrap_or_default();
let session_config = SessionConfig::from(config_options);
let ctx = SessionContext::new_with_config_rt(session_config, runtime);
```

For going broader across the codebase, you could adapt `DeltaSessionConfig` to load the config from env by default on creation:
```
 pub fn new() -> Self {
      // Start with environment variables if available, otherwise use DataFusion defaults
      let base_config = datafusion::common::config::ConfigOptions::from_env()
          .map(SessionConfig::from)
          .unwrap_or_default();

      // Apply delta-specific overrides
      let inner = base_config.set_bool("datafusion.sql_parser.enable_ident_normalization", false);

      DeltaSessionConfig { inner }
  }
```

rtyler:
@fvaleye I have taken some time to cnosider the suggestion. I am reluctant to add environmental configuration loading into delta-rs from a general standpoint. I think that can lead to surprising behavior for end-users who are not expecting it.

I feel that the builder pattern does give end-users the ability to do that if
they _want_. 

I am not thrilled with how invesaive adding this has to be, but I don't have a
better idea in mind to exposing this API without touching too much

fvaleye:
> @fvaleye I have taken some time to cnosider the suggestion. I am reluctant to add environmental configuration loading into delta-rs from a general standpoint. I think that can lead to surprising behavior for end-users who are not expecting it.
> 
> I feel that the builder pattern does give end-users the ability to do that if they _want_.
> 
> I am not thrilled with how invesaive adding this has to be, but I don't have a better idea in mind to exposing this API without touching too much

Fair point!
It would actually change the current behaviour, I agree.

The builder's new option is the more convenient approach.

PR #3793, opened 2025-09-29
chore(ci): add automatic cache cleanup for closed main branch PRs

fvaleye:
# Description
Add a step to clean the cache entries of GitHub actions:
- Integrate cleanup-cache job into dev_pr.yml workflow  
- Delete up to 100 cache entries per closed PR targeting main
- Prevent cache thrashing with proper error handling and logging

# Documentation
- https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches#force-deleting-cache-entries


.github/workflows/dev_pr.yml
```diff
@@ -8,6 +8,7 @@ on:
       - edited
       - synchronize
       - reopened
+      - closed
```
rtyler: Because there are other types of targets in this list, I think this is going to purge the caches for _any_ pull request action rather than just "closed". 

I don't think there's a downside to creating a separate workflow file just for the cleanup and then that can uave

```yaml
on:
  pull_request_target:
    - closed

```

fvaleye: Right, thanks @rtyler

Issue #3300, opened 2025-03-05
Feature: Support Parquet Modular Encryption


corwinjoy:
[Google Doc Link for Comments](https://docs.google.com/document/d/1MUg1J7u5VdLkgejJ4ybzfZt1OmwhQkq2iGPxsn4gqLI/edit?usp=sharing)

# Description

Support Parquet Modular Encryption in delta-rs. ([https://parquet.apache.org/docs/file-format/data-pages/encryption/](https://parquet.apache.org/docs/file-format/data-pages/encryption/))

# Use Cases:

1. Client side encryption for enhanced security.  
2. Encryption of selected columns to protect sensitive data such as personally identifiable information and/or to meet regulatory requirements such as GDPR.  
3. Different levels of access for different groups of users.

# Background:

We are in the process of completing a pull request for arrow-rs to add low-level encryption and decryption support to the library.  
[https://github.com/apache/arrow-rs/pull/6637](https://github.com/apache/arrow-rs/pull/6637)

What we hope to do is “bubble up” these new encryption configuration settings to the top level delta-rs calls. We have an outline of how to do this (and draft working code) but are not all that familiar with the delta-rs architecture and could use a design review and suggestions.

We’d also like feedback on whether this is a feature that is desirable to the rest of the Delta community, and how it should be implemented. Our proposed approach allows integration with Parquet modular encryption in data files, without changing any Delta specific metadata, but there could be an argument for more integrated columnar encryption in the Delta Lake format. For example, the proposed encryption feature for Apache Iceberg v3 includes encryption of column statistics in metadata files.

# Related Reading:

- Encrypted Delta Lakes: [https://github.com/delta-io/delta/issues/2269](https://github.com/delta-io/delta/issues/2269)  
- Apply encryption to PII fields of delta tables: [https://medium.com/@data\_engineering\_0216/apply-encryption-to-pii-fields-of-delta-tables-a61a8e7bef93](https://medium.com/@data_engineering_0216/apply-encryption-to-pii-fields-of-delta-tables-a61a8e7bef93)  
- Column level encryption with PII: [https://www.databricks.com/blog/2020/11/20/enforcing-column-level-encryption-and-avoiding-data-duplication-with-pii.html](https://www.databricks.com/blog/2020/11/20/enforcing-column-level-encryption-and-avoiding-data-duplication-with-pii.html)  
[...]

PR #3794, opened 2025-09-29
feat: add framework for File Format Options

corwinjoy:
# Description
This PR adds encryption support and other advanced file options to `delta-rs` by implementing a comprehensive framework for file format settings. The changes enable users to configure encryption settings, customize writer properties, and apply file-level formatting options when reading and writing Delta tables.

- Introduces a `FileFormatOptions` trait and related infrastructure to handle file-specific configurations
- Adds support for both simple property-based encryption and KMS-based encryption through new factory patterns
- Updates all operation builders to accept and propagate file format options throughout the write/read pipeline

In general, we have added a new trait called `FileFormatOptions` at the root `DeltaTable` level to unify how files within a delta table are read and written with specific formatting. The idea is that you can apply these settings once, at the top level, and then seamlessly perform any operations with the necessary settings.

This PR leverages the DataFusion `TableOptions` structure to support format options for multiple underlying file formats. (The idea being that `delta-rs` may eventually want to support storage formats beyond _Parquet,_ such as _Vortex_ or _Lance_.) Additionally, it centralizes file format options in a single, consistent location. This avoids the current difficulties where one has to separately set `WriterProperties`; then reader properties as part of the `SessionState`. (This is in line with comments from @roeap about how file configuration might be improved: https://github.com/delta-io/delta-rs/issues/3300#issuecomment-3024736525). We would also like to eventually extend this upgrade to add notations about these file configurations to the delta table properties. For example, if the files are encrypted, one could add a KMS configuration for where to retrieve encryption keys.

## Review Suggestion
This PR turned out to be larger than we hoped, so apologies for that, but I don't know how to split it into smaller pieces.
When reviewing, we suggest starting with the file `crates/core/src/table/file_format_options.rs` to get an overview of the new file format trait that can be applied to delta tables.

# Related Issue(s)
Support Parquet Modular Encryption:
https://github.com/delta-io/delta-rs/issues/3300

# Documentation

Parquet Modular Encryption: https://docs.google.com/document/d/1MUg1J7u5VdLkgejJ4ybzfZt1OmwhQkq2iGPxsn4gqLI/edit?tab=t.0#heading=h.34wvmhc1zdch

# Attribution
This PR was created in collaboration with @adamreeve


corwinjoy:
Note that fully supporting Parquet encryption requires being able to get write and read properties per-file, which is why the existing ability to set `WriterProperties` isn't sufficient, and why `WriterPropertiesFactory::create_writer_properties` is called per file and requires a file path. This allows generating new random data encryption keys per file and performing tasks such as specifying a per-file AAD prefix or supporting the external storage of encryption keys that can be looked up using the file path.

crates/core/src/operations/optimize.rs
```diff
@@ -331,27 +347,21 @@ impl<'a> std::future::IntoFuture for OptimizeBuilder<'a> {
             let operation_id = this.get_operation_id();
             this.pre_execute(operation_id).await?;
 
-            let writer_properties = this.writer_properties.unwrap_or_else(|| {
-                WriterProperties::builder()
-                    .set_compression(Compression::ZSTD(ZstdLevel::try_new(4).unwrap()))
-                    .set_created_by(format!("delta-rs version {}", crate_version()))
-                    .build()
-            });
```
corwinjoy, line 380: TODO.  Right now, we're breaking the existing behaviour of using zstd for compaction operations and this will use snappy by default. This is OK for getting feedback on the initial PR, but we should make sure to address this before merging.

corwinjoy:
@rtyler @roeap @alamb Tagging you here per our previous discussion on adding encryption support to delta-rs.

crates/core/src/operations/optimize.rs
corwinjoy: OK. I have updated the constructor to use zstd compression by default, so that should address this.

rtyler:
I have marked this pull request as draft. This does not compile as is, I can come back to it once it is able to compile and pass unit tests

corwinjoy:
> I have marked this pull request as draft. This does not compile as is, I can come back to it once it is able to compile and pass unit tests

@rtyler OK. It seems that when I auto-merged the main branch it introduced a build error. I have resolved this and the code is once again building and passing unit tests.

crates/core/src/operations/add_feature.rs
```diff
@@ -144,6 +144,7 @@ impl std::future::IntoFuture for AddTableFeatureBuilder {
             Ok(DeltaTable::new_with_state(
                 this.log_store,
                 commit.snapshot(),
+                None,
```
ion-elgreco: Let's not change the function signature here and in the other builders

crates/core/src/operations/delete.rs
```diff
@@ -190,12 +204,13 @@ impl ExtensionPlanner for DeleteMetricExtensionPlanner {
 #[allow(clippy::too_many_arguments)]
 async fn execute_non_empty_expr(
     snapshot: &DeltaTableState,
+    file_format_options: Option<FileFormatRef>,
```
ion-elgreco: I think it would be better if this pushed into the LoadConfig and not passed through each function

ion-elgreco (changes requested):
I can see the benefit but we really need to reduce the surface of change that are being introduced

roeap:
@corwinjoy - awesome to see this come to fruition! Will find some time to give this a review hopefully tomorrow.

At first glance one quick question. Do we see a way to "bundle" the datafusion specific stuff a bit more? It's a bit hard to keep track of all the individual flags while reviewing :)

crates/core/src/operations/add_feature.rs
corwinjoy: Could you be more specific about what you are looking for? Checking the code base, I see 25 calls to this constructor, and in 10/25 of the cases, I need to pass `file_format_options` to maintain the needed settings. I guess I could eliminate the 15 cases where I pass None by splitting this into two named constructors...

crates/core/src/operations/delete.rs
corwinjoy: Can you elaborate a bit here? There are a lot of these different execute functions (one for every option) and I agree that it would be nice if they had a common configuration structure rather than the somewhat long list of arguments they take. That would probably be independent of this PR. For now, we have just added an additional argument to pass the needed settings through to execution.

corwinjoy:
@roeap 
> At first glance one quick question. Do we see a way to "bundle" the datafusion specific stuff a bit more? It's a bit hard to keep track of all the individual flags while reviewing :)

What we did to minimize this dependency is define an abstract `FileFormatOptions` trait. Everything just passes around a `FileFormatRef` defined as `Arc<dyn FileFormatOptions>`. Then, only when needed, do we grab final table options or writer properties. Furthermore, we've gated these instances of getting final details behind three function calls in `file_format_options.rs`:
```
pub fn build_writer_properties_factory_ffo(
    file_format_options: Option<FileFormatRef>,
) -> Option<Arc<dyn WriterPropertiesFactory>> {...}

pub fn to_table_parquet_options_from_ffo(
    file_format_options: Option<&FileFormatRef>,
) -> Option<TableParquetOptions> {...}

pub fn state_with_file_format_options(
    state: SessionState,
    file_format_options: Option<&FileFormatRef>,
) -> DeltaResult<SessionState> {...}

```

There might be some ways to refine this further, but in general we've tried to isolate and abstract these file properties where possible and not require datafusion.

corwinjoy:
@roeap From a user point of view, we've tried hard to make the settings as easy as possible. This can be seen in `crates/deltalake/examples/basic_operations_encryption.rs`. Here, we demonstrate different kinds of operations on tables. (We have a more formal unit test at `crates/core/tests/commands_with_encryption.rs`). Thes code examples all look like ordinary operations; all we needed was a common function call when creating `DeltaOps`:
```
async fn ops_with_crypto(
    uri: &str,
    file_format_options: &FileFormatRef,
) -> Result<DeltaOps, DeltaTableError> {
    let prefix_uri = format!("file://{}", uri);
    let url = Url::parse(&*prefix_uri).unwrap();
    let ops = DeltaOps::try_from_uri(url).await?;
    Ok(ops.with_file_format_options(file_format_options.clone()))
}
```

Calling `with_file_format_options` is sufficient to apply the needed encryption settings for all operations.

crates/core/src/operations/add_feature.rs
ion-elgreco: You need to use DeltaTableConfig for this: https://github.com/delta-io/delta-rs/blob/18f949efba220f9b6840a3a991e6d0726198fa18/crates/core/src/table/builder.rs#L30-L58

crates/core/src/operations/delete.rs
ion-elgreco: See ping above, it needs to be `DeltaTableConfig`

crates/core/src/operations/add_feature.rs
corwinjoy: OK. I think you are suggesting that I add `file_format_options` to the `config` member of `DeltaTable`.
In fact, that was the approach I tried initially but I had to back that out and add this as a direct member for the following reasons:
1. The file format options don't really seem to fit properly into `DeltaTableConfig` since these options seem to be more about managing the logfile.
2. We need to preserve the formatting options when going from `DeltaTable` to `DeltaOps` and back. Right now, the config gets lost when `new_with_state` is called. See below where the `config` gets reset to default:
https://github.com/delta-io/delta-rs/blob/18f949efba220f9b6840a3a991e6d0726198fa18/crates/core/src/table/mod.rs#L128
We need to preserve these settings throughout any chained operations. This means we still need an extra parameter in `new_with_state` to preserve any existings config. Also, I felt it was cleaner to create a new entry and directly set and pass it. Possibly we could move this into `DeltaTableConfig` but I think this may be more of a hindrance than a help.

@adamreeve It was a couple of months ago that we moved this, do you remember anything else from our discussion? See commit below:
https://github.com/delta-io/delta-rs/commit/666f0ba264659e10bdb59f835fe66ee60cc4327f

ion-elgreco: 1. Correct but they are related to loading the table, which file formats (encryption) should belong as well
2. DeltaTableConfig tags along with the snapshot: https://github.com/delta-io/delta-rs/blob/a61ac169dcf2982e78ee1bdbaaa9b5e8dd5fe019/crates/core/src/table/state.rs#L73-L76

so that shouldn't be an issue to allow it to stay there when you do new_with_state

corwinjoy: OK. I think it makes sense to move this here. I will investigate this week. The main issue is whether I have to support serialization out of the gate to make this work / at what points config gets reconstructed from serialized properties. The reason why is that serialization of the `FileFormatOptions` will be a bit tricky because:
1. I'm not sure if `TableOptions` fully supports serialization.
2. For direct encryption and decryption properties, we will need to modify the serialization to make sure that passwords don't get serialized.

corwinjoy:
Still working on this to move the file config to `DeltaTableConfig`, but doing this in a separate branch to keep things clean. It's fairly tricky, so it will take a bit. I also plan to improve the unit tests to confirm that files are really being encrypted.

corwinjoy:
@ion-elgreco OK. I have migrated these file options to the config property in `DeltaTable`. This definitely reduced the changes quite a bit so thanks for the suggestion! I think it looks pretty solid and look forward to your feedback when you are ready.
@adamreeve

crates/core/src/operations/mod.rs
```diff
@@ -181,6 +185,28 @@ impl DeltaOps {
...
+                    &self.0.log_store,
+                    self.0.config.clone(),
+                    Some(self.0.state.unwrap().version()),
+                )
+                .await?,
+            );
+        }
+        Ok(self)
+    }
+
```
corwinjoy, line 368: @ion-elgreco We could use some feedback on the best way to set the config for an existing table. In this design, we wanted to:
1. Be able to set the config at runtime, not just at table construction. This is important for fields like passwords, where we will not want them to be serialized, or other options that we may want to change at runtime.
2. Make the user interface easy to use.
With this design, setting the config for any operation looks like: (from crates/deltalake/examples/basic_operations_encryption.rs)
```
    let ops = DeltaOps::try_from_uri(url).await?;
    let ops = ops
        .with_file_format_options(file_format_options.clone())
        .await?;
```

We also considered a design where you could only set this via `DeltaTableBuilder` but I wanted to make this feature easy to use.  See e.g. the following diff where this function is removed.

https://github.com/adamreeve/delta-rs/commit/ec40881890012fa056e1afcb7dd74cf8cffc065a

@adamreeve

ion-elgreco:
> @ion-elgreco OK. I have migrated these file options to the config property in `DeltaTable`. This definitely reduced the changes quite a bit so thanks for the suggestion! I think it looks pretty solid and look forward to your feedback when you are ready. @adamreeve

I'll do another review over the weekend

crates/core/src/operations/mod.rs
adamreeve: I've just pushed a fixed commit that handles when the load fails if the table isn't created yet: https://github.com/adamreeve/delta-rs/commit/0d551a50ac02ab0f074eb3b026b4d041cb3f3813

This matches the behaviour of `DeltaOps::try_from_uri`. Maybe there could be a `DeltaOps::try_from_table` or `TryFrom<DeltaTable> for DeltaOps` implementation to handle that scenario.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -45,6 +46,7 @@ pub(crate) async fn find_files(
     snapshot: &EagerSnapshot,
     log_store: LogStoreRef,
     session: &dyn Session,
+    file_format_options: Option<&FileFormatRef>,
```
ion-elgreco: This is not needed anymore, since you can access the snapshot.load_config() to access the `file_format_options`

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -221,6 +229,7 @@ async fn find_files_scan(
...
```
ion-elgreco: Same here it's already available in the snapshot. so we can defer at latest stage to grab these from the load_config

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -250,10 +259,13 @@ async fn find_files_scan(
     // Add path column
     used_columns.push(logical_schema.index_of(scan_config.file_column_name.as_ref().unwrap())?);
 
+    let table_parquet_options = to_table_parquet_options_from_ffo(file_format_options);
```
ion-elgreco: I prefer we `Impl Into` here between these two structs.

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -373,6 +374,7 @@ pub(crate) struct DeltaScanBuilder<'a> {
     limit: Option<usize>,
     files: Option<&'a [Add]>,
     config: Option<DeltaScanConfig>,
+    parquet_options: Option<TableParquetOptions>,
```
ion-elgreco: I suggest we remove this here. We can move all this logic into:

DeltaScanConfigBuilder.build(), there we can introspect the TableLoadConfig and set the parquetOptions on the DeltaScanConfig

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -730,9 +755,17 @@ impl TableProvider for DeltaTable {
...
+            format_options.update_session(session)?;
+        }
         let filter_expr = conjunction(filters.iter().cloned());
 
         let scan = DeltaScanBuilder::new(self.snapshot()?.snapshot(), self.log_store(), session)
+            .with_parquet_options(
+                crate::table::file_format_options::to_table_parquet_options_from_ffo(
+                    self.config.file_format_options.as_ref(),
+                ),
+            )
```
ion-elgreco: This can be removed since this will happen inside the DeltaScanBuilder with my suggestion above

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -763,6 +796,7 @@ pub struct DeltaTableProvider {
     config: DeltaScanConfig,
     schema: Arc<Schema>,
     files: Option<Vec<Add>>,
+    file_format_options: Option<FileFormatRef>,
```
ion-elgreco: Same here, not needed anymore, since we can pass it through the DeltaScanConfig

crates/core/src/operations/mod.rs
ion-elgreco: I am having hard time understanding this usecase. Why wouldn't it be enough to update the DeltaTableConfig, I would assume without setting the encryption you couldn't load the table at all

crates/core/src/operations/optimize.rs
```diff
@@ -1080,6 +1119,29 @@ async fn build_zorder_plan(
     Ok((operation, metrics))
 }
 
+async fn get_file_decryption_properties(
```
ion-elgreco, line 1175: Why can this return None as well?

crates/core/src/table/file_format_options.rs
```diff
@@ -0,0 +1,344 @@
...
+        // Ensure DataFusion has the encryption factory registered
+        session.runtime_env().register_parquet_encryption_factory(
+            &self.encryption_factory_id,
+            Arc::clone(self.table_encryption.encryption_factory()),
+        );
+        Ok(())
+    }
+}
+
+/// Extension trait to obtain a `WriterPropertiesBuilder` from an existing
```
ion-elgreco: 56.2.0 has landed so this can be removed

crates/core/src/table/file_format_options.rs
```diff
@@ -0,0 +1,344 @@
...
+    fn table_options(&self) -> TableOptions {
+        self.table_options.clone()
+    }
+
+    fn writer_properties_factory(&self) -> WriterPropertiesFactoryRef {
+        build_writer_properties_factory_tpo(&Some(self.table_options.parquet.clone())).unwrap()
+    }
+}
+
+pub fn build_writer_properties_factory_ffo(
```
ion-elgreco: Do we really need a function for this 🤷

crates/core/src/table/file_format_options.rs
```diff
@@ -0,0 +1,344 @@
...
+    file_format_options: Option<FileFormatRef>,
+) -> Option<WriterPropertiesFactoryRef> {
+    file_format_options.map(|ffo| ffo.writer_properties_factory())
+}
+
+pub fn build_writer_properties_factory_or_default_ffo(
+    file_format_options: Option<FileFormatRef>,
+) -> WriterPropertiesFactoryRef {
+    build_writer_properties_factory_ffo(file_format_options)
+        .unwrap_or_else(|| build_writer_properties_factory_default())
```
ion-elgreco: unwrap_or_default is more idiomatic

crates/core/src/table/file_format_options.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+
+#[cfg(feature = "datafusion")]
+pub fn to_table_parquet_options_from_ffo(
```
ion-elgreco: same applies here, I dont think we need a function for just a map()

ion-elgreco:
@corwinjoy looks already better but I think we can reduce the amount of line changes even more!

I still have to take a better look at why we need a WriterPropertiesFactory :s but maybe you can explain it shortly for me?

adamreeve:
> I still have to take a better look at why we need a WriterPropertiesFactory :s but maybe you can explain it shortly for me?

Corwin is busy travelling this week so might be slow to reply, but I can help answer this part. For some use cases it might be fine to have a single `WriterProperties` instance for all files, but there are a few reasons why you could want to generate new encryption properties per-file so need a factory:
* To generate new random data encryption keys per file, as it's good security practice to limit how widely one key is used
* To set a different [AAD prefix](https://github.com/apache/parquet-format/blob/master/Encryption.md#441-aad-prefix) per-file, which prevents attackers from being able to swap out encrypted modules between files and tamper with data
* Be able to handle schema changes like adding columns while specifying per-column encryption keys
* Enables use of [external key material](https://g-research.github.io/ParquetSharp/guides/Encryption.html#external-metadata-and-key-rotation), which is when you write a JSON file alongside each Parquet file containing the key metadata. This allows rotation of master keys without having to re-write Parquet files, only the JSON files need to be rewritten. We don't implement this in the Rust parquet-key-management crate but it's supported by the Java and C++/Python Parquet implementations.

This also aligns with the DataFusion [`EncryptionFactory`](https://github.com/apache/datafusion/blob/37aad28424da933e8966a032b502b9a0cf4b7507/datafusion/execution/src/parquet_encryption.rs#L37) trait that takes the file path as a parameter when creating encryption and decryption properties.

crates/core/src/table/file_format_options.rs
corwinjoy: Not quite sure how to do that since `WriterPropertiesFactoryRef` is just a typedef and not a true type.

crates/core/src/table/file_format_options.rs
corwinjoy: Removed.

crates/core/src/delta_datafusion/find_files.rs
corwinjoy: This was removed and replaced with a direct call to map.

crates/core/src/delta_datafusion/find_files.rs
corwinjoy: OK. Removed.

crates/core/src/delta_datafusion/find_files.rs
corwinjoy: Removed.

crates/core/src/operations/optimize.rs
corwinjoy: This can return None because `FileFormatOptions` may be set in order to specify a particular parquet formatting. But, this may not include any encryption. `FileFormatOptions` is for more than just encryption.

crates/core/src/delta_datafusion/table_provider.rs
corwinjoy: OK. I have moved these `parquet_options` to DeltaScanConfig.

crates/core/src/delta_datafusion/table_provider.rs
corwinjoy: OK. Moved as suggested.

crates/core/src/delta_datafusion/table_provider.rs
corwinjoy: Removed as suggested.

crates/core/src/operations/mod.rs
corwinjoy: @ion-elgreco The use case is this.
1. We have an existing table on the file system. We want to load it and apply some operations.
2. But, we need to set some parquet file options at runtime. These may be passwords that we don't want serialized. Or, they could be specific parquet options that we want to apply for the operations we invoke.
3. The above pattern handles this use case e.g.
```
    let ops = DeltaOps::try_from_uri(url).await?;
    let ops = ops
        .with_file_format_options(file_format_options.clone())
        .await?;
```

adamreeve: I think my example can also handle this use case though. You can specify the file format options using `DeltaTableBuilder::with_file_format_options` when loading the table at runtime, and there's no serialization involved. 

> I would assume without setting the encryption you couldn't load the table at all

Only the Parquet data files are encrypted so you can still load the table without needing any encryption options set, as loading the table only needs to read the log files.

I noticed discussion about consolidating DeltaOps and DeltaTable in https://github.com/delta-io/delta-rs/discussions/3866 that might be relevant to this. If ops are merged into DeltaTable and the DeltaOps type is removed then I guess this becomes a non-issue.

crates/core/src/table/file_format_options.rs
corwinjoy: Removed.

crates/core/src/operations/add_feature.rs
corwinjoy: Fixed.

crates/core/src/operations/delete.rs
corwinjoy: Fixed.

corwinjoy:
@ion-elgreco OK. I have applied the changes you suggested from two weeks ago and re-merged the latest from main. So, I am ready for another review when you get the chance. Thanks again for the suggestions!

crates/core/src/writer/record_batch.rs
```diff
@@ -223,7 +236,8 @@ impl RecordBatchWriter {
 
     /// Sets the writer properties for the underlying arrow writer.
     pub fn with_writer_properties(mut self, writer_properties: WriterProperties) -> Self {
-        self.writer_properties = writer_properties;
+        let writer_properties_factory = build_writer_properties_factory_wp(writer_properties);
+        self.writer_properties_factory = writer_properties_factory;
```
ion-elgreco: I prefer the use of extension traits here, example:

```rust
pub trait IntoWriterPropertiesFactoryRef {
    fn into_factory_ref(self) -> WriterPropertiesFactoryRef;
}

impl IntoWriterPropertiesFactoryRef for WriterProperties {
    fn into_factory_ref(self) -> WriterPropertiesFactoryRef {
        Arc::new(SimpleWriterPropertiesFactory::new(self))
    }
}
```

And this can be applied across the board where we do func(a) -> b

crates/core/src/operations/write/writer.rs
```diff
@@ -404,22 +418,28 @@ pub struct PartitionWriter {
 
 impl PartitionWriter {
     /// Create a new instance of [`PartitionWriter`] from [`PartitionWriterConfig`]
-    pub fn try_with_config(
+    pub async fn try_with_config(
```
ion-elgreco: Why has this been made async?

crates/core/src/table/file_format_options.rs
```diff
@@ -0,0 +1,268 @@
...
+    ) -> DeltaResult<WriterProperties> {
+        debug!("Called create_writer_properties for file: {file_path}");
+        Ok(self.writer_properties.clone())
+    }
+}
+
+// More advanced factory with KMS support
+#[cfg(feature = "datafusion")]
+#[derive(Clone, Debug)]
+pub struct KMSWriterPropertiesFactory {
```
ion-elgreco: I think this can be moved  (all KMS stuff) to your own private crate Since we have the WriterPropetiesFactory trait

crates/core/src/table/file_format_options.rs
```diff
@@ -0,0 +1,268 @@
...
+
+/// Convenience alias for file format options reference used across the codebase
+pub type FileFormatRef = Arc<dyn FileFormatOptions>;
+
+/// Convenience alias for writer properties factory reference used across the codebase
+pub type WriterPropertiesFactoryRef = Arc<dyn WriterPropertiesFactory>;
+
+#[cfg(feature = "datafusion")]
+#[derive(Clone, Debug, Default)]
+pub struct SimpleFileFormatOptions {
```
ion-elgreco, line 45: What's the purpose for this for other delta-rs users?

corwinjoy: `SimpleFileFormatOptions` is a helper for users to easily create a `FileFormatRef` from `TableOptions` in order to set their read and write properties. We use it like:
```
let file_format_options = Arc::new(SimpleFileFormatOptions::new(tbl_options)) as FileFormatRef;
```
This is needed if users want to set their own `TableOptions` for reading and writing. In our examples, we only use this for encryption. But the idea is that users may set their own parquet options at the `DeltaTable` level. This default class is the easiest way to do it, holding `TableOptions` for reading and writing.

crates/core/src/operations/write/writer.rs
corwinjoy: Good point. I had switched this to async since we were calling the underlying writers directly but with the latest changes to main this is no longer needed. Removed.

crates/core/src/writer/record_batch.rs
corwinjoy: OK. Switched this to extension trait as well as `build_writer_properties_factory_or_default_ffo`.

crates/core/src/table/file_format_options.rs
corwinjoy: @ion-elgreco Could you give more details about what you are looking for here?  For us, KMS integration is a key part of this feature set. The reason for this is that a large driver of this feature set is corporate compliance requirements. (Similar to GDPR in Europe). This means not only securely encrypting sensitive columns but also KMS integration for the proper management of keys (secure key storage, key rotation, user access controls, etc.). As a result, no matter what, we want to make sure that tests to check the KMS features and correctness become part of the unit tests.

Here is one possible way to do this:
1. `parquet-key-management ` is already only a dev dependency.
2. The class `KMSWriterPropertiesFactory` is used in both the example and tests to implement KMS integration.
3. Possibly move `KMSWriterPropertiesFactory` to `crates/core/src/test_utils`.
4. Or, maybe duplicate the definition of `KMSWriterPropertiesFactory` in both the test and example file.

Either way, we want to preserve the KMS tests in `basic_operations_encryption.rs`.
@adamreeve

ion-elgreco: I think we should do option 3 and keep it in tests only, and then put a reference to example integration. The whole encryption part is a very niche use case, so would like to keep the maintenance burden as minimal for us

corwinjoy: OK. This class + `TableEncryption` has been moved into `test_utils/kms_encryption.rs` and is now only refrenced by the encryption example and test.

corwinjoy:
@ion-elgreco OK. I have added changes to address your latest requests.
In addition, I have merged the latest main.
Plus, I have added the update from @jgiannuzzi to support async KMS for the tests and example.
Please let me know if you have any further questions or are ready for another review.

ion-elgreco:
@roeap @roeap for me it looks in a mergable state, maybe you guys can give a final thought on it?

Perhaps we should put this behind an encryption feature gate

corwinjoy:
I had thought about doing this with an encryption feature gate, but left it out for this PR for a few reasons:
1. Adding the gate makes both the code and usage more complex. Making it more of a hassle both to use and debug this feature.
2. Relative to the size of the delta-rs project, disabling the encryption feature won't save much on size.
3. This may be cleaner to add as a follow-up rather than mixed in with an already longish PR.

corwinjoy:
I went ahead and merged latest main again.  @roeap @rtyler any thoughts on this?

corwinjoy:
@roeap @rtyler any thoughts on this?

corwinjoy:
@roeap @rtyler @hntd187  Any chance I can get a second review or merge on this, please? It is somewhat time-consuming to maintain repeated merges of main.

adamreeve:
Sorry to keep bugging people but it would be great to get this PR in to provide more control over Parquet properties and enable using Parquet encryption. Is there another reviewer who can take a look?

I think the two outstanding questions it would be good to have a decision on are whether this needs to be behind a feature flag (and if that can be in a follow up PR If necessary), and this discussion about setting the file format options when creating DeltaOps: https://github.com/delta-io/delta-rs/pull/3794#discussion_r2430571990

corwinjoy:
Updated PR with latest version of main.
1. Apply crypto to new table scan code.
2. Migrate crypto test and example to use methods directly on DeltaTable rather than deprecated DeltaOps.

crates/core/src/operations/mod.rs
corwinjoy: I think that with the move towards using `DeltaTable` directly this is not much of an issue anymore. Right now in our example we can just use `with_file_format_options` directly. From the example file we have:
```
fn table_with_crypto(
    uri: &str,
    file_format_options: &FileFormatRef,
) -> Result<DeltaTable, DeltaTableError> {
    let prefix_uri = format!("file://{}", uri);
    let url = Url::parse(&*prefix_uri).unwrap();
    let table = DeltaTableBuilder::from_url(url)?
        .with_file_format_options(file_format_options.clone())
        .build()?;
    Ok(table)
}
```

corwinjoy:
Just to give an overview, here is roughly how these code changes break down..
```
git diff main file_format_options_squashed --numstat | awk '{ print ($1 + $2) "\t" $0 }' | sort -nr | cut -f 2-
// New code for tests and examples
// inserted deleted filename
502	0	crates/core/tests/commands_with_encryption.rs
338	0	crates/deltalake/examples/basic_operations_encryption.rs

// Code to support encryption and extended file format traits
244	0	crates/core/src/test_utils/kms_encryption.rs
180	0	crates/core/src/table/file_format_options.rs

// Code needed to add encryptions to writes that do not use datafusion
96	31	crates/core/src/operations/optimize.rs
45	26	crates/core/src/writer/json.rs
40	25	crates/core/src/writer/record_batch.rs
38	20	crates/core/src/operations/write/writer.rs
43	13	crates/core/src/operations/delete.rs
43	4	crates/core/src/delta_datafusion/table_provider.rs
37	10	crates/core/src/operations/update.rs
31	11	crates/core/src/operations/merge/mod.rs
30	2	crates/core/src/operations/mod.rs
24	4	crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs

// Code to pass config through from DeltaTable and replace FileFormat with a FileFormatFactory
13	13	crates/core/src/operations/write/execution.rs
13	11	crates/core/src/writer/utils.rs
19	0	crates/core/src/table/builder.rs
13	5	crates/core/src/operations/write/mod.rs
15	2	crates/core/src/operations/load.rs
13	0	crates/core/src/delta_datafusion/table_provider/next/mod.rs
11	2	crates/core/src/operations/create.rs
11	1	crates/core/src/kernel/snapshot/mod.rs
4	4	crates/core/tests/command_optimize.rs
6	1	crates/core/Cargo.toml
6	0	crates/deltalake/Cargo.toml
2	2	Cargo.toml
2	0	crates/core/src/test_utils/mod.rs
1	0	crates/core/src/table/mod.rs
```

the-reacher-data:
Following up on our comment in #3300 — we're evaluating this PR as the prerequisite for Parquet Modular Encryption, and did a detailed review focused on the encryption plumbing, verified against branch head `2c8e486` (signatures cross-checked against the `datafusion` 52.1 / `parquet` 57 versions the branch compiles against).

Overall we think the design is right: the per-file async `WriterPropertiesFactory` is exactly what per-file DEKs/AAD prefixes need, the DataFusion `EncryptionFactory` integration is clean, and `DeltaTableConfig::file_format_options` is the right injection point. Below are five findings with proposed fixes, most severe first.

---

### 1. Make the "next" scan path fail loudly when a configured encryption factory is missing

The classic scan path treats an unregistered factory as a hard error (`table_provider.rs` L625–634 uses `.map(...).transpose()?`), while the kernel path swallows it and continues **without** decryption, so the user gets an opaque parquet decode failure instead of "encryption factory not found". Two read paths for the same table shouldn't disagree on whether a config error is fatal. Hoisting the lookup out of the per-store loop also mirrors the classic path and makes the error deterministic even when a store group is empty.

`crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs` L312–344:

```diff
@@ -312,6 +312,16 @@ async fn get_read_plan(
     let pq_options = table_parquet_options
         .cloned()
         .unwrap_or_else(|| state.table_options().parquet.clone());

+    // A configured-but-unregistered encryption factory is a configuration error.
+    // Silently reading on without it only surfaces later as an opaque parquet
+    // decode failure. Mirrors `DeltaScanBuilder::build`.
+    let encryption_factory = pq_options
+        .crypto
+        .factory_id
+        .as_ref()
+        .map(|factory_id| state.runtime_env().parquet_encryption_factory(factory_id))
+        .transpose()?;
+
     let mut full_read_schema = SchemaBuilder::from(parquet_read_schema.as_ref().clone());
@@ -334,14 +344,9 @@ async fn get_read_plan(
             .with_table_parquet_options(pq_options.clone())
            .with_parquet_file_reader_factory(reader_factory);

-        // Set encryption factory if configured
-        if let Some(factory_id) = &pq_options.crypto.factory_id {
-            if let Ok(encryption_factory) =
-                state.runtime_env().parquet_encryption_factory(factory_id)
-            {
-                file_source = file_source.with_encryption_factory(encryption_factory);
-            }
-        }
+        if let Some(encryption_factory) = &encryption_factory {
+            file_source = file_source.with_encryption_factory(Arc::clone(encryption_factory));
+        }
```

`get_read_plan` returns `datafusion::common::Result<_>` and `RuntimeEnv::parquet_encryption_factory` returns the same `Result`, so `?` needs no conversion; `ParquetSource::with_encryption_factory` takes the `Arc` by value, hence `Arc::clone`. No new imports needed.

---

### 2. Remove the reachable `panic!` in `build_writer_properties_tpo` by validating options at construction

`WriterPropertiesBuilder::try_from(&tpo).expect(...)` turns user configuration into a process abort, and it is re-executed on **every** `writer_properties_factory()` call (once per operation builder). Rather than making the trait method fallible — which would ripple through 7 call sites plus `IntoWriterPropertiesFactoryRef` — we'd suggest validating once in the constructor and storing the factory, exactly like `KmsFileFormatOptions` already does (`kms_encryption.rs` L147–158).

`crates/core/src/table/file_format_options.rs` L43–71 and L95–119:

```diff
@@ -43,29 +43,32 @@
 #[cfg(feature = "datafusion")]
-#[derive(Clone, Debug, Default)]
+#[derive(Clone)]
 pub struct SimpleFileFormatOptions {
     table_options: TableOptions,
+    writer_properties_factory: WriterPropertiesFactoryRef,
 }

 #[cfg(feature = "datafusion")]
 impl SimpleFileFormatOptions {
-    pub fn new(table_options: TableOptions) -> Self {
-        Self { table_options }
-    }
+    /// Build file format options from DataFusion table options.
+    ///
+    /// Writer properties are materialized once here so that an invalid parquet
+    /// configuration is reported to the caller instead of aborting later on a
+    /// write path.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`crate::DeltaTableError`] if the parquet options cannot be
+    /// converted into parquet writer properties.
+    pub fn try_new(table_options: TableOptions) -> DeltaResult<Self> {
+        let writer_properties_factory =
+            build_writer_properties_factory_tpo(&table_options.parquet)?;
+        Ok(Self {
+            table_options,
+            writer_properties_factory,
+        })
+    }
 }

 #[cfg(feature = "datafusion")]
 impl FileFormatOptions for SimpleFileFormatOptions {
     fn table_options(&self) -> TableOptions {
         self.table_options.clone()
     }

     fn writer_properties_factory(&self) -> WriterPropertiesFactoryRef {
-        // SAFETY: unwrap is safe because build_writer_properties_factory_tpo only returns None
-        // when given None. Since we pass Some(...), it will always return Some.
-        build_writer_properties_factory_tpo(&Some(self.table_options.parquet.clone())).unwrap()
+        Arc::clone(&self.writer_properties_factory)
     }
 }
@@ -95,25 +98,22 @@
 #[cfg(feature = "datafusion")]
 fn build_writer_properties_tpo(
-    table_parquet_options: &Option<TableParquetOptions>,
-) -> Option<WriterProperties> {
-    table_parquet_options.as_ref().map(|tpo| {
-        let mut tpo = tpo.clone();
-        tpo.global.skip_arrow_metadata = true;
-        let mut wp_build = WriterPropertiesBuilder::try_from(&tpo)
-            .expect("Failed to convert TableParquetOptions to ParquetWriterOptions");
-        if let Some(enc) = tpo.crypto.file_encryption {
-            wp_build = wp_build.with_file_encryption_properties(Arc::new(enc.into()));
-        }
-        wp_build.build()
-    })
+    table_parquet_options: &TableParquetOptions,
+) -> DeltaResult<WriterProperties> {
+    let mut tpo = table_parquet_options.clone();
+    tpo.global.skip_arrow_metadata = true;
+    let mut wp_build = WriterPropertiesBuilder::try_from(&tpo)?;
+    if let Some(enc) = tpo.crypto.file_encryption {
+        wp_build = wp_build.with_file_encryption_properties(Arc::new(enc.into()));
+    }
+    Ok(wp_build.build())
 }

 #[cfg(feature = "datafusion")]
 fn build_writer_properties_factory_tpo(
-    table_parquet_options: &Option<TableParquetOptions>,
-) -> Option<WriterPropertiesFactoryRef> {
-    let props = build_writer_properties_tpo(table_parquet_options);
-    props.map(|wp| Arc::new(SimpleWriterPropertiesFactory::new(wp)) as WriterPropertiesFactoryRef)
+    table_parquet_options: &TableParquetOptions,
+) -> DeltaResult<WriterPropertiesFactoryRef> {
+    Ok(Arc::new(SimpleWriterPropertiesFactory::new(
+        build_writer_properties_tpo(table_parquet_options)?,
+    )))
+}
```

Both existing call sites (`commands_with_encryption.rs` L279, `basic_operations_encryption.rs` L287) are already in functions returning `Result<_, DeltaTableError>`, so they only need `SimpleFileFormatOptions::try_new(tbl_options)?`. `WriterPropertiesBuilder: TryFrom<&TableParquetOptions, Error = DataFusionError>` and `From<DataFusionError> for DeltaTableError` exists under the same feature gate, so `?` resolves. Both helpers are private to this file. Since both types are new in this PR, no released API breaks.

**Residual panic, upstream:** `enc.into()` goes through DataFusion's `From<ConfigFileEncryptionProperties> for FileEncryptionProperties`, which does `hex::decode(..).unwrap()` / `.expect("Invalid column encryption key")` — bad hex in user config still aborts, and DataFusion 52.1 exposes no fallible counterpart. Probably worth an upstream issue; a cheap even-length + `is_ascii_hexdigit` guard before `.into()` returning `DeltaTableError::Generic` would close it locally in the meantime.

---

### 3. Two data paths bypass the plumbing: CDF reads and the `DataSink` insert path

CDC files are **written** through the factory (`write_execution_plan_cdc` → `WriterConfig`), so with encryption enabled `_change_data` files get encrypted — but `CdfLoadBuilder::build` builds its `TableParquetOptions` from session defaults only (`load_cdf.rs` L415–418) and never calls `update_session`, so encrypted CDF output can never be read back, silently. We'd suggest wiring it rather than erroring out: everything needed is already on the snapshot, the pattern is a verbatim copy of `DeltaScanBuilder::build`, and an explicit "not supported" error would still need the same lookup to detect the encrypted case — same cost, less delivered.

`crates/core/src/operations/load_cdf.rs`, L348 and L415–425:

```diff
@@ -345,7 +345,15 @@ impl CdfLoadBuilder {
         session: &dyn Session,
         filters: Option<&Arc<dyn PhysicalExpr>>,
     ) -> DeltaResult<Arc<dyn ExecutionPlan>> {
-        let snapshot = resolve_snapshot(&self.log_store, self.snapshot.clone(), true, None).await?;
+        // Preserve the table's load config (and therefore its file format options)
+        // when the builder has to materialize its own snapshot.
+        let base_config = self.snapshot.as_ref().map(|s| s.load_config());
+        let snapshot = resolve_snapshot_with_config(
+            &self.log_store,
+            self.snapshot.clone(),
+            true,
+            None,
+            base_config,
+        )
+        .await?;
         PROTOCOL.can_read_from(&snapshot)?;
@@ -415,17 +423,36 @@ impl CdfLoadBuilder {
-        let parquet_options = TableParquetOptions {
-            global: session.config().options().execution.parquet.clone(),
-            ..Default::default()
-        };
-
-        let mut cdc_source = ParquetSource::new(cdc_table_schema)
-            .with_table_parquet_options(parquet_options.clone());
-        let mut add_source = ParquetSource::new(add_table_schema)
-            .with_table_parquet_options(parquet_options.clone());
-        let mut remove_source =
-            ParquetSource::new(remove_table_schema).with_table_parquet_options(parquet_options);
+        // Reuse the table's file format options so encrypted tables stay readable
+        // through the CDF path (mirrors `DeltaScanBuilder::build`).
+        let file_format_options = snapshot.load_config().file_format_options.clone();
+        if let Some(ffo) = &file_format_options {
+            ffo.update_session(session)?;
+        }
+        let parquet_options = file_format_options
+            .as_ref()
+            .map(|ffo| ffo.table_options().parquet)
+            .unwrap_or_else(|| TableParquetOptions {
+                global: session.config().options().execution.parquet.clone(),
+                ..Default::default()
+            });
+
+        let encryption_factory = parquet_options
+            .crypto
+            .factory_id
+            .as_ref()
+            .map(|factory_id| session.runtime_env().parquet_encryption_factory(factory_id))
+            .transpose()?;
+
+        let new_source = |table_schema: TableSchema| {
+            let source = ParquetSource::new(table_schema)
+                .with_table_parquet_options(parquet_options.clone());
+            match &encryption_factory {
+                Some(factory) => source.with_encryption_factory(Arc::clone(factory)),
+                None => source,
+            }
+        };
+
+        let mut cdc_source = new_source(cdc_table_schema);
+        let mut add_source = new_source(add_table_schema);
+        let mut remove_source = new_source(remove_table_schema);
```

(`resolve_snapshot_with_config` is already `pub(crate) use`d from `crate::kernel`; `EagerSnapshot::load_config()` is `pub`; the trait-object calls compile the same way `optimize.rs` L1179/L1187 already do.)

**The `DataSink` insert path has the same hole today, on this branch**: `DeltaDataSink::write_all` passes `None` as `writer_properties_factory` to `WriterConfig::new` (`data_sink.rs` L118–129), so `INSERT INTO` on an encryption-configured table silently writes **plaintext** files — a data-protection failure rather than an error. The sink owns an `EagerSnapshot`, so the fix is one expression, same idiom as `DeleteBuilder`/`WriteBuilder`:

```diff
@@ -118,7 +118,11 @@ impl DataSink for DeltaDataSink {
         let config = WriterConfig::new(
             self.snapshot.read_schema(),
             partition_columns.clone(),
-            None,
+            self.snapshot
+                .load_config()
+                .file_format_options
+                .as_ref()
+                .map(|ffo| ffo.writer_properties_factory()),
             Some(table_props.target_file_size().get() as usize),
             None,
             table_props.num_indexed_cols(),
```

The `encryption_tests!` macro covers create/read/optimize/update/delete/merge but not `INSERT INTO` or CDF reads — which is exactly why these slipped through; adding those two cases would lock this in.

Also worth a docs note for users: min/max statistics of encrypted columns still land in cleartext in the Delta log — expected at this layer (that's what the protocol RFC in delta-io/delta#6195 addresses), but users should know.

---

### 4. Redact key material from `Debug`

`SimpleFileFormatOptions` derives `Debug` over `TableOptions`, whose `parquet.crypto.file_encryption` / `file_decryption` carry a plaintext `footer_key_as_hex: String` and per-column keys; `SimpleWriterPropertiesFactory` likewise derives `Debug` over a `WriterProperties` that can carry `FileEncryptionProperties`. Both live inside `DeltaTableConfig` (derived `Debug`), reachable from `Snapshot`/`EagerSnapshot` (both `Debug`) — and there is already a live log site that prints it: `trace!("checking to see if {snapshot:?} can write timestampntz")` (`kernel/transaction/protocol.rs` L104). `KmsFileFormatOptions` already does the right thing with a manual redacting impl; the two `Simple*` types need the same treatment. Allowlisting chosen non-secret fields (rather than skipping known-bad ones) also keeps future DataFusion/parquet field additions from leaking:

```rust
impl std::fmt::Debug for SimpleFileFormatOptions {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        // Never print `table_options`: `parquet.crypto` carries raw hex key material.
        let crypto = &self.table_options.parquet.crypto;
        f.debug_struct("SimpleFileFormatOptions")
            .field("encryption_factory_id", &crypto.factory_id)
            .field("has_file_encryption", &crypto.file_encryption.is_some())
            .field("has_file_decryption", &crypto.file_decryption.is_some())
            .finish_non_exhaustive()
    }
}
```

```rust
impl std::fmt::Debug for SimpleWriterPropertiesFactory {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        // `WriterProperties` derives Debug over `FileEncryptionProperties`, which
        // prints footer and per-column keys as raw bytes.
        f.debug_struct("SimpleWriterPropertiesFactory")
            .field("created_by", &self.writer_properties.created_by())
            .field("writer_version", &self.writer_properties.writer_version())
            .field(
                "has_file_encryption",
                &self
                    .writer_properties
                    .file_encryption_properties()
                    .is_some(),
            )
            .finish_non_exhaustive()
    }
}
```

Both traits require `Debug` as a supertrait, so the manual impls satisfy the bounds; only the derive lists change (combined with finding 2, `#[derive(Clone, Debug, Default)]` → `#[derive(Clone)]`).

---

### 5. `JsonWriter` should derive writer properties from the partition-stripped schema

Two of the three `create_writer_properties` call sites pass the physical file schema — `record_batch.rs` L215 passes `arrow_schema_without_partitions(...)` and `write/writer.rs` L370 passes `config.file_schema` — while `json.rs` L357 passes the full logical schema even though the file it then opens is built from the stripped `schema` (L346, L359). A KMS `EncryptionFactory` derives per-column keys from `file_schema`, so the JSON writer currently asks for keys for columns that will never exist in the file, and the three writers disagree about what "file schema" means.

`crates/core/src/writer/json.rs` L357:

```suggestion
                        .create_writer_properties(&path, &schema)
```

(`schema` at L346 is the stripped `ArrowSchemaRef`, matching the trait signature; the borrow ends before `schema` moves into `DataArrowWriter::new` at L359.)

---

### Sequencing vs #4568

#4568 introduces the `datafile` seam (`DataFileWriter`/`DataFileReader`) that is the structural fix for the root cause behind findings 3 and 5 — but it's cut from `main` and contains zero references to `file_format_options`, so the two PRs will conflict in exactly the files above. Our suggestion: land this PR first (it's additive and defines the option surface that #4568's `ReaderProperties {}` placeholder wants), then rebase #4568 so `ReaderProperties` carries `Option<FileFormatRef>` and centralizes the `TableParquetOptions` + `EncryptionFactory` wiring in one place — at which point the per-call-site wiring from findings 1 and 3 gets deleted rather than duplicated a fourth time. If #4568 is judged closer to merge instead, this PR should re-target the `datafile` seam — either way, deciding before the next rebase saves a full rework.

One more thing the #4568 seam would also absorb: `get_file_decryption_properties` (`operations/optimize.rs` L1176–1196) constructs a fresh `SessionContext::default()` **per file** during compaction purely to look up the encryption factory, discarding the caller's session and its registry — O(files) session construction on the hot compaction path; at minimum worth hoisting out of the loop.

---

Happy to re-test all of this against a real S3 + AWS KMS setup once the branch is rebased (it's currently conflicting with `main` and the Rust CI matrix hasn't run on the current HEAD). Hope this helps with the second review this PR has been waiting for.

<sub>Per the contributing guide's note on AI usage: AI tooling assisted this analysis; we reviewed and verified every finding and diff against the branch ourselves and stand behind them.</sub>

corwinjoy:
We're abandoning this PR because it ended up having to touch too many places. Unfortunately, with the current delta-rs structure, there is no way to avoid that because there are 6-7 different read/write paths that need to be accounted for. So, we're trying a more staged approach:
### Stage 1 — Refactor the read/write interface
- Consolidate and refactor the read / write paths behind a more consistent interface. #4568 . This was the approach we discussed with @rtyler at the recent conference and in Slack: https://delta-users.slack.com/archives/C013LCAEB98/p1781571891482589
Still waiting on a review. (The hope was for simplification but it still touches a number of places).

The further stages are on my fork waiting on this first step. 

### Stage 2 — Read/write properties (encryption foundation)
  - [corwinjoy#39 — feat(encryption): foundation types — `EncryptionConfig` and `WriterPropertiesFactory`](https://github.com/corwinjoy/delta-rs/pull/39)
    Introduces the configuration types for reader/writer properties that later stages thread through the consolidated read/write paths.

### Stage 3 — Columnar encryption
  - [corwinjoy#40 — feat(encryption): write path — factory threading, KMS resolver, global registry](https://github.com/corwinjoy/delta-rs/pull/40)
  - [corwinjoy#41 — feat(encryption): read path — DataFusion scan decryption, optimize, full round-trip tests](https://github.com/corwinjoy/delta-rs/pull/41)

Issue #3780, opened 2025-09-22
post_commithook_properties checkpoint flag is not honored


havrylenkok:
# Environment

**Delta-rs version**: 1.1.4

**Binding**: 

**Environment**:
- **Cloud provider**: AWS
- **OS**:
- **Other**: python 3.11, table is written to S3 with DDB locking

***
# Bug

**What happened**:

```
write_deltalake(
    dt,
    arrow_table,
    mode="append",
    schema_mode=None,
    storage_options=storage_options,
    writer_properties=writer_properties,
    post_commithook_properties=PostCommitHookProperties(create_checkpoint=False),
    partition_by=["partition_date"],
)
```

creates checkpoints every 100 commits

**What you expected to happen**:

no automatic checkpoints are created

**How to reproduce it**:

run the snippet to write to the table and create 101 commits

**More details**:

docs on these https://delta-io.github.io/delta-rs/api/transaction/#deltalake.PostCommitHookProperties

Reading the table created with `deltalake` in S3 via Athena often doesn't work because of disagreement on contents of last checkpoint file in this library and trino https://github.com/trinodb/trino/issues/18760 

I'm trying to disable any kind of automatic checkpoint creation so then when I create them manually via `table.create_checkpoint` I fix the file in S3 immediately and do not break queries.

fvaleye:
Hi @havrylenkok 👋,
Thank you for the report.
Could you please provide a code snippet that reproduces the error you’re experiencing?

havrylenkok:
@fvaleye I tried to reproduce it in a simple script from scratch and couldn't. Neither when creating a table in local fs, nor in S3 with ddb locking and code close to what I'm using in my app.

Sorry for a not reproducible report, should be fine to close.

For posterity: in my app I can reliably see it creating checkpoints for commits 99 and 199. However after that the table, as I expected it to, continues for hundreds of commits without new checkpoints. While I reported it originally for `write_deltalake`, digging deeper into the logs, those first two checkpoints get created at some point in the middle of looping over created partitions and running `table.optimize.z_order` and `table.optimize.compact` with `partition_filters`.

fvaleye:
> @fvaleye I tried to reproduce it in a simple script from scratch and couldn't. Neither when creating a table in local fs, nor in S3 with ddb locking and code close to what I'm using in my app.
> 
> Sorry for a not reproducible report, should be fine to close.
> 
[...]

PR #3795, opened 2025-09-30
feat(tracing): add tracing spans to all I/O sections

### LINKED ISSUE #3641: Introduce tracing spans to all I/O sections
We already pull in the `tracing` library but are really only using it for logging at the moment.

For this issue, I would like each code block where we have operations hitting the object store for reads or writes to add a tracing span. This will help with larger scale performance analysis

There are a number of compute heavy sections in the `DeltaOps` module as well, those are not my current focus but a good bonus!

fvaleye:
# Description
To better understand performance in the `delta-rs` crate, I added additional tracing to capture more detailed debug-level performance information.

Python now uses `OpenTelemetry` to collect tracing data emitted from Rust.
With this change, we gain true end-to-end visibility: Python spans can serve as parents of Rust spans (and vice versa), ensuring a continuous trace across both runtimes.

# Related Issue(s)
-  close #3641

# Documentation
- https://docs.rs/tracing/latest/tracing/

### DIFF
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -227,7 +227,10 @@ async fn find_files_scan(
         .log_data()
         .iter()
         .map(|f| f.add_action())
-        .map(|add| (add.path.clone(), add.to_owned()))
+        .map(|add| {
+            let path = add.path.clone();
+            (path, add)
+        })
         .collect();
 
     Span::current().record("total_files", candidate_map.len());
@@ -256,8 +259,7 @@ async fn find_files_scan(
     let scan = Arc::new(scan);
 
     let config = &scan.config;
-    let input_schema = scan.logical_schema.as_ref().to_owned();
-    let input_dfschema = input_schema.clone().try_into()?;
+    let input_dfschema = scan.logical_schema.as_ref().to_owned().try_into()?;
 
     let predicate_expr = session
         .create_physical_expr(Expr::IsTrue(Box::new(expression.clone())), &input_dfschema)?;
@@ -329,7 +331,10 @@ async fn scan_memory_table(snapshot: &EagerSnapshot, predicate: &Expr) -> DeltaR
 
     let map = actions
         .into_iter()
-        .map(|action| (action.path.clone(), action))
+        .map(|action| {
+            let path = action.path.clone();
+            (path, action)
+        })
         .collect::<HashMap<String, Add>>();
 
     join_batches_with_add_actions(batches, map, PATH_COLUMN, false)


crates/core/src/kernel/transaction/mod.rs
```diff
@@ -619,14 +619,21 @@ impl<'a> std::future::IntoFuture for PreparedCommit<'a> {
             let commit_or_bytes = this.commit_or_bytes;
 
             if this.table_data.is_none() {
+                tracing::debug!("committing initial table version 0");
```
rtyler: :wave: in a lot of places we actually just import `use tracing::*;` which means this module prefix isn't necessary.

IMHO it's reasonable for us to pull `tracing::*` into most modules :smile:

crates/core/src/delta_datafusion/mod.rs
```diff
@@ -2388,4 +2442,86 @@ mod tests {
...
+        );
+
+        tracing::info!(
+            candidates = 12,
+            scan_type = "physical",
+            "file pruning completed"
+        );
+
+        tracing::debug!(matching_files = 12, "physical scan completed");
+    }
```
rtyler: :laughing: this test doesn't actually do anything

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -967,4 +1026,58 @@ mod tests {
...
+            "transaction committed successfully"
+        );
+        tracing::warn!(
+            base_version = 5,
+            latest_version = 7,
+            "table updated during transaction"
+        );
+        tracing::error!(version = 8, "commit failed, aborting");
+        tracing::debug!(conflicts_checked = 3, "all conflicts resolved");
+    }
```
rtyler: This test also doesn't seem to actually do anything :upside_down_face:

rtyler (dismissed):
Generally very excited to have more tracing in place. Once tests pass I'm comfortable merging this

crates/core/src/delta_datafusion/mod.rs
fvaleye: I'm working on it... I removed some examples where I tested the execution.
I wanted to see the test coverage results first.

crates/core/src/kernel/transaction/mod.rs
fvaleye: I'm working on it... I removed some examples where I tested the execution.
I wanted to see the test coverage results first.

crates/core/src/kernel/transaction/mod.rs
fvaleye: Okay! It was a way to differentiate `log` from `tracing`.

fvaleye: Just imported with `use tracing::*;`, if you found it's ok for readability.
I didn't want to mix `log` and `tracing` crates, but it seems reasonable for now! We can tighten the imports or add explicit bridging later if the project grows.

fvaleye:
> Generally very excited to have more tracing in place. Once tests pass I'm comfortable merging this

Same!

I’m re-requesting a review based on your feedback. I’m still waiting for the code coverage results, and we can proceed with a merge!

ion-elgreco:
This is stdout/err tracing only, right?

fvaleye:
> This is stdout/err tracing only, right?

I used `warning`, `info`, and `debug` levels together with `tracing`, adding `Spans` with custom fields to track performance.

`tracing` isn’t limited to stdout or stderr. It’s a subscriber-agnostic instrumentation framework that emits structured events with typed fields, not just plain text logs. 

Where these traces go is entirely up to the user: with `tracing-subscriber`, traces can be sent to distributed tracing backends like Jaeger, Zipkin, Tempo, Honeycomb, or Datadog, to OpenTelemetry collectors, to custom backends, or nowhere at all.

Example:
```
// Option 1: Console output
let subscriber = tracing_subscriber::fmt()
    .with_env_filter("deltalake=debug")
    .finish();
tracing::subscriber::set_global_default(subscriber)?;

```

**Side note:** Importantly, if no subscriber is configured, the overhead is nearly zero.

ion-elgreco:
I get that, my point is rather, how would one configure it to use push to an OTEL endpoint from python ^^

fvaleye:
> I get that, my point is rather, how would one configure it to use push to an OTEL endpoint from python ^^

Ah, sorry! 😀
Indeed, we might need to wire `tracing-opentelemetry` + `opentelemetry-otlp` so spans are pushed to an OTLP endpoint (e.g. http://localhost:4317). On the Python side, we just point OpenTelemetry SDK at the same collector. Both Rust and Python traces then land in the same backend.

In Python, we might need to add a small API like deltalake.init_tracing("otel_endpoint") that sets up the exporter inside the Rust runtime, WDYT? 

I can take time to add it, since we have many cases where the performances is evaluated from Python!

rtyler:
I'm marking this as a draft, please change that when you're ready for this to be merged @fvaleye

fvaleye:
Ready for review @rtyler @ion-elgreco 

@ion-elgreco I added what was required from Python to collect data from `tracing` crate using `OpenTelemetry`.

**Side notes:** I fixed `typos` issues, but the CI is still failling...

python/src/lib.rs
```diff
@@ -2159,6 +2162,39 @@ fn rust_core_version() -> &'static str {
...
+        let rt = Runtime::new().map_err(|e| {
+            PyRuntimeError::new_err(format!("Failed to create Tokio runtime: {}", e))
+        })?;
+        rt.block_on(async {
+            tracing_otlp::init_otlp_tracing(&endpoint).map_err(|e| {
+                PyRuntimeError::new_err(format!("Failed to initialize tracing: {}", e))
+            })
+        })
+    }
+    #[cfg(not(feature = "otel"))]
```
ion-elgreco: We don't really not feature gates for python tbh

python/src/lib.rs
```diff
@@ -2159,6 +2162,39 @@ fn rust_core_version() -> &'static str {
     deltalake::crate_version()
 }
 
+/// Initialize OpenTelemetry tracing with OTLP exporter
+#[pyfunction]
+#[pyo3(signature = (endpoint))]
+fn init_tracing(endpoint: String) -> PyResult<()> {
+    #[cfg(feature = "otel")]
+    {
+        let rt = Runtime::new().map_err(|e| {
```
ion-elgreco: We have the rt() helper function to create a runtime, can you use that instead

python/deltalake/__init__.py
```diff
@@ -16,6 +24,41 @@
...
+        deltalake.init_tracing("http://localhost:4317")
+
+        deltalake.write_deltalake("my_table", data)
+        dt = deltalake.DeltaTable("my_table")
+
+        deltalake.shutdown_tracing()
+        ```
+    """
+    _init_tracing(otel_endpoint)
+    atexit.register(_shutdown_tracing)
```
ion-elgreco, line 77: Do we still need to explicitly shutdown tracing if we have this?

python/deltalake/__init__.py
```diff
@@ -16,6 +24,41 @@
     write_deltalake,
 )
 
+
+def init_tracing(otel_endpoint: str = "http://localhost:4317") -> None:
+    """
+    Initialize OpenTelemetry tracing for delta-rs operations.
+
+    Args:
+        otel_endpoint: The OTLP gRPC endpoint URL (default: http://localhost:4317)
```
ion-elgreco: I would expect in enterprise use cases that you could define api key, can we use the normal OTEL env params for this? Or would we want to allow it to be set in init_tracing as well?

python/deltalake/__init__.py
fvaleye: Yes, you are right!

python/src/lib.rs
fvaleye: Yes, thanks!

python/deltalake/__init__.py
fvaleye: I added documentation for these use cases!

crates/core/src/tracing_otlp.rs
abhiaagarwal, line 1: Why does this live in the `deltalake_core` crate instead of the python implementation? There are more idiomatic ways to do this in rust-land, and if you've already initialized tracing in your app for other components, this will conflict

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -637,9 +644,20 @@ impl<'a> std::future::IntoFuture for PreparedCommit<'a> {
...
 
+            let commit_span = info_span!(
+                "commit_with_retries",
+                base_version = read_snapshot.version(),
+                max_retries = this.max_retries,
+                attempt = field::Empty,
+                target_version = field::Empty,
+                conflicts_checked = 0
+            );
+            let _enter = commit_span.enter();
```
abhiaagarwal: Entering a span manually doesn't really work in async code. Instead you can spawn an anonymous coroutine and await it. see: https://docs.rs/tracing/latest/tracing/span/struct.EnteredSpan.html#in-asynchronous-code

crates/core/src/tracing_otlp.rs
fvaleye: It's a valid point @abhiaagarwal, thanks!
There is no need to include this part in the `deltalake_core`, as it's more related to the Python binding. I will move this part.

crates/core/src/kernel/transaction/mod.rs
fvaleye: Good catch, I will rework this!

abhiaagarwal: thanks for the quick responses :) I've been instrumenting delta in production through hacky means, very excited to get more metrics when this lands!

ion-elgreco:
@fvaleye is this in a state it can be reviewed again?

fvaleye:
> @fvaleye is this in a state it can be reviewed again?

Yes!

crates/core/src/logstore/mod.rs
```diff
@@ -706,21 +718,28 @@ pub async fn write_commit_entry(
         .map_err(|err| -> TransactionError {
             match err {
                 ObjectStoreError::AlreadyExists { .. } => {
+                    warn!("commit entry already exists");
```
roeap, line 720: does this need to be a warning? conflicting commits might be considered something we expect in the "normal" flow of things.

crates/core/src/kernel/transaction/mod.rs
ion-elgreco: Can we do it now to be more explicit or is tracing a very small library?

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -619,14 +619,21 @@ impl<'a> std::future::IntoFuture for PreparedCommit<'a> {
...
-                    create_checkpoint: false,
-                    cleanup_expired_logs: None,
+                    create_checkpoint: this
+                        .post_commit
+                        .map(|v| v.create_checkpoint)
+                        .unwrap_or_default(),
+                    cleanup_expired_logs: this
+                        .post_commit
+                        .map(|v| v.cleanup_expired_logs)
+                        .unwrap_or_default(),
```
ion-elgreco: This shouldn't be changed

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -637,102 +644,158 @@ impl<'a> std::future::IntoFuture for PreparedCommit<'a> {
...
-
-                    // Need to check for conflicts with each version between the read_snapshot and
-                    // the latest!
-                    while steps != 0 {
-                        let summary = WinningCommitSummary::try_new(
-                            this.log_store.as_ref(),
-                            latest_version - steps,
-                            (latest_version - steps) + 1,
-                        )
+            async move {
```
ion-elgreco, line 650: The additional async move doesn't seem necessary, why this change?

crates/core/src/logstore/mod.rs
```diff
@@ -680,19 +680,31 @@ pub async fn get_latest_version(
...
+            let bytes = res.bytes().await?;
+            debug!(size = bytes.len(), "commit entry read successfully");
+            Ok(Some(bytes))
+        }
+        Err(ObjectStoreError::NotFound { .. }) => {
+            debug!("commit entry not found");
+            Ok(None)
+        }
+        Err(err) => {
+            error!(error = %err, "failed to read commit entry");
```
ion-elgreco: Does it make sense to add the version number into these logs or is it done already?

crates/core/src/logstore/storage/runtime.rs
```diff
@@ -185,6 +186,7 @@ impl<T: ObjectStore + Clone> std::fmt::Display for DeltaIOStorageBackend<T> {
 
 #[async_trait::async_trait]
 impl<T: ObjectStore + Clone> ObjectStore for DeltaIOStorageBackend<T> {
+    #[instrument(skip(self, bytes), fields(path = %location, size = bytes.content_length()))]
```
ion-elgreco: This wrapper is almost never hit at this point, maybe we need a simple wrapper during storage creation that wraps and instruments these methods.

I also wonder if the object store doesn't have some native tracing?

crates/core/src/operations/write/mod.rs
ion-elgreco, line 1: Why are there changes being made here? Seems like a wrong resolution of merge conflict

crates/core/src/operations/write/writer.rs
```diff
@@ -418,19 +424,24 @@ impl PartitionWriter {
...
+            part_count += 1;
         }
 
         if !buffer.is_empty() {
             let upload_future = multi_part_upload.put_part(buffer.into());
             tasks.spawn(upload_future);
+            part_count += 1;
         }
 
+        debug!(parts = part_count, "uploading multipart file");
```
ion-elgreco: I guess the same here, should the path be mentioned for context?

crates/core/src/operations/write/mod.rs
```diff
@@ -416,328 +417,340 @@ impl std::future::IntoFuture for WriteBuilder {
...
-            let (predicate_str, predicate) = match this.predicate {
-                Some(predicate) => {
-                    let pred = match predicate {
-                        Expression::DataFusion(expr) => expr,
-                        Expression::String(s) => {
-                            let df_schema = DFSchema::try_from(schema.as_ref().to_owned())?;
-                            parse_predicate_expression(&df_schema, s, &state)?
+                let source = LogicalPlan::Extension(Extension {
+                    node: Arc::new(MetricObserver {
+                        id: "write_source_count".into(),
```
roeap, line 556: this has likely nothing todo with this PR, but there should be an const specified somewhere with this value. Since the planners do some internal matching on these ids, this can easily get out of sync... could we as a drive-by replace this, or add a TODO comment?

python/pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = ["arro3-core>=0.5.0", "deprecated>=1.2.18"]
 [project.optional-dependencies]
 pandas = ["pandas"]
 pyarrow = ["pyarrow>=16"]
+opentelemetry = ["opentelemetry-sdk>=1.20.0", "opentelemetry-api>=1.20.0"]
```
ion-elgreco: If this is just a dev dependency, we should move it to the uv dev dependency group

python/README.md
```diff
@@ -38,21 +38,48 @@ NOTE: official binary wheels are linked against openssl statically for remote
 objection store communication. Please file Github issue to request for critical
 openssl upgrade.
 
+## Tracing and Observability
```
roeap, line 41: I do agree that it is high time to update our README, but is this critical enough for people coming in, or should this be in the docs (or both?)

python/tests/test_tracing.py
```diff
@@ -0,0 +1,206 @@
...
+        }
+    )
+
+    table_path = str(tmp_path / "test_table")
+
+    with tracer.start_as_current_span("delta_write_operation") as span:
+        span.set_attribute("table.path", table_path)
+        span.set_attribute("operation.mode", "overwrite")
+
+        write_deltalake(table_path, sample_data, mode="overwrite")
```
ion-elgreco, line 67: I guess a follow up pr could be to do this for users directly?

python/README.md
```diff
@@ -38,21 +38,48 @@ NOTE: official binary wheels are linked against openssl statically for remote
...
-```sh
+````sh
 curl https://sh.rustup.rs -sSf | sh -s
 
 Then you can build wheels for your own platform like so:
 
 ```sh
 uvx maturin build --release --out wheels
-```
+````
```
roeap, line 82: I think we have some backticks too many?

crates/core/src/kernel/transaction/mod.rs
fvaleye: Oh, it was part of another [investigation](https://github.com/delta-io/delta-rs/issues/3780)!
Thanks, removing it 😄

python/pyproject.toml
fvaleye: `OpenTelemetry` is not a development dependency. It's more of an optional dependency that must be installed if the user wants to enable Rust tracing from Python. It is gated, meaning it’s only required when tracing functionality is activated...

python/tests/test_tracing.py
fvaleye: Yes, good idea! 👍

python/README.md
fvaleye: Not critical, but I wanted to create a section to improve our issue resolution by adding instructions on how to enable tracing. This would provide additional context when investigating issues.
I can definitely move this somewhere else.

python/README.md
fvaleye: Yes 😄

crates/core/src/operations/write/mod.rs
fvaleye: Right, let me take a look!

crates/core/src/kernel/transaction/mod.rs
fvaleye: It's more for adding an instrument to this part. instrument (commit_span).await
It ensures the span is correctly attached to the future and doesn't cause issues with await points.
But, I can extract the retry logic into a separate function and use `#[instrument]` macro on it, which would be cleaner.

crates/core/src/logstore/mod.rs
fvaleye: Yes, good idea 👍 
It's actually tracked: `version = version`.
However, we could also add it to the log.

crates/core/src/logstore/mod.rs
fvaleye: Good question!
I wanted to inform the user about the `AlreadyExists` error encountered, which falls between `info` and `warn`.

crates/core/src/logstore/storage/runtime.rs
fvaleye: Good point!
I don't know, but we could add an issue to track this need.

crates/core/src/operations/write/writer.rs
fvaleye: Yes, good!

crates/core/src/operations/write/mod.rs
fvaleye: Actually, there is already a const `SOURCE_COUNT_ID`, adding it!

crates/core/src/operations/write/mod.rs
fvaleye: Changed it, but there is a significant change due to the async traced block.

crates/core/src/kernel/transaction/mod.rs
ion-elgreco: Ah ok, yeah the diff was so big I didn't see that

crates/core/src/logstore/storage/runtime.rs
ion-elgreco: @fvaleye yeah sound goods

python/pyproject.toml
ion-elgreco: Hmm so the rust _init_tracing doesn't work if you don't have the python dependency installed?

crates/core/src/logstore/storage/runtime.rs
```diff
@@ -15,6 +15,7 @@ use object_store::{
 use object_store::{MultipartUpload, PutMultipartOptions};
 use serde::{Deserialize, Serialize};
 use tokio::runtime::{Builder as RuntimeBuilder, Handle, Runtime};
+use tracing::*;
```
ion-elgreco: Unused import warning I can see from the CI

fvaleye: Thanks for catching this!

python/pyproject.toml
fvaleye: You're right! The Rust `init_tracing` works without the Python OpenTelemetry packages. We don’t need to install anything extra to use `deltalake.init_tracing()`.

It's only required for running the OpenTelemetry tests, and they’re already included in the dev dependency group, so we are good!

Thanks!

crates/core/src/kernel/transaction/mod.rs
fvaleye: Normal, with all these whitespaces 😄

ion-elgreco:
> Just flushing some first pass comments. awesome stuff!!
> 
> For everyone reviewing this, I recommend disabling whitespace changes.

Yeah that definetly helps, should have enabled that earlier :P

python/pyproject.toml
ion-elgreco: Can you remove it from the toml dependencies then? :)

Issue #3733, opened 2025-09-03
[Epic]: Kernel based table scans


roeap:
> [!IMPORTANT]
> This is a living issue to track the work required to fully kernalize our scan planning. New sub-issues might be added as out pathfinding continues.

# Description

Delta-rs recently migrated to kernel based log replay where we use delta-kernel-rs for log replay. After log replay we have access to all relevant file actions. Currently we use this data to manually plan our table scans within the `TableProvider` implementations. This however means that we need to still interpret much of the delta metadata to enable specific features.

Kernel exposes a more generic API to plan scans which handles much of the metadata interpretation for us. We should refactor our scans such that we take advantage of these more powerful APis.

Once we have an up-to-date snapshot of the table (embedded in out snapshots), we can start planning a scan. This is a two step process.

1. Invoke `scan_metadat[_from]` on the snapshot to get the latest log state
2. Visit the log data to get read tasks

There are a few related issues we likely need to tackle for this as well.

* per-table object store handling
* ...

WE are also faced with an addition issue, which is snapshot freshness. As we execute our operations, we create ad-how datafusion sessions to run the query. However if our table provider is used within an external session, it might become stale, as we do not ensure latest state during the scan. There are several ways to mitigate this some going a bit further as in managing this via datafusion catalog abstractions.

A rough sketch of where we ned to end up.

1. get and process the kernel metadata to get the raw scan plan. below is a version from the [deltalake-datafusion](https://github.com/delta-incubator/deltalake-datafusion) crate which does not take advantage yet of our materialised/cached state.

```rs
async fn scan_metadata(
    engine: Arc<dyn Engine>,
    snapshot: &Arc<Snapshot>,
    projection: Option<&Vec<usize>>,
    predicate: Arc<Predicate>,
    table_schema: &ArrowSchemaRef,
) -> Result<TableScan> {
    // set up the scan with projection an predicate
    let projected_delta_schema = project_delta_schema(table_schema, snapshot.schema(), projection);
    let scan = snapshot
        .clone()
        .scan_builder()
        .with_schema(projected_delta_schema)
        .with_predicate(predicate)
        .build()
        .map_err(to_df_err)?;

    let table_root = scan.table_root().clone();
    let physical_schema: ArrowSchemaRef =
        Arc::new(scan.physical_schema().as_ref().try_into_arrow()?);
    let physical_predicate = scan.physical_predicate();
[...]

PR #3796, opened 2025-10-01
refactor: use EagerSnapshot in datafusion module

roeap:
# Description

While migrating to kernel log replay we took on a lot of tech dept, that we no need to clean up :).

One reason for bloat and is the similar nature of `DeltaTableState`, `EagerSnapshot`, and `Snapshot`. In this PR we reduce the API surface that use `DeltaTableState` in favour of using `EagerSnapshot`. While we still require some pathfinding, the most likely candidate to consolidate is using `EagerSnapshot` and getting rid of the others.

Almost all operations are migrated, except `Vacuum` which would have required too much changes to business logic and will be migrated later.

# Related Issue(s)

related #3733

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -68,7 +68,7 @@ tokio = { workspace = true, features = [
 ] }
 
 # caching
-foyer = { version = "0.17.2", optional = true, features = ["serde"] }
+foyer = { version = "0.20.0", optional = true, features = ["serde"] }
 tempfile = { version = "3.19.1", optional = true }
 
 # other deps (these should be organized and pulled into workspace.dependencies as necessary)


crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -652,22 +652,38 @@ impl EagerSnapshot {
...
+    ) -> BoxStream<'_, DeltaResult<LogicalFileView>> {
+        if filters.is_empty() {
+            return self.files(log_store, None);
+        }
+        let predicate = match to_kernel_predicate(filters, self.snapshot.schema()) {
+            Ok(predicate) => Arc::new(predicate),
+            Err(err) => return Box::pin(futures::stream::once(async { Err(err) })),
+        };
+        self.files(log_store, Some(predicate))
+    }
```
roeap, line 683: this is a migration helper. We should move the conversion of `PartitionFilter` to predicate either to the python crate, or even better translate DNF directly to kernel predicates. Getting rid of this completely would have required a much larger change.

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -488,13 +485,12 @@ impl<'a> DeltaScanBuilder<'a> {
...
         };
 
-        let context = SessionContext::new();
         let df_schema = logical_schema.clone().to_dfschema()?;
 
         let logical_filter = self
             .filter
             .clone()
-            .map(|expr| simplify_expr(&context, &df_schema, expr));
+            .map(|expr| simplify_expr(self.session, &df_schema, expr));
```
roeap, line 493: This is a drive-by fix and a good example of what @rtyler raised w.r.t. LogStores. We were creating a new session while also tracking a session on the operation. i.e. using inconsistent datafusion sessions in the same operation.

crates/core/src/table/state.rs
```diff
@@ -218,6 +222,10 @@ impl DeltaTableState {
     /// ## Returns
     ///
     /// A stream of logical file views that match the partition filters.
+    #[deprecated(
+        since = "0.30.0",
+        note = "Use `.snapshpt().files(log_store, predicate)` with a kernel predicate instead."
```
ion-elgreco:
```suggestion
        note = "Use `.snapshot().files(log_store, predicate)` with a kernel predicate instead."
```

ion-elgreco (dismissed):

crates/benchmarks/src/bin/merge.rs
```diff
@@ -199,7 +199,7 @@ async fn benchmark_merge_tpcds(
     let table = DeltaTableBuilder::from_uri(table_url)?.load().await?;
 
     let provider = DeltaTableProvider::try_new(
-        table.snapshot()?.clone(),
+        table.snapshot()?.snapshot().clone(),
```
rtyler, line 202: :laughing:

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -652,22 +652,36 @@ impl EagerSnapshot {
...
-            files
+            }
         } else {
             self.files.clone()
         };
         let iter = (0..data.num_rows()).map(move |i| Ok(LogicalFileView::new(data.clone(), i)));
         futures::stream::iter(iter).boxed()
     }
 
+    #[deprecated(since = "0.30.0", note = "Use `files` with kernel predicate instead.")]
```
rtyler, line 669: :laughing: what version number is this?

crates/core/src/logstore/mod.rs
```diff
@@ -341,7 +341,7 @@ pub trait LogStore: Send + Sync + AsAny {
 
     #[deprecated(
         since = "0.1.0",
```
rtyler, line 343: :laughing:

crates/core/src/operations/cdc.rs
```diff
@@ -113,7 +113,8 @@ mod tests {
             .await
             .expect("Failed to make a table");
         table.load().await.expect("Failed to reload table");
-        let result = should_write_cdc(table.snapshot().unwrap()).expect("Failed to use table");
+        let result =
+            should_write_cdc(&table.snapshot().unwrap().snapshot).expect("Failed to use table");
```
rtyler: When we were hacking yesterday this kind of accessing into the `snapshot` field of the `EagerSnapshot` is something we should probably be removing.

Is it possible to put a `#[cfg(test)]` around that to keep its isolation into tests?

crates/core/src/kernel/snapshot/mod.rs
roeap: was hoping the net one we release 😆

crates/core/src/logstore/mod.rs
roeap: well ... i feel like we knew this was coming early on.

roeap:
@rtyler @ion-elgreco - addressed your feedback and would appreciate some new stamps :).

crates/core/src/operations/cdc.rs
roeap: I was able to get rid of some more call sites of this bit not all. its `pub(crate)` for now and should go away as we consolidate.

ion-elgreco (approved):
☑️

Issue #3767, opened 2025-09-18
larger table causes overflow deep in arrow land


rtyler:
# Environment

**Delta-rs version**: 0.28.1

**Binding**: 🦀 

***
# Bug

**What happened**:

```
thread 'main' (3742440) panicked at /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-array-55.2.0/src/builder/generic_bytes_builder.rs:164:35:
attempt to add with overflow
```

This occurs when opening a table that has ~5.9GB of transaction log data, which I don't consider particularly large 😄 

**What you expected to happen**:

Well, not panic obviously 😆 

**How to reproduce it**:

Work at my company, and then open this specific table 😆  The backtrace shows:

```
thread 'main' (3742440) panicked at /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-array-55.2.0/src/builder/generic_bytes_builder.rs:164:35:
attempt to add with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/69b76df90c7ea63b5350d1865f92902a0b27c9a2/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/69b76df90c7ea63b5350d1865f92902a0b27c9a2/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_add_overflow
             at /rustc/69b76df90c7ea63b5350d1865f92902a0b27c9a2/library/core/src/panicking.rs:175:17
   3: <i32 as core::ops::arith::Add>::add
             at /rustc/69b76df90c7ea63b5350d1865f92902a0b27c9a2/library/core/src/ops/arith.rs:106:45
   4: arrow_array::builder::generic_bytes_builder::GenericByteBuilder<T>::append_array
             at /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-array-55.2.0/src/builder/generic_bytes_builder.rs:164:35
   5: arrow_select::concat::concat_bytes
             at /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-select-55.2.0/src/concat.rs:239:17
   6: arrow_select::concat::concat
             at /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-select-55.2.0/src/concat.rs:425:27
[...]

PR #3801, opened 2025-10-02
fix: avoid overflow for large table state

roeap:
# Description

When loading the active add files into memory, we concatenate the batches read from the log. For very large logs, we may exceed the admissible size for an individual array, specifically likely for large stats fields.

@rtyler - mind checking out if this fixes the issue we see on large tables? And, do we have an issue for this?


rtyler:
#3767

rtyler:
I can confirm that this does not exhaust memory on said table:

roeap:
@rtyler - the PR grew a bit in size, but hopefully for good cause. We may now see that we are no longer using as much memory, since we are no longer tracking the serialised stats as part of the file data.

Would you mind confirming? 😄

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -651,9 +649,21 @@ impl<'a> DeltaScanBuilder<'a> {
...
+                let batch_mask = &mask[mask_offset..mask_offset + batch_size];
+                let batch_mask_array = BooleanArray::from(batch_mask.to_vec());
+                let pruned_batch = filter_record_batch(batch, &batch_mask_array)?;
+                if pruned_batch.num_rows() > 0 {
+                    pruned_batches.push(pruned_batch);
+                }
+                mask_offset += batch_size;
+            }
+
+            LogDataHandler::new(&pruned_batches, es.table_configuration()).statistics()
```
roeap, line 666: This was not really nice before and unfortunately got a bit less nice.

Longer therm i think we may have to decide if we need additional skipping from datafsuion, or rely on the file skipping in delta-kernel to be selective (we have no reason to believe it would not be :)).

crates/core/src/kernel/arrow/engine_ext.rs
```diff
@@ -150,14 +150,15 @@ impl SnapshotExt for Snapshot {
     /// scan row (file data).
     fn scan_row_parsed_schema_arrow(&self) -> DeltaResultLocal<ArrowSchemaRef> {
         let mut fields = SCAN_ROW_ARROW_SCHEMA.fields().to_vec();
+        let stats_idx = SCAN_ROW_ARROW_SCHEMA.index_of("stats").unwrap();
 
         let stats_schema = self.stats_schema()?;
         let stats_schema: ArrowSchema = stats_schema.as_ref().try_into_arrow()?;
-        fields.push(Arc::new(Field::new(
+        fields[stats_idx] = Arc::new(Field::new(
```
roeap, line 151: we are now replacing the existing stats field with the stats_parsed rather than amending the parsed ones.

crates/core/src/kernel/snapshot/iterators/scan_row.rs
roeap, line 1: While not entirely clean yet, we aim to isolate processing of the data we get from kernels log replay in this module. Essentially we need to revert when we do when receiving data when we feed it back into a scan / replay.

crates/core/src/kernel/snapshot/iterators/scan_row.rs
```diff
@@ -0,0 +1,455 @@
...
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.project();
+        match this.stream.poll_next(cx) {
+            Poll::Ready(Some(Ok(batch))) => match parse_stats_column(&this.snapshot, &batch) {
+                Ok(batch) => Poll::Ready(Some(Ok(batch))),
+                Err(err) => Poll::Ready(Some(Err(err))),
+            },
+            other => other,
+        }
+    }
```
roeap, line 60: it seems work has started to support async/streams directly from kernel. As such we start to move some processing onto streams rather that doing it in iterator world.

This should also align well when we work on the datafusion integrations, since there we find the same model.

crates/core/src/kernel/snapshot/iterators.rs
```diff
@@ -158,12 +157,12 @@ impl LogicalFileView {
...
-            self.files
-                .column(*FIELD_INDICES.get(FIELD_NAME_STATS).unwrap()),
-            self.index,
-        )
+    pub fn stats(&self) -> Option<String> {
+        let stats = self.stats_parsed()?.slice(self.index, 1);
+        let value = to_json(&stats)
+            .ok()
+            .map(|arr| arr.as_string::<i32>().value(0).to_string());
+        value.and_then(|v| (!v.is_empty()).then_some(v))
```
roeap, line 165: we now need to serialise individual fields to get the json stats (for add actions) since we are no longer stacking the stats column.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -602,81 +631,34 @@ impl EagerSnapshot {
...
     ///
     /// * `log_store` - A reference to a [`LogStore`] implementation.
     /// * `predicate` - An optional predicate to filter the files.
     ///
-    /// # Returns
+    /// ## Returns
     ///
     /// A stream of [`LogicalFileView`] objects.
-    pub fn files(
+    pub fn file_views(
```
roeap, line 642: renamed this for consistency since it is returning file views after all.

crates/core/src/table/state.rs
```diff
@@ -347,13 +350,18 @@ impl EagerSnapshot {
...
             ARROW_HANDLER.new_expression_evaluator(input_schema, expression.into(), table_schema);
-        let result = evaluator.evaluate_arrow(actions)?;
+
+        let results = self
+            .files
+            .iter()
+            .map(|file| evaluator.evaluate_arrow(file.clone()))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let result = concat_batches(results[0].schema_ref(), &results)?;
```
roeap, line 368: we still concatenate the add actions table. in a follow-up we should also move this to a stream and expose that via record batch readers in python.

Opened #3811 to track this.

crates/core/Cargo.toml
```diff
@@ -80,6 +80,7 @@ indexmap = "2.2.1"
 itertools = "0.14"
 parking_lot = "0.12"
 percent-encoding = "2"
+pin-project-lite = "^0.2.7"
```
rtyler, line 83: I'm not sure why this dependency crept back in, I'll just have to remove it again :laughing:

crates/core/src/kernel/snapshot/iterators/scan_row.rs
rtyler: `[citation needed]` :laughing: 

There's been a lot of talk that I have heard but I haven't seen any concrete changes, do you have some to link?

PR #3802, opened 2025-10-02
feat: allow passing a `SessionState` into a `OptimizeBuilder`

### LINKED ISSUE #3797: Allow passing `SessionState` into `OptimizeBuilder`
# Description

**Use Case**

I set a global `RuntimeEnv` with a memory pool to limit the amount of ram datafusion is allowed to use globally in my application to prevent Kubernetes OOMs since I'm working with multiple tables concurrently. This works wonderfully for `WriteBuilder` / `MergeBuilder` / `DeleteBuilder` which have a `with_session_state` builder. The one holdout is `OptimizeBuilder`, which as of #3751 allows passing a `SessionConfig`, but not a runtime env, which I'd like to globally limit the amount of ram it uses.

**Related Issue(s)**

abhi-airspace-intelligence:
# Description

This allows passing a `SessionState` into an `OptimizeBuilder`, mirroring the other operation APIs. Very useful if running concurrent delta table operations in a kubernetes pod, as you can limit Datafusion to use a certain amount of ram between all shared contexts to prevent OOMs

# Related Issue(s)

Closes #3797, effectively reverts/supersedes #3751

### DIFF
--- a/crates/core/src/operations/optimize.rs
+++ b/crates/core/src/operations/optimize.rs
@@ -27,7 +27,10 @@ use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use arrow_array::RecordBatch;
 use arrow_schema::SchemaRef as ArrowSchemaRef;
-use datafusion::prelude::SessionConfig;
+use datafusion::execution::context::SessionState;
+use datafusion::execution::memory_pool::FairSpillPool;
+use datafusion::execution::runtime_env::RuntimeEnvBuilder;
+use datafusion::execution::SessionStateBuilder;
 use delta_kernel::engine::arrow_conversion::TryIntoArrow as _;
 use delta_kernel::expressions::Scalar;
 use delta_kernel::table_properties::DataSkippingNumIndexedCols;
@@ -214,8 +217,8 @@ pub struct OptimizeBuilder<'a> {
     max_spill_size: usize,
     /// Optimize type
     optimize_type: OptimizeType,
-    /// Optional [SessionConfig] for users that want more control over the Datafusion execution
-    session_config: Option<SessionConfig>,
+    /// Datafusion session state relevant for executing the input plan
+    state: Option<SessionState>,
     min_commit_interval: Option<Duration>,
     custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
 }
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/optimize.rs
```diff
@@ -310,9 +316,9 @@ impl<'a> OptimizeBuilder<'a> {
...
     }
 
-    /// Add a custom [SessionConfig] to the optimize plan execution
-    pub fn with_session_config(mut self, config: SessionConfig) -> Self {
-        self.session_config = Some(config);
+    /// A session state accompanying a given input plan, containing e.g. registered object stores
+    pub fn with_input_session_state(mut self, state: SessionState) -> Self {
+        self.state = Some(state);
         self
     }
```
abhi-airspace-intelligence, line 324: Note that the name of this function is inconsistent across the various builder APIs, happy to open a follow-up PR to rename everything to either `with_input_session_state` or `with_session_state` and deprecate the other.

abhi-airspace-intelligence:
@ion-elgreco I'll do that in a follow-up PR to minimize blast radius!

crates/core/src/operations/optimize.rs
```diff
@@ -215,8 +218,8 @@ pub struct OptimizeBuilder<'a> {
     max_spill_size: usize,
     /// Optimize type
     optimize_type: OptimizeType,
-    /// Optional [SessionConfig] for users that want more control over the Datafusion execution
-    session_config: Option<SessionConfig>,
+    /// Datafusion session state relevant for executing the input plan
+    state: Option<SessionState>,
```
roeap, line 222: Haven't looked at all the APIs, but would we get away with tracking something like `Arc<dyn Session>` here?

roeap (changes requested):

roeap:
@abhi-airspace-intelligence - blocked the PR not b/c its a pre-requisite, just wanted to ask. If not feasible we can likely merge as is.

crates/core/src/operations/optimize.rs
abhi-airspace-intelligence: Nice, looks like that's preferred by Datafusion to pass this instead. https://docs.rs/datafusion-session/50.1.0/datafusion_session/session/trait.Session.html#migration-from-sessionstate. Let me give it a shot

abhi-airspace-intelligence: @roeap I will resolve this in a follow-up PR, since I'm unifying all the interfaces anyways

roeap (approved):

Issue #3799, opened 2025-10-02
Consolidate Datafusion session management


roeap:
# Description

Delta-rs uses several advanced datafusion features / integrations such as customising query planning, custom logical / physical plan nodes etc. Additionally we use ad-hoc session updates to e.g. register dedicated stores for tables.

Currently, we sometimes allow users to also pass in their own session or create sessions on the fly when needed.

This fragmented session management adds complexity and brittleness to our codebase. We should consolidate creating datafusion sessions or extending existing sessions with delta-rs constructs.

**Related Issue(s)**

#3797

PR #3804, opened 2025-10-03
refactor: consolidate extension planners

roeap:
# Description

We currently create a dedicated session with a dedicated planner in some operations to collect metrics. This creates a situation where - even if we pass a session - we need to recreate the session on the fly, potentially replacing an external planner and thus potentially loosing support for UDFs or other custom nodes.

This PR consolidates the various extension planner we currently have into a single planner to eventually make it easy for us to centralise session creation or allow others to extend their own planner with the delta extension planners.

cc @abhi-airspace-intelligence

# Related Issue(s)

part-of: #3799


ion-elgreco:
Correct me if I am wrong, but adding all extensions planner together will cause the plan become messy and do multiple metric counts

roeap:
> Correct me if I am wrong, but adding all extensions planner together will cause the plan become messy and do multiple metric counts

The individual planners, while all downcasting to the same `MetricObserver` node, also match on unqiue IDs on the node for their respective metrics, as long as these don't overlap, only one planner should ever return a plan..

In case there are duplicate IDs we would potentially match wrong depending on the first planner that signifies it can plan the extension node (returning `Some(...)`). However we do always plan a node at most once .

ion-elgreco:
> > Correct me if I am wrong, but adding all extensions planner together will cause the plan become messy and do multiple metric counts
> 
> The individual planners, while all downcasting to the same `MetricObserver` node, also match on unqiue IDs on the node for their respective metrics, as long as these don't overlap, only one planner should ever return a plan..
> 
> In case there are duplicate IDs we would potentially match wrong depending on the first planner that signifies it can plan the extension node (returning `Some(...)`). However we do always plan a node at most once .

Yeah, I thought our IDs werent unique but it's been a while that I've touched these extension planners

roeap:
> it's been a while that I've touched these extension planners

same 😆 

In a post kernelized world, I do believe we'll see much more of the likes though ...

ion-elgreco (approved):

PR #3806, opened 2025-10-03
ci: split out integration tests

roeap:
# Description

This is another attempt at making our CI a bit more pleasant to work with. The basic idea is - more smaller jobs as well as avoiding some redundant work.

we moved integration tests into their own job and streamlined the build job.

- `build/check`: fmt, clippy (with --test), docs, on linux only
- `build/build`: previous buid and check commands but without `--test`, all os
- `build/test`: unit tests on all os

In integration we now run a job for every crate that has integration tests. azure, aws, gcp, hdfs, lakefs

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -16,9 +16,6 @@ runs:
   using: "composite"
 
   steps:
-    - name: checkout
-      uses: actions/checkout@v5
-
     - name: Install uv
       uses: astral-sh/setup-uv@v3
 
@@ -35,3 +32,5 @@ runs:
       with:
         toolchain: ${{ inputs.rust-toolchain }}
         cache: true
+        # if we ever get rid of all the clippy warnings, we can remove this line
+        rustflags: ""
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -17,7 +17,7 @@ runs:
 
   steps:
     - name: checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     - name: Install uv
       uses: astral-sh/setup-uv@v3
@@ -27,14 +27,11 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
 
-    - name: Install Rust toolchain
-      uses: actions-rs/toolchain@v1
-      with:
-        profile: default
-        toolchain: ${{ inputs.rust-toolchain }}
-        override: true
-        components: rustfmt, clippy
+    - name: Setup sccache
+      uses: mozilla-actions/sccache-action@v0.0.9
 
-    - uses: Swatinem/rust-cache@v2
+    - name: Setup Rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@v1
       with:
-        cache-targets: false
+        toolchain: ${{ inputs.rust-toolchain }}
+        cache: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,16 +8,35 @@ on:
   merge_group:
 
 env:
-  DEFAULT_FEATURES: "azure,datafusion,s3,gcs,glue,hdfs "
-  # Disable full debug symbol generation to speed up CI build and keep memory down
-  RUSTFLAGS: -C debuginfo=line-tables-only
-  # Disable incremental builds by cargo for CI which should save disk space
-  # and hopefully avoid final link "No space left on device"
-  CARGO_INCREMENTAL: 0
+  DEFAULT_FEATURES: "azure,datafusion,s3,gcs,glue,hdfs"
   SCCACHE_GHA_ENABLED: "true"
   RUSTC_WRAPPER: "sccache"
 
 jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Setup sccache
+        uses: mozilla-actions/sccache-action@v0.0.9
+
+      - name: Setup Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+          cache: true
+
+      - name: Format
+        run: cargo fmt -- --check
+
+      - name: Run Machete to check for unused dependencies
+        uses: bnjbvr/cargo-machete@aa070c61641e482bc2b5d41c97b496ee4755bf37
+
+      - name: build and lint with clippy
+        run: cargo clippy --features ${{ env.DEFAULT_FEATURES }} --tests
+
   # run various build configurations, fmt, and clippy.
   build:
     strategy:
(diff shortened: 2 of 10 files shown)


abhiaagarwal:
I read this article the other day which might be helpful to avoid CI running out of space: https://kobzol.github.io/rust/2025/09/22/reducing-binary-size-of-rust-programs-with-debuginfo.html. You can compress debuginfo to keep tracebacks with `RUSTFLAGS="-Clink-arg=-Wl,--compress-debug-sections=zlib"`

roeap:
@abhiaagarwal - awesome, thanks! Will take a look.

@rtyler @fvaleye @ion-elgreco - not sure how far we would wnat to go (if we follow this approach at all), but for now I tried to get the failures done and reduce runtime ...

In teh current config we run only linux tests on PRs and windows / osx only on main. WIndows beings by far the biggest offender. I think here caches are kicking in more effectively and most builds are done after about 2mins. The full unit tests still take about 5 mins (10+ mins on windows).

In the many CI runs I observed, i never saw an out of memory exception happening though!

Thoughts?

.github/actions/setup-env/action.yml
```diff
@@ -16,9 +16,6 @@ runs:
   using: "composite"
 
   steps:
-    - name: checkout
-      uses: actions/checkout@v4
-
```
roeap, line 21: we need to checkout the code before we can use local actions. so this must have been unused anyhow.

.github/actions/setup-env/action.yml
```diff
@@ -27,14 +24,13 @@ runs:
...
-        profile: default
-        toolchain: ${{ inputs.rust-toolchain }}
-        override: true
-        components: rustfmt, clippy
+    - name: Setup sccache
+      uses: mozilla-actions/sccache-action@v0.0.9
 
-    - uses: Swatinem/rust-cache@v2
+    - name: Setup Rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@v1
```
roeap, line 31: The `actions-rs/toolchain` seems unmaintained albeit still widely used. This action is maintained and has the additional benefit of also setting up `Swatinem/rust-cache`. My assumtion beeing that they spend much more time thinging about how than I am hoping to invest 😆.

.github/workflows/build.yml
```diff
@@ -8,328 +8,133 @@ on:
   merge_group:
 
 env:
-  DEFAULT_FEATURES: "azure,datafusion,s3,gcs,glue,hdfs "
-  # Disable full debug symbol generation to speed up CI build and keep memory down
-  RUSTFLAGS: -C debuginfo=line-tables-only
-  # Disable incremental builds by cargo for CI which should save disk space
-  # and hopefully avoid final link "No space left on device"
-  CARGO_INCREMENTAL: 0
```
roeap, line 16: I deliberately removed most our rust flags etc. to see how well we perform out of the box. SO far we seem to be doing well.

One question came up along the way. Could it also be that `CARGO_INCREMENTAL: 0` hurts reusability of cached data? not sure what cargo / rustc are writing out in the different modes, but if its is fewer larger files vs. more smaller files (which are in total more) it feel intuitive that this might be the case.

Just speculating though :).

.github/workflows/build.yml
```diff
@@ -8,328 +8,133 @@ on:
...
+          cargo llvm-cov --features ${{ env.DEFAULT_FEATURES }} \
+            --workspace \
             --codecov \
-            --output-path codecov.json
+            --output-path codecov.json \
+            --exclude deltalake \
+            --exclude deltalake-azure \
+            --exclude deltalake-aws \
+            --exclude deltalake-hdfs \
+            --exclude deltalake-lakefs
```
roeap, line 137: I was hoping that we would not build these packages, but looking at the logs it seems we still do. This was an attempt to get build time on windows down, but seems to not have much of an impact. SO we ended at just removing them.

.github/workflows/dev_pr.yml
```diff
@@ -82,11 +82,3 @@ jobs:
...
             })
             core.setFailed(message)
-
-  typos:
-    name: Spell Check
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - name: Check spelling
-        uses: crate-ci/typos@v1
```
roeap, line 92: This one is on me ... This file triggers on the pull request target, which is why we checking the errors on the main branch and seeing failures if we changed files that had spelling errors on main, even if we fixed the errors in the PR.

Now this is in a dedicated workflow that triggers on pull request.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,207 @@
...
+          cache: true
+
+      - name: Start emulated services
+        run: docker compose up -d
+
+      - name: Run tests with native-tls
+        run: |
+          gmake setup-dat
+          cargo test \
+            --package deltalake-aws \
```
roeap, line 113: i looked through our crates and it seems that we are using a native-tls flag only in the AWS crate. We should make this a matrix if there are ever any more or I missed some here.

.github/workflows/python_build.yml
```diff
@@ -43,9 +31,6 @@ jobs:
           uv sync --no-install-project
           make check-python
 
-      - name: Check Rust
-        run: make check-rust
```
roeap, line 47: Our rust checks already check the python crate as well, so no need to repeat that check. Alternative is to also install clippy in case someone feels this check adds value.

.github/workflows/build.yml
abhiaagarwal: Incremental cargo builds don't work with sccache. So it's the right decision.

Also, I think incremental builds theoretically are non-reproducible, so it's probably correct to not run it in CI. This job automatically does it https://github.com/actions-rust-lang/setup-rust-toolchain/blob/063a3b947b5c5bf7d5f87076c3e5e9784b776aa8/action.yml#L118

roeap: Thanks @abhiaagarwal - if I read the linked code correctly, this setting is also applied internally,, so we don't need to re-set this here. Or should we also be explicitly setting this?

abhi-airspace-intelligence: The job already sets `CARGO_INCREMENTAL=0`, so you don't need to be explicitly setting it.

ion-elgreco (approved):

ion-elgreco:
> @abhiaagarwal - awesome, thanks! Will take a look.
> 
> @rtyler @fvaleye @ion-elgreco - not sure how far we would wnat to go (if we follow this approach at all), but for now I tried to get the failures done and reduce runtime ...
> 
> In teh current config we run only linux tests on PRs and windows / osx only on main. WIndows beings by far the biggest offender. I think here caches are kicking in more effectively and most builds are done after about 2mins. The full unit tests still take about 5 mins (10+ mins on windows).
> 
> In the many CI runs I observed, i never saw an out of memory exception happening though!
> 
> Thoughts? 

> I'm fine with not running windows and Mac in PRS

PR #3812, opened 2025-10-05
refactor: use EagerSnapshot in vacuum operation

roeap:
# Description

The vacuum operations was the last one still using `DeltaTableState`. Now that we have updated tombstone access, we can also migrate this one to use `EagerSnapshot`.


ion-elgreco:
@roeap it's getting difficult to keep up haha

PR #3814, opened 2025-10-05
feat: allow passing a custom runtime into various `DeltaOps`

### LINKED ISSUE #3800: Allow passing a different tokio runtime into `DeltaOps` to move task there
# Description

**Use Case**

Per datafusion's [recommendations](https://docs.rs/datafusion/latest/datafusion/#network-io-and-cpu-intensive-tasks), spawning a separate tokio runtime for CPU-bound tasks can lead to better tail latencies. Therefore, I propose that `DeltaOps`/`Merge/Write/etcBuilder` take an optional `with_runtime` parameter that allows passing an optional handle to another runtime. Then, delta-rs can spawn the task on a separate joinset, await it in there, and pass the result back to the io-bound handle via an oneshot channel.

This requires a bit more plumbing due to object_store calls inside of various executors, but that can be worked across with [SpawnedReqwestConnector](https://docs.rs/object_store/latest/object_store/client/struct.SpawnedReqwestConnector.html).

This could also potentially be benchmarked for the python impl to see if it substantially improves performance.

**Related Issue(s)**

Related to #3640.

abhiaagarwal:
# Description

This is based on some code I've written at work to execute various `DeltaOps` on a separate tokio runtime. I've noticed that this leads to fewer kubernetes kills due to the main tokio IO runtime not getting blocked on cpu-bound work , as well as reducing tail latencies.

I've only done this to `CreateBuilder` as a POC.

Marked as a draft because there are a few open questions:

- Is this the right API?
- Can we benchmark this empirically? 
- Can we override the relevant objectstore if `runtime` is set to ensure that it's using a `SpawnedReqwestConnector`? 

# Related Issue(s)

Closes #3800 

# Documentation

Vendors this example from datafusion-examples: https://github.com/apache/datafusion/tree/main/datafusion-examples/examples/thread_pools.rs

### DIFF
--- a/crates/core/src/operations/create.rs
+++ b/crates/core/src/operations/create.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 use delta_kernel::schema::MetadataValue;
 use futures::future::BoxFuture;
 use serde_json::Value;
+use tokio::runtime::Handle;
 use tracing::log::*;
 use uuid::Uuid;
 
@@ -66,6 +67,8 @@ pub struct CreateBuilder {
     commit_properties: CommitProperties,
     raise_if_key_not_exists: bool,
     custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
+    /// Optional tokio handle to execute the operation on
+    handle: Option<Handle>,
 }
 
 impl super::Operation<()> for CreateBuilder {
@@ -102,6 +105,7 @@ impl CreateBuilder {
             commit_properties: CommitProperties::default(),
             raise_if_key_not_exists: true,
             custom_execute_handler: None,
+            handle: None,
         }
     }
 
@@ -244,6 +248,12 @@ impl CreateBuilder {
         self
     }
 
+    /// Set a custom tokio handle to execute the operation on
+    pub fn with_handle(mut self, handle: Handle) -> Self {
+        self.handle = Some(handle);
+        self
+    }
+
     /// Consume self into uninitialized table with corresponding create actions and operation meta
     pub(crate) async fn into_table_and_actions(
         mut self,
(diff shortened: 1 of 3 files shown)


rtyler:
We _kind_ of have some parts of this running already already with `DeltaTableBuilder.with_io_runtime`. Basically somebody would have to do:

```rust
let table = DeltaTableBuilder::from_url(blah).with_io_runtime(other_blah).build()?
let op = DeltaOps::from(table).create().etc().etc()
```

That said, the operations aren't all using the IO runtime from the table :unamused:  so there's.. room for improvement!

abhiaagarwal:
Yep, but I think it's kind of the opposite solution that I want. It's pretty common to run IO on the main tokio runtime and spawn a dedicated CPU runtime and use it sparingly. The `with_io_runtime` pattern encourages the opposite approach.

rtyler:
@abhiaagarwal if you're going that approach, why not just spawn the tasks needing DeltaOps into that CPU intensive runtime outside of the API?

Our APIs here are already messy, I'm trying to reel that back in whenever possible :fishing_pole_and_fish:

abhiaagarwal:
I'm already doing that :) just figured I'd upstream my code to make it easier for anyone else.

If you feel this is API surface bloat, feel free to close it!

ion-elgreco:
> Yep, but I think it's kind of the opposite solution that I want. It's pretty common to run IO on the main tokio runtime and spawn a dedicated CPU runtime and use it sparingly. The `with_io_runtime` pattern encourages the opposite approach.

It's also common with datafusion to do CPU on main runtime and IO on separate, the with_io_runtime has a smaller api footprint then doing this per operation for cpu runtime

ion-elgreco:
Closing this since the inverse is possible and a normal pattern to use

PR #3816, opened 2025-10-05
Unify inconsistent SessionState 

### LINKED ISSUE #3799: Consolidate Datafusion session management

abhi-airspace-intelligence:
# Description

Unifies the SessionState management for the various DeltaOps by moving it to use the `dyn Session` interface, downcasting to `SessionContext` when relevant.

# Related Issue(s)

Closes #3799 

# Documentation

`Session` trait in datafusion: https://docs.rs/datafusion/50.0.0/datafusion/catalog/trait.Session.html


rtyler:
Marking this as draft until the conflicts are resolved :see_no_evil:

abhi-airspace-intelligence:
Should be good to go.

@roeap do you have a preference on what the var should be for datafusion state? In some places it's `session`, `state`, or `df_state`. Would prefer to have one term.

roeap (changes requested):
Looking great @abhi-airspace-intelligence. 

As for the name i think `session` is the way to go, as this is the traits name. state would be second, since we often used the implementation `SessionState`.

We also do have some fixable lints (unused imports). Granted these are hard to track in our codebase right now, since there are SOOOO many deprecations. but maybe try to fix these :). - i.e. whatever makes check fail.

I'll remove the linked issue, since I believe we should also centralise setting up the session in a dedicated function o.a. since we still need to create session in several places. Important step towards consolidation though!!

abhi-airspace-intelligence:
@roeap I will fix the lints and update the name tomorrow!

PR #3818, opened 2025-10-06
feat: shim kernel Scan and ScanBuilder

roeap:
# Description

part-of: #3733

Most friction in integrating kernel arises when we need to translate kernels blocking iterators to streams to conform to our compute model.

Thus far we had an extension trait for kernel scans to translate the scan iterators into arrow record batches. However when building our own scans, we need to process more of the scan metadata.

To facilitate implementation of kernel based scans we now provide shims around kernels `Scan` and `ScanBuilder` to expose their operations via more idiomatic abstractions without altering the returned data.


roeap:
The test coverage dropped a bit - looking at the report though, this is mainly the unused "proxy-methods" that directly delegate ti fields in the inner types. THis will get covered as we start to use these abstractions for our own scans.

rtyler (approved):

PR #3822, opened 2025-10-07
feat: expose arrow schema on snapshots

roeap:
# Description

In many places we require the arrow schema of a snapshot. As such we replicate converting schemas in many places. In this PR we track a reference to an arrow schema on our innermost snapshot and start using it in some places.

We also now expose the kernel schema as an arc to better alogn with how delta kernel and datafusion expose schemas.

part-of: #3733


ion-elgreco (approved):

roeap:
@ion-elgreco - there was some confusion with the `arrow_schema` field on the datafusion mixin. this is fixed now and we also made the schema methods on the mixin infallible.

mind stamping again? 😄

PR #3826, opened 2025-10-08
fix(performance): improve CPU and memory efficiency through several optimizations

fvaleye:
# Description
I spent some time identifying areas where we could reduce memory usage, such as avoiding unnecessary cloning or pre-allocating memory. I also analyzed CPU-intensive operations to look for potential improvements.

This work is mainly intended to spark discussion on whether small code changes could lead to performance benefits.

**1. JSON parsing:**
- Reduce String allocation by using bytes and slice operations

**2. ShareableBuffer with BytesMut**
- Mostly to use zero-copy operations (previously with intermediate `to_vec` operations)

**3. Partition processing**
- Pre-compute sorted columns once, then use `slice()` instead of `take()` per partition

**4. HashMap pre-allocation**
- Use more `HashMap::with_capacity(estimated_capacity)` when known

**5. String Clone Reduction** 

**6. Memory Take Optimization**
- Use `std::mem::take()` to move files instead of cloning

### DIFF
--- a/crates/core/src/logstore/mod.rs
+++ b/crates/core/src/logstore/mod.rs
@@ -48,7 +48,6 @@
 //! ## Configuration
 //!
 use std::collections::HashMap;
-use std::io::{BufRead, BufReader, Cursor};
 use std::sync::{Arc, LazyLock};
 
 use bytes::Bytes;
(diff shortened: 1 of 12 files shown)


crates/core/src/logstore/mod.rs
```diff
@@ -569,20 +568,27 @@ pub async fn get_actions(
...
-            line,
-            version,
-        })?;
-        actions.push(action);
-    }
-    Ok(actions)
+
+    tokio::task::spawn_blocking(move || {
+        let mut actions = Vec::new();
+        for line_bytes in commit_log_bytes.split(|&b| b == b'\n') {
```
abhi-airspace-intelligence: Has this been benchmarked? I'd imagine the overhead of transferring the actions to the current context might be higher than just leaving it as-is.

abhi-airspace-intelligence: Since this is ndjson, might be worth trying using this: https://docs.rs/serde_json/latest/serde_json/struct.StreamDeserializer.html, which would also allow returning actions as a lazy iterator.

ion-elgreco:
I think it would be more practical if we have each section in a separate pr and then some benchmarks of before and after, then it's easier to see the added benefit

crates/core/src/logstore/mod.rs
fvaleye: Yes, it would be better to adapt the `StreamDeserializer` to use `from_slice` to take advantage of serde_json's performance.

fvaleye:
> I think it would be more practical if we have each section in a separate pr and then some benchmarks of before and after, then it's easier to see the added benefit

Sure! I’d like to use this PR as an opportunity to discuss. 
I'm hoping to have `tracing` capabilities before enabling any changes. 
I’ll also add some local, isolated benchmarks to evaluate the potential benefits.

roeap:
I do remember that @rtyler recommended a crate/allocator that helps with tracking memory consumption.

rtyler:
If you have a moment to resolve the conflicts @fvaleye I am comfortable merging this change as is

fvaleye:
> If you have a moment to resolve the conflicts @fvaleye I am comfortable merging this change as is

I’m splitting the work into separate PRs to better isolate the performance improvements.
I’ve already implemented the most impactful ones, and once everything is complete, I’ll close this PR 😄
I’m keeping it open for now to track the changes I’ve investigated.

fvaleye:
Closing this PR, as all current performance improvements have been implemented.

PR #3830, opened 2025-10-09
chore(performance): optimize JSON parsing in get_actions and snapshot reading

fvaleye:
# Description
Following this [PR](https://github.com/delta-io/delta-rs/pull/3826), I took more time to investigate how to optimize the JSON parsing in `get_actions()`.
I added a bench to evaluate the performance gains by comparing the baseline to the new implementation (leveraging `Deserializer::from_slice` from serde_json).

## Benchmark results 
Only for  the`get_actions()` method
My hardware: Mac Book Pro M1
The performance boost of `13-27%` depending on JSON complexity

### **Performance Results (150 samples, 10s measurement)**
1. Simple Actions (1000 add actions)
Baseline: ~360-398 µs (2.51-2.77 Melem/s)
New version: ~311-312 µs (3.20-3.21 Melem/s)
Improvement: `~17-18%` faster

2. With Stats (1000 actions with stats)
Baseline: ~940 µs (1.06 Melem/s)
New version: ~688-717 µs (1.39-1.45 Melem/s)
Improvement: `~26-27%` faster

3. Full Complexity (1000 complex actions)
Baseline: ~1.41 ms (708-710 Kelem/s)
New version: ~1.22 ms (817-819 Kelem/s)
Improvement: `~13-15%` faster

## Changes
- Leverage the streaming `Deserializer::from_slice()`
- Avoids allocating intermediate String objects for each line
- Passing `&bytes::Bytes` avoids atomic ref-counting overhead

_Side notes: I will remove the `get_actions_baseline()`, but we could keep the bench test_

### DIFF
--- a/crates/catalog-unity/src/datafusion.rs
+++ b/crates/catalog-unity/src/datafusion.rs
@@ -183,14 +183,22 @@ impl UnitySchemaProvider {
         table: &str,
     ) -> Result<TemporaryTableCredentials, UnityCatalogError> {
         tracing::debug!("Fetching new credential for: {catalog}.{schema}.{table}",);
-        self.client
-            .get_temp_table_credentials(catalog, schema, table)
-            .map(|resp| match resp {
-                Ok(TableTempCredentialsResponse::Success(temp_creds)) => Ok(temp_creds),
-                Ok(TableTempCredentialsResponse::Error(err)) => Err(err.into()),
-                Err(err) => Err(err),
-            })
+        match self
+            .client
+            .get_temp_table_credentials_with_permission(catalog, schema, table, "READ_WRITE")
             .await
+        {
+            Ok(TableTempCredentialsResponse::Success(temp_creds)) => Ok(temp_creds),
+            Ok(TableTempCredentialsResponse::Error(_err)) => match self
+                .client
+                .get_temp_table_credentials(catalog, schema, table)
+                .await?
+            {
+                TableTempCredentialsResponse::Success(temp_creds) => Ok(temp_creds),
+                _ => Err(UnityCatalogError::TemporaryCredentialsFetchFailure),
+            },
+            Err(err) => Err(err),
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/core/src/logstore/mod.rs
```diff
@@ -566,23 +566,22 @@ pub fn to_uri(root: &Url, location: &Path) -> String {
 /// Reads a commit and gets list of actions
 pub async fn get_actions(
     version: i64,
-    commit_log_bytes: bytes::Bytes,
+    commit_log_bytes: &bytes::Bytes,
 ) -> Result<Vec<Action>, DeltaTableError> {
```
abhi-airspace-intelligence, line 570: I wonder why this function is async. There's nothing async inside of it. Not your fault as the base function was also async, but probably some legacy tech debt? I can imagine there was a world where this function took an async bytes stream instead of all the bytes.

abhi-airspace-intelligence: per git blame, this function was implemented two years ago as async even though there were no async in it at any time.

fvaleye: Yes, I kept the function `async`. 
Removing `async` would be a minor breaking change, as it would also require removing `.await` from the callers.
Let's see what @roeap and @rtyler think about this!

roeap: more generally speaking, I see most call sites disappearing short term, since log replay nor produces record batches that we extract data from (i.e. `LogFileView` et. al.) avoiding copies whenver possible.

Exception being calling this in `commit_infos`.

Sine we are passing in `Bytes`, I see no reason why we should be doing IO in this function, and with that also little reason for it to be async ... maybe in a follow-up we can make it sync.

roeap: Also, let's see when we have fully kernelized conflict resolution. There might be a few surprises lurking 😆.

fvaleye: > Also, let's see when we have fully kernelized conflict resolution. There might be a few surprises lurking 😆.

Yay!
Let's keep it like this for now and make it sync later.
I will create an issue for tracking this need.

PR #3831, opened 2025-10-09
feat: datafusion based kernel engine

roeap:
# Description

This is the first direct step toward kernelized scans. The core of this will be a new `TableProvider`. This new engine directly integrates with the datafusion session (more specifically, `TaskContext`). I thought about making a more generic engine that we could also use for non-datafusion code-paths. However, we would like to exchange the "inner" implementations to also leverage datafusion abstractions, at which point these would likely diverge. Thus, hoping to integrate this deeper first, before seeing if and how we can share between implementations.

Why do we need a dedicated engine? URL support and moving away from object store paths.

I have this integrated with a table provider locally, but I am trying to keep things reviewable.


crates/core/src/delta_datafusion/engine/file_formats.rs
```diff
@@ -0,0 +1,176 @@
...
+            RuntimeFlavor::MultiThread => Arc::new(DefaultJsonHandler::new(
+                store,
+                Arc::new(TokioMultiThreadExecutor::new(self.handle.clone())),
+            )),
+            RuntimeFlavor::CurrentThread => Arc::new(DefaultJsonHandler::new(
+                store,
+                Arc::new(TokioBackgroundExecutor::new()),
+            )),
+            _ => panic!("unsupported runtime flavor"),
+        };
```
roeap, line 97: We replicate this logic inside functions to avoid the generic parameter spilling out into our implementations. This tends to bubble up quite quickly since `TaskExecutor` is not `dyn`.

Once we move to datafusion specific implementations, these will not be generic over the runtime since datafusion and we are tied to tokio.

crates/core/src/delta_datafusion/engine/file_formats.rs
```diff
@@ -0,0 +1,176 @@
...
+                        physical_schema.clone(),
+                        predicate.clone(),
+                    )
+                })
+                // TODO: this should not do any blocking operations, since this should
+                // happen when the iterators are polled and we are just creating a vec of iterators.
+                // Is this correct?
+                .try_collect::<_, Vec<_>, _>()?
+                .into_iter()
+                .flatten(),
```
roeap, line 127: Would love some feedback on this comment if I made the right assumptions here. We can also handle this in the `map`, but this gets a bit messy...

crates/core/src/delta_datafusion/engine/storage.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+pub(crate) fn group_by_store<T: IntoIterator<Item = impl AsObjectStoreUrl>>(
+    files: T,
+) -> std::collections::HashMap<ObjectStoreUrl, Vec<T::Item>> {
+    files
+        .into_iter()
+        .map(|item| (item.as_object_store_url(), item))
+        .into_group_map()
+}
```
roeap, line 145: This is the core piece for routing things to individual sub-instances scoped to an individual store.

There is a caveat in there in that we could be changing the order of inputs, which, of course, hurts log replay. But will look closer into this. The good news is that logs reside in a single store; only data files are potentially distributed across stores.

PR #3834, opened 2025-10-09
feat: enable ability to do writes through Unity Catalog

hntd187:
# Description
Now if the user has permissions to do writes (and actually does a write) we will request a write permission first instead of just read only permissions. When this fails we will go back to the normal path of requesting a read-only cred. 

# Related Issue(s)
None I'm aware of.

# Documentation
https://docs.databricks.com/api/workspace/temporarytablecredentials/generatetemporarytablecredentials


ion-elgreco:
@hntd187 there is one test failing, but otherwise LGTM

PR #3836, opened 2025-10-11
chore: modified add_actions_table to stream

### LINKED ISSUE #3811: Move `add_actions_table` onto stream
# Description

delta logs can become fairly large so fully materialising them should be done only when really required.

Whenever we use `concat_batches` this is likely the case. an example being the `add_actions_table` method, which currently returns a single record batch.

We should change that method to return `BoxStream<...>` instead and use the same lazy readers we use for scans to expose it in python.

zen-zap:
# Description
Changed `add_actions_table` to return BoxStream in `crates/core/src/table/state.rs`. This helps by making sure we don't load everything into memory all at once. Added `BoxStreamToReaderAdapter` in `python/src/reader.rs`  to convert the BoxStreams into RecordBatchReader. Modified the python bindings and tests to use `.read_all()` to fetch them all at once.

# Related Issue(s)
Some tests fail I think. I got this: 
438 passed, 4 skipped, 2 xfailed, 5 xpassed, 22 warnings in 14.72s

- closes #3811

### DIFF
--- a/crates/core/src/table/state.rs
+++ b/crates/core/src/table/state.rs
@@ -12,7 +12,7 @@ use futures::{future::ready, stream, stream::BoxStream, StreamExt as _, TryStrea
 use object_store::path::Path;
 use serde::{Deserialize, Serialize};
 
-use super::DeltaTableConfig;
+use super::{builder, DeltaTableConfig};
 use crate::kernel::arrow::engine_ext::{ExpressionEvaluatorExt, SnapshotExt};
 #[cfg(test)]
 use crate::kernel::Action;
@@ -20,10 +20,10 @@ use crate::kernel::{
     Add, DataType, EagerSnapshot, LogDataHandler, LogicalFileView, Metadata, Protocol,
     TombstoneView, ARROW_HANDLER,
 };
-use crate::logstore::LogStore;
+use crate::logstore::{LogStore, LogStoreRef};
 use crate::partitions::PartitionFilter;
 use crate::table::config::TablePropertiesExt;
-use crate::{DeltaResult, DeltaTableError};
+use crate::{open_table, DeltaResult, DeltaTableError};
 
 /// State snapshot currently held by the Delta Table instance.
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


python/src/lib.rs
```diff
@@ -1434,18 +1434,39 @@ impl RawDeltaTable {
...
-                .map_err(PythonError::from)?)
+        
+        let batches = self.with_table(|t| {
+            let snapshot = t.snapshot()
+                .map_err(PythonError::from)
+                .map_err(PyErr::from)?;
+            let stream = snapshot.add_actions_table(flatten);
+            
+            // Collect batches from stream
+            let batches: Vec<RecordBatch> = rt()
```
ion-elgreco: This action already materializes all batches into memory

zen-zap: Sorry, I missed that. 

I did make some changes to give BoxStream a 'static lifetime parameter and tried to have it stream instead of collecting it all.

crates/core/src/table/state.rs
```diff
@@ -343,34 +349,45 @@ impl EagerSnapshot {
...
+        let evaluator =
+            ARROW_HANDLER.new_expression_evaluator(input_schema, expression.into(), table_schema);
 
-        if flatten {
-            Ok(result.normalize(".", None)?)
-        } else {
-            Ok(result)
-        }
+        let files = self.files.clone();
+        stream::iter(files)
```
abhiaagarwal: I don't think this PR is safe.

Sure, the `'static` lifetime helps avoid the lifetime issues, but at the cost of essentially cloning the table's file state at "a point in time". This means that if you get the record batch stream, then perform operations on the delta table (which may delete a file, say a compact) and _then_ you try to consume the stream, you could be reading a file that got invalidated.

Someone should double-check my logic, but a `'static` lifetime is a bit scary unless you can prove it really _does_ live for the duration of the program.

abhiaagarwal: I guess that these are log files, it's pretty rare they get deleted, but it still would get stale table state if another action added some files. Maybe that's the right abstraction on the python side, but definitely not the rust side.

zen-zap: Maybe I could take a snapshot like - get the current version of the delta table and start the stream from that? If I tie it to the version, it should be okay I think.. 

Do you have any suggestions?

abhiaagarwal: The only real way to do it in a safe way would be to have the `BoxStream` lifetime be `'a` which is tied into `&'a self`, so the boxstream borrows from self. 

```
pub fn add_actions_table<'a>(
    &'a self,
    flatten: bool,
) -> BoxStream<'a, DeltaResult<arrow::record_batch::RecordBatch>> {
    ...
}
```

abhiaagarwal: Now that I think about it, in theory, the `self.files` parameter _should_ never change per snapshot so perhaps this is safe, but it's not enforced in the type system. On the python side though, this race condition absolutely does exist, though I'm not sure what the proper semantics are anyways

zen-zap: I think version pinning would help dealing with the race condition on rust side. The python side uses this so that should get fixed. Let me know if you have any ideas about this. Thanks!

Issue #3843, opened 2025-10-13
Port benchmarks from spark delta to delta-rs


abhiaagarwal:
# Description

**Use Case**

The upstream spark delta lake has a suite of benchmarks [here](https://github.com/delta-io/delta/tree/master/benchmarks). The original benchmark setup in #1857 was a port of the merge benchmarks. We should port the remaining benchmarks.

**Related Issue(s)**

Follow up to #3839, related to #3640

abhiaagarwal:
take

rtyler:
I believe this was done with #3845

PR #3845, opened 2025-10-14
feat(bench): add TPC-DS benchmarks

abhiaagarwal:
# Description

Adds benchmarks based on the [TPS-DS queries](https://github.com/delta-io/delta/blob/master/benchmarks/src/main/scala/benchmark/TPCDSBenchmark.scala) used in the delta-spark repository. This is definitely closer to a "datafusion benchmark" over a "delta benchmark", but it provides a decent baseline for optimization opportunities.

# Related Issue(s)

Progress towards #3843


roeap (dismissed):
Awesome @abhiaagarwal!

Read through the code and looks great. Really stepping up the benchmarking game here!

There are a few minor CI failures to fix, but we can merge otherwise.

One question that came up, especially considering @alexwilcoxson-rel  latest PR. Is there a way we can track additional metrics we collect during query execution as benchmarking results? I.e. teacking how well we do file skipping / reading so files skipped or bytes read ...

rtyler (dismissed):

abhiaagarwal:
> Awesome @abhiaagarwal!
> 
> Read through the code and looks great. Really stepping up the benchmarking game here!
> 
> There are a few minor CI failures to fix, but we can merge otherwise.
> 
> One question that came up, especially considering @alexwilcoxson-rel latest PR. Is there a way we can track additional metrics we collect during query execution as benchmarking results? I.e. teacking how well we do file skipping / reading so files skipped or bytes read ...

I looked into this, and I believe not :( I believe this issue in divan would be what we're looking for https://github.com/nvzqz/divan/issues/34

PR #3847, opened 2025-10-14
feat(datafusion): add max_temp_directory_size parameter for z-order and compact operations for DataFusion

### LINKED ISSUE #3833: Asked to increase `max_temp_directory_size` in the disk manager configuration when optimizing large table
# Environment

**Delta-rs version**: 1.1.4

# Bug

**What happened**:

When I tried to optimize a large delta table (10G of parquet), I get the following error on my M4 Mac:

_internal.DeltaError: Failed to parse parquet: Parquet error: Z-order failed while scanning data: ResourcesExhausted("The used disk space during the spilling process has exceeded the allowable limit of 100.0 GB. Try increasing the `max_temp_directory_size` in the disk manager configuration.")

Spelunking I'm pretty sure this is a datafusion limit. Is there any way to increase it? I tried to tweak all the possible settings to no avail:

```py
#!/usr/bin/env python

import sys
from deltalake import WriterProperties

SPILL_SIZE = 40 * 1024 * 1024 * 1024  # 40 gb

agg = DeltaTable(sys.argv[1], storage_options={"AWS_DEFAULT_REGION": "us-west-2"})
wp = WriterProperties(compression="ZSTD", compression_level=3)
print("Optimizing...")
# agg.optimize.compact()
# agg.vacuum(retention_hours=0, enforce_retention_duration=False, dry_run=False)

agg.optimize.z_order(
    columns=["object_id"],
    writer_properties=wp,
    max_concurrent_tasks=1,
    max_spill_size=SPILL_SIZE,
)
```

Allowed to either increase the limit or have less spillage

**How to reproduce it**:

Not sure, dataset is proprietary. Probably could create a reproduction case with safe data if it helps though.

**More details**:

fvaleye:
# Description
- Add `max_temp_directory_size` parameter to the Python bindings to optimize DataFusion operations.
- Add `max_spill_size` to the Python binding (available in Rust)

# Related Issue(s)
- closes #3833


crates/core/src/operations/optimize.rs
```diff
@@ -247,6 +250,7 @@ impl<'a> OptimizeBuilder<'a> {
             preserve_insertion_order: false,
             max_concurrent_tasks: num_cpus::get(),
             max_spill_size: 20 * 1024 * 1024 * 1024, // 20 GB.
+            max_temp_directory_size: 100 * 1024 * 1024 * 1024, // 100 GB.
```
ion-elgreco: Maybe not expose it here if we immediately deprecate it, but rather create a proper configured Session in the python binding when calling optimize and passing that around

fvaleye: I took the simple approach.
I can take more time to create a properly configured session, something like this:
```
fn create_optimize_session(
    max_spill_size: usize,
    max_temp_directory_size: u64,
) -> SessionState {
```
WDYT?

ion-elgreco: Looks good!

fvaleye: Okey dokey!
I will refactor this to remove depreciation and configure a proper session. I will ping you when it's ready.

abhiaagarwal:
I looked into this similarly and unfortunately, there doesn't seem to be a way to allow passing in a datafusion SessionState from the python side to the rust side, which would give python callers the maximum flexibility :/

ion-elgreco:
> I looked into this similarly and unfortunately, there doesn't seem to be a way to allow passing in a datafusion SessionState from the python side to the rust side, which would give python callers the maximum flexibility :/

Not a state, but you can consume FFISessionConfig, not sure if it has all the some buttons and knobs as state/context

abhiaagarwal:
> > I looked into this similarly and unfortunately, there doesn't seem to be a way to allow passing in a datafusion SessionState from the python side to the rust side, which would give python callers the maximum flexibility :/
> 
> 
> 
> Not a state, but you can consume FFISessionConfig, not sure if it has all the some buttons and knobs as state/context

It accomplishes the goal on the per-operation level, but if you wanted to share a SessionContext across multiple operations (say, to limit the total amount of ram for working with multiple tables concurrently) you're out of luck. 

I guess the rust side could store a SessionContext as an optional in RawDeltaTable and pass that in every operation to achieve the same goal, though. Definitely not done in this ticket, though :D

fvaleye:
> > I looked into this similarly and unfortunately, there doesn't seem to be a way to allow passing in a datafusion SessionState from the python side to the rust side, which would give python callers the maximum flexibility :/
> 
> Not a state, but you can consume FFISessionConfig, not sure if it has all the some buttons and knobs as state/context

Something like:
```
config = SessionConfig()
config.set_config("datafusion.execution.target_partitions", "8")
config.set_config("datafusion.optimizer.skip_failed_rules", "false")

dt.optimize.z_order(
    columns=["id"],
    session_config=config,  # Advanced DataFusion config
)
```
We can have a follow-up PR for this, WDYT?

ion-elgreco:
> > > I looked into this similarly and unfortunately, there doesn't seem to be a way to allow passing in a datafusion SessionState from the python side to the rust side, which would give python callers the maximum flexibility :/
> > 
> > 
> > Not a state, but you can consume FFISessionConfig, not sure if it has all the some buttons and knobs as state/context
> 
> Something like:
> 
> ```
> config = SessionConfig()
> config.set_config("datafusion.execution.target_partitions", "8")
> config.set_config("datafusion.optimizer.skip_failed_rules", "false")
> 
> dt.optimize.z_order(
>     columns=["id"],
>     session_config=config,  # Advanced DataFusion config
> )
> ```
> 
> We can have a follow-up PR for this, WDYT?

Yeah might be nice for interopability with datafusion-python

roeap:
@fvaleye @ion-elgreco @abhiaagarwal - I just opened #3860 to create a more narrow waist when creating datafusion sessions. Going forward we also need to wire our object-store and log store factories with the datafusion session.

Would it be possible to base this PR on the one above to properly configure the session?

fvaleye:
> @fvaleye @ion-elgreco @abhiaagarwal - I just opened #3860 to create a more narrow waist when creating datafusion sessions. Going forward we also need to wire our object-store and log store factories with the datafusion session.
> 
> Would it be possible to base this PR on the one above to properly configure the session?

Yes, I will do that @roeap 👍

ion-elgreco (dismissed):

ion-elgreco:
@fvaleye can you resolve the conflict, then we are good to go :)

fvaleye:
> @fvaleye can you resolve the conflict, then we are good to go :)

I'm on it!

ion-elgreco (approved):

PR #3848, opened 2025-10-14
feat: add num_retries metric to write operation metrics

vsmanish1772:
# Description

Track the number of commit retry attempts in operationMetrics for write operations.

## Changes
I have added num_retries to WriteMetrics struct which will be available inside operationMetrics inside delta log. I have used #[serde(default)] for backward compatibility with old delta logs. 

The reason for the current approach was because operationMetrics is written to the delta log before we know how many retries were needed, making it difficult to include in the persisted log. Hence I do reserialise the commit log again. 
## Result

Delta log now includes num_retries in operationMetrics: {
  "operationMetrics": {
    "num_added_files": 5,
    "num_retries": 2,
    ...
  }
}

# Related Issue(s)

# Documentation


vsmanish1772:
I'm not really happy with the approach or is it a right approach but I could think of any other solution as it was tricky. let me know if we can do something better. 

Also I would like to add this change for all different operations i.e delete, merge, update etc. To keep the pr small I have started with write operation and will bring in subsequent pr's once we agree with approach here
@rtyler @ion-elgreco

ion-elgreco:
Hey @vsmanish1772, as you are saying as well, I don't believe this should be the approach. We need to somehow defer the bytes creation at a later stage but I also wonder this will give overhead to re-update the commit bytes each time you retry, so maybe this should be an opt-in functionality only (@roeap @rtyler, thoughts?)

rtyler (approved):
@ion-elgreco With the optimization I just added, I think this is safe to merge and an interesting bit of information to log. The recalculating of the bytes is going to pale in comparison to the actual I/O call in terms of the wall-clock time consumed here.

I think it's worth proceeding

PR #3850, opened 2025-10-14
fix: enable default features on `WriteBuilder` datafusion context

abhi-airspace-intelligence:
# Description

Ran into this edge case where if writing to a table that has a generated column defined by a SQL function (say, as a partition column, `to_char(timestamp)`, the write builder will fail since a default SessionState does not have UDFs unless `new_with_default_features` is specified.

# Related Issue(s)

Regression from #3816 

# Documentation

Left as draft since I want to write a regression test to assert this doesn't happen again


rtyler (approved):

roeap:
@abhi-airspace-intelligence - This might be also addressed in #3860 where we centralise creating the sessions a bit more.

PR #3851, opened 2025-10-15
chore: modified add_actions_table to use streams

### LINKED ISSUE #3811: Move `add_actions_table` onto stream

zen-zap:
# Description
changed `add_actions_table` to return BoxStream. This helps by making sure we don't load everything into memory all at once. Added `BoxStreamToReaderAdapter` to convert BoxStreams to RecordBatchReader. Streaming with a specific version helps avoid race conditions but we depend on the retention period of the versions.

# Related Issue(s)
None that I'm aware of.

- closes #3811


rtyler:
converting to a draft due to the conflicts

zen-zap:
Should I try to resolve them and open it again?

hntd187:
This PR is almost a year old with no movement in a long time. If you wish to continue this work please open a new PR.

Issue #3855, opened 2025-10-15
Compacting a delta table consistently undershoots the `target_file_size`, creating an unnecessary extra file


abhi-airspace-intelligence:
# Environment

**Delta-rs version**: git latest

**Binding**: Rust

**Environment**:
- **Cloud provider**: AWS
- **OS**: Linux
- **Other**: Writing Parquet 2.0 files with ZSTD

***
# Bug

**What happened**:

When compacting a table, delta-rs seems to consistently undershoot the target file size, leading to it creating a file size of ~98MB and another file of roughly 1MB. 

**What you expected to happen**:

It should write one file with 100MB, rather than splitting into a 98MB file and a ~1.8MB file.

**How to reproduce it**:

Working on a repro right now. 

**More details**:

My guess is that this occurs only with Parquet V2 since usage of it in the wild is rare, and the naming indicates this codepath is getting hit https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/write/writer.rs#L478-L486

<img width="633" height="464" alt="Image" src="https://github.com/user-attachments/assets/9ec26d46-a52d-4a20-931d-8056f0188017" />

NSchrading:
I'm also seeing this issue. I think a fix here would be a huge improvement for systems that are compacting pretty large tables. I think it might also resolve https://github.com/delta-io/delta-rs/issues/2591. I think due to this bug, the compaction logic attempts to compact the same files over and over because it's leaving small leftover files every time. Without partition filters, this means the compaction logic will need to rewrite more and more data each time, eventually using all of your available disk space and failing.

I see there is a potential fix on https://github.com/delta-io/delta-rs/pull/3901. I would love if that could get prioritized and in for the next release.

PR #3857, opened 2025-10-16
feat: in-flight, streaming `PartitionWriter`

### LINKED ISSUE #3578: Refactor delta writer to not use buffer

abhiaagarwal:
# Description

This was based on me taking a closer look at the `PartitionWriter` code, and realizing it could be made a bit more efficient by deferring the parquet writing to a background task and removing the async locking associated with `AsyncShareableBuffer`. 

Instead, it switches the underlying writer to use a `ParquetObjectWriter`, which internally uses a `BufWriter`, which keeps a certain capacity of bytes in-memory until reaching a threshold, where it starts a multipart write. Additionally, once a writer has enough bytes in-progress/about to be flushed, we finish its writing on a background tokio task instead of yielding while putting new writes on the new file.

# Related Issue(s)

Probably closes #3578, as it still uses a small buffer internally through `ParquetObjectWriter` but adaptively flushes via a multipart upload using `BufWriter` for streaming writes. I'm also hoping this solves the issue I faced in #3855, though I'll aim to test this in production.

- closes https://github.com/delta-io/delta-rs/pull/3675 (superseded)

# Documentation

[`ParquetObjectWriter`](https://docs.rs/parquet/56.2.0/parquet/arrow/async_writer/struct.ParquetObjectWriter.html)
[`BufWriter`](https://docs.rs/object_store/0.12.4/object_store/buffered/struct.BufWriter.html)

### DIFF
--- a/crates/core/src/operations/write/writer.rs
+++ b/crates/core/src/operations/write/writer.rs
@@ -35,6 +35,7 @@ use parquet::format::FileMetaData;
 const DEFAULT_TARGET_FILE_SIZE: usize = 104_857_600;
 const DEFAULT_WRITE_BATCH_SIZE: usize = 1024;
 const DEFAULT_UPLOAD_PART_SIZE: usize = 1024 * 1024 * 5;
+const DEFAULT_MAX_CONCURRENCY_TASKS: usize = 10;
 
 fn upload_part_size() -> usize {
     static UPLOAD_SIZE: OnceLock<usize> = OnceLock::new();
@@ -56,6 +57,16 @@ fn upload_part_size() -> usize {
                 }
             })
             .unwrap_or(DEFAULT_UPLOAD_PART_SIZE)
+    })
+}
+
+fn get_max_concurrency_tasks() -> usize {
+    static MAX_CONCURRENCY_TASKS: OnceLock<usize> = OnceLock::new();
+    *MAX_CONCURRENCY_TASKS.get_or_init(|| {
+        std::env::var("DELTARS_MAX_CONCURRENCY_TASKS")
+            .ok()
+            .and_then(|s| s.parse::<usize>().ok())
+            .unwrap_or(DEFAULT_MAX_CONCURRENCY_TASKS)
     })
 }
 
@@ -296,7 +307,7 @@ pub struct PartitionWriterConfig {
     /// determine how fine granular we can track / control the size of resulting files.
     write_batch_size: usize,
     /// Concurrency level for writing to object store
-    max_concurrency: usize,
+    max_concurrency_tasks: usize,
 }
 
 impl PartitionWriterConfig {
@@ -307,7 +318,7 @@ impl PartitionWriterConfig {
         writer_properties: Option<WriterProperties>,
         target_file_size: Option<usize>,
         write_batch_size: Option<usize>,
-        max_concurrency: Option<usize>,
+        max_concurrency_tasks: Option<usize>,
     ) -> DeltaResult<Self> {
         let part_path = partition_values.hive_partition_path();
         let prefix = Path::parse(part_path)?;
@@ -326,7 +337,7 @@ impl PartitionWriterConfig {
             writer_properties,
             target_file_size,
             write_batch_size,
-            max_concurrency: max_concurrency.unwrap_or(10),
+            max_concurrency_tasks: max_concurrency_tasks.unwrap_or_else(get_max_concurrency_tasks),
         })
     }
 }
@@ -346,7 +357,7 @@ impl LazyArrowWriter {
                         path.clone(),
                         upload_part_size(),
                     )
-                    .with_max_concurrency(config.max_concurrency),
+                    .with_max_concurrency(config.max_concurrency_tasks),
                 );
                 let mut arrow_writer = AsyncArrowWriter::try_new(
                     writer,


crates/core/src/operations/write/writer.rs
```diff
@@ -57,6 +58,52 @@ fn upload_part_size() -> usize {
...
+    arrow_writer: ArrowWriter<Vec<u8>>,
+    path: Path,
+    object_store: ObjectStoreRef,
+) -> DeltaResult<(Path, i64, FileMetaData)> {
+    // Finalize the arrow writer and get the buffer
+    let mut arrow_writer = arrow_writer.into_serialized_writer()?.0;
+    let metadata = arrow_writer.finish()?;
+    // SAFETY: The buffer was already constructed, but we can't consume with `into_inner` since
+    // `finish` finalizes the writer.
+    let buffer = unsafe { &*(arrow_writer.inner_mut().as_slice() as *const [u8]) };
```
rtyler: :scream_cat:

crates/core/src/operations/write/writer.rs
```diff
@@ -57,6 +58,52 @@ fn upload_part_size() -> usize {
...
+        let part = buffer_bytes.slice(offset..end);
+        let upload_future = multi_part_upload.put_part(PutPayload::from(part));
+
+        // wait until one spot frees up before spawning new task
+        if tasks.len() >= max_concurrent_tasks {
+            tasks.join_next().await;
+        }
+        tasks.spawn(upload_future);
+        offset = end;
+    }
```
rtyler: I have talked about similar approaches with @ion-elgreco in the past. My belief based on what I have seen is that parallelizing I/O operations in this way doesn't buy too much since the upload of one part/file _typically_ is going to go at line speed, splitting into parts will be `line speed / num parts`

My hunch is that if we're able to parallelize the parquet encoding then we'd see some good improvements.

Do you have some measurements to share to help inform the discussion?

crates/core/src/operations/write/writer.rs
abhi-airspace-intelligence: (work account)

yeah, unfortunately there's a weird problem where calling `into_inner` calls `finish` but it errors if `finish` was already called, but I need the `FileMetadata`, which I can only get from manually calling `finish`. So this is safe but working around a limited API.

https://docs.rs/parquet/latest/src/parquet/arrow/arrow_writer/mod.rs.html#378-404

crates/core/src/operations/write/writer.rs
abhi-airspace-intelligence: So this code was copied from the original implementation, I don't have any opinion on whether this works or not. I know `ParquetObjectWriter` does something similar internally.

The more benefit of this PR is that rather than yielding on writing the new parquet file to the store before starting to write the second one, we can kick off the write job for the first file in the background and immediately start writing to the second one.

As for benchmarks, I'm writing a python-based one. But I might just give this PR a whirl in my dev environment anyways for my own curiosity with AWS :)

abhiaagarwal:
```
------------------------------------------------------------------------------------------------- benchmark 'optimize': 8 tests --------------------------------------------------------------------------------------------------
Name (time in ms)                                          Min                   Max                  Mean              StdDev                Median                 IQR            Outliers     OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_optimize[5] (0001_writer-)             697.8591 (1.0)        783.6447 (1.0)        730.9472 (1.0)       33.9840 (1.39)       715.5890 (1.0)       45.2321 (2.69)          1;0  1.3681 (1.0)           5           1
test_benchmark_optimize[5] (0002_main)                733.2482 (1.05)       826.2599 (1.05)       766.5344 (1.05)      41.3734 (1.69)       739.8363 (1.03)      63.7684 (3.79)          1;0  1.3046 (0.95)          5           1
test_benchmark_optimize[1] (0001_writer-)           2,296.7281 (3.29)     2,818.7333 (3.60)     2,646.8827 (3.62)     202.7652 (8.27)     2,713.1137 (3.79)     173.5503 (10.31)         1;1  0.3778 (0.28)          5           1
test_benchmark_optimize[1] (0002_main)              2,379.6139 (3.41)     2,511.3386 (3.20)     2,425.0909 (3.32)      50.5637 (2.06)     2,412.8268 (3.37)      46.4805 (2.76)          1;0  0.4124 (0.30)          5           1
test_benchmark_optimize_minio[5] (0001_writer-)     2,901.4343 (4.16)     3,096.3666 (3.95)     2,996.6734 (4.10)      89.5425 (3.65)     2,983.5820 (4.17)     168.9837 (10.04)         2;0  0.3337 (0.24)          5           1
test_benchmark_optimize_minio[5] (0002_main)        2,945.3638 (4.22)     3,008.1905 (3.84)     2,965.4372 (4.06)      24.5123 (1.0)      2,958.0412 (4.13)      16.8391 (1.0)           1;1  0.3372 (0.25)          5           1
test_benchmark_optimize_minio[1] (0002_main)        5,607.0593 (8.03)     5,673.8537 (7.24)     5,631.1967 (7.70)      25.2358 (1.03)     5,625.3738 (7.86)      22.5687 (1.34)          1;0  0.1776 (0.13)          5           1
test_benchmark_optimize_minio[1] (0001_writer-)     5,739.4105 (8.22)     5,972.5491 (7.62)     5,805.5987 (7.94)      96.2346 (3.93)     5,780.0473 (8.08)      96.6722 (5.74)          1;0  0.1722 (0.13)          5           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------------------------------------------- benchmark 'write': 4 tests ----------------------------------------------------------------------------------------------
Name (time in ms)                                  Min                   Max                  Mean             StdDev                Median                 IQR            Outliers     OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_write (0001_writer-)           266.1552 (1.0)        347.1750 (1.03)       289.3167 (1.0)      34.6414 (3.54)       269.6400 (1.0)       41.8909 (2.53)          1;0  3.4564 (1.0)           5           1
test_benchmark_write (0002_main)              315.7212 (1.19)       337.5128 (1.0)        323.9466 (1.12)      9.7868 (1.0)        319.3020 (1.18)      16.5525 (1.0)           1;0  3.0869 (0.89)          5           1
test_benchmark_write_minio (0001_writer-)     914.5869 (3.44)       999.5264 (2.96)       948.1451 (3.28)     33.1406 (3.39)       944.4835 (3.50)      45.2663 (2.73)          2;0  1.0547 (0.31)          5           1
test_benchmark_write_minio (0002_main)        973.5743 (3.66)     1,105.6136 (3.28)     1,035.1443 (3.58)     60.9761 (6.23)     1,026.6726 (3.81)     115.4668 (6.98)          2;0  0.9660 (0.28)          5           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

about a ~5-10% increase in write performance

crates/core/src/operations/write/writer.rs
abhiaagarwal: This was removed entirely, as I decided to use `ParquetObjectWriter` instead.

crates/core/src/operations/write/writer.rs
```diff
@@ -305,10 +326,54 @@ impl PartitionWriterConfig {
             writer_properties,
             target_file_size,
             write_batch_size,
+            max_concurrency: max_concurrency.unwrap_or(10),
```
ion-elgreco: Can we make this configurable with an EnvVar? :)

crates/core/src/operations/write/writer.rs
```diff
@@ -305,10 +326,54 @@ impl PartitionWriterConfig {
...
+enum LazyArrowWriter {
+    Initialized(Path, ObjectStoreRef, PartitionWriterConfig),
+    Writing(Path, AsyncArrowWriter<ParquetObjectWriter>),
+}
+
+impl LazyArrowWriter {
+    async fn write_batch(&mut self, batch: &RecordBatch) -> DeltaResult<()> {
+        match self {
+            LazyArrowWriter::Initialized(path, object_store, config) => {
+                let writer = ParquetObjectWriter::from_buf_writer(
```
ion-elgreco, line 354: Nice, didn't even know Parquet crate had this streamed write functionality

ion-elgreco:
@abhiaagarwal we might still suffer from smaller files, we create a Writer for each datafusion partition output and then call close at the end of the execution of each inner plan. Depending on how datafusion partitioned this, it might still be below the file-size even though combined it's above the filesize

abhiaagarwal:
> @abhiaagarwal we might still suffer from smaller files, we create a Writer for each datafusion partition output and then call close at the end of the execution of each inner plan. Depending on how datafusion partitioned this, it might still be below the file-size even though combined it's above the filesize

Yeah, my suspicion is that when the datafusion plan is created it uses `target_file_size` and then when it executes it also uses `target_file_size`, where the best thing would be to use it only for the planning and not the writing on the compact/zorder plan.

crates/core/src/operations/write/writer.rs
abhiaagarwal: Sure! `DELTA_RS_MAX_CONCURRENCY_TASKS` sound good, to mirror the upload_part function?

crates/core/src/operations/write/writer.rs
abhiaagarwal: Me neither, it's pretty new and I only found it after starting this PR

ion-elgreco:
> > @abhiaagarwal we might still suffer from smaller files, we create a Writer for each datafusion partition output and then call close at the end of the execution of each inner plan. Depending on how datafusion partitioned this, it might still be below the file-size even though combined it's above the filesize
> 
> Yeah, my suspicion is that when the datafusion plan is created it uses `target_file_size` and then when it executes it also uses `target_file_size`, where the best thing would be to use it only for the planning and not the writing on the compact/zorder plan.

I think we can solve it if we have a single global writer here:

https://github.com/delta-io/delta-rs/blob/e6d9a2acbe226b224926f504631c6ef493b7e0c7/crates/core/src/operations/write/execution.rs#L283-L308

Instead a new writer for each inner plan

crates/core/src/operations/write/writer.rs
abhiaagarwal: 464e2ba6a6b7db81e1103878da087db4428f16cd

PR #3872, opened 2025-10-19
feat: allow for lazy loading files in operations

roeap:
# Description

Thus far we had a bunch of `unwraps` in our `DeltaOps` that would panic for a missing `EagerSnapshot`. In this PR we handle missing snapshot by resolving the options inside an operations future. This also made it quite simple to re-load files in case we need them but `require_files=false`. This effectively allows lazy laoding the files while we still default to eagerly loading the data.

The PR does look quite large, but the patterns are very repetitive.
* make eager snapshot optional on builders
* resolve snapshot at beginning of operation future

To safeguard against trying to execute operations without files loaded, we pushed the respective check on the delta config into the methods accessing the files.

### DIFF
--- a/crates/core/src/operations/add_column.rs
+++ b/crates/core/src/operations/add_column.rs
@@ -29,7 +29,7 @@ pub struct AddColumnBuilder {
     custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
 }
 
-impl Operation<()> for AddColumnBuilder {
+impl Operation for AddColumnBuilder {
     fn log_store(&self) -> &LogStoreRef {
         &self.log_store
     }
--- a/crates/core/src/operations/add_feature.rs
+++ b/crates/core/src/operations/add_feature.rs
@@ -29,7 +29,7 @@ pub struct AddTableFeatureBuilder {
     custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
 }
 
-impl super::Operation<()> for AddTableFeatureBuilder {
+impl super::Operation for AddTableFeatureBuilder {
     fn log_store(&self) -> &LogStoreRef {
         &self.log_store
     }
--- a/crates/core/src/operations/constraints.rs
+++ b/crates/core/src/operations/constraints.rs
@@ -41,7 +41,7 @@ pub struct ConstraintBuilder {
     custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
 }
 
-impl super::Operation<()> for ConstraintBuilder {
+impl super::Operation for ConstraintBuilder {
     fn log_store(&self) -> &LogStoreRef {
         &self.log_store
     }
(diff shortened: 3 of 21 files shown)


crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -463,7 +463,23 @@ impl Snapshot {
...
+    if let Some(snapshot) = maybe_snapshot {
+        if require_files {
+            snapshot.with_files(log_store).await
+        } else {
+            Ok(snapshot)
+        }
+    } else {
+        EagerSnapshot::try_new(log_store, Default::default(), None).await
+    }
 }
```
roeap, line 485: This function along with the new methods on `EagerSnapshot` are the main pieces of new logic.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -463,7 +463,25 @@ impl Snapshot {
...
+        if require_files {
+            snapshot.with_files(log_store).await
+        } else {
+            Ok(snapshot)
+        }
+    } else {
+        let mut config = DeltaTableConfig::default();
+        config.require_files = require_files;
+        EagerSnapshot::try_new(log_store, config, None).await
+    }
```
roeap, line 484: This function along with the new methods on EagerSnapshot are the main pieces of new logic.

doing this again b/c force push.

rtyler (commented):
All of these builders are having a constructor change, which I don't think is necessarily a problem but _should_ they all have `pub` constructors? This change forces an semver change due to breaking public APIs. :cry: 

If we're breaking, why not just take them away?

roeap:
@rtyler - I really like the idea of hiding the constructors. rigth now we need them in the python bindings.

Happy to look into this in a follow-up?

abhiaagarwal:
This is DEFINITELY something that shouldn't be part of this PR but I was looking through this code the other day and was perhaps wondering if it's possible to get the best of both worlds? Store a `BoxStream` internally, and as callers request more operation data, advance the stream and cache prior results in a Vec per an EagerSnapshot. So you load what you ask for. It becomes a bit murky with the lifetimes but definitely is possible, you'd probably want to Arc<EagerSnapshot> so all users share the underlying cache.

roeap:
@rtyler - turns out there surprisingly is a way ... using our intended APIs 😆. So most of the builder constructors are private now, exceptions being `WriteBuilder` and `MergeBuilder` which require some more refactoring.

@abhiaagarwal - we absolutely need to look into some caching strategies and the way we are moving is to do as much work on streams as possible. The main blocker in this is that most operations do their planning assuming the full state is available. So even if they consume the stream based apis on Snapshots for operations planning, they usually just collect the stream. My best guess right now is that we do some caching in the engine/object-store level (e.g. caching json bytes or deletion vector data) and treat our current `files` state as a cache as well. We do need to re-think some things , e.g.  how we provide statistics to datafusion w/o a-priori log replay, but there is precedence in datafusions Parquet handling.

PR #3879, opened 2025-10-20
feat: add multiple constraints at once

### LINKED ISSUE #1986: Add multiple constraints at once
# Description
Currently add constraints is limited to one named expression. But we can add a hash map as well in one go.

JustinRush80:
# Description
This PR allow users to add multiple constraints at once using an HashMap.

# Related Issue(s)

- closes #1986 

# Documentation

### DIFF
--- a/python/tests/test_alter.py
+++ b/python/tests/test_alter.py
@@ -53,16 +53,6 @@ def test_add_constraint(tmp_path: pathlib.Path, sample_table: Table):
         write_deltalake(tmp_path, data, mode="append")
 
 
-def test_add_multiple_constraints(tmp_path: pathlib.Path, sample_table: Table):
-    write_deltalake(tmp_path, sample_table)
-
-    dt = DeltaTable(tmp_path)
-
-    with pytest.raises(ValueError):
-        dt.alter.add_constraint(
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/constraints.rs
```diff
@@ -67,11 +65,12 @@ impl ConstraintBuilder {
     /// Specify the constraint to be added
     pub fn with_constraint<S: Into<String>, E: Into<Expression>>(
```
ion-elgreco, line 66: Maybe one small nit, i think we can keep this where this just inserts in the hashmap and then we also have `with_constraints` which is the hashmap version. This way we don't entirely break the signature

ion-elgreco (changes requested):
Good stuff! Just one comment for small improvement :)

crates/core/src/operations/constraints.rs
JustinRush80: Ah yes that makes sense! Do we want to restrict using both with_constraint and with_constraints at the same time? At first I had with_multiple_constraints that extend the hashmap just in case someone used both builder functions

ion-elgreco: Yeah you could have them both extend it, always

PR #3880, opened 2025-10-21
fix: remove manylinux 217 builds for aarch64

ion-elgreco:
Can't be supported anymore due to compilation issues, see https://github.com/delta-io/delta-rs/actions/runs/18479050964/job/53251499488

### DIFF
--- a/.github/workflows/python_release.yml
+++ b/.github/workflows/python_release.yml
@@ -88,30 +88,6 @@ jobs:
           command: publish
           args: --skip-existing -m python/Cargo.toml
 
-  release-pypi-manylinux-217-aarch64:
-    needs: validate-release-tag
-    name: PyPI release manylinux-2_17 aarch64
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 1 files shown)


roeap:
IIRC, maturin can generate a pretty nice ci setup via `maturin generate-ci github`. Maybe worthwhile to just run that and see if we like it better that want we have?

fvaleye:
> IIRC, maturin can generate a pretty nice ci setup via `maturin generate-ci github`. Maybe worthwhile to just run that and see if we like it better that want we have?

There is also this GitHub Action available: https://github.com/PyO3/maturin-action with built-in support for cross compilation

ion-elgreco:
> > IIRC, maturin can generate a pretty nice ci setup via `maturin generate-ci github`. Maybe worthwhile to just run that and see if we like it better that want we have?
> 
> There is also this GitHub Action available: https://github.com/PyO3/maturin-action with built-in support for cross compilation

This is already used in the CI

PR #3883, opened 2025-10-21
fix: display kernel-rs errors better

ion-elgreco:
# Description
Objectstore errors surfacing from kernel are not displayed nicely in python :)


roeap (approved):
👍

PR #3889, opened 2025-10-22
feat: generate an Symlink Manifest for External Engines

### LINKED ISSUE #2375: deltaTable.generate("symlink_format_manifest")
# Description
Looks like generation of manifest files is not yet available in Python deltalake

**Use Case**
To generate manifests for tools like BigQuery to use DeltaLake

**Related Issue(s)**

JustinRush80:
# Description
Add the GENERATE command that create an symlink manifest. 

I have done an very dumb solution for my day job but it definitely not scalable. So creating this draft to first understand the expected behavior of this command and then turn my initial solution to fit the code standard. 

# Related Issue(s)

- closes #2375

# Documentation

https://books.japila.pl/delta-lake-internals/post-commit-hooks/GenerateSymlinkManifest/


rtyler:
@JustinRush80 I have finished this functionality but would request that you test the generated manifest with an engine that requires it.

@roeap this is ready to merge once it's green in CI

JustinRush80:
> @JustinRush80 I have finished this functionality but would request that you test the generated manifest with an engine that requires it.
> 
> 
> 
> @roeap this is ready to merge once it's green in CI
> 
> 

Sure thing! Is this functionality able to incrementally update the manifest? I am still learning the code base but how does it handle partitions?

rtyler:
@JustinRush80 I have no idea what partitions are supposed to look like, if there is a specification of what this manifest is supposed to look like, I'm happy to help make this implementation compliant

rtyler:
@ion-elgreco this is ready for your consideration, I have done as much as I think we can so I believe this is ready for merge/release.

python/deltalake/table.py
```diff
@@ -507,6 +507,15 @@ def protocol(self) -> ProtocolVersions:
         """
         return ProtocolVersions(*self._table.protocol_versions())
 
+    def generate(self) -> None:
+        """
+        Generate symlink manifest for engines that cannot read native Delta Lake tables.
```
ion-elgreco, line 512: In the rust code its mentioned where this is placed, would be useful for the python users as well :)

JustinRush80:
I don't know if you wanted to have an fast follower after release: But I believe symlink manifest cannot be generated if Column Mapping is enabled based on delta-spark code.   `  /**
   * Generating manifests, when column mapping used is not supported,
   * because external systems will not be able to read Delta tables that leverage
   * column mapping correctly.
   */
  private def checkColumnMappingMode(metadata: Metadata): Unit = {
    if (metadata.columnMappingMode != NoMapping) {
      throw DeltaErrors.generateManifestWithColumnMappingNotSupported
    }
  }`

ion-elgreco:
> I don't know if you wanted to have an fast follower after release: But I believe symlink manifest cannot be generated if Column Mapping is enabled based on delta-spark code. ` /**
> 
> * Generating manifests, when column mapping used is not supported,
> * because external systems will not be able to read Delta tables that leverage
> * column mapping correctly.
>   */
>   private def checkColumnMappingMode(metadata: Metadata): Unit = {
>   if (metadata.columnMappingMode != NoMapping) {
>   throw DeltaErrors.generateManifestWithColumnMappingNotSupported
>   }
>   }`

Please do make a follow up to check if this feature is enabled and error out

PR #3890, opened 2025-10-22
feat: add support for blind appends

### LINKED ISSUE #2700: Optimizer Compact not running parallel to append writers
# Environment

**Delta-rs version**: 0.18.2

**Binding**: 

**Environment**: 
- **Cloud provider**: AWS
- **OS**: Mac OS
- **Other**:

***
# Bug

**What happened**:
When compaction is happening parallel to append writers, the compaction fails due to CommitFailedError.
```
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "local_testing.py", line 505, in optimizers
    dt.optimize.compact()
  File "/Users/code/delta-rs/venv/lib/python3.9/site-packages/deltalake/table.py", line 1899, in compact
    metrics = self.table._table.compact_optimize(
_internal.CommitFailedError: Failed to commit transaction: 15
```

Compaction is successful.

**How to reproduce it**:

```
import pandas as pd
import pyarrow as pa
import os
import random
import multiprocessing as mp
from time import sleep

# Creating a dataframe with 50 columns and 10000 rows
num_columns = 50
num_rows = 10000

data = {
    f'col{i + 1}': [f'row{random.randint(1, 2)}' for j in range(num_rows)] for i in range(num_columns)
}

df = pd.DataFrame(data)

table = pa.Table.from_pandas(df)
table_path = "s3a:/deltalake_test"
os.environ["AWS_ACCESS_KEY_ID"] = "***"
os.environ["AWS_ENDPOINT_URL"] = "***"
os.environ["AWS_SECRET_ACCESS_KEY"] = "***"

partition_cols = ["col1", "col2", "col3", "col4"]

storage_options = {
    'AWS_S3_LOCKING_PROVIDER': 'dynamodb',
    'DELTA_DYNAMO_TABLE_NAME': 'delta_log',
}

def writers():
    for i in range(10):
        write_deltalake(table_path, table, partition_by=partition_cols
                        , mode="append", engine="rust", storage_options=storage_options)


def optimizers():
    print("RUNNING OPTIMIZERS")
    dt = DeltaTable(table_path, storage_options=storage_options)
    dt.optimize.compact()
    print("COMPLETED OPTIMIZERS")


if __name__ == '__main__':

    tc = DeltaTable.create(
        table_uri=table_path,
        schema=table.schema,
        mode="append",
        partition_by=partition_cols,
        storage_options=storage_options,
    )

    processes = []
    pn = [f"process{i}" for i  in range(7)]

    for name in pn:
        process = mp.Process(target=writers)
        processes.append(process)

    optimizer = mp.Process(target=optimizers)

    for process in processes:
        process.start()

    sleep(10) # just to make it start after some data has been added
    optimizer.start()

    for process in processes:
        process.join()

    optimizer.join()

```

**More details**:

abhi-airspace-intelligence:
# Description

Consider the following case:

1) I kick off a compaction of a delta table
2) Simultaneously, I start a write
3) The compaction finishes, logically removing/adding some files
4) I go to commit my write

This will fail, since by default, the conflict checker assumes that the write depended on all the files in the table at a given state, this with `ConcurrentDeleteRead`. However, if it is a true blind append, the writer never depended on the state of the table. 

Instead, this PR marks a write commit as a blind append if it is in mode Append and there are no remove actions.

Marked as draft bc I'd like to write tests and confirm this is the correct behavior.

# Related Issue(s)

Likely closes #2700 

# Documentation

Reference here: https://books.japila.pl/delta-lake-internals/OptimisticTransactionImpl/?h=blind+app#isBlindAppend
Planning on diving into the delta-spark code when I have the time to ensure this is the correct heuristic.

### DIFF
--- a/crates/core/src/kernel/models/actions.rs
+++ b/crates/core/src/kernel/models/actions.rs
@@ -988,7 +988,9 @@ pub struct CommitInfo {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub isolation_level: Option<IsolationLevel>,
 
-    /// TODO
+    /// A flag indicating if the commit is a blind append.
+    /// A blind append is a write operation with mode Append that does not include any Remove actions.
+    /// https://books.japila.pl/delta-lake-internals/OptimisticTransactionImpl/?h=blind+app#isBlindAppend
     #[serde(skip_serializing_if = "Option::is_none")]
     pub is_blind_append: Option<bool>,
 
--- a/crates/core/src/kernel/transaction/conflict_checker.rs
+++ b/crates/core/src/kernel/transaction/conflict_checker.rs
@@ -513,6 +513,21 @@ impl<'a> ConflictChecker<'a> {
     fn check_for_deleted_files_against_current_txn_read_files(
         &self,
     ) -> Result<(), CommitConflictError> {
+        // For blind append writes, do not consider delete-read conflicts, since the
+        // transaction did not logically read any prior files.
+        let is_blind_append = self
+            .txn_info
+            .actions
+            .iter()
+            .find_map(|a| match a {
+                Action::CommitInfo(ci) => ci.is_blind_append,
+                _ => None,
+            })
+            .unwrap_or(false);
+        if is_blind_append {
+            return Ok(());
+        }
+
         // Fail if files have been deleted that the txn read.
         let read_file_path: HashSet<String> = self
             .txn_info
--- a/crates/core/src/kernel/transaction/mod.rs
+++ b/crates/core/src/kernel/transaction/mod.rs
@@ -96,8 +96,8 @@ use crate::kernel::{Action, CommitInfo, EagerSnapshot, Metadata, Protocol, Trans
 use crate::logstore::ObjectStoreRef;
 use crate::logstore::{CommitOrBytes, LogStoreRef};
 use crate::operations::CustomExecuteHandler;
-use crate::protocol::DeltaOperation;
 use crate::protocol::{cleanup_expired_logs_for, create_checkpoint_for};
+use crate::protocol::{DeltaOperation, SaveMode};
 use crate::table::config::TablePropertiesExt as _;
 use crate::table::state::DeltaTableState;
 use crate::{crate_version, DeltaResult};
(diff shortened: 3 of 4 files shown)


rtyler:
:wave: Thanks for starting the PR in draft to get the discussion going!

Blind appends are ... tricky :smile: 

For your specific situation I am curious if setting the `appendOnly` table property would suffice:
[append
only](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#append-only-tables)
If that table property doesn't solve the issue you describe, then I would say
we have a bug in our commit code that should be rectified, because on an
appendOnly table there should be no problem writing.

There's not really anything in the protocol that states what the rules should
be for a blind append. In the scenario that you describe it is probably safe
enough to issue an append but i'm not sure I would call that a blind append.

When I think of a blind append, I think about something that doesn't
care/consider whether the schema has been changed underneath either, i.e. isn't
considering if a metadata action happened between its current state and
whatever `$latest_version` is.

abhiaagarwal:
@rtyler the offending code is here:

https://github.com/delta-io/delta-rs/blob/main/crates/core/src/kernel/transaction/conflict_checker.rs#L209-L212

Yes, my tables are already marked as append only, but I still run into the conflict bug. Where if there is no read predicate specified, then all the files that were read in the snapshot are returned. However, in the case of blind appends, while the table state _was_ read, it doesn't have any actual impact on what was written. Additionally, even if a table is not AppendOnly by property, I'd still say that marking commits as blind appends if they are is a decent practice.

`isBlindAppend` isn't specified in the protocol, but delta-spark has special handling for it based on my look at the codebase.

abhiaagarwal:
For what it's worth, this PR has solved my problem entirely of being able to run concurrent appends + compactions as a background tokio task without any sort of concurrency issues.

abhi-airspace-intelligence:
I opened an issue in the delta-spark repository

PR #3894, opened 2025-10-23
feat: tracing spans across threadpool

ion-elgreco:
# Description

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/kernel/snapshot/stream.rs
+++ b/crates/core/src/kernel/snapshot/stream.rs
@@ -1,12 +1,13 @@
 //! the code in this file is hoisted from datafusion with only slight modifications
 //!
-use std::pin::Pin;
-
 use arrow::{datatypes::SchemaRef, record_batch::RecordBatch};
 use futures::stream::BoxStream;
 use futures::{Future, Stream, StreamExt};
+use std::pin::Pin;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio::task::JoinSet;
+use tracing::dispatcher;
+use tracing::Span;
 
 use crate::errors::DeltaResult;
 use crate::DeltaTableError;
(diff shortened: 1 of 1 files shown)


chiragjn:
I tried this PR, I still see disconnection after `Snapshot::try_new_with_engine`

I have forked kernel and added decorators on `for_snapshot` and `read_metadata`

image

I am mostly fumbling around at this point and my observation is for `spawn_blocking`, I need to wrap like this to make it work

```
let dispatch = tracing::dispatcher::get_default(|d| d.clone());
let span = tracing::Span::current();
let result = spawn_blocking(move || {
    tracing::dispatcher::with_default(&dispatch, || {
        let _enter = span.enter();
        // actual logic
```

ion-elgreco:
@chiragjn hmm I see, I've updated the code to reflect that now, can you try again? I guess the key thing is that .instrument only enters the current span when polling the future but the dispatcher also needs to be re-used which doesn't happen with tracing-futures :(

chiragjn:
Tried the PR again, works fine 🎉 

I think a few places are missing - `replay_for_scan_metadata` from kernel shows up disconnected 

I believe these are from `scan.rs`
https://github.com/delta-io/delta-rs/blob/acfaee5c07e47e9dadb3bbb5201ab75b245bf9b1/crates/core/src/kernel/snapshot/scan.rs#L136-L145

I am able to nest correctly on my fork where I have changed all `spawn_blocking` in `scan.rs`

ion-elgreco:
@chiragjn have added that as well, should be good to go now

ion-elgreco:
@roeap can you stamp it again? :)

rtyler (approved):
I want to incorporate this in 0.29.2 which I'll be releasing today

PR #3895, opened 2025-10-23
feat(memory): optimize collection preallocation where capacity is known

fvaleye:
# Description
Following [this](https://github.com/delta-io/delta-rs/pull/3826), this work implements a first round of preallocating collections when their size is known, to avoid unnecessary reallocations. This applies to `Vec`, `HashMap` (with_capacity), and other dynamically sized collections. 
Preallocating memory improves efficiency and reduces overhead when the final size is known.


roeap (approved):
🚀

PR #3901, opened 2025-10-27
feat: consolidate `target_file_size` and allow unbounded writes

### LINKED ISSUE #3855: Compacting a delta table consistently undershoots the `target_file_size`, creating an unnecessary extra file

abhi-airspace-intelligence:
# Description

I was trying to investigate the issue in 3855, and I believe I've figured it out. The problem is that when we perform a compaction, we create a bin of files where their total sum is > the target file size, and then, when we go to write the file, we _also_ apply the exact same target file size. This means if, say, the size of all the files in your bin is 105MB, you write a 100MB file _and_ a 5MB file, which I would argue is undesired behavior. Additionally, due to compression/rounding, I've observed that the 100MB file _might_ be written as 97MB and 4MB respectively, which then leads to infinite compactions on the same set of file that generates wasteful IO.

Instead, this changes compactions to allow a target_file_size to be unbounded when actually performing the rewrite operation. To do this, I had to replumb a lot of the code since everything makes the assumption of target-size writes. Additionally, I found there are three separate places where a default target size is set, so I consolidated them to just one place.

# Related Issue(s)

Closes #3855 

# Documentation

### DIFF
--- a/crates/core/src/operations/write/mod.rs
+++ b/crates/core/src/operations/write/mod.rs
@@ -617,12 +617,7 @@ impl std::future::IntoFuture for WriteBuilder {
                     .map(|snapshot| snapshot.table_properties());
 
                 let target_file_size = this.target_file_size.unwrap_or_else(|| {
-                    Some(
-                        this.snapshot
-                            .as_ref()
-                            .map(|snapshot| snapshot.table_properties().target_file_size())
-                            .unwrap_or(DEFAULT_TARGET_FILE_SIZE),
-                    )
+                    Some(super::get_target_file_size(config, &this.configuration))
                 });
 
                 let (num_indexed_cols, stats_columns) =
--- a/crates/core/src/operations/mod.rs
+++ b/crates/core/src/operations/mod.rs
@@ -7,6 +7,8 @@
 //! with a [data stream][datafusion::physical_plan::SendableRecordBatchStream],
 //! if the operation returns data as well.
 use std::collections::HashMap;
+#[cfg(feature = "datafusion")]
+use std::num::NonZeroU64;
 use std::sync::Arc;
 
 #[cfg(feature = "datafusion")]
@@ -404,6 +406,23 @@ pub fn get_num_idx_cols_and_stats_columns(
             .clone()
             .map(|v| v.iter().map(|v| v.to_string()).collect::<Vec<String>>()),
     )
+}
+
+/// Get the target_file_size from the table configuration in the sates
+/// If table_config does not exist (only can occur in the first write action) it takes
+/// the configuration that was passed to the writerBuilder.
+#[cfg(feature = "datafusion")]
+pub(crate) fn get_target_file_size(
+    config: Option<&TableProperties>,
+    configuration: &HashMap<String, Option<String>>,
+) -> NonZeroU64 {
+    match &config {
+        Some(conf) => conf.target_file_size(),
+        _ => configuration
+            .get("delta.targetFileSize")
+            .and_then(|v| v.clone().and_then(|v| v.parse::<NonZeroU64>().ok()))
+            .unwrap_or(crate::table::config::DEFAULT_TARGET_FILE_SIZE),
+    }
 }
 
 #[cfg(feature = "datafusion")]


crates/core/src/operations/write/mod.rs
```diff
@@ -143,7 +145,8 @@ pub struct WriteBuilder {
     /// When using `Overwrite` mode, replace data that matches a predicate
     predicate: Option<Expression>,
     /// Size above which we will write a buffered parquet file to disk.
-    target_file_size: Option<usize>,
+    /// If None, the writer will not create a new file until the writer is closed.
+    target_file_size: Option<Option<NonZeroUsize>>,
```
ion-elgreco: Maybe just flatten it instead of nesting

crates/core/src/operations/write/mod.rs
```diff
@@ -616,9 +619,16 @@ impl std::future::IntoFuture for WriteBuilder {
                     .as_ref()
                     .map(|snapshot| snapshot.table_properties());
 
-                let target_file_size = this.target_file_size.or_else(|| {
-                    Some(super::get_target_file_size(config, &this.configuration) as usize)
+                let target_file_size = this.target_file_size.unwrap_or_else(|| {
+                    this.snapshot.as_ref().and_then(|snapshot| {
```
ion-elgreco: I think this will cause a first write to be unbounded, no? Not sure if thats something we want

crates/core/src/operations/optimize.rs
```diff
@@ -549,7 +549,8 @@ impl MergePlan {
             task_parameters.file_schema.clone(),
             partition_values.clone(),
             Some(task_parameters.writer_properties.clone()),
-            Some(task_parameters.input_parameters.target_size as usize),
+            // Since we know the total size of the bin, we can set the target file size to None.
```
ion-elgreco, line 554: 👍

crates/core/src/operations/write/mod.rs
abhi-airspace-intelligence: It was done this way to support the following usecases:

`None`: bounded writes, fallback to the delta table configuration
`Some(None)`: unbounded writes
`Some(val)`: bounded writes, overriding delta table configuration

crates/core/src/operations/write/mod.rs
abhi-airspace-intelligence: Good catch, it should fallback to the default size.

abhi-airspace-intelligence:
This moves a bunch of a bunch of parameters to be `Option<NonZeroUsize>`, which enables niche optimization, but is also a breaking API change. https://doc.rust-lang.org/std/num/type.NonZeroUsize.html

abhi-airspace-intelligence:
Also, now that I think about this, using `usize` for a file size is probably an antipattern... perhaps, it should be switched to an `u64`? `object_store` docs reference this here. https://docs.rs/object_store/latest/object_store/struct.ObjectMeta.html#structfield.size.

@rtyler thoughts?

crates/core/src/operations/write/mod.rs
ion-elgreco: I see, but maybe we can use an enum for this, it's not quite apparent when you look at it :)

ion-elgreco:
@abhi-airspace-intelligence any updates on this PR?

abhiaagarwal:
Hey there, I've been using this in production, it works and solves my use case. Problem is mostly that I haven't written tests for it yet. I'll see if I can do that this weekend.

abhiaagarwal:
Gonna close this PR on my work account and instead open it on my personal, since I rarely have time to work on delta-rs professionally.

PR #3903, opened 2025-10-28
perf(snapshot): minor memory allocation and usage reduction without cloning

fvaleye:
# Description
Following this [PR](https://github.com/delta-io/delta-rs/pull/3826):
- use `std::mem:take` to reduce memory allocation and consumption without cloning in updating `EagerSnapshot`

# Benchmark
From 100 rows to 1k rows, on my local laptop with a simple divan script.

Test case with a small table (a few Parquet files):
- 5-10 files
- 100-1,000 rows created only a few files

Results:
- Fewer allocations: 16-56 allocations eliminated per update() call
- Less memory: 2-10 KB saved per operation
- Fewer deallocations: 114-154 fewer deallocations

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,4 +116,3 @@ inherits = "release"
 opt-level = 3
 codegen-units = 1
 lto = "fat"
-strip = "debuginfo"


crates/core/src/kernel/snapshot/mod.rs
hntd187, line 42: I think github really wants you to remove this import.

hntd187 (approved):

fvaleye:
Thanks @hntd187!

Issue #3899, opened 2025-10-26
[Feature]: split `delta-rs` into multiple crates


abhiaagarwal:
### Is your feature request related to a problem?

Developing on `delta-rs` can be a tough experience, owning to the long compilation times. This is particularly exacerbated by the datafusion dependency, which itself is huge and blocks compiling the rest of the crate.

### Describe the solution you'd like

Instead, we should split delta-rs into multiple crates. For example, the "protocol" folder could become `deltalake-protocol`, the "logstore" folder could become `deltalake-logstore`, `deltalake-operations` / `deltalake-datafusion-operations`, `deltalake-datafusion`. This way, we benefit from build parallelization and can avoid huge recompiles.

### Describe alternatives you've considered

N/A

### Priority

Medium - Would be helpful

### Contribution

- [x] I'm willing to submit a pull request for this feature
- [ ] I can help with testing this feature
- [ ] I can help with documentation for this feature

rtyler:
I did the original crate refactor to reduce our dependency scope across the board. I also do a fair bit of complaining to @roeap about the datafusion dependency 😆 

I am not in favor of further breaking things apart into crates since it leads to non-trivial release engineering work for me. 

What can help would be refactoring into smaller modules. Smaller modules leads to smaller compilation units and less that needs to be re-compiled. Ensuring datafusion is only pulled in when it needs to be also helps.  The link times still will suffer because of the long serial chain of datafusion dependencies. 😦 

Some tips to help:

* use `mold` as a linker
* use sccache
* cargo-nextest

The arrow and datafusion dependencies are kind of a big linear boat anchor on our build process, and there's not many _easy_ fixes

abhiaagarwal:
> Smaller modules leads to smaller compilation units

The crate is the compilation unit in rust, so the only real way is to split things up into crates. Illustrative article: https://www.feldera.com/blog/cutting-down-rust-compile-times-from-30-to-2-minutes-with-one-thousand-crates

rtyler:
The feldera folks are doing some similar wildness to what the [arroyo](https://arroyo.dev) team was doing. I'm happy to not have to worry about those kind of compilation woes 😄 

I re-ran some of my `cargo build --timings` testing and really the fastest possible build is: sccache + cranelift + parallelized frontend  without the `datafusion` feature flag. The datafusion build pipeline adds around 100% more compilation/link time compared to building core without it.

[...]

PR #3913, opened 2025-11-02
feat: extract log store into its own crate, `deltalake-logstore`

abhiaagarwal:
# Description

Extracts the logstore delta lake component into its own crate, making core take a dependency on it. This allows the aws, gcp, hdfs, etc implementations to rely exclusively on `deltalake-logstore` instead of depending on core, flipping the dependency graph and decreasing compile times (about a 30% decrease on my system).

# Related Issue(s)

Progress towards #3899 

# Documentation

Still a WIP, but compiles locally without fixing the tests (which mostly require changing changing `DeltaResult` to `LogStoreResult`.


abhiaagarwal:
Timings of this branch: 
[timing-log-store.html](https://github.com/user-attachments/files/23291519/timing-log-store.html)

Timings of main: 
[timing-main.html](https://github.com/user-attachments/files/23291520/timing-main.html)

While the overall timing is the same (due to all the build targets such as tests and python still requiring `deltalake_core`), you can see that this drops `deltalake_core`'s overall compile time from 20s -> 15s and enables it to be compiled 10s earlier due to increased parallelization. The biggest benefit is the improvement in rust-analyzer re-running on code changes, since the aws/gcp/etc (unfortunately, LakeFS wasn't possible) doesn't require a re-compile anymore.

(all after a cargo clean, running `cargo build --all-targets --timings`)

PR #3916, opened 2025-11-04
fix: update stats serialization logic for scale-0 decimals

DrakeLin:
# Description
Creating Delta tables with Decimal(precision, scale=0) columns fails since in stats code, we store decimals as f64. serde_json converts f64 values like 1234.0 to 1234.0 for scale=0. Arrow Parser then rejects it since it expects 1234.

Modified StatsScalar enum to preserve scale information:

  - Changed from Decimal(f64) to Decimal { value: f64, scale: i32 }
  - Updated serialization logic to output integers for scale=0 decimal

Tested by building delta-rs and running the example script linked in the issue.

# Related Issue(s)

https://github.com/delta-io/delta-kernel-rs/issues/1431

# Documentation

### DIFF
--- a/crates/core/src/writer/stats.rs
+++ b/crates/core/src/writer/stats.rs
@@ -244,7 +244,8 @@ enum StatsScalar {
     Date(chrono::NaiveDate),
     Timestamp(chrono::NaiveDateTime),
     // We are serializing to f64 later and the ordering should be the same
-    Decimal(f64),
+    // Scale is stored to handle scale=0 serialization correctly
+    Decimal { value: f64, scale: i32 },
     String(String),
     Bytes(Vec<u8>),
     Uuid(uuid::Uuid),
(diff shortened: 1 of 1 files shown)


rtyler:
@DrakeLin is this independent of any kernel changes, or is this pending something in kernel to be released?

DrakeLin:
@rtyler This is independent of any kernel changes. The bug seen in https://github.com/delta-io/delta-kernel-rs/issues/1431 didn't turn out to be a kernel issue (delta-rs doesn't use kernel for writes), so putting up a quick fix in delta-rs. The bug, as the original author pointed out, is due to an issue in serde_json.

This bug caused delta-rs to write stats with scale 0 out as 1234.0. The kernel issue we saw was due to kernel trying to read this and failing. I think there is a kernel fix we can push out as well to relax our constraints on read, but that's independent of this change.

PR #3922, opened 2025-11-06
fix: handle empty tables in get_add_actions()

### LINKED ISSUE #3918: [Bug]: `get_add_actions()` panics with "index out of bounds" when table has no data files
### What happened?

When calling `get_add_actions()` on a Delta table that has no add actions (i.e., no data files), the function panics with `exception: PanicException`
`exc_value: index out of bounds: the len is 0 but the index is 0`

This is a regression from version 1.1.4, where this scenario worked correctly without errors.

**Root Cause:**
The issue is in `crates/core/src/table/state.rs` at line 358 in the `EagerSnapshot::add_actions_table()` function:

```rust
let result = concat_batches(results[0].schema_ref(), &results)?;
```

This code attempts to access `results[0]` without checking if the `results` vector is empty. When there are no files in the table, `self.files()?` returns an empty iterator, resulting in an empty `results` vector, which causes the index out of bounds panic.

**Expected behavior**

The function should handle empty tables gracefully and return an empty RecordBatch with the appropriate schema, similar to how it behaved in version 1.1.4.

**Binding**
Python (also affects Rust)

**Bindings Version**
Versions after 1.1.4 (current main branch affected)

**Steps to reproduce**

1. Create or load a Delta table with no data files (e.g., a newly initialized table or a table where all data has been deleted)
2. Call `get_add_actions()` on the table
3. Observe the panic

Example code:
```python

# Load a table with no data files
dt = DeltaTable("path/to/empty/table")
dt.get_add_actions()  # This will panic
```

### Expected behavior

The function should handle empty tables gracefully and return an empty RecordBatch with the appropriate schema, similar to how it behaved in version 1.1.4.

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

_No response_

### Relevant logs

```shell

```

vsmanish1772:
## Description

This PR fixes a panic that occurs when `get_add_actions()` is called on a Delta table that has no add actions (i.e., no data files).

### Problem

When calling `get_add_actions()` on a table with no files, the function panics with:
```
index out of bounds: the len is 0 but the index is 0
```

This happens because the code attempts to access `results[0]` without checking if the `results` vector is empty.

### Solution

- Added early return check for empty files before creating the evaluator
- When files are empty, create an empty RecordBatch with the correct schema
- This matches the behavior from version 1.1.4 where empty tables were handled gracefully

### Changes
- Check if files are empty before processing
- Return empty RecordBatch with correct schema when no files exist

### Testing

- Added unit test that covers the empty table scenario
- Test verifies both flattened and non-flattened schemas

Fixes #3918

### DIFF
--- a/crates/core/src/table/state.rs
+++ b/crates/core/src/table/state.rs
@@ -383,3 +383,25 @@ impl EagerSnapshot {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{DeltaOps, DeltaResult};
+
+    /// <https://github.com/delta-io/delta-rs/issues/3918>
+    #[tokio::test]
+    async fn test_add_actions_empty() -> DeltaResult<()> {
+        let table = DeltaOps::new_in_memory()
+            .create()
+            .with_column(
+                "id",
+                DataType::Primitive(delta_kernel::schema::PrimitiveType::Long),
+                true,
+                None,
+            )
+            .await?;
+        let _actions = table.snapshot()?.add_actions_table(false)?;
+        Ok(())
+    }
+}


vsmanish1772:
can someone look into this please? The build complains about formatting but locally it doesnt throw that error

rtyler:
@vsmanish1772 what version of Rust are you building with? CI runs with 1.91 or whatever is stable, but I see the issues with 1.90 as well

vsmanish1772:
> @vsmanish1772 what version of Rust are you building with? CI runs with 1.91 or whatever is stable, but I see the issues with 1.90 as well

@rtyler  I'm using rustc 1.91.0 (f8297e351 2025-10-28)

PR #3929, opened 2025-11-11
fix: correctly rectify Urls with dots in DeltaTableBuilder

rtyler:
While doing some unrelated work I did some dry-by test coverage addition and discovered [this issue](https://github.com/servo/rust-url/issues/1086) with some of our DAT tests.

The only way I believe we can be certain our path segments don't have `..` in them is to re-parse the `Url` :cry:

### DIFF
--- a/crates/core/src/kernel/snapshot/log_data.rs
+++ b/crates/core/src/kernel/snapshot/log_data.rs
@@ -16,29 +16,32 @@ pub(crate) trait PartitionsExt {
 
 impl PartitionsExt for IndexMap<String, Scalar> {
     fn hive_partition_path(&self) -> String {
-        let fields = self
-            .iter()
-            .map(|(k, v)| {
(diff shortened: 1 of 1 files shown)


crates/core/src/kernel/snapshot/log_data.rs
```diff
@@ -16,29 +16,32 @@ pub(crate) trait PartitionsExt {
 
 impl PartitionsExt for IndexMap<String, Scalar> {
     fn hive_partition_path(&self) -> String {
-        let fields = self
-            .iter()
-            .map(|(k, v)| {
+        let sep = String::from('/');
```
hntd187, line 19: Would this clone the string on every interspersion? Pedantic here, but maybe a `const SEP: &str = "/";` would make more sense?

hntd187 (approved):
One question, but otherwise GTG

PR #3940, opened 2025-11-19
fix: remove clippy warnings

### LINKED ISSUE #3907: [Chore]: Proposal, remove clippy warnings

In working with the project to add enhancements, I've been struggling with the large number of warnings returned by
```
make check-rust
```
I would like to propose a PR to clean up the warnings but am not sure how it would be received. What I would propose is:
1. Set a flag to turn off the deprecation warnings since they make the other warnings hard to spot. Something in the build properties equivalent to:
```
RUSTFLAGS=-Adeprecated
```
2. Do a PR to eliminate the remaining warnings so we can get a clean clippy build back.

I think I am not the only developer struggling with this since I see the number of clippy warnings are increasing so I just wanted to see what others thought. I do think clippy can be helpful, but right now it is hard to use the output with so many warnings.

Clippy warnings removed.

_No response_

### Priority

None

### Additional context

_No response_

### Contribution

corwinjoy:
# Description
1. Set a flag to turn off the deprecation warnings. They clutter the error messages and we can turn this back on when we are ready to remove the deprecated functions.
2. Go through the remaining clippy warnings and either fix them or disable them if it would mean changing the interface.

# Related Issue(s)
Closes #3907

### DIFF
--- a/crates/core/src/logstore/config.rs
+++ b/crates/core/src/logstore/config.rs
@@ -236,9 +236,7 @@ impl StorageConfig {
     }
 }
 
-pub(super) fn try_parse_impl<T, K, V, I>(
-    options: I,
-) -> DeltaResult<(T, HashMap<String, String>)>
+pub(super) fn try_parse_impl<T, K, V, I>(options: I) -> DeltaResult<(T, HashMap<String, String>)>
 where
     I: IntoIterator<Item = (K, V)>,
     K: AsRef<str> + Into<String>,


crates/core/src/kernel/schema/mod.rs
```diff
@@ -2,6 +2,7 @@ use std::any::Any;
 
 pub mod cast;
 pub mod partitions;
+#[allow(clippy::module_inception)]
```
corwinjoy, line 5: Clippy does not like the repeated name per https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#module_inception

But, changing this could break API usage so I just disabled the warning.

hntd187:
I'm not generally a fan of this. I think we should fix the deprecations rather than just silence the warnings. That's the entire reason they exist in the first place. Additionally, why are there changes in this PR entirely unrelated to the clippy warnings?

corwinjoy:
@hntd187  Here is my reasoning on the deprecation flag.
1.  Many of these deprecation warnings are similar to
```
warning: use of deprecated method `kernel::snapshot::EagerSnapshot::file_views_by_partitions`: Use `files` with kernel predicate instead.
    --> crates/core/src/operations/optimize.rs:1057:36
     |
1057 |     let mut file_stream = snapshot.file_views_by_partitions(log_store,...
```
I think these came about with the new kernel, but I'm not sure this update has been finished yet, so a new method may not exist.
2. There are so many deprecation warnings that it really clutters the warning space. Making it very hard to spot any new warnings when developing on this project. What I have seen over the last month or so is that the number of clippy warnings is growing, since I think folks have just started to ignore this.
3. Adding a deprecation flag helps to break this PR into more manageable pieces. When we are ready to address the deprecations (not sure if this is possible now due to the kernel stuff), we can turn these back on.

--------------------------------------------------------------------------------------------------------------------------------------------

> Additionally, why are there changes in this PR entirely unrelated to the clippy warnings?

In terms of your other question, as far as I can tell every one of these changes should be associated with a clippy warning message. I am running a double check to confirm. But, maybe I missed something? If so, could you please point out the incorrect change?

corwinjoy:
@adamreeve @abhiaagarwal

crates/benchmarks/src/main.rs
```diff
@@ -197,7 +197,7 @@ async fn run_merge_with_params(
 
 async fn run_merge_case(case: &MergeTestCase, parquet_dir: &Path) -> anyhow::Result<()> {
     let tmp_dir = tempfile::tempdir()?;
-    let (source, table) = prepare_source_and_table(&case.params, &tmp_dir, &parquet_dir).await?;
+    let (source, table) = prepare_source_and_table(&case.params, &tmp_dir, parquet_dir).await?;
```
corwinjoy, line 200: Crate: crates/benchmarks File: crates/benchmarks/src/main.rs (run_merge_case) Clippy warning fixed: "this expression creates a reference which is immediately dereferenced by the compiler --> crates/benchmarks/src/main.rs:200:76 help: change this to: parquet_dir" Change: removed the needless borrow and passed parquet_dir directly to prepare_source_and_table.

crates/core/src/delta_datafusion/planner.rs
```diff
@@ -39,7 +39,7 @@ use crate::operations::merge::MergeMetricExtensionPlanner;
 use crate::operations::update::UpdateMetricExtensionPlanner;
 use crate::operations::write::metrics::WriteMetricExtensionPlanner;
 
-const DELTA_EXTENSION_PLANNERS: LazyLock<Vec<Arc<dyn ExtensionPlanner + Send + Sync>>> =
+static DELTA_EXTENSION_PLANNERS: LazyLock<Vec<Arc<dyn ExtensionPlanner + Send + Sync>>> =
```
corwinjoy, line 42: Crate: crates/core File: crates/core/src/delta_datafusion/planner.rs Clippy warnings fixed: "named constant with interior mutability --> crates/core/src/delta_datafusion/planner.rs:42:7 = help: did you mean to make this a static item" and "named constant with interior mutability --> crates/core/src/delta_datafusion/planner.rs:52:7" Change: converted the LazyLock const items to static (e.g., static DELTA_EXTENSION_PLANNERS and static DELTA_PLANNER) so interior mutability is handled correctly.

crates/core/src/kernel/arrow/engine_ext.rs
```diff
@@ -399,7 +399,7 @@ pub(crate) fn kernel_to_arrow(metadata: ScanMetadata) -> DeltaResult<ScanMetadat
     )?;
     Ok(ScanMetadataArrow {
         scan_files,
-        scan_file_transforms,
+        _scan_file_transforms: scan_file_transforms,
```
corwinjoy, line 402: Crate: crates/core File: crates/core/src/kernel/arrow/engine_ext.rs (ScanMetadataArrow) Clippy warning fixed: "field scan_file_transforms is never read --> crates/core/src/kernel/arrow/engine_ext.rs:44:5" Change: renamed the field to _scan_file_transforms and updated construction to _scan_file_transforms: scan_file_transforms to mark it intentionally unused.

crates/core/src/kernel/models/actions.rs
```diff
@@ -1060,11 +1057,6 @@ pub enum IsolationLevel {
 
 // Spark assumes Serializable as default isolation level
 // https://github.com/delta-io/delta/blob/abb171c8401200e7772b27e3be6ea8682528ac72/core/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala#L1023
-impl Default for IsolationLevel {
-    fn default() -> Self {
-        Self::Serializable
-    }
-}
```
corwinjoy, line 1067: Crate: crates/core File: crates/core/src/kernel/models/actions.rs Clippy warnings fixed: "associated function new is never used --> crates/core/src/kernel/models/actions.rs:271:19" and "this impl can be derived --> crates/core/src/kernel/models/actions.rs:706:1" Change: added #[allow(dead_code)] to the unused ProtocolInner::new method; replaced manual impl Default with #[derive(Default)] and marked default variants with #[default] for StorageType and IsolationLevel per clippy suggestions.

crates/core/src/kernel/snapshot/log_data.rs
```diff
@@ -211,7 +211,7 @@ mod datafusion {
 
                 if let Some(mut accumulator) = accumulator {
                     return accumulator
-                        .update_batch(&[array.clone()])
+                        .update_batch(std::slice::from_ref(array))
```
corwinjoy, line 214: Crate: crates/core File: crates/core/src/kernel/snapshot/log_data.rs Clippy warnings fixed: "unnecessary use of to_vec --> crates/core/src/kernel/snapshot/log_data.rs:102:18 help: use: self.data.iter().cloned()" and "this call to clone can be replaced with std::slice::from_ref --> crates/core/src/kernel/snapshot/log_data.rs:214:39 help: try: std::slice::from_ref(array)" Change: updated IntoIterator signature to carry the lifetime (impl<'a> IntoIterator for LogDataHandler<'a> and type IntoIter = Box<dyn Iterator<Item = Self::Item> + 'a>), replaced .to_vec().into_iter() with self.data.iter().cloned(), and replaced .update_batch(&[array.clone()]) with .update_batch(std::slice::from_ref(array)).

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -479,8 +479,10 @@ pub(crate) async fn resolve_snapshot(
             Ok(snapshot)
         }
     } else {
-        let mut config = DeltaTableConfig::default();
-        config.require_files = require_files;
+        let config = DeltaTableConfig {
+            require_files,
+            ..Default::default()
+        };
```
corwinjoy, line 485: Crate: crates/core File: crates/core/src/kernel/snapshot/mod.rs (resolve_snapshot) Clippy warning fixed: "field assignment outside of initializer for an instance created with Default::default() --> crates/core/src/kernel/snapshot/mod.rs:483:9 note: consider initializing the variable with DeltaTableConfig { require_files: require_files, ..Default::default() }" Change: constructed config with struct update syntax: let config = DeltaTableConfig { require_files, ..Default::default() }; instead of creating default and reassigning a field.

crates/core/src/kernel/snapshot/stream.rs
```diff
@@ -89,6 +89,7 @@ impl<O: Send + 'static> ReceiverStreamBuilder<O> {
 
     /// Spawn task that will be aborted if this builder (or the stream
     /// built from it) are dropped
+    #[allow(dead_code)]
```
corwinjoy, line 92: Crate: crates/core File: crates/core/src/kernel/snapshot/stream.rs (ReceiverStreamBuilder) Clippy warning fixed: "method spawn is never used --> crates/core/src/kernel/snapshot/stream.rs:92:12" Change: added #[allow(dead_code)] to the unused spawn method to silence the dead-code lint.

crates/core/src/logstore/config.rs
```diff
@@ -236,14 +236,12 @@ impl StorageConfig {
...
-pub(super) fn try_parse_impl<T: std::fmt::Debug, K, V, I>(
-    options: I,
-) -> DeltaResult<(T, HashMap<String, String>)>
+pub(super) fn try_parse_impl<T, K, V, I>(options: I) -> DeltaResult<(T, HashMap<String, String>)>
 where
     I: IntoIterator<Item = (K, V)>,
     K: AsRef<str> + Into<String>,
     V: AsRef<str> + Into<String>,
-    T: TryUpdateKey,
+    T: TryUpdateKey + std::fmt::Debug,
```
corwinjoy, line 244: Crate: crates/core File: crates/core/src/logstore/config.rs (try_parse_impl signature) Clippy warning fixed: "bound is defined in more than one place --> crates/core/src/logstore/config.rs:239:30 = help: for further information visit https://rust-lang.github.io/rust-clippy/...#multiple_bound_locations" Change: consolidated type bounds by moving std::fmt::Debug into the where-clause so T bounds are defined in one place: pub(super) fn try_parse_impl<T, K, V, I>(...) -> ... where ... T: TryUpdateKey + std::fmt::Debug, ....

crates/core/src/operations/constraints.rs
```diff
@@ -177,8 +178,8 @@ impl std::future::IntoFuture for ConstraintBuilder {
                 tracing::warn!("Skipping constraint '{name}': identical constraint already exists with expression '{}'",constraints_sql_mapper[name]);
             }
             let constraints_checker: Vec<Constraint> = constraints_sql_mapper
-                .iter()
-                .map(|(_, sql)| Constraint::new("*", sql))
+                .values()
+                .map(|sql| Constraint::new("*", sql))
```
corwinjoy, line 182: Crate: crates/core File: crates/core/src/operations/constraints.rs Clippy warnings fixed: "iterating on a map's keys --> crates/core/src/operations/constraints.rs:131:17 help: try: this.check_constraints.keys().map(|name| ...)" and "iterating on a map's values --> crates/core/src/operations/constraints.rs:179:56 help: try: constraints_sql_mapper.values().map(|sql| ...)" Change: iterated over keys/values directly: replaced .iter().map(|(name, _)| ...) with .keys().cloned().map(|name| ...), and .iter().map(|(_, sql)| ...) with .values().map(|sql| ...).

crates/core/src/operations/vacuum.rs
```diff
@@ -279,7 +279,7 @@ impl VacuumBuilder {
         };
 
         let expired_tombstones =
-            get_stale_files(&snapshot, retention_period, now_millis, &self.log_store).await?;
+            get_stale_files(snapshot, retention_period, now_millis, &self.log_store).await?;
```
corwinjoy, line 282: Crate: crates/core File: crates/core/src/operations/vacuum.rs Clippy warning fixed: "this expression creates a reference which is immediately dereferenced by the compiler --> crates/core/src/operations/vacuum.rs:282:29 help: change this to: snapshot" Change: removed the needless borrow and passed snapshot directly to get_stale_files(snapshot, ...).

crates/core/src/operations/write/execution.rs
```diff
@@ -561,6 +561,6 @@ pub(crate) async fn write_execution_plan_v2(
             write_time_ms,
         };
 
-        return Ok((actions, metrics));
+        Ok((actions, metrics))
```
corwinjoy, line 564: Crate: crates/core File: crates/core/src/operations/write/execution.rs Clippy warnings fixed: "unneeded return statement --> crates/core/src/operations/write/execution.rs:372:9" and "unneeded return statement --> crates/core/src/operations/write/execution.rs:564:9" Change: removed redundant return keywords and returned the expressions directly (Ok((actions, metrics))).

crates/core/src/operations/write/mod.rs
```diff
@@ -82,9 +81,6 @@ pub(crate) enum WriteError {
     #[error("No data source supplied to write command.")]
     MissingData,
 
-    #[error("Failed to execute write task: {source}")]
-    WriteTask { source: tokio::task::JoinError },
-
```
corwinjoy, line 87: Crate: crates/core File: crates/core/src/operations/write/mod.rs Clippy warnings fixed: "unused import: crate::delta_datafusion::planner::DeltaPlanner --> crates/core/src/operations/write/mod.rs:59:5" and "variant WriteTask is never constructed --> crates/core/src/operations/write/mod.rs:86:5" Change: removed the unused DeltaPlanner import and deleted the unused WriteTask { source: tokio::task::JoinError } enum variant.

crates/core/src/kernel/snapshot/stream.rs
corwinjoy: Not changed since I don't want to potentially break the API.

crates/core/src/kernel/models/actions.rs
```diff
@@ -268,6 +268,7 @@ impl Default for ProtocolInner {
 
 impl ProtocolInner {
     /// Create a new protocol action
+    #[allow(dead_code)]
```
corwinjoy, line 271: Crate: crates/core File: crates/core/src/kernel/models/actions.rs Clippy warnings fixed: "associated function new is never used --> crates/core/src/kernel/models/actions.rs:271:19" 

Not removed because this could break the API.

corwinjoy:
OK. With the help of copilot I have added comments listing the warning message and what was done for each of these.

crates/core/src/delta_datafusion/planner.rs
hntd187: Where is this modified such that interior mutability is relevant here?

corwinjoy: Here is the full warning message. I tried just assigning to a local variable but the fix is a bit trickier than that I guess.

```
warning: borrow of a named constant with interior mutability
  --> crates/core/src/delta_datafusion/planner.rs:98:28
   |
98 |         for ext_planner in DELTA_EXTENSION_PLANNERS.iter() {
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: there is a compiler inserted call to `Deref::deref` here
   = help: this lint can be silenced by assigning the value to a local variable before borrowing
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#borrow_interior_mutable_const
   = note: `#[warn(clippy::borrow_interior_mutable_const)]` on by default
```

corwinjoy: @hntd187 I could use some help on a better way to fix this error since simple assignment to a temporary doesn't quite do it. Here are the docs on this
https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#borrow_interior_mutable_const
So, potentially, we could mark this to be ignored but that doesn't seem too great.

hntd187: My point was I don't believe this lint is correct. We don't modify this value here, or anywhere I'm aware of. Perhaps just 
```rust
let planners = DELTA_EXTENSION_PLANNERS.clone();
for planner in planners.iter() {
  ...
}
```
might work but I'd have to look into it myself.

corwinjoy: Hmm. That also doesn't work. But, I think this should actually be a `static` here because of the use of `LazyLock`.
`const` here means a compile-time constant. (See: https://stackoverflow.com/questions/37405835/populating-a-static-const-with-an-environment-variable-at-runtime-in-rust). But `LazyLock` is something different; it is actually initialized on first usage.
https://doc.rust-lang.org/std/sync/struct.LazyLock.html

corwinjoy:
Closing due to lack of response. Hopefully the maintainers take a look at the clippy warnings because the long list makes it hard to spot new messages and maintain code quality.

Issue #930, opened 2022-11-10
Support column mapping


dennyglee:
# Description
Per [1076](https://github.com/delta-io/delta/pull/1076), it would be great for delta-rs to support column mapping.  With Delta tables that support column mapping, the altering or dropping of columns is a metadata only operation.

**Use Case**
As folks can enable column mapping for Delta tables, it would be great for delta-rs to support this feature.

**Related Issue(s)**
[#1076](https://github.com/delta-io/delta/pull/1076), [#1064](https://github.com/delta-io/delta/pull/1064)

aersam:
On python side, I guess it's not possible to implement with the current `to_pyarrow_dataset` method, since Dataset does not support a distinction between physical and logical column names. I opened an issue about that: https://github.com/apache/arrow/issues/36593

With pyarrow Table would be fairly straight forward, one could probably also expose a [RecordBatchReader](https://arrow.apache.org/docs/python/generated/pyarrow.RecordBatchReader.html#pyarrow.RecordBatchReader.from_batches) based on a dataset which uses physical column names.

keen85:
@rtyler, are there any plans to support [column mappings (reader v2, writer v5)](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-mapping) in the future?
It is not on the [roadmap](https://github.com/orgs/delta-io/projects/10/views/1)

danieljsc:
This is open since 2022, will this be tackled anytime soon or are we deprecating delta-rs?

roeap:
we are currently updating delta-kernel-rs with some requirements for deta-rs. We already have column mapping support in kernel.

It will take some more time to wire everything up in the read paths, but in principle this work is ongoing.

keen85:
fast forward 4 months: any updates here?
@roeap, @rtyler, @ion-elgreco

ion-elgreco:
@keen85 none from my side.

ohadmata:
After the last kernel update - It looks like the version in the main supports Column mapping feature.

FrankPortman:
@ohadmata is that in both Python and Rust or just Rust?

ohadmata:
@FrankPortman both python&rust. Only for main branch for now.

ion-elgreco:
@ohadmata column mapping is not supported for either. Kernel can handle it but delta-rs doesn't do anything with this information yet

keen85:
but @rtyler and @roeap are working on it, right?
[...]

PR #3954, opened 2025-11-27
feat: add column mapping support when reading tables

JonatanMartens:
# Description
This pr adds the ability to read tables with the column mapping feature enabled. 

# Related Issue(s)
Partially implements #930, but lacks the ability to write to tables with column mapping.

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/schema_adapter.rs
+++ b/crates/core/src/delta_datafusion/schema_adapter.rs
@@ -144,8 +144,41 @@ pub(crate) struct SchemaMapping {
 
 impl SchemaMapper for SchemaMapping {
     fn map_batch(&self, batch: RecordBatch) -> Result<RecordBatch> {
-        let record_batch = cast_record_batch(&batch, self.logical_schema.clone(), false, true)?;
-        Ok(record_batch)
+        // Cast individual columns to match the target schema types
+        // This handles type differences like timestamp precision mismatches
+        let batch_columns = batch.columns().to_vec();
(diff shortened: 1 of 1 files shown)


python/deltalake/table.py
```diff
@@ -922,15 +922,6 @@ def to_pyarrow_dataset(
...
-        if (
-            table_protocol.reader_features
-            and "columnMapping" in table_protocol.reader_features
-        ):
-            raise DeltaProtocolError(
-                "The table requires reader feature 'columnMapping' "
-                "but this is not supported using pyarrow Datasets."
-            )
-
         if (
```
ion-elgreco, line 862: Why is this removed? The pr itself doesn't provide support for reading column mapping with pyarrow

JonatanMartens: My primary use case is loading delta tables with datafusion (in python not rust), so must have missed it. 
Will look into adding support for reading with pyarrow

JonatanMartens: I have opened a pr to arrow https://github.com/apache/arrow/pull/48289, once that is merged we can support column mapping in datasets.

abhiaagarwal: I think it's worthwhile to get this merged in and throw an error on pyarrow usage, I have no clue what percentage of using pyarrow but I think it's likely considered legacy

JonatanMartens: @abhiaagarwal sounds good to me.
@ion-elgreco what do you think? We could print a warning when initializing the table in python (if it column mapping is enabled) and throw an error if the user tries to use the pyarrow methods

PR #3962, opened 2025-12-02
chore: removing APIs and deprecation warnings: 0.30.x here we come

rtyler:
:bomb: :boom: 

This is a pretty massive change, regrettably, but moves a tremendous amount of code from the :crab:  from using conventional/goofy `&str` implementations to using `url::Url`. This ended up surfacing a number of [Url oddities](https://brokenco.de/2025/12/03/about-url.html) that needed to be accounted for, most notably the semantic importance of a trailing `/` on our use of `Url` since we heavily rely on `Url::join`. 

There are _additional_ changes here as a result of a Rust edition upgrade to `2024` and the corresponding lints and clippy checks.

To boot, a lot of deprecated code is hereby removed in this change, thus the big negative number :hammer:

PR #3963, opened 2025-12-03
feat: support full path URIs

### LINKED ISSUE #865: Delta Log entities with absolute path not supported

corwinjoy:
# Description
Add support for absolute file paths and full path URIs to hold parquet file names in the delta logstore.

# Related Issue(s)
Closes #865

# Documentation
These kind of paths occur when creating shallow clones via Java. In addition there may be other cases where we wish to specify the files as full URIs.

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider.rs
+++ b/crates/core/src/delta_datafusion/table_provider.rs
@@ -579,6 +579,7 @@ impl<'a> DeltaScanBuilder<'a> {
         let table_partition_cols = &self.snapshot.metadata().partition_columns();
 
         let root = self.log_store.config().location.clone();
+        let path_handler = crate::logstore::DeltaPathHandler::new(root.clone());
 
         // SECURITY WARNING:
         // Registering a root object store for the file scheme allows reading ANY file on the filesystem,
@@ -596,10 +597,8 @@ impl<'a> DeltaScanBuilder<'a> {
         for action in files.iter() {
             // Normalize path using shared utility to support full path URIs
             let mut action_with_normalized_path = action.clone();
-            let (normalized_path, needs_bucket_root) = crate::logstore::normalize_add_path_for_scan(
-                &root,
-                action_with_normalized_path.path.as_str(),
-            );
+            let (normalized_path, needs_bucket_root) =
+                path_handler.normalize_add_path_for_scan(action_with_normalized_path.path.as_str());
             action_with_normalized_path.path = normalized_path;
             need_bucket_root_store |= needs_bucket_root;
 
@@ -609,7 +608,7 @@ impl<'a> DeltaScanBuilder<'a> {
                 &action_with_normalized_path,
                 table_partition_cols,
                 &schema,
-                &root,
+                &path_handler,
                 allow_unrestricted_file_access,
             );
 
@@ -1183,7 +1182,7 @@ fn partitioned_file_from_action(
     action: &Add,
     partition_columns: &[String],
     schema: &Schema,
-    root: &url::Url,
+    path_handler: &crate::logstore::DeltaPathHandler,
     allow_unrestricted_file_access: bool,
 ) -> PartitionedFile {
     let partition_values = partition_columns
@@ -1219,10 +1218,10 @@ fn partitioned_file_from_action(
             .naive_utc(),
     );
 
-    let object_meta = if root.scheme() == "file" && allow_unrestricted_file_access {
+    let object_meta = if path_handler.root().scheme() == "file" && allow_unrestricted_file_access {
         // Use a root-scoped filesystem store; provide absolute path without leading '/'
         ObjectMeta {
-            location: crate::logstore::object_store_path_for_file_root(root, action.path.as_str()),
+            location: path_handler.object_store_path_for_file_root(action.path.as_str()),
             last_modified,
             size: action.size as u64,
             e_tag: None,
@@ -1378,7 +1377,9 @@ mod tests {
 
         let part_columns = vec!["year".to_string(), "month".to_string()];
         let root = url::Url::parse("s3://test-bucket/").unwrap();
-        let file = partitioned_file_from_action(&action, &part_columns, &schema, &root, false);
+        let path_handler = crate::logstore::DeltaPathHandler::new(root);
+        let file =
+            partitioned_file_from_action(&action, &part_columns, &schema, &path_handler, false);
         let ref_file = PartitionedFile {
             object_meta: object_store::ObjectMeta {
                 location: Path::from("year=2015/month=1/part-00000-4dcb50d3-d017-450c-9df7-a7257dbd3c5d-c000.snappy.parquet".to_string()),
--- a/crates/core/src/logstore/mod.rs
+++ b/crates/core/src/logstore/mod.rs
@@ -87,6 +87,7 @@ pub use self::storage::utils::is_absolute_uri_or_path;
 pub use self::storage::utils::{
     normalize_add_path_for_scan, normalize_path_for_file_scheme, object_store_path_for_file_root,
     relativize_path_for_file_scheme, relativize_uri_to_bucket_root, strip_table_root_from_full_uri,
+    DeltaPathHandler,
 };
 pub use self::storage::{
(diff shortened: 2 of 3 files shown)


corwinjoy:
This is a draft PR for discussion.
1. I have added tests to check for absolute paths under a local file system, S3, Azure and GCP.
2. Existing tests check for escape characters in paths and escape characters in partitions when optimizing.
3. A caveat, I used AI to help code this so it is a bit verbose and I may have missed something.
4. The details get a bit messy here because we have re-encoding both by the object_store::Path and in the JSON serialization of the logstore paths.
5. Better might be to create some kind of wrapper around object_store::Path to better control operations but it's a bit tough because of the use of structures like ObjectMeta which directly use an object_store::Path. Still, if you see a good way to do this we could try it.
6. Finally, allowing full file path usage adds a potential security risk so this could use some discussion. I have put this behind a flag for filesystems, this may not be the best approach. Also, there is a similar potential risk for bucket stores.

Overall I think this gives a good picture of what would be needed for full path support.

corwinjoy:
@hntd187 @adamreeve  tagging due to related discussion on shallow clones.

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -662,24 +702,58 @@ impl<'a> DeltaScanBuilder<'a> {
...
+            }
+        } else {
+            if need_bucket_root_store {
+                // Build a root-scoped store identifier unique per (scheme, host)
+                let loc = &self.log_store.config().location;
+                let host = loc.host_str().unwrap_or("-");
+                // TODO. May be a security risk to allow absolute paths outside the bucket/account.
+                // When need_bucket_root_store is true, the code registers a bucket-root object
+                // store that could access ANY object in the bucket, not just those in the
+                // table directory. This has similar security implications to the file-root store above.
```
corwinjoy, line 726: Note. I have not added an environment variable to allow this like I did for filesystems. I'm not as sure this is necessary since object stores have more detailed bucket permissions. But, I guess it could still be a problem for delta-rs running as a superuser with malicious delta tables.

corwinjoy:
Also tagging @roeap since he has done some related work in the past:
https://github.com/delta-io/delta-rs/pull/3456

crates/core/src/kernel/models/actions.rs
```diff
@@ -1145,7 +1145,8 @@ pub(crate) mod serde_path {
         .add(b'{')
         .add(b'^')
         .add(b'}')
-        .add(b'%')
+        // NOTE: '%' is NOT included here to avoid double-encoding.
+        // Paths already contain valid percent-encoded characters from partition values.
```
corwinjoy, line 1149: Note that this may represent an existing bug in the coding.

rtyler:
@corwinjoy I believe this is now a redundant PR with the monster #3962 merged

corwinjoy:
> @corwinjoy I believe this is now a redundant PR with the monster #3962 merged

Thanks @rtyler ! I did a quick branch of main and re-ran the new path tests as posted here: https://github.com/delta-io/delta-rs/pull/3996

These full path tests fail. From what I can tell, there is still a significant logic gap.
Pieces that are missing:
1. Recognition of full path specification when producing URL output in get_file_uris().
2. Rebasing of object store as needed. The object store can only access files underneath its root directory. So, to use full paths the store needs to be rebased to a new root directory. In this full path draft PR there are new logic routines to recognize full paths and rebase the object store when needed. I believe you will need some similar logic in order to really support full paths. Hopefully some of these logic and parsing routines will be useful to you.

crates/core/Cargo.toml
```diff
@@ -92,6 +92,9 @@ criterion = "0.5"
 ctor = "0"
 datatest-stable = "0.3"
 deltalake-test = { path = "../test" }
+deltalake-aws = { path = "../aws" }
+deltalake-azure = { path = "../azure" }
+deltalake-gcp = { path = "../gcp" }
```
abhiaagarwal, line 97: Compile times are already pretty bad, can you not move these tests as integration tests in their respective tests?

Issue #3939, opened 2025-11-18
[Bug]: _internal.DeltaError: Generic DeltaTable error: Unable to map __delta_rs_path to action.


ion-elgreco:
### What happened?

No clue, tried our latest release and it throws this error on second write:

`_internal.DeltaError: Generic DeltaTable error: Unable to map __delta_rs_path to action.`

### Expected behavior

don't fail duh

### Operating System

None

### Binding

None

### Relevant logs

```shell

```

rtyler:
@ion-elgreco for a second write, can you share the `write_to_deltalake` invocation you're using?

ion-elgreco:
> @ion-elgreco for a second write, can you share the `write_to_deltalake` invocation you're using?

Dont have my device at hand for coming days but it was only write_deltalake with mode overwrite and predicate provided

dlstadther:
I have also experienced this error.
* python 3.12 (but also confirmed in 3.11 and 3.13)
* pyarrow 22.0.0
* pandas 2.3.3
* deltalake 1.2.1

For my case, I've found the error was related to a partition being overwritten by the predicate where the partition value contains url-encoded characters (spaces in my case).

The following is an example to recreate the issue.

```python
"""
Demonstrate writing to a new dataset: first write without predicate, second write with predicate, the final writes with predicate with partial upsert.
The first write succeeds because it doesn't use a predicate on an empty dataset.
The second write uses a predicate to update specific partitions.
The third write overlaps with the first write's "profile 1" / "2024-01-15" partition and raises an error.
"""

import os
import shutil

import pandas as pd
from deltalake import DeltaTable, Field, Schema, write_deltalake


def main():
    # Define the dataset path
    dataset_path = "./delta_dataset"

    # Delete the dataset if it already exists
    if os.path.exists(dataset_path):
        shutil.rmtree(dataset_path)
        print(f"Deleted existing dataset at {dataset_path}")

    partition_columns = ["profile_name", "report_date"]

    # Create a dummy pandas dataframe with partition columns
    # Include at least one profile_name with a space
    dummy_data_1 = {
        "profile_name": ["profile 1", "profile2", "profile 1", "profile3"],
        "report_date": ["2024-01-15", "2024-01-15", "2024-01-16", "2024-01-16"],
        "value": [100, 200, 150, 300],
        "metric": ["A", "B", "A", "C"],
    }
    df1 = pd.DataFrame(dummy_data_1)

    # Convert report_date to date type
[...]

PR #3976, opened 2025-12-10
fix: decode path before lookup

ion-elgreco:
# Description
Paths in the json log are double encoded, when we go over logical_file_view.path() the paths are decoded but when you access them directly in the add_actions_table they aren't so we need to decode as well here to be able to look up in the map properly.

Imho, the double encoding in the logs seems like a bug but seems to have been prevalent for a long time and this is just a tape fix to get old behavior back.

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/3939

### DIFF
--- a/crates/core/src/delta_datafusion/engine/expressions/mod.rs
+++ b/crates/core/src/delta_datafusion/engine/expressions/mod.rs
@@ -1,0 +1,79 @@
+pub(crate) use self::to_datafusion::*;
+pub(crate) use self::to_kernel::*;
+
+mod to_datafusion;
+mod to_json;
+mod to_kernel;
+
+#[cfg(test)]
+mod tests {
+    use std::ops::Not;
+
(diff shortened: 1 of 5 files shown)


corwinjoy:
In the draft PR where I examine adding support for full path URIs (https://github.com/delta-io/delta-rs/pull/3963/file)  this double encoding (once for JSON, and once for the object_store Path) is a major source of complications. If we could remove this, it would be great. If not, I would suggest a test where you use escape characters because this is where the double encoding gets complicated and tends to have problems. In addition, in the draft PR, AI suggested that there may be an existing bug with encoding the '%' character which it would be nice to test.
https://github.com/delta-io/delta-rs/pull/3963/files#r2609031682

rtyler (approved):
Test demonstrates the bug, thanks for adding @ion-elgreco

PR #3996, opened 2025-12-17
test: Add tests for full path URIS

corwinjoy:
# Description
Add the test for full path URIs to the main branch.
For discussion pursuant to https://github.com/delta-io/delta-rs/pull/3963

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,7 +111,7 @@ tpe = "tpe"
 # Usage: cargo test (automatic)
 # <https://github.com/delta-io/delta-rs/issues/1550>
 [profile.test]
-debug = "line-tables-only"
+debug = true
 
 # Benchmark profile with debug symbols for flamegraphs
 # Usage: cargo bench (automatic)
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -92,6 +92,9 @@ criterion = "0.5"
 ctor = "0"
 datatest-stable = "0.3"
 deltalake-test = { path = "../test" }
+deltalake-aws = { path = "../aws" }
+deltalake-azure = { path = "../azure" }
+deltalake-gcp = { path = "../gcp" }
 dotenvy = "0"
 fs_extra = "1.2.0"
 pretty_assertions = "1.2.1"
@@ -101,6 +104,8 @@ serial_test = "3"
 tempfile = { workspace = true }
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+
+
 
 [features]
 default = ["rustls"]
@@ -148,6 +153,10 @@ required-features = ["datafusion"]
 [[test]]
 name = "commit_info_format"
 required-features = ["datafusion"]
+
+[[test]]
+name = "full_path"
+required-features = ["datafusion", "cloud"]
 
 [[test]]
 name = "integration"
--- a/crates/core/tests/full_path.rs
+++ b/crates/core/tests/full_path.rs
@@ -1,0 +1,844 @@
+// Verifies that tables with fully-qualified file paths behave as expected
+// and that reading them yields the same data as the corresponding table
+// with relative file paths.
+use serde_json::Value;
+use std::collections::HashMap;
+use std::fs;
+use std::path::{Path, PathBuf};
+use url::Url;
+
+// Small utility to build original/cloned URIs from a base URI and two prefixes
+fn build_prefixed_uris(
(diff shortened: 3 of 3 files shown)


corwinjoy:
This draft PR is to help with the discussion at: https://github.com/delta-io/delta-rs/pull/3963#issuecomment-3667535808

Cargo.toml
```diff
@@ -111,7 +111,7 @@ tpe = "tpe"
 # Usage: cargo test (automatic)
 # <https://github.com/delta-io/delta-rs/issues/1550>
 [profile.test]
-debug = "line-tables-only"
+debug = true
```
corwinjoy, line 114: In RustRover "line-tables-only" makes it so you can't use variable debugging, which makes it hard to debug any failing tests. Turned this on so that I can see what is happening in the tests.

rtyler: In the past I had set this to line-tables-only because the tests were taking up massive amounts of disk space in Github Action Runners. Since @fvaleye added the `ci` profile this should probably be fine to return :thinking:

crates/core/Cargo.toml
```diff
@@ -92,6 +92,9 @@ criterion = "0.5"
 ctor = "0"
 datatest-stable = "0.3"
 deltalake-test = { path = "../test" }
+deltalake-aws = { path = "../aws" }
+deltalake-azure = { path = "../azure" }
+deltalake-gcp = { path = "../gcp" }
```
rtyler, line 97: I think the test would be better suited to add into the ``crates/deltalake` directory, since that's where these dependencies already are present.

I'm a little surprised the tests compile because of the circular dependency here, but I guess cargo just figures it out :shrug:

crates/core/tests/full_path.rs
```diff
@@ -0,0 +1,850 @@
...
+    fn drop(&mut self) {
+        use std::process::Command;
+        let _ = Command::new(self.program).args(&self.args).status();
+    }
+}
+
+// Test to use absolute file paths
+#[tokio::test]
+async fn compare_table_with_full_paths() {
+    // std::env::set_var("DELTA_RS_ALLOW_UNRESTRICTED_FILE_ACCESS", "1");
```
rtyler, line 47: When I run these tests locally this test fails, is that to be expected with this pull request? I.e. is this supposed to be a demonstration of a bug?

crates/core/tests/full_path.rs
```diff
@@ -0,0 +1,850 @@
...
+
+    // Compare the cloned table against expected URIs and data from the original table.
+    assert_table_uris_and_data(&cloned_dir, &tmpdir.path(), &expected_abs).await;
+}
+
+// Test to use absolute file paths pointing to an original source table directory
+// Similar to what a shallow clone with full paths would produce.
+#[tokio::test]
+async fn compare_table_with_full_paths_to_original_table() {
+    // std::env::set_var("DELTA_RS_ALLOW_UNRESTRICTED_FILE_ACCESS", "1");
```
rtyler, line 65: This test also fails for me

corwinjoy:
@rtyler 
1. These tests check for full path URI suuport. They fail against the main branch, indicating full paths are not supported yet. (These tests pass for https://github.com/delta-io/delta-rs/pull/3963). So, yes, you should see these tests failing.

To be concrete, suppose we have a file entry in the logstore with a full path specification of the form `/a/b/c.parquet` and the test table directory is `/tmp/delta_test`. Then, in main
a. `table.get_file_uris()` returns `/tmp/delta_test/a/b/c.parquet` which is incorrect.
b.. When we attempt to read the table we get an error like `ParquetError(External(NotFound { path: "/tmp/delta_tests/a/b/c/parquet", source: Os { code: 2, kind: NotFound, message: "No such file or directory" } })) ` 
This is partly due to the incorrect path. But also, the object store will only open files that are a subdirectory of the root directory for the object store.  `/a/b/c.parquet`  is not in a subdirectory of  `/tmp/delta_test`.

So, to support full path URIs, I believe that delta-rs should:
1. Return the correct path when  `table.get_file_uris()`  is called.
2. Should be able to read data from a parquet file when an absolute (or full) path is used.
3. Be able to handle escape characters as needed (existing tests).
4. Be able to write to a parquet file with an absolute path if needed (no test for this here).

These tests check for 1 & 2 and also absolute file specifications like:
`file:///a/b/c.parquet`
`s3:///a/b/c.parquet`
`gs:///a/b/c.parquet`

The draft PR, https://github.com/delta-io/delta-rs/pull/3963 , has additional logic to handle these cases. Hopefully some of these pieces will be helpful to you.

Issue #3980, opened 2025-12-13
TableProvider via KernelScans


roeap:
### Is your feature request related to a problem?

Currently our datafusion table scans are build without accounting for several transformations that need to be applied on top of the data.

Additionally we can apply some optimisations for specific queries that can be evaluated on metadata only - e.g. some `COUNT(*)` cases.

### Describe the solution you'd like

A new `TableProvider` implementation that is deigned with the kernel APIs in mind.

### Describe alternatives you've considered

migrating the existing providers is quite cumbersome and error prone as they do many things that are now covered in kernel and are largely based on the older internal APIs.

### Priority

High - Important for my use case

### Contribution

- [x] I'm willing to submit a pull request for this feature
- [ ] I can help with testing this feature
- [ ] I can help with documentation for this feature

PR #3998, opened 2025-12-19
feat: kernel expression conversion

roeap:
# Description

This PR add basic conversion of expressions to the internal engine implementation. We still have some more complex cases (i.e. opaque expressions) to handle but since its isolated and not in any code path yet, I hope we can get this merged to get started.

# Related Issue(s)

part-of: #3980
part-of: #3999

### DIFF
--- a/crates/core/src/delta_datafusion/engine/expressions/to_kernel.rs
+++ b/crates/core/src/delta_datafusion/engine/expressions/to_kernel.rs
@@ -124,6 +124,31 @@ pub(crate) fn to_delta_expression(expr: &Expr) -> DFResult<Expression> {
         Expr::Not(expr) => Ok(Expression::Predicate(Box::new(Predicate::Not(Box::new(
             Predicate::BooleanExpression(to_delta_expression(expr.as_ref())?),
         ))))),
+        Expr::Between(between) => {
+            let expr = to_delta_expression(&between.expr)?;
+            let (high_op, low_op) = if between.negated {
+                (BinaryPredicateOp::LessThan, BinaryPredicateOp::GreaterThan)
+            } else {
+                (BinaryPredicateOp::GreaterThan, BinaryPredicateOp::LessThan)
+            };
+            Ok(Expression::Predicate(Box::new(Predicate::Junction(
+                JunctionPredicate {
+                    op: JunctionPredicateOp::And,
+                    preds: vec![
+                        Predicate::Binary(BinaryPredicate {
+                            left: Box::new(expr.clone()),
+                            op: low_op,
+                            right: Box::new(to_delta_expression(&between.low)?),
+                        }),
+                        Predicate::Binary(BinaryPredicate {
+                            left: Box::new(expr),
+                            op: high_op,
+                            right: Box::new(to_delta_expression(&between.high)?),
+                        }),
+                    ],
+                },
+            ))))
+        }
         _ => Err(DataFusionError::NotImplemented(format!(
             "Unsupported expression: {:?}",
             expr
--- a/crates/core/src/delta_datafusion/engine/expressions/mod.rs
+++ b/crates/core/src/delta_datafusion/engine/expressions/mod.rs
@@ -1,5 +1,5 @@
-pub use self::to_datafusion::*;
-pub use self::to_kernel::*;
+pub(crate) use self::to_datafusion::*;
+pub(crate) use self::to_kernel::*;
 
 mod to_datafusion;
 mod to_json;


crates/core/src/delta_datafusion/engine/expressions/to_datafusion.rs
ion-elgreco, line 1: Extreme small nit haha, but maybe call it to_datafusion. For me df reads as dataframe

crates/core/src/delta_datafusion/engine/expressions/to_datafusion.rs
```diff
@@ -0,0 +1,874 @@
...
+                .zip(values.into_iter())
+                .fold(ScalarStructBuilder::new(), |builder, (field, value)| {
+                    builder.with_scalar(field, value)
+                })
+                .build()?
+        }
+        Scalar::Array(_) => {
+            return Err(DataFusionError::NotImplemented(
+                "Array scalar values not implemented".into(),
+            ));
```
ion-elgreco, line 92: Any reason, you are holding this off?

crates/core/src/delta_datafusion/engine/expressions/to_df.rs
```diff
@@ -0,0 +1,874 @@
...
+    match expr {
+        Expression::Literal(scalar) => scalar_to_df(scalar).map(lit),
+        Expression::Column(name) => {
+            let mut name_iter = name.iter();
+            let base_name = name_iter.next().ok_or_else(|| {
+                DataFusionError::Internal("Expected at least one column name".into())
+            })?;
+            Ok(name_iter.fold(col(base_name), |acc, n| acc.field(n)))
+        }
+        Expression::Predicate(expr) => predicate_to_df(expr, output_type),
```
ion-elgreco, line 36:
```suggestion
        Expression::Predicate(expr) => to_datafusion_predicate(expr, output_type),
```
Nitpicking at best here

crates/core/src/delta_datafusion/engine/expressions/to_datafusion.rs
roeap: just that current tests did not need it :).

but this will get an overhaul very soon ...

crates/core/src/delta_datafusion/engine/expressions/to_datafusion.rs
roeap: actually a good point for a non-confusing API :). updated.

roeap:
@ion-elgreco - addressed your feedback. specifically we no longer use df in place of datafusion in public APIs. Its particularly confusion since there are DataFrames in datafusion :D

Also added a custom udf to implement `to_json`, still more work, but worth landing I believe. 🤞

crates/core/src/delta_datafusion/engine/expressions/to_kernel.rs
```diff
@@ -0,0 +1,551 @@
...
+
+pub fn to_predicate(expr: &Expr) -> DFResult<Predicate> {
+    match to_delta_expression(expr)? {
+        Expression::Predicate(pred) => Ok(pred.as_ref().clone()),
+        expr => Ok(Predicate::BooleanExpression(expr)),
+    }
+}
+
+/// Convert a DataFusion expression to a Delta expression.
+pub fn to_delta_expression(expr: &Expr) -> DFResult<Expression> {
```
rtyler: Of all the newly added public APIs here, this is the only one that's documented. :sob: 

Please please please add some rustdoc for what all this new public API surface is supposed to be used for. While we haven't explicitly discussed it, I think we should adopt a "no new public APIs without X" type rule.

At the low end X could just mean "add some rustdoc" at the high end would be "rustdoc and at least one doctest" but the latter would be a pretty big pain in the ass given our current bad habits :laughing:

roeap: @rtyler - excellent point...

Added some doc strings to the exposed functions, but also made them `pub(crate)` since I am now unsure that we actually want to expose them, so better safe that sorry.

roeap: while this also increases our warning count, we will soon recover as we start using these.

ion-elgreco:
Slight naming suggestion: `to_datafusion_X` -> `to_datafusion_scalar`

roeap:
@ion-elgreco - good point, there is just something about consistency in public APIs :).

... updated.

ion-elgreco:
> @ion-elgreco - good point, there is just something about consistency in public APIs :).
> 
> ... updated.

I think it applies for all functions, I guess the scalar one is was just an example

roeap:
> I think it applies for all functions, I guess the scalar one is was just an example

I just did it for the accessible ones.

roeap:
> we're back to previous level after the next PR :).

PR #4014, opened 2025-12-27
fix: null handling when extracting scalars

roeap:
# Description

Currently we do not validate values we access when extracting a value from an array. This PR adds a call to `is_valid` before we create a new scalar, creating a NULL scalar otherwise.

I also asked a friendly AI helper to write some testing code for the various methods in the scalars module.


crates/core/src/kernel/scalars.rs
```diff
@@ -344,6 +346,14 @@ impl ScalarExt for Scalar {
     }
 }
 
+fn checked(arr: &dyn Array, idx: usize, val: Scalar) -> Scalar {
```
rtyler, line 349: @roeap  this function is a good candidate for [inlining](https://nnethercote.github.io/perf-book/inlining.html) for future reference.

There may be a couple others already in the codebase, but good to know about in the future :wink:

Issue #3825, opened 2025-10-07
use generic session trait tor tracking Datafusion sessions.


roeap:
# Description

Datafusion recommends using the `Session` trait whenever possible. We should do so whenever possible.

PR #4024, opened 2025-12-30
feat: use datafusion logical plans more, materialize as late as possible

abhiaagarwal:
# Description

This mostly started as an exercise to see how much direct datafusion dependencies we can remove throughout the codebase to improve compile times by using only the subcrates. The only leftovers were `SessionState` and `DataFrame`. As I was looking through our usage of `DataFrame`, I realized we use them as intermediary objects purely for the syntax with logical plans, so I decided to change the logic to directly manipulate the logical plans and avoid any dataframe usage. This means we can drastically reduce our usage of `SessionState` since `DataFrame` needs a `SessionState`, but we can rely on `&dyn Session` to convert it into a physical plan.

# Related Issue(s)

Ties into #2006, #3825, and @roeap's work in general

# Documentation

### DIFF
--- a/crates/core/src/operations/write/mod.rs
+++ b/crates/core/src/operations/write/mod.rs
@@ -654,8 +654,8 @@ impl std::future::IntoFuture for WriteBuilder {
                                     source_logical_plan =
                                         LogicalPlanBuilder::from(source_logical_plan)
                                             .with_column(
-                                                "insert",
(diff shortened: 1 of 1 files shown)


abhiaagarwal:
I'm gonna close this, as @roeap has carried on my legacy 🫡

PR #4030, opened 2026-01-01
feat(core): add BlindDeltaTable for stats-free append writes

### LINKED ISSUE #3979: delta.appendOnly tables should not bother parsing stats when opened for write

When loading extremely large tables I believe we're doing a "full replay" so to speak, with all the bells and whistles on the log. This can be unnecessarily slow when all you want is an append.

When `delta.appendOnly` is set to `true` on the table properties and the user has indicated that the table is being opened for a **write operation** the parsing of stats should be omitted.

Thanks to some investigation work done by @libc with a massive proprietary table we have observed that disabling stats results in a **85%** reduction in time-to-open metrics on append only write operations.

I think this functionality might be additionally beneficial for work that @abhi-airspace-intelligence has been doing as well.

_No response_

### Priority

None

### Additional context

_No response_

### Contribution

seokjin0414:
# Description
Add `BlindDeltaTable` struct optimized for blind append operations that skips file statistics parsing during table load.                                          
                                                                                                                                                                    
  - Add `BlindDeltaTable` with `try_new()`, `commit()` methods                                                                                                      
  - Add `RecordBatchWriter::for_blind_appends()` for writer integration                                                                                             
  - Use lazy `Snapshot` with `require_files=false` to skip stats parsing                                                                                            
  - Use Kernel Transaction API directly (bypasses `CommitBuilder`)                                                                                                  
  - Provides compile-time safety by not exposing `files()`/`log_data()` methods                                                                                     
  - 14 unit tests                                                                                                                                                   
                                                                                                                                                                    
This addresses the performance issue where large tables with many files experience slow load times due to unnecessary stats parsing when only append writes are needed.

# Related Issue(s)
  - closes #3979

# Documentation
  - Usage example in module docs: `crates/core/src/table/blind.rs`


abhiaagarwal:
I'm skeptical this actually works. Anecdotally, I've tested in a development environment with `require_files=False` on append-only workloads, and it still fails.

hntd187:
This is missing a lot, and I believe we already support appends. What does this provide other than skipping conflict checking and any table configuration that might affect appends?

rtyler:
@seokjin0414 our timezones not be too helpful for cooperation but next Tuesday at 7am PST (15:00 UTC), @roeap and I have a standing "Delta Lake Live" meeting where we pair on stuff like this, if you stop by the [#delta-rs](https://delta-users.slack.com/archives/C013LCAEB98) channel maybe we could pair on this implementation?

seokjin0414:
@rtyler Thank you for the great suggestion! The time works for me.

rtyler:
There is some similar work happening that we can take inspiration from in #4040 on how to handle the kernel commit APIs for pushing data through.

From our live session there are a couple things we discussed:

* Dropping `EagerSnapshot` in favor of `Snapshot` to ensure maximum laziness
* @seokjin0414 is going to explore swapping out `flush_and_commit` in the `BlindDeltaTable` with the Kernel `Commit` API . We were looking at [this snippet of code](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/examples/write-table/src/main.rs#L90-L113) and discussing using our existing parquet writer code instead of the `write_parquet` function but otherwise using everything else in the write-table example.
* @rtyler needs to explore what happens if we tie the DynamoDBLogStore logic into the kernel Commit API in a manner similar to how ExtendedStats are being done in #4040 to provide the necessary custom behavior there.
* @roeap is determined to own LakeFS and all its necessary code to change out `CommitBuilder` in favor of the commit API. :smiling_imp:

seokjin0414:
@rtyler
Updated based on our discussion:                                                                                                                                                                                                                                                                                 
  - Changed EagerSnapshot → Snapshot
  - Replaced CommitBuilder with Kernel Transaction API                                                                                                                                                                                                                                                         

One question: Used a fixed constant (`BLIND_COMMIT_MAX_RETRIES = 15`) for retry limit. Also considered adding `with_max_retries()` builder pattern. Which approach is preferred?

rtyler (approved):
Let's give it a try!

ethan-tyler (approved):
I don't see anything crazy here - approving and if anything comes up we can patch it

PR #4036, opened 2026-01-02
feat: kernel based table scans

roeap:
We introduce a new TableProvider implementation that leverages the kernel's capabilities for supporting advanced table features. A key design goal is to fully utilize kernels and datafusions capabilities and only implement what is strictly necessary in our codebase.

Getting to a point where we can defer materializing the log to as late as possible and only for the parts that are strictly necessary is a key goal for performance and efficiency. As such, the table provider can work with both Snapshot and EagerSnapshot implementations.

The table provider itself does not directly expose stats, but stats are available on the generated execution plan, so this should not limit datafusions ability to optimize plans.

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/mod.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/mod.rs
@@ -38,7 +38,8 @@ use dashmap::DashMap;
 use datafusion::catalog::memory::DataSourceExec;
 use datafusion::common::stats::Precision;
 use datafusion::common::{
-    ColumnStatistics, DataFusionError, HashMap, HashSet, Result, Statistics, plan_err,
+    ColumnStatistics, DataFusionError, HashMap, HashSet, Result, Statistics, plan_datafusion_err,
+    plan_err,
 };
 use datafusion::datasource::TableType;
 use datafusion::datasource::listing::PartitionedFile;
@@ -90,6 +91,7 @@ mod scan_meta;
 pub type ScanMetadataStream =
     Pin<Box<dyn Stream<Item = Result<ScanMetadata, DeltaTableError>> + Send>>;
 
+/// Default column name for the file id column we add to files read from disk.
 const FILE_ID_COLUMN_DEFAULT: &str = "__delta_rs_file_id__";
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
@@ -305,18 +307,6 @@ impl TableProvider for DeltaScan {
             self.snapshot.table_configuration(),
         ))
     }
-
-    /// Insert the data into the delta table
-    /// Insert operation is only supported for Append and Overwrite
-    /// Return the execution plan
-    async fn insert_into(
-        &self,
-        _state: &dyn Session,
-        input: Arc<dyn ExecutionPlan>,
-        insert_op: InsertOp,
-    ) -> Result<Arc<dyn ExecutionPlan>> {
-        todo!("implement insert into")
-    }
 }
 
 /// Internal representation of a scan plan based on Kernel's Scan abstraction
@@ -374,9 +364,9 @@ impl KernelScanPlan {
                 .read_schema()
                 .column_with_name(col)
                 .ok_or_else(|| {
-                    DataFusionError::Plan(format!(
+                    plan_datafusion_err!(
                         "Column '{col}' referenced in filter but not found in schema"
-                    ))
+                    )
                 })?;
             projection.push(idx);
         }
@@ -427,11 +417,11 @@ impl KernelScanPlan {
     ///
     /// It may still be impossible to perform a metadata-only scan if the
     /// file statistics are not sufficient to satisfy the query.
-    pub fn is_metadata_only(&self) -> bool {
+    fn is_metadata_only(&self) -> bool {
         self.scan.physical_schema().fields().len() == 0
     }
 
-    pub fn table_configuration(&self) -> &TableConfiguration {
+    fn table_configuration(&self) -> &TableConfiguration {
         self.scan.snapshot().table_configuration()
     }
 }
@@ -692,14 +682,6 @@ mod tests {
                 .clone_inner()
                 .sum_by_name("bytes_scanned");
             self.total_bytes_scanned = pq_metrics.map(|v| v.as_usize());
-
-            // if let Some(parquet_source) = scan_config
-            //     .file_source
-            //     .as_any()
-            //     .downcast_ref::<ParquetSource>()
-            // {
-            //     parquet_source
-            // }
 
             Ok(true)
         }
--- a/crates/core/src/delta_datafusion/table_provider/next/replay.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/replay.rs
@@ -134,7 +134,7 @@ impl<S> ScanFileStream<S> {
             dv_stream: ReceiverStreamBuilder::<(Url, Option<Vec<bool>>)>::new(100),
             engine,
             table_root: scan.table_root().clone(),
-            kernel_scan: scan.inner.clone(),
+            kernel_scan: scan.inner().clone(),
             stream,
         }
     }
--- a/crates/core/src/delta_datafusion/table_provider/next/scan.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan.rs
@@ -31,6 +31,10 @@ use crate::delta_datafusion::table_provider::next::KernelScanPlan;
 use crate::kernel::ARROW_HANDLER;
 use crate::kernel::arrow::engine_ext::ExpressionEvaluatorExt;
 
+/// Execution plan for scanning a Delta table.
+///
+/// This plan wraps another execution plan that reads the raw data from data files,
+/// and applies per-file transformations and selection vectors to yield the logical data.
 #[derive(Clone, Debug)]
 pub struct DeltaScanExec {
     scan_plan: Arc<KernelScanPlan>,
@@ -38,7 +42,7 @@ pub struct DeltaScanExec {
     input: Arc<dyn ExecutionPlan>,
     /// Transforms to be applied to data eminating from individual files
     transforms: Arc<HashMap<String, ExpressionRef>>,
-    /// Deletion vectors for the table
+    /// Selection vectors to be applied to data read from individual files
     selection_vectors: Arc<DashMap<String, Vec<bool>>>,
     /// Execution metrics
     metrics: ExecutionPlanMetricsSet,
--- a/crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
@@ -33,9 +33,6 @@ use crate::delta_datafusion::table_provider::next::KernelScanPlan;
 use crate::kernel::ARROW_HANDLER;
 use crate::kernel::arrow::engine_ext::ExpressionEvaluatorExt;
 
-// TODO:
-// - implement limit/fetch pushdown
-
 /// Execution plan for scanning a Delta table based only on file metadata.
 ///
 /// For certain queries (e.g. `COUNT(*)`) we may be able to satisfy the query
--- a/crates/core/src/kernel/snapshot/scan.rs
+++ b/crates/core/src/kernel/snapshot/scan.rs
@@ -70,7 +70,7 @@ impl ScanBuilder {
 
 #[derive(Debug)]
 pub struct Scan {
-    pub(crate) inner: Arc<KernelScan>,
+    inner: Arc<KernelScan>,
 }
 
 impl From<KernelScan> for Scan {
@@ -88,6 +88,11 @@ impl From<Arc<KernelScan>> for Scan {
 }
 
 impl Scan {
+    /// Get a shared reference to the inner [`KernelScan`].
+    pub(crate) fn inner(&self) -> &Arc<KernelScan> {
+        &self.inner
+    }
+
     /// The table's root URL. Any relative paths returned from `scan_data` (or in a callback from
     /// [`ScanMetadata::visit_scan_files`]) must be resolved against this root to get the actual path to
     /// the file.


crates/core/src/delta_datafusion/table_provider/next/replay.rs
```diff
@@ -0,0 +1,431 @@
...
+            size: inner.size,
+            transform: inner.transform,
+            stats,
+            partitions,
+        }
+    }
+}
+
+/// Metadata to read a data file from object storage.
+struct ScanFileContextInner {
```
hntd187, line 369: Maybe a better name for this is `ScanContextFile`? "Inner" is a bit confusing here I think.

crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
```diff
@@ -0,0 +1,531 @@
...
+        } else {
+            batch
+        };
+
+        // TODO: all casting should be done in the expression evaluator
+        let result = cast_record_batch(&batch, self.scan_plan.result_schema.clone(), true, true)?;
+
+        if let Some(file_id_field) = &self.file_id_field {
+            let file_id_array =
+                StringArray::from_iter_values(std::iter::repeat_n(file_id, result.num_rows()));
```
hntd187, line 290: I remember there being a way for arrow to generate arrays like this that were all the same singular value. Maybe I'm misremembering but wouldn't that be more desirable than creating this from an iterator?

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -691,6 +727,133 @@ impl<'a> DeltaScanBuilder<'a> {
...
+        let mut builder = TableProviderBuilder::new();
+        if let Ok(state) = self.snapshot() {
+            builder = builder.with_eager_snapshot(state.snapshot().clone());
+        } else {
+            builder = builder.with_log_store(self.log_store());
+        }
+        builder
+    }
+
+    pub fn update_datafsuion_session(&self, session: &dyn Session) -> DeltaResult<()> {
```
hntd187:
```suggestion
    pub fn update_datafusion_session(&self, session: &dyn Session) -> DeltaResult<()> {
```

crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
roeap: it gets dictionary encoded later, but creating dict arrays by hand is a major pain.
i think Datum does not help us here unfortunately. This si covering the very rare case where we wnat a file id coulumn in a metadata only scan - so we should aready be relative effcient here and can take the hit?

hntd187: Fair enough, thanks for clarifying

hntd187 (approved):
I'm good thanks for the updates :D

crates/core/src/delta_datafusion/table_provider/next/replay.rs
roeap: was also not too happy with the name. but since its not pub in any way was not going deeper on it :).

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+pub type ScanMetadataStream =
+    Pin<Box<dyn Stream<Item = Result<ScanMetadata, DeltaTableError>> + Send>>;
+
+const FILE_ID_COLUMN_DEFAULT: &str = "__delta_rs_file_id__";
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub(crate) enum SnapshotWrapper {
+    Snapshot(Arc<Snapshot>),
+    EagerSnapshot(Arc<EagerSnapshot>),
+}
```
rtyler, line 101: :face_with_spiral_eyes:

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+
+mod predicate;
+mod replay;
+mod scan;
+mod scan_meta;
+
+pub type ScanMetadataStream =
+    Pin<Box<dyn Stream<Item = Result<ScanMetadata, DeltaTableError>> + Send>>;
+
+const FILE_ID_COLUMN_DEFAULT: &str = "__delta_rs_file_id__";
```
rtyler, line 95: For what purpose is the file_id? Rustdoc would help me here :wink:

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+    ) -> Result<Vec<TableProviderFilterPushDown>> {
+        Ok(can_pushdown_filters(
+            filter,
+            self.snapshot.table_configuration(),
+        ))
+    }
+
+    /// Insert the data into the delta table
+    /// Insert operation is only supported for Append and Overwrite
+    /// Return the execution plan
```
rtyler:
```suggestion
    /// This has yet to be implemented
```

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+                .column_with_name(col)
+                .ok_or_else(|| {
+                    DataFusionError::Plan(format!(
+                        "Column '{col}' referenced in filter but not found in schema"
+                    ))
+                })?;
+            projection.push(idx);
+        }
+
+        drop(columns_referenced);
```
rtyler, line 374: There are a couple of `drop` statements here, I don't understand why the explicit drops are necessary. Can you add more comments to this function? There's a lot of logic here for a `new` function which I am trying to wrap my head around

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+
+    /// Denotes if the scan can be resolved using only file metadata
+    ///
+    /// It may still be impossible to perform a metadata-only scan if the
+    /// file statistics are not sufficient to satisfy the query.
+    pub fn is_metadata_only(&self) -> bool {
+        self.scan.physical_schema().fields().len() == 0
+    }
+
+    pub fn table_configuration(&self) -> &TableConfiguration {
```
rtyler: Please add Rustdoc to public functions :smiling_imp:

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+
+    let metrics = ExecutionPlanMetricsSet::new();
+    MetricBuilder::new(&metrics)
+        .global_counter("count_files_scanned")
+        .add(stream.metrics.num_scanned);
+
+    Ok((files, transforms, dvs, metrics))
+}
+
+type FilesByStore = (ObjectStoreUrl, Vec<(PartitionedFile, Option<Vec<bool>>)>);
```
rtyler, line 580: This type is only used down below in `get_read_plan`, is it just for ergonomics of that function?

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+    let mut plans = Vec::new();
+
+    for (store_url, files) in files_by_store.into_iter() {
+        // state.ensure_object_store(store_url.as_ref()).await?;
+
+        let store = state.runtime_env().object_store(&store_url)?;
+        let _reader_factory = source
+            .parquet_file_reader_factory()
+            .cloned()
+            .unwrap_or_else(|| Arc::new(DefaultParquetFileReaderFactory::new(store)));
```
rtyler, line 601: :confused: is this supposed to be used? Or is this work partway done?

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+            self.total_bytes_scanned = pq_metrics.map(|v| v.as_usize());
+
+            // if let Some(parquet_source) = scan_config
+            //     .file_source
+            //     .as_any()
+            //     .downcast_ref::<ParquetSource>()
+            // {
+            //     parquet_source
+            // }
+
```
rtyler:
```suggestion
```

Can this be nuked?

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -0,0 +1,775 @@
...
+        let provider = DeltaScan::builder().with_snapshot(snapshot).await?;
+
+        let session = Arc::new(create_session().into_inner());
+        session.register_table("delta_table", provider).unwrap();
+
+        let df = session.sql("SELECT * FROM delta_table").await.unwrap();
+        let batches = df.collect().await?;
+
+        let expected = vec![
+            "+----+", "| id |", "+----+", "| 5  |", "| 7  |", "| 9  |", "+----+",
```
rtyler, line 719: :laughing:

crates/core/src/delta_datafusion/table_provider/next/replay.rs
```diff
@@ -0,0 +1,431 @@
...
+        #[pin]
+        stream: S,
+    }
+}
+
+impl<S> ScanFileStream<S> {
+    pub(crate) fn new(engine: Arc<dyn Engine>, scan: &Arc<Scan>, stream: S) -> Self {
+        Self {
+            metrics: ReplayStats::new(),
+            dv_stream: ReceiverStreamBuilder::<(Url, Option<Vec<bool>>)>::new(100),
```
rtyler, line 134: This hard-coded magic number that we're now using in a few places.   Perhaps `ReceiverStreamBuilder` should have a `Default` impl which sets the capacity to your desired 100.

crates/core/src/delta_datafusion/table_provider/next/replay.rs
```diff
@@ -0,0 +1,431 @@
...
+}
+
+fn parse_path(url: &Url, path: &str) -> DeltaResult<Url, DataFusionError> {
+    Ok(match Url::parse(path) {
+        Ok(url) => url,
+        Err(_) => url
+            .join(path)
+            .map_err(|e| DataFusionError::External(Box::new(e)))?,
+    })
+}
```
rtyler, line 414: This function really concerns me, what's the purpose of this? I did not realize that [ScanFile](https://docs.rs/delta_kernel/latest/delta_kernel/scan/state/struct.ScanFile.html) uses a `path` which is `String`

Doesn't this kind of explode the progress we just made at Url'ing everything in core? :cry:

crates/core/src/delta_datafusion/table_provider/next/scan.rs
```diff
@@ -0,0 +1,830 @@
...
+use delta_kernel::{EvaluationHandler, ExpressionRef};
+use futures::stream::{Stream, StreamExt};
+
+use crate::cast_record_batch;
+use crate::delta_datafusion::table_provider::next::KernelScanPlan;
+use crate::kernel::ARROW_HANDLER;
+use crate::kernel::arrow::engine_ext::ExpressionEvaluatorExt;
+
+#[derive(Clone, Debug)]
+pub struct DeltaScanExec {
```
rtyler, line 39: Please add some Rustdoc for this new public struct :sunglasses:

crates/core/src/delta_datafusion/table_provider/next/scan.rs
```diff
@@ -0,0 +1,830 @@
...
+    scan_plan: Arc<KernelScanPlan>,
+    /// Execution plan yielding the raw data read from data files.
+    input: Arc<dyn ExecutionPlan>,
+    /// Transforms to be applied to data eminating from individual files
+    transforms: Arc<HashMap<String, ExpressionRef>>,
+    /// Deletion vectors for the table
+    selection_vectors: Arc<DashMap<String, Vec<bool>>>,
```
rtyler: Are these deletion vectors or selection vectors? :laughing: I think the rustdoc is just typoed

crates/core/src/delta_datafusion/table_provider/next/scan.rs
```diff
@@ -0,0 +1,830 @@
...
+            | DisplayFormatType::Verbose
+            | DisplayFormatType::TreeRender => {
+                write!(f, "DeltaScanExec: file_id_column={}", self.file_id_column)
+            }
+        }
+    }
+}
+
+impl DeltaScanExec {
+    pub(super) fn new(
```
rtyler, line 73: I have typically found `pub(super)` leads to a pretty gnarly web of inter-module dependencies. IMHO this either be crate visible or not at all. But also, how does the pub on `DeltaScanExec`, interact here?

crates/core/src/delta_datafusion/table_provider/next/scan.rs
```diff
@@ -0,0 +1,830 @@
...
+
+        if config.is_feature_enabled(&TableFeature::ColumnMapping) {
+            let get_index = |name| {
+                if let Some(logical) = self.scan_plan.scan.logical_schema().field(name) {
+                    let physical = logical.make_physical(config.column_mapping_mode());
+                    self.input.schema().index_of(physical.name()).ok()
+                } else {
+                    None
+                }
+            };
```
rtyler, line 129: This is a pattern I see a lot inside of delta-kernel and I think it's really unmaintainable. Inline closures like this are very untestable. 

There is a lot overlap between this branch and the else statement, as far as I can tell the difference is one side grabs the index from `physical_schema` and the other is `logical_schema`?

crates/core/src/delta_datafusion/table_provider/next/scan.rs
```diff
@@ -0,0 +1,830 @@
...
+    }
+
+    fn children(&self) -> Vec<&Arc<dyn ExecutionPlan>> {
+        vec![&self.input]
+    }
+
+    // TODO: setting this will fail certain tests, but why
+    // fn maintains_input_order(&self) -> Vec<bool> {
+    //     vec![true]
+    // }
```
rtyler, line 182: :laughing: 

BUT WHYYYYYYYYY

crates/core/src/delta_datafusion/table_provider/next/scan.rs
```diff
@@ -0,0 +1,830 @@
...
+
+    fn gather_filters_for_pushdown(
+        &self,
+        _phase: FilterPushdownPhase,
+        parent_filters: Vec<Arc<dyn PhysicalExpr>>,
+        _config: &ConfigOptions,
+    ) -> Result<FilterDescription> {
+        // TODO(roeap): this will likely not do much for column mapping enabled tables
+        // since the default methods determines this based on existence of columns in child
+        // schemas. In the case of column mapping all columns will have a different name.
```
rtyler, line 276: Does this mean that the filters returned may be _wrong_ for column mapped tables, leading to lower performance, or is it kind of a :shrug: situation?

If it's a potentially slower scan, I recommend adding a `warn!()` here to say something like "Yo we can't push filters down with column mapped tables, sorry this might be less efficient than it could be"

crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
```diff
@@ -0,0 +1,531 @@
...
+use futures::stream::Stream;
+use itertools::Itertools;
+
+use crate::cast_record_batch;
+use crate::delta_datafusion::table_provider::next::KernelScanPlan;
+use crate::kernel::ARROW_HANDLER;
+use crate::kernel::arrow::engine_ext::ExpressionEvaluatorExt;
+
+// TODO:
+// - implement limit/fetch pushdown
```
rtyler: :thinking: what's this doing floating around up here?

crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
```diff
@@ -0,0 +1,531 @@
...
+                        .map(|f| f.name().to_owned())
+                        .unwrap_or("<none>".to_string())
+                )
+            }
+        }
+    }
+}
+
+impl DeltaScanMetaExec {
+    pub(super) fn new(
```
rtyler, line 78: Same concern as before about messy module interdependency with `pub(super)`

crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    fn gather_filters_for_pushdown(
+        &self,
+        _phase: FilterPushdownPhase,
+        parent_filters: Vec<Arc<dyn PhysicalExpr>>,
+        _config: &ConfigOptions,
+    ) -> Result<FilterDescription> {
+        // TODO(roeap): this will likely not do much for column mapping enabled tables
+        // since the default methods determines this based on existence of columns in child
+        // schemas. In the case of column mapping all columns will have a different name.
```
rtyler, line 215: :detective:

crates/core/src/kernel/snapshot/scan.rs
```diff
@@ -70,7 +70,7 @@ impl ScanBuilder {
 
 #[derive(Debug)]
 pub struct Scan {
-    inner: Arc<KernelScan>,
+    pub(crate) inner: Arc<KernelScan>,
```
rtyler: Since this member is already in an `Arc<T>` my preference would be that we avoid making struct fields visible and instead just create an accessor function.   That has made some of our refactoring easier further down the line with `LogStore` as an example

crates/core/src/delta_datafusion/table_provider/next/replay.rs
```diff
@@ -0,0 +1,431 @@
...
+                        max_value,
+                        min_value,
+                        sum_value: Precision::Absent,
+                        distinct_count: Precision::Absent,
+                    }
+                })
+                .collect_vec();
+
+            Some((
+                parse_path(scan.snapshot().table_root(), view.path_raw().as_ref()).ok()?,
```
rtyler, line 303: I am very confused by this line. There's a `?` but the `extract_file_statistics` function doesn't propagate a result type, it returns a `HashMap` :confused:

crates/core/src/kernel/snapshot/iterators.rs
```diff
@@ -113,6 +115,16 @@ impl LogicalFileView {
         percent_decode_str(raw).decode_utf8_lossy()
     }
 
+    /// Returns the raw file path as stored in the log, without URL decoding.
+    pub(crate) fn path_raw(&self) -> &str {
```
rtyler, line 119: This function makes me uneasy. I feel like it's practically guaranteed to cause a bug later.  I don't have a good suggestion here unfortunately.  This seems like a hack to work around the hack that's four lines above it  :sob:

crates/core/src/delta_datafusion/table_provider/next/mod.rs
roeap: will add more comments. in short though, since all these sets work on references we need to drop them to not get into lifetime issues.

crates/core/src/delta_datafusion/table_provider/next/mod.rs
roeap: granted ... the struct itself is not pub at all, so I'll just remove the pub here as well :).

crates/core/src/delta_datafusion/table_provider/next/mod.rs
roeap: yes

crates/core/src/delta_datafusion/table_provider/next/mod.rs
roeap: partial stuff ... the parquet plan can be do much more, but for this I wanted a dedicated review.

crates/core/src/delta_datafusion/table_provider/next/replay.rs
roeap: good idea!

crates/core/src/delta_datafusion/table_provider/next/scan.rs
roeap: the doc is wrong. these are selection vectors.

crates/core/src/delta_datafusion/table_provider/next/scan.rs
roeap: only we can create it ... I was contemplating if we need the DeltaScanExec to be pub. In the end its mostly needed for downcasting if people build visitors and want to extract stats etc. So it needs to be public, but I felt there is too much in flux that we wnat to allow people to construct it themselves just yet.

crates/core/src/delta_datafusion/table_provider/next/replay.rs
roeap: this also works for options :). 

since statistics are always optional, i thought this would be to just swallow an error here. so if this errors we cannot process the stats and they will default to unknown when we aggregate stats on a plan level.

crates/core/src/delta_datafusion/table_provider/next/scan.rs
roeap: scans will not be affected. this is when internal operators a generating new filters as certain merge operators might.

To be honest, I need too understand what this is doing a bit more to clean up the implementation. I think @alexwilcoxson-rel added this functionality to our existing providers and might know more when this is important.

crates/core/src/delta_datafusion/table_provider/next/replay.rs
roeap: well ... we capture this as the data enters our realm of influence, so I still think all the work was and is great :).

everything downstream should work against urls and also assume they might be pointing at multiple stores.

crates/core/src/delta_datafusion/table_provider/next/scan.rs
roeap: w.r.t. `(super)`, IMO the tighter we can scope things the better, no? At least in this case, the struct is really only useful for being created via our DeltaScan - i.e. the Exec that DeltaScan produces.

Since there is a clear hierarchy, i thought it made sense?

crates/core/src/kernel/snapshot/scan.rs
roeap: will update.

crates/core/src/kernel/snapshot/iterators.rs
roeap: Could not agree more! In fact, seeing encoding errors show up was quite a shock moment!

The good news is our legacy codebase (relying on `Add`s) is shrinking quickly by now. and in these places we need to do encoding back and forth. So there might be light at then end of the tunnel.

crates/core/src/delta_datafusion/table_provider/next/mod.rs
roeap: i think this is the price we have to pay on the journey to squashing these 😢.

crates/core/src/delta_datafusion/table_provider/next/scan_meta.rs
roeap: removed and created an [issue](https://github.com/delta-io/delta-rs/issues/4037) to track this,

crates/core/src/delta_datafusion/table_provider/next/mod.rs
roeap: gone!

crates/core/src/delta_datafusion/table_provider/next/scan.rs
roeap: done 👍

crates/core/src/delta_datafusion/table_provider/next/scan.rs
roeap: to be the code just looked really messy otherwise as we go quite deep in nesting.

The branches are quite similar indeed, but differed enough that I saw no obvious way to share more code.

crates/core/src/delta_datafusion/table_provider/next/mod.rs
roeap: removed the method all together. I do have a commit queued up to add a kernel commit based insert_into method. But thats for another PR :).

PR #4040, opened 2026-01-03
feat: insert into via kernel transaction

roeap:
# Description

Some explorations that might eventually get merged around writing to delta tables using the Kernel APIs.

Writing commits through kernel would enable writing to tables with row tracking or other advanced features enabled. Although there is missing stats support, which is something we cannot accept. This PR contains a workaround/hack for this, by injecting stats when writing the data to disk.

@rtyler - another big question is how we can integrate with dynamo DB locking. Would a custom `Committer` implementation help us here maybe?

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/data_sink.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/data_sink.rs
@@ -1,0 +1,488 @@
+use std::any::Any;
+use std::fmt;
+use std::sync::Arc;
+
+use arrow::array::{AsArray, Int64Builder, MapBuilder, StringBuilder};
+use arrow::datatypes::SchemaRef;
+use arrow_array::{ArrayRef, RecordBatch, StructArray};
+use arrow_schema::{DataType, Field, Schema};
+use datafusion::common::stats::Precision;
+use datafusion::common::{Statistics, exec_datafusion_err, exec_err};
+use datafusion::config::TableParquetOptions;
+use datafusion::datasource::file_format::parquet::ParquetSink;
+use datafusion::datasource::physical_plan::FileSinkConfig;
+use datafusion::datasource::physical_plan::parquet::metadata::DFParquetMetadata;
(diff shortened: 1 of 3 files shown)


ion-elgreco:
Why would we want to use a parquet sink and not our own writer, which supports many things in terms of file size limits, concurrency configs etc

roeap:
> Why would we want to use a parquet sink and not our own writer, which supports many things in terms of file size limits, concurrency configs etc

Not sure yet if we should :). The motivation behind implementing it this way is to get a feel how much benefit the (IMO) very advanced architecture behind the datafusion implementation can give us. Spiecifically how they handle sending data to various partiton writers i found very interesting.. It is exactly that process though that we need to control if we ever what liquid clustering. so we'll see :)

rtyler:
> another big question is how we can integrate with dynamo DB locking. Would a custom Committer implementation help us here maybe?

@roeap It's not _locking :smile:  it's the LogStore implementation which uses DynamoDB. If delta-kernel is supposed to be taking writes via this Transaction, then I am inclined to say we cannot/should not bring it in until it is at feature parity with our existing writer implementations which includes the DynamoDBLogStore implementation. 

Otherwise this introduces a broken writer implementation into delta-rs. 

I think this is really interesting work nonetheless, but I am **very** suspicious of the write support in kernel and its compatibility with existing writer implementations in the ecosystem. Right now I think this would offer S3 users a very easy way to potentially corrupt a table without proper LogStore support (or equivalent).

djouallah:
@rtyler for my own education, why do we still need a LogStore , i thought S3 has put if absent thing ?

rtyler:
> @rtyler for my own education, why do we still need a LogStore , i thought S3 has put if absent thing ?

Delta/Spark still (:scream:) does not support put-if-absent to my knowledge. The DynamoDBLogStore was created specifically to support multiple writers with Spark and delta-rs. Previously we had our own mechanism which was more akin to an actual lock but moved away from that in order to be cooperative with the implementation in Delta/Spark.

Issue #4041, opened 2026-01-04
[Bug]: Linux ARM wheels missing for 1.3.0


potiuk:
### What happened?

This is continuation of #4029  - while .sdist and AMD linux whl packages have been uploaded, the ARM linux whl packages are still missing - and it causes problems in CI systems that run tests on ARM in linux, also it adds a lot of overhead for anyone running linux docker on Mac OS ARM devices (i.e. pretty much all of modern Mac OS).

This is quite a big issue for Airflow - we not only run our CI with ARM, but we also run tests in parallel and a lot of our contributors use ARM MacOS devices - and we have there `breeze` development environment that uses docker-compose to be able to reproduce CI environment and we are using debian-bookworm based images for this environment.

The effect of runnig CI in parallel for deltalake is that when we are building several builds in parallel, and they want to build deltalake in parallel, we are sharing `uv` cache between the processes and in such case, only one process gets a lock and rebuilds deltalake - this lock however is kept for the whole time of building deltalake and - with multiple processes - the build takes more than default `uv` 300 seconds and build fails:

Example here: https://github.com/apache/airflow/actions/runs/20687292566/job/59390484354

```
    × Failed to download and build `deltalake==1.3.0`
    ├─▶ Failed to acquire lock on the distribution cache
    ├─▶ Could not acquire lock
    ╰─▶ Timeout (300s) when waiting for lock on
        `/root/.cache/uv/sdists-v9/pypi/deltalake/1.3.0` at
        `/root/.cache/uv/sdists-v9/pypi/deltalake/1.3.0/.lock`, is another uv
        process running? You can set `UV_LOCK_TIMEOUT` to increase the timeout.
```

We could probably increase the lock timeout, but actually this is an indication that not having ARM whls increases the time to run our CI tests quite dramatically - it's not cool to delay most of our build CI by more than 300 seconds  (5 mins) just because a .whl file is missing.

For now we simply exclude 1.3.0 version of deltalake, and hope that if not for 1.3.0, this will be fixed for 1.3.1

### Expected behavior

Whl file should be present for ARM linux and `pip install deltalake` or `uv pip install deltalake` should not take several minutes.

### Operating System

Linux

### Binding

Python

### Bindings Version

1.3.0

### Steps to reproduce

1. Use MacOS ARM device and have docker desktop or similar installed.
2. Run `docker run -it python:3.12 bash`
3. In docker run `pip install deltalake==1.2.1` and observe that it installs quickly from whl (seconds)
[...]

PR #4043, opened 2026-01-04
chore(ci): use ubuntu-arm images for linux arm builds

abhiaagarwal:
# Description

The PyPI builds for Linux Arm64 have been OOMing recently. This is perhaps due to cross compilation. GHA now has native arm images, let's see if they they improve things.

# Related Issue(s)

Potential solution to #4041 

# Documentation

https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md

### DIFF
--- a/.github/workflows/python_release.yml
+++ b/.github/workflows/python_release.yml
@@ -73,7 +73,7 @@ jobs:
 
   release-pypi-manylinux-x86-64:
     needs: validate-release-tag
-    name: PyPI release on Linux (x86_64-manylinux-2014)
+    name: PyPI release on Linux (x86_64-manylinux-2_17)
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v5
@@ -91,11 +91,11 @@ jobs:
           command: build
           container: off
           args:  -m python/Cargo.toml --zig --profile python-release
-          manylinux: "2014"
+          manylinux: "2_17"
 
   release-pypi-manylinux-228-aarch64:
     needs: validate-release-tag
-    name: PyPI release on Linux (aarch64-manylinux-2014)
+    name: PyPI release on Linux (aarch64-manylinux-2_28)
     runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v5
@@ -113,7 +113,7 @@ jobs:
           command: build
           container: off
           args:  -m python/Cargo.toml --zig --profile python-release
-          manylinux: "2014"
+          manylinux: "2_28"
 
   release-pypi-musl-x86-64:
     needs: validate-release-tag


.github/workflows/cleanup-pr-cache.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   cleanup:
     name: Cleanup PR Caches
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
```
abhiaagarwal, line 11: Also took the opportunity to switch to the ubuntu-slim images for jobs that are just doing API calls, which spin up quicker. I also have the suspicion that if we use this, Github will schedule more runners for us since we're giving up some capacity https://github.blog/changelog/2025-10-28-1-vcpu-linux-runner-now-available-in-github-actions-in-public-preview/

.github/workflows/python_release.yml
```diff
@@ -89,12 +89,14 @@ jobs:
         with:
           target: x86_64-unknown-linux-gnu
           command: publish
+          container: off
           args: --skip-existing -m python/Cargo.toml --no-sdist --profile python-release
+          manylinux: "2_17"
```
abhiaagarwal, line 94: Per maturin's docs, it by default builds inside a docker container. You can disable this and build on the host, which will use zig as the linker. Shouldn't be a real downside to doing this.

rtyler:
Can you change on of the Python build jobs to run this in the PR to test?

rtyler:
@abhiaagarwal this looks like it's now correctly running everything, will you have time to re-set the release workflow for merge or would you like me to do that?

.github/workflows/python_release.yml
```diff
@@ -88,49 +88,52 @@ jobs:
...
           MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
         with:
           target: aarch64-unknown-linux-gnu
-          command: publish
-          args: --skip-existing -m python/Cargo.toml --no-sdist --profile python-release
-          manylinux: "2_28"
+          command: build
+          container: off
+          args:  -m python/Cargo.toml --zig --profile python-release
+          manylinux: "2014"
```
ion-elgreco: Why are we removing 2_28, we need wheels with glibc 2.28+ due to some funky issues with signatures.

See the pr where I introduced this:

https://github.com/delta-io/delta-rs/pull/3263

ion-elgreco (changes requested):

.github/workflows/python_release.yml
ion-elgreco: We need to revert this before merging, otherwise we resurface old issues again

PR #4050, opened 2026-01-06
feat: move data validation on a stream

roeap:
# Description

Moves data validation into a Datafusion execution plan and associated stream. We make this change in the hopes of fully leveraging Datafusion's execution model and more seamlessly integrating validation in other code paths.

We now compile all collected predicates we need to evaluate into a single, optimised physical predicate. This however means, we do loose the ability to track back failures to individual predicates.

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider.rs
+++ b/crates/core/src/delta_datafusion/table_provider.rs
@@ -12,7 +12,7 @@ use datafusion::catalog::TableProvider;
 use datafusion::catalog::memory::DataSourceExec;
 use datafusion::common::pruning::PruningStatistics;
 use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion};
-use datafusion::common::{Column, DFSchema, Result, Statistics, ToDFSchema};
+use datafusion::common::{Column, DFSchemaRef, Result, Statistics, ToDFSchema};
 use datafusion::config::{ConfigOptions, TableParquetOptions};
 use datafusion::datasource::TableType;
 use datafusion::datasource::physical_plan::{
@@ -21,7 +21,6 @@ use datafusion::datasource::physical_plan::{
 use datafusion::datasource::physical_plan::{FileScanConfigBuilder, ParquetSource};
 use datafusion::datasource::sink::{DataSink, DataSinkExec};
 use datafusion::error::DataFusionError;
-use datafusion::execution::context::ExecutionProps;
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
 use datafusion::logical_expr::dml::InsertOp;
 use datafusion::logical_expr::simplify::SimplifyContext;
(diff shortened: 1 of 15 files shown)


crates/core/src/operations/constraints.rs
```diff
@@ -147,13 +148,12 @@ impl std::future::IntoFuture for ConstraintBuilder {
...
                 .session
                 .unwrap_or_else(|| Arc::new(create_session().into_inner().state()));
-            register_store(this.log_store.clone(), session.runtime_env().as_ref());
+            update_datafusion_session(this.log_store.as_ref(), session.as_ref())?;
 
-            let scan = DeltaScanBuilder::new(&snapshot, this.log_store.clone(), session.as_ref())
-                .build()
+            let proivider = DeltaScanNext::builder()
+                .with_eager_snapshot(snapshot.clone())
                 .await?;
```
roeap, line 160: as a side effect we are also moving the `AddConstraint` operation to use the new table provider.

rtyler:
Please pull this out of draft once CI is green and I'll take a pass

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -1191,14 +1201,14 @@ fn df_logical_schema(
...
-    context.create_physical_expr(simplified, df_schema).unwrap()
+pub(crate) fn simplify_expr(
+    session: &dyn Session,
+    df_schema: DFSchemaRef,
+    expr: Expr,
+) -> Result<Arc<dyn PhysicalExpr>> {
+    let context = SimplifyContext::new(&session.execution_props()).with_schema(df_schema.clone());
+    let simplifier = ExprSimplifier::new(context).with_max_cycles(10);
+    session.create_physical_expr(simplifier.simplify(expr)?, df_schema.as_ref())
 }
```
roeap, line 1212: there was a subtle bug here. when optimising expressions, DF may replace things like `now()` with literals, the value of which is tracked on the state/`ExecutionProps`. By creating a new instance within the function, we would use different values in different invocations.

roeap:
@rtyler @ion-elgreco - moved this out of draft, ready for review :).

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -87,7 +111,8 @@ pub struct DeltaScan {
 
 impl DeltaScan {
     // create new delta scan
-    pub fn new(snapshot: SnapshotWrapper, config: DeltaScanConfig) -> Result<Self> {
+    pub fn new(snapshot: impl Into<SnapshotWrapper>, config: DeltaScanConfig) -> Result<Self> {
+        let snapshot = snapshot.into();
```
roeap, line 115: This is an attempt to limit the exposure of `SnapshotWrapper`, while it remains `pub`, users have now way to `use` it, since it's not exported in any public module.

ion-elgreco:
Hmm wanted to still leave a comment about not being able to see the actual invariant that has failed now @roeap, I do use this quite frequently

roeap:
I feared this might come up @ion-elgreco 😄. 

Lets maybe open an issue which information you would want and how..

Gut reaction is we could still benefit from improved performance and do a detailed pass if we find an error?

Happy to look into this, bit also think a good opportunity to do some designing?

roeap:
Another issue was that we are now raising things in DF land -- so we need to raise their errors.

Since this is for debugging, would it help to maybe just print some failed rows? So you would not see it in the error, but could inspect some rows that violated?

PR #4054, opened 2026-01-08
chore(deps): upgrade to DataFusion 52

ethan-tyler:
### Description
Upgrade DataFusion from 51.0.0 to 52.0.0 (branch-52)

Ref: apache/datafusion#18566

### Changes

- Replace SchemaAdapter with TableSchema API in next scan path
- Wire PhysicalExprAdapterFactory to honor schema_force_view_types
- Normalize predicates for Utf8/Utf8View type compatibility
- Add session plumbing for parquet options (pushdown, view types)
- Update FFI_TableProvider::new signature (3 → 5 args)
- Fix FFI lifetime: wrap provider + context in capsule

### Out of scope
- DV mask padding handled separately in #4058.

### Testing
```
cargo test -p deltalake-core --features datafusion --test integration_datafusion
cargo test -p deltalake-core --features datafusion --test datafusion_table_provider
```


crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -337,11 +333,13 @@ impl DeltaScanStream {
...
-                let mut sv: Vec<bool> = selection_vector.drain(0..sel_len).collect();
-                sv.extend(vec![true; remaining]);
-                Some(sv)
+                return Err(internal_datafusion_err!(
+                    "Selection vector length ({}) is shorter than batch size ({}) for file '{}'. \
+                     This indicates a bug in deletion vector processing.",
+                    selection_vector.len(),
+                    batch.num_rows(),
+                    file_id
+                ));
```
roeap: i believe that the deletion vectors do not encode all trailing non-delete flag values, as a result, we may get selection vectors that are too short and that need to be extended.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -236,6 +236,38 @@ impl DeltaScanConfig {
...
+    ) -> Result<SchemaRef> {
+        // Use base types for parquet schema; view conversion happens after reading
+        let cols = table_config.metadata().partition_columns();
+        let table_schema = Arc::new(Schema::new(
+            base.fields()
+                .iter()
+                .map(|f| self.map_field_for_parquet(f.clone(), cols))
+                .collect_vec(),
+        ));
+        Ok(table_schema)
```
roeap: It was my understanding that, it is in fact desirable to have the parquet reader read data directly into view arrays, as discussed [here](https://datafusion.apache.org/blog/2024/09/13/string-view-german-style-strings-part-1/#faster-parquet-loading) it seems to avoid additional data processing?

WHat is the motivation behind this change or am I mistaken?

roeap (commented):
Thanks for taking care of this @ethan-tyler.

Left a question around parquet reading and view types, that I'd love to understand better.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
alamb: Reading directly into StringViewArray or BinaryViewArray typically can avoid a copy and thus is often faster than StringArray

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
ethan-tyler: Agreed. DVs encode deleted row positions only and a selection mask shorter than the file/batch row count is valid when max_deleted_index < num_rows. I’ll update the drain logic to treat missing positions as true (pad the remainder) and add a regression test for this case.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: The base-then-cast was a “minimize risk” choice, but it may give up the zero-copy Parquet decode win from view arrays. Agreed - I'll honor datafusion.execution.parquet.schema_force_view_types in parquet_read_schema and only rewrite view literals to base types when the scan schema is base-typed.

ethan-tyler:
Note: DV semantics fix is handled in #4058; this PR remains DF52-only. I validated DF52+DV locally on branch df52-integration with cherry-pick f602342e (tests: integration_datafusion, datafusion_table_provider).

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -40,17 +40,7 @@ use crate::kernel::arrow::engine_ext::ExpressionEvaluatorExt;
 /// Wraps a Parquet reader execution plan and applies Delta Lake protocol transformations
 /// to produce the logical table data. This includes:
 ///
-/// - **Column mapping**: Translates physical column names to logical names
```
hntd187, line 77: Why were these docs removed?

ethan-tyler: Thanks for the catch, my mistake. I'll restore the full docs

rtyler:
Thanks for starting the work @ethan-tyler! I hope this isn't going to require arrow 58, if it is we're going to get caught by another tedious bit dependency juggling :juggling_person:

ethan-tyler:
> Thanks for starting the work @ethan-tyler! I hope this isn't going to require arrow 58, if it is we're going to get caught by another tedious bit dependency juggling :juggling_person: 

Happy to help. No risk here, DF52 is pinned to Arrow 57.1. Arrow 57.2 is coming soon and should be a drop-in for DF52. 

Arrow 58 DF upgrade is in progress and @alamb can speak better than me on the timeline. 

I’d recommend landing this now and treating Arrow upgrades separately. Let me know if you prefer a different approach.

alamb:
> Thanks for starting the work @ethan-tyler! I hope this isn't going to require arrow 58, if it is we're going to get caught by another tedious bit dependency juggling 🤹

No, DataFusion 52 still uses arrow 57 et al -- thus there will be no needed arrow juggling upgrades until DataFusion 53 😬

rtyler:
@ethan-tyler 52 was released to crates.io yesterday. If you have time to prepare this PR to merge, that'd be fantastic. If not I think I can square this away on Wednesday morning

ethan-tyler:
> @ethan-tyler 52 was released to crates.io yesterday. If you have time to prepare this PR to merge, that'd be fantastic. If not I think I can square this away on Wednesday morning

yes sir - been chipping away at this today, should be ready to merge soon.

ethan-tyler:
Several builders accept `Arc<dyn Session> `but some operations actually need a concrete SessionState for runtime/execution context. The API is misleading - it looks like any Session works, but we sometimes have to downcast.

I initially tightened this to require SessionState and error otherwise. Backed that out as it's a breaking change for anyone passing a custom Session impl. Current behavior stays: if we can't downcast to SessionState, we fall back to an internal one.

That fallback keeps things compatible but can bite us because any config, runtime settings, or registrations on the caller's session get silently ignored when we fall back. Will open a follow up to make this explicit.

issue created: 
- #4081

ethan-tyler:
Python DF tests are skipped by default, PyPI is not upgraded to DF52 and the FFI mismatch will segfault. Tests gated behind DELTALAKE_RUN_DATAFUSION_TESTS=1.

Will re-enable once PyPI DF52 hits

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -213,7 +219,7 @@ impl DeltaScanConfig {
             file_column_name: None,
             wrap_partition_values: true,
             enable_parquet_pushdown: true,
-            schema_force_view_types: true,
+            schema_force_view_types: false,
```
ion-elgreco, line 222: Why is this being set to false

crates/core/src/delta_datafusion/schema_adapter.rs
```diff
@@ -1,82 +0,0 @@
...
-        ))
-    }
-}
-
-#[derive(Debug)]
-pub(crate) struct SchemaMapping {
-    projected_schema: SchemaRef,
-}
-
-impl SchemaMapper for SchemaMapping {
```
ion-elgreco, line 70: Why is this schema mapper removed?

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -538,12 +544,74 @@ impl<'a> DeltaScanBuilder<'a> {
 
         let stats = stats.unwrap_or(Statistics::new_unknown(&schema));
 
+        // Remap stats to TableSchema order: file_schema columns, then partition columns
```
ion-elgreco, line 547: Why is the order relevant for the stats?

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -709,6 +800,11 @@ impl DeltaTable {
         builder
     }
 
+    /// Build a provider that honors session-level DataFusion configuration.
+    pub fn table_provider_with_session(&self, session: Arc<dyn Session>) -> TableProviderBuilder {
```
ion-elgreco, line 804: Can we remove this, you can already do with_session

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -208,6 +208,10 @@ impl TableProvider for DeltaScan {
         &self,
         filter: &[&Expr],
     ) -> Result<Vec<TableProviderFilterPushDown>> {
+        if self.config.schema_force_view_types {
+            // Avoid a separate FilterExec when view types require in-scan filtering.
+            return Ok(vec![TableProviderFilterPushDown::Exact; filter.len()]);
```
ion-elgreco, line 213: Do you have a reference to docs why this is possible?

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -412,6 +420,20 @@ impl DeltaScanStream {
...
+                let normalized = normalize_predicate_to_schema(filter_expr.clone(), &schema)?;
+                let physical = logical2physical(&normalized, schema.as_ref());
+                self.filter_physical = Some(physical);
+            }
+            if let Some(filter) = self.filter_physical.as_ref() {
+                let selection = filter.evaluate(&result)?.into_array(result.num_rows())?;
+                let selection = as_boolean_array(selection.as_ref())?;
+                result = filter_record_batch(&result, selection)?;
+            }
+        }
```
ion-elgreco, line 435: This change seems unrelated to DF52, or am I missing something? You can you please explain

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -280,12 +280,16 @@ impl DeltaScanMetaStream {
...
-            let missing = batch.num_rows() - selection.len();
-            let filter = if missing > 0 {
-                BooleanArray::from_iter(selection.iter().chain(std::iter::repeat_n(&true, missing)))
-            } else {
-                BooleanArray::from_iter(selection.iter())
-            };
+            if selection.len() != batch.num_rows() {
+                return Err(internal_datafusion_err!(
+                    "Selection vector length ({}) does not match row count ({}) for file '{}'. \
+                     This indicates a bug in deletion vector processing.",
```
ion-elgreco, line 286: Makes sense to error out but not sure why this exists in this PR?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -196,7 +203,9 @@ async fn get_data_scan_plan(
         }
         .into();
         partitioned_file = partitioned_file.with_statistics(Arc::new(f.stats));
-        partitioned_file.partition_values = vec![ScalarValue::Utf8(Some(f.file_url.to_string()))];
+        partitioned_file.partition_values = vec![wrap_partition_value_in_dict(ScalarValue::Utf8(
```
ion-elgreco, line 206: Why are you making this change

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -220,6 +229,13 @@ async fn get_data_scan_plan(
     } else {
         scan_plan.parquet_predicate.as_ref()
     };
+    let has_selection_vectors = files_by_store
+        .values()
+        .any(|files| files.iter().any(|(_, selection)| selection.is_some()));
+    let parquet_pushdown_active =
+        enable_parquet_pushdown && predicate.is_some() && !has_selection_vectors;
+    let skip_filter_predicate =
+        parquet_pushdown_active && scan_plan.parquet_predicate_covers_all_filters;
```
ion-elgreco, line 238: Again, why is this change in a DF52 pr? Don't get me wrong this seems to make sense to push to parquet reading when possible but not sure why you are making the change here in this PR

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -83,12 +86,26 @@ impl KernelScanPlan {
...
         // At this point we should only have supported predicates, but we decide where
         // when can handle them (kernel scan and/or parquet scan)
-        let (kernel_predicate, parquet_predicate) = process_filters(filters, table_config)?;
+        let (kernel_predicate, parquet_predicate, parquet_predicate_covers_all_filters) =
+            process_filters(filters, table_config)?;
+        let filter_predicate = if config.schema_force_view_types {
+            conjunction(filters.iter().cloned())
+        } else {
+            None
+        };
```
ion-elgreco, line 95: Can you explain this code?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -236,69 +259,150 @@ impl DeltaScanConfig {
...
                     field
                         .as_ref()
                         .clone()
                         .with_data_type(wrap_partition_type_in_dict(field.data_type().clone()))
                         .into()
                 }
+                DataType::Utf8View => field
+                    .as_ref()
+                    .clone()
+                    .with_data_type(wrap_partition_type_in_dict(DataType::Utf8))
```
ion-elgreco, line 368: Is there any point in wrapping view types as dictionary? As far as I understood view types they are already an efficient way of storing the data

python/src/lib.rs
```diff
@@ -13,7 +13,17 @@ mod writer;
 use arrow::pyarrow::PyArrowType;
 use arrow_schema::SchemaRef;
 use chrono::{DateTime, Duration, FixedOffset, Utc};
+use datafusion_ffi::execution::FFI_TaskContextProvider;
 use datafusion_ffi::table_provider::FFI_TableProvider;
+use deltalake::datafusion::execution::TaskContextProvider;
+
+/// FFI capsule wrapper: `provider` MUST be first field (repr(C) offset 0) for pointer aliasing.
+/// `_ctx` prevents SessionContext drop while capsule is alive (FFI_TaskContextProvider uses Weak).
```
ion-elgreco, line 21: Do you have a reference for this, interesting in learning more how this works :)

python/tests/test_datafusion.py
```diff
@@ -1,17 +1,37 @@
...
+        pytest.skip(
+            "DataFusion Python integration requires datafusion>=52 wheels (PyPI currently provides 51.x)"
+        )
     nrows = 5
     table = Table(
         {
             "id": Array(
                 ["1", "2", "3", "4", "5"],
-                Field("id", type=DataType.string_view(), nullable=True),
+                Field("id", type=DataType.string(), nullable=True),
```
ion-elgreco, line 34: We should keep the view types, this was a welcomed change for performance

ion-elgreco (changes requested):

roeap:
>Several builders accept Arc<dyn Session> but some operations actually need a concrete SessionState for runtime/execution context. The API is misleading - it looks like any Session works, but we sometimes have to downcast.

>I initially tightened this to require SessionState and error otherwise. Backed that out as it's a breaking change for anyone passing a custom Session impl. Current behavior stays: if we can't downcast to SessionState, we fall back to an internal one.

>That fallback keeps things compatible but can bite us because any config, runtime settings, or registrations on the caller's session get silently ignored when we fall back. Will open a follow up to make this explicit.

This is major pain point right now and also cause of concern - as you mention we might get inconsistent results as the state holds important information foe eval of predicates etc. We also don't leverage any caches o.a. that are tracked on the session.

As we work through the current refactoring, we should also focus on pushing physical planning up the stack and do whatever we can using logical plans as this does not require a session.

We also need to create a very narrow waist for DF session management since we do need to register our planner for some executions.

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: Intentionally false for the legacy provider- forcing view types changes the read schema but the legacy path doesn't normalize predicates/literals to match. The next-scan path does normalize and view types are safe there. The struct doc calls out the limitation at crates/core/src/delta_datafusion/table_provider.rs:197. 

If we want view types on by default, should be next-scan only unless we port normalization to legacy.

crates/core/src/delta_datafusion/schema_adapter.rs
ethan-tyler: DF52 removed SchemaAdapterFactory/SchemaMapper; replacement is PhysicalExprAdapterFactory (planning-time expr rewrite vs runtime batch mapping). Logic moved to expr_adapter.rs, wired via with_expr_adapter.

Ref: https://github.com/apache/datafusion/blob/main/docs/source/library-user-guide/upgrading.md

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: Statistics::column_statistics is positional- one entry per field, aligned to schema order. If we supply the stats in the wrong order, the wrong column will get them and prune incorrectly. That's why we remap to TableSchema order.

Ref: https://github.com/apache/datafusion/blob/main/datafusion/common/src/stats.rs

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: Agreed - this is redundant with table_provider().with_session(session)

crates/core/src/delta_datafusion/table_provider/next/mod.rs
ethan-tyler: When we return Exact, DF won't add a FilterExec above the scan. Here we apply the filter inside DeltaScanExec after normalizing literals to the view-typed schema (exec.rs:423) and returning Exact is valid.

Ref: https://github.com/apache/datafusion/blob/main/docs/source/library-user-guide/custom-table-providers.md#filter-pushdown

That said, blanket Exact for all filters is risky if any aren't fully enforced. Will tighten to per-filter.

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
ethan-tyler: This evaluates filters against the post-transform batch schema (after partition injection, column mapping, view-type conversion). DF52 removed the SchemaMapper hooks we relied on, we now normalize predicates to the actual scan schema and compile to a physical expr. Without this, predicates can be evaluated against mismatched schemas/types.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
ethan-tyler: Previous code did batch.num_rows() - selection.len() which underflows if the mask is longer than the batch. This turns a potential panic into a clear error.

Landed while refactoring DF52 scan paths. Happy to move to a follow up PR if you want

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: The file_id partition field is defined as Dictionary(UInt16, Utf8) in DeltaScanConfig::file_id_field. DF requires partition_values types to match partition column types. Without the wrapper, the type mismatch breaks extract_file_id which expects Dictionary<UInt16, Utf8>.

Ref: https://docs.rs/datafusion-datasource/latest/datafusion_datasource/file_scan_config/fn.wrap_partition_value_in_dict.html

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Fair point on scope. This is about correctness with selection vectors and avoiding double apply. It came up during the DF52 scan-path refactor but isn’t required to compile. Happy to split into a follow up PR.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: Stores the full conjunction so we can re-evaluate inside the scan against the post transform schema with literal normalization (when view types enabled). When the view types are off, we skip this path to avoid double apply.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: Partition columns are typically dictionary encoded in DF (see wrap_partition_type_in_dict). When the logical schema uses Utf8View, we still use Dictionary(UInt16, Utf8) and partition_values match the declared field type exactly as DF requires type alignment. View types are beneficial for data read from Parquet; partition cols are per file constants.

Ref: https://docs.rs/datafusion-datasource/52.0.0/datafusion_datasource/file_scan_config/fn.wrap_partition_type_in_dict.html

python/src/lib.rs
ethan-tyler: DF Python expects __datafusion_table_provider__ to return a PyCapsule that's directly cast to FFI_TableProvider* via capsule.reference(). That requires FFI_TableProvider at offset 0, hence repr(C) with it as the first field. The _ctx field keeps the SessionContext alive since FFI_TaskContextProvider uses a Weak reference.

Refs:
DF Python table provider: https://github.com/apache/datafusion-python/blob/main/docs/source/user-guide/io/table_provider.rst
 https://datafusion.apache.org/python/contributor-guide/ffi.html
Rust repr(C) layout: https://doc.rust-lang.org/reference/type-layout.html#reprc-structs

python/tests/test_datafusion.py
ethan-tyler: Agreed - it shouldn't normalize away view types. Will update to assert view types are preserved when configured.

ethan-tyler:
> > Several builders accept Arc but some operations actually need a concrete SessionState for runtime/execution context. The API is misleading - it looks like any Session works, but we sometimes have to downcast.
> 
> > I initially tightened this to require SessionState and error otherwise. Backed that out as it's a breaking change for anyone passing a custom Session impl. Current behavior stays: if we can't downcast to SessionState, we fall back to an internal one.
> 
> > That fallback keeps things compatible but can bite us because any config, runtime settings, or registrations on the caller's session get silently ignored when we fall back. Will open a follow up to make this explicit.
> 
> This is major pain point right now and also cause of concern - as you mention we might get inconsistent results as the state holds important information foe eval of predicates etc. We also don't leverage any caches o.a. that are tracked on the session.
> 
> As we work through the current refactoring, we should also focus on pushing physical planning up the stack and do whatever we can using logical plans as this does not require a session.
> 
> We also need to create a very narrow waist for DF session management since we do need to register our planner for some executions.

Agreed that the silent fallback is a problem. I opened #4081 to track this, my plan is to make fallback explicit, add a SessionState-only API, and define a narrow session waist for planner registration. Good point on pushing physical planning up and using logical plans where we can.

For this PR I'll keep current behavior but add a warning on fallback.

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
ion-elgreco: Ok thats clear! thanks for explaining :)

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ion-elgreco: Ok got it!

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ion-elgreco: Would be great if you can split in a separate pr stacked on top :)

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ion-elgreco: I understood it was not just for reading but also in memory execution is also faster with string views

ion-elgreco: I understood it was not just for reading but also in memory execution is also faster with string views

ethan-tyler: I agree- view types help execution too. Dictionary(UInt16, Utf8View) is probably the right choice when schema_force_view_types=true. Literal normalization should be fine (it recurses into dictionary value types), but I need to confirm our per-file transform and cast_record_batch can materialize Dictionary<_, Utf8View>. Will add tests and follow up.

ethan-tyler: I looked into this more and verified that predicate normalization handles dict-of-view fine (it recurses into dictionary value types), but the blocker is casting. Arrow currently can't dictionary-pack into Utf8View (CastError: Unsupported output type for dictionary packing: Utf8View).

Supporting dict-of-view would require either building the dictionary directly during transform (skip the cast) or an upstream Arrow fix.

crates/core/src/delta_datafusion/expr_adapter.rs
roeap, line 1: Are we fixing something by doing a custom implementation of this?. I did add several tests that validate the deafult impl can handle things. 

Admittedly it does not cover cases like structs in lists etc.

Do we have some failing tests somewhere that warrant this addition?

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
roeap: Hmm .. the new table provider never actually made use of the old schema mapper, since we push the desired schema all teh way into the parquet read. the logcal to physical translation of the predicate should also already happen. 

Form any predicate that we cannot evaluate `Exact`ly - any that reference non partition columns or cannot be converted to a kernel predicate - should be included in the outer `FilterExec` that datafusion wil generate during filter planning.

DO we have substantial changes in DF52 that alter, or - also absolutely possible - did i miss something? :).

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
roeap: In this case the batch does not originate from a file but we generate the number of row from the file metadata. As such there is always exactly one batch per file and if written correctly we should never have an oversized DV. We may have valid undersized ones.

Should we just move the error up and check `election.len() > batch.num_rows()` and error out in this case?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
roeap: yes, would be great to split this out.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -301,31 +321,41 @@ async fn get_read_plan(
...
-        let config =
-            FileScanConfigBuilder::new(store_url, physical_schema.clone(), Arc::new(file_source))
-                .with_file_groups(file_groups)
-                .with_statistics(statistics)
-                .with_table_partition_cols(vec![file_id_field.as_ref().clone()])
-                .with_limit(limit)
-                .build();
+        statistics
+            .column_statistics
+            .push(ColumnStatistics::new_unknown());
```
roeap, line 351: we need this for the file id column?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -375,27 +405,212 @@ fn cast_record_batch(batch: RecordBatch, target_schema: &SchemaRef) -> Result<Re
...
-        &DataType::Struct(target_schema.fields().clone()),
-        &options,
-    )?
-    .as_struct()
-    .into())
+    crate::kernel::schema::cast::cast_record_batch(&batch, Arc::clone(target_schema), true, true)
+        .map_err(|err| DataFusionError::External(Box::new(err)))
+}
+
+pub(crate) fn normalize_predicate_to_schema(expr: Expr, schema: &SchemaRef) -> Result<Expr> {
```
roeap, line 412: are there now some tests / cases failing? We "should" already use an expression taht corresponds to the corrrect schemas. Where are we diverging?

roeap (changes requested):
@ethan-tyler - sorry for the delay in the review.

There is a lot of stuff happening here that is hard to discern where the need originated, which makes it very hard to review this. Unfortunately this happens in some of the most critical code paths - i.e. when building the scan.

Would it be possible to try and do the very minimal to get things compiled and all current tests passing with DF 52.

In subsequent PRs we can then discuss more fixes and improvements?

crates/core/src/operations/update.rs
```diff
@@ -514,8 +512,14 @@ impl std::future::IntoFuture for UpdateBuilder {
...
-            register_store(this.log_store.clone(), session.runtime_env().as_ref());
-            let state = session_state_from_session(session.as_ref())?;
+
+            let state = session
+                .as_any()
+                .downcast_ref::<SessionState>()
+                .cloned()
+                .unwrap_or_else(|| create_session().state());
+
+            register_store(this.log_store.clone(), state.runtime_env().as_ref());
```
roeap, line 522: this is interesting. what is different here in terms of behavior? `session_state_from_session` should do more or less the same case, and if we call `register_store` before or after casting should have no effect?

WE use the function which is already deprecated to help us keep track of where we need to remove more explicit dependence on `SessionSate`.

ethan-tyler:
> @ethan-tyler - sorry for the delay in the review.
> 
> There is a lot of stuff happening here that is hard to discern where the need originated, which makes it very hard to review this. Unfortunately this happens in some of the most critical code paths - i.e. when building the scan.
> 
> Would it be possible to try and do the very minimal to get things compiled and all current tests passing with DF 52.
> 
> In subsequent PRs we can then discuss more fixes and improvements?

Thanks for keeping me honest, and agree that this got away from me. I'll open a minimal DF52 PR with just what's needed to compile and pass existing tests. Then split the rest into focused follow-ups with proper context. I appreciate for the patience! :)

PR #4057, opened 2026-01-08
feat: expose newest table provider to python

ion-elgreco:
# Description

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/mod.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/mod.rs
@@ -54,7 +54,7 @@ mod scan;
 const FILE_ID_COLUMN_DEFAULT: &str = "__delta_rs_file_id__";
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub(crate) enum SnapshotWrapper {
+pub enum SnapshotWrapper {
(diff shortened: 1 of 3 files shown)


crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -54,7 +54,7 @@ mod scan;
 const FILE_ID_COLUMN_DEFAULT: &str = "__delta_rs_file_id__";
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub(crate) enum SnapshotWrapper {
+pub enum SnapshotWrapper {
```
roeap, line 57: The snapshot wrapper is meant to be a purely temporary thing as we consolidate snapshots - do you think we can keep it `pub(crate)` and hide the wrpping in a new constructor?

roeap (commented):
🎉

crates/core/src/delta_datafusion/table_provider/next/mod.rs
ion-elgreco: I think you do this somwhat in TableProviderBuilder but that was also not exposed..

roeap (approved):

PR #4058, opened 2026-01-08
fix: handle DV mask exhaustion and short masks in batch_project

ethan-tyler:
# Description
This fixes deletion vector selection mask handling across RecordBatch boundaries during scans.

- Add DvMaskResult and consume_dv_mask() to encapsulate mask consumption logic
-  Remove exhausted selection vector entries from DashMap after use
- Pad short masks with true (keep rows) instead of erroring
- Guard against empty batches to prevent panic in extract_file_id
- Add unit tests for mask drain, exhaustion, and empty batch handling

# Related Issue(s)
- Relates to #1094

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/data_validation.rs
+++ b/crates/core/src/delta_datafusion/data_validation.rs
@@ -1,0 +1,867 @@
+use std::any::Any;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::{Context, Poll};
+
+use arrow::array::AsArray;
+use arrow_array::RecordBatch;
+use arrow_schema::{DataType, SchemaRef};
+use datafusion::catalog::Session;
+use datafusion::common::{DFSchema, Statistics, exec_err, plan_datafusion_err, plan_err};
+use datafusion::config::ConfigOptions;
+use datafusion::error::{DataFusionError, Result};
+use datafusion::execution::{RecordBatchStream, SendableRecordBatchStream, TaskContext};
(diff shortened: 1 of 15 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -329,25 +363,29 @@ impl DeltaScanStream {
...
         {
-            if selection_vector.len() >= batch.num_rows() {
-                let sv: Vec<bool> = selection_vector.drain(0..batch.num_rows()).collect();
-                Some(sv)
-            } else {
-                let remaining = batch.num_rows() - selection_vector.len();
-                let sel_len = selection_vector.len();
-                let mut sv: Vec<bool> = selection_vector.drain(0..sel_len).collect();
-                sv.extend(vec![true; remaining]);
-                Some(sv)
```
roeap, line 344: just did a quick pass. would it already suffice to remove the vector from the map once we needed to extend it?

ethan-tyler: Thanks for the response and quick pass! Removing on the extend path helps, but it doesn’t cover the case where we drain exactly batch.num_rows() and the vector becomes empty, or when an empty vector is already present. The fix needs to remove entries whenever the selection vector is exhausted, regardless of which branch got us there.

roeap: makes sense! approving!

PR #4061, opened 2026-01-09
feat: expose new table provider in query builder

ion-elgreco:
# Description
@roeap  how do we disable partition columns to be returned as a dictionary?

I don't think we should want to do this by default.

### DIFF
--- a/python/src/query.rs
+++ b/python/src/query.rs
@@ -40,11 +40,10 @@ impl PyQueryBuilder {
     pub fn register(&self, table_name: &str, delta_table: &RawDeltaTable) -> PyResult<()> {
         let snapshot = delta_table.cloned_state()?;
         let log_store = delta_table.log_store()?;
-        let object_store_url = log_store.object_store_url();
-        let url: &Url = object_store_url.as_ref();
+        let url = log_store.root_url();
 
         self.ctx
-            .register_object_store(url, log_store.object_store(None));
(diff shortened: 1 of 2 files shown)


roeap:
the config gets applied there..

https://github.com/delta-io/delta-rs/blob/642febb05f566e0d31bad2b152d02dce3b0e7322/crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs#L240-L253

I think I updated the default impl for the `DeltaScanConfig` to use new() ...

https://github.com/delta-io/delta-rs/blob/642febb05f566e0d31bad2b152d02dce3b0e7322/crates/core/src/delta_datafusion/table_provider.rs#L378-L387

I am in slight favour of using dict encoded types by default, but only very slightly, so feel free to update :).

ion-elgreco:
@roeap the reason I am against is, it's a differnt type then what we have in the schema. In polars dictionary encoded types become categorical which acts differently, so its weird if you cant roundtrip your data. I will just update the behavior from the python side only

ion-elgreco:
@roeap we are ready to go now :)

roeap:
I think we need to fix the lakefs tests, otherwise, yes :).

ion-elgreco:
@roeap hmm did we lose support for lakefs with the new provider? 

Edit: nvm, I see what to do now

ion-elgreco:
Ok done for tonight, I expexcted just registering the object store would have been enough..

PR #4075, opened 2026-01-13
chore: clean up older ignored tests for deletion vectors and column mapping

rtyler:
These tests were relying on add_actions_table() which is an interface
that @roeap and I both want to get rid of in the near future!

These at least make sure that we can open the tables with these table
features correctly

### DIFF
--- a/.github/workflows/python_release.yml
+++ b/.github/workflows/python_release.yml
@@ -112,7 +112,7 @@ jobs:
           target: aarch64-unknown-linux-gnu
           command: publish
           container: off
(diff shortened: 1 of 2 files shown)


roeap:
looking at all the deleted code we would have had to do a bit of work on this if we wnated to produce this output. i.e. stats are always stored in physical names, and we also do pruning on that level, so converting the stats into that format woild require some more mapping things.

so glad we are removing it. :)

ion-elgreco (approved):

PR #4080, opened 2026-01-14
feat: wip to remove EagerSnapshot

hntd187:
# Description
This begins the Purge of the artist formerly known as `EagerSnapshot` but the concept doesn't really go away as it just rolls it into `Snapshot` and `Snapshot` is meant to fulfill both purposes.

### DIFF
--- a/crates/aws/tests/integration_s3_dynamodb.rs
+++ b/crates/aws/tests/integration_s3_dynamodb.rs
@@ -421,7 +421,7 @@ async fn create_incomplete_commit_entry(
     };
     let prepared = CommitBuilder::default()
         .with_actions(actions)
-        .build(Some(table.snapshot()?), table.log_store(), operation)
+        .build(Some(table.table_state()?), table.log_store(), operation)
         .into_prepared_commit_future()
         .await?;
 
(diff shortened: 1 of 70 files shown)


hntd187:
So @roeap I think a lot of tests are failing because the testing setup is broken from the changes, any ideas what might be the issue here? I know I'm missing like 1 place where eager snapshot was updated after a table operation but I'm not sure where.

roeap:
@hntd187 - i beleive before removing eager snapshot, we should make sure there are no more call sites of `log_data`, then remove `DeltaTableProvider` in favour of our new provider.
Haven't looked in the individual failures, but since its mostly (all) tests for operations, this wpould be consistent since this is the place where the old provider is still being used.

roeap:
also i needed to update your comment to capitalize "Purge" since this has resemblence to the Purge as we march through the codebase ad get right of too many snapshots :D.

hntd187:
> @hntd187 - i beleive before removing eager snapshot, we should make sure there are no more call sites of `log_data`, then remove `DeltaTableProvider` in favour of our new provider. Haven't looked in the individual failures, but since its mostly (all) tests for operations, this wpould be consistent since this is the place where the old provider is still being used.

Good you mention log_data because it's the source of every single failure, I'm just struggling to figure out why.

PR #4084, opened 2026-01-15
refactor: do some logical planning in find_files

roeap:
# Description

This PR refactors `find_files` to make use of the new table provider and makes some improvements to the internal logic along the way. Specifically we are now creating a logical plan rather than a physical one which also lets the datafusion planner handle most of the projection/schema logic.

We also remove one more use of `EagerSnapshot` which leads to updating several call sites and makes the change spread across a few more files.

### DIFF
--- a/crates/core/src/operations/write/generated_columns.rs
+++ b/crates/core/src/operations/write/generated_columns.rs
@@ -28,7 +28,8 @@ pub fn with_generated_columns(
         .map(|gc| {
             Ok::<_, DeltaTableError>((
                 gc.get_name(),
-                parse_predicate_expression(plan.schema(), &gc.generation_expr, session)?,
+                parse_predicate_expression(plan.schema(), &gc.generation_expr, session)?
+                    .alias(gc.get_name()),
             ))
         })
         .try_collect()?;


crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -60,7 +60,7 @@ pub(crate) fn consume_dv_mask(
             should_remove: is_empty,
         }
     } else {
-        let mut sv: Vec<bool> = selection_vector.drain(..).collect();
+        let mut sv: Vec<bool> = std::mem::take(selection_vector);
```
roeap, line 63: This is a drive-by change resulting from running `clippy`.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -236,6 +235,18 @@ async fn find_files_scan(
     session: &dyn Session,
     expression: Expr,
 ) -> DeltaResult<Vec<Add>> {
+    update_datafusion_session(&log_store, session, None)?;
```
roeap, line 193: the fact that we need to update the df session so deep into the stack is not great. We should probably also deprecate the `update_datafusion_session` function since this is mostly another stop-gap as we migrate.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -245,51 +256,38 @@ async fn find_files_scan(
...
+        provider_as_source(provider),
+        projection,
+        vec![expression.clone()],
+    )?
+    .aggregate([col(PATH_COLUMN)], [first_value(lit(1_i32), vec![])])?
+    .build()?;
+    let exec = session.create_physical_plan(&logical_plan).await?;
+
+    let path_batches = collect(exec, session.task_ctx()).await?;
+    let result = join_batches_with_add_actions(path_batches, candidate_map, true, false)?;
```
roeap: This is the main functional change in this PR. I was quite happy to see that this change simplified the code a bit as well.

The plan we generate is not quite the same plan as we did before. Previously we applied a `LocalLimitExec` to find individual files. This operator applies a limit to each partition. This must have thus far always corresponded to individual files, but I am not totally sure that this is something we can guarantee, especially since we now have additional rewrites of the plan as it moves through DFs planning. Instead we now do a `GROUP BY` on the file column with the `first_value` as aggregate function.

crates/core/tests/integration_datafusion.rs
```diff
@@ -1839,7 +1841,8 @@ mod insert_into_tests {
         )?;
 
         let ctx = SessionContext::new();
-        let scan_config = DeltaScanConfigBuilder::new().build(table.snapshot()?.snapshot())?;
+        let scan_config =
+            DeltaScanConfigBuilder::new().build(table.snapshot()?.snapshot().snapshot())?;
```
roeap, line 1845: Don't we just love the `.snapshot()?.snapshot().snapshot()` calls? 🤣

crates/core/src/operations/update.rs
```diff
@@ -271,24 +271,8 @@ async fn execute(
...
-    // our schemas for Lists due to issues discussed
-    // [here](https://github.com/delta-io/delta-rs/pull/2886#issuecomment-2481550560>
-    let rules: Vec<Arc<dyn datafusion::optimizer::OptimizerRule + Send + Sync>> = session
-        .optimizers()
-        .iter()
-        .filter(|rule| {
-            rule.name() != "optimize_projections" && rule.name() != "simplify_expressions"
-        })
-        .cloned()
-        .collect();
```
roeap, line 285: it seems this has been fixed upstream and we no longer need to limit optimisations. I assume tests were failing if we included the change?

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -245,51 +256,29 @@ async fn find_files_scan(
...
 
-    let path_batches = datafusion::physical_plan::collect(limit, session.task_ctx()).await?;
+    // Scan the table with the predicate applied to identify files with matching rows.
+    // We are only interested if any matching row exists in the file, so we
+    // can use a simple first_value aggregate to minimize data movement.
+    let logical_plan = LogicalPlanBuilder::scan("source", provider_as_source(provider), None)?
+        .filter(expression.is_true())?
+        .aggregate([col(PATH_COLUMN)], [first_value(lit(1_i32), vec![])])?
+        .build()?;
+    let exec = session.create_physical_plan(&logical_plan).await?;
```
roeap: This is the main functional change in this PR. I was quite happy to see that this change simplified the code a bit as well.

The plan we generate is not the same plan as we did before. Previously we applied a `LocalLimitExec` to find individual files. This operator applies a limit to each partition. This must have thus far always corresponded to individual files, but I am not totally sure that this is something we can guarantee, especially since we now have additional rewrites of the plan as it moves through DFs planning. Instead we now do a GROUP BY on the file column with the first_value as aggregate function.

PR #4089, opened 2026-01-17
feat: use new provider and logical planning in find_files

roeap:
# Description

The last attempt (#4084) got a bit out of hand, so hoping to land some initial changes. i.e. we are migrating the `find_files` method to use the new table provider and make more use of logical planning.

### DIFF
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -2,26 +2,24 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::sync::Arc;
 
-use arrow_array::{Array, RecordBatch, StringArray};
-use arrow_schema::{DataType, Field, Schema, SchemaRef};
+use arrow_array::RecordBatch;
 use datafusion::catalog::Session;
-use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion, TreeNodeVisitor};
-use datafusion::datasource::MemTable;
-use datafusion::execution::context::SessionContext;
-use datafusion::logical_expr::{Expr, Volatility, col};
-use datafusion::physical_plan::ExecutionPlan;
-use datafusion::physical_plan::filter::FilterExec;
-use datafusion::physical_plan::limit::LocalLimitExec;
+use datafusion::common::tree_node::{TreeNode as _, TreeNodeRecursion, TreeNodeVisitor};
+use datafusion::datasource::provider_as_source;
+use datafusion::functions_aggregate::expr_fn::first_value;
+use datafusion::logical_expr::{Expr, LogicalPlanBuilder, Volatility, col};
+use datafusion::physical_plan::{ExecutionPlan, collect};
+use datafusion::prelude::lit;
 use futures::TryStreamExt;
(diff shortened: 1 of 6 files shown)


crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -225,153 +180,76 @@ fn join_batches_with_add_actions(
...
-    }
-
-    fields.push(Arc::new(Field::new(file_column_name, DataType::Utf8, true)));
+    // Scan the table with the predicate applied to identify files with matching rows.
+    // We are only interested if any matching row exists in the file, so we
+    // can use a simple first_value aggregate to minimize data movement.
+    let logical_plan = LogicalPlanBuilder::scan("source", provider_as_source(provider), None)?
+        .filter(expression.is_true())?
+        .aggregate([col(PATH_COLUMN)], [first_value(lit(1_i32), vec![])])?
+        .build()?;
```
roeap, line 252: its a bit hidden among the changes, but this is the heart of this change, where we create the new logical plan.

This plan is a bit different than what we were doing before. Prior we applied a `LocalLimitExec` which will apply a limit to each partition. I am not sure that we can promise one partition will always correspond to exactly one file. I hope doing an GROUP BY is more robust.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -225,153 +180,76 @@ fn join_batches_with_add_actions(
...
-    let result =
-        join_batches_with_add_actions(path_batches, candidate_map, &file_column_name, true, false)?;
+    let path_batches = collect(exec, session.task_ctx()).await?;
+    let result = join_batches_with_add_actions(path_batches, candidate_map)?;
 
     Span::current().record("matching_files", result.len());
     Ok(result)
 }
 
-async fn scan_memory_table(snapshot: &EagerSnapshot, predicate: &Expr) -> DeltaResult<Vec<Add>> {
```
roeap, line 298: we should not need to disambiguate between metadata-only scans and table scans anymore, since the new table provider will make this choice internally.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -225,153 +180,76 @@ fn join_batches_with_add_actions(
...
     )
 )]
 async fn find_files_scan(
-    snapshot: &EagerSnapshot,
+    snapshot: SnapshotWrapper,
     log_store: LogStoreRef,
     session: &dyn Session,
     expression: Expr,
 ) -> DeltaResult<Vec<Add>> {
+    update_datafusion_session(&log_store, session, None)?;
```
roeap, line 193: i think we should maybe take another hit in cargo warnings and deprecate this function. IF we no longer need this its a good sign we have our session management under control.

roeap:
@ion-elgreco - i cannot reproduce the python failure locally right now, it works on my machine 😆. Any chance you have some insights into what's going on here?

ion-elgreco:
> @ion-elgreco - i cannot reproduce the python failure locally right now, it works on my machine 😆. Any chance you have some insights into what's going on here?

Looks to me that load_cdf has a bug now because you are seeing one more row than you should

PR #4090, opened 2026-01-17
refactor!: more logical writes

roeap:
# Description

It seems we may have found an avenue into moving further into logical planning that does not break stuff 😆. This PR updates the write operation to no longer make use of `DataFrame` and use logical plans.

Turns out that handling generated columns is a significant part of this. I'll try and remove the the usage of current gc-related function in a follow-up. This touches merge though, so keeping it separate.

### DIFF
--- a/crates/core/src/operations/write/generated_columns.rs
+++ b/crates/core/src/operations/write/generated_columns.rs
@@ -1,26 +1,72 @@
-use crate::kernel::EagerSnapshot;
-use datafusion::common::ScalarValue;
-use datafusion::logical_expr::{ExprSchemable, col, when};
+use arrow_schema::Schema;
+use datafusion::catalog::Session;
+use datafusion::common::{HashMap, Result, ScalarValue, plan_err};
+use datafusion::logical_expr::{ExprSchemable, LogicalPlan, LogicalPlanBuilder, col, when};
 use datafusion::prelude::lit;
 use datafusion::{execution::SessionState, prelude::DataFrame};
 use delta_kernel::engine::arrow_conversion::TryIntoArrow as _;
+use delta_kernel::table_features::TableFeature;
+use itertools::Itertools as _;
 use tracing::debug;
 
-use crate::{DeltaResult, kernel::DataCheck, table::GeneratedColumn};
+use crate::{
+    DeltaResult, DeltaTableError,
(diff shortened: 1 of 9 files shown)


abhiaagarwal:
> Turns out that handling generated columns is a significant part of this. 

When I was working on my branch, this was also a problem. I think the right way to solve it would be to use a session at ingestion time to convert the SQL -> logical plan and store that. That, of course, could lead to inconsistencies between the functions that the generated columns had access to and what we use in planning.

roeap:
Yes @abhiaagarwal! In general we need to defer physical planning as much as possible and see that we have clear APIs and guidelines for managing sessions.

crates/core/src/operations/write/mod.rs
```diff
@@ -430,29 +439,35 @@ impl std::future::IntoFuture for WriteBuilder {
                 // and should be created
                 let mut actions = this.check_preconditions().await?;
 
+                let Some(mut source) = this.input.take() else {
+                    return Err(WriteError::MissingData.into());
```
ion-elgreco, line 443: Check_preconditions already does this check

crates/core/src/operations/write/generated_columns.rs
```diff
@@ -1,26 +1,72 @@
...
-        if snapshot.protocol().min_writer_version() == 7
-            && !features.contains(&delta_kernel::table_features::TableFeature::GeneratedColumns)
-        {
-            return Ok(false);
-        }
+#[inline]
+pub fn gc_is_enabled(snapshot: &EagerSnapshot) -> bool {
+    snapshot
+        .table_configuration()
+        .is_feature_enabled(&TableFeature::GeneratedColumns)
```
ion-elgreco, line 24: Will this feature be in the snapshot with protocol v4 writer? If not then the behavior is not the same

roeap: good point! and yes, this method will check feature enablement via version and feature lists.

crates/core/src/operations/write/mod.rs
roeap: yes, saw that when looking into that fn ... but i wanted to cleanly take the input and not have the unwrap ... so purely cosmetic :).

PR #4092, opened 2026-01-18
chore(deps): upgrade datafusion to 52.0.0

ethan-tyler:
# Description
Minimal upgrade to compile and pass existing tests on DF52.
Supersedes #4054.

## Changes
**Dependencies:**
- Bump datafusion* crates to 52.0.0

**DF52 API migrations:**
- Remove SchemaAdapter (DF52 removed it)
- Update scan builders to TableSchema + FileScanConfigBuilder (legacy, next-scan, CDF paths)
- Update ColumnStatistics struct literals
- Remove session_state_from_session shim, error if caller supplies non-SessionState (no silent fallback)

**DF52/Arrow57 scalar compatibility:**
- Normalize Utf8View/BinaryView literals in predicates for pushdown compatibility
- Arrow57 requires matching scalar types for comparison kernels

**Correctness fix:** 
DF52 enables batch coalescing by default, producing batches with rows from multiple files. Next-scan assumed one file per batch.

- Split batches by contiguous file_id runs before applying per-file DVs/transforms
- Queue multiple output batches from single input
- Add test for dictionary-based file_id extraction

**Test updates:**
- Loosen stats assertions for DF52 Precision::Inexact

**Follow Ups (PR Splits):**
- PhysicalExprAdapter / expr adapter work
- schema_force_view_types changes
- Schema evolution extras
- Python surface-area changes

**Verification:**
- `cargo test -p deltalake-core --features datafusion --test integration_datafusion`

### Related Issue(s)
- Supersedes #4054

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,4 @@
+[profile.ci]
+fail-fast = false
+retries = 1
+slow-timeout = "60s"
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -39,6 +39,7 @@ jobs:
       AZURE_STORAGE_ALLOW_HTTP: "1"
       AZURITE_BLOB_STORAGE_URL: "http://localhost:10000"
       AZURE_STORAGE_CONNECTION_STRING: "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;"
+      NEXTEST_PROFILE: 'ci'
 
     steps:
       - uses: actions/checkout@v5
@@ -54,6 +55,7 @@ jobs:
 
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
+      - uses: taiki-e/install-action@nextest
 
       - name: Start emulated services
         run: docker compose up -d
@@ -61,7 +63,7 @@ jobs:
       - name: Run tests with rustls (default)
         run: |
           gmake setup-dat
-          cargo llvm-cov --profile=ci \
+          cargo llvm-cov nextest --cargo-profile=ci \
             --package deltalake-${{ matrix.target.name }} \
             --features integration_test \
             --codecov \
@@ -101,13 +103,15 @@ jobs:
           toolchain: stable
           cache: true
 
+      - uses: taiki-e/install-action@nextest
+
       - name: Start emulated services
         run: docker compose up -d
 
       - name: Run tests with native-tls
         run: |
           gmake setup-dat
-          cargo test --profile=ci \
+          cargo nextest r --cargo-profile=ci \
             --package deltalake-aws \
             --no-default-features \
             --features integration_test,native-tls
@@ -129,6 +133,7 @@ jobs:
 
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
+      - uses: taiki-e/install-action@nextest
 
       # Install Java and Hadoop for HDFS integration tests
       - uses: actions/setup-java@v4
@@ -145,7 +150,7 @@ jobs:
       - name: Run tests with rustls (default)
         run: |
           gmake setup-dat
-          cargo llvm-cov --profile=ci \
+          cargo llvm-cov nextest --cargo-profile=ci \
             --features integration_test \
             --package deltalake-hdfs \
             --codecov \
@@ -177,6 +182,7 @@ jobs:
 
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
+      - uses: taiki-e/install-action@nextest
 
       - name: Download Lakectl
         run: |
@@ -190,7 +196,7 @@ jobs:
       - name: Run tests
         run: |
           gmake setup-dat
-          cargo llvm-cov --profile=ci \
+          cargo llvm-cov nextest --cargo-profile=ci \
             --package deltalake-lakefs \
             --features integration_test_lakefs \
             --codecov \
--- a/crates/aws/tests/common.rs
+++ b/crates/aws/tests/common.rs
@@ -3,6 +3,7 @@ use deltalake_aws::constants;
 use deltalake_aws::register_handlers;
 use deltalake_aws::storage::*;
 use deltalake_test::utils::*;
+use rand::RngCore;
 use rand::random;
 use std::process::{Command, ExitStatus, Stdio};
 
@@ -14,8 +15,9 @@ pub struct S3Integration {
 impl Default for S3Integration {
     fn default() -> Self {
         register_handlers(None);
+        let task_id = format!("{}", rand::thread_rng().next_u64());
         Self {
-            bucket_name: format!("test-delta-table-{}", Utc::now().timestamp()),
+            bucket_name: format!("delta-table-{}-{task_id}", Utc::now().timestamp()),
         }
     }
 }
--- a/crates/azure/tests/context.rs
+++ b/crates/azure/tests/context.rs
@@ -2,6 +2,7 @@ use azure_storage_blobs::prelude::*;
 use chrono::Utc;
 use deltalake_azure::register_handlers;
 use deltalake_test::utils::*;
+use rand::RngCore;
 use std::path::PathBuf;
 use std::process::ExitStatus;
 use tokio::runtime::Handle;
@@ -17,7 +18,8 @@ pub enum MsftIntegration {
 impl Default for MsftIntegration {
     fn default() -> Self {
         register_handlers(None);
-        Self::Azure(format!("test-delta-table-{}", Utc::now().timestamp()))
+        let task_id = format!("{}", rand::thread_rng().next_u64());
+        Self::Azure(format!("delta-table-{}-{task_id}", Utc::now().timestamp()))
     }
 }
 


ion-elgreco:
"Arrow57 requires matching scalar types for comparison kernels"

@alamb is there a reason why arrow 57 is more restrictive with literal comparison?

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -538,12 +539,82 @@ impl<'a> DeltaScanBuilder<'a> {
...
+        let mut reordered_stats = file_col_stats;
+        reordered_stats.extend(partition_col_stats);
+
+        let stats = Statistics {
+            num_rows: stats.num_rows,
+            total_byte_size: stats.total_byte_size,
+            column_statistics: reordered_stats,
+        };
+
+        // Add unknown stats for file_column if present (it's added as partition field but not in original schema)
```
ion-elgreco, line 594: Just curious, what does datafusion do if you have a column in the schema but no statistics at all, not even a mention of the column?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -236,6 +236,43 @@ impl DeltaScanConfig {
         Ok(table_schema)
     }
 
+    fn parquet_file_schema(
+        &self,
+        table_config: &TableConfiguration,
+        base: &Schema,
+    ) -> Result<SchemaRef> {
+        // IMPORTANT: This schema is used for Parquet reading and predicate pushdown.
+        // It must NOT apply view-type conversions (Utf8View/BinaryView) because:
```
ion-elgreco: So what happens if you do not that?

python/tests/test_lakefs.py
```diff
@@ -516,7 +516,7 @@ def test_update(lakefs_path, sample_table_update: Table, lakefs_storage_options)
         {
             "id": Array(
                 ["1", "2", "3", "4", "5"],
-                type=ArrowField("id", DataType.string_view(), nullable=True),
+                type=ArrowField("id", DataType.string(), nullable=True),
```
ion-elgreco: We should revert to view types for Python

alamb:
> "Arrow57 requires matching scalar types for comparison kernels"
> 
> @alamb is there a reason why arrow 57 is more restrictive with literal comparison?

I am not sure what this is referring to

I don't think arrow changed in this regards

DataFusion has a bunch of code to coerce literals to the corrct type, and I think some of the work to change how pushdown worked results in maybe not calling that code implicitly anymore -- you may have to call it explicitly

Maybe @adriangb  can point you at the correct APIs etc (e.g. the SchemaAdapter stuff)

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -359,8 +362,65 @@ struct DeltaScanStream {
 }
 
 impl DeltaScanStream {
-    /// Apply the per-file transformation to a RecordBatch.
-    fn batch_project(&mut self, batch: RecordBatch) -> Result<RecordBatch> {
+    fn batch_project(&mut self, batch: RecordBatch) -> Result<Vec<RecordBatch>> {
```
roeap: Not sure what is the exact case, but I was under the assumption that `ParquetExec` would not coalesce batches and we could expect that one batch always originates from one file. if this is not something we can rely on on, we should at the very least be able to rely on that consecutive roes will be from one file. In that case we might want to look into the `partition` kernel which we also use for when writing data to partitions.

ethan-tyler:
> > "Arrow57 requires matching scalar types for comparison kernels"
> > @alamb is there a reason why arrow 57 is more restrictive with literal comparison?
> 
> I am not sure what this is referring to
> 
> I don't think arrow changed in this regards
> 
> DataFusion has a bunch of code to coerce literals to the corrct type, and I think some of the work to change how pushdown worked results in maybe not calling that code implicitly anymore -- you may have to call it explicitly
> 
> Maybe @adriangb can point you at the correct APIs etc (e.g. the SchemaAdapter stuff)

@ion-elgreco: With DF52, view types are more prevalent (schema_force_view_types defaults true). In next-scan we keep the Parquet pushdown schema as base Utf8/Binary and apply view conversions only in the result schema. Predicates may contain view-typed literals from the query, we normalize those to base types before pushdown.

@alamb: In delta-rs we build pushdown predicates and call logical2physical directly, which bypasses the analyzer's coercion passes. DF's physical planning assumes coercion is already done and logical2physical won't coerce for us, we normalize literals to match the scan schema explicitly before calling it. If there's a canonical DF52 API for predicate coercion we should use instead, happy to switch.

Will update the PR description to clarify this is a pushdown plumbing issue, not Arrow strictness.

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: Good question - Statistics.column_statistics is positional, must align 1:1 with TableSchema order. Missing entries break downstream assumptions, we remap by name and fill gaps with ColumnStatistics::new_unknown().

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: Pushdown errors on type mismatch. Parquet stores Utf8/Binary, if we told ParquetSource to operate on Utf8View/BinaryView, the comparison kernels would see different types. We keep base types for the Parquet schema, apply view conversions only in the result schema, and normalize view-typed literals before pushdown.

python/tests/test_lakefs.py
ethan-tyler: Agreed. The DataType.string() changes were a workaround. The Rust side literal normalization is already in place, reverting to DataType.string_view() should work.

alamb:
> @alamb: In delta-rs we build pushdown predicates and call logical2physical directly, which bypasses the analyzer's coercion passes. DF's physical planning assumes coercion is already done and logical2physical won't coerce for us, we normalize literals to match the scan schema explicitly before calling it. If there's a canonical DF52 API for predicate coercion we should use instead, happy to switch.

DataFusion now does something similar to this in 52

Specifically, it uses [`PhysicalExprAdapter`](https://docs.rs/datafusion/latest/datafusion/physical_expr_adapter/trait.PhysicalExprAdapter.html) to rewrite a physical expression that was in terms of the table schema (logical schema) in to an expression in terms of the file schema (that might have different types than the table schema) 

There is a corresponding [`PhysicalExprSimplifier` ](https://docs.rs/datafusion/latest/datafusion/physical_expr/struct.PhysicalExprSimplifier.html) that then performs some base optimizations on this expression (like unwraping column casts and applying the cast to the literal)

So it sounds like (without looking at the code in detail) you may be able to use those two methods now instead of custom logic in delta-rs

ethan-tyler:
> 
> DataFusion now does something similar to this in 52
> 
> Specifically, it uses [`PhysicalExprAdapter`](https://docs.rs/datafusion/latest/datafusion/physical_expr_adapter/trait.PhysicalExprAdapter.html) to rewrite a physical expression that was in terms of the table schema (logical schema) in to an expression in terms of the file schema (that might have different types than the table schema)
> 
> There is a corresponding [`PhysicalExprSimplifier` ](https://docs.rs/datafusion/latest/datafusion/physical_expr/struct.PhysicalExprSimplifier.html) that then performs some base optimizations on this expression (like unwraping column casts and applying the cast to the literal)
> 
> So it sounds like (without looking at the code in detail) you may be able to use those two methods now instead of custom logic in delta-rs

Thanks @alamb! Super helpful context, and I agree this is the right direction.

Why I went with the literal normalization shim for now: next-scan builds the Parquet pushdown predicate already in file-schema space (base Utf8/Binary), then calls logical2physical. The adapter expects a table-schema expression to translate and using it would require restructuring the predicate planning flow.

I had adapter wiring in #4054, but per guidance I split it to keep the DF52 upgrade reviewable. This PR is the minimal extraction; the adapter approach is planned as a focused follow up.

Question for that follow up: Would you recommend wiring DefaultPhysicalExprAdapterFactory and PhysicalExprSimplifier directly (mirroring DF's Parquet scan path), or is a custom adapter more appropriate given next-scan's kernel/DV constraints?

ethan-tyler:
@roeap @ion-elgreco - confirming alignment on path forward:

This PR: Minimal DF52 upgrade (API migrations, batch coalescing fix) and a narrow, temporary shim to normalize Utf8View/BinaryView literals for next-scan Parquet pushdown. Keeps base-typed Parquet read schema and avoids type mismatch failures.

Proposed plan:
- Re-introduce PhysicalExprAdapterFactory and PhysicalExprSimplifier pipeline from #4054 as a focused PR for next-scan pushdown correctness and robustness
- Decide and document schema_force_view_types behavior in next-scan (provider schema vs parquet read schema). This touches kernel/DV compatibility and perf tradeoffs
- Expand schema evolution and pushdown regression tests (nested fields, missing cols, view/base casts) once the adapter path lands

Does this plan work, or would you prefer different ordering or scope?

alamb:
BTW @adriangb  added a helper here for batch conversion here:
- https://github.com/apache/datafusion/pull/19877

This will be released as part of the 52.1.0 release, which I anticipate being released towards the end of this week, if that is helpful

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -316,31 +318,48 @@ async fn get_read_plan(
...
         let file_group: FileGroup = files.into_iter().map(|file| file.0).collect();
-        let (file_groups, statistics) =
-            compute_all_files_statistics(vec![file_group], full_read_schema.clone(), true, false)?;
+        let (file_groups, mut statistics) =
+            compute_all_files_statistics(vec![file_group], physical_schema.clone(), true, false)?;
+
+        // Add unknown stats for file_id column since TableSchema includes it as a partition field
+        statistics
+            .column_statistics
+            .push(ColumnStatistics::new_unknown());
```
roeap: I a lo al attempt i found it advantageous to push this right when we create the partitioned file and also assign the other stats.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -403,9 +431,137 @@ fn cast_record_batch(batch: RecordBatch, target_schema: &SchemaRef) -> Result<Re
...
+                let dt = strip_view_types_from_data_type(f.data_type());
+                if &dt == f.data_type() {
+                    f.as_ref().clone()
+                } else {
+                    f.as_ref().clone().with_data_type(dt)
+                }
+            })
+            .collect::<Vec<_>>(),
+    )
+}
```
roeap: I somehow cannot accept that we need to do this?

What starts failing if we don't. I just trying to get rid of more and more of our custom casting code. WHat spoecifically starts failing so that we need to add this?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -847,4 +1010,201 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_view_literal_conversion_in_predicate_pushdown() -> TestResult {
```
roeap: this test name make me think it's the new physical expression conversion we need?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -847,4 +1010,201 @@ mod tests {
...
+            FILE_ID_COLUMN_DEFAULT,
+            DataType::Dictionary(DataType::UInt16.into(), DataType::Utf8.into()),
+            false,
+        ));
+
+        let predicate = col("name").eq(lit(ScalarValue::Utf8View(Some("bob".to_string()))));
+        let plan = get_read_plan(
+            &session.state(),
+            files_by_store,
+            &arrow_schema,
```
roeap, line 1001: the schema we pass into `get_read_plan` should always be the most specific physical schema around ... i.e. contain VIew types. 

If this is what we are testing here we need to update schema we are passing to have view types.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -275,6 +277,72 @@ impl DeltaScanConfig {
...
+                DataType::LargeList(self.map_field_for_parquet(inner.clone(), partition_cols))
+            }
+            DataType::ListView(inner) => {
+                DataType::ListView(self.map_field_for_parquet(inner.clone(), partition_cols))
+            }
+            // Always use base types for Parquet reads/predicate evaluation.
+            // DataFusion's Parquet reader does not reliably produce view arrays even when enabled,
+            // and requesting view types here can cause schema/array mismatches.
+            DataType::Utf8View => DataType::Utf8,
+            DataType::BinaryView => DataType::Binary,
```
roeap: I don't think we want to do this or should? 

Can you help me understand the motivation?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -431,7 +511,14 @@ fn process_predicate<'a>(
...
             (TableProviderFilterPushDown::Inexact, None)
         } else {
-            (TableProviderFilterPushDown::Inexact, Some(expr))
+            (
+                if parquet_pushdown_exact {
+                    TableProviderFilterPushDown::Exact
+                } else {
+                    TableProviderFilterPushDown::Inexact
+                },
+                Some(expr),
```
roeap: DO we really want to declare more as exact. While this might be true, i did not find a precedent in datafusion where we set this to be exact for anything other then partition columns.

crates/core/src/delta_datafusion/session.rs
```diff
@@ -48,7 +48,16 @@ impl Default for DeltaSessionConfig {
     fn default() -> Self {
         DeltaSessionConfig {
             inner: SessionConfig::default()
-                .set_bool("datafusion.sql_parser.enable_ident_normalization", false),
+                .set_bool("datafusion.sql_parser.enable_ident_normalization", false)
+                .set_bool("datafusion.execution.parquet.schema_force_view_types", true)
+                // Work around DataFusion 52.0.0 bug: hash-join dynamic filtering (IN-list
```
roeap: do we have a link to this bug?

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -538,12 +539,82 @@ impl<'a> DeltaScanBuilder<'a> {
 
         let stats = stats.unwrap_or(Statistics::new_unknown(&schema));
 
+        // DF52's TableSchema outputs columns as: file_schema + partition_columns
+        // Source stats are indexed by TableConfiguration.schema() field order, which may differ
+        // from the scan schema order. We need name-based remapping, not index-based.
```
roeap, line 544: where does this reordering come in? My hope would be that the schema we push into the various operations pushes through ... 

That said, the table provider needs to go and any changes that are purely motivated by needing to migrate this provider should ideally be included in a pre-factor o.a. so that we can revert them easily. I just a few commits away from removing that thing entirely.

python/tests/test_table_read.py
```diff
@@ -1127,11 +1126,11 @@ def test_read_query_builder_join_multiple_tables(tmp_path):
             {
                 "date": Array(
                     ["2021-01-01", "2021-01-02", "2021-01-03", "2021-12-31"],
-                    ArrowField("date", type=DataType.string_view(), nullable=True),
+                    ArrowField("date", type=DataType.string(), nullable=True),
```
roeap, line 1129: we really want to keep outputting StingViews, why is this change needed?

python/src/datafusion.rs
```diff
@@ -266,6 +266,10 @@ mod tests {
                 Ok(None)
             }
         }
+
+        fn reset_state(&self) -> Arc<RwLock<dyn LazyBatchGenerator>> {
+            Arc::new(RwLock::new(TestBatchGenerator::new(self.data.clone())))
+        }
```
roeap, line 273: where is this functionality needed (in DF?)? not sure how we would even reset that but this just copies it?=

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
adriangb: I don't think this is true, it the schema is view types the parquet reader will produce view types. Now if the table schema is `Utf8` then it will produce `Utf8` even when the physical type in the file is `Utf8View`. However a predicate like `lit(Utf8('foo')) = utf8view_col` will be evaluated as all `Utf8View` without casts.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
adriangb: TableProvider implementations should only declare as exact predicates they can completely evaluate during the call to `scan` (which should be replaced with `scan_with_args` btw). Everything else gets resolved later.

crates/core/src/delta_datafusion/session.rs
adriangb: I can very much believe that dictionary arrays are not handled correctly (we've also found plenty of panics in DataFusion because of it) but it would be great to track the issue upstream.

python/src/datafusion.rs
adriangb: IIRC recursive CTEs

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
ethan-tyler: You're right that `ParquetExec` typically yields single file batches and the common case is one output batch. We return `Vec<RecordBatch>` defensively since DF can hand us a multi file batch (e.g., after coalescing), and DV masks + per-file transforms must apply in file context. Implementation does a linear scan for `file_id` run boundaries and zero-copy `slice()` per run with minimal overhead.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Fixed. I moved `file_id` stats to `PartitionedFile` construction time as you suggested. We were double appending `file_id` stats. Now we use `partition_values` and `with_statistics` for per file stats and call `compute_all_files_statistics` with the full table schema.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Removed `strip_view_types_from_schema`. The issue was parquet pushdown breaks when DF produces `Utf8View`/`BinaryView` literals but our read schema is base `Utf8`/`Binary`. DF's `DefaultPhysicalExprAdapter` casts columns but not literals (upstream TODO). We now use DF's adapter plus a literal rewrite (view to base) for pushdown. I also added regression tests in scan/mod.rs.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Renamed. Test covers the literal rewrite for Parquet pushdown when physical schema uses base types.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Intentionally passing base types here. `get_read_plan` builds `parquet_read_schema`, which is used for parquet reads and predicate pushdown evaluation. Pushdown happens before DF's schema casts are applied, if we pass view types but the Parquet files use base types, we hit `Utf8 == Utf8View` mismatches during filter eval. DF coercions also aren't reliably recursive for nested types.

View types are still exposed at the `DeltaScan` boundary via `table_schema` just not in the read/pushdown path. We adapt predicates via `with_expr_adapter` to rewrite view literals to base for pushdown.

If you'd rather pass view types here and handle mismatches differently, open to discussing, this was the least invasive fix I found.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: @adriangb is right - the comment is inaccurate and I updated it. The parquet reader does produce view arrays when the schema requests them. My motivation behind this is pushdown stability as described above.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: Agreed with @adriangb - partition only predicates (fully evaluated during `scan`) now return `Exact`. Everything else returns `Inexact`, including Parquet pushdown predicates (still attempted when safe, but DF retains a post scan Filter as the correctness boundary). Renamed `parquet_pushdown_exact` to `parquet_pushdown_enabled` for clarity.

crates/core/src/delta_datafusion/session.rs
ethan-tyler: I will create an issue and link it.

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: Stats reordering happens in `table_provider.rs` via `FileScanConfigBuilder::with_statistics`. DF52 expects `column_statistics` indexed as `file_schema || partition_columns`, but kernel returns them in `TableConfiguration.schema()` order. We rebuild by name lookup. Scoped to current provider and should be easy to drop when provider is removed.

python/tests/test_table_read.py
ethan-tyler: You're right, this change was an overcorrection and reverted. We still output `StringView` (via `schema_force_view_types=true`). The base types in `parquet_read_schema` are only for the pushdown path; user visible output remains view types.

python/src/datafusion.rs
ethan-tyler: @adriangb is right again :) -`reset_state` is part of DF's `LazyBatchGenerator` contract and  DF calls it when it needs to rerun a scan from the beginning.

For this test generator, we just clone the backing `Vec<RecordBatch>`. For real Python Arrow streams (which can't rewind), `ArrowStreamBatchGenerator::reset_state` returns an `ExhaustedStreamGenerator` that errors with guidance to buffer input if plan re-execution is required.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -199,15 +205,10 @@ async fn get_data_scan_plan(
...
-        f.stats.column_statistics.push(ColumnStatistics {
-            null_count: Precision::Exact(0),
-            max_value: Precision::Exact(file_value.clone()),
-            min_value: Precision::Exact(file_value.clone()),
-            sum_value: Precision::Absent,
-            distinct_count: Precision::Absent,
-        });
+        // NOTE: `PartitionedFile::with_statistics` appends exact stats for partition columns based
+        // on `partition_values`, so partition values must be set first.
+        partitioned_file.partition_values = vec![file_value.clone()];
```
roeap, line 210: So by invterting the order of the assignment we get stats auto-assigned? nice 👍

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -403,9 +422,104 @@ fn cast_record_batch(batch: RecordBatch, target_schema: &SchemaRef) -> Result<Re
...
+        if self.physical_has_view_types {
+            Ok(expr)
+        } else {
+            // DataFusion's DefaultPhysicalExprAdapter inserts casts on columns, but does not
+            // currently rewrite literals to match the physical schema. Parquet predicate
+            // evaluation/pushdown happens against the physical file schema prior to those casts.
+            //
+            // See upstream TODO:
+            // datafusion-physical-expr-adapter/src/schema_rewriter.rs
+            // "TODO: ... move the cast from the column to literal expressions ..." (52.1.0).
```
roeap: this is something we can expect in 52.1, right? i.e. the base optimizer that does this?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -533,39 +648,6 @@ mod tests {
...
-        assert_batches_sorted_eq!(&expected, &batches);
-        assert!(matches!(
-            batches[0].column(1).data_type(),
-            DataType::Utf8View
-        ));
-        assert!(matches!(
-            batches[0].column(2).data_type(),
-            DataType::Utf8View
-        ));
-
```
roeap, line 568: can we not fulfil this test anymore?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
roeap: for the partition columns yes, but especially for the read paths they are interesting? At least that where we are getting most of the benefit i believe?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -275,6 +277,76 @@ impl DeltaScanConfig {
...
+            }
+            DataType::LargeList(inner) => {
+                DataType::LargeList(self.map_field_for_parquet(inner.clone(), partition_cols))
+            }
+            DataType::ListView(inner) => {
+                DataType::ListView(self.map_field_for_parquet(inner.clone(), partition_cols))
+            }
+            // Always use base types for the Parquet read schema.
+            DataType::Utf8View => DataType::Utf8,
+            DataType::BinaryView => DataType::Binary,
```
roeap, line 320: i am so sorry to keep pestering about this - but we especially want them in the parquet read, no?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -374,9 +453,13 @@ fn process_filters(
         .file_column_name
         .as_deref()
         .unwrap_or(FILE_ID_COLUMN_DEFAULT);
+
+    let parquet_pushdown_enabled = scan_config.enable_parquet_pushdown
+        && !config.is_feature_enabled(&TableFeature::RowTracking)
+        && !config.is_feature_enabled(&TableFeature::DeletionVectors);
```
roeap, line 442: oh yes, this would have been a nice way to mess up data without this check!!

crates/core/src/delta_datafusion/table_provider.rs
roeap: but parquet should have no awareness of partition columns. at the very least in the new provider. Partition columns should be 100% handled on the table provider level. 

The only partition column we assign on the parquet level is the file id columns.

At least in my mind the parquet reader would read the file as whatever schema (physical types and odering) we push into it? of course sans any partition column. Where these appear would ultimately be decided in the expression we apply from the scan metadata.

roeap: Come to think of it, could it be that some of the changes to the parquet schema in `KernelScanPlan` are used for this deprecated provider?

If so, can we keep that separate. It just hard to believe that DF would get so much worse in one release 😆.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -453,7 +543,7 @@ fn process_predicate<'a>(
     ProcessedPredicate {
         pushdown: TableProviderFilterPushDown::Inexact,
         kernel_predicate: None,
-        parquet_predicate: Some(expr),
+        parquet_predicate: parquet_pushdown_enabled.then_some(expr),
```
roeap, line 529: do we need to enciode this twice? i.e. we are checking the config parquet pushdown enabled also when we create the scan. so i think its fine we still pass the predicate if we can create it and then make the decision as we set uop the scan intead of here?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -203,7 +203,7 @@ impl KernelScanPlan {
         } else {
             result_schema.clone()
         };
-        let parquet_read_schema = config.physical_arrow_schema(
+        let parquet_read_schema = config.parquet_file_schema(
```
roeap: Sorry for all the noise.

I went through this a bit more comprehensive, and ultimately it's this one line change and its implications that got me worried. This is fairly direct translation from delta kernel and might eventually contain additional metadata columns that need to get passed through that relate to specific features. So any transformation on this is worry-some. (we do transform the physical types as optimization). In any case this should not be processed based on partition columns - the "physical" specifically refers to only being related to physical data. partition columns should have no role at this level.

The idea is that the schema we specify here flows through and gets amended. We do need to look into kernel expression evaluation to fully make sure we don't do additional costly transformations there.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: No, sorry that comment is confusing and adjusted it. Adapter still casts columns but not literals, move cast from column to literal remains a TODO upstream (there was an earlier attempt). We should keep the workaround until there is an upstream fix.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: I don't think `get_read_plan()` should be expected to produce `Utf8View` types. The next
provider forces the Parquet read/predicate schema to base types to keep predicate pushdown stable and view types are exposed only at the DeltaScan boundary.

I updated this by:
- keeping the missing column NULL/backfill assertion in the get_read_plan() test, but dropping the Utf8View type assertions there
- adding a boundary level assertion in KernelScanPlan that result_schema can have view types when enabled, while parquet_read_schema stays base types (crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs)

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Yep - the read paths are actually the main win here. We always dictionary encode the synthetic file-id (file_url) when building PartitionedFile.partition_values and the file-id column itself is Dictionary<UInt16, Utf8>, we avoid repeating the full URL for every row. Partition cols are handled above Parquet via kernel transforms in the next provider and they don’t materially change the Parquet read plan.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: Why base types in parquet_read_schema are needed:

`schema_force_view_types` exists for the perf win, but pushdown/pruning correctness blocks it  in DF 52.1.0. `Utf8 == Utf8View` is a hard error at the Arrow kernel level (arrow-rs #6283), and DF's physical expr adapter has known gaps around literal/cast rewriting (workaround at mod.rs:455-469, tests at :944/:1022).

For delta-rs, `__delta_rs_file_id__` must be `Dictionary<_, Utf8>`
because Arrow can't dictionary-pack to `Utf8View` (plan.rs:232-242, arrow-rs
#9219). View typing the read schema without addressing this reintroduces mixed
base/view predicates on file-id.

Options I see:

1. Keep current behavior: Parquet read schema stays base for
   stable pruning/pushdown and view types applied at DeltaScan boundary.

2. Enable view types with safety shim: Keep file-id as `Dictionary<_, Utf8>`,
   adjust expr adapter to normalize view literals against base columns. Add
   regression test filtering on `__delta_rs_file_id__ = Utf8View(...)`.

3. Enable view types end-to-end: After arrow-rs #9220 ships, switch
   file-id to `Dictionary<_, Utf8View>` and remove mixed schema handling.

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: Agreed Parquet shouldn't know about Delta partition columns. This remap handles
DF's TableSchema/FileScanConfig expectations. DF52 appends partition fields
after the file schema (filled from PartitionedFile.partition_values), and
Statistics.column_statistics is positional which requires name based reordering when
source stats are in TableConfiguration.schema() order. In next, Parquet partition
fields are file_id only; Delta partition columns are handled above Parquet.

KernelScanPlan parquet schema changes stay separate. The remapping fixes a
latent stats-index vs scan-schema-index mismatch that DF52 exposes.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: Agree on single decision point. plan.rs is currently source of truth for
enable_parquet_pushdown and DV/RowTracking gates. mod.rs has an extra RowTracking
guard that doesn't mirror the config/DV checks. Removing the then_some(...) gate
in plan.rs requires moving the full condition to the read plan setup (keeping
has_selection_vectors guard).

Here's the options I can think of:
1. Keep plan.rs gate, drop or convert mod.rs RowTracking check to invariant
2. Always carry predicate, centralize all gating in get_read_plan and remove
    redundant mod.rs check

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
ethan-tyler: Agreed - partition columns shouldn’t influence the Parquet/physical read schema. I updated parquet_read_schema to be derived only from scan.physical_schema() and removed the partition column dependent wrapping from parquet_file_schema. 

The only remaining transformation there is Utf8View/BinaryView -> Utf8/Binary to keep DF predicate pushdown stable and partition columns are still injected above Parquet via the kernel transform path. This should also be safer if kernel later adds additional metadata columns to the physical schema, since we’re no longer keying off partition columns at this level.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
adriangb: PhysicalExprSimplifier moves the casts from columns to literals. Maybe look at BatchAdapter (either to use it or as inspo; it’s in main and will be in 52.1.0 not released yet)?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
adriangb: Note for completeness: the Inexact filters may be removed by the filter pushdown physical optimizer if the data source accepts them as exact…

crates/core/src/delta_datafusion/table_provider.rs
adriangb: > but parquet should have no awareness of partition columns. at the very least in the new provider. Partition columns should be 100% handled on the table provider level

`ParquetOpener` *does* does have awareness of DataFusion's partition columns. Before we made that change filters like `part_col = 1 OR file_col = 2` would give wrong results when filter Pushdown was enabled. Not sure if Delta partition columns are handled completely separately...

ethan-tyler: @adriangb - The `ParquetOpener` partition handling you're referencing (`part_col = 1 OR file_col = 2` correctness) applies to DF's built in partitioning. Next-scan sidesteps this entirely- Delta partition columns are handled outside DF's partition machinery:

1. Parquet layer: only sees `file_id` (provider metadata). No Delta partition columns in `parquet_read_schema`.
2. Value injection: kernel transforms add partition values post read, bypassing `FileScanConfig.partition_values`.
3. Filter pushdown: partition predicates use kernel's stats based file skipping, not `ParquetOpener`'s partition aware path.

Kernel owns partition semantics and injects values above the scan.

The stats remapping in the legacy provider exists because it *does* use DF's partition column machinery and must align with DF52's `file_schema + partition_columns` index expectations. This will go away once the legacy provider is removed.

PR #4096, opened 2026-01-19
feat: improve logical planning and migrate update op

roeap:
> [!IMPORTANT]
> Please challenge the bold claims below when reviewing.

# Description

This PR brings us some large steps closer to converging on the new `TableProvider` to support modern features in all operations. While I believe this is a strong foundation for further performance improvements and should yield improved performance in many scenarios, there are cases that will see performance degradation - at least for a while.

As a first of the "file-aware" operations we update the `UPDATE` operation to make use of the new `DeltaScan`. We do plan based on file origin whenever we need to re-write data (MERGE, OPTIMIZE, DELETE, ...). In these cases we do not only want to rely on best-effort file skipping for only re-writes files with concrete matches. For this we have been using `find_files` which produces a collection of files to scan. The current `DeltatableProvider` allows hardcoding the files to scan. We did not want to apply the same logic in the new provider because:
* It's hard (and ugly) to re-wire our internal Add actions into the kernel file metadata replay to support DVs etc. 
* using Add actions directly blocs us from using more logical planning.

One important aspect of this is that if we have a match in a file, we do want to read ALL the data from the file, since we are rewriting it to new files.

The relevance of this is PR is then that we establish an updated way to integrate this logic. Specifically we no longer specify the refined file list (those that definitely contain data matching a predicate) via `Add` actions but via the following approach.
* we push the initial predicate (now called `skipping_predicate`) into the kernel scan, since we just need to consider files that MIGHT contain data.
  * this predicate does NOT get pushed into the parquet scan
* We identify all files that contain relevant data via the same aggregate query we perform in `find_files`.
* We create an updated predicate `<file id column> IS IN [.. valid files ..]` which we push into the parquet scan when we are re-writing the data.
  * since `<file id column>` is a metadata-only column we don't need to materialize data to evaluate the predicate.

In follow-ups I'll be applying the same logic to the missing file-aware ops.

> [!IMPORTANT]
> To create the delete actions we are now performing a new log replay which may incur duplicate work. I'll be looking into how we can re-use a scoped (with predicate applied) log replay within the scope of an operation as a follow-up.

### DIFF
--- a/python/src/merge.rs
+++ b/python/src/merge.rs
@@ -2,7 +2,8 @@ use deltalake::arrow::datatypes::Schema as ArrowSchema;
 use deltalake::datafusion::catalog::TableProvider;
 use deltalake::datafusion::datasource::MemTable;
 use deltalake::datafusion::physical_plan::memory::LazyBatchGenerator;
-use deltalake::datafusion::prelude::SessionContext;
+use deltalake::delta_datafusion::create_session;
+use deltalake::delta_datafusion::DeltaSessionContext;
 use deltalake::kernel::EagerSnapshot;
 use deltalake::logstore::LogStoreRef;
 use deltalake::operations::merge::MergeBuilder;
@@ -54,7 +55,7 @@ impl PyMergeBuilder {
         commit_properties: Option<PyCommitProperties>,
         custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
     ) -> DeltaResult<Self> {
-        let ctx = SessionContext::new();
+        let ctx = create_session().into_inner();
 
         let source = source
             .into_reader()
--- a/python/tests/test_merge.py
+++ b/python/tests/test_merge.py
@@ -2058,6 +2058,65 @@ def test_merge_field_special_characters_delete_2438(tmp_path: pathlib.Path):
     ) == expected
 
 
+@pytest.mark.parametrize("streaming", (True, False))
+def test_merge_preserves_casing_in_quoted_identifiers(
+    tmp_path: pathlib.Path, streaming: bool
+):
+    data = Table(
+        {
+            "fooBar": Array(
+                ["A001", "A002"],
+                ArrowField("fooBar", type=DataType.string_view(), nullable=True),
+            )
+        }
+    )
+
+    write_deltalake(tmp_path, data, mode="append")
+
+    dt = DeltaTable(tmp_path)
+    source_table = Table(
+        {
+            "fooBar": Array(
+                ["A003"],
+                ArrowField("fooBar", type=DataType.string_view(), nullable=True),
+            )
+        }
+    )
+
+    (
+        dt.merge(
+            source=source_table,
+            predicate='target."fooBar" = source."fooBar"',
+            source_alias="source",
+            target_alias="target",
+            streamed_exec=streaming,
+        )
+        .when_not_matched_insert_all()
+        .execute()
+    )
+
+    expected = Table(
+        {
+            "fooBar": Array(
+                ["A001", "A002", "A003"],
+                ArrowField("fooBar", type=DataType.string_view(), nullable=True),
+            )
+        }
+    )
+
+    result = (
+        QueryBuilder()
+        .register("tbl", dt)
+        .execute('select * from tbl order by "fooBar" asc')
+        .read_all()
+    )
+
+    last_action = dt.history(1)[0]
+
+    assert last_action["operation"] == "MERGE"
+    assert result == expected
+
+
 @pytest.mark.pandas
 @pytest.mark.pyarrow
 def test_struct_casting(tmp_path: pathlib.Path):


crates/core/src/delta_datafusion/session.rs
```diff
@@ -19,19 +17,6 @@ pub fn create_session() -> DeltaSessionContext {
...
-#[deprecated(
-    since = "0.29.1",
-    note = "Stop gap to get rid of all explicit session state references"
-)]
-pub(crate) fn session_state_from_session(session: &dyn Session) -> DataFusionResult<&SessionState> {
-    session
-        .as_any()
-        .downcast_ref::<SessionState>()
-        .ok_or_else(|| exec_datafusion_err!("Failed to downcast Session to SessionState"))
-}
```
roeap, line 33: 🔥

crates/core/src/kernel/scalars.rs
```diff
@@ -251,6 +252,32 @@ impl ScalarExt for Scalar {
                     StructData::try_new(struct_fields, values).ok()?,
                 ))
             }
+            Dictionary(kt, dt) if matches!(kt.as_ref(), UInt16) => {
```
roeap, line 255: @ethan-tyler this, or rather the lack of this MAY be one of the causes for the need to do so much wrapping etc. in the df52 PR.

crates/core/src/operations/update.rs
```diff
@@ -272,137 +283,138 @@ async fn execute(
...
+
+    // Scan the delta table with a dedicated predicate applied for file skipping
+    // and with the source file path exosed as column.
+    let table_source = provider_as_source(
+        DeltaScanNext::builder()
+            .with_eager_snapshot(snapshot.clone())
+            .with_file_skipping_predicates(skipping_pred.clone())
+            .with_file_column(FILE_ID_COLUMN_DEFAULT)
+            .await?,
+    );
```
roeap, line 316: This is where our new `find_files` logic begins with a scan that has a static file skipping predicate for kernel scans. and exposes the file id column.

crates/core/src/operations/update.rs
```diff
@@ -272,137 +283,138 @@ async fn execute(
...
+                .iter()
+                .flatten()
+                .map(|s| s.to_string())
+                .collect_vec()
+        })
+        .try_collect::<Vec<_>>()
+        .await?
+        .into_iter()
+        .flatten()
+        .collect();
```
roeap, line 350: we apply the same logic as in find find files, to check the actual data for predicate matches to the the exact files we need to process. Since we are running this through the can, we get DV etc. applied here.

crates/core/src/operations/update.rs
```diff
@@ -272,137 +283,131 @@ async fn execute(
...
+    // Run a table scan limiting the data to that originating from valid files.
+    let file_list = valid_files
+        .iter()
+        .cloned()
+        .map(|v| lit(wrap_partition_value_in_dict(ScalarValue::Utf8(Some(v)))))
+        .collect_vec();
+    let plan = LogicalPlanBuilder::scan("target", table_source, None)?
+        .filter(col(FILE_ID_COLUMN_DEFAULT).in_list(file_list, false))?
+        .project_away([FILE_ID_COLUMN_DEFAULT])?
+        .build()?;
```
roeap, line 365: here we convert the valid files into a predicate on the file id column, which will get pushed into the parquet scan. Since the file column is defined as partition column, we should not need to materialise any data to evaluate that predicate in the parquet scan.

@alamb - not sure if you can tell from the limited context here, but in case you can, would this be correct? :)

crates/core/src/operations/update.rs
```diff
@@ -420,52 +425,50 @@ async fn execute(
...
+                .clone()
+                .join(f.path_raw())
+                .map_err(|e| exec_datafusion_err!("{e}"))?;
+            valid.contains(url.as_ref());
+            Ok(valid
+                .contains(url.as_ref())
+                .then(|| Action::Remove(f.remove_action(true))))
+        })
+        .try_collect()
+        .await?;
```
roeap, line 458: TO get the remove actions we run a new log replay here where we agin filter against the valid files.

@rtyler - for situations like this it would be great to have some sort of replay-from existing data logic included to not pay another repaly.

Wirign things through from prior scans is another route we can go.. 

I'll do some analysis on actual cases to see how we can make sure we are not doing excessive work.

crates/core/src/delta_datafusion/logical.rs
```diff
@@ -1,8 +1,83 @@
...
+        let mut projection = self
+            .schema()
+            .iter()
+            .map(|(_, f)| col(Column::from_name(f.col_name())))
+            .collect_vec();
+        projection.push(expr.alias(name));
+        self.project(projection)
+    }
+
+    fn project_away(self, cols: impl IntoIterator<Item = impl ColumnReference>) -> Result<Self> {
```
ion-elgreco: So `drop`?

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -654,6 +658,68 @@ impl TableProviderBuilder {
         self.file_column = Some(file_column.to_string());
         self
     }
+
+    /// Add predicates applied only during file skipping.
```
ion-elgreco, line 661: The old table provider has this for merge, but there it's just a flag to disable pushdown into parquet. Do you think it's necessary on expr level to distinguish it?

crates/core/src/delta_datafusion/utils.rs
```diff
@@ -0,0 +1,86 @@
...
+#[derive(Default)]
+pub(crate) struct DeltaScanVisitor {
+    pub(crate) delta_plan: Option<KernelScanPlan>,
+}
+
+impl DeltaScanVisitor {
+    fn pre_visit_delta_scan(&mut self, delta_scan_exec: &DeltaScanExec) -> Result<bool> {
+        // let Some(_metrics) = delta_scan_exec.metrics() else {
+        //     return Ok(true);
+        // };
```
ion-elgreco: Needs to be removed?

crates/core/src/operations/write/configs.rs
```diff
@@ -20,4 +24,15 @@ impl WriterStatsConfig {
             stats_columns,
         }
     }
+
+    pub fn from_config(config: &TableConfiguration) -> Self {
```
ion-elgreco, line 28: 🤘

crates/core/src/delta_datafusion/table_provider.rs
roeap: This is intentionally a `pub(crate)` API since its most useful when used correctly :). Having it on the expression level helped integrate this into the scan planning in `KernelScanPlan` more seamlessly.

For our internal ops planning we now actually want predicates to be pushed down into the scan, just the right one...

its essentially a replacement for the `with_files` api on the old provider.

crates/core/src/delta_datafusion/logical.rs
roeap: ii guess you could call it that 🤣

ethan-tyler:
> @ethan-tyler sorry for leaving you hanging on the DF 52 PR. This work here is the major source of my concern. @alamb pointed out, a lot of the type coercion issues we have been seeing should not appear if things properly flow through datafusions planning. The migration to the new table provider and the associated refactoring of the operations aims to achieve exactly that.
> 
> In the hopes of not being too demanding, could we try and land this prior to getting the DF 52 stuff in the hopes of addressing some of the sources that warranted larger refactors?

@roeap - No worries at all, I know these things take time to review! I agree, if predicates flow through DF’s planning and adaptation pipeline properly, the coercion issues should resolve naturally.

This is the cleaner path and lets us rely on DF’s adapter logic. I’d rather fix it properly here than carry workarounds in the DF52 PR.

More than happy to help get this into shape, appreciate including me. I’ll review this afternoon (dad duty calls first).

crates/core/src/delta_datafusion/logical.rs
ion-elgreco: Haha let's call it that, bit more clear

crates/core/src/delta_datafusion/table_provider.rs
ion-elgreco: I see that makes sense

ion-elgreco: I see that makes sense

crates/core/src/kernel/scalars.rs
ethan-tyler: Agreed. Would be good to add a few tests around dictionary edge cases (nulls, empty, unexpected value types) since we're relying on UInt16 conversions in several places now.

crates/core/src/operations/update.rs
ethan-tyler: Design looks right, but worth we add a regression test asserting file_id IN (...) is treated as partition pruning and not pushed as a Parquet row predicate. Keeps us safe across DF versions.

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: I think the pub(crate) is fine. However my concern is what's the contract when an expr isn't kernel-convertible?

I would treat skipping as best-effort optimization. If conversion fails, drop that term (skip fewer files), log at debug level. We should avoid skipping more aggressive than the original predicate.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -101,13 +101,23 @@ pub(crate) async fn find_files(
     }
 }
 
-struct FindFilesExprProperties {
+pub(crate) struct FindFilesExprProperties {
```
ethan-tyler, line 104: Consider keeping fields private and expose a helper like `validate_file_skipping_expr(&Expr) -> DeltaResult<()> `or read-only accessors. Avoids accidental coupling callers to internals like partition_only.

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -654,6 +658,68 @@ impl TableProviderBuilder {
...
+                ));
+                }
+            }
+        };
+
+        let mut provider = next::DeltaScan::new(snapshot, config)?;
+        if let Some(skipping) = self.file_skipping_predicates {
+            // validate that the expressions contain no illegal variants
+            // that are not eligible for file skipping, e.g. volatile functions.
+            let _valid: Vec<_> = skipping
```
ethan-tyler: nit: This allocates Vec just to validate and discard. Something like `skipping.iter().try_for_each(...) `avoids the allocation. Ensure errors are user-facing (include function name, volatility, etc).

crates/core/src/operations/update.rs
```diff
@@ -272,137 +283,131 @@ async fn execute(
...
-        .build(&snapshot)?;
+    // the kernel scan only provides a best effort file skipping, in this case
+    // we want to determine the file we certainly need to rewrite. For this
+    // we perform an initial aggreagte scan to see if we can quickly find
+    // at least one matching record in the files.
+    let files_plan = LogicalPlanBuilder::scan("scan", table_source.clone(), None)?
+        .filter(predicate.clone())?
+        .aggregate(
+            [col(FILE_ID_COLUMN_DEFAULT)],
+            [first_value(lit(1_i32), vec![])],
```
ethan-tyler, line 326: Is `first_value(lit(1), …)` here for optimizer reasons or just "one row per file"?

crates/core/src/operations/update.rs
```diff
@@ -272,137 +283,131 @@ async fn execute(
...
+        )?
+        .build()?;
+    let files_exec = session.create_physical_plan(&files_plan).await?;
+    let valid_files: HashSet<_> = execute_stream(files_exec, session.task_ctx())?
+        .map_ok(|f| {
+            let dict_arr = f.column(0).as_dictionary::<UInt16Type>();
+            let typed_dict = dict_arr.downcast_dict::<StringArray>().unwrap();
```
ethan-tyler: This unwrap() can panic from user input. The hardcoded UInt16 key type could break silently or panics if upstream changes to Dictionary(Int32, Utf8) or plain Utf8.

I would accept Utf8 and Dictionary(_, Utf8), return error with observed type on mismatch.

crates/core/src/operations/update.rs
```diff
@@ -420,52 +425,48 @@ async fn execute(
...
+            root_url,
+            Arc::new(valid_files),
+        ))))
+        .map(|(f, u)| f.map(|f| (f, u)))
+        .try_filter_map(|(f, (root, valid))| async move {
+            let url = root
+                .clone()
+                .join(f.path_raw())
+                .map_err(|e| exec_datafusion_err!("{e}"))?;
+            let is_valid = valid.contains(url.as_ref());
```
ethan-tyler, line 454: Membership relies on `f.file_url.to_string()` == `table_root.join(f.path_raw()).as_ref()`. 
These can diverge on encoding/normalization (see :189 re double-encoding).

I would add a regression test with a path requiring encoding, or compare canonical components instead of raw strings. Mismatch silently skips removes.

crates/core/src/delta_datafusion/find_files.rs
roeap: This is a good point, Personally I like using the visitors directly and we do so in several places (rewriting schemas and expression rewrites). That said, making sure we are not exposing too much is a good thing :).

I am currently working on migrating the remaining ops, where these APIs (find_files-related) are in the center. I'll keep an eye out when doing this.

crates/core/src/delta_datafusion/logical.rs
roeap: I like the `project_away` name b/c it reminds me of a language I liked - but you are right and i renamed it :)>

crates/core/src/delta_datafusion/table_provider.rs
roeap: This is indeed the case. File skipping during log replay has to always be considered best effort. This is just to reduce the work we need to do finding the actual matches.

> We should avoid skipping more aggressive than the original predicate.

This is very true of course, since otherwise we would be producing bad results.

We might be able to also use the original predicate to do more file-level skipping before setting up the scan, I would hold off with that though until we can prove it helps us.

crates/core/src/delta_datafusion/table_provider.rs
roeap: good point. updated to use a simple for loop.

crates/core/src/delta_datafusion/utils.rs
roeap: gone!

crates/core/src/kernel/scalars.rs
roeap: Right now a bunch of code should run through this.

I do have more PRs queued up, that will improve the kernel implementations and improve our coverage both w.r.t. to covered logic as well as test coverage.

crates/core/src/operations/update.rs
roeap: we really just want the quickest row we can get for every file that contains matched data. Not even that - the quickest indication that such a row exists.

I thought `first_value` without any ordering would be the cheapest, but more that happy to update to a better way.

crates/core/src/operations/update.rs
roeap: unwraps are a big no-no - agreed :). Updated this to manage the error.

In this case we are not processing user input though. The file id column on this code path is hidden from use space and projected away before we expose any of this data to outside code.

crates/core/src/operations/update.rs
roeap: In this case the column os not part of the physical data could that even happen then?

After we are consolidated I hope to dive a bit deeper into some of this, but I think we would have to set up a failing store that checks for the skipped file path.

While worthwhile, I hope to defer this to a later point in time :).

crates/core/src/operations/update.rs
roeap: Well, you found it - the HACK that hides in the code 🤣.

`path_raw` was introduced earlier specifically for the purpose of joining with the table URL. It's also `pub(crate)` as to not spill out.

Btw. this will also break if we track full URLs in the add action. This of course is a problem ...

The proper way - IMO - to solve this is to normalize the URLs as we receive the data in the inner streams. in fact we do some processing there already to handle stats parsing.

SO getting rid of this is very much on my agenda. Ultimately we need to do the same thing as when we are defining the partition value on the partitioned file, which right now is this. but we also need to make sure that this does not diverge.

crates/core/src/operations/update.rs
ethan-tyler: Fair question. `FILE_ID_COLUMN_DEFAULT` isn't a physical Parquet column, but it's an Arrow array flowing through DataFusion, and DF/Arrow can choose different representations over time (`Utf8`, `Dictionary(Int32, Utf8)`, `Utf8View`, etc.).

I actually got the idea for this working on the DF52 upgrade and opened a PR (apache/arrow-rs#9220). It adds support for packing/casting to `Dictionary(_, Utf8View/BinaryView)`. If this lands, hardcoded assumptions like `downcast_dict::<StringArray>()` become more fragile, even for metadata-only columns.

I agree the regression test is non trivial to assert pruning behavior without observing which files were opened. A wrapper store that records get/head calls is probably the cleanest approach. Fine deferring and can open an issue so it doesn't get lost.

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: Agreed - skipping during log replay is best-effort.

IMO to make the contract more concrete - When a skipping term isn't kernel-convertible, I'd prefer dropping that term (scan more files) rather than erroring. Preserves best-effort semantics.

Totally good on deferring "use original predicate for more pre-scan skipping" until we can benchmark it. Happy to open a follow up issue for this if not already tracked.

crates/core/src/operations/update.rs
ethan-tyler: That intent makes sense ("files with ≥1 match"). But `first_value(..)` won't short-circuit per group and still flows through hash aggregation and processes all matching rows.

The clearest DF-native way is `SELECT DISTINCT file_id`: `filter(predicate) -> project([file_id]) -> distinct()`. Avoids the dummy `first_value(lit(1_i32), vec![])` and makes intent obvious, same underlying plan shape.
```rust
let files_plan = LogicalPlanBuilder::scan("target", table_source, None)?
    .filter(predicate)?
    .project(vec![col(FILE_ID_COLUMN_DEFAULT)])?
    .distinct()?
    .build()?;
```

I can add this as a follow up if agreed.

PR #4098, opened 2026-01-19
fix: properly simplify delete predicate expressions for Datafusion

### LINKED ISSUE #4093: [Bug]: arrow_cast not not using ExprSimplifier
### What happened?

See https://github.com/apache/datafusion/issues/19866#issuecomment-3765438348 for details

```bash
  File "demo.py", line 64, in <module>
    main()
  File "demo.py", line 60, in main
    table.delete(predicate=predicate)
  File ".venv/lib/python3.10/site-packages/deltalake/table.py", line 1112, in delete
    metrics = self._table.delete(
_internal.DeltaError: Generic DeltaTable error: Internal error: arrow_cast should have been simplified to cast.
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/datafusion/issues

```

### Expected behavior

use `ExprSimplifier` and simplify the `arrow_cast` to `cast`

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

see https://github.com/apache/datafusion/issues/19866#issuecomment-3765438348

### Relevant logs

```shell

```

rtyler:
Fixes #4093

### DIFF
--- a/crates/core/src/operations/delete.rs
+++ b/crates/core/src/operations/delete.rs
@@ -33,20 +33,24 @@
 //! # })
 //! ````
 
+use arrow_schema::Schema;
 use async_trait::async_trait;
 use datafusion::catalog::Session;
-use datafusion::common::ScalarValue;
+use datafusion::common::{DFSchema, ScalarValue};
 use datafusion::datasource::provider_as_source;
 use datafusion::error::Result as DataFusionResult;
-use datafusion::execution::context::SessionState;
+use datafusion::execution::context::{ExecutionProps, SessionState};
+use datafusion::logical_expr::simplify::SimplifyContext;
 use datafusion::logical_expr::{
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/delete.rs
```diff
@@ -207,12 +211,19 @@ impl std::future::IntoFuture for DeleteBuilder {
...
-                },
+                Some(predicate) => {
+                    let expr = match predicate {
+                        Expression::DataFusion(expr) => expr,
+                        Expression::String(s) => {
+                            snapshot.parse_predicate_expression(s, session.as_ref())?
+                        }
+                    };
+                    let df_schema = DFSchema::try_from(snapshot.arrow_schema())?;
+                    let props = ExecutionProps::new();
```
roeap: we need to get the execution props from an active session. otherwise we might get inconsistent values for `now()` expressions that get optimised into constants ...

Issue #4097, opened 2026-01-19
[Bug]: Python merge operation ignores enable_ident_normalization setting, causing case-sensitive column name errors


bellshun:
### Environment:

deltalake version: 1.3.2
Python version: 3.11+
OS: Windows
Bug Summary:
The Python DeltaTable.merge() operation fails when using CamelCase column names (e.g., hogeNo), even though Rust's merge implementation correctly handles them by using DeltaSessionContext with enable_ident_normalization = false.

### Error Message:

> Generic DeltaTable error: Schema error: No field named hogeno. Column names are case sensitive. 
You can use double quotes to refer to the "hogeno" column or set the 
datafusion.sql_parser.enable_ident_normalization configuration. Did you mean 'hogeNo'?
Root Cause:
In the Python bindings (python/src/lib.rs), the merge operation uses the standard SessionContext::new() instead of DeltaSessionContext::default().

The standard SessionContext has enable_ident_normalization = true by default, which normalizes identifiers to lowercase. This contradicts the Rust implementation where DeltaSessionContext explicitly sets enable_ident_normalization = false in crates/core/src/delta_datafusion/session.rs:

```
impl Default for DeltaSessionConfig {
    fn default() -> Self {
        DeltaSessionConfig {
            inner: SessionConfig::default()
                .set_bool(
                    "datafusion.sql_parser.enable_ident_normalization", 
                    false
                ),
        }
    }
}
```

However, in the Python bindings, a standard SessionContext is created:

```
let ctx = SessionContext::new();
Reproduction:


from deltalake import DeltaTable, write_deltalake
import pyarrow as pa

# Create table with CamelCase column
schema = pa.schema([('hogeNo', pa.string())])
data = pa.table({'hogeNo': ['A001', 'A002']})
write_deltalake('/tmp/test_table', data)

# Try to merge
dt = DeltaTable('/tmp/test_table')
new_data = pa.table({'hogeNo': ['A003']})

dt.merge(
    source=new_data,
    predicate='target."hogeNo" = source."hogeNo"',
    source_alias="source",
    target_alias="target"
).when_not_matched_insert_all().execute()
# -> Fails with "No field named hogeno"
```

### Related Issues:

https://github.com/delta-io/delta-rs/issues/1797
https://github.com/delta-io/delta-rs/issues/1954

### Expected behavior

The merge operation should preserve column name casing, consistent with the Rust implementation.

### Suggested Fix:
[...]

PR #4102, opened 2026-01-20
fix(python): disable ident normalization in merge

bellshun:
fix(python): disable ident normalization in merge

Use DeltaSessionContext in Python merge bindings to preserve column case and match Rust merge behavior (enable_ident_normalization = false). 

Fixes https://github.com/delta-io/delta-rs/issues/4097

### DIFF
--- a/python/src/merge.rs
+++ b/python/src/merge.rs
@@ -55,7 +55,7 @@ impl PyMergeBuilder {
         commit_properties: Option<PyCommitProperties>,
         custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
     ) -> DeltaResult<Self> {
-        let ctx: SessionContext = create_session().into();
+        let ctx = create_session().into_inner();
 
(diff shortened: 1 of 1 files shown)


python/src/merge.rs
```diff
@@ -54,7 +55,7 @@ impl PyMergeBuilder {
         commit_properties: Option<PyCommitProperties>,
         custom_execute_handler: Option<Arc<dyn CustomExecuteHandler>>,
     ) -> DeltaResult<Self> {
-        let ctx = SessionContext::new();
+        let ctx: SessionContext = DeltaSessionContext::default().into();
```
roeap: Do you mind updating this to use the `create_session()` method. Right now these are equivalent, but I am planning on integrating a bit more logic there.

bellshun: hello, @roeap.
Sure — I updated it to use create_session(). Thanks for the heads-up.

PR #4106, opened 2026-01-20
feat: centralize predicate parsing with literal coercion

roeap:
# Description

We recently landed a fix to simplify predicates before using them... This PR consolidates that logic and adds another fix we requires in follow ups to properly coerce literals.

This is since we are also translating the predicate to kernel where no coercion happens.

### DIFF
--- a/crates/core/src/operations/write/execution.rs
+++ b/crates/core/src/operations/write/execution.rs
@@ -1,18 +1,16 @@
 use std::sync::{Arc, OnceLock};
 use std::vec;
 
-use arrow::compute::concat_batches;
 use arrow::datatypes::Schema;
 use arrow_array::RecordBatch;
 use datafusion::catalog::{Session, TableProvider};
 use datafusion::common::ToDFSchema;
 use datafusion::datasource::{MemTable, provider_as_source};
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/merge/mod.rs
```diff
@@ -2972,12 +2969,6 @@ mod tests {
         let last_commit = table.last_commit().await.unwrap();
         let parameters = last_commit.operation_parameters.clone().unwrap();
         assert!(!parameters.contains_key("predicate"));
-        assert_eq!(
-            parameters["mergePredicate"],
-            "target.id = source.id AND \
-            target.id IN (source.id, source.modified, source.value) AND \
-            target.modified IN ('2021-02-02')"
-        );
```
roeap, line 2980: since we are re-writing things now, checking against a specific string is brittle ...

ion-elgreco: Is it though, it's still deterministic, right?

roeap: within a version of DF yes.. across versions things might change.

PR #4111, opened 2026-01-21
refactor: move files scan to separate function

roeap:
## Description

When refactoring the UPDATE operation, we also introduced a new way to find files
containing data which matches a predicate - i.e. our current find_files function.
In this PR we factor this out into a separate function for reuse in other operations.

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
@@ -4,14 +4,15 @@
 //! and applies Delta Lake protocol transformations to produce logical table data.
 
 use std::any::Any;
+use std::collections::VecDeque;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
-use arrow::array::{ArrayAccessor, AsArray, RecordBatch, StringArray};
+use arrow::array::{ArrayAccessor, RecordBatch, StringArray};
 use arrow::compute::filter_record_batch;
 use arrow::datatypes::{SchemaRef, UInt16Type};
-use arrow_array::BooleanArray;
+use arrow_array::{Array, BooleanArray};
 use dashmap::DashMap;
 use datafusion::common::config::ConfigOptions;
 use datafusion::common::error::{DataFusionError, Result};
@@ -283,6 +284,7 @@ impl ExecutionPlan for DeltaScanExec {
             selection_vectors: Arc::clone(&self.selection_vectors),
             file_id_column: self.file_id_column.clone(),
             return_file_ids: self.retain_file_ids,
+            pending: VecDeque::new(),
         }))
     }
 
@@ -342,8 +344,8 @@ impl ExecutionPlan for DeltaScanExec {
 /// - Schema projection: Projects to requested columns only
 /// - Type casting: Ensures output matches expected logical schema
 ///
-/// Each batch is processed independently, with transformations looked up by file ID
-/// from the batch's partition values.
+/// Input batches may contain rows from multiple file IDs (e.g., due to upstream coalescing).
+/// The stream splits such batches by contiguous file-id runs and applies per-file transforms.
 struct DeltaScanStream {
     scan_plan: Arc<KernelScanPlan>,
     /// Kernel data type for the data after transformations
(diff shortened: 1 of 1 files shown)


crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,160 @@ fn df_logical_schema(
...
+    // we want to determine the file we certainly need to rewrite. For this
+    // we perform an initial aggreagte scan to see if we can quickly find
+    // at least one matching record in the files.
+    let files_plan = LogicalPlanBuilder::scan("files_scan", table_source.clone(), None)?
+        .filter(predicate.clone())?
+        .project([
+            cast(col(FILE_ID_COLUMN_DEFAULT), DataType::Utf8View).alias(FILE_ID_COLUMN_DEFAULT)
+        ])?
+        .distinct()?
+        .build()?;
```
roeap, line 503: @ethan-tyler - I did see your comment on the update PR and took the liberty of integrating that here.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,160 @@ fn df_logical_schema(
...
+    //     files_plan.schema().inner().clone(),
+    //     files_data,
+    // )?));
+    // let files_plan = LogicalPlanBuilder::scan("files_scan", files_source, None)?.build()?;
+    // let join_expr = col(Column::new(Some("scan"), FILE_ID_COLUMN_DEFAULT))
+    //     .eq(col(Column::new(Some("files_scan"), FILE_ID_COLUMN_DEFAULT)));
+    // let plan = LogicalPlanBuilder::scan("scan", table_source, None)?
+    //     .join_on(files_plan, JoinType::LeftAnti, [join_expr])?
+    //     .drop_columns([FILE_ID_COLUMN_DEFAULT])?
+    //     .build()?;
```
roeap: @ethan-tyler - also took a look at you comment regarding the use of `in_list`, which I agree will not necessarily scale. This was an attempt at something better, but did not work out of the box.

This might be a good thing for a follow up?

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,160 @@ fn df_logical_schema(
...
+
```
rtyler:
```suggestion
```

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,160 @@ fn df_logical_schema(
...
+    // Crate a table scan limiting the data to that originating from valid files.
+    let file_list = valid_files
+        .iter()
+        .flat_map(|arr| arr.iter().flatten().map(|v| v.to_string()))
+        .map(|v| lit(wrap_partition_value_in_dict(ScalarValue::Utf8(Some(v)))))
+        .collect_vec();
```
rtyler, line 525: Nobody loves an iter chain like Robbie Pack

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,160 @@ fn df_logical_schema(
...
+}
+
+impl MatchedFilesScan {
+    pub(crate) fn scan(&self) -> &LogicalPlan {
+        &self.plan
+    }
+
+    /// Create a HashSet containing all files included in this scan.
+    ///
+    /// Files are referenced using fully qualified URLs.
```
rtyler, line 423: I don't believe this comment. :laughing:  I was looking at how `valid_files` is produced below, and that's just coming off a string view from the underlying batch from the kernel scan? How do we enforce that any of this is an actual `URL`?

roeap: This data does not originate in kernel. We add a virtual column in the parquet scan with a normalized URL .. its this file id column we get the data from.

roeap:
> I pushed a commit to prune commented out code and a warning that I saw around an unused struct member 😄

That is needed in follow ups, but no issue at all to add it back once we do :).

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,141 @@ fn df_logical_schema(
...
+    pub(crate) fn files_set(&self) -> HashSet<String> {
+        self.valid_files
+            .iter()
+            .flat_map(|arr| arr.iter().flatten().map(|v| v.to_string()))
```
hntd187, line 427: flat map the flattened map? how many levels deep does this go? This line is a bit confusing.

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,141 @@ fn df_logical_schema(
...
+) -> Result<Option<MatchedFilesScan>> {
+    let skipping_pred = simplify_predicates(split_conjunction_owned(predicate))?;
+
+    let partition_columns = snapshot
+        .table_configuration()
+        .metadata()
+        .partition_columns();
+    // validate that the expressions contain no illegal variants
+    // that are not eligible for file skipping, e.g. volatile functions.
+    for term in &skipping_pred {
```
hntd187, line 461: could you reuse the object here and just reset the result on every iteration? so you don't clone partition columns on every iteration

crates/core/src/delta_datafusion/find_files.rs
ethan-tyler: Thanks for carrying that forward, this is exactly what I was hoping for here!

crates/core/src/delta_datafusion/find_files.rs
ethan-tyler: LGTM for the refactor. I saw you tried the MemTable and join approach. What specifically didn't work? Planning limitation, schema mismatch, DF behavior?

For a follow up, execution-time HashSet filtering in the scan exec layer might be simpler than plan-time `IN (...)`, especially since `split_by_file_id_runs()` infrastructure already exists. Happy to open a issue with options, lmk your thoughts.

roeap: @ethan-tyler - an issue with options would be great ... not sure and it might even work and I had some other bugs around during refactor - many tests did pass :).

But if there is an issue, it might be a good opportunity to think on the best approach, as I know there are even more variants to do in-list logic :).

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -385,3 +397,141 @@ fn df_logical_schema(
...
+
+impl MatchedFilesScan {
+    pub(crate) fn scan(&self) -> &LogicalPlan {
+        &self.plan
+    }
+
+    /// Create a HashSet containing all files included in this scan.
+    ///
+    /// Files are referenced using fully qualified URLs.
+    pub(crate) fn files_set(&self) -> HashSet<String> {
```
ethan-tyler, line 424: update.rs:373 compares these strings against `root.join(f.path_raw()).as_ref()`; if representations diverge, removes can be silently skipped. Might be worth a small regression test here.

crates/core/src/delta_datafusion/find_files.rs
roeap: good point, this will undergo some more changes as we move along. 

With your approach its also easier to just borrow the DVs. Is it OK if I add that in a follow up?

crates/core/src/delta_datafusion/find_files.rs
roeap: Currently working on a follow up, where we consolidate this logic (the whole remove replay part); Ok if I add the additional testing there?

ethan-tyler: Yep, that’s fine by me :)

crates/core/src/delta_datafusion/find_files.rs
ethan-tyler: Opened - #4113

PR #4112, opened 2026-01-21
fix(datafusion): handle coalesced multi-file batches in next-scan

ethan-tyler:
# Description

Fix next-scan execution when upstream coalescing produces batches with rows from multiple files.

Changes:
- Split incoming batches into contiguous file_id runs before applying DV masks/transforms
- Buffer fan-out outputs via VecDeque to preserve row order
- Return `internal_datafusion_err!` on unexpected file_id column type instead of panicking
- Add tests for interleaved file IDs, fanout, and invalid/null file_id paths

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
@@ -4,14 +4,15 @@
 //! and applies Delta Lake protocol transformations to produce logical table data.
 
 use std::any::Any;
+use std::collections::VecDeque;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
-use arrow::array::{ArrayAccessor, AsArray, RecordBatch, StringArray};
+use arrow::array::{ArrayAccessor, RecordBatch, StringArray};
 use arrow::compute::filter_record_batch;
 use arrow::datatypes::{SchemaRef, UInt16Type};
-use arrow_array::BooleanArray;
+use arrow_array::{Array, BooleanArray};
 use dashmap::DashMap;
 use datafusion::common::config::ConfigOptions;
 use datafusion::common::error::{DataFusionError, Result};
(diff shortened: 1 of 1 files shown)


ethan-tyler:
@roeap - This handles the case where DF coalesces batches across file boundaries. Splits on file_id runs, applies DV/transforms per chunk and queues the fanout output.

Nice tie in with the DML refactors, since those implicitly assume per file processing remains correct even when DF gets aggressive with batching.  

Do we want “mixed file_id batches” to be a supported long term contract for DeltaScanExec, or should the scan try to enforce single file batches and do its own coalescing for perf?

roeap:
> Do we want “mixed file_id batches” to be a supported long term contract for DeltaScanExec, or should the scan try to enforce single file batches and do its own coalescing for perf?

There are many performance sesnsitive users (myself included :)) of delta-rs, so in general performance is always welcome :).

That said, right now we have to establish a baseline through this migration where in many places we need to clean up "what is happening" before going into too deep with optimisation. 

So for now I would go the simpler route, and start optimizing once we have consolidated the pre and post kernel world :)

ethan-tyler:
> There are many performance sesnsitive users (myself included :)) of delta-rs, so in general performance is always welcome :).
> 
> That said, right now we have to establish a baseline through this migration where in many places we need to clean up "what is happening" before going into too deep with optimisation.
> 
> So for now I would go the simpler route, and start optimizing once we have consolidated the pre and post kernel world :)

I agree - just wanted to see what your thoughts were on it. Thanks for the context and will keep this in mind   post migration.

ethan-tyler:
@roeap thanks for the review and merge!

On the ordering question - yes, for this specific fix we're only dealing with batch boundary changes from coalescing, which preserves input order. DV row position semantics are maintained.

The broader concern is completely valid. If an optimizer introduces operators that can reorder rows (repartition, interleave, etc.), DV application and log replay risk breaking. Making the ordering contract explicit via DF plan ordering requirements/guarantees seems like the right direction. For log replay, we'd need a total order (e.g. `(version, action_index)`), which likely implies single partition processing or an ordered merge.

I did a lot of research on this tonight - I'm going to open an issue to document this and discuss options. Would love your input there.

PR #4116, opened 2026-01-23
feat: add logical DataValidation node

roeap:
## Description

* https://github.com/delta-io/delta-rs/pull/4119
* https://github.com/delta-io/delta-rs/pull/4117
* __->__ https://github.com/delta-io/delta-rs/pull/4116

When writing data to delta tables we support writing this a change set containing both data to be written to data files and cdf data to be written to CDF files. When rewriting files we currently need to separate the writes for data we keep and data we update due to predicate handling.

To better reuse these advanced writing features it is useful if we can perform data validation on subtrees (e.g. only on newly inserted data) in our plans. For this we add a logical data validation node we can conveniently integrate into our logical planning.

### DIFF
--- a/crates/core/src/delta_datafusion/data_validation.rs
+++ b/crates/core/src/delta_datafusion/data_validation.rs
@@ -1,4 +1,5 @@
 use std::any::Any;
+use std::collections::HashSet;
 use std::pin::Pin;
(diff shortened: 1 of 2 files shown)


roeap:
* merged via #4117

Issue #3687, opened 2025-08-18
Full partition delete generates data into **_change_data** directory.


detoxifiedplant:
# Environment

**Delta-rs version**:
0.27.0

**Binding**: 
Rust

**Environment**:
- **Cloud provider**:
- **OS**: MacOS
- **Other**:

***
# Bug

**What happened**:
Running a delete query in partition while having Change data feed enabled, it generates data in **_change_data** directory.

**What you expected to happen**:
As per the [documentation](https://docs.delta.io/latest/delta-change-data-feed.html) here, **full partition delete will not generate data in the _change_data directory.**

**How to reproduce it**:

<details>
<summary>Example</summary>

```rust
use std::fs;
use std::path::Path;
use std::sync::Arc;
use std::time::Duration;

use anyhow::Result;
use chrono::NaiveDate;
use datafusion::prelude::*;
use deltalake::arrow::array::{Date32Array, Int64Array, RecordBatch};
use deltalake::arrow::datatypes::{DataType, Field, Schema};
use deltalake::kernel::{PrimitiveType, StructField, StructType};
use deltalake::parquet::file::properties::WriterProperties;
use deltalake::{DeltaOps, DeltaTable};
use tokio::time::sleep;

const TABLE_PATH: &str = "/tmp/users_delta_table/";
const TEST_DATE: &str = "2025-04-12";
const ROWS_PER_ITERATION: usize = 200_000;
const TOTAL_ITERATIONS: usize = 5;

#[tokio::main]
async fn main() -> Result<()> {
    if Path::new(TABLE_PATH).exists() {
        fs::remove_dir_all(TABLE_PATH)?;
        println!("Cleaned up existing table at {}", TABLE_PATH);
    }

    // create table
    let table = create_users_table().await?;

    // Verify partition columns
    assert_eq!(
        table.metadata()?.partition_columns(),
        &vec!["date"],
        "Partition columns not set correctly"
    );

    println!(
        "Partition information: {:?}",
        table.metadata()?.partition_columns(),
    );

    // insert data for same partition(date column)
    let table = insert_test_data(table).await?;
    println!(
        "Number of inserted rows {}",
        ROWS_PER_ITERATION * TOTAL_ITERATIONS,
    );

    // confirm number of inserted records
    let records = records_count(&table).await?;
    assert_eq!(ROWS_PER_ITERATION * TOTAL_ITERATIONS, records as usize);

    // delete records by partition value (same date value)
[...]

PR #4117, opened 2026-01-23
feat: migrate delete

roeap:
# Description

- fixes https://github.com/delta-io/delta-rs/issues/3687

* https://github.com/delta-io/delta-rs/pull/4119
* __->__ https://github.com/delta-io/delta-rs/pull/4117
* https://github.com/delta-io/delta-rs/pull/4116

This PR migrates the delete operation to do more logical planning and to fully leverage or more advanced change-set writing capabilities.

in the `replaceWhere` case we were previously doing 2 write operations, 1) to re-write the data that we need to retain from the deleted files, and another for writing change data if needed.

Our current v2 write code is already able to handle this in a single pass by writing a change set. WE now just embrace it more :)

### DIFF
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -411,6 +411,10 @@ pub(crate) struct MatchedFilesScan {
     /// we do not (yet) have full coverage to translate datafusion to
     /// kernel predicates.
     pub(crate) delta_predicate: Arc<Predicate>,
+    /// The predicate contains only partition column references
+    ///
+    /// This implies that for each matched file all data matches.
+    pub(crate) partition_only: bool,
 }
 
 impl MatchedFilesScan {
@@ -458,12 +462,13 @@ pub(crate) async fn scan_files_where_matches(
         .partition_columns();
     // validate that the expressions contain no illegal variants
     // that are not eligible for file skipping, e.g. volatile functions.
+    let mut visitor = FindFilesExprProperties {
+        partition_columns: partition_columns.clone(),
+        partition_only: true,
+        result: Ok(()),
+    };
     for term in &skipping_pred {
-        let mut visitor = FindFilesExprProperties {
-            partition_columns: partition_columns.clone(),
-            partition_only: true,
-            result: Ok(()),
-        };
+        visitor.result = Ok(());
         term.visit(&mut visitor)?;
         visitor.result?
     }
@@ -533,5 +538,6 @@ pub(crate) async fn scan_files_where_matches(
         valid_files,
         predicate,
         delta_predicate,
+        partition_only: visitor.partition_only,
     }))
 }
--- a/crates/core/src/kernel/transaction/mod.rs
+++ b/crates/core/src/kernel/transaction/mod.rs
@@ -306,7 +306,8 @@ impl CommitData {
             );
             app_metadata.extend(commit_info.info);
             commit_info.info = app_metadata.clone();
-            actions.push(Action::CommitInfo(commit_info))
+            // commit info should be the first action to support in-commit timestamps.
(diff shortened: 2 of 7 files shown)


crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -458,12 +462,13 @@ pub(crate) async fn scan_files_where_matches(
...
+    };
     for term in &skipping_pred {
-        let mut visitor = FindFilesExprProperties {
-            partition_columns: partition_columns.clone(),
-            partition_only: true,
-            result: Ok(()),
-        };
+        visitor.result = Ok(());
         term.visit(&mut visitor)?;
         visitor.result?
```
roeap, line 473: @hntd187 - incorporated you feedback re re-using the visitor here.

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -306,7 +306,8 @@ impl CommitData {
             );
             app_metadata.extend(commit_info.info);
             commit_info.info = app_metadata.clone();
-            actions.push(Action::CommitInfo(commit_info))
+            // commit info should be the first action to support in-commit timestamps.
+            actions.insert(0, Action::CommitInfo(commit_info));
```
roeap, line 310: Just a drive-by fix in fear of forgetting it. But when inCommit timestamps are supported, we need to write the commit info as first object always. Might as well do this now.

crates/core/src/operations/write/execution.rs
```diff
@@ -321,6 +322,55 @@ pub(crate) async fn write_execution_plan_v2(
...
 }
 
+pub(crate) async fn write_exec_plan(
+    session: &dyn Session,
+    log_store: &dyn LogStore,
+    table_config: &TableConfiguration,
+    exec: Arc<dyn ExecutionPlan>,
+    operation_id: Option<Uuid>,
+    write_as_cdc: bool,
+) -> DeltaResult<(Vec<Action>, WriteExecutionPlanMetrics)> {
```
roeap, line 332: Sorry for introducing yet another `write_*` function. However this one has a simpler API as it does the extraction of writer properties from the session etc. internally. 

It also allows us to plan validation in the execution plan a bit more fine granularly. Will consolidate and add more comments in one of the next PRs.

python/tests/test_cdf.py
```diff
@@ -496,17 +495,20 @@ def test_delete_partitioned_cdf(tmp_path, sample_data_pyarrow: "pa.Table"):
...
-    cdc_data = pq.read_table(cdc_path, schema=table_schema)
+    cdc_data = (
+        pa.table(dt.load_cdf().read_all())
+        .filter(pc.field("_change_type") == "delete")
+        .select(["int64", "_change_type"])
+    )
 
-    assert os.path.exists(cdc_path), "_change_data doesn't exist"
-    assert len(os.listdir(cdc_path)) == 2
-    assert cdc_data == expected_data
```
roeap, line 509: Turns out the `_change_data` folder should actually not exist in this case. since we are partitioned on a column - due to the structure of the source data we actually have one row per partition here. So when deleting we are only deleting whole files and do not need to write any CDC data.

Changed the assertion now that we are producing the right change data in `load_cdf`.

crates/core/src/operations/delete.rs
```diff
@@ -1124,6 +1084,81 @@ mod tests {
...
+                    Some("2020"),
+                    Some("2024"),
+                    Some("2025"),
+                ])),
+                Arc::new(Int32Array::from(vec![Some(1), Some(2), Some(3), Some(4)])),
+            ],
+        )
+        .unwrap();
+
+        println!("protodol: {:?}", table.snapshot().unwrap().protocol());
```
ion-elgreco, line 1118: Typo 😜

PR #4119, opened 2026-01-24
feat: migrate write operation

roeap:
# Description

* __->__ https://github.com/delta-io/delta-rs/pull/4119

This PR migrates the write operation to leverage more logical planning and the new table provider / "find-files" logic.

* remove `find_files` and associated functions which heavily relied on both the current table provider and explicit `Add` actions.


crates/core/src/operations/write/plan.rs
```diff
@@ -0,0 +1,895 @@
...
+    log_store: &dyn LogStore,
+    snapshot: &EagerSnapshot,
+    plan: LogicalPlan,
+    where_clause: Expr,
+) -> Result<(Vec<Action>, LogicalPlan, bool)> {
+    let maybe_files_scan = scan_files_where_matches(session, snapshot, where_clause).await?;
+    let Some(files_scan) = maybe_files_scan else {
+        // No files match the predicate, so we can just return the original plan
+        return Ok((vec![], plan, false));
+    };
```
ethan-tyler, line 438: Early return bypasses replaceWhere validation. Incoming rows not checked
against where_clause when no files match. I would validate plan before file scan.

Something like - `) -> Result<(Vec<Action>, LogicalPlan, bool)> {`

crates/core/src/operations/write/mod.rs
```diff
@@ -334,63 +359,69 @@ impl WriteBuilder {
...
-            .ok_or::<DeltaTableError>(WriteError::MissingData.into())?;
-        let schema: StructType = input.schema().as_arrow().try_into_kernel()?;
+        let Some(source) = self.input.take() else {
+            return Err(WriteError::MissingData.into());
+        };
+
+        let schema: StructType = source.schema().as_arrow().try_into_kernel()?;
 
-        match &self.snapshot {
+        match self.snapshot.take() {
```
ethan-tyler, line 406: `take()` leaves `self.snapshot` as None. Commit path later calls
`snapshot.as_ref()` and gets None, skips protocol checks. Clone or thread through would be what I'd do.

crates/core/src/operations/write/plan.rs
```diff
@@ -0,0 +1,895 @@
...
+            if &expr.get_type(source_schema)? != field.data_type() {
+                if safe_cast {
+                    expr = try_cast(expr, field.data_type().clone());
+                } else {
+                    expr = expr.cast_to(field.data_type(), source_schema)?;
+                }
+            }
+            expr = expr.alias(field.name());
+            leaf_transforms.push((leaf, expr, true));
+        } else if let Some(gen_expr) = generated_columns.remove(&leaf) {
```
ethan-tyler, line 291: The "pick from source" path casts to `field.data_type()` when needed, but the generated column path pushes `gen_expr` as is.

crates/core/src/delta_datafusion/engine/expressions/to_kernel.rs
```diff
@@ -23,9 +23,7 @@ pub(crate) fn to_delta_predicate(expr: &Expr) -> Result<Predicate> {
 /// Converts a DataFusion expression to a Delta kernel expression.
 pub(crate) fn to_delta_expression(expr: &Expr) -> Result<Expression> {
     match expr {
-        Expr::Column(column) => Ok(Expression::Column(ColumnName::from_naive_str_split(
-            &column.name,
-        ))),
+        Expr::Column(column) => Ok(Expression::Column(ColumnName::new([&column.name]))),
```
ethan-tyler, line 26: What's the expected contract for `Expr::Column(column.name)` at this engine boundary?

crates/core/src/operations/write/plan.rs
```diff
@@ -0,0 +1,895 @@
...
+        let projection = project_for_insert(&table_schema, &source_schema, gen_lookup, safe_cast)?;
+        let plan = plan.into_builder().project(projection)?.build()?;
+        (plan, SchemaAction::NoChange)
+    } else if missing_in_table > 0 {
+        // The source schema has new fields not present in the table schema
+        // We need to evolve the table schema to include those new fields
+        // The source data already has all the necessary fields
+        validate_schema_evolution(plan.schema().as_arrow(), table_config, &gen_lookup)?;
+        (plan, SchemaAction::Evolve(source_schema))
+    } else {
```
ethan-tyler, line 165: `SchemaAction::Evolve(source_schema)` replaces with source schema, which may not carry existing table field metadata.

Consider merging into*the existing schema rather than replacing with source schema.

crates/core/src/operations/write/mod.rs
```diff
@@ -755,6 +595,158 @@ impl std::future::IntoFuture for WriteBuilder {
...
+    let (source, schema_action) =
+        plan_insert(session, source, table_config, schema_mode, safe_cast)?;
+
+    // Determine if the required schema changes can be applied to the table
+    // and generate the necessary actions to update the table protocol/metadata.
+    let mut actions = vec![];
+    match schema_action {
+        SchemaAction::NoChange => (),
+        SchemaAction::Evolve(to_schema) => {
+            if table_config.is_feature_enabled(&TableFeature::TypeWidening) {
```
ethan-tyler, line 645: Schema evolution is blanket blocked when `TypeWidening` is enabled. If this blocks purely additive evolution (no widening), that's likely too strict.

Consider limiting the failure to actual widening changes only.

crates/core/src/operations/write/execution.rs
ethan-tyler, line 1: Per-write tuning knobs appear dropped

`WriteBuilder::with_target_file_size` and `with_write_batch_size` are [documented public APIs](https://docs.rs/deltalake/latest/deltalake/operations/write/struct.WriteBuilder.html), but the new orchestration path does not appear to thread them through.

- `write_exec_plan(...)` now derives `target_file_size` only from table properties
- `write_batch_size = None` is passed into both `write_cdc_plan` and `write_data_plan`

IMO:
- Either thread these knobs through the new execution path, or
- Explicitly deprecate/remove them with release notes and migration guidance

I couldn't find the exact lines in the diff because the relevant code was deleted/replaced. The new execution path in `execution.rs` doesn't appear to accept these parameters.

ethan-tyler (commented):
Thanks for this work- the architectural direction is coherent and should reduce duplicated DML machinery. The projection builder is thoughtful, especially the struct assembly optimization.

Provided some comments below.

ethan-tyler:
Nullability inference safety concern

`DataValidation::try_new(...)` walks validation expressions and tightens schema via `make_fields_non_nullable(...)`. The `NotNullExtractor` treats any `Expr::IsNotNull(...)` or `NOT (.. IS NULL)` occurrence as proof the referenced column is non-nullable.

IMO - This inference is not generally sound for arbitrary boolean forms (e.g., `a IS NOT NULL OR b IS NOT NULL`, `CASE`, nested negations), and tightening Arrow nullability based on it can mislead downstream planning.

- What expressions does `DataValidation` actually produce?
- If it only generates simple conjunctions of `col IS NOT NULL` terms, narrow the inference to that subset
- If users can inject arbitrary predicates, this needs hardening

Not introduced by this PR, but worth noting: the expanded use of DataValidation makes this pre-existing concern more relevant..

PR #4122, opened 2026-01-25
chore: allow integration tests to be run in parallel with nextest

rtyler:
For the AWS integration tests this is a 77% speedup on my machine when
running the tests. Compile/link times are not changed here, but at least
the tests run faster!

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,4 @@
+[profile.ci]
+fail-fast = false
+retries = 1
+slow-timeout = "60s"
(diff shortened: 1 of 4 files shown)


rtyler:
Comparing some of these integration test runs to the others and this is looking like a really positive improvement! LakeFS integration tests cut in half (6->3m), aws tests cut by ~30% (9.5->6)

The azure integration tests are only moderately sped up since there are few integration tests and the build (datafusion!) is substantially slower than the others.

PR #4127, opened 2026-01-27
fix(datafusion): resolve DML predicates against execution scan schema

### LINKED ISSUE #4125: [Bug]: DeltaTable.delete fails during equality comparison with a string
### What happened?

`DeltaTable.delete` fails with a equality predicate like `col = 'foo'`

I believe #4098 might have caused this (but not sure) 

### Expected behavior

It should work

### Operating System

None

### Binding

None

### Bindings Version

1.4.0

### Steps to reproduce

```py

df = pl.DataFrame({"id": ["a", "b"], "status": ["active", "inactive"]})

with tempfile.TemporaryDirectory() as tmpdir:
    df.write_delta(tmpdir)
    deltalake.DeltaTable(tmpdir).delete("status = 'inactive'")

```

### Relevant logs

```shell
  File "/home/dan/code/github/anam-org/metaxy/repro_delta_utf8.py", line 21, in <module>
    deltalake.DeltaTable(tmpdir).delete("status = 'inactive'")
  File "/home/dan/code/github/anam-org/metaxy/.venv/lib/python3.10/site-packages/deltalake/table.py", line 1080, in delete
    metrics = self._table.delete(
_internal.DeltaError: Generic DeltaTable error: Parquet error: External: Compute error: Error evaluating filter predicate: ArrowError(InvalidArgumentError("Invalid comparison operation: Utf8 == Utf8View"), Some(""))
```

ethan-tyler:
# Description
After the DataFusion 52 upgrade, DELETE and UPDATE operations fail with:

`Invalid comparison operation: Utf8 == Utf8View`

Error: `Arrow error: Invalid comparison operation: Utf8 == Utf8View`

This occurs when using string equality predicates like table.delete().with_predicate("name = 'foo'").

## Root Cause:
With DataFusion 52, Parquet string/binary columns may be represented as view types (`Utf8View`/`BinaryView`) when `datafusion.execution.parquet.schema_force_view_types` is enabled (default `true`). 

DELETE/UPDATE were resolving predicates and SET expressions against `snapshot.arrow_schema()` (base `Utf8`/`Binary`), but executing against the scan/provider schema (which can include view types). This mismatch caused filter evaluation errors like `Utf8 == Utf8View` (and similarly `Binary == BinaryView`).

## Fix:
Resolve predicates against DeltaScanConfig::table_schema() instead of snapshot.arrow_schema(). This ensures predicate literals are coerced to match the actual execution schema (view types when enabled, base types when disabled).

## Changes:
- delete.rs: Use DeltaScanConfig::table_schema() for predicate resolution
- update.rs: Use DeltaScanConfig::table_schema() for both SET expressions and WHERE predicates
- expr.rs: Add ScalarValue::Dictionary support to fmt_expr_to_sql (partition columns are dictionary-encoded in execution schema)

##  Tests:
- test_delete_string_equality_utf8view_regression_4125
- test_update_string_equality_non_partition
- test_delete_partition_string_predicate_dictionary_formatting
- test_delete_binary_equality_non_partition
- test_delete_custom_session_schema_force_view_types_disabled
- Dictionary scalar formatting unit tests in expr.rs

# Related Issue(s)
- Fixes #4125

# Documentation


crates/core/src/operations/delete.rs
```diff
@@ -200,7 +201,13 @@ impl std::future::IntoFuture for DeleteBuilder {
 
             let predicate = this
                 .predicate
-                .map(|p| p.resolve(session.as_ref(), snapshot.arrow_schema().to_dfschema_ref()?))
+                .map(|p| {
+                    let scan_config = DeltaScanConfig::new_from_session(session.as_ref());
```
roeap, line 205: Could we move the construction of `DeltaScanConfig` out of the iterator and re-use it?

roeap: Also, just ignore me as this is options not iterators 🤦.

PR #4132, opened 2026-01-27
feat: update asserted nullability in DataValidation output schema

roeap:
# Description

Following our newly introduces AI disclosure guidelines - I used AI to write the recursive methods that update field nullability.

In the current refactoring efforts, we are combining more and more operations into logical plans and further defer doing physical work. In some cases we combine data that may differ in the logical nullability, but be equivalent in practical terms - i.e. no null values present, in nullable field.

We assert the absence of null values in the `DataValidationExec`, to be able to combine different sources (usually inserted data and data rescued from exisiting files) we also need to update the schema to be matching in null values.

### DIFF
--- a/crates/core/src/delta_datafusion/data_validation.rs
+++ b/crates/core/src/delta_datafusion/data_validation.rs
@@ -661,7 +661,7 @@ pub(crate) fn collect_non_nullable_fields(schema: &Schema) -> Vec<ColumnName> {
     let mut non_nullable_paths = Vec::new();
 
     for field in schema.fields() {
-        let path = vec![field.name().to_string()];
+        let path = vec![field.name()];
         collect_non_nullable_fields_recursive(field, path, &mut non_nullable_paths);
     }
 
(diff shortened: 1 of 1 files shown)


crates/core/src/delta_datafusion/data_validation.rs
```diff
@@ -118,62 +228,76 @@ impl ExtensionPlanner for DataValidationExtensionPlanner {
...
+    let non_nullable_source: HashSet<_> = collect_non_nullable_fields(source_schema.as_arrow())
+        .into_iter()
+        .collect();
+    let mut validations: Vec<_> = non_nullable_table
+        .difference(&non_nullable_source)
+        .map(|col| {
+            to_datafusion_expr(
+                &Expression::Column(col.clone()),
+                &delta_kernel::schema::DataType::BOOLEAN,
+            )
```
roeap, line 281: we run this through kernel conversion since field access is a bit cumbersome to resolve on the DF side and its already handled there.

crates/core/src/delta_datafusion/data_validation.rs
```diff
@@ -472,22 +638,203 @@ where
...
+                let mut child_path = current_path.clone();
+                child_path.push(child.name().to_string());
+                collect_non_nullable_fields_recursive(child, child_path, non_nullable_paths);
+            }
+        }
+        // DataType::List(child) | DataType::LargeList(child) => {
+        //     let mut child_path = current_path.clone();
+        //     child_path.push("element".to_string());
+        //     collect_non_nullable_fields_recursive(child, child_path, non_nullable_paths);
+        // }
```
rtyler: por que?

If this code is meant to be implemented later, I would rather see a `todo!()`

Issue #4113, opened 2026-01-22
[Feature]: Matched file filtering via IN (...) doesn't scale for large file sets


ethan-tyler:
### Is your feature request related to a problem?

PR #4111 adds `scan_files_where_matches()` to identify files containing matching rows for DML rewrites (UPDATE now, MERGE/DELETE next). The second scan is restricted to matched files via a plan-time `IN (...)` predicate:
```rust
let file_list = valid_files
    .iter()
    .flat_map(|arr| arr.iter().flatten().map(|v| v.to_string()))
    .map(|v| lit(wrap_partition_value_in_dict(ScalarValue::Utf8(Some(v)))))
    .collect_vec();

let plan = LogicalPlanBuilder::scan("source", table_source, None)?
    .filter(col(FILE_ID_COLUMN_DEFAULT).in_list(file_list, false))?
    // ...
```

This builds one `Expr::Literal` per matched file. For rewrites touching thousands of files, that means large expression trees, slow planning, and memory overhead. This is a file level restriction implemented as a DF expression over the synthetic file-id column, not a Parquet row predicate.

### Describe the solution you'd like

**Scan-planning allowlist:** 

Instead of constructing `Expr::InList` with one literal per matched file, plumb an allowlist of matched file ids into the scan planning path and filter the candidate files before building Parquet `FileGroups` / `PartitionedFiles`. This avoids plan-time bloat while preserving true file pruning (don't read Parquet for excluded files).

**Optional safety net:** 

exec-time membership filtering in `DeltaScanExec` (e.g., per run via `split_by_file_id_runs()` from #4112) to guard against mixed batch edge cases. However, this happens after Parquet reads and doesn't provide I/O savings.

**Something like:**

1. Add `file_id_allowlist: Option<Arc<HashSet<String>>>` to `DeltaScanConfig`
2. Add `with_file_id_allowlist()` builder method
3. Filter in scan planning when assembling file list / file groups (before Parquet read plan)
4. Optionally also guard in `DeltaScanExec` for mixed-batch correctness

### Describe alternatives you've considered

Alternatives considered:
- **Semi-join against MemTable** - avoids literal list, uses DF join machinery. 
- **IN (subquery)** - depends on DF planner behavior
- **Chunked IN (...)** - still builds large expressions, just delays the cliff

### Priority

Medium - Would be helpful

### Additional context

**Related:**
- #4111 
- #4112

### Contribution

- [x] I'm willing to submit a pull request for this feature
[...]

PR #4134, opened 2026-01-27
feat: migrate optimize operation

roeap:
# Description

migrating the optimize operation. We do introduce a new fn `scan_file_list` which will likely be consolidated as we look into #4113. 

Other that that fairly straight forward refactor.

### DIFF
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -8,7 +8,7 @@ use arrow_array::{Array, GenericByteViewArray, RecordBatch, StringArray};
 use arrow_schema::{DataType, Field, Schema, SchemaRef};
 use datafusion::catalog::Session;
 use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion, TreeNodeVisitor};
-use datafusion::common::{HashSet, Result};
+use datafusion::common::{HashSet, Result, plan_err};
 use datafusion::datasource::{MemTable, provider_as_source};
 use datafusion::execution::context::SessionContext;
 use datafusion::logical_expr::utils::{conjunction, split_conjunction_owned};
@@ -25,6 +25,7 @@ use futures::TryStreamExt as _;
 use itertools::Itertools;
 use percent_encoding::percent_decode_str;
 use tracing::*;
+use url::Url;
 
 use crate::delta_datafusion::engine::to_delta_predicate;
 use crate::delta_datafusion::logical::LogicalPlanBuilderExt as _;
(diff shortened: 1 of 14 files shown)


crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -541,3 +542,62 @@ pub(crate) async fn scan_files_where_matches(
...
     }))
 }
+
+pub(crate) async fn scan_file_list(
+    snapshot: &EagerSnapshot,
+    file_list: impl IntoIterator<Item = impl ToString>,
+) -> Result<LogicalPlan> {
+    let table_source = provider_as_source(
+        DeltaScanNext::builder()
+            .with_eager_snapshot(snapshot.clone())
```
rtyler, line 553: sus

ion-elgreco:
@roeap looks good for z-order but for compaction we should go through table provider now as well otherwise we run into issues if the table has deletion vectors

rtyler (changes requested):
I would like to approve this PR before merge, will have time later today (I hope!)

crates/core/src/delta_datafusion/find_files.rs
roeap: The `TableProviderBuilder` behind this will also accept a `Snapshot` via the `with_snapshot` method, it's been specifically designed to only use things we can also do via `Snapshot`.

roeap:
>  but for compaction we should go through table provider now as well otherwise we run into issues if the table has deletion vectors

good catch - i was focussed on the removal of `DeltaTableProvider`, but of course this need updates too.

crates/core/src/operations/optimize.rs
```diff
@@ -694,48 +674,36 @@ impl MergePlan {
...
+                            ));
+                            util::flatten_join_error(rewrite_result).await
+                        }
                     })
                     .boxed()
             }
         };
 
-        let mut stream = stream.buffer_unordered(max_concurrent_tasks);
+        // let mut stream = stream.buffer_unordered(max_concurrent_tasks);
```
ion-elgreco: Old?

crates/core/src/operations/optimize.rs
```diff
@@ -561,7 +563,7 @@ impl MergePlan {
             task_parameters.stats_columns.clone(),
         )?;
 
-        let mut read_stream = read_stream.await?;
+        // let mut read_stream = read_stream.await?;
```
ion-elgreco: Old?

crates/core/src/operations/optimize.rs
roeap: actually i wanted to raise this one ... previously we had a stream of futures which we could execute with this buffered approach. But we also did manual parquet reading.

I am not sure if we are loosing performance or if we will have even better concurrency due to using more datafusion to just leverage all the core within individual tasks but then execute tasks esentially in sequence.

crates/core/src/operations/optimize.rs
roeap: yes, removed.

crates/core/src/operations/optimize.rs
ion-elgreco: Hmm not sure, but I assume datafusion would do it concurrently when there are multiple files in a bin.

I guess instead of doing this per partition sequentially, could we get more performance if we union all the partition plans into one plan and then pipe it through write_execution_plan

crates/core/src/operations/optimize.rs
```diff
@@ -312,6 +317,10 @@ impl<'a> OptimizeBuilder<'a> {
     }
 
     /// Max number of concurrent tasks
+    #[deprecated(
+        since = "0.31.0",
+        note = "This setting is no longer used and handled on the daatfusion session."
+    )]
     pub fn with_max_concurrent_tasks(mut self, max_concurrent_tasks: usize) -> Self {
         self.max_concurrent_tasks = max_concurrent_tasks;
         self
```
roeap, line 329: @ion-elgreco - i deprecated the setting since we are no longer using it. If we are concerned about the performance though, I can try to get this to work

crates/core/src/operations/optimize.rs
roeap: >I guess instead of doing this per partition sequentially, could we get more performance if we union all the partition plans into one plan and then pipe it through write_execution_plan

TO be honest, the compaction could maybe be replaced by exactly that, we may not even have to bin, but rather just sort the data if needed and run it all through the V2 writer. Ultimately its about getting better file bins, but why not distrubute even on the record batch level which should probably also generate more even file sizes.

ion-elgreco: Yeah that's the only thing though, I'm not sure if this still holds since we have lazy writes, but before having unsorted partitions would lead to many partition writers being opened and holding a large buffer in memory

ion-elgreco: This should be fairly easy to test with dummy data though, just do alot of small appends to 1000 partitions and then try compaction with current setup and a version that just pipes everything through write_execution_plan

ion-elgreco: This should be fairly easy to test with dummy data though, just do alot of small appends to 1000 partitions and then try compaction with current setup and a version that just pipes everything through write_execution_plan

rtyler:
@roeap What's the status of this pull request, it's ready to review but it looks like it is not compatible with `main` at the moment.

ion-elgreco:
> @roeap What's the status of this pull request, it's ready to review but it looks like it is not compatible with `main` at the moment.
> 

The code is fine but we need some benchmarks before merging

PR #4140, opened 2026-01-28
fixup! [HSTACK] - propagate creds for object store in distributed exec

mihai-bleont:
Move the registration of the delta object_store to cobweb. The changes in cobweb are here: https://github.com/AdobeAnalytics/axle-cobweb/pull/359


mihai-bleont:
Opened this PR in the wrong place

Issue #4135, opened 2026-01-28
[Bug]: Segmentation fault when registering table to datafusion session


JonatanMartens:
### What happened?

When registering a table in a datafusion session it fails with segmentation fault.

It might be because the datafusion version I use is 51 while delta-rs uses 52, but the python bindings for datafusion 52 have not yet been released.

### Expected behavior

The table registers successfully

### Operating System

Linux

### Binding

Python

### Bindings Version

1.4.0

### Steps to reproduce

```python
from deltalake import DeltaTable
from datafusion import SessionContext

context = SessionContext()
table = DeltaTable("./path/to/table")
context.register_table("test", table)
```

### Relevant logs

```shell
[1]    38512 segmentation fault
```

ion-elgreco:
This is afaik because there was a breaking change in the ffi api

JonatanMartens:
So once datafusion 52 is released this should work?

ion-elgreco:
> So once datafusion 52 is released this should work?

It should yeah

ethan-tyler:
The DF52 wheels should be released soon. I've been engaged and helping with the final testing, will keep you posted.

ethan-tyler:
Added a runtime guard: `__datafusion_table_provider__()` now errors when `datafusion` major version isn't 52 instead of segfaulting.

**Workaround:** use [`QueryBuilder`](https://delta-io.github.io/delta-rs/api/query/) (no DataFusion-Python required):
```python
from deltalake import DeltaTable, QueryBuilder
dt = DeltaTable("path/to/table")
data = QueryBuilder().register("tbl", dt).execute("SELECT * FROM tbl")
```

PR #4142, opened 2026-01-28
fix(python): guard DataFusion FFI export on datafusion major version

ethan-tyler:
# Description

- Add a runtime version check in __datafusion_table_provider__ to prevent FFI ABI mismatch segfaults
- Block capsule export when installed datafusion major != 52
- Provide actionable error text with QueryBuilder workaround

Changes:

- lib.rs: add REQUIRED_DATAFUSION_PY_MAJOR, datafusion_python_version(), guard at method start
- test_datafusion.py: add incompatible version and not installed tests

Note: This guard is a temporary safety net to prevent segfaults until DataFusion 52 Python wheels are available on PyPI. Once wheels land, users can install datafusion==52.* and use SessionContext registration normally.

# Related Issue(s)
- #4135 

# Documentation


rtyler (approved):

rtyler:
@ion-elgreco if this version of the python package were used in the future with a datafusion-ffi of `53` wouldn't there still likely be a SIGSEGV due to ABI incompatibilities? 

Basically as I understand the problem here, this version number will have to change as newer major releases are built into our wheels

ion-elgreco:
> @ion-elgreco if this version of the python package were used in the future with a datafusion-ffi of `53` wouldn't there still likely be a SIGSEGV due to ABI incompatibilities? 
> 
> Basically as I understand the problem here, this version number will have to change as newer major releases are built into our wheels

Only if the ffi interface changes but that's assuming we will have breaking changes every major release of datafusion-ffi. But before it was working for multiple major versions

rtyler:
@ion-elgreco in the future we can update the check with the new major versions, but I think we should be strict until newer versions are proven to work. The consequences of failing is a complete segfault which is about as disastrous of an outcome as it gets

ion-elgreco:
@rtyler alright that's fair

Issue #4139, opened 2026-01-28
update_datafusion_session has the wrong order of arguments


rtyler:
### Describe the solution you'd like

This was discussed on a live-stream with @roeap, but `update_datafusion_session` is dyslexic in that it updates the _second_ argument's state (`Session`).

Functions which update _something_ should conventionally make the subject of the update the first argument, i.e. if we're updating the logstore(s) on the session, the session should be the first argument.

### Priority

None

### Contribution

- [ ] I'm willing to submit a pull request for this feature
- [x] I can help with testing this feature
- [x] I can help with documentation for this feature

pauldouane:
take

Issue #4081, opened 2026-01-14
with_session_state silently falls back when Session isn't SessionState


ethan-tyler:
## What happened?
Builders like OptimizeBuilder, MergeBuilder, UpdateBuilder accept Arc<dyn Session> via with_session_state(...), but internally require a concrete SessionState. When the downcast fails, they silently fall back to an internal default:

```
session.as_any().downcast_ref::<SessionState>()
    .cloned()
    .unwrap_or_else(|| create_internal_session_state(...))

```

Relevant paths:
- crates/core/src/operations/optimize.rs
- crates/core/src/operations/merge/mod.rs
- crates/core/src/operations/update.rs

The operation "works", but caller-provided config/runtime/registrations get silently ignored.

## Expected behavior
Either accept SessionState directly, error when downcast fails, or at minimum log/document that fallback happened.

## Operating System
All (API-level behavior)

## Binding
Rust

## Bindings Version
Observed on `main` / DF52 upgrade work (DataFusion 52).

## Steps to reproduce

1. Create a session with custom config (runtime env, target partitions, UDF registrations, etc.)
2. Pass it to OptimizeBuilder::with_session_state(...) as Arc<dyn Session> where concrete type isn't SessionState
3. Operation runs with internal defaults instead of caller's session - no warning

## Relevant logs
N/A (the problem is the absence of any warning/error when fallback happens)

ethan-tyler:
take

Issue #4138, opened 2026-01-28
Runtime config and resource management


roeap:
### Is your feature request related to a problem?

In the early days delta-rs has been build as more or less standalone library to read and write Delta Tables. Then case the rise of more complex operations like MERGE, OPTIMIZE et. al. These operations often require custom datafusion extension to work properly. Examples being the logical `MetricObserver` or `DataValidation` operators which require extension planners to be translated to physical plans. Datafusion sessions also carry other important resources like object stores and caches.

As a result we started creating dedicated Datafusion sessions internally to execute these operations. However creating dedicated sessions brings several challenges that are hard to manage properly. Particularly we see
* Parquet footer caching will not work
* We need to handle resources (object stores) inside operations
* Memory management on the session (e.g. spill size management) gets ignored
* ... and many more

In several places we already allow callers to provide their own sessions, but without clear guidelines on how to properly set up a session such that it can be used with all delta-rs operations.

There are similar problems when it comes to configuration. We collect several configurations (e.g. processing batch size) that can also be configured via the Datafusion session. 

Even more so there are advanced configurations like parquet encryption (#3300) that need to be wired through to all parquet reads.

The whole issue gets even more complicated since we have two execution layers - the kernel engine and our processing - which need to be aligned.

While it seems that the current code "does what it should" I believe we should formulate a more coherent strategy to manage our runtime environment.

### Describe the solution you'd like

There are at least 3 usage scenarios we should support: 
1. Users who do not care about the session and just want to interact with a delta table.
2. Users who are integrating into a datafusion based application
3. Advanced users who maintain their own `Session` implementation.

For each of these we should have clear guidelines on the appropriate APIs and usage patterns.

The first user is looking for an equivalent of the datafusion `DataFrame` API which I beleive we are serving with the `DeltaTable`. WHich would currently look similar to:

```rs
// our open table APIs provide a minimal API for opening tables.
// The DeltaTableBuilder is another simple way to create new tables
let dt = open_table(...);
let (dt, result) = dt.scan_table().await?;
let (dt, metrics) = dt.write(result).await?
```

While I believe these APIs are somewhat convenient, most basic users are not interested in the metrics. While I believe this is bearable for all users, we could consider providing a simplified API without directly exposing metrics (but keeping them accessible).

Some of these APIs allow you to pass a DataFrame which is essentially a logical plan coupled with a session. In these cases we could propagate the session. For example in the merge operation:

```rs
#[must_use]
[...]

PR #4145, opened 2026-01-29
feat: session-first DataFusion integration + session resolution policies

ethan-tyler:
# Description

Make delta-rs' DataFusion integration consistently honor caller provided sessions and introduce a session first API for registering Delta object stores.

**Changes:** 

Session resolution: 
- Add `resolve_session_state(...)` & `SessionFallbackPolicy` (`InternalDefaults` / `DeriveFromTrait` / `RequireSessionState`)
- Builders expose `with_session_fallback_policy(...)` to control strictness
- Migrate operations to use resolver (optimize/merge/update/write/delete)

Session first registration:
- Add `DeltaSessionExt` trait:
  - `ensure_object_store_registered(...)`
  - `ensure_log_store_registered(...)`
- Deprecate `DeltaTable::update_datafusion_session(...)` (shim kept for compatibility)

Predicate parsing:
- Non-`SessionState` sessions can preserve UDFs when configured via `DeriveFromTrait`

**Compatibility:**
- Default is backward compatible: `SessionFallbackPolicy::InternalDefaults` warns but doesn't break
- Strict mode available: `with_session_fallback_policy(RequireSessionState)` errors instead of falling back
- `DeltaTable::update_datafusion_session` remains but is deprecated

**Tests:**
- Regression tests for fallback policy wiring across builders (`RequireSessionState` path)
- Existing `deltalake-core` DataFusion test suite passes with `--features datafusion`

# Related Issue(s)
**Addresses:** 
- #4081
- #4139

# Documentation

# Follow ups:

- Flip default to `RequireSessionState` (breaking change)
- Remove deprecated `DeltaTable::update_datafusion_session` after deprecation window

### DIFF
--- a/crates/core/src/delta_datafusion/session.rs
+++ b/crates/core/src/delta_datafusion/session.rs
@@ -49,6 +49,9 @@ pub(crate) trait DeltaSessionExt: DataFusionSession {
     /// This is primarily a migration helper for legacy DataFusion integrations that expect
     /// table-relative paths.
     ///
+    /// Only non-migrated/legacy operations should rely on this behavior; the newer table provider
+    /// path does not rely on these internal/special `delta-rs://...` URLs.
+    ///
     /// This does not support fully-qualified file URLs (e.g. shallow clones). Prefer
     /// `ensure_object_store_registered` in new code.
     ///
--- a/crates/core/src/logstore/mod.rs
+++ b/crates/core/src/logstore/mod.rs
@@ -374,6 +374,10 @@ pub trait LogStore: Send + Sync + AsAny {
     /// registering/fetching. In our case the scheme is hard-coded to "delta-rs", so to get a unique
     /// host we convert the location from this `LogStore` to a valid name, combining the
     /// original scheme, host and path with invalid characters replaced.
+    ///
+    /// This is a legacy/migration helper for delta-rs DataFusion integrations that use a
+    /// synthetic per-table `delta-rs://...` URL mapping to a table-scoped (prefixed) object store.
+    /// It will not work correctly with fully-qualified file URLs (e.g. shallow clones).
     fn object_store_url(&self) -> ObjectStoreUrl {
         crate::logstore::object_store_url(&self.config().location)
     }


crates/core/src/delta_datafusion/session.rs
```diff
@@ -10,13 +11,216 @@ use datafusion::{
...
+use crate::delta_datafusion::engine::AsObjectStoreUrl;
 use crate::delta_datafusion::planner::DeltaPlanner;
+use crate::errors::{DeltaResult, DeltaTableError};
+use crate::logstore::LogStore;
 
 pub fn create_session() -> DeltaSessionContext {
     DeltaSessionContext::default()
 }
 
+pub trait DeltaSessionExt: DataFusionSession {
```
roeap: can we make this `pub(crate)` for now?

I do believe we should have this pub asap, but while these are only crate scoped, we have more freedom to break things!

crates/core/src/delta_datafusion/session.rs
```diff
@@ -10,13 +11,216 @@ use datafusion::{
...
+where
+    T: DataFusionSession + ?Sized,
+{
+    fn ensure_object_store_registered(
+        &self,
+        log_store: &dyn LogStore,
+        operation_id: Option<Uuid>,
+    ) -> DeltaResult<()> {
+        let url = log_store.root_url().as_object_store_url();
+        if self.runtime_env().object_store(&url).is_err() {
```
roeap, line 80: You may be able to drop the `?Sized` constraint by ...

```suggestion
        if T::runtime_env(self).object_store(&url).is_err() {
```

crates/core/src/delta_datafusion/session.rs
```diff
@@ -10,13 +11,216 @@ use datafusion::{
...
+    tracing::warn!(
+        operation = ctx.operation,
+        table_uri = ?table_uri,
+        session_id = %session.session_id(),
+        cdc = ctx.cdc,
+        "Provided DataFusion Session is not a SessionState; falling back to internal defaults. This may ignore the caller's runtime/config/execution props/UDF registries/caches. Catalogs are not transferable via the Session trait. See delta-io/delta-rs#4081."
+    );
+}
+
+fn derive_session_state_from_trait(
```
roeap, line 176: nice, this has been a challenge to do this cast.

Ultimately I do hope that our operators will not rely on `SessionState`, but this is definitely MUUCH better that trying to cast things all over the place.

hntd187:
I like the idea behind this, very cool. @roeap weren't we going to eventually do something with the datafusion session such that we might not need to do this much resolution on it?

crates/core/src/delta_datafusion/session.rs
```diff
@@ -10,13 +11,216 @@ use datafusion::{
...
+    /// URL. If the session already has a (stale/incorrect) object store registered for the table
+    /// URL, it is up to the caller to update it via `RuntimeEnv::register_object_store`.
+    fn ensure_object_store_registered_for_table(
+        &self,
+        table: &crate::DeltaTable,
+        operation_id: Option<Uuid>,
+    ) -> DeltaResult<()> {
+        let log_store = table.log_store();
+        self.ensure_object_store_registered(log_store.as_ref(), operation_id)
+    }
```
roeap: Can we do without this one? 

Conceptually I believe `DeltaTable` should move more and more towards something like `DataFrame` in DF, which is naturally tied to a session. So it feels a bit awkward to update the seesion with that. 

API wise callers would need to then use the `LogStore` route, but that's just one method?

roeap (commented):
Just flushing some initial comments.

Really looking forward to getting our sessions straight, and this is a big step!

While writing https://github.com/delta-io/delta-rs/issues/4138 I started wondering about something more fundamental. I.e. where should we depend on `Session` and where can we use `SessionState` and `SessionContext`?

IF we expose an API that can consume `DataFrame`, then we will also always have `SessionState` in our dependencies.

Folks who implement their own session would probably(?) not go through the top level api, but rather build plans using the lower level stuff like `TableProvider`.

If we end up in a place where our builders can produce logical plans  they are then free to execute that using their session and we should not.

To be clear, we should still push put `SessionState` from the operations!

Just some thoughts that came up :).

crates/core/src/delta_datafusion/expr.rs
```diff
@@ -194,30 +194,34 @@ pub(crate) struct DeltaContextProvider<'a> {
 }
 
 impl<'a> DeltaContextProvider<'a> {
-    fn new(session: &'a dyn Session) -> Self {
-        // default planners are [CoreFunctionPlanner, NestedFunctionPlanner, FieldAccessPlanner,
-        // UserDefinedFunctionPlanner]
+    fn new(session: &'a dyn Session) -> DeltaResult<Self> {
```
roeap: nit:
```suggestion
    fn try_new(session: &'a dyn Session) -> DeltaResult<Self> {
```

crates/core/src/delta_datafusion/session.rs
ethan-tyler: Extension needs to work on `&dyn Session` (e.g. `DeltaTable::update_datafusion_session`), impl must support unsized receivers. UFCS is fine but doesn't remove the `?Sized` requirement unless we add `impl DeltaSessionExt for dyn DataFusionSession` or change the shim to take concrete `SessionState`/`SessionContext`.

crates/core/src/delta_datafusion/session.rs
```diff
@@ -10,13 +11,205 @@ use datafusion::{
...
+    /// Delta-rs registers a per-table object store URL (using the `delta-rs://...` scheme) so
+    /// DataFusion can resolve table-relative paths consistently across storage backends.
+    ///
+    /// This method is idempotent and will not overwrite an existing object store mapping for the
+    /// table's delta-rs object store URL.
+    fn ensure_log_store_registered(&self, log_store: &dyn LogStore) -> DeltaResult<()> {
+        let object_store_url = log_store.object_store_url();
+        if self.runtime_env().object_store(&object_store_url).is_err() {
+            self.runtime_env()
+                .register_object_store(object_store_url.as_ref(), log_store.object_store(None));
```
roeap, line 64: As you mention, the `object_store` method returns a pre-fixed object store that points to a path inside the root store and we create this cryptic URL. However this will never work properly if we have delta logs with fully qualified URLs (e.g. shallow clones). So I personally consider this a deprecated behaviour.

Not sure how to best do this, but I believe we should either directly deprecate this method (mark it as migration helper) or keep `register_store` around and deprecate that?

crates/core/src/delta_datafusion/session.rs
```diff
@@ -10,13 +11,205 @@ use datafusion::{
...
+                default_state(),
+                ResolvedSessionStateKind::InternalDefaultsIncompatibleSession,
+            ))
+        }
+        SessionFallbackPolicy::DeriveFromTrait => {
+            let derived = derive_session_state_from_trait(ctx, provided);
+            Ok((derived, ResolvedSessionStateKind::DerivedFromTrait))
+        }
+        SessionFallbackPolicy::RequireSessionState => Err(DeltaTableError::generic(format!(
+            "{operation}: provided DataFusion Session (session_id={session_id}) is not a SessionState. To fix: pass a concrete SessionState (e.g. Arc::new(create_session().state()) or Arc::new(session_ctx.state())). Catalogs are not transferable via the Session trait. See delta-io/delta-rs#4081.",
```
roeap: can we break this and the other long messages up in more lines. These long strings used to (or may still do) break `cargo fmt` on my machine :).

crates/core/src/operations/write/mod.rs
```diff
@@ -250,12 +255,28 @@ impl WriteBuilder {
...
-    /// The Datafusion session state to use
+    /// Set the DataFusion session used for planning and execution.
+    ///
+    /// The provided `session` should wrap a concrete `datafusion::execution::context::SessionState`.
+    ///
+    /// If `session` is not a `SessionState`, the default policy is to log a warning and fall back to
+    /// internal defaults. To make this strict (error instead), set
+    /// `with_session_fallback_policy(SessionFallbackPolicy::RequireSessionState)`.
+    ///
+    /// Example: `Arc::new(create_session().state())`.
```
roeap, line 266: seeing this made me wonder if `create_session` should just return the DF session? Do we even invoke any method from our session wrapper ever? I feel like we almost always immediately convert this to a DF session?

crates/core/src/delta_datafusion/session.rs
roeap: only non-migrated operations should rely on this behavior. The new table provider does not make use of these internal / special.

Not sure if this already landed, but in my more experimental code, we are building the log store from the session (actually `TaskContext`) and pass it in our code. deep in the log handling, a scoped store is still convenient. so i.e. 

```
session + table__url = log_store
```

crates/core/src/delta_datafusion/session.rs
roeap: let's just keep it like this :)

crates/core/src/delta_datafusion/session.rs
ethan-tyler: Done!

crates/core/src/delta_datafusion/expr.rs
ethan-tyler: Done!

PR #4146, opened 2026-01-29
[RFC] use datafusion sql parsing

roeap:
# Description

After some recent discussions, it turns out, that we may have to look at how we want to use `SessionState` - a core challenge being SQL parsing is not possible with just `dyn Session`.

Then I also wanted to know if we still need to have this custom parsing code.

So my question is, should we in some places just go ahead and depend on `SessionState` after all, and just have everything after we created logical plans (or at least parsed all we need into `Expr`s) be `Session` only?

### DIFF
--- a/crates/core/src/delta_datafusion/expr.rs
+++ b/crates/core/src/delta_datafusion/expr.rs
@@ -21,247 +21,16 @@
 
 //! Utility functions for Datafusion's Expressions
 use std::fmt::{self, Display, Error, Formatter, Write};
-use std::marker::PhantomData;
-use std::sync::Arc;
 
-use arrow_array::{Array, GenericListArray};
-use arrow_schema::{DataType, Field};
 use chrono::{DateTime, NaiveDate};
 use datafusion::catalog::Session;
-use datafusion::common::Result as DFResult;
-use datafusion::common::{DFSchema, Result, ScalarValue, TableReference, config::ConfigOptions};
+use datafusion::common::{DFSchema, Result, ScalarValue};
 use datafusion::execution::context::SessionState;
-use datafusion::execution::session_state::SessionStateBuilder;
-use datafusion::functions_nested::make_array::MakeArray;
-use datafusion::functions_nested::planner::{FieldAccessPlanner, NestedFunctionPlanner};
 use datafusion::logical_expr::expr::InList;
-use datafusion::logical_expr::planner::ExprPlanner;
-use datafusion::logical_expr::{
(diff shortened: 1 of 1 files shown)


roeap:
@rtyler @ion-elgreco - I think we needed all this custom parsing to have inner fields named a specific way. Is this something for us internally, or just a service to external callers?

ion-elgreco:
> @rtyler @ion-elgreco - I think we needed all this custom parsing to have inner fields named a specific way. Is this something for us internally, or just a service to external callers?

It's to create compliant parquet files

roeap:
> It's to create compliant parquet files

Datafusion is writing non-compliant parquet files?

ion-elgreco:
> > It's to create compliant parquet files
> 
> Datafusion is writing non-compliant parquet files?

Not datafusion but the parquet crate just writes arrow data as is, so the arrow data needs to have a compliant schema already

roeap:
> Not datafusion but the parquet crate just writes arrow data as is, so the arrow data needs to have a compliant schema already

this was around `item` vs `element` for the internal names in lists? What would i need to assert? that the arrow data read from a written parquet file has the expected inner field name?

ion-elgreco:
> > Not datafusion but the parquet crate just writes arrow data as is, so the arrow data needs to have a compliant schema already
> 
> this was around `item` vs `element` for the internal names in lists? What would i need to assert? that the arrow data read from a written parquet file has the expected inner field name?

This is mostly relevant when writing, whatever incorrect schema we read we have to cast it correctly before writing

ethan-tyler:
Thanks for putting this together @roeap! Here are my thoughts:

I like this approach as it aligns with the “narrow waist” / session consolidation direction we’ve discussed before and avoids us carrying custom Tokenizer/SqlToRel wiring.

A couple invariants/contract questions:

1. **Session contract:** this now effectively requires `SessionState` (downcast from `&dyn Session` at `crates/core/src/delta_datafusion/expr.rs:41`). Is that the intended public API?

- If yes, I’d rather make it explicit (take `&SessionState`, or clearly error) so we don’t repeat the “caller session gets ignored / surprising behavior” problem we’ve been tracking. 
- If no, we should define an explicit resolution/fallback policy rather than an implicit downcast.

3. **Parquet/Delta schema compliance:** the removed `MakeParquetArray` / `plan_array_literal` override existed to force list inner field name `element` (vs Arrow/DataFusion default `item`). Since parquet writes Arrow schema “as is”, can we point to where we normalize this today (e.g. write time cast), and add a regression test proving we still write/read compliant schemas?

Also: `create_logical_expr` uses the session’s SQL parser options (dialect / ident normalization). Should predicate parsing honor caller config, or enforce Delta defaults?

PR #4150, opened 2026-01-31
fix(delete): use Add metadata for partition only DELETE

### LINKED ISSUE #4149: [Bug]: Delete regression in 1.4.0: partition‑only predicate removes too few files
### What happened?

Delete file counts regressed between 1.2.1 and 1.4.0 for a partition‑only predicate. In 1.4.0, the scan‑based matching path appears to return a much smaller matched file set than the partition‑metadata path. This is only a direction/hypothesis, not a confirmed root cause.

Observed metrics:
- 1.2.1: {"num_added_files": 0, "num_removed_files": 7703, "num_deleted_rows": 0, "num_copied_rows": 0, "execution_time_ms": 59, "scan_time_ms": 56, "rewrite_time_ms": 2}
- 1.4.1: {"num_added_files": 0, "num_removed_files": 641, "num_deleted_rows": 0, "num_copied_rows": 0, "execution_time_ms": 7086, "scan_time_ms": 3677, "rewrite_time_ms": 0}

i have also confirmed that actual files to be deleted are accurate with 1.2.1

### Expected behavior

Delete with a partition‑only predicate (dt < '2025‑11‑13') should remove all files in matching partitions consistent with 1.2.1 behavior and partition metadata counts. 

scan_files_where_matches appears to use a scan‑based matching path even when the predicate is partition‑only. I think there something wrong going on when it finds the files that matches row that satisfy the predicate

### Operating System

None

### Binding

Python

### Bindings Version

1.4.0

### Steps to reproduce

```python

# table partitioned by dt, hour
# predicate is partition-only
print(table.delete(predicate="dt < '2025-11-13'"))
```

### Relevant logs

```shell

```

ethan-tyler:
# Description
Fix regression where `DELETE` with partition only predicates failed to remove empty files in matching partitions.

**Root Cause:**
When the delete predicate references only partition columns, file removal should be decided from log metadata alone. Prior implementation relied on scan derived results and empty files (zero rows) produced no matches. They weren't removed even though their partition satisfied the predicate.

**Changes:**
- Add partition only delete fast path: evaluates predicate against `Add.partition_values` (metadata only)
- Refine partition only predicate detection, remove redundant validation
- Add `remove_from_add` helper for removing files via their `Add` actions
- Add helper for partition predicate file finding

**Tests:**
Regression coverage for:
- Partition only deletes remove empty files in matching partitions (repro for #4149)
- NULL partition values handled correctly
- Partition-only path avoids unnecessary scanning
```bash
cargo fmt --check
cargo test --workspace
cargo test -p deltalake-core --features datafusion
```

# Related Issue(s)
- Fixes #4149

# Documentation

Notes

- Scoped to partition only predicates; intended semantics preserved
- NULL partition handling explicitly tested

### DIFF
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -410,10 +410,6 @@ pub(crate) struct MatchedFilesScan {
     /// we do not (yet) have full coverage to translate datafusion to
     /// kernel predicates.
     pub(crate) delta_predicate: Arc<Predicate>,
-    /// The predicate contains only partition column references
-    ///
-    /// This implies that for each matched file all data matches.
-    pub(crate) partition_only: bool,
 }
 
 impl MatchedFilesScan {
@@ -467,9 +463,8 @@ pub(crate) async fn scan_files_where_matches(
         result: Ok(()),
     };
     for term in &skipping_pred {
-        visitor.result = Ok(());
         term.visit(&mut visitor)?;
-        visitor.result?
+        std::mem::replace(&mut visitor.result, Ok(()))?;
     }
 
     // convert to a delta predicate that can be applied to kernel scans.
@@ -554,7 +549,6 @@ pub(crate) async fn scan_files_where_matches(
         valid_files,
         predicate,
         delta_predicate,
-        partition_only: visitor.partition_only,
     }))
 }
 
--- a/crates/core/src/operations/delete.rs
+++ b/crates/core/src/operations/delete.rs
@@ -378,9 +378,8 @@ async fn execute(
         result: Ok(()),
     };
     for term in &skipping_pred {
-        props.result = Ok(());
         term.visit(&mut props)?;
-        props.result?;
+        std::mem::replace(&mut props.result, Ok(()))?;
     }
 
     if props.partition_only {
@@ -466,15 +465,6 @@ async fn execute(
         .try_collect()
         .await?;
     metrics.num_removed_files = removes.len();
-
-    if files_scan.partition_only {
-        // Note: Partition-only predicates are handled earlier using log metadata.
-        // This branch is kept defensively in case `scan_files_where_matches` ever determines
-        // a predicate is partition-only after simplification.
-        // if we are deleting entire files only, no need to rescue any data or write cdc files.
-        metrics.execution_time_ms = Instant::now().duration_since(exec_start).as_millis() as u64;
-        return Ok((removes, metrics));
-    }
 
     let counted_scan = LogicalPlan::Extension(Extension {
         node: Arc::new(MetricObserver {


crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -113,6 +113,19 @@ pub(crate) async fn find_files(
...
+/// Assumes the caller already validated that `predicate`:
+/// - references only partition columns
+/// - is deterministic
+pub(crate) async fn find_files_by_partition_predicate(
+    snapshot: &EagerSnapshot,
+    predicate: &Expr,
+) -> DeltaResult<Vec<Add>> {
+    scan_memory_table(snapshot, predicate).await
+}
+
```
roeap: here is where the above comment should have gone :)

ion-elgreco:
@roeap I think we should start doing benchmarks moving forward with these PRs, also with optimize for example. These issues show that we are regressing in scan speed and overall execution speed

roeap:
Agreed, will try and set something up.

The main regression in speed is likely thought since we recently disabled use of the cached action even for EagerSnapshot. So every call to file_views does a full log replay and we use that quite heavily in operations.

roeap:
@ion-elgreco - do you habe an example perf regression bug?

This one is corretness only?

vsmanish1772:
> @ion-elgreco - do you habe an example perf regression bug?
> 
> This one is corretness only?

@roeap 
Would this comment suffice an example? I mean the execution_time_ms between 1.2.1 and 1.4.0 is huge. 
https://github.com/delta-io/delta-rs/issues/4149#issuecomment-3825201497

roeap:
Absolutely!! Thanks @vsmanish1772!

ethan-tyler:
Did some digging and here's what I'm seeing:

In `python-v1.2.1`, DELETE calls `find_files(...)` which detects partition only predicates and evaluates them against an in-memory table built from Add-action metadata (paths and `partition.*` columns). Returns matching files without touching parquet.

In `python-v1.4.0`, the DML/delete migration (#4117) switched DELETE to call `scan_files_where_matches(...)` even for partition only predicates. That helper finds files by running `.filter(predicate)` then `project(file_id)` then `distinct()`.

**Why I think this causes the regression**

The `distinct(file_id)` step requires reading parquet to produce rows, even though partition only predicates can be answered from log metadata alone. This turned a metadata-only operation into parquet I/O and aggregation and explains the scan time jump.

Also explains "removes too few files": empty parquet files produce no rows, they never appear in `distinct(file_id)` and get skipped even though their partition values match.

Partition only predicates regressed because they now take the scan based path instead of the Add action metadata path.

Does this match what you're seeing? @roeap @ion-elgreco

ion-elgreco:
Why are we reading parquet? The logical plan is just a scan of the json files right?

ethan-tyler:
> Why are we reading parquet? The logical plan is just a scan of the json files right?

No - the scan reads `_delta_log` only to build the file list and metadata (and maybe prune via file skipping), but the DataFusion `scan(...)` still executes a `ParquetExec` over data files to produce rows. `filter(predicate)` then `distinct(file_id)` needs actual row evaluation unless we short circuit on partition metadata.

That's also why empty files get missed- they produce no rows and they never appear in `distinct(file_id)` even though their partition values match.

roeap:
@ethan-tyler - maybe ... The new table provider falls back to metadata only scans internally if neither the predicate nor the projection ask for non-metadata columns... which may kick in in this case or not.

The most obvious reason I would investigate first is [this](https://github.com/delta-io/delta-rs/commit/fd79507cc73218999eb14fb205f8be39c4aba7e5) commit. There we effectively disable the use of pre-loaded log data. This means we now do up to 3 full log replays from scratch in DELETE including parsing stats twice per replay - prior it would have been one, but still doing some redundant stats serde.

But if there is one thing I learned, when it comes to performance, gut feelings don't mean much and we need to validate this 😄.

Hoping I can support in this later today.

ethan-tyler:
Good point - validated this locally.

The physical plan for `scan_files_where_matches` (`filter` then `project(file_id)` then `distinct()`) does not use `ParquetSource/DataSourceExec` in the partition only case. It uses `DeltaScanMetaExec` (metadata only), my earlier claim isn't accurate for this shape.

That said, I still think the regression comes from the same high level change where we moved from directly filtering `Add.partitionValues` (old `find_files` path) to a DF plan that derives matches from rows, even if those rows are synthesized from stats.

**Correctness**: metadata only execution still yields 0 rows for empty files (`numRecords=0`), they never contribute a `file_id` to `distinct(file_id)` and still get missed. Matches the "removes too few files" symptom.

**Performance**: even metadata only, we're executing a DF plan with repartition, aggregate, and stats parsing, plus potential log replay depending on snapshot/cache usage. Consistent with the reported `scan_time_ms` jump, but should confirm exact replay count with tracing.

roeap:
@ethan-tyler @vsmanish1772 @ion-elgreco - I tried reproducing this bug [here](https://github.com/delta-io/delta-rs/pull/4158), but so far to no avail. I was thinking lets fix the bug first and then look into performance? The linked PR tests delete with and without partition only predicates as well as partitioning on a timestamp column. 

@vsmanish1772 - could you provide some more context for the repro-case? maybe even a small script?

One suspicion might be that our path matching is tripping up due to maybe something we would only see in the prefix?

vsmanish1772:
> @ethan-tyler @vsmanish1772 @ion-elgreco - I tried reproducing this bug [here](https://github.com/delta-io/delta-rs/pull/4158), but so far to no avail. I was thinking lets fix the bug first and then look into performance? The linked PR tests delete with and without partition only predicates as well as partitioning on a timestamp column.
> 
> @vsmanish1772 - could you provide some more context for the repro-case? maybe even a small script?
> 
> One suspicion might be that our path matching is tripping up due to maybe something we would only see in the prefix?

Hmm, not really sure how do i provide more context. I can mention few points from our discussions
1) All i did was generate some data for 30 days partitioned on `dt` and `hour` and run delete()
2) There are no empty tables/files in my data set, every file has some data for sure.
2) We partition data based of dt and hour column
3) The path of my table looks like `s3://manishsogiyawar-test/data/id=SOME_UUID/stage=foo/tbl=blah/`
and then there is `delta_log` and folder like `dt=2025-12-01` (partitioned data) under that path

ion-elgreco:
@vsmanish1772 can you give us an MRE? So we are running the same code as you are

roeap:
If it's just generated data, any chance you can share a script that shows the bahvior?

Maybe its also somehow related to our expression parsing? I'll dive deeper and see what i can find ...

vsmanish1772:
Sure il try to get something up. Generated data in the sense i have some data in dev for 30 days that was sitting there. Il try to get a script that generates same data with fakes values and run deletion on it.

Another thing to note. we use 1.1.4 for append writes and i was using 1.4.0 to delete the data

vsmanish1772:
@ethan-tyler @roeap i have pulled this pr in my local branch and can confirm the files removed after change matches with 1.2.1 and 1.3.2 version
The correctness might have been solved but performance is still the same

`{'num_added_files': 0, 'num_removed_files': 33189, 'num_deleted_rows': 0, 'num_copied_rows': 0, 'execution_time_ms': 7977, 'scan_time_ms': 7977, 'rewrite_time_ms': 0}
(deltalake) msogiyawar@MAC-BFFF3B delta-rs % uv pip install deltalake==1.3.2 
Using Python 3.14.2 environment at: python/.venv
Resolved 4 packages in 327ms
Uninstalled 1 package in 2ms
Installed 1 package in 5ms
- deltalake==1.4.1 (from file:///Users/msogiyawar/Workspace/delta-rs/python)
+ deltalake==1.3.2
(deltalake) msogiyawar@MAC-BFFF3B delta-rs % /Users/msogiyawar/Workspace/delta-rs/python/.venv/bin/python /Users/msogiyawa
r/Workspace/delta-rs/test.py
30746
{'num_added_files': 0, 'num_removed_files': 33189, 'num_deleted_rows': 0, 'num_copied_rows': 0, 'execution_time_ms': 61, 'scan_time_ms': 52, 'rewrite_time_ms': 6}`

crates/core/src/operations/delete.rs
```diff
@@ -359,14 +362,88 @@ async fn execute(
...
+        .metadata()
+        .partition_columns()
+        .clone();
+    let mut props = crate::delta_datafusion::FindFilesExprProperties {
+        partition_columns,
+        partition_only: true,
+        result: Ok(()),
+    };
+    for term in &skipping_pred {
+        props.result = Ok(());
```
ion-elgreco: Why is this needed? Won't it already fail on the first err?

crates/core/src/operations/delete.rs
```diff
@@ -391,6 +468,9 @@ async fn execute(
     metrics.num_removed_files = removes.len();
 
     if files_scan.partition_only {
+        // Note: Partition-only predicates are handled earlier using log metadata.
+        // This branch is kept defensively in case `scan_files_where_matches` ever determines
+        // a predicate is partition-only after simplification.
         // if we are deleting entire files only, no need to rescue any data or write cdc files.
         metrics.execution_time_ms = Instant::now().duration_since(exec_start).as_millis() as u64;
         return Ok((removes, metrics));
```
ion-elgreco: Hmm can you find a test that ever hits here?

crates/core/src/operations/delete.rs
```diff
@@ -459,6 +539,84 @@ async fn execute(
...
+    use arrow_schema::Schema;
+    use datafusion::logical_expr::LogicalPlanBuilder;
+    use datafusion::logical_expr::col;
+
+    use crate::delta_datafusion::PATH_COLUMN;
+    use crate::errors::DeltaTableError;
+    use datafusion::datasource::{MemTable, provider_as_source};
+    use datafusion::physical_plan::collect;
+
+    let batch = snapshot.add_actions_table(true)?;
```
ion-elgreco, line 550: In case @vsmanish1772 code is merged after, do you mind doing a follow up that uses the add_action_batches?

crates/core/src/operations/delete.rs
ethan-tyler: You’re right, that reset is redundant. Removing it, thanks for the catch!

crates/core/src/operations/delete.rs
ethan-tyler: I couldn’t find a test that reaches that branch, it looks unreachable after the earlier partition only fast path. I’ll remove it.

crates/core/src/operations/delete.rs
ethan-tyler: aye aye captain

vsmanish1772: You guys still working at this time!!! 
Yes sure will do @ion-elgreco

PR #3707, opened 2025-08-25
feat(url): use Url in Rust for accessing to DeltaTable, use only string-based api in Python

fvaleye:
# Description
Use the URL crate in Rust for accessing the DeltaTable. Use only the string-based API in Python.
- Rust APIs now take `url::Url` for opening/creating tables.
- Python keeps str at the boundary and parses to URL immediately.
- String-based Rust helpers still exist and are marked deprecated (for migration).

# Related Issue(s)
- closes #3615 

# Documentation
- Use [URL](https://docs.rs/url/latest/url/) in Rust instead of using `&str`
- Keep `&str` for the Python binding with helper methods
- Keep backward compatibility by making the URI deprecated.

**Side notes:** 💚 
@rtyler, I started contributing to delta-rs with my first [PR](https://github.com/delta-io/delta-rs/commit/35ae40949116d32b43fa71c7a156e2f3c8cac797) regarding URI.
So, as a "comeback", I'm taking this opportunity to improve it 😄

Issue #3615, opened 2025-07-23
Adopt url::Url based interfaces for referring to tables


rtyler:
We have a substantial amount of complexity for handling string encodings of paths of all flavors which stem from an original sin of `deltalake::open_table` accepting `&str`. 

We can simplify a lot of our testing and internals by simply removing this option entirely, as I have discussed with @roeap , and requiring our public interfaces to accept `url::Url`. This would ensure that the end-user is responsibly constructing the URLs they need for referring to Delta tables of various forms.

From the Python standpoint, I think it is fine to still accept `str` Python types _but_ those should be parsed immediately as `url::Url` types before passing into the core layer.

PR #4154, opened 2026-02-01
fix(docs): open_table examples to use URLs instead of str

khalidmammadov:
# Description

This is to fix all examples and code samples of open_table helper function where it's still using str as input. 

- Whereever `Url::from_directory_path` is used I tried use absolute path to make inline with URL API requirements.
- In other places I also used Path and derived cononical path to satisfy again URL requirements.
- Some other places I just used plain variable replacement i.e. "tmp" -> delta_path to preserve simplicity

# Related Issue(s)
https://github.com/delta-io/delta-rs/issues/3615

https://github.com/delta-io/delta-rs/pull/3707

# Documentation

Tested locally via project toolset and imperically checked via serving in browser.


khalidmammadov:
@rtyler can you please take a look one more time?

Issue #4160, opened 2026-02-03
[Bug]: Rust CdfLoadBuilder.build function is crate private and prohibits using feature


khalidmammadov:
### What happened?

In Rust API CdfLoadBuilder.build function is declared crate private and prohibits using feature.

And hence example provided in the docs for usage is not working i.e.:

    let cdf = table.scan_cdf()
        .with_starting_version(0)
        .with_ending_version(4)
        .build(&ctx.state(), None) <----Private
        .await?;

### Expected behavior

It should be accessible to be able to construct plan

### Operating System

None

### Binding

None

### Relevant logs

```shell

```

khalidmammadov:
take

PR #4161, opened 2026-02-03
fix(cdf): make cdf builders build function accessible

khalidmammadov:
# Description

In Rust API CdfLoadBuilder.build function is declared crate private and prohibits using feature in other Rust projects.

Although below example works within delta-rs crate it's not working for external projects:

docs/src/rust/read_cdf.rs
```
... 
let cdf = table.scan_cdf()
    .with_starting_version(0)
    .with_ending_version(4)
    .build(&ctx.state(), None) <----Private
    .await?;

...
```

# Related Issue(s)

https://github.com/delta-io/delta-rs/issues/4160

# Documentation

Updated example that goes into docs and locally tested. 
Also, current test cases covers actual functionality.

### DIFF
--- a/crates/aws/Cargo.toml
+++ b/crates/aws/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-aws"
-version = "0.13.0"
+version = "0.13.1"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -69,7 +69,8 @@ native-tls = [
 ]
 rustls = [
     "aws-config/client-hyper",
-    "aws-config/rustls",
-    "aws-sdk-dynamodb/rustls",
-    "aws-sdk-sts/rustls",
+    # Use the new AWS SDK default HTTPS client stack (hyper 1.x + rustls via aws-lc-rs),
+    # not the legacy `rustls` feature which pins us to hyper 0.14.
+    "aws-sdk-dynamodb/default-https-client",
+    "aws-sdk-sts/default-https-client",
 ]
--- a/crates/catalog-glue/Cargo.toml
+++ b/crates/catalog-glue/Cargo.toml
@@ -17,8 +17,14 @@ aws-config = { version = "1.8", default-features = false, features = [
     "behavior-version-latest",
     "rt-tokio",
     "credentials-process",
+    "client-hyper",
 ] }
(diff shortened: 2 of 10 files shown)


hntd187:
I'll be honest the interface for CDF has been the same for a long time. Is there a reason you want to change it?

khalidmammadov:
> I'll be honest the interface for CDF has been the same for a long time. Is there a reason you want to change it?

I think I was misguided by this `docs/src/rust/read_cdf.rs` example in the docs which is not updated for the newer implementation where CDF Builder interface is partially exposed. Also, I think it's odd that it's CDF configuration API called builder that implies user can call build() which is typical pattern, TBH I would call it something like `CdfScanOptions` instead to convey clarity.  

Anyway, I will update PR with docs example fix only to use uptodate API.

PS: Thanks for looking at and pointing out!

roeap:
@khalidmammadov - i am not 100% sure, but I think I might have added that method recently and deliberately kept it `pub(crate)` for now.

The main reason was that there is some uncertainty as to how we want to evolve our operations APIs. As @rtyler and others also point out, the `IntoFuture` pattern we have been using is somewhat limiting, as we need to manage the session alongside the operations, which is unintuitive if we think of this more in terms of logical planning.

I guess what I am saying is, that from my end it would be fine to expose that method, with the explicit caveat, that this might look very different as we re-work some of the operations APIs.

hntd187:
> @khalidmammadov - i am not 100% sure, but I think I might have added that method recently and deliberately kept it `pub(crate)` for now.
> 
> The main reason was that there is some uncertainty as to how we want to evolve our operations APIs. As @rtyler and others also point out, the `IntoFuture` pattern we have been using is somewhat limiting, as we need to manage the session alongside the operations, which is unintuitive if we think of this more in terms of logical planning.
> 
> I guess what I am saying is, that from my end it would be fine to expose that method, with the explicit caveat, that this might look very different as we re-work some of the operations APIs.

Right and I'd prefer as few breaking changes as possible as it's harder to take something away than it is to just open it up. @khalidmammadov I appreciate you updating the examples though it's been a much needed thing through this entire process.

rtyler:
a colleague of mine recently ran into this too, thanks for picking it up. I'm going to bring it back to 0.30.2 too

PR #4166, opened 2026-02-05
test: reproduce partition only under delete with empty file

ethan-tyler:
# Description
Reproduces the error in #4149. Partition only DELETE can remove too few files when a matching partition includes a 0-row (empty) Parquet file. This fails on main and passes with the fix in #4150.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/operations/delete.rs
+++ b/crates/core/src/operations/delete.rs
@@ -1463,4 +1463,150 @@ mod tests {
         assert_eq!(table.version(), Some(2), "The delete failed to execute");
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_delete_partition_only_removes_empty_files_in_matching_partitions()
+    -> DeltaResult<()> {
+        use std::collections::HashMap;
+        use std::sync::Arc;
+
+        use arrow_array::{Int32Array, RecordBatch, StringArray};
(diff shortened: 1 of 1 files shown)


roeap:
I still feel like an empty file should not be written in the first place. and since there is no data to delete, I think it's consistent that this file is also not touched.

ion-elgreco:
> I still feel like an empty file should not be written in the first place. and since there is no data to delete, I think it's consistent that this file is also not touched.

That should be addressed in the writer somehow

roeap:
> That should be addressed in the writer somehow

in fact I believe #4119 might actually address this. IIRC there was a failing test with 0 matched rows, where stats were off and I updated after reviewing it.

ion-elgreco:
> > That should be addressed in the writer somehow
> 
> in fact I believe #4119 might actually address this. IIRC there was a failing test with 0 matched rows, where stats were off and I updated after reviewing it.

But which Part of that pr or addresses it? Then we can merge it isolated

roeap:
> But which Part of that pr or addresses it? Then we can merge it isolated

I thinks its just that more stuff is running through the V2 write code. So the fix is probably to use our current best writing consistently :).

Issue #4115, opened 2026-01-22
Next-scan ordering contracts and DV masking hazard mitigation


ethan-tyler:
### Is your feature request related to a problem?

#4112 fixed a correctness bug when DF coalesces batches across file boundaries, assuming stream order is preserved. However, @roeap pointed out we're still relying on ordering invariants for DV row position semantics and log replay that DF doesn't actually guarantee. 

## Problem

**DVs:**

`consume_dv_mask` assumes rows show up in the order they exist in the file,  it drains a per-file selection vector from the front. This works fine if:

1. Rows for a file come through in order (doesn't have to be contiguous, just monotonic)
2. A file doesn't get split across multiple streams/partitions

The DF optimizer can introduce plan shapes that reorder rows or split files across partitions. If it does, we'll silently apply DV bits to the wrong rows. Not great.

#4112 handles mixed `file_id` batches under order preserving coalesce, but does not protect against upstream operators that break monotonicity or split files across partitions.

**Log replay:**

This one's more theoretical right now, but if we ever push log replay into a DF plan we need global ordering by commit version for "last write wins" to work correctly. `required_input_ordering` only gives us per-partition ordering which isn't sufficient, we'd need either single partition execution or a sort preserving merge to get one globally ordered stream.

If multiple actions share a commit version, a deterministic tie breaker (e.g. action index) may also be needed.

## What I think the contract should be

For DVs: we need monotonic per-file row order and a file can't be split across streams. Interleaving is fine if monotonicity preserved because contiguity isn't required. We should reject plan shapes that imply reordering.

For log replay:  We need global total order by version. Local ordering must be paired with global merge or single-partition execution.

**Fail-fast**
I think we should default to fail-fast on detected violations. Only recover when we can prove order is preserved.

Things that should be fine:
- Batch coalescing (rows stay in order)
- Mixed-file batches that we split by file_id runs

Things that should fail-fast:
- Upstream operator that doesn't preserve order before DV application
- Same file_id showing up in multiple partitions
- repartition_file_scans with DVs active

Detecting "rows arrived out of order" at runtime is not reliable without row position metadata. We need to catch bad plan shapes before execution instead.

Maybe also worth adding a strict/debug mode that errors on file_id reappearance as extra defense? Not sure if that's overkill.

## Potential Solutions:

**Short term**
- Document these ordering contracts
- Add checks that fail on unsafe plan shapes when DVs are active
- Runtime check that `file_id` only shows up in one stream
[...]

PR #4167, opened 2026-02-05
fix: align file_id with DataFusion UInt16 dictionary

ethan-tyler:
# Description
The synthetic `file_id` column type was defined independently allowing silent drift between `Int32` and `UInt16`. This causes type coercion failures in DML paths that build `file_id IN (...)` predicates.

**Changes:**
Adds `file_id.rs` as single source of truth for the column name, data type,
field constructor, and dictionary aligned literal wrapper. All call sites now go
through these helpers.

Adds a guard to chunk `FileGroup`s so the per-group partition dictionary
cannot exceed the `UInt16` keyspace, with two counters
(`count_file_groups_planned`, `count_file_group_chunks`) to observe when
chunking triggers.

`file_id` remains an internal correlation mechanism. This reduces its surface
area and makes removal easier once `ParquetAccessPlan` based scan filtering
lands.

# Related Issue(s)

Related: 
- #4113
- #4115

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
@@ -38,7 +38,6 @@ use datafusion::{
         union::UnionExec,
     },
     prelude::Expr,
-    scalar::ScalarValue,
 };
 use datafusion_datasource::{
     PartitionedFile, TableSchema, compute_all_files_statistics, file_groups::FileGroup,
@@ -60,6 +59,7 @@ use crate::{
     delta_datafusion::{
         DeltaScanConfig,
         engine::{AsObjectStoreUrl as _, to_datafusion_scalar},
+        file_id::wrap_file_id_value,
     },
 };
 
@@ -198,7 +198,7 @@ async fn get_data_scan_plan(
             version: None,
         }
         .into();
-        let file_value = wrap_file_id_partition_value(f.file_url.as_str());
+        let file_value = wrap_file_id_value(f.file_url.as_str());
         // NOTE: `PartitionedFile::with_statistics` appends exact stats for partition columns based
         // on `partition_values`, so partition values must be set first.
         partitioned_file.partition_values = vec![file_value.clone()];
@@ -217,18 +217,6 @@ async fn get_data_scan_plan(
         .try_collect::<_, Vec<_>, _>()?
         .into_iter()
         .into_group_map();
-
-    // Track planned file-group chunking so we can observe if/when it triggers.
-    let planned_file_groups = files_by_store
-        .values()
-        .map(|files| files.len().div_ceil(MAX_PARTITION_DICT_CARDINALITY))
-        .sum::<usize>();
-    MetricBuilder::new(&metrics)
-        .global_counter("count_file_groups_planned")
-        .add(planned_file_groups);
-    MetricBuilder::new(&metrics)
-        .global_counter("count_file_group_chunks")
-        .add(planned_file_groups.saturating_sub(files_by_store.len()));
 
     // TODO(roeap); not sure exactly how row tracking is implemented in kernel right now
     // so leaving predicate as None for now until we are sure this is safe to do.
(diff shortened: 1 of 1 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -344,9 +379,15 @@ async fn get_read_plan(
                 .with_pushdown_filters(true);
         }
 
-        let file_group: FileGroup = files.into_iter().map(|file| file.0).collect();
+        // Partition values are dictionary encoded using a UInt16 key (DataFusion's default
+        // `wrap_partition_type_in_dict`). Keep file groups small enough that the file-id partition
+        // dictionary doesn't exceed the key space (one distinct value per file).
```
roeap: nit: this comment comes up several times, maybe just put it to where we use it (.chunks()). IMO, its such an edge case, that its is enough this explanation shows up once.

crates/core/src/delta_datafusion/file_id.rs
roeap, line 1: is this file better placed inside the scan module?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -291,6 +307,25 @@ fn update_partition_stats(
...
 
 type FilesByStore = (ObjectStoreUrl, Vec<(PartitionedFile, Option<Vec<bool>>)>);
+
+/// Maximum number of distinct values representable by DataFusion's default partition dictionary
+/// encoding (`Dictionary<UInt16, _>`).
+const MAX_PARTITION_DICT_CARDINALITY: usize = (u16::MAX as usize) + 1;
+
+fn partitioned_files_to_file_groups(
+    files: impl IntoIterator<Item = PartitionedFile>,
+    max_files_per_group: usize,
```
roeap: does this need to be a parameter? I.e. could we just always chunk by `MAX_PARTITION_DICT_CARDINALITY`?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -221,6 +218,21 @@ async fn get_data_scan_plan(
...
+        .values()
+        .map(|files| files.len().div_ceil(MAX_PARTITION_DICT_CARDINALITY))
+        .sum::<usize>();
+    MetricBuilder::new(&metrics)
+        .global_counter("count_file_groups_planned")
+        .add(planned_file_groups);
+    MetricBuilder::new(&metrics)
+        .global_counter("count_file_group_chunks")
+        .add(planned_file_groups.saturating_sub(files_by_store.len()));
+
```
roeap: are these interesting stats?

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -254,6 +266,10 @@ async fn get_data_scan_plan(
     Ok(Arc::new(exec))
 }
 
+fn wrap_file_id_partition_value(file_url: &str) -> ScalarValue {
+    crate::delta_datafusion::file_id::wrap_file_id_value(file_url)
+}
```
roeap: why do we need this wrapper?

ethan-tyler: No strong reason, I will remove it

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: These were mainly tracking an edge case and don’t add value. Will remove them.

crates/core/src/delta_datafusion/file_id.rs
ethan-tyler: `file_id.rs` is shared across `find_files.rs`, `table_provider.rs` (legacy path), and the next-provider scan/planning flow. Moving it under `table_provider/next/scan` would force non-scan code to depend on scan internals or duplicate the file-id logic. Keeping it at the `delta_datafusion` level avoids that coupling.

roeap (approved):

PR #4168, opened 2026-02-06
feat: expose DV metadata and payloads as Arrow streams

### LINKED ISSUE #4159: Expose deletion vectors information to Python
Polars OSS would like to support reading (and eventually writing) files with deletion vectors directly. To that end, we would like to see deletion vectors exposed through the Python bindings. Is that in the plan / can that be arranged? Thanks in advance.

xref https://github.com/pola-rs/polars/issues/26369

_Originally posted by @kdn36 in https://github.com/delta-io/delta-rs/issues/1094#issuecomment-3841064636_

ethan-tyler:
# Description
Adds `DeltaTable.deletion_vectors() -> RecordBatchReader` returning one row per data file with a deletion vector. 

Schema: `filepath: utf8`, `selection_vector: list[bool]` (true = keep, false = deleted).

Reuses the existing DataFusion replay path via `replay_deletion_vectors(...)`.  Results are deterministic and sorted by filepath.

**Core changes:**
- `DeletionVectorSelection` struct, `DeltaScan::deletion_vectors()`, shared
  `scan_metadata_stream()` helper to avoid drift between scan paths
- Replaced internal DV `expect(...)` with typed error propagation

**Python binding:**
- `cloned_table_and_state()` to avoid TOCTOU on table + snapshot
- Chunked Arrow batch output with non-nullable list items
- Preserves `without_files` guard behavior

# Related Issue(s)
- Closes #4159

# Documentation

cc @ion-elgreco

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
@@ -37,7 +37,7 @@ use pin_project_lite::pin_project;
 use url::Url;
 
 use crate::{
-    DeltaResult,
+    DeltaResult, DeltaTableError,
     delta_datafusion::{DeltaScanConfig, engine::to_datafusion_scalar},
     kernel::{
         LogicalFileView, ReceiverStreamBuilder, Scan, StructDataExt,
@@ -195,12 +195,11 @@ where
                     scan_data
                 };
 
-                let mut ctx = ScanContext::new(this.table_root.clone());
-                ctx = match scan_data.visit_scan_files(ctx, visit_scan_file) {
-                    Ok(ctx) => ctx,
-                    Err(err) => return Poll::Ready(Some(Err(err.into()))),
-                };
-                ctx = match ctx.error_or() {
+                let ctx = match scan_data
+                    .visit_scan_files(ScanContext::new(this.table_root.clone()), visit_scan_file)
+                    .map_err(|err| DataFusionError::from(DeltaTableError::from(err)))
+                    .and_then(ScanContext::error_or)
+                {
                     Ok(ctx) => ctx,
                     Err(err) => return Poll::Ready(Some(Err(err.into()))),
                 };
@@ -535,13 +534,11 @@ fn apply_file_selection(
             continue;
         }
 
-        let keep = parse_path(
+        let file_url = parse_path(
             table_root,
             LogicalFileView::new(batch.clone(), idx).path_raw(),
-        )
-        .map(|url| file_selection.contains(url.as_str()))
-        .unwrap_or(false);
-        selection_vector[idx] = keep;
+        )?;
+        selection_vector[idx] = file_selection.contains(file_url.as_str());
     }
 
     scan_data.scan_files =
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -29,7 +29,7 @@ use deltalake::delta_datafusion::{
 };
 
 use deltalake::arrow::array::{
-    ArrayRef, BooleanBuilder, ListBuilder, RecordBatchIterator, StringBuilder,
+    ArrayRef, BooleanBuilder, LargeStringBuilder, ListBuilder, RecordBatchIterator,
 };
 use deltalake::errors::DeltaTableError;
 use deltalake::kernel::scalars::ScalarExt;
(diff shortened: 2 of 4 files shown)


ion-elgreco:
@ethan-tyler thanks for picking this up!

I think we can simplify the pr quite abit. For Python we only want to have a recordbatchreadwr with two columns: filepath<str>, selection_vector<list[str]>

We only need a log replay as is done in here and then convert the dv_hashmap into a recordbatchreader:

https://github.com/delta-io/delta-rs/blob/main/crates%2Fcore%2Fsrc%2Fdelta_datafusion%2Ftable_provider%2Fnext%2Fscan%2Fmod.rs#L81-L82

ethan-tyler:
@ion-elgreco - Agreed on simplifying the surface and it was actually my first pass. After implementing and validating,  I saw real limitations with a `filepath + selection_vector[list[bool]]` boundary.

If this goes through DataFusion scan internals, that's an architectural mismatch for Polars and any other non-DF consumers. 
- cross engine API stability couples to DF internals/refactors
- semantics defined by one engine's execution path, not a neutral Delta boundary
- Inherent maintenance cost increases for non-DF consumers

IMO for cross engine consumers- a protocol/kernel level DV boundary (or binding owned transform) is a better feature.

Since we're already in the internals, I'd rather preserve this core and simplify UX on top:
- I've done the complex parts with snapshot validation, descriptor parsing, bounded IO/concurrency, DV decoding, error handling.
- We add a `filepath + dv_roaring_bytes` is a thin wrapper on current output.
- `filepath + selection_vector[list[bool]]` can layer on top by decoding/expanding, keeping the compact backend as the efficient path.

Work is reusable either way. Happy to split into smaller PRs if scope is the concern. Lmk your thoughts and happy to discuss further.

ion-elgreco:
> If this goes through DataFusion scan internals, that's an architectural mismatch for Polars and any other non-DF consumers. 

I dont really see how this is an issue. Consumers on the Python side receive a recordbatchreader including Polars. The dv hashmap is already materialized so we can actually even return an arrow table. How it's executed in rust should have no influence how Python consumers take in the deletion vectors

> 
> IMO for cross engine consumers- a protocol/kernel level DV boundary (or binding owned transform) is a better feature.
> 
> Since we're already in the internals, I'd rather preserve this core and simplify UX on top:
> - I've done the complex parts with snapshot validation, descriptor parsing, bounded IO/concurrency, DV decoding, error handling.
> - We add a `filepath + dv_roaring_bytes` is a thin wrapper on current output.
> - `filepath + selection_vector[list[bool]]` can layer on top by decoding/expanding, keeping the compact backend as the efficient path.
> 

The Python binding api surface should be minimal, this custom reader should never exist solely for the Python binding. That's why I'm suggesting to go through the replay_files.

ethan-tyler:
ok that's fair, I still think there’s value in this approach in a different context but I’m aligned. I’ll shelf this and push the simplified version​​​​​​​​​​​​​​​​. Thanks for the discussion, appreciate you engaging on this

crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
```diff
@@ -186,6 +186,10 @@ where
                     Ok(ctx) => ctx,
                     Err(err) => return Poll::Ready(Some(Err(err.into()))),
                 };
+                ctx = match ctx.error_or() {
+                    Ok(ctx) => ctx,
+                    Err(err) => return Poll::Ready(Some(Err(err.into()))),
+                };
```
ion-elgreco: Why is the same thing done here?

python/src/lib.rs
```diff
@@ -117,6 +122,62 @@ type StringVec = Vec<String>;
...
 
+/// Maximum number of file-level deletion vector entries per Arrow RecordBatch when returning
+/// results from `DeltaTable.deletion_vectors()`.  Each entry is one (filepath, selection_vector)
+/// pair, so this bounds memory per batch — not per row within a file's mask.
+const DELETION_VECTOR_BATCH_SIZE: usize = 1024;
+
+fn deletion_vector_schema() -> Arc<arrow::datatypes::Schema> {
+    use arrow::datatypes::{DataType, Field, Schema};
+    Arc::new(Schema::new(vec![
+        Field::new("filepath", DataType::Utf8, false),
```
ion-elgreco: Let's use large utf8 or utf8view to be on safer side

crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
ethan-tyler: lol right, missed this in my refactoring. Thanks for the call out, will fix it

ethan-tyler:
@ion-elgreco is codecov a blocker on this one? Happy to add more tests if needed, just lmk.

ion-elgreco:
> @ion-elgreco is codecov a blocker on this one? Happy to add more tests if needed, just lmk. 

Nope :)

PR #4172, opened 2026-02-06
refactor: add FileSelection to next provider for DeltaTableProvider removal

ethan-tyler:
# Description
DML operations scope scans to specific files via `DeltaTableProvider::with_files(Vec<Add>)`. The next provider (`DeltaScan`)
has no equivalent. This PR adds that foundation.

Introduces `FileSelection` a `pub(crate)` struct holding normalized file IDs
(`HashSet<String>`) that scopes scans to specific files. Unselected files are
excluded in the replay phase before DV tasks or data reads are produced.

No callers yet and will be added when DML operations migrate.

**Migration path**
1. This PR: introduce `FileSelection`, wire into `DeltaScan` pipeline
2. Convert DML ops from `DeltaTableProvider::with_files(Vec<Add>)` to
   `DeltaScan::with_file_selection(FileSelection::from_adds(...))`
3. Convert optimize's partition filtered scans to `FileSelection::from_paths`
4. Remove `DeltaTableProvider` and `with_files`

CC: @roeap 

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
@@ -23,6 +23,7 @@ use datafusion::{
 use delta_kernel::{
     Engine, ExpressionRef,
     engine::{arrow_conversion::TryIntoArrow, arrow_data::ArrowEngineData},
+    engine_data::FilteredEngineData,
     expressions::{Scalar, StructData},
     scan::{
         Scan as KernelScan, ScanMetadata,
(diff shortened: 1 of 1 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
```diff
@@ -187,6 +191,12 @@ where
                     Err(err) => return Poll::Ready(Some(Err(err.into()))),
                 };
 
+                if let Some(selection) = this.file_selection {
+                    ctx.files
+                        .retain(|file| selection.contains(file.file_url.as_str()));
+                    ctx.count = ctx.files.len();
```
ion-elgreco: I believe this should be done before visiting the scan files to prevent any work on kernel side for files that are being excluded anyways

ethan-tyler: Agreed and implemented the change

crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
```diff
@@ -489,14 +496,21 @@ fn parse_path(url: &Url, path: &str) -> DeltaResult<Url, DataFusionError> {
     })
 }
 
-fn visit_scan_file(ctx: &mut ScanContext, scan_file: ScanFile) {
+fn visit_scan_file(ctx: &mut ScanContext<'_>, scan_file: ScanFile) {
```
ion-elgreco: Correct me if I'm wrong but it looks like in this callback that ScanFile has already been visited in ScanData. I think we need to do it even earlier

ethan-tyler: You’re right. Filtering in the callback was still after `visit_scan_files` row visitation. I changed it to apply file selection to `scan_data.scan_files` first before calling `visit_scan_files`. Excluded files are skipped before ScanFile materialization/callback work.

Issue #4175, opened 2026-02-07
[Bug]: is_deltatable creates paths for non existing delta table location


khalidmammadov:
### What happened?

Using DeltaTable.is_deltatable(non_existing_path) creates paths which is not expected behaviour.

This is similar to old issue:
https://github.com/delta-io/delta-rs/pull/3504

### Expected behavior

It should not create and raise an error instead

### Operating System

None

### Binding

None

### Relevant logs

```shell

```

khalidmammadov:
take

PR #4176, opened 2026-02-07
fix: is_deltatable must not to create paths for not existing tables paths

khalidmammadov:
# Description

Calling `DeltaTable.is_deltatable(non_existing_path)` creates paths which is not expected behaviour.
This is a check function and path creation is not an expected behaviour.

After fix it will return False for invalid path and not create any new paths.

# Related Issue(s)
- closes [#106](https://github.com/delta-io/delta-rs/issues/4175)

# Documentation
Added notes into docs and checked locally by serving

### DIFF
--- a/python/tests/test_table_read.py
+++ b/python/tests/test_table_read.py
@@ -1062,14 +1062,18 @@ def test_is_deltatable_empty_path(tmp_path: Path):
     not_delta_path = tmp_path / "not_delta_table"
     # Ensure path exists
     not_delta_path.mkdir()
-
     assert not DeltaTable.is_deltatable(str(not_delta_path))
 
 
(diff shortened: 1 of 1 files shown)


ion-elgreco:
This function should not raise, if it doesn't exist it should return false.

khalidmammadov:
> This function should not raise, if it doesn't exist it should return false.

fixed, can you please take a look?

ion-elgreco:
You need to create a test that actually checks whether no path is created for local storage.

khalidmammadov:
> You need to create a test that actually checks whether no path is created for local storage.

Thanks, missed that one (the main one :) )! 

Added.

PR #4184, opened 2026-02-09
refactor: add insert_into to DeltaScan, replace file_id IN(...) with FileSelection

ethan-tyler:
# Description

Migration step toward `DeltaTableProvider` removal. Adds write path support to `DeltaScan` and replaces the plan time `file_id IN (...)` literal list with provider level file selection.

Adds `TableProvider::insert_into` to `DeltaScan` using `DeltaDataSink` + `DataSinkExec`. Runtime `LogStoreRef` is
carried on `DeltaScan` with `#[serde(skip)]`. Deserialized providers are read only unless re-wired at runtime.

Replaces matched file `IN (...)` construction in `scan_files_where_matches` with `FileSelection::from_paths(...)` forwarded into next scan planning/replay. Uses `MissingFilePolicy::Ignore` for matched file rewrites.

# Related Issue(s)
- part of #4113

# Documentation

PR #4187, opened 2026-02-10
feat: add DeltaScan insert_into with runtime log_store

ethan-tyler:
## Summary
- Add `TableProvider::insert_into` to next-provider `DeltaScan` (Append/Overwrite); reject Replace.
- Carry a runtime `LogStoreRef` on `DeltaScan` behind `#[serde(skip)]` so deserialized providers remain read-only unless re-wired at runtime.
- Wire `TableProviderBuilder` / `DeltaTable::table_provider()` to seed `DeltaScan` with the runtime log store for write support.

Migration step toward removing `DeltaTableProvider`: bring write-path parity to the next provider without expanding public API.

Split from #4184 per request

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider.rs
+++ b/crates/core/src/delta_datafusion/table_provider.rs
@@ -647,20 +647,28 @@ impl TableProviderBuilder {
     }
 
     pub async fn build(self) -> Result<next::DeltaScan> {
+        let TableProviderBuilder {
+            log_store,
+            snapshot,
+            file_column,
+            table_version,
+            file_skipping_predicates,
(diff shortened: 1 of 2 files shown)


ion-elgreco:
Thanks! We should do this though eventually: https://github.com/delta-io/delta-rs/issues/4189

ion-elgreco:
Thanks! We should do this though eventually: https://github.com/delta-io/delta-rs/issues/4189

ethan-tyler:
> Thanks! We should do this though eventually: #4189

@ion-elgreco  yesss sir

PR #4191, opened 2026-02-10
fix: preserve generated column metadata during schema merge

### LINKED ISSUE #4186: [Bug]: `write_deltalake` with `schema_mode="merge"` drops generated column metadata
### What happened?

Follow up from #4156. 

Appending data to a delta table when the input data omits generated columns and nullable non-generated columns now runs without an error. However, the generated column expressions are stripped from the table schema when appending data. This causes subsequent writes to fail to compute generated column values, resulting in NULL instead of the expected generated value. 

### Expected behavior

The generated column metadata is preserved during schema merge operations.

### Operating System

macOS

### Binding

Python

### Bindings Version

1.4.2

### Steps to reproduce

```python
from datetime import datetime, timezone
from uuid import uuid4
from pathlib import Path
import tempfile
import pyarrow as pa

# Schema with a generated column and a nullable column
schema = pa.schema([
    pa.field("uuid", pa.string(), nullable=False),
    pa.field("start", pa.timestamp("us", tz="UTC"), nullable=False),
    pa.field("year", pa.int32(), nullable=True,
             metadata={"delta.generationExpression": "EXTRACT(YEAR FROM start)"}),
    pa.field("user", pa.string(), nullable=True),
])

with tempfile.TemporaryDirectory() as tmp_dir:
    table_path = Path(tmp_dir) / "test_table"
    DeltaTable.create(str(table_path), schema=schema)

    # Table creation
    dt = DeltaTable(str(table_path))
    field = dt.schema().to_arrow().field("year")
    print(f"On table creation - year metadata: {field.metadata}") # {"delta.generationExpression": "EXTRACT(YEAR FROM start)"}

    # Write data
    data = pa.Table.from_pydict({
        "uuid": [str(uuid4())],
        "start": [datetime.now(timezone.utc)],
    })
    write_deltalake(str(table_path), data, mode="append", schema_mode="merge")

    # After first append with schema_mode=merge
    dt = DeltaTable(str(table_path))
    field = dt.schema().to_arrow().field("year")
    table = dt.to_pyarrow_table()
    print(f"After first append - year value: {table['year'][0].as_py()}") # 2026
    print(f"After first append - year metadata: {field.metadata}") # {}

    # On subsequent appends with schema_mode=merge
    write_deltalake(str(table_path), data, mode="append", schema_mode="merge")
    dt = DeltaTable(str(table_path))
    field = dt.schema().to_arrow().field("year")
    table = dt.to_pyarrow_table()
    print(f"On subsequent appends - year value: {table['year'][0].as_py()}") # None
    print(f"On subsequent appends - year metadata: {field.metadata}") # {}
```

### Relevant logs

```shell
On table creation - year metadata: {b'delta.generationExpression': b'EXTRACT(YEAR FROM start)'}
After first append - year value: 2026
After first append - year metadata: {}
On subsequent appends - year value: None
On subsequent appends - year metadata: {}
```

ethan-tyler:
# Description
Fixes a regression where `schema_mode="merge"` appends strip `delta.generationExpression` from the table schema. Once lost, subsequent writes compute NULL instead of generated values.

# Related Issue(s)
- closes #4186 

# Documentation

### DIFF
--- a/python/tests/test_generated_columns.py
+++ b/python/tests/test_generated_columns.py
@@ -295,32 +295,35 @@ def test_raise_when_gc_passed_merge_statement_during_schema_evolution(
         )
 
 
-def test_raise_when_gc_altered_merge_statement_during_schema_evolution(
-    table_with_gc: DeltaTable,
-):
+def test_schema_evolution_does_not_override_existing_gc_expression(tmp_path):
+    table_schema = DeltaSchema(
+        [
+            Field(name="id", type=PrimitiveType("integer"), nullable=True),
+            Field(
+                name="gc",
+                type=PrimitiveType("integer"),
+                nullable=True,
+                metadata={"delta.generationExpression": "5"},
+            ),
+            Field(name="user", type=PrimitiveType("string"), nullable=True),
+        ]
+    )
+    dt = DeltaTable.create(tmp_path, schema=table_schema)
+
     id_col = ArrowField("id", DataType.int32(), nullable=True)
     altered_gc_col = ArrowField("gc", DataType.int32(), nullable=True).with_metadata(
         {"delta.generationExpression": "id * 10"}
     )
     altered_gc_data = Table.from_pydict(
-        {"id": Array([1, 2], type=id_col), "gc": Array([10, 20], type=altered_gc_col)},
+        {"id": Array([1, 2], type=id_col), "gc": Array([5, 5], type=altered_gc_col)},
     )
 
-    with pytest.raises(
-        SchemaMismatchError,
-        match="Schema evolved fields cannot have generated expressions. Recreate the table to achieve this.",
-    ):
-        (
-            table_with_gc.merge(
-                altered_gc_data,
-                predicate="s.id = t.id",
-                source_alias="s",
-                target_alias="t",
-                merge_schema=True,
-            )
-            .when_not_matched_insert_all()
-            .execute()
-        )
+    # Force schema evolution by omitting nullable `user`, and ensure batch-side
+    # generationExpression metadata does not override table metadata for `gc`.
+    write_deltalake(dt, mode="append", data=altered_gc_data, schema_mode="merge")
+    dt = DeltaTable(tmp_path)
+    fields_by_name = {field.name: field for field in dt.schema().fields}
+    assert fields_by_name["gc"].metadata == {"delta.generationExpression": "5"}
 
 
 def test_merge_with_gc_invalid(table_with_gc: DeltaTable, invalid_gc_data):


ion-elgreco:
Did this work prior to the refactoring from 1.2.x and onwards?

I don't have my laptop in the coming weeks so reviewing is a bit difficult. What worries me is that  our schema evolution code was fine afaik before all the refactoring

ethan-tyler:
This specific path wasn't reachable before the recent refactors. Before #4162 these appends failed earlier on missing columns and the metadata loss never surfaced. #4090 introduced the DF plan schema derivation in the evolution path, and #4162 unblocked it. The schema evolution code was fine, this is a latent interaction between the older merge logic and the newer write/evolution paths.

ion-elgreco:
> This specific path wasn't reachable before the recent refactors. Before #4162 these appends failed earlier on missing columns and the metadata loss never surfaced. #4090 introduced the DF plan schema derivation in the evolution path, and #4162 unblocked it. The schema evolution code was fine, this is a latent interaction between the older merge logic and the newer write/evolution paths.

I mean before all the refractors. In 1.0.0 this never surfaced so I'm curious why

ethan-tyler:
Looks like this was already in v1.0.0, but the tests didn't cover the subsequent append edge case. One of those cases where it didn't surface until later on when that path became reachable.

ion-elgreco:
Alright, can you check if there is a test that covers not overriding generated expressions in schema evolution? 

When I added support for this, I intented It should only be possible to add generated expressions for new columna. Altering and removing should be done through a metadata operation only

ethan-tyler:
> Alright, can you check if there is a test that covers not overriding generated expressions in schema evolution?

Yeah, covered:

- `test_merge_arrow_schema_ignores_batch_generation_expression_conflicts` and `test_merge_arrow_schema_preserves_table_field_metadata_when_batch_missing` in `kernel/schema/cast/mod.rs`
- `test_schema_merge_append_missing_nullable_column_with_generated_columns` in integration tests
- `test_raise_when_gc_passed_during_schema_evolution` and `test_raise_when_gc_passed_merge_statement_during_schema_evolution` in Python tests

> When I added support for this, I intented It should only be possible to add generated expressions for new columna. Altering and removing should be done through a metadata operation only

No test for explicit remove/alter of an existing expression though. I'll add this.

ion-elgreco:
Thanks! @ethan-tyler

python/tests/test_generated_columns.py
```diff
@@ -293,6 +295,34 @@ def test_raise_when_gc_passed_merge_statement_during_schema_evolution(
...
+    altered_gc_col = ArrowField("gc", DataType.int32(), nullable=True).with_metadata(
+        {"delta.generationExpression": "id * 10"}
+    )
+    altered_gc_data = Table.from_pydict(
+        {"id": Array([1, 2], type=id_col), "gc": Array([10, 20], type=altered_gc_col)},
+    )
+
+    with pytest.raises(
+        SchemaMismatchError,
+        match="Schema evolved fields cannot have generated expressions. Recreate the table to achieve this.",
```
ion-elgreco: How do we even hit this error, this branch is only hit when the batch field name is not found in the table fields and a GC expression exists. But this field name already exists on the table fields :S

ethan-tyler: thanks Ion- I got myself all mixed up into testing the error message instead of the branch lol. Fixing it now.

ion-elgreco (approved):

ion-elgreco:
Thanks @ethan-tyler ! Nice work again

PR #4193, opened 2026-02-11
perf: parallel partition writers via per-stream JoinSet

fvaleye:
# Summary
Introducing parallelized partitioned writes in the DataFusion write path:
- For partitioned tables, we hash-repartition by partition columns and run one writer task per stream (`JoinSet`), then merge produced add actions.
- For unpartitioned tables, we keep the existing single-writer fan-in path unchanged.

Tried to keep behavior stable while improving partitioned write.

## Benchmark 
Local benchmark on tables to see the improvements
| Scenario | main | branch | speedup vs main |
|---|---:|---:|---:|
| partitioned 1M / 10 | 89.132 | 22.305 | +74.98% |
| partitioned 1M / 100 | 146.300 | 48.960 | +66.53% |
| partitioned 5M / 10 | 371.270 | 76.867 | +79.30% |
| unpartitioned 1M | 30.491 | 32.490 | -6.56% |

Partitioned writes are faster in these runs.
Unpartitioned path is functionally unchanged.

## Notes
- I tried not to introduce public API changes.
- On writer task failure, remaining tasks are aborted (`abort_all()`), but files could have already written (will be vacuumed later)

### DIFF
--- a/crates/aws/Cargo.toml
+++ b/crates/aws/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-aws"
-version = "0.13.1"
+version = "0.14.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -13,7 +13,7 @@ rust-version.workspace = true
 
 [dependencies]
 # path dependencies
-deltalake-core = { version = "0.30.0", path = "../core" , features = ["cloud"]}
+deltalake-core = { version = "0.31.0", path = "../core" , features = ["cloud"]}
 
 # workspace dependencies
 async-trait = { workspace = true }
@@ -30,19 +30,19 @@ uuid = { workspace = true, features = ["serde", "v4"] }
 url = { workspace = true }
 
 # crates.io dependencies
-aws-smithy-runtime-api = { version = "1.8" }
-aws-smithy-runtime = { version = "1.8", optional = true }
+aws-smithy-runtime-api = { version = "1.11" }
+aws-smithy-runtime = { version = "1.10", optional = true }
 aws-credential-types = { version = "1.2", features = ["hardcoded-credentials"] }
 aws-config = { version = "1.8", features = [
     "behavior-version-latest",
     "rt-tokio",
     "credentials-process",
 ] }
-aws-sdk-dynamodb = { version = "1.93.0", default-features = false, features = [
+aws-sdk-dynamodb = { version = "1.105.0", default-features = false, features = [
     "behavior-version-latest",
     "rt-tokio",
 ] }
-aws-sdk-sts = { version = "1.86.0", default-features = false, features = [
+aws-sdk-sts = { version = "1.98.0", default-features = false, features = [
     "behavior-version-latest",
     "rt-tokio",
 ] }
--- a/crates/azure/Cargo.toml
+++ b/crates/azure/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-azure"
-version = "0.13.0"
+version = "0.14.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -12,7 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-deltalake-core = { version = "0.30.0", path = "../core", features = ["cloud"] }
+deltalake-core = { version = "0.31.0", path = "../core", features = ["cloud"] }
 
 # workspace depenndecies
 bytes = { workspace = true }
@@ -22,7 +22,7 @@ tokio = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
-deltalake-core = { version = "0.30.0", path = "../core", features = [
+deltalake-core = { version = "0.31.0", path = "../core", features = [
     "datafusion",
 ] }
 chrono = { workspace = true }
--- a/crates/catalog-glue/Cargo.toml
+++ b/crates/catalog-glue/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-catalog-glue"
-version = "0.14.0"
+version = "0.15.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -25,7 +25,7 @@ aws-sdk-glue = { version = "1.120", default-features = false, features = [
     # Use the new default HTTPS client stack (no hyper 0.14).
     "default-https-client",
 ] }
-deltalake-core = { version = "0.30.0", path = "../core" }
+deltalake-core = { version = "0.31.0", path = "../core" }
 thiserror = { workspace = true }
 
 [dev-dependencies]
--- a/crates/catalog-unity/Cargo.toml
+++ b/crates/catalog-unity/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-catalog-unity"
-version = "0.14.1"
+version = "0.15.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -21,10 +21,10 @@ futures.workspace = true
 chrono.workspace = true
 tracing.workspace = true
 typed-builder = { workspace = true }
-deltalake-core = { version = "0.30.0", path = "../core" }
-deltalake-aws = { version = "0.13.0", path = "../aws", optional = true }
-deltalake-azure = { version = "0.13.0", path = "../azure", optional = true }
-deltalake-gcp = { version = "0.14.0", path = "../gcp", optional = true }
+deltalake-core = { version = "0.31.0", path = "../core" }
+deltalake-aws = { version = "0.14.0", path = "../aws", optional = true }
+deltalake-azure = { version = "0.14.0", path = "../azure", optional = true }
+deltalake-gcp = { version = "0.15.0", path = "../gcp", optional = true }
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "json", "http2"] }
 reqwest-retry = "0.7"
 reqwest-middleware = { version = "0.4.0", features = ["json"] }
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-core"
-version = "0.30.1"
+version = "0.31.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -15,7 +15,7 @@ rust-version.workspace = true
 features = ["datafusion", "json"]
 
 [dependencies]
-deltalake-derive = { version = "0.30.0", path = "../derive" }
+deltalake-derive = { version = "0.31.0", path = "../derive" }
 
 delta_kernel.workspace = true
 
--- a/crates/deltalake/Cargo.toml
+++ b/crates/deltalake/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake"
-version = "0.30.1"
+version = "0.31.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -25,14 +25,14 @@ features = [
 ]
 
 [dependencies]
-deltalake-core = { version = "0.30.0", path = "../core", default-features = false }
-deltalake-aws = { version = "0.13.0", path = "../aws", default-features = false, optional = true }
-deltalake-azure = { version = "0.13.0", path = "../azure", optional = true }
-deltalake-gcp = { version = "0.14.0", path = "../gcp", optional = true }
-deltalake-hdfs = { version = "0.14.0", path = "../hdfs", optional = true }
-deltalake-lakefs = { version = "0.13.0", path = "../lakefs", optional = true }
-deltalake-catalog-glue = { version = "0.14.0", path = "../catalog-glue", optional = true }
-deltalake-catalog-unity = { version = "0.14.0", path = "../catalog-unity", optional = true }
+deltalake-core = { version = "0.31.0", path = "../core", default-features = false }
+deltalake-aws = { version = "0.14.0", path = "../aws", default-features = false, optional = true }
+deltalake-azure = { version = "0.14.0", path = "../azure", optional = true }
+deltalake-gcp = { version = "0.15.0", path = "../gcp", optional = true }
+deltalake-hdfs = { version = "0.15.0", path = "../hdfs", optional = true }
+deltalake-lakefs = { version = "0.14.0", path = "../lakefs", optional = true }
+deltalake-catalog-glue = { version = "0.15.0", path = "../catalog-glue", optional = true }
+deltalake-catalog-unity = { version = "0.15.0", path = "../catalog-unity", optional = true }
 delta_kernel = { workspace = true }
 ctor = "0.6"
 
--- a/crates/derive/Cargo.toml
+++ b/crates/derive/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-derive"
-version = "0.30.0"
+version = "0.31.0"
 description = "Dervice macros for use in delta ecosystem crates"
 authors.workspace = true
 rust-version.workspace = true
--- a/crates/gcp/Cargo.toml
+++ b/crates/gcp/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-gcp"
-version = "0.14.0"
+version = "0.15.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -12,7 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-deltalake-core = { version = "0.30.0", path = "../core", features = ["cloud"] }
+deltalake-core = { version = "0.31.0", path = "../core", features = ["cloud"] }
 
 # workspace depenndecies
 async-trait = { workspace = true }
--- a/crates/hdfs/Cargo.toml
+++ b/crates/hdfs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-hdfs"
-version = "0.14.0"
+version = "0.15.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -12,7 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-deltalake-core = { version = "0.30.0", path = "../core"}
+deltalake-core = { version = "0.31.0", path = "../core"}
 hdfs-native-object-store = "0.15"
 
 # workspace dependencies
--- a/crates/lakefs/Cargo.toml
+++ b/crates/lakefs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-lakefs"
-version = "0.13.0"
+version = "0.14.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -12,7 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-deltalake-core = { version = "0.30.0", path = "../core", features = ["cloud"]}
+deltalake-core = { version = "0.31.0", path = "../core", features = ["cloud"]}
 # workspace dependencies
 async-trait = { workspace = true }
 bytes = { workspace = true }
--- a/crates/mount/Cargo.toml
+++ b/crates/mount/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-mount"
-version = "0.14.0"
+version = "0.15.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -12,7 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-deltalake-core = { version = "0.30.0", path = "../core" }
+deltalake-core = { version = "0.31.0", path = "../core" }
 errno = "0.3"
 
 # workspace depenndecies
--- a/crates/test/Cargo.toml
+++ b/crates/test/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
 name = "deltalake-test"
-version = "0.13.0"
+version = "0.14.0"
 edition = "2021"
 publish = false
 
 [dependencies]
-deltalake-core = { version = "0.30.0", path = "../core", features = [
+deltalake-core = { version = "0.31.0", path = "../core", features = [
     "integration_test",
 ] }
 


crates/core/src/operations/write/execution.rs
```diff
@@ -396,63 +424,126 @@ async fn write_data_plan(
...
-        let tx_clone = tx.clone();
-        let handle = tokio::task::spawn(async move {
-            while let Some(maybe_batch) = partition_stream.next().await {
+
+    let mut join_set = JoinSet::new();
+    for mut stream in partition_streams {
+        let store = object_store.clone();
+        let writer_config = config.clone();
+        join_set.spawn(async move {
+            let mut writer = DeltaWriter::new(store, writer_config);
```
ethan-tyler: This spawns one writer task per output partition with no upper bound. On large plans that's a lot of DeltaWriters, each with its own multipart concurrency. I worry this risks OOM/FD exhaustion/rate limiting. Worth adding a bounded concurrency mechanism (cap or semaphore), same for the CDC parallel branch.

crates/core/src/operations/write/mod.rs
```diff
@@ -1107,6 +1107,117 @@ mod tests {
...
+        let parallel_write_metrics: WriteMetrics = get_write_metrics(&parallel_table).await;
+        let single_writer_metrics: WriteMetrics = get_write_metrics(&single_writer_table).await;
+        assert_eq!(
+            parallel_write_metrics.num_added_files,
+            single_writer_metrics.num_added_files
+        );
+        assert_eq!(parallel_write_metrics.num_added_files, 2);
+    }
+
+    #[tokio::test]
```
ethan-tyler, line 1177: New CDC parallel path in write_cdc_plan doesn't have direct test coverage. I think this needs a multi-stream/partitioned/CDC test to guard against potential regressions.

crates/core/src/operations/write/execution.rs
fvaleye: The number of writer tasks is bounded by the physical plan output partition count (`plan.output_partitioning().partition_count()`), and `repartition_by_partition_columns` reuses that same count rather than introducing additional fan-out.
This is commonly influenced by DataFusion `target_partitions` (DataFusion [configs/session](https://datafusion.apache.org/user-guide/configs.html)), though the upstream plan, which can also determine final partitioning independently.

The previous implementation already spawned one task per partition stream (feeding a single mpsc writer), so stream-level parallelism existed before. This change moves write work into those per-stream tasks instead of funneling through one writer.

I tried first a semaphore-based approach, but it would trade off some partitioned-write throughput by reducing concurrent writers...
That said, I’m open to adding a bounded-concurrency cap (max_writers environment variable, semaphore...) if we want a stronger operational safety guardrail 😄

crates/core/src/operations/write/mod.rs
fvaleye: Good point, very true.
It was not previously covered, but I can cover `write_cdc_plan` as well!

crates/core/src/operations/write/execution.rs
ethan-tyler: Thanks, this clarification helps and I agree repartition isn’t adding extra fanout beyond output partition count. 

My remaining concern is the partitioned path now does writes directly in each stream task instead of the old single writer & bounded channel model. That changes the resource profile at high partition counts, especially with CDC overhead and object store upload concurrency.

I think it's still worth adding a bounded guardrail for both data and CDC paths, not blocking just my thoughts.

crates/core/src/operations/write/execution.rs
```diff
@@ -375,7 +380,30 @@ pub(crate) async fn write_exec_plan(
     }
 }
 
-// We drive partition streams concurrently and centralize writes via an mpsc channel.
+/// Hash repartitions the plan by partition columns so each stream
+/// writes to disjoint Delta partitions, avoiding duplicate small files
```
ion-elgreco: Small files problem was already resolved

crates/core/src/operations/write/execution.rs
```diff
@@ -41,6 +45,7 @@ fn channel_size() -> usize {
         std::env::var("DELTARS_WRITER_BATCH_CHANNEL_SIZE")
             .ok()
             .and_then(|s| s.parse::<usize>().ok())
+            .filter(|size| *size > 0)
```
ion-elgreco: I think channel size 0 is a legit value, it means nothing is buffered when that Is set

crates/core/src/operations/write/execution.rs
fvaleye: > Thanks, this clarification helps and I agree repartition isn’t adding extra fanout beyond output partition count.

You're welcome!

> My remaining concern is the partitioned path now does writes directly in each stream task instead of the old single writer & bounded channel model. That changes the resource profile at high partition counts, especially with CDC overhead and object store upload concurrency.
> 
> I think it's still worth adding a bounded guardrail for both data and CDC paths, not blocking just my thoughts.

And thanks for sharing! I didn't want to be too defensive, but your comment helps me rethink and see what's feasible to give more developers' control over this!

crates/core/src/operations/write/execution.rs
fvaleye: You're right!

crates/core/src/operations/write/execution.rs
fvaleye: Checking this again, I thought `0` was invalid. Thx, removing this!

crates/core/src/operations/write/mod.rs
fvaleye: Done!

fvaleye:
> This is great! But I have some concerns/theories, which I would like to see validated :)

Sure, and feedback is always appreciated!

> 1. Before we have one writer, but we would push multiple partition streams to a single channel. Partition streams were not defined by partition columns so the data could come in an unsorted order if you look at it from a Delta partition columns perspective. This means we open multiple partition_writers inside Delta writer. Even though we stream to disk, we still have multiple open so in theory this could also grow in memory if the amount of partitions grows. Could you check what sorting the datafusion plan by the partition columns while still keeping the old behavior of having one writer?

After checking it locally:
When one `DeltaWriter` with rows unsorted, multiple internal `PartitionWriters` can remain open, which could increase memory usage as partition cardinality grows.

I validated this with the one-writer behavior, comparing unsorted vs sorted-by-partition input.
Using the previous single-writer system, sorting input by partition columns improves performance, but only modestly (1M rows / 100 partitions)
- unsorted: `1.58s`
- sorted by partition column: `1.43s`
So sorting helps (~9.6%) with identical output cardinality, but it's not comparable to the performance gains of multiple writers.

> 2. Now we have a writer task per actual Delta partition stream, but repartitioning in datafusion should also come at a cost I assume. I can believe their repartitioning is less memory intensive than what we do in our writer but I still like to see this in a benchmark. So can you rerun the benchmarks but also with a memory profile graph, and peak rss?

Sure!

I took the time to run the benchmarks sequentially and locally.

**Throughput** (criterion, 10 samples):
| Scenario | Main | Branch | Speedup |
|---|---|---|---|
| 1M rows / 10 partitions | 159.82 ms | 65.95 ms | **2.42x** |
| 1M rows / 100 partitions | 227.25 ms | 113.74 ms | **2.00x** |
| 5M rows / 10 partitions | 491.09 ms | 142.23 ms | **3.45x** |
| 1M rows unpartitioned | 60.30 ms | 53.83 ms | ~1.12x |
| 5M rows unpartitioned | 204.39 ms | 200.47 ms | ~1.02x |

**Peak RSS** (`/usr/bin/time -l`):
| Scenario | Main | Branch | Delta |
|---|---|---|---|
| 1M rows / 10 partitions | 607 MB | 626 MB | +3.1% |
| 1M rows / 100 partitions | 656 MB | 666 MB | +1.4% |
| 5M rows / 10 partitions | 712 MB | 979 MB | +37.5% |
| 1M rows unpartitioned | 448 MB | 445 MB | -0.6% |
| 5M rows unpartitioned | 700 MB | 480 MB | -31.4% |

**Charts generated with Python/matplotlib**

rss_timeseries

peak_rss_and_throughput

> 3. Since datafusion drives the partition stream according to our delta lake partition columns, we might be able to increase performance by opening a PartitionWriter directly instead of a DeltaWriter. Can you also test this?

Good idea, it’s definitely better with a quick local running (same work, 2 adds), the numbers were:
- median `DeltaWriter`: ~654.9 ms
- median with `PartitionWriter`: ~564.5 ms
So, I implemented it!

The bump in memory might come from DataFusion's repartition buffers (shuffle) before writing.

crates/core/src/operations/write/execution.rs
```diff
@@ -396,63 +469,144 @@ async fn write_data_plan(
...
+            write_time_ms,
+        };
+
+        let actions = adds.into_iter().map(Action::Add).collect::<Vec<_>>();
+        return Ok((actions, metrics));
+    }
+
+    let file_schema = arrow_schema_without_partitions(&plan.schema(), &partition_columns);
+    let plan = repartition_by_partition_columns(plan, &partition_columns)?;
     let partition_streams = execute_stream_partitioned(plan, session.task_ctx())?;
```
ion-elgreco, line 554: Can we not already know which partition sits in which partition stream? Or just peak the first row of each batch, instead of calling divide_by_partition_values

ion-elgreco (dismissed):
Thanks for the benchmarks! Just curious in your examples the partitions are evenly distributed? Can you have a look at larger evenly distributed and skewed distributions? 

Curious what it does there because maybe for skewed source data it might make sense to have the old behavior

ion-elgreco:
Looking at our DeltaWriter, I think we can actually just call DeltaWriter::write_partition, and you manually give the correct values here inside the writer task. I believe this will yield same performance while avoiding the divide_by_partition_values!

fvaleye:
> Thanks for the benchmarks! Just curious in your examples the partitions are evenly distributed? Can you have a look at larger evenly distributed and skewed distributions?
> 
> Curious what it does there because maybe for skewed source data it might make sense to have the old behavior

I ran some additional benchmarks covering both when data is evenly distributed and when it's `90%` skewed distribution. 

Results:
| Scenario | Main (time) | Branch (time) | Speedup | Main (RSS) | Branch (RSS) | RSS Delta |
|---|---|---|---|---|---|---|
| 10M rows / 100 partitions (even) | 1247.6 ms | 151.9 ms | **8.21x** | 2641 MB | 2327 MB | **-11.9%** |
| 10M rows / 1000 partitions (even) | 3238.4 ms | 620.1 ms | **5.22x** | 3451 MB | 2676 MB | **-22.5%** |
| 1M rows / 10 partitions (90% skew) | 52.8 ms | 38.5 ms | **1.37x** | 1398 MB | 1527 MB | +9.3% |
| 1M rows / 100 partitions (90% skew) | 92.2 ms | 52.1 ms | **1.77x** | 1519 MB | 2118 MB | +39.5% |
| 5M rows / 10 partitions (90% skew) | 252.4 ms | 171.3 ms | **1.47x** | 1890 MB | 2245 MB | +18.8% |
| 10M rows / 100 partitions (90% skew) | 549.5 ms | 343.8 ms | **1.60x** | 1890 MB | 2467 MB | +30.6% |
| 10M rows / 1000 partitions (90% skew) | 1103.2 ms | 492.8 ms | **2.24x** | 2095 MB | 2997 MB | +43.1% |

DataFusion's repartitioning is very efficient when the data is evenly spread. And when it's not, the memory is increased and might come from DataFusion's repartition buffers.
Even in the worst-case skew scenario, it's 1.37x faster.

fvaleye:
> Looking at our DeltaWriter, I think we can actually just call DeltaWriter::write_partition, and you manually give the correct values here inside the writer task. I believe this will yield same performance while avoiding the divide_by_partition_values!

After `RepartitionExec` by partition columns, hash collisions can cause a stream to contain multiple partition values, so we can't reliably peek at the first row or skip partitioning. `divide_by_partition_values` is still needed to split batches that contain multiple partitions (hash collisions), but the sort cost per batch is tiny compared to the actual writes.                                 
We can have a check that the batch is single-partition, then call `write_partition` directly, falling back to `divide_by_partition_values` when it's not. But I don't know if it's worth it.

crates/core/src/operations/write/execution.rs
fvaleye: There is a risk of hash colission, I replied [here](https://github.com/delta-io/delta-rs/pull/4193#issuecomment-3902124586)

ion-elgreco:
> > Looking at our DeltaWriter, I think we can actually just call DeltaWriter::write_partition, and you manually give the correct values here inside the writer task. I believe this will yield same performance while avoiding the divide_by_partition_values!
> 
> After `RepartitionExec` by partition columns, hash collisions can cause a stream to contain multiple partition values, so we can't reliably peek at the first row or skip partitioning. `divide_by_partition_values` is still needed to split batches that contain multiple partitions (hash collisions), but the sort cost per batch is tiny compared to the actual writes.                                 
> We can have a check that the batch is single-partition, then call `write_partition` directly, falling back to `divide_by_partition_values` when it's not. But I don't know if it's worth it.

Hmm but what are the chances of hash collision?

Hey! @alamb , in terms of repartitioning in datafusion how likely is it that a partition stream contains parts from another stream?

ion-elgreco:
> > Looking at our DeltaWriter, I think we can actually just call DeltaWriter::write_partition, and you manually give the correct values here inside the writer task. I believe this will yield same performance while avoiding the divide_by_partition_values!
> 
> After `RepartitionExec` by partition columns, hash collisions can cause a stream to contain multiple partition values, so we can't reliably peek at the first row or skip partitioning. `divide_by_partition_values` is still needed to split batches that contain multiple partitions (hash collisions), but the sort cost per batch is tiny compared to the actual writes. We can have a check that the batch is single-partition, then call `write_partition` directly, falling back to `divide_by_partition_values` when it's not. But I don't know if it's worth it.

The docs and code imply that this doesn't happen, and that each partition stream has exclusively data from that partition. I posted the question as well on discord for extra verification, so lets see. If that's true we can use write_partition directly

fvaleye:
> > > Looking at our DeltaWriter, I think we can actually just call DeltaWriter::write_partition, and you manually give the correct values here inside the writer task. I believe this will yield same performance while avoiding the divide_by_partition_values!
> > 
> > 
> > After `RepartitionExec` by partition columns, hash collisions can cause a stream to contain multiple partition values, so we can't reliably peek at the first row or skip partitioning. `divide_by_partition_values` is still needed to split batches that contain multiple partitions (hash collisions), but the sort cost per batch is tiny compared to the actual writes. We can have a check that the batch is single-partition, then call `write_partition` directly, falling back to `divide_by_partition_values` when it's not. But I don't know if it's worth it.
> 
> The docs and code imply that this doesn't happen, and that each partition stream has exclusively data from that partition. I posted the question as well on discord for extra verification, so lets see. If that's true we can use write_partition directly

I think there is no strong guarantee using `HashRepartition` of one partition value per stream.
The concern I have is the following: `RepartitionExec` assigns rows via `partition_index = hash(partition_columns) % num_partitions`

And, `nums_partitions` comes from DataFusion's `target_partitions` configuration, which is by default the number of available CPU cores.

If we have 8 cores on a machine, then we have `num_partitions` equal to 8, and if we have 100 distinct partition values in a Delta Table... we only have 8 buckets for 100 values.

```
Example:
- hash(year=2000) % 8 = 4
- hash(year=2001) % 8 = 4 -> same stream, different partition
```

In any case, it is worth checking if we require `divide_by_partition_values`, let me know 👍

ethan-tyler:
@ion-elgreco - I did some research into this and here's my understanding. Hash repartitioning gives bucket exclusivity, not partition value exclusivity. All rows for a given partition key land in the same stream, but a stream can contain multiple partition values when distinct keys > N (normal bucket collision from % N).

I don't think we can peek the first row and assume the whole stream is one partition and still need divide_by_partition_values. The value here is that no two streams can write the same partition key concurrently.

Let see what responses the DF discord gets too.

crates/core/src/operations/write/execution.rs
```diff
@@ -375,7 +386,69 @@ pub(crate) async fn write_exec_plan(
...
+                )?;
+                e.insert(PartitionWriter::try_with_config(
+                    store.clone(),
+                    config,
+                    writer_stats_config.num_indexed_cols,
+                    writer_stats_config.stats_columns.clone(),
+                )?)
+            }
+        };
+        writer.write(&part.record_batch).await?;
```
rtyler: As best as I can tell this is _also_ a potential serialization point. Is there a reason why the JoinSet approach wasn't applied here as well? 🤔

crates/core/src/operations/write/execution.rs
```diff
@@ -396,63 +469,144 @@ async fn write_data_plan(
         writer_stats_config.stats_columns.clone(),
     );
 
-    // sync channel for batches produced by partition stream
-    let (tx, mut rx) = mpsc::channel::<RecordBatch>(channel_size());
+    // Keep the previous single-writer fan-in path for unpartitioned tables.
+    if partition_columns.is_empty() {
```
rtyler, line 483: These functions are quite lengthy, I don't see a compelling reason to keep this divergent branch around even if the section below is only going to end up with a JoinSet containing one task?

Was there severely different performance that would justify the pretty big deviation between these two code paths?

crates/core/src/operations/write/execution.rs
```diff
@@ -504,53 +658,210 @@ async fn write_cdc_plan(
...
-
-    let normal_writer_handle = tokio::task::spawn(async move {
-        let mut writer = DeltaWriter::new(object_store, normal_config);
-        let mut total_write_ms: u64 = 0;
-        while let Some(batch) = rx_normal.recv().await {
-            let wstart = std::time::Instant::now();
-            writer.write(&batch).await?;
-            total_write_ms += wstart.elapsed().as_millis() as u64;
+    // Keep the previous single-writer fan-in path for unpartitioned tables.
+    if partition_columns.is_empty() {
```
rtyler, line 654: Same concern about fairly redundant but different branches between writers with partitions and those without

crates/core/src/operations/write/execution.rs
```diff
@@ -504,53 +658,210 @@ async fn write_cdc_plan(
...
+                        normal_batch.remove_column(idx.ok_or(DeltaTableError::generic(
+                            "idx of _change_type col not found. This shouldn't have happened.",
+                        ))?);
+
+                        txn.send(normal_batch).await.map_err(|_| {
+                            DeltaTableError::Generic(
+                                "normal writer closed unexpectedly".to_string(),
+                            )
+                        })?;
+                    }
```
rtyler, line 739: I'm not a datafusion wizard.  🪄  Is there a strong reason why this couldn't/shouldn't be operations on the `normal_df` that get handled by Datafusion in a way that adds some potential performance or parallelism gains?

I don't have a great mental model of how the `normal_df` plan would be executed by `execute_stream` but I would assume it does _something_ (🤞) in parallel, which would mean that there's another potential serialization point here where the stream gets collapsed to a single task only to fill up `txn` one by one 🤔

rtyler (commented):
Looks cool @fvaleye ! 

I have some questions inline, which if they are just a gap in my own understanding, no need to explain in the PR, but code comments to remind us later would be apprecaited

crates/core/src/operations/write/execution.rs
fvaleye: The parallelism is set at the stream level (one task per stream via `JoinSet`), while within the stream, writes are ordered (sequential) to keep each partition writer's state consistent. If we wanted parallelism here, it would require a synchronization (like a lock) that would hurt the performance. And the performance gain would be minimal because after hash repartitioning, each batch typically contains just a small number of partition fragments, so the loop is very short. I can add a comment if you want!

crates/core/src/operations/write/execution.rs
fvaleye: You're right, I wanted to split the paths intentionally between partition / no partition. That said, we can refactor duplicated code into helpers!

crates/core/src/operations/write/execution.rs
fvaleye: Agreed!

crates/core/src/operations/write/execution.rs
fvaleye: I wrapped up this code (with JoinSet) to ensure that we follow the same path as before (it already existed, u).
We have two options if we want to improve this:
1. Don't rely on Datafusion and use Arrow compute, replace the per-batch Datafusion with direct `arrow::compute::filter_record_batch` using a boolean mask on the _change_type column.                                                                                                                                                              
3. Rely more on DataFusion plan restructuring: instead of the upstream `UNION ALL` that merges normal + CDC rows with a `_change_type` tag, keep them as two separate execution plans from the start to avoid downstream splitting.

ion-elgreco:
> > > > Looking at our DeltaWriter, I think we can actually just call DeltaWriter::write_partition, and you manually give the correct values here inside the writer task. I believe this will yield same performance while avoiding the divide_by_partition_values!
> > > 
> > > 
> > > After `RepartitionExec` by partition columns, hash collisions can cause a stream to contain multiple partition values, so we can't reliably peek at the first row or skip partitioning. `divide_by_partition_values` is still needed to split batches that contain multiple partitions (hash collisions), but the sort cost per batch is tiny compared to the actual writes. We can have a check that the batch is single-partition, then call `write_partition` directly, falling back to `divide_by_partition_values` when it's not. But I don't know if it's worth it.
> > 
> > 
> > The docs and code imply that this doesn't happen, and that each partition stream has exclusively data from that partition. I posted the question as well on discord for extra verification, so lets see. If that's true we can use write_partition directly
> 
> I think there is no strong guarantee using `HashRepartition` of one partition value per stream. The concern I have is the following: `RepartitionExec` assigns rows via `partition_index = hash(partition_columns) % num_partitions`
> 
> And, `nums_partitions` comes from DataFusion's `target_partitions` configuration, which is by default the number of available CPU cores.
> 
> If we have 8 cores on a machine, then we have `num_partitions` equal to 8, and if we have 100 distinct partition values in a Delta Table... we only have 8 buckets for 100 values.
> 
> ```
> Example:
> - hash(year=2000) % 8 = 4
> - hash(year=2001) % 8 = 4 -> same stream, different partition
> ```
> 
> In any case, it is worth checking if we require `divide_by_partition_values`, let me know 👍

If thats the case, and num_partitions is always set by the default available cpu cores. Then you should see roughly the same performance without hash repartioning, right? Do you mind running the same benchmarks without the repartitioning step?

crates/core/src/operations/write/execution.rs
fvaleye: Added `DELTARS_MAX_CONCURRENT_WRITERS` as safeguard

rtyler (approved):
yolo

ethan-tyler:
> yolo

To the moon!!

Issue #4189, opened 2026-02-10
[Feature]: refactor DeltaDataSink to go through delta writer builder


ion-elgreco:
### Describe the solution you'd like

DeltaDataSink should pipe through WriteBuilder

### Priority

None

### Contribution

- [ ] I'm willing to submit a pull request for this feature
- [ ] I can help with testing this feature
- [ ] I can help with documentation for this feature

ethan-tyler:
Take

ethan-tyler:
@ion-elgreco  @roeap - Should we hold off on this until after the write migration #4119 lands to avoid conflicts or land this now and rebase on top?

ion-elgreco:
We can land this now, there shouldn't be any conflicts imho

ethan-tyler:
Landing now works if this stays local to DeltaDataSink and calls existing write internals. Anything else should belong in #4119 IMO.

ion-elgreco:
> Landing now works if this stays local to DeltaDataSink and calls existing write internals. Anything else should belong in #4119 IMO.

I agree, that's why going through the WriteBuilder should not cause any conflicts.

PR #4194, opened 2026-02-11
feat: route DeltaDataSink through shared write_streams

ethan-tyler:
# Description
Replaces the local `DeltaWriter` loop in `DeltaDataSink::write_all` with
`operations::write::execution::write_streams(...)`. Same change applied to
`write_data_plan`.

# Related Issue(s)
- closes: #4189 

# Documentation

### DIFF
--- a/crates/core/src/operations/write/execution.rs
+++ b/crates/core/src/operations/write/execution.rs
@@ -78,8 +78,8 @@ mod tests {
     use object_store::memory::InMemory;
 
     use super::{
-        DEFAULT_WRITER_BATCH_CHANNEL_SIZE, ObjectStoreRef, SendableRecordBatchStream, WriterConfig,
-        parse_channel_size, write_streams,
+        DEFAULT_WRITER_BATCH_CHANNEL_SIZE, ObjectStoreRef, SendableRecordBatchStream,
+        WRITER_TASK_CLOSED_UNEXPECTEDLY_MSG, WriterConfig, parse_channel_size, write_streams,
     };
 
     #[test]
@@ -189,9 +189,16 @@ mod tests {
         let result =
             write_streams(vec![pending_stream, failing_stream], object_store, config).await;
 
+        let err = result
+            .expect_err("expected writer failure when stream schema mismatches writer config");
+        let err_msg = err.to_string();
         assert!(
-            result.is_err(),
-            "expected writer failure when stream schema mismatches writer config"
+            err_msg.contains("Unexpected Arrow schema"),
+            "expected writer schema mismatch error, got: {err_msg}"
+        );
+        assert!(
+            !err_msg.contains(WRITER_TASK_CLOSED_UNEXPECTEDLY_MSG),
+            "expected primary writer failure, got channel-close fallback: {err_msg}"
         );
         assert!(
             dropped.load(Ordering::SeqCst),


ethan-tyler:
Note: this overlaps with #4193 in operations/write/execution.rs. I had these changes done before. Will rebase after #4193 lands.

ion-elgreco:
@ethan-tyler I meant we should do this instead: https://docs.rs/deltalake/latest/deltalake/operations/write/struct.WriteBuilder.html

This would reduce the surface of sink_all to just a couple lines

ethan-tyler:
> @ethan-tyler I meant we should do this instead: https://docs.rs/deltalake/latest/deltalake/operations/write/struct.WriteBuilder.html
> 
> This would reduce the surface of sink_all to just a couple lines

Agree on single source of truth. However `WriteBuilder` doesn't accept a `SendableRecordBatchStream`. `with_input_batches(...)` would require collecting the full stream into memory, losing backpressure.

I choose to go with `write_all` streaming by delegating to `write_streams` (same core as the write execution path). 

If we want `write_all` to call `WriteBuilder`, we likely need a stream native `WriteBuilder` input API first.

ion-elgreco:
> > @ethan-tyler I meant we should do this instead: https://docs.rs/deltalake/latest/deltalake/operations/write/struct.WriteBuilder.html
> > 
> > This would reduce the surface of sink_all to just a couple lines
> 
> Agree on single source of truth. However `WriteBuilder` doesn't accept a `SendableRecordBatchStream`. `with_input_batches(...)` would require collecting the full stream into memory, losing backpressure.
> 
> I choose to go with `write_all` streaming by delegating to `write_streams` (same core as the write execution path). 
> 
> If we want `write_all` to call `WriteBuilder`, we likely need a stream native `WriteBuilder` input API first.

You should be able to route it through LazyMemoryExec again, see here: https://github.com/delta-io/delta-rs/blob/main/python%2Fsrc%2Fdatafusion.rs#L32-L122

ion-elgreco:
> > > @ethan-tyler I meant we should do this instead: https://docs.rs/deltalake/latest/deltalake/operations/write/struct.WriteBuilder.html
> > > 
> > > This would reduce the surface of sink_all to just a couple lines
> > 
> > Agree on single source of truth. However `WriteBuilder` doesn't accept a `SendableRecordBatchStream`. `with_input_batches(...)` would require collecting the full stream into memory, losing backpressure.
> > 
> > I choose to go with `write_all` streaming by delegating to `write_streams` (same core as the write execution path). 
> > 
> > If we want `write_all` to call `WriteBuilder`, we likely need a stream native `WriteBuilder` input API first.
> 
> You should be able to route it through LazyMemoryExec again, see here: https://github.com/delta-io/delta-rs/blob/main/python%2Fsrc%2Fdatafusion.rs#L32-L122

We can also move this wrapping of Sendablerecordbatchstream inside writer builder::with_recordbatch_stream() which converts it to lazytableprovider and then logicalplan :)

ethan-tyler:
> We can also move this wrapping of Sendablerecordbatchstream inside writer builder::with_recordbatch_stream() which converts it to lazytableprovider and then logicalplan :)

Looked into this --`LazyBatchGenerator::generate_next_batch()` is sync, but `DataSink::write_all` gets an async `SendableRecordBatchStream`. Bridging needs either blocking the runtime or prebuffering and both defeat streaming. The Python path avoids this because it wraps a synchronous `RecordBatchReader`.

Even with a working bridge, data round trips. `WriteBuilder` also reruns a bunch of lines of setup that don't apply in the sink context.

For now I think sharing at the I/O layer is the right call given`write_streams()` is what `WriteBuilder` delegates to internally anyway. Both paths get the same writer, channel, backpressure, and error handling.

A `WriteBuilder::with_streams()` / `with_physical_plan()` that skips the logical planning cycle would be the cleaner single source of truth path and always happy to build that as a follow up. :)

ion-elgreco:
> > We can also move this wrapping of Sendablerecordbatchstream inside writer builder::with_recordbatch_stream() which converts it to lazytableprovider and then logicalplan :)
> 
> Looked into this --`LazyBatchGenerator::generate_next_batch()` is sync, but `DataSink::write_all` gets an async `SendableRecordBatchStream`. Bridging needs either blocking the runtime or prebuffering and both defeat streaming. The Python path avoids this because it wraps a synchronous `RecordBatchReader`.
>

An upstream refactor seems to be acceptable, see: https://github.com/apache/datafusion/issues/13614

That would simplify it alot

> Even with a working bridge, data round trips. `WriteBuilder` also reruns a bunch of lines of setup that don't apply in the sink context.
> 

That's fair, perhaps we should have some code moved into functions so we can reuse, which will actually make it collide with the work of @roeap 

> For now I think sharing at the I/O layer is the right call given`write_streams()` is what `WriteBuilder` delegates to internally anyway. Both paths get the same writer, channel, backpressure, and error handling.
> 
Alright!

> A `WriteBuilder::with_streams()` / `with_physical_plan()` that skips the logical planning cycle would be the cleaner single source of truth path and always happy to build that as a follow up. :) 

Let's avoid that :p

ethan-tyler:
@rtyler - I will need rebase after #4193 lands

crates/core/src/delta_datafusion/table_provider/data_sink.rs
```diff
@@ -111,10 +112,23 @@ impl DataSink for DeltaDataSink {
         let target_schema = self.snapshot.input_schema();
         let table_props = self.snapshot.table_configuration().table_properties();
 
-        let mut stream = self.create_converted_stream(data, target_schema.clone());
+        let operation_id = Uuid::new_v4();
+        let stream = self.create_converted_stream(data, target_schema.clone());
         let partition_columns = self.snapshot.metadata().partition_columns();
-        let object_store = self.log_store.object_store(None);
+        let object_store = self.log_store.object_store(Some(operation_id));
```
ion-elgreco, line 118: Operation id is not really going to do much here tbh

ethan-tyler: You're right, we should actually revert - I think there's a potential panic for LakeFS because it expects a transaction store created by pre_execute

ethan-tyler: @ion-elgreco - let me know if this is a valid concern and I can do a quick follow up

ion-elgreco: No it's fine, lakefs can't be supported in this write path

PR #4195, opened 2026-02-12
refactor: use BatchAdapterFactory for scan adaptation

ethan-tyler:
# Description
Follow up to the DF 52 upgrade. Replaces custom scan batch casting logic
with DataFusion's `BatchAdapterFactory` via `datafusion-physical-expr-adapter`.

Adds hardening tests for schema evolution and DV scan behavior

Keeps DF default behavior with no custom compatibility mode

# Related Issue(s)

# Documentation


ethan-tyler:
FYI I have a follow up hardening PR  on top of this

PR #4197, opened 2026-02-13
fix(datafusion): avoid overflow when scanning add actions

vsmanish1772:
# Description
In 1.3.2 when we perform delete with partition only predicate, when we try to find_files() -> it calls get_add_action() which concatenates all file metadata into one giant RecordBatch. Arrow's StringArray uses 32-bit offsets, so a single array can hold at most ~2GB of string data. When the table is huge which cannot be held into single array.

if approved this will be back ported to 1.3.3

# Related Issue(s)

# Documentation


ion-elgreco:
@vsmanish1772 good! I would like to see that we do a breaking change in Python as well, but we should not backport that part

This should return an Arro3 table that is constructed with the Vec of recordbatches :), see below where it should happen.

https://github.com/delta-io/delta-rs/blob/main/python%2Fsrc%2Flib.rs#L1634

vsmanish1772:
> @vsmanish1772 good! I would like to see that we do a breaking change in Python as well, but we should not backport that part
> 
> This should return an Arro3 table that is constructed with the Vec of recordbatches :), see below where it should happen.
> 
> https://github.com/delta-io/delta-rs/blob/main/python%2Fsrc%2Flib.rs#L1634

okay makes sense. in that case a second pr then. If i understand right we will back port this pr to 1.3.3 and then a follow up pr in main which will change python API to use new `get_add_action_batches()` which will return ArrowTable. is my understanding correct?

crates/core/src/table/state.rs
```diff
@@ -231,6 +231,55 @@ impl EagerSnapshot {
         &self,
         flatten: bool,
     ) -> Result<arrow::record_batch::RecordBatch, DeltaTableError> {
+        let batches = self.add_actions_batches(flatten)?;
+        if batches.is_empty() {
+            // Return an empty batch with the correct schema
+            return self.add_actions_batches_empty(flatten);
```
ethan-tyler: `add_actions_expr_and_schema` called twice on the empty path. I would compute once and thread through.

crates/core/src/table/state.rs
```diff
@@ -231,6 +231,55 @@ impl EagerSnapshot {
...
+        }
+        Ok(concat_batches(batches[0].schema_ref(), &batches)?)
+    }
+
+    /// Get add action data as a list of [arrow::record_batch::RecordBatch] without
+    /// concatenating them into a single batch. This avoids the 2GB Arrow offset
+    /// limit for tables with a very large number of files.
+    ///
+    /// See [Self::add_actions_table] for schema details.
+    pub fn add_actions_batches(
```
ethan-tyler: I'd change this to `pub(crate)`, the only caller is `find_files.rs`. We can widen later if needed, but no reason to commit to a pub api before you have to.

crates/core/src/table/state.rs
```diff
@@ -343,4 +380,42 @@ mod tests {
...
+
+        let snapshot = table.snapshot()?.snapshot();
+        assert!(snapshot.add_actions_batches(false)?.is_empty());
+        assert!(snapshot.add_actions_batches(true)?.is_empty());
+        Ok(())
+    }
+
+    #[cfg(feature = "datafusion")]
+    #[tokio::test]
+    async fn test_add_actions_batches_non_empty_flatten_has_partition_columns() -> DeltaResult<()> {
```
ethan-tyler, line 448: Worth adding an equivalence assertion:

    let concatenated = concat_batches(
        flattened_batches[0].schema_ref(),
        &flattened_batches,
    )?;
    let single = snapshot.add_actions_table(true)?;
    assert_eq!(concatenated, single);

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -323,33 +323,45 @@ async fn scan_memory_table(snapshot: &EagerSnapshot, predicate: &Expr) -> DeltaR
...
+    let mut mem_schema: Option<Arc<Schema>> = None;
 
-    for field in schema.fields() {
-        if let Some(name) = field.name().strip_prefix("partition.") {
-            arrays.push(batch.column_by_name(field.name()).unwrap().to_owned());
-            fields.push(field.as_ref().clone().with_name(name));
+    for batch in &all_batches {
+        let schema = batch.schema();
+        let mut arrays = Vec::new();
+        let mut fields = Vec::new();
```
ethan-tyler: Is there a reason you removed `with_capacity` here where `schema.fields().len()` is known?

crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -323,33 +323,45 @@ async fn scan_memory_table(snapshot: &EagerSnapshot, predicate: &Expr) -> DeltaR
...
+        fields.push(Field::new(PATH_COLUMN, DataType::Utf8, false));
+
+        for field in schema.fields() {
+            if let Some(name) = field.name().strip_prefix("partition.") {
+                arrays.push(batch.column_by_name(field.name()).unwrap().to_owned());
+                fields.push(field.as_ref().clone().with_name(name));
+            }
         }
+
+        let batch_schema = mem_schema.get_or_insert_with(|| Arc::new(Schema::new(fields.clone())));
```
ethan-tyler: nit: `fields` recomputed every iteration but is only used on the first. I would pull schema construction out of the loop.

ion-elgreco:
> > @vsmanish1772 good! I would like to see that we do a breaking change in Python as well, but we should not backport that part
> > 
> > This should return an Arro3 table that is constructed with the Vec of recordbatches :), see below where it should happen.
> > 
> > https://github.com/delta-io/delta-rs/blob/main/python%2Fsrc%2Flib.rs#L1634
> 
> okay makes sense. in that case a second pr then. If i understand right we will back port this pr to 1.3.3 and then a follow up pr in main which will change python API to use new `get_add_action_batches()` which will return ArrowTable. is my understanding correct?

Yes can be done as followup!

crates/core/src/table/state.rs
vsmanish1772: agreed!

crates/core/src/delta_datafusion/find_files.rs
vsmanish1772: i think that was oversight from my end assuming i do not know how much space to allocate but i was wrong

ethan-tyler: You're good, just wanted to ask if it was intentional.

vsmanish1772:
@ethan-tyler @ion-elgreco i'm ready for another round of reviews. Thanks for the suggestions

crates/core/src/table/state.rs
ethan-tyler: Sorry I didn't think of this earlier and since we are in the weeds, could you add the same check for flatten = false? If someone changes `add_actions_batches_with_schema` and accidentally alters non flattenedschema/ordering/nullability, CI may not catch it.

crates/core/src/table/state.rs
```diff
@@ -231,6 +231,65 @@ impl EagerSnapshot {
...
+
+        let input_schema = self.snapshot().inner.scan_row_parsed_schema_arrow()?;
+        let input_schema = Arc::new(input_schema.as_ref().try_into_kernel()?);
+        let evaluator = ARROW_HANDLER.new_expression_evaluator(
+            input_schema,
+            expression.into(),
+            table_schema,
+        )?;
+
+        let results = files
```
ethan-tyler, line 277: For uncheckpointed tables with lots of small commits this could theoretically produce thousands of single row batches. If it's not too much trouble I think it's worth a rechunk step or at least a note that it's a known tradeoff.

ion-elgreco: I agree but afaik arrow-rs doesn't have any nice way to do this inside their api, right?

At least in polars-arrow rechunking is something handled by their chunkedArrays

ion-elgreco: I agree but afaik arrow-rs doesn't have any nice way to do this inside their api, right?

At least in polars-arrow rechunking is something handled by their chunkedArrays

ethan-tyler: @ion-elgreco - You're right, however  I did find that arrow-rs has a native batch coalescing api `arrow_select::coalesce::BatchCoalescer` exposed via `arrow::compute::BatchCoalescer`

https://docs.rs/arrow-select/latest/arrow_select/coalesce/struct.BatchCoalescer.html

For this backport I'm okay keeping current scope, but we can add a follow up that coalesces tiny metadata batches before MemTable construction and benchmark it.

ion-elgreco: Ah sweet! That we can definitely use!

vsmanish1772: Ah i agree, i didnt know about this BatchCoalescer. Small thousands of micro batches are always problem (overhead) :D haha

vsmanish1772: okay i think i understand now how it works and i can wire `BatchCoalescer` inside inside add_actions_batches  right before returning. That way both callers (scan_memory_table and any future caller) get properly sized batches automatically. Hmm but i wonder if eagerSnapshot would still be used in future

vsmanish1772:
> Nice work! Left a few comments and suggestions

i have attempted the suggestions you have made. let me know what you think

crates/core/src/table/state.rs
vsmanish1772: i realised little later that we would like to do a breaking change in python where get_add_actions() will start using add_action_batches() which will return arrow table as per @ion-elgreco suggestion. I will probably make it public api in the pr that does breaking change

ethan-tyler: yes that make sense and what I had in mind. Thanks!

PR #4199, opened 2026-02-13
chore: harden scan adapter caching and DV mask edge cases

ethan-tyler:
# Description
Follow up to the DF scan adapter integration.

Adds a cached schema adapter path in scan execution to avoid rebuilding adapters for repeated source schemas. Updates scan finalization to use the cached path.

Fixes DV mask handling edge cases with perserving remainder across multi batch file reads, pads short masks with implicit true, errors when mask length exceeds batch row count.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -410,10 +410,6 @@ pub(crate) struct MatchedFilesScan {
     /// we do not (yet) have full coverage to translate datafusion to
     /// kernel predicates.
     pub(crate) delta_predicate: Arc<Predicate>,
-    /// The predicate contains only partition column references
-    ///
-    /// This implies that for each matched file all data matches.
-    pub(crate) partition_only: bool,
 }
 
 impl MatchedFilesScan {
@@ -467,9 +463,8 @@ pub(crate) async fn scan_files_where_matches(
         result: Ok(()),
     };
     for term in &skipping_pred {
-        visitor.result = Ok(());
         term.visit(&mut visitor)?;
-        visitor.result?
+        std::mem::replace(&mut visitor.result, Ok(()))?;
     }
 
     // convert to a delta predicate that can be applied to kernel scans.
@@ -554,7 +549,6 @@ pub(crate) async fn scan_files_where_matches(
         valid_files,
         predicate,
         delta_predicate,
-        partition_only: visitor.partition_only,
     }))
 }
 
--- a/crates/core/src/operations/delete.rs
+++ b/crates/core/src/operations/delete.rs
@@ -37,10 +37,13 @@ use std::time::Instant;
 
 use async_trait::async_trait;
 use datafusion::catalog::Session;
+use datafusion::common::tree_node::TreeNode;
 use datafusion::common::{ToDFSchema as _, exec_datafusion_err};
 use datafusion::error::Result as DataFusionResult;
 use datafusion::execution::context::SessionState;
+use datafusion::logical_expr::utils::{conjunction, split_conjunction_owned};
 use datafusion::logical_expr::{Extension, LogicalPlan, UserDefinedLogicalNode, lit};
+use datafusion::optimizer::simplify_expressions::simplify_predicates;
 use datafusion::physical_plan::{ExecutionPlan, metrics::MetricBuilder};
 use datafusion::physical_planner::{ExtensionPlanner, PhysicalPlanner};
 use datafusion::prelude::Expr;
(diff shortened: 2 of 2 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -1478,4 +1512,21 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn test_dv_long_mask_retains_remainder_for_next_batch() {
+        use super::{DvMaskResult, consume_dv_mask};
```
ion-elgreco, line 1518: Loving the tests

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -343,13 +355,46 @@ impl DeltaScanMetaStream {
...
+    let padded_rows = batch.num_rows() - selection.len();
+    // Delta Kernel may emit short keep-masks; missing trailing entries are
+    // implicitly `true` (row is kept).
+    let filter = BooleanArray::from_iter(
+        selection
+            .iter()
+            .copied()
+            .chain(std::iter::repeat_n(true, padded_rows)),
+    );
+    Ok((filter_record_batch(&batch, &filter)?, padded_rows))
```
ion-elgreco: Small nit for readability 

```suggestion
    let n_rows_to_pad = batch.num_rows() - selection.len();
    // Delta Kernel may emit short keep-masks; missing trailing entries are
    // implicitly `true` (row is kept).
    let filter = BooleanArray::from_iter(
        selection
            .iter()
            .copied()
            .chain(std::iter::repeat_n(true, n_rows_to_pad)),
    );
    Ok((filter_record_batch(&batch, &filter)?, n_rows_to_pad))
```

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -504,14 +509,47 @@ fn finalize_transformed_batch(
...
+        let needs_rebuild = !can_reuse;
+        if needs_rebuild {
+            let adapter = self.factory.make_adapter(Arc::clone(&source_schema))?;
+            self.cached_source = Some(source_schema);
+            self.cached_adapter = Some(adapter);
+        }
+        match self.cached_adapter.as_ref() {
+            Some(adapter) => adapter.adapt_batch(&batch),
+            None => plan_err!(
+                "schema adapter cache entry missing for source schema: {:?}",
```
ion-elgreco, line 548: We can just do unreachable!() Here right? Or just unwrap directly

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -650,6 +727,57 @@ mod tests {
...
+        let id = adapted
+            .column(0)
+            .as_any()
+            .downcast_ref::<Int64Array>()
+            .unwrap();
+        assert_eq!(id.values(), &[2i64]);
+    }
+
+    #[test]
+    fn test_schema_adapter_recovers_from_inconsistent_cache_state() {
```
ion-elgreco: The thing is this will never happen but if it did then the execution plan crashed

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Fair point,  it's unreachable right now- however I'd lean toward keeping `plan_err!` since we're in execution and panicking kills the query. It costs nothing while giving you a recoverable error if someone refactors `adapt()` later. But I don't feel strongly, happy either way.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Yes it's unreachable today, just like above I kept it as a safety net in case `adapt()` gets refactored and the cache population changes. If you'd rather drop it I won't fight for it. just lmk

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ion-elgreco: Alright let's keep it

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: Honestly I'm overthinking it, I'll remove it to keep it clean

ion-elgreco (approved):
Thanks 🙏

PR #4204, opened 2026-02-16
refactor(python): get add action return arrow table

vsmanish1772:
# Description
**Breaking change (Python API):** `DeltaTable.get_add_actions()` now returns `arro3.core.Table` instead of `arro3.core.RecordBatch`.

This change switches Python bindings to use batched add-actions (`add_actions_batches`) to avoid single-batch overflow risks and construct an Arro3 table from `Vec<RecordBatch>`.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/kernel/snapshot/stream.rs
+++ b/crates/core/src/kernel/snapshot/stream.rs
@@ -2,7 +2,7 @@
 //!
 use arrow::{datatypes::SchemaRef, record_batch::RecordBatch};
 use futures::stream::BoxStream;
-use futures::{Future, Stream, StreamExt};
+use futures::{Stream, StreamExt};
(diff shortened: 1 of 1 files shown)


ion-elgreco:
@rtyler Polars heavily relies on our add_actions_table for their `scan_delta` :P So for python in some form we should expose these add actions with the stats

rtyler:
@ion-elgreco I understand that, but we're basically leaking implementation details all the way from kernel, to engine, to delta-rs :crab:  to delta-rs :snake:  for Polars to then be able to consume it again? 

That's insanity, and not sustainable. We can offer these APIs at the Rust layer which makes more sense for an implementation like Polars to consume, but pulling all this up to Python is fraught with problems

PR #4206, opened 2026-02-16
fix: enforce file-id filter semantics in scan planning

ethan-tyler:
# Description

**Last DF 52 upgrade clean up PR**
 
`file_id` is a synthetic scan column and is not a valid kernel/parquet pushdown target. Predicates referencing `file_id` are now classified as `Unsupported` and excluded from kernel/parquet predicate generation, keeping them as residual DataFusion filters.

Non-`file_id` behavior unchanged: partition only filters remain `Exact`, data-column-only filters remain `Inexact`.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/tests/datafusion_table_provider.rs
+++ b/crates/core/tests/datafusion_table_provider.rs
@@ -58,21 +58,50 @@ fn file_id_at_row(batch: &RecordBatch, row: usize) -> Option<String> {
         .column(file_id_idx)
         .as_any()
         .downcast_ref::<DictionaryArray<UInt16Type>>()?;
+    if row >= dict.len() {
+        return None;
+    }
     if dict.is_null(row) {
         return None;
     }
 
     let key = dict.keys().value(row) as usize;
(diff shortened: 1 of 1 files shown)


crates/core/tests/datafusion_table_provider.rs
```diff
@@ -49,6 +52,29 @@ async fn collect_plan(
...
+    let file_id_idx = batch.schema().index_of("file_id").ok()?;
+    let dict = batch
+        .column(file_id_idx)
+        .as_any()
+        .downcast_ref::<DictionaryArray<UInt16Type>>()?;
+    if dict.is_null(row) {
+        return None;
+    }
+
+    let key = dict.keys().value(row) as usize;
```
rtyler, line 68: `value()` can panic if `row` isn't a valid offset. Can this bail earlier if the offset doesn't fit in the length of `keys()`?

ethan-tyler:
Thank you sir @rtyler!

PR #4208, opened 2026-02-17
chore: set compression for partition optimization as well

rtyler:
This is a good example of different routes for partitioned/non-partitioned tables gets us in trouble


ethan-tyler:
I did an in depth and thorough review and LGTM

PR #4209, opened 2026-02-17
chore: enable snappy compression on checkpoints

rtyler:
our checkpoints are larger than they need to be, and this can lead to longer I/O times when loading large unoptimized checkpoints from storage


ion-elgreco:
Interesting, we had this default compression before, I guess it got lost in the refactors :)

PR #4210, opened 2026-02-17
feat: log compaction

ion-elgreco:
# Description

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/protocol/log_compaction.rs
+++ b/crates/core/src/protocol/log_compaction.rs
@@ -10,8 +10,9 @@ use uuid::Uuid;
 use crate::kernel::{Snapshot, spawn_blocking_with_span};
 use crate::logstore::LogStore;
 use crate::protocol::to_rb;
-
 use crate::{DeltaResult, DeltaTable, DeltaTableError};
+use arrow_json::LineDelimitedWriter;
+use object_store::MultipartUpload;
 
 #[tracing::instrument(skip(log_store, snapshot), fields(operation = "log_compaction", start_version = start_version, end_version = end_version, table_uri = %log_store.root_url()))]
 pub(crate) async fn compact_logs_for(
@@ -51,11 +52,11 @@ pub(crate) async fn compact_logs_for(
 
     let root_store = log_store.root_object_store(operation_id);
 
-    let mut writer = arrow_json::LineDelimitedWriter::new(Vec::new());
+    let mut upload = root_store.put_multipart(&lc_path).await?;
+    let mut buffer = Vec::with_capacity(8 * 1024 * 1024);
 
-    let mut current_batch;
     loop {
-        (current_batch, lc_data) = spawn_blocking_with_span(move || {
+        let (current_batch, lc_data_next) = spawn_blocking_with_span(move || {
             let Some(first_batch) = lc_data.next() else {
                 return Ok::<_, DeltaTableError>((None, lc_data));
             };
@@ -64,16 +65,26 @@ pub(crate) async fn compact_logs_for(
         .await
         .map_err(|e| DeltaTableError::Generic(e.to_string()))??;
 
+        lc_data = lc_data_next;
+
         let Some(batch) = current_batch else {
             break;
         };
+
+        let mut writer = LineDelimitedWriter::new(&mut buffer);
         writer.write(&batch)?;
+        writer.finish()?;
+
+        if buffer.len() >= 5 * 1024 * 1024 {
+            upload.put_part(std::mem::take(&mut buffer).into()).await?;
+        }
     }
-    writer.finish()?;
 
-    let bytes = writer.into_inner();
+    if !buffer.is_empty() {
+        upload.put_part(buffer.into()).await?;
+    }
 
-    root_store.put(&lc_path, bytes.into()).await?;
+    upload.complete().await?;
 
     Ok(())
 }
--- a/python/tests/test_log_compaction.py
+++ b/python/tests/test_log_compaction.py
@@ -1,3 +1,4 @@
+import json
 import pathlib
 from typing import TYPE_CHECKING
 
@@ -19,6 +20,7 @@ def test_log_compaction(tmp_path: pathlib.Path, sample_table: Table):
         / "_delta_log"
         / "00000000000000000000.00000000000000000003.compacted.json"
     )
+
     for i in range(4):
         write_deltalake(str(tmp_table_path), sample_table, mode="append")
 
@@ -28,3 +30,18 @@ def test_log_compaction(tmp_path: pathlib.Path, sample_table: Table):
     delta_table.compact_logs(starting_version=0, ending_version=3)
 
     assert compaction_log_path.exists()
+
+    with open(compaction_log_path) as fp:
+        compaction_data = fp.readlines()
+        compaction_data = [json.loads(row) for row in compaction_data]
+
+        action_types = {}
+        for action in compaction_data:
+            key = next(iter(action.keys()))
+            if key not in action_types:
+                action_types[key] = 1
+            else:
+                action_types[next(iter(action.keys()))] += 1
+
+        assert len(compaction_data) == 6
+        assert action_types == {"add": 4, "protocol": 1, "metaData": 1}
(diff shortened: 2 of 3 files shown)


ethan-tyler:
I'm gonna review today

crates/core/src/protocol/log_compaction.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    if start_version >= end_version {
+        return Err(DeltaTableError::Generic(format!(
+            "Invalid version range: end_version {end_version} must be greater than start_version {start_version}"
+        )));
+    }
+
+    let new_snapshot = KernelSnapshot::builder_from(snapshot.inner.clone())
+        .at_version(end_version)
+        .build(task_engine.as_ref())?;
```
ethan-tyler: `builder_from` is for equal/newer snapshot evolution but fails when the table is at version N and end_version < N.

crates/core/src/protocol/log_compaction.rs
```diff
@@ -0,0 +1,91 @@
...
+    let mut lc_writer = new_snapshot.log_compaction_writer(start_version, end_version)?;
+
+    let lc_url = lc_writer.compaction_path();
+    let lc_path = Path::from_url_path(lc_url.path())?;
+
+    let mut lc_data = lc_writer.compaction_data(engine.as_ref())?;
+
+    let root_store = log_store.root_object_store(operation_id);
+
+    let mut writer = arrow_json::LineDelimitedWriter::new(Vec::new());
```
ethan-tyler: I worried about unbounded memory for writes with large ranges. Would stream to object store be better? If not a valid concern, can iqnore

crates/core/src/protocol/log_compaction.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::{DeltaResult, DeltaTable, DeltaTableError};
+
+pub(crate) async fn compact_logs_for(
+    start_version: u64,
+    end_version: u64,
+    log_store: &dyn LogStore,
+    operation_id: Option<Uuid>,
+    snapshot: &Snapshot,
+) -> DeltaResult<()> {
+    let engine = log_store.engine(operation_id);
```
ethan-tyler, line 25: The checkpoint path has `#[tracing::instrument]` / `spawn_blocking_with_span`, I think this should too.

python/src/lib.rs
```diff
@@ -1558,6 +1559,60 @@ impl RawDeltaTable {
...
+                .map_err(PythonError::from)?;
+            }
+
+            #[allow(clippy::await_holding_lock)]
+            let result = rt().block_on(async {
+                match self._table.lock() {
+                    Ok(table) => compact_logs(
+                        &table,
+                        starting_version as u64,
+                        ending_version as u64,
```
ethan-tyler: Both version params cast i64 -> u64 without validation and negative values wrap silently.

python/tests/test_log_compaction.py
```diff
@@ -0,0 +1,30 @@
...
+from deltalake import (
+    DeltaTable,
+    write_deltalake,
+)
+
+if TYPE_CHECKING:
+    pass
+
+
+def test_log_compaction(tmp_path: pathlib.Path, sample_table: Table):
```
ethan-tyler, line 16: this test only verifies that the compacted file exists, however there's no test that verifies compaction correctness or safety.

python/deltalake/table.py
```diff
@@ -1009,6 +1009,12 @@ def create_checkpoint(self) -> None:
         """
         self._table.create_checkpoint()
 
+    def compact_logs(self, starting_version: int, ending_version: int) -> None:
```
ethan-tyler, line 1020: Do we need docs for this new pub api?

crates/core/src/protocol/log_compaction.rs
```diff
@@ -0,0 +1,91 @@
...
+    writer.finish()?;
+
+    let bytes = writer.into_inner();
+
+    root_store.put(&lc_path, bytes.into()).await?;
+
+    Ok(())
+}
+
+/// Creates checkpoint at current table version
```
ethan-tyler: this is log compaction, not checkpointing

```suggestion
/// Creates a log compaction file for a specified version range
```

python/tests/test_log_compaction.py
ion-elgreco: > this test only verifies that the compacted file exists, however there's no test that verifies compaction correctness or safety. 

I didn't feel the need to test that because that's kernels responsibility :p

rtyler (commented):
Quite straight-forward and simple. I do think @ethan-tyler has made some points worth addressing.

I don't really understand the point of these [log compaction](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#log-compaction-files) files compared to checkpoints, but that's not an @ion-elgreco question to resolve 😆  Seems like more over complication of the protocol. 🤷 🙈 

If these do provide reader benefits, what if we fire-and-forget a task that tries to generate on of these files in the background after a write on `version % 25` or something?  If we succeed in writing a log compaction file, great, if we don't, who cares 😆

ion-elgreco:
> Quite straight-forward and simple. I do think @ethan-tyler has made some points worth addressing.
> 
> I don't really understand the point of these [log compaction](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#log-compaction-files) files compared to checkpoints, but that's not an @ion-elgreco question to resolve 😆 Seems like more over complication of the protocol. 🤷 🙈
> 
> If these do provide reader benefits, what if we fire-and-forget a task that tries to generate on of these files in the background after a write on `version % 25` or something? If we succeed in writing a log compaction file, great, if we don't, who cares 😆

I assume kernel log replay makes use of these if it encounters them. I can do this in a commit hook in a follow up pr

crates/core/src/protocol/log_compaction.rs
ion-elgreco: But snapshotbuilder is not async, so why would you want to do spawn_blocking?

crates/core/src/protocol/log_compaction.rs
ion-elgreco: Do you have reference to docs or code where it would fail, can't really find this

python/deltalake/table.py
ion-elgreco: We can add that later stage ;) for now its a hidden gem

ion-elgreco:
Exact matching of our json logs vs this compaction log is not possible because in the compaction log we drop null columns, suggested by kernel, but our own action to bytes serializer does not. Maybe a small optimization we can do in the future is to also drop null columns

> NOTE: Null columns should not be written to the JSON file. For example, if a row has columns [“a”, “b”] and the value of “b” is null, the JSON object should be written as { “a”: “…” }. Note that including nulls is technically valid JSON, but would bloat the log, therefore we recommend omitting them.

hntd187:
> Exact matching of our json logs vs this compaction log is not possible because in the compaction log we drop null columns, suggested by kernel, but our own action to bytes serializer does not. Maybe a small optimization we can do in the future is to also drop null columns
> 
> > NOTE: Null columns should not be written to the JSON file. For example, if a row has columns [“a”, “b”] and the value of “b” is null, the JSON object should be written as { “a”: “…” }. Note that including nulls is technically valid JSON, but would bloat the log, therefore we recommend omitting them.

This will eventually manifest as a bug of some sort since we're doing different things from kernel.

ethan-tyler:
> This will eventually manifest as a bug of some sort since we're doing different things from kernel.

I see your concern but I don't think this is a correctness bug. Both encodings are semantically equivalent for Delta replay. You should be able to validate this with replay/state equivalence and not byte matching. Worth a follow up to unify serializers across paths, but IMO not a blocker.

hntd187:
> > This will eventually manifest as a bug of some sort since we're doing different things from kernel.
> 
> I see your concern but I don't think this is a correctness bug. Both encodings are semantically equivalent for Delta replay. You should be able to validate this with replay/state equivalence and not byte matching. Worth a follow up to unify serializers across paths, but IMO not a blocker.

Yeah, it's probably fine, it's more we don't know what we don't know. I'm just saying there is prolly some corner case that'll come up as a bug eventually on this. This still LGTM

PR #4211, opened 2026-02-18
fix: delete partition fallback batching and add action coalescing

ethan-tyler:
# Description
some follow up/hardening changes from the partition only delete work done recently 

DELETE partition only fallback and add action evaluation could materialize all actions into a single batch, which breaks on large tables

Changes:
- DELETE fallback uses batched partition metadata instead of single batch materialization
- Shared partition metadata MemTable builder across scan and DELETE paths
- Snapshot fast path for partition only column projection
- add_actions coalescing streams directly into BatchCoalescer instead of pre-collecting
- Python docs note get_add_actions() return type migration

# Documentation

### DIFF
--- a/python/docs/source/usage.rst
+++ b/python/docs/source/usage.rst
@@ -320,21 +320,6 @@ the add actions data using :meth:`DeltaTable.get_add_actions`:
     0  part-00000-c9b90f86-73e6-46c8-93ba-ff6bfaf892a...         440 2021-03-06 15:16:07         True            2                 0          0          2
     1  part-00000-04ec9591-0b73-459e-8d18-ba5711d6cbe...         440 2021-03-06 15:16:16         True            2                 0          2          4
 
-.. note::
-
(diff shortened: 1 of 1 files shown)


crates/core/src/delta_datafusion/find_files.rs
```diff
@@ -316,55 +327,78 @@ async fn find_files_scan(
...
-    if all_batches.is_empty() {
-        return Ok(vec![]);
+/// Build a partition-metadata MemTable from snapshot add actions.
+///
+/// Projects only the `path` column and `partition.*` columns for evaluating
+/// partition predicates without materializing full add-action metadata.
+/// Returns `None` when the snapshot contains no add actions.
+pub(crate) fn add_actions_partition_mem_table(
+    snapshot: &(impl PartitionAddActionsProvider + ?Sized),
+) -> DeltaResult<Option<MemTable>> {
```
ethan-tyler, line 337: this decouples helper input from EagerSnapshot for when we nuke it

python/docs/source/usage.rst
ion-elgreco, line 1: These are the old sphinx docs, no need to update them. We still need to remove them

PR #4219, opened 2026-02-20
feat: added disk spilling for merge

### LINKED ISSUE #4217: [Feature]: Write incrementally to table when merging

If I create a big-ish table like her where there are 200 rows per `id` and 20M rows in total. The write is created such that each id should end up (roughly) in each parquet in the file:

```python
import polars as pl
import numpy as np

ids_size = 100_000
n_col_float = 100
size = 1_000_000
reps = size // ids_size

n_writes_to_delta = 20

def get_one_row_per_id(n):
    df_ids = pl.DataFrame({"id": pl.int_range(0, n, eager=True)})
    df_random = pl.DataFrame(np.random.rand(n, n_col_float))
    return pl.concat([df_ids, df_random], how="horizontal")

for i in range(n_writes_to_delta):
    print(f"Write {i+1}/{n_writes_to_delta}")
    df_multiplier =  pl.DataFrame({"multiplier": pl.int_range(0, reps, eager=True)})
    df = df_multiplier.join(get_one_row_per_id(ids_size), how="cross")
    df.write_delta("delta_table", mode="append")


print(f"Total rows: {n_writes_to_delta * size}. {n_writes_to_delta * size // ids_size} rows per id.")
```

and then try to update it with a new data for just one `id` (so only updating 200 rows) it will OOM 
```python
df_small = get_one_row_per_id(2)

res = DeltaTable("delta_table").merge(
    df_small,"s.id = t.id and t.id in (1,2)", 
    target_alias="t", 
    source_alias="s"
```
As I see it it would be possible to do the merge file per file (or with a some buffer) and write the files incrementally to the table given that the needed DataFusion tooling exists. 

If there was any failure before committing the transaction in the end the written files would just get removed in next vacuum. 

Merge file per file (or with a some buffer) and write the files incrementally to the table

Find a way to spill more to disk such that less is needed to stay in memory.

### Priority

None

### Additional context

_No response_

### Contribution

thomasfrederikhoeck:
Added disk spilling for `merge `similar to optimize functions to allow for merges which touches many files in the target

# Description
I have added functionality for spilling to disk similar to how it works in the optimize functions. If nothing is provided it works as before.

I have added test similar to those for the other spill functions.

I have tested my cases in #4217 which now successfully completes the merge without OOM.

I have used AI (Opus 4.6) for getting a overview of the project structure and for writing most of the code. I have review and verified the code myself. 

Work done:

- create `create_session_state_with_spill_config` (which is just a move and rename of `create_session_state_for_optimize`)
- use `create_session_state_with_spill_config` in existing optimize functions
- use `create_session_state_with_spill_config` for `merge`

# Related Issue(s)
Closes #4217

# Documentation


thomasfrederikhoeck:
I believe the failed test are due to arro3-core==0.7.0 being released. If I downgrad to 0.6.5 locally they run succesfully.

hntd187:
Is there a reason you do not use the existing methods of setting these builder config values? https://github.com/delta-io/delta-rs/blob/main/crates/core/src/delta_datafusion/session.rs#L289 for reference.

hntd187:
I looked a bit further, it looks like this basically did what optimize does, which is fine, but would you be able to remove the optimize and the new path you added and unify them all around the single builder creation path that already exists?

ion-elgreco:
Maybe we should expose the datafusion session configuration instead into Python and then allow to pass that session into any operation.

thomasfrederikhoeck:
@ion-elgreco my gut feeling is that would probably hurt discovery / add complexity of settings a bit for users using delta via other libs such as polars. Are there other examples of settings on the session that you think would be beneficial from the Python side?

ion-elgreco:
> @ion-elgreco my gut feeling is that would probably hurt discovery / add complexity of settings a bit for users using delta via other libs such as polars. Are there other examples of settings on the session that you think would be beneficial from the Python side?

It would simplify our operations api in Python way more since there is one SessionConfig

thomasfrederikhoeck:
@ion-elgreco done. Changes to compare with arro3 types.

rtyler:
These Python test failures I am seeing across multiple pull requests right now, so I'm going to self-assign this, figure it out, and then merge once I have a fix :smile:

thomasfrederikhoeck:
arro3 0.8.0 is released which rolls back some changes such that the comparison should work again between pyarrow and arro3: https://github.com/kylebarron/arro3/pull/483

https://github.com/kylebarron/arro3/pull/483/changes#diff-aac088c485a53a5add25496062171f00197641b1f6a95445110eed12de5259f4

I will change back the tests

EDIT: It looks like the release of 0.8.0 failed https://github.com/kylebarron/arro3/actions/runs/22292112536/job/64481282723

thomasfrederikhoeck:
@ion-elgreco I removed my changes to the tests related to arro3 since 0.8.0 was released. Can you give a review again? :-)

ion-elgreco (approved):

PR #4221, opened 2026-02-22
chore: pin arro3-core under the 0.7.0 release

rtyler:
The 0.7.0 release seems to have broken a number of our tests which
perform tyep checks,  e.g.

        >       assert table_schema.field("price").type == sample_table["price"].type
        E       AssertionError: assert DataType(int64) == arro3.core.DataType<Int64>\n
        E        +  where DataType(int64) = pyarrow.Field<price: int64>.type
        E        +    where pyarrow.Field<price: int64> = field('price')
        E        +      where field = id: string\nprice: int64\nsold: int32\ndeleted: bool.field
        E        +  and   arro3.core.DataType<Int64>\n = arro3.core.ChunkedArray<Int64>\n[\n  [\n    0,\n    1,\n    2,\n    3,\n    4,\n  ]\n]\n.type

        tests/test_merge.py:2400: AssertionError

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -52,19 +52,16 @@ jobs:
 
       - name: Build and install deltalake
         run: |
-          cd python
-          python -m venv venv
-          source venv/bin/activate
           make ${{ env.BUILD_ARGS }}
 
       - name: Install dependencies
         run: |
-          source python/venv/bin/activate
+          source python/.venv/bin/activate
           pip install -r docs/requirements.txt
 
       - name: Build
         run: |
-          source python/venv/bin/activate
+          source python/.venv/bin/activate
           mkdocs build
 
       - name: Deploy


rtyler:
Reported [upstream here](https://github.com/kylebarron/arro3/issues/482)

ion-elgreco (commented):
This pins the package, but we should only pin the dev dependency

rtyler:
@ion-elgreco if this release breaks our tests because datatypes are no longer equal, I don't believe we can trust it.  I'm not sure what benefit only pinning the development dependency would be

ion-elgreco:
> @ion-elgreco if this release breaks our tests because datatypes are no longer equal, I don't believe we can trust it.  I'm not sure what benefit only pinning the development dependency would be

We don't do any comparisons between pyarrow and arro3 types so if should be fine!

ion-elgreco:
Looks like 0.8 got released, lets wait for that then it resolves itself

PR #4223, opened 2026-02-23
fix: generated column expr with SchemaMode::Merge handles missing columns

### LINKED ISSUE #4169: bug: generated column expr fails when SchemaMode::Merge would add referenced column
### What happened?

`SchemaMode::Merge` append fails at planning if a generated column expression
references a nullable column omitted from the input batch.

`with_generated_columns()` resolves expressions against
`plan.schema()` before schema evolution adds missing columns.
`generated_columns.rs:26` calls `parse_predicate_expression(...)` against the
pre-evolution schema, any column that schema evolution would insert as NULL is
unresolvable.

Related: #4156

### Expected behavior

Missing nullable base columns resolve as NULL during generated column
computation. Missing non-nullable base columns fail as a validation error.

A fix would look something like: reorder so generated columns compute after schema evolution, similar to the
merge path's `add_missing_generated_columns`.

### Operating System

None

### Binding

Rust

### Bindings Version

_No response_

### Steps to reproduce

1. Table schema: `id INT NOT NULL`, `user STRING NULL`,
   `computed STRING GENERATED ALWAYS AS (user)`
2. Append batch with only `id`, `SchemaMode::Merge`
3. Planning fails: `user` not found during expr resolution

### Relevant logs

```shell

```

veeceey:
Fixes #4169

When appending with `SchemaMode::Merge`, generated column expressions can reference nullable columns that aren't in the input batch. The problem was that `with_generated_columns()` resolves expressions against the plan schema *before* schema evolution runs -- so any column that schema evolution would add as NULL doesn't exist yet, and `parse_predicate_expression` fails.

The fix catches expression resolution errors in `with_generated_columns()` and falls back to a typed NULL placeholder. This lets the pipeline continue through schema evolution, which adds the missing columns. The `DataValidationExec` then sees `NULL IS NOT DISTINCT FROM NULL = true`, correctly passing validation.

This mirrors the approach already used in `add_missing_generated_columns()` (the merge path), which also inserts NULL placeholders for missing generated columns.

### Test plan
- Added `test_generated_column_referencing_missing_column_uses_null_placeholder` test that reproduces the exact failure described in the issue
- All 7 generated_columns tests pass
- `cargo test -p deltalake-core --lib --features datafusion -- "generated_columns::tests"` passes cleanly

### DIFF
--- a/crates/core/src/operations/write/generated_columns.rs
+++ b/crates/core/src/operations/write/generated_columns.rs
@@ -23,6 +23,18 @@ pub fn gc_is_enabled(snapshot: &EagerSnapshot) -> bool {
         .is_feature_enabled(&TableFeature::GeneratedColumns)
 }
 
+/// Returns `true` when the error indicates that a column referenced in the
+/// generation expression could not be found in the current plan schema.
+/// This happens during `SchemaMode::Merge` when the input batch omits
+/// nullable columns that schema evolution will add later.  All other errors
+/// (e.g. SQL syntax errors, type mismatches) should still be surfaced.
+fn is_column_resolution_error(err: &crate::DeltaTableError) -> bool {
+    let msg = err.to_string();
+    // DataFusion emits "No field named ..." for unresolved column references
+    // and "Schema error: ..." for broader schema resolution failures.
+    msg.contains("No field named") || msg.contains("Schema error")
+}
+
 pub fn with_generated_columns(
     session: &dyn Session,
     plan: LogicalPlan,
@@ -68,11 +80,11 @@ pub fn with_generated_columns(
                 }
                 e
             }
-            Err(_) => {
+            Err(ref err) if is_column_resolution_error(err) => {
                 debug!(
-                    "Could not resolve generation expression for column {}, \
+                    "Could not resolve generation expression for column {} ({}), \
                      inserting NULL placeholder (will be resolved after schema evolution).",
-                    name
+                    name, err
                 );
                 // Use the target data type from the table schema if available,
                 // otherwise fall back to a bare NULL.
@@ -82,6 +94,7 @@ pub fn with_generated_columns(
                     lit(ScalarValue::Null).alias(name)
                 }
             }
+            Err(err) => return Err(err.into()),
         };
         projection.push(expr);
     }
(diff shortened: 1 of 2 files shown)


veeceey:
### Test Results

```
$ cargo test -p deltalake-core --lib --features datafusion -- "generated_columns::tests"

running 7 tests
test operations::write::generated_columns::tests::test_empty_generated_columns ... ok
test operations::write::generated_columns::tests::test_existing_columns_pass_through ... ok
test operations::write::generated_columns::tests::test_multiple_generated_columns ... ok
test operations::write::generated_columns::tests::test_generated_column_referencing_missing_column_uses_null_placeholder ... ok
test operations::write::generated_columns::tests::test_missing_non_generated_nullable_column_does_not_error ... ok
test operations::write::generated_columns::tests::test_add_missing_generated_column ... ok
test operations::write::generated_columns::tests::test_mixed_existing_and_generated_columns ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 630 filtered out; finished in 0.09s
```

crates/core/src/operations/write/generated_columns.rs
```diff
@@ -49,12 +49,40 @@ pub fn with_generated_columns(
...
+            session,
+        ) {
+            Ok(resolved) => {
+                let mut e = resolved.alias(name);
+                if let Ok(field) = table_schema.field_with_name(name) {
+                    e = e.cast_to(field.data_type(), plan.schema())?;
+                }
+                e
+            }
+            Err(_) => {
```
ethan-tyler: This is too broad. We should only catch errors during the schema merge and everything else should surface.

crates/core/src/operations/write/generated_columns.rs
```diff
@@ -49,12 +49,40 @@ pub fn with_generated_columns(
...
+                debug!(
```
ethan-tyler, line 84: Bind the error and include it in the log

crates/core/src/operations/write/generated_columns.rs
```diff
@@ -358,4 +386,61 @@ mod tests {
                 .is_err()
         );
     }
+
+    /// Test that a generated column referencing a column not in the input batch
+    /// does not fail, but instead produces a NULL placeholder.
+    /// This is the core fix for #4169.
+    #[test]
+    fn test_generated_column_referencing_missing_column_uses_null_placeholder() {
```
ethan-tyler, line 407: Nice test. It would be nice to have an e2e test through WriteBuilder &
SchemaMode::Merge that asserts actual write time values and not just
schema shape.

ethan-tyler:
Thanks for the changes, lgtm. Just fix the DCO sign off and we should be good to go. Thanks again @veeceey 

Looks like you need to fmt as well

PR #4227, opened 2026-02-23
feat: vacuum lite mode to avoid storage listing

khalidmammadov:
# Description
Vacuum Lite Mode only deletes Stale Tombstone files, but current implementation does full file listing regardless of Lite or Full mode. 

This change avoids listing storage for Lite mode and tries to simlify and clarify logic by segregating concerns for each mode.

# Related Issue(s)
- closes [#106](https://github.com/delta-io/delta-rs/issues/4228)

# Documentation

Added test cases to test and clarify intend

### DIFF
--- a/crates/core/src/operations/vacuum.rs
+++ b/crates/core/src/operations/vacuum.rs
@@ -537,13 +537,11 @@ fn ok_to_delete(
     keep_files: &HashSet<String>,
     partition_columns: &[String],
 ) -> Result<bool, DeltaTableError> {
-    if valid_files.contains(location) // file is still being tracked in table
+    Ok(
+        !(valid_files.contains(location) // file is still being tracked in table
         || keep_files.contains(&location.to_string()) // file is associated with a version that we are keeping
-        || is_hidden_directory(partition_columns, location)?
-    {
-        return Ok(false);
-    }
-    Ok(true)
+        || is_hidden_directory(partition_columns, location)?),
+    )
 }
 
 /// List files no longer referenced by a Delta table and are older than the retention threshold.


rtyler (approved):

khalidmammadov:
@rtyler thanks for fixing those formatting issues, somehow missed it, even though I remember running those lint checks.. 😔

crates/core/src/operations/vacuum.rs
```diff
@@ -530,13 +528,31 @@ fn is_hidden_directory(partition_columns: &[String], path: &Path) -> Result<bool
...
+/// associated with a kept version, or is a hidden directory.
+fn ok_to_delete(
+    location: &Path,
+    valid_files: &HashSet<Path>,
+    keep_files: &HashSet<String>,
+    partition_columns: &[String],
+) -> Result<bool, DeltaTableError> {
+    if valid_files.contains(location) // file is still being tracked in table
+        || keep_files.contains(&location.to_string()) // file is associated with a version that we are keeping
+        || is_hidden_directory(partition_columns, location)?
```
hntd187: You can probably simplify this.

```rust
Ok(valid_files.contains(location)
        || keep_files.contains(&location.to_string())
        || is_hidden_directory(partition_columns, location)?)
```

crates/core/src/operations/vacuum.rs
```diff
@@ -288,66 +290,62 @@ impl VacuumBuilder {
...
-                        &obj_meta.location
-                    );
-                    continue;
-                } else {
+                        &valid_files,
+                        &keep_files,
+                        partition_columns,
+                    )?
+                {
+                    // For files without tombstones (uncommitted or orphaned files),
```
hntd187, line 328: So I believe if the file is not referenced in any transaction logs then it's not to be touched as part of vacuum lite.

crates/core/src/operations/vacuum.rs
khalidmammadov: good spot, fixed

crates/core/src/operations/vacuum.rs
khalidmammadov: yes, this behaviour not changed.

khalidmammadov:
> Looks good, 

Thanks! 

> the part of removing files not referenced in the transaction log needs to be removed.

Removing orphaned/unrelated is beahivour of Full mode (unchanged in this PR) and is standard feature of Delta (verified with Spark just now). 

And, Lite mode is not (and was not) removing them as it only consulting logs (stale tombstones) for decision.

It's just the fact that Light mode was packed inside Full and hence was going through List operation as well that is now being removed. 

Hope it makes sense, but please do let me know if any followups

hntd187 (approved):
Thanks for addressing, sorry I made a mistake reading the original PR, thanks for clarifying. Everything LGTM

PR #4088, opened 2026-01-16
Upsert

adampolomski:
# Description
The description of the main changes of your pull request

# Related Issue(s)

# Documentation

PR #4230, opened 2026-02-24
feat: optimized upsert operation

adampolomski:
# Description
This introduces a dedicated upsert operation to Delta‑RS, implemented separately from the existing merge logic. The goal is to provide a more efficient path for the common “insert‑or‑update” workload without incurring the full overhead of a general merge.

To validate this, I benchmarked the new operation against the current merge implementation. The results show meaningful improvements in both execution time and memory usage. Benchmark output is attached for reference:
[benchresults.txt](https://github.com/user-attachments/files/25522008/benchresults.txt)

A previous attempt to introduce this functionality was made here: https://github.com/delta-io/delta-rs/pull/4088.
At the time, the maintainers preferred not to add a separate operation. I’m reopening the discussion with concrete performance data to help evaluate whether a dedicated upsert path could be beneficial.

I’m fully open to feedback on whether this should exist as a standalone operation or be integrated as an optimized code path within merge when the semantics match an upsert. My goal is to contribute a measurable performance improvement for a common workload, and I’m happy to adapt the implementation to align with project direction.

# Related Issue(s)
n/a

# Documentation
n/a


ethan-tyler:
Hey @adampolomski  - Thanks for the continued work on this and backing it with performance data. However, Im not convinced we should add a new pub api yet.

I have a hard time justifying the added maintenance and semantic drift risks for a performance optimization that I think we can capture with the existing merge api. 

I think a better first step would be to keep the same core performance idea and add an internal fast path when merge clauses match pure upsert semantics. this should provide us measurable gains without committing to another api contract.

If that path proves stable and broadly valuable, I would be happy to revisit a dedicated upsert api later.

PR #4232, opened 2026-02-25
fix: change visibility of add_action method to public

lizardoluis:
# Description
This pull request makes a small change to the visibility of the `add_action` method in the `LogicalFileView` implementation. This follow the same standard for the `remove_action` method and allows extracting the Add from a LogicalFileView.

### DIFF
--- a/crates/core/src/kernel/snapshot/iterators.rs
+++ b/crates/core/src/kernel/snapshot/iterators.rs
@@ -283,7 +283,10 @@ impl LogicalFileView {
     }
 
     /// Converts this file view into an Add action for log operations.
-    #[deprecated(since = "0.31.0", note = "Use Arrow arrays directly instead of converting to Add actions.")]
+    #[deprecated(
+        since = "0.31.0",
+        note = "Use Arrow arrays directly instead of converting to Add actions."
+    )]
     pub fn add_action(&self) -> Add {
         Add {
             path: self.path().to_string(),


roeap:
Thanks for opening this @lizardoluis. 

Before merging, it would be great to get a sense for the use case to use add actions directly. I.e. we have been trying to get rid of all internal call sites of `add_action` (and `remove_action` for that matter) in favour of processing arrow data directly. In the near-ish future we will have no more use for this in the delta-rs codebase and wanted to remove it entirely (hence `pub(crate)` :)) 

That said, we certainly want to accommodate valid use-cases out there...

lizardoluis:
Hi @roeap , 

What I'm trying to do is to build a `DeltaTableProvider` to hit only some files. The `with_files` method receives a `Vec<Add>`.

rtyler (approved):

lizardoluis:
Hi @rtyler , thanks for approving it. 
I updated the branch now with main, so I think it requires a new approval to run the workflows.

lizardoluis:
Hi @roeap @rtyler , can you guys give a look again please?

roeap:
> What I'm trying to do is to build a DeltaTableProvider to hit only some files. The with_files method receives a Vec<Add>.

Yeah, it was sort of a mistake to carry an API that requires `Add` over to the new table provider. But thats where we are...

Any chance you could mark the method ad deprecated as well so I can just yank it once I cleaned things up 😆?

lizardoluis:
@roeap done. 
I'm not sure though if the deprecated message is proper.

roeap:
Thanks for updating @lizardoluis - i think the message is fine for now. Hopefully it won't be around for long 😆.

roeap:
We do need to run `fmt` one more time it seems.

lizardoluis:
@roeap sorry about that. 
It should be OK now

PR #4233, opened 2026-02-25
feat: optimized upsert path within merge

adampolomski:
# Description
Adds a dedicated upsert‑optimized execution path within the existing merge implementation. When the merge semantics match a pure “insert‑or‑update” pattern, the engine now takes a more efficient route that avoids the additional work required for fully general merge logic.

To validate the impact, I benchmarked the merge behavior before and after this change:

- [before](https://github.com/user-attachments/files/25550070/oldbench.txt)
- [after](https://github.com/user-attachments/files/25550069/newbench3.txt)

This work was originally proposed as a standalone upsert operation: https://github.com/delta-io/delta-rs/pull/4230
Following reviewer feedback, this PR integrates the optimized path directly into merge instead of introducing a new API surface. The goal is to deliver the same performance benefits while keeping the public interface unchanged.

# Related Issue(s)
n/a

# Documentation
n/a

### DIFF
--- a/crates/benchmarks/src/merge.rs
+++ b/crates/benchmarks/src/merge.rs
@@ -460,8 +460,8 @@ pub async fn prepare_source_and_table(
         .filter(expr_fn::random().lt_eq(lit(params.sample_matched_rows)))?;
 
     let rand = cast(
-        expr_fn::random() * lit(i32::MAX),
-        arrow::datatypes::DataType::Int32,
+        expr_fn::random() * lit(u32::MAX),
+        arrow::datatypes::DataType::Int64,
     );
     let not_matched = source
         .filter(expr_fn::random().lt_eq(lit(params.sample_not_matched_rows)))?


crates/core/src/operations/merge/mod.rs
```diff
@@ -98,6 +98,274 @@ use crate::{DeltaResult, DeltaTable, DeltaTableError};
...
+// ---------------------------------------------------------------------------
+// Upsert-path detection
+// ---------------------------------------------------------------------------
+
+/// Attempt to detect whether the merge configuration matches "upsert" semantics:
+///   - Exactly one `when_matched_update` clause with **no** predicate.
+///   - Exactly one `when_not_matched_insert` clause with **no** predicate.
+///   - **No** `when_not_matched_by_source_*` clauses.
+///   - No schema-evolution (`merge_schema == false`).
+///   - The join predicate is a pure AND-conjunction of `source.col = target.col`
```
thomasfrederikhoeck, line 110: How does this align with https://delta-io.github.io/delta-rs/usage/merging-tables/#1-add-partition-columns-to-predicates : 

> As you can see, your filter should specify the partition column(s) and the value(s) you want to target during the merge operation. This is especially important when using the default argument streamed_exec=True in the merge method which disables the use of source table statistics to derive an early pruning predicate. Without these statistics, explicit predicates in your merge condition are required for file pruning.

adampolomski: > predicate="s.id = t.id AND s.month_id = t.month_id AND t.month_id = 202501",  # month_id is a partition column

Nothing changes. Regular merge path is executed.

> predicate="s.id = t.id AND s.month_id = t.month_id",  # month_id is a partition column

Optimized upsert path is executed. Scanning is limited only to relevant partitions. It all happens here:
https://github.com/delta-io/delta-rs/pull/4233/changes#diff-2feaa40bff0c5fb01d5b907aada630dbe61ab747b63cecda05411e20c98f9de0R153

thomasfrederikhoeck: Okay. Will it then work for streaming sources too?

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,597 @@
...
+                            delta_kernel::expressions::Scalar::Null(_) => None,
+                            _ => None,
+                        };
+                        (k.name().to_string(), value)
+                    })
+                    .collect()
+            })
+            .unwrap_or_default();
+
+        Action::Remove(Remove {
```
ethan-tyler: use existing remove_action(true) instead of custom remove

crates/core/src/operations/merge/mod.rs
```diff
@@ -98,6 +98,266 @@ use crate::{DeltaResult, DeltaTable, DeltaTableError};
...
+        return None;
+    }
+    if !matches!(not_match_op.r#type, OperationType::Insert) {
+        return None;
+    }
+
+    // 5. The update must be a full-row replacement: every schema field must have
+    //    an entry in the operations HashMap.  Partial updates cannot be safely
+    //    routed through UpsertBuilder (which always replaces whole rows).
+    if !operations_cover_all_columns(&match_op.operations, schema_fields) {
```
ethan-tyler: this only checks column presence and not that expressions are identify mappings. We should add a check for non-identity merges

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,597 @@
...
+        app_metadata.insert("readVersion".to_owned(), self.snapshot.version().into());
+        if let Ok(metrics_json) = serde_json::to_value(metrics) {
+            app_metadata.insert("operationMetrics".to_owned(), metrics_json);
+        }
+
+        let mut commit_properties = self.commit_properties.clone();
+        commit_properties.app_metadata = app_metadata;
+
+        let partition_columns: Vec<String> = self.snapshot.metadata().partition_columns().to_vec();
+        let operation = DeltaOperation::Write {
```
ethan-tyler: Users are calling merge() but fast path commits as DeltaOperation::Write

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,597 @@
...
+            mode: SaveMode::Append,
+            partition_by: if partition_columns.is_empty() {
+                None
+            } else {
+                Some(partition_columns)
+            },
+            predicate: None,
+        };
+
+        let commit = CommitBuilder::from(commit_properties)
```
ethan-tyler, line 611: the post commit hook is not threaded through here. Canonical
merge passes it at mod.rs:1746. Same handler should flow through
fast path

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,597 @@
...
+        let (add_actions, write_metrics) = write_execution_plan_v2(
+            Some(&self.snapshot),
+            state,
+            physical_plan,
+            partition_columns,
+            self.log_store.object_store(None),
+            Some(self.snapshot.table_properties().target_file_size().get() as usize),
+            None,
+            self.writer_properties.clone(),
+            WriterStatsConfig::new(self.snapshot.table_properties().num_indexed_cols(), None),
```
ethan-tyler, line 296: cdc flag is hardcoded to false regardless of whether the table has CDF enabled. I would either disable fast path when CDF is active or implement CDF correct behavior.

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    /// Commit all changes to the Delta log.
+    async fn commit_changes(
+        &self,
+        actions: Vec<Action>,
+        metrics: &UpsertMetrics,
+        operation_id: Uuid,
+    ) -> DeltaResult<DeltaTable> {
+        let mut app_metadata = self.commit_properties.app_metadata.clone();
```
ethan-tyler: We should mirror canonical merge behavior and skip commits on empty actions to avoid unnecessary version bumps

crates/core/src/operations/merge/mod.rs
adampolomski: Ah, I see your point. No, the "upsert path" is pretty much ignorant about the streaming flag. It will always scan the source extracting all partition column values.
I'll route the streaming cases through the regular merge path for now.

thomasfrederikhoeck: Thanks for clarifying! :-) I think this should be reflected in the docs because I think the different knobs to dial related to merge performance is some of fine details that are hard to discover.

adampolomski:
> Thanks @adampolomski for moving this under merge, I think this is the right direction.
> 
> Did a first pass and left a few comments. I also think we could use some negative tests for where we do not want the fast path to trigger.

Thanks a ton for your review @ethan-tyler. I've addressed all your comments.

crates/core/src/operations/merge/mod.rs
```diff
@@ -98,6 +98,331 @@ use crate::{DeltaResult, DeltaTable, DeltaTableError};
...
+/// This is a best-effort string-based parser that handles the common patterns
+/// produced by the Python helpers and benchmark code.
+fn extract_join_keys_from_str(
+    predicate: &str,
+    source_alias: Option<&str>,
+    target_alias: Option<&str>,
+) -> Option<Vec<String>> {
+    let normalised = predicate.trim().to_lowercase();
+
+    let clauses: Vec<&str> = normalised.split(" and ").collect();
```
hntd187, line 364: What would happen for a expr that is 

``` 
"""s.id = t.id AND
s.something = t.something"""
```

or were these removed at some prior point that I missed?

crates/core/src/operations/merge/upsert.rs
hntd187: +1 this sadly have to be CDC aware or as Ethan mentioned disable the fast path.

ion-elgreco:
@adampolomski  are you using AI to generate this code?

adampolomski:
> @adampolomski are you using AI to generate this code?

@ion-elgreco I'm using the coding agent in my IDE, so it is in some part AI generated.

crates/core/src/operations/merge/upsert.rs
adampolomski: It's disabled. I don't want to expand the scope at this point.

ion-elgreco:
> > @adampolomski are you using AI to generate this code?
> 
> @ion-elgreco I'm using the coding agent in my IDE, so it is in some part AI generated.

Let me remind you of this: https://github.com/delta-io/delta-rs/blob/main/CONTRIBUTING.md#ai-generated-code

crates/core/src/operations/merge/mod.rs
adampolomski: Added line break removal.

I just noticed there was no test case for string predicate. Added one with all possible variations here:
https://github.com/delta-io/delta-rs/pull/4233/changes#diff-12f59fe3c4440b7ae4ee1a5ac810b42c1d7357c246aae7b5770e840e52d3ec52R5139

adampolomski:
> > > @adampolomski are you using AI to generate this code?
> > 
> > 
> > @ion-elgreco I'm using the coding agent in my IDE, so it is in some part AI generated.
> 
> Let me remind you of this: https://github.com/delta-io/delta-rs/blob/main/CONTRIBUTING.md#ai-generated-code

Got it. It's surely not "fully generated by AI with little to no engagement from the author". 
I leveraged AI mostly to adapt my originally standalone upsert op into merge flow. I didn't just blindly ship the generated code though.

Anyway, if there are any remarks regarding the quality of this PR I'm more than happy to address it.

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+                        // Try to parse as integer first, then as string
+                        if let Ok(int_val) = v.parse::<i32>() {
+                            lit(int_val)
+                        } else if let Ok(int64_val) = v.parse::<i64>() {
+                            lit(int64_val)
+                        } else {
+                            lit(v.clone())
+                        }
+                    })
+                    .collect();
```
ethan-tyler: This tries int parse before string, something like "001" becomes lit(1) for a Utf8 partition column. You now have a mismatch and get a duplicate row instead of an update.

I think we should use the column's schema dtype to construct the literal instead

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+                            all_values.extend(uint32_array.iter().flatten().map(|v| v.to_string()));
+                        } else if let Some(uint64_array) =
+                            column.as_any().downcast_ref::<arrow::array::UInt64Array>()
+                        {
+                            all_values.extend(uint64_array.iter().flatten().map(|v| v.to_string()));
+                        } else {
+                            return Err(DeltaTableError::Generic(format!(
+                                "Unsupported partition column type for '{}'",
+                                partition_col
+                            )));
```
ethan-tyler: I don't think we should hard error here. We should fall back to canonical merge and not fail the operation.

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+        state: &SessionState,
+    ) -> DeltaResult<(Vec<Action>, UpsertMetrics)> {
+        // Get unique partition values from source to limit scan scope.
+        // Only consider partition columns that are also join keys.
+        let partition_filters: HashMap<String, Vec<String>> = self
+            .extract_partition_filters()
+            .await?
+            .into_iter()
+            .filter(|(k, _)| self.join_keys.contains(k))
+            .collect();
```
ethan-tyler: Partition columns that are irrelevant to the fast path shouldn't be able to error it out. We should run join key filtering first and then run extract_partition_filters on that subset.

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+            .collect::<Vec<_>>()
+            .join(" AND ")
+    }
+
+    fn build_merge_operation(&self) -> DeltaOperation {
+        const UPDATE_ACTION_TYPE: &str = "update";
+        const INSERT_ACTION_TYPE: &str = "insert";
+
+        DeltaOperation::Merge {
+            predicate: None,
```
ethan-tyler: This marks as read_whole_table for conflict checking. We should mirror canonical merge behavior and narrow the predicate scope to avoid unnecessary conflicts.

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+                for batch in batches {
+                    if let Ok(column_index) = batch.schema().index_of(partition_col) {
+                        let column = batch.column(column_index);
+
+                        if let Some(int_array) =
+                            column.as_any().downcast_ref::<arrow::array::Int32Array>()
+                        {
+                            all_values.extend(int_array.iter().flatten().map(|v| v.to_string()));
+                        } else if let Some(str_array) =
+                            column.as_any().downcast_ref::<arrow::array::StringArray>()
```
ethan-tyler: DF 52+ can produce StringViewArray (Utf8View), and source columns can be LargeUtf8. I would add both to avoid runtime failures.

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+            let file_path_col = batch.column(0);
+
+            if let Some(dict_array) = file_path_col
+                .as_any()
+                .downcast_ref::<arrow::array::DictionaryArray<arrow::datatypes::UInt16Type>>()
+            {
+                let keys = dict_array.keys();
+                let values = dict_array.values();
+                if let Some(str_values) =
+                    values.as_any().downcast_ref::<arrow::array::StringArray>()
```
ethan-tyler: Same StringViewArray gap as the partition extraction path above

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+        let physical_plan = state.create_physical_plan(&logical_plan).await?;
+
+        let partition_columns: Vec<String> = self.snapshot.metadata().partition_columns().to_vec();
+
+        let (add_actions, write_metrics) = write_execution_plan_v2(
+            Some(&self.snapshot),
+            state,
+            physical_plan,
+            partition_columns,
+            self.log_store.object_store(None),
```
ethan-tyler: We should thread operation_id through to match canonical merge behavior 
```suggestion
            self.log_store.object_store(Some(operation_id)),
```

crates/core/src/operations/merge/mod.rs
```diff
@@ -1570,38 +1887,93 @@ impl std::future::IntoFuture for MergeBuilder {
...
+
+                let merge_metrics = MergeMetrics {
+                    num_target_rows_updated: upsert_metrics.num_conflicting_records,
+                    num_target_files_added: upsert_metrics.num_added_files,
+                    num_target_files_removed: upsert_metrics.num_removed_files,
+                    execution_time_ms: upsert_metrics.execution_time_ms,
+                    scan_time_ms: upsert_metrics.scan_time_ms,
+                    rewrite_time_ms: upsert_metrics.write_time_ms,
+                    ..Default::default()
+                };
```
ethan-tyler, line 1949: this is only a subset of MergeMetrics 

All of the fields are listed under pub struct MergeMetrics
https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/merge/mod.rs

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+        while let Some(file_view) = file_stream.next().await {
+            let file_view = file_view?;
+            let path = file_view.path().to_string();
+            if conflicting_file_names.contains(&path) {
+                remove_actions.push(Action::Remove(file_view.remove_action(true)));
+            }
+        }
+
+        Ok(remove_actions)
+    }
```
ethan-tyler, line 399: nit: I think it would be better to swap Vec for Hashset on the conflicting file names before entering the loop. This would prevent scanning linearly for every file in the stream, which could bloat at scale.

crates/core/src/operations/merge/upsert.rs
```diff
@@ -0,0 +1,624 @@
...
+        conflicts_df: &DataFrame,
+    ) -> Result<Vec<String>, DeltaTableError> {
+        use std::collections::HashSet;
+
+        let conflicting_paths = conflicts_df
+            .clone()
+            .select(vec![col(FILE_PATH_COLUMN)])?
+            .distinct()?
+            .collect()
+            .await?;
```
ethan-tyler, line 453: nit: you're deduping twice here

ethan-tyler (changes requested):
Thanks for the follow up changes and the additional tests look good.

Did a deeper pass and found some more spots where the fast path drifts from canonical merge. I left comments on what I was able to find plus a couple other notes. 

I think the key thing is making sure users get the same experience no matter which path runs. We should be improving performance on top of what exists today and not changing behavior. It worth doing another pass with the canonical merge code/docs side by side.

crates/core/src/operations/merge/mod.rs
hntd187: Thanks for updating, 

`lines().collect::<Vec<_>>().join(" ")` would probably be more appropriate here so all different OS line breaks are accounted.

https://doc.rust-lang.org/std/primitive.str.html#method.lines

adampolomski: Thanks, fixed. I also changed the testing strategy. Extracting join keys for upsert is now tested in isolation from merge.

crates/core/src/operations/merge/upsert.rs
adampolomski: As far as I understand  the conflict checker, the predicate for upsert case should be in the form of:
```
partition_a IN (...) AND ... partition_n IN (...) AND source.join_key_a = target.join_key_b AND ... source.join_key_n = target.join_key_b"
```
Did I get this right?

crates/core/src/operations/merge/upsert.rs
adampolomski: I reworked this part not to "if else" through various types and use ScalarValue instead.

ethan-tyler:
Thanks for the effort here @adampolomski. This is a meaningful area and we appreciate the continued interest in contributing.

After multiple review rounds, the maintainers don’t think this PR is ready to merge.

The scope is complex, and the PR is still changing significantly during review (new commits are introducing/expanding behavior while feedback is still being addressed). That makes it a moving target and is not yet in a review ready state.

Even when using AI tooling, we expect PR authors to fully understand and be able to justify their implementation during review. When a PR appears primarily AI generated without clear author ownership of the design and tradeoffs, it doesn’t serve the contributor or maintainers. The main value of review is shared understanding and long term maintainability.

Suggested path forward:
- Move this PR back to draft
- Pick up a “good first issue” to build familiarity with the codebase and contribution workflow
- Revisit this direction via smaller, focused PRs that are easier to review and reason about

The enhancement direction is welcome and we want to see it succeed. It just needs to land in a way that is sustainable for maintainers and contributors. Happy to point you to a good first issue if helpful.

adampolomski:
> Thanks for the effort here @adampolomski. This is a meaningful area and we appreciate the continued interest in contributing.
> 
> 
> 
> After multiple review rounds, the maintainers don’t think this PR is ready to merge.
> 
> 
> 
> The scope is complex, and the PR is still changing significantly during review (new commits are introducing/expanding behavior while feedback is still being addressed). That makes it a moving target and is not yet in a review ready state.
> 
> 
> 
> Even when using AI tooling, we expect PR authors to fully understand and be able to justify their implementation during review. When a PR appears primarily AI generated without clear author ownership of the design and tradeoffs, it doesn’t serve the contributor or maintainers. The main value of review is shared understanding and long term maintainability.
> 
> 
> 
> Suggested path forward:
> 
> - Move this PR back to draft
> 
> - Pick up a “good first issue” to build familiarity with the codebase and contribution workflow
> 
> - Revisit this direction via smaller, focused PRs that are easier to review and reason about
> 
> 
> 
> The enhancement direction is welcome and we want to see it succeed. It just needs to land in a way that is sustainable for maintainers and contributors. Happy to point you to a good first issue if helpful.

@ethan-tyler Thanks for the feedback. I appreciate the time and care put into the reviews. I’m still motivated to contribute this work, as I’m convinced the optimization brings real value. At the same time, I understand the reviewability concerns around scope and the PR becoming a moving target. In hindsight, I should have stabilized and narrowed it earlier. The review comments were genuinely helpful in surfacing those issues. I’m happy to reset and build more familiarity with the codebase first. I’ll pick up a “good first issue” as suggested — pointers would be very welcome. 

For this area specifically, I’d like to propose a smaller, focused first step: adding comprehensive test coverage for merge/upsert scenarios. My goal would be to clearly define and lock in the invariants around merge/upsert behavior, which should make future changes in this area (including performance work) easier and safer to review. Coverage there also seems relatively thin today, so this should provide standalone value. Does that sound like a reasonable first step before revisiting the optimization in smaller, incremental PRs?

PR #4236, opened 2026-02-26
fix: pad DV keep masks to numRecords

### LINKED ISSUE #4235: [Bug]: Python `delection_vectors()` truncates bool list.
### What happened?

MRE
Step 1 - generate a delta with deletion vectors:
```python
import shutil
import os

# Path to the warehouse
warehouse_path = os.path.join(os.getcwd(), "spark-warehouse", "dv_force")

# 1. Clean up the physical directory if it exists
if os.path.exists(warehouse_path):
    shutil.rmtree(warehouse_path)

spark = (
    SparkSession.builder
    .appName("delta-dv-test")
    .config("spark.jars.packages", "io.delta:delta-spark_2.13:4.1.0")
    .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
    .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog")
    .config("spark.databricks.delta.properties.defaults.enableDeletionVectors", "true")
    .getOrCreate()
)

# 2. Drop the metadata
spark.sql("DROP TABLE IF EXISTS dv_force")

# 3. Create data and write it (using overwrite mode to be safe)
df = spark.range(10).coalesce(1)
df.write.format("delta").mode("overwrite").saveAsTable("dv_force")

# 4. Perform the deletion (This should now trigger the DV)
spark.sql("DELETE FROM dv_force WHERE id = 5")
spark.sql("DELETE FROM dv_force WHERE id = 2")

# 5. Show results
spark.sql("SELECT * FROM dv_force").show()

# check json log file for 'deletionVectors'
```
and read in the deletion vector:
```python

table_path = "spark-warehouse/dv_force"
dt = DeltaTable(table_path)

vectors = dt.deletion_vectors()
table = vectors.read_all()

print(table["selection_vector"])
```
This returns a truncated set of bools (observed length 6 bools, expected length 10):
```
arro3.core.ChunkedArray<List(Field { name: "item", data_type: Boolean, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} })>
[
  [
    [true, true, false, true, true, false],
  ]
]
```
The list gets truncated after the highest/last deletion index.

### Expected behavior

The returned bool list length matches the row count.

Ref https://github.com/delta-io/delta-rs/pull/4168

### Operating System

Linux

### Binding

Python

### Bindings Version

main

### Steps to reproduce

See above

### Relevant logs

```shell
See above.
```

ethan-tyler:
# Description
Fixes #4235  - `DeltaTable.deletion_vectors()` returned truncated selection vectors when the highest deleted row index was below the file's total row count.

Kernel returns a sparse DV mask (up to highest deleted index). The api returned that raw mask directly, which could be shorter than numRecords.

**What Changed**
- Plumb `num_records: Option<u64>` through scan replay DV side channel
- Pad short masks with `true` up to `numRecords` at the API boundary
- Error if mask exceeds `numRecords` or `numRecords` is missing

This is now a stricter contract with `deletion_vectors()` now failing if a DV file is missing `numRecords` instead of returning a truncated mask.

**Upstream Kernel Note**
If kernel can return full length selection vectors, this normalization will not be needed. Will look into if an upstream feature on delta-kernel is welcomed for a length aware selection vector api

# Related Issue(s)
- #4235  

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
@@ -49,6 +49,7 @@ use delta_kernel::{
 use futures::{Stream, TryStreamExt as _, future::ready};
 use itertools::Itertools as _;
 use object_store::{ObjectMeta, path::Path};
+use url::Url;
 
 pub use self::exec::DeltaScanExec;
 use self::exec_meta::DeltaScanMetaExec;
@@ -156,9 +157,11 @@ pub(super) async fn replay_deletion_vectors(
     // Only files with `dv_info.has_vector()` spawn tasks, so every item should carry a DV.
     // Guard with a typed error (instead of panic) in case that invariant drifts.
     let dvs: DashMap<_, _> = dv_stream
-        .and_then(|(url, dv)| {
+        .and_then(|(url, dv, num_records)| {
             ready(match dv {
-                Some(keep_mask) => Ok((url.to_string(), keep_mask)),
+                Some(keep_mask) => normalize_dv_keep_mask_for_api(keep_mask, num_records, &url)
+                    .map(|mask| (url.to_string(), mask))
+                    .map_err(DeltaTableError::from),
                 None => Err(DeltaTableError::generic(
                     "Invariant violation: DV task spawned for file without deletion vector",
                 )),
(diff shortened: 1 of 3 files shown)


ion-elgreco (commented):
Maybe we can add a flag: sparse, with default to False in case people want the sparse version of it?

rtyler:
> Maybe we can add a flag: sparse, with default to False in case people want the sparse version of it?

@ion-elgreco I'm not sure I understand the potential use-case for a sparse version of these.  I think it's fine to only add flags when somebody has a strong use-case to justify it though :smile:

ion-elgreco:
> > Maybe we can add a flag: sparse, with default to False in case people want the sparse version of it?
> 
> @ion-elgreco I'm not sure I understand the potential use-case for a sparse version of these.  I think it's fine to only add flags when somebody has a strong use-case to justify it though :smile: 

Haha yeah that's fair

ethan-tyler:
I agree with @rtyler - Happy to revisit in a follow up if there is a compelling use case

kdn36:
Thanks!

PR #4240, opened 2026-02-28
fix(python): handle void type in schema Arrow conversion

### LINKED ISSUE #1947: python DeltaTable.schema().to_pyarrow() does not support data type `void`
# Environment

pip datalake==0.14.0

**Environment**:
- **Cloud provider**: /
- **OS**: Linux

***
# Bug

**What happened**:
I wanted to read the schema from a delta table in pyarrow format via
``` python

file_uri = "..."
deltaTable = DeltaTable(file_uri, storage_options=...)
pyarrow_schema = deltaTable.schema().to_pyarrow()
```

It failed with error:
```
[...]
  File "my_python_file.py", line 63, in __
    pyarrow_schema = deltaTable.schema().to_pyarrow()
Exception: Schema error: Invalid data type for Arrow: void
```

pyarrow has a [null](https://arrow.apache.org/docs/python/generated/pyarrow.null.html#) data type. I would expect that in case of a `void` type in delta, the pyarrow schema returns the column with `null` data type instead of throwing an exception.

**How to reproduce it**:
-

**More details**:
-

ayzmkk-rongbi:
# Description

   Fix `schema().to_pyarrow()` failing with `ArrowError("Invalid data type for Arrow: void")` when a Delta table
  contains void type columns.

   Per the Delta protocol, void data type columns may exist in tables and it is recommended to drop them on reads.
  This PR handles void type at the Python binding layer by filtering void fields during Arrow conversion and mapping
  individual void types to Arrow Null.

   # Related Issue(s)

   - closes #1947

   # Documentation

   - [Delta Protocol - Primitive Types](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types)

### DIFF
--- a/python/src/schema.rs
+++ b/python/src/schema.rs
@@ -12,13 +12,13 @@ use deltalake::kernel::{
 };
 use pyo3::exceptions::{PyException, PyNotImplementedError, PyTypeError, PyValueError};
 use pyo3::types::PyCapsule;
-use pyo3::{IntoPyObjectExt, prelude::*};
-use pyo3_arrow::PyDataType;
-use pyo3_arrow::PyField;
-use pyo3_arrow::PySchema as PyArrow3Schema;
+use pyo3::{prelude::*, IntoPyObjectExt};
 use pyo3_arrow::error::PyArrowResult;
 use pyo3_arrow::export::{Arro3DataType, Arro3Field, Arro3Schema};
 use pyo3_arrow::ffi::to_schema_pycapsule;
+use pyo3_arrow::PyDataType;
+use pyo3_arrow::PyField;
+use pyo3_arrow::PySchema as PyArrow3Schema;
 use std::collections::HashMap;
 use std::sync::Arc;
 
@@ -31,6 +31,24 @@ use crate::utils::warn;
 // See: https://github.com/PyO3/pyo3/issues/1836
 
 // Decimal is separate special case, since it has parameters
+
+fn is_void_type(data_type: &DataType) -> bool {
+    matches!(data_type, DataType::Primitive(p) if p.to_string() == "void")
+}
+
+/// Filter out void fields from a struct type.
+/// Per the Delta protocol: "Existing tables may have void data type columns.
+/// Behavior is undefined for void data type columns but it is recommended
+/// to drop any void data type columns on reads."
+fn filter_void_fields(struct_type: &DeltaStructType) -> PyResult<DeltaStructType> {
+    let fields: Vec<StructField> = struct_type
+        .fields()
+        .filter(|f| !is_void_type(f.data_type()))
+        .cloned()
+        .collect();
+    DeltaStructType::try_new(fields)
+        .map_err(|e| PyException::new_err(format!("Failed to filter void fields: {e}")))
+}
 
 fn schema_type_to_python(schema_type: DataType, py: Python<'_>) -> PyResult<Bound<'_, PyAny>> {
     match schema_type {
@@ -698,7 +716,8 @@ impl StructType {
 
     #[pyo3(text_signature = "($self)")]
     fn to_arrow(&self) -> PyResult<Arro3DataType> {
-        let inner_type = DataType::Struct(Box::new(self.inner_type.clone()));
+        let filtered = filter_void_fields(&self.inner_type)?;
+        let inner_type = DataType::Struct(Box::new(filtered));
         let arrow_type: ArrowDataType = (&inner_type)
             .try_into_arrow()
             .map_err(|err: ArrowError| PyException::new_err(err.to_string()))?;
@@ -717,7 +736,8 @@ impl StructType {
     }
 
     fn __arrow_c_schema__<'py>(&self, py: Python<'py>) -> PyArrowResult<Bound<'py, PyCapsule>> {
-        let inner_type = DataType::Struct(Box::new(self.inner_type.clone()));
+        let filtered = filter_void_fields(&self.inner_type)?;
+        let inner_type = DataType::Struct(Box::new(filtered));
         let arrow_type: ArrowDataType = (&inner_type)
             .try_into_arrow()
             .map_err(|err: ArrowError| PyException::new_err(err.to_string()))?;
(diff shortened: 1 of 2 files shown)


ion-elgreco:
I would like to see an integration test where you create a table with void columns. Because I don't see this even being supported in kernel so it might fail earlier.

python/tests/test_schema.py
```diff
@@ -238,3 +238,64 @@ def test_field_serialization():
...
+def test_void_type_to_arrow():
+    """Test that void type fields are handled correctly in Arrow conversion.
+
+    Per the Delta protocol, void data type columns may exist in tables and
+    it is recommended to drop them on reads.
+    """
+    import pyarrow as pa
+
+    # PrimitiveType("void").to_arrow() should return Arrow Null type
+    void_type = PrimitiveType("void")
```
ethan-tyler: this is rejected before reaching `to_arrow()` which makes this path not reachable

python/src/schema.rs
```diff
@@ -32,6 +32,23 @@ use crate::utils::warn;
...
+/// Per the Delta protocol: "Existing tables may have void data type columns.
+/// Behavior is undefined for void data type columns but it is recommended
+/// to drop any void data type columns on reads."
+fn filter_void_fields(struct_type: &DeltaStructType) -> DeltaStructType {
+    let fields: Vec<StructField> = struct_type
+        .fields()
+        .filter(|f| !is_void_type(f.data_type()))
+        .cloned()
+        .collect();
+    DeltaStructType::try_new(fields).unwrap_or_else(|_| struct_type.clone())
```
ethan-tyler: `unwrap_or_else(|_| struct_type.clone())` is silently masking errors and returns unfiltered fields

python/src/schema.rs
```diff
@@ -149,6 +166,10 @@ impl PrimitiveType {
 
     #[pyo3(text_signature = "($self)")]
     fn to_arrow(&self) -> PyResult<Arro3DataType> {
+        // Handle void type by mapping to Arrow Null type
+        if self.inner_type.to_string() == "void" {
```
ethan-tyler: This handles void in `to_arrow()` but `__arrow_c_schema__()` doesn't have the same check. Callers get different behavior depending on which conversion path they use.

python/src/schema.rs
```diff
@@ -572,6 +593,15 @@ impl Field {
     #[pyo3(text_signature = "($self)")]
     fn to_arrow(&self) -> PyResult<Arro3Field> {
         let inner_type = self.inner.clone();
+        // Handle void type by mapping to Arrow Null type
+        if is_void_type(inner_type.data_type()) {
```
ethan-tyler: same thing here as comment at line 170

python/src/schema.rs
```diff
@@ -698,7 +728,8 @@ impl StructType {
 
     #[pyo3(text_signature = "($self)")]
     fn to_arrow(&self) -> PyResult<Arro3DataType> {
-        let inner_type = DataType::Struct(Box::new(self.inner_type.clone()));
+        let filtered = filter_void_fields(&self.inner_type);
```
ethan-tyler: I am seeing two different policies here. This drops void fields, but PrimitiveType/Field.to_arrow() maps void to Arrow Null.

CONTRIBUTING.md
ethan-tyler, line 1: not sure how these changes are in scope for this PR

ethan-tyler (changes requested):
Thanks for the patch @ayzmkk-rongbi!. Left a few comments

ayzmkk-rongbi:
Thanks for the thorough review @ethan-tyler and @ion-elgreco! I have addressed all the feedback:

1. **CONTRIBUTING.md** - Removed, out of scope.
2. **PrimitiveType/Field void-to-Null** (lines 170, 597) - Removed the dead code. Void type is rejected at construction, so these paths were unreachable.
3. **unwrap_or_else** (line 49) - Changed filter_void_fields to return PyResult and propagate errors properly.
4. **Inconsistent policies** (line 731) - Now there is one consistent policy: **drop void fields** at Schema/StructType level only.
5. **__arrow_c_schema__ consistency** - Both to_arrow() and __arrow_c_schema__() use the same filter_void_fields() path for Schema and StructType.
6. **Test rewritten** - Uses StructType.from_json() to simulate void fields coming from Delta log metadata (the real-world scenario), instead of trying to construct them via Python constructors.

@ion-elgreco regarding integration test with creating a table with void columns - I will look into whether kernel supports this. The current test uses from_json() to simulate what happens when reading an existing table with void columns in its metadata.

ayzmkk-rongbi:
@ion-elgreco You were right. I created an integration test that injects a void column into the Delta log metadata, and the kernel rejects it at `DeltaTable(path)` construction time:

```
_internal.DeltaError: Kernel error: data did not match any variant of untagged enum DataType at line 1 column 201
```

This means the kernel does not support void as a DataType variant, so our Python-level filter_void_fields code is currently unreachable. The fix for #1947 would need to start at the kernel level (delta-kernel-rs) to accept void as a valid DataType before the Python bindings can handle it.

Should I:
1. Close this PR and open an issue on delta-kernel-rs to add void type support, or
2. Keep this PR as-is so the Python layer is ready once kernel support lands?

Happy to help with whichever approach you prefer.

ion-elgreco:
@ayzmkk-rongbi you should start in delta-kernel-rs, once that is merged and release we can support it here as well

ayzmkk-rongbi:
@ion-elgreco Thanks for the direction! I found that there's already an active PR on delta-kernel-rs: [delta-io/delta-kernel-rs#1858](https://github.com/delta-io/delta-kernel-rs/pull/1858) by @cdelmonte-zg, which adds void type support to the kernel. It's been through several rounds of review and was updated on Mar 1 with the latest feedback from maintainers — should be close to merging.

Once that PR is merged and released, I'll update this delta-rs PR to work with the new kernel void support. In the meantime, I'll keep this PR open as a draft so the Python-side changes are ready to go.

Should I close this PR for now and reopen once kernel support lands, or keep it open as-is?

ion-elgreco:
@ayzmkk-rongbi you can keep it in draft until it lands ;)

PR #4248, opened 2026-03-02
fix: move table builder local path check guard to open_table

khalidmammadov:
# Description
Allows DeltaTableBuilder be created from non existent local path. This allows already existing but non functioning feature of building of an uninitialised DeltaTable and that allows calling various builder methods like CreateBuilder for table creation. 

For example, below does not work at the moment due to eager local path check fails. 

```
    let mut table = 
        DeltaTable::try_from_url(new_local_delta_path).await?
        .create()
        .with_table_name("some-table")
        .with_save_mode(SaveMode::Overwrite)
        .with_columns(
            StructType::try_new(vec![
                StructField::new(
                    "x".to_string(),
                    DataType::Primitive(PrimitiveType::Integer),
                    true,
                ),
                StructField::new(
                    "y".to_string(),
                    DataType::Primitive(PrimitiveType::Integer),
                    true,
                ),
            ])?
            .fields()
            .cloned(),
        )
        .await?;
```
`
Error: InvalidTableLocation("Local path \"/tmp/my_table/\" does not exist or you don't have access!")
`

This PR, moves check from builder to open_table where it was originally [reported](https://github.com/delta-io/delta-rs/issues/1916) and [added](https://github.com/delta-io/delta-rs/pull/1917/changes) and modified several times.

The check is moving back to where it's expected (open_table methods) to be inline with original intent. 

After this change DeltaTable can be constructed with an empty state to allow other builder methods be called for local paths sa well.

Adds a new test case and existing/original `test_fail_fast_on_not_existing_path` test passes as before.

# Related Issue(s)

# Documentation


khalidmammadov:
@rtyler can you please review?

khalidmammadov:
_PreS, below NOT AI generated._

> Especially something that adds a new function for validating paths. 

To be fair, I dont actually add a new function as such here but rather merging one logic that span accross two places into one common function.

> Tyler did a large refactoring a few months ago to align a ton of our path parsing to be more consistent. Are the current functions for validating a correct path not sufficient for this purpose? 

They are correct, but there is this old logic for eager check for only local paths is misplaced (from outset) and stops builder usage. Which then stops other functionality builder provides inaccessible. 
 
Result of this PR is that it adds/modifies **below user facing behaviour**: 
1. Since DeltaTableBuilder if pub and if it's used directlty, i.e. outside this crate then _it will not fail for non-existent local paths anymore_. If user relies on this feature then it will be _breaking change_ for them.
2. Conversaly, it allows DeltaTableBuilder to be created for non-existent local paths now, so other functionality of builder can be utilised (like example in the diesc)  
3. Effectively, DeltaTableBuilder becomes lazy (as it was intended originally) now and any illegal operation (like loading/deleting/history/restore/vacuum etc. on table from non-existent path) would fail during that operation with `Error: NotATable("Generic delta kernel error: No files in log segment")` error as they all need state to work on. 

And there is no behavioural change on **open_table** methods.

Also, I went through all usages of `from_url` within the crate and dont see any impact.

On point 1: Alternative approach would be to leave `from_url` alone and add try_from_url that excludes this check and behaves generic way regardless if it's local or cloud storage. Then from_url can be marked as deprecated. 

WDYT?

hntd187:
I appreciate the detail and sounds good if Tyler is good with it so am I.

PR #4249, opened 2026-03-04
feat(python): add post_commithook_properties to alter metadata apis

vsmanish1772:
# Description
adds `post_commithook_properties` support to the missing `table.alter` metadata methods, so cleanup and checkpoint creation behavior can be controlled consistently via `cleanup_expired_logs` and `create_checkpoint` respectively

## Validation
- `make tests`
- `cargo check -p deltalake-python`
- `cargo fmt --all --check`


vsmanish1772:
@ion-elgreco  @rtyler if this pr looks good. can we merge it please?

PR #4250, opened 2026-03-04
refactor: add insert_into and file selection write path to DeltaScan

ethan-tyler:
# Description
Pre-req scaffolding for Merge migration

Adds insert_into support for the next DeltaScan provider and wires file selection through the write path, allowing targeted file rewrites to use DeltaScan directly

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -573,7 +573,7 @@ pub(crate) async fn scan_files_where_matches(
         .build()?
         .table_root()
         .clone();
-    let file_selection = FileSelection::from_paths(
+    let file_selection = FileSelection::from_file_paths(
         valid_files
             .iter()
             .flat_map(|arr| arr.iter().flatten().map(|v| v.to_string())),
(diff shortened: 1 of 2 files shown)


ethan-tyler:
> So curious why you chose paths here and not say partition values for insert into. I think partition values make sense they are a pretty universal concept but I think paths are a potentially slippery slope. What do you think @ethan-tyler ?

I agree partition values are the right user facing abstraction @hntd187

This is plumbing the existing matched file rewrite path from #4188, into `DeltaScanNext` . However, that path is already file granular. Multiple files can share the same partition values, and also needs to work for non partitioned tables. Internally we need concrete file selection once we've resolved which files to rewrite.

I agree raw paths would be a slippery slope if they became the pub contract or if we started deriving partition semantics from path structure. My intent is partition predicates at the API boundary for pruning, file selection internally for execution.

Happy to rename `from_paths` to something like `from_file_ids` if the current naming reads too path centric.

hntd187:
Well whether a partition value write is one file or N is an implementation detail the user shouldn’t care about if they want to tune file size there are configuration properties to help with that. Basically I’m saying partition values should be the only exposed knob to turn on targeted inserts like this. So yeah if it’s just a matter of being more “on the nose” with the naming I’m fine with just renaming.

ethan-tyler:
@hntd187 @ion-elgreco - this good to approve? I have the merge migration queued up, thanks!

ion-elgreco (approved):

PR #4254, opened 2026-03-06
fix: add a central arrow delta type normalization

### LINKED ISSUE #1721: Support parquet files with nanoseconds timestamps when converting a Parquet table to a Delta table

### LINKED ISSUE #3877: [Bug]: `write_deltalake` fails when writing date64 type
### What happened?

The below Python code raises

>_internal.DeltaError: Kernel error: Json error: whilst decoding field 'minValues': whilst decoding field 'sales_date': failed to parse 1760918400000 as Date32"

on python 3.13.7, deltalake 1.2.0 and pyarrow 21.0.0.

```python
from datetime import date
from tempfile import TemporaryDirectory

import pyarrow as pa

with TemporaryDirectory() as tmp_dir:
    delta_table_path = f"{tmp_dir}/DELTA"
    schema = pa.schema([pa.field("sales_date", pa.date64())])
    table = pa.table([pa.array([date(2025, 10, 20)])], schema=schema)

    write_deltalake(delta_table_path, table)
```

When changing the type to pa.date32 it works as expected. Explicitly creating the data as date64 before writing (`pa.table([pa.array([date(2025, 10, 20)], type=pa.date64())], schema=schema)`) does not make a difference.

### Expected behavior

The table should be written without error. Reading it back with 
```python
Deltalake(delta_table_path).to_pandas()
```
should result in a dataframe with 1 row and a column "sales_date" with value "2025-10-20"

### Operating System

macOS

### Binding

Python

### Bindings Version

1.2.0

### Steps to reproduce

Execute my code example as is

### Relevant logs

```shell

```

fvaleye:
# Description
When we write data with unsupported Arrow types (`Date64`, `Timestamp(ns)`), writes either fail with confusing kernel errors or produce tables with incompatible schemas.
The goal was to centralize the normalization (including the opportunity to refactor `convert_to_delta`) into `normalize_for_delta` to convert unsupported Arrow types to their Delta-compatible equivalents before writing.
 This matches the Delta protocol specification and is consistent with how Spark handles these types.

# Related Issue(s)
- Fixes #3877
- Fixes #1721

## Types conversion
| Arrow type | Delta-compatible type |
|---|---|
| `Date64` | `Date32` |
| `Timestamp(s/ms/ns, tz)` | `Timestamp(us, tz)` |

Nested types (Struct, List, FixedSizeList, Map) are normalized recursively.


ion-elgreco:
All the large and view types shouldn't be normalized. The casting logic has been built in such a way that large and view types can pass through so we can avoid costly casting!

python/tests/test_writer.py
ion-elgreco, line 1: Actually normalization we already did for python in here: `delta-rs/python/deltalake/writer/_conversion.py` but it doesn't do it for date64-> date32

fvaleye: Yes, I think it's better to keep normalization on the Rust side and keep Python's specific conversions. The data always passes through Rust for writing, so it's better covered!

roeap:
@fvaleye @ion-elgreco @ethan-tyler 

Just a quick question regarding this.

Personally (and this is just me) I would not want this stuff to happen auto-magically as these transformations are not all loss-less.

The date conversion is fine (unless it panics when we convert too large/small dates), but we are changing data when re-writing timestamps. To me this is an error in business logic that callers need to handle, but we should not take an opinion on?

I worked on projects on the past, where this kind of truncation would lead lead to massively erroneous results in downstream analyses.

ion-elgreco:
> @fvaleye @ion-elgreco @ethan-tyler
> 
> Just a quick question regarding this.
> 
> Personally (and this is just me) I would not want this stuff to happen auto-magically as these transformations are not all loss-less.
> 
> The date conversion is fine (unless it panics when we convert too large/small dates), but we are changing data when re-writing timestamps. To me this is an error in business logic that callers need to handle, but we should not take an opinion on?
> 
> I worked on projects on the past, where this kind of truncation would lead lead to massively erroneous results in downstream analyses.

For python this has been the default for a very long time now :P I personally prefer it because having to expclitily cast timestamps each time because delta protocol is so limiting is annoying

fvaleye:
> @fvaleye @ion-elgreco @ethan-tyler
> 
> Just a quick question regarding this.
> 
> Personally (and this is just me) I would not want this stuff to happen auto-magically as these transformations are not all loss-less.
> 
> The date conversion is fine (unless it panics when we convert too large/small dates), but we are changing data when re-writing timestamps. To me this is an error in business logic that callers need to handle, but we should not take an opinion on?
> 
> I worked on projects on the past, where this kind of truncation would lead lead to massively erroneous results in downstream analyses.

Hey @roeap 👋,
This is a good point!
I wanted to ensure every write (in Rust) follows the [protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md), and enforces what the Delta protocol requires. The alternative is to error out and force users to manually cast, which is what currently happens and might be a pain.

Looking closely, we have:
**Lossless conversions** 
- `Date64` -> `Date32`: same logical day precision, lossless (Date64 stores milliseconds since epoch but still represents dates)
- `Timestamp(s)` -> `Timestamp(us)`: lossless (upscale)
- `Timestamp(ms)` ->`Timestamp(us)`: lossless (upscale)
                                                                                                                                      
**Lossy conversions**
- `Timestamp(ns)`->`Timestamp(us)`: truncates nanoseconds, but **matches the Delta Lake's protocol**

For reference (need verification):
- Spark silently truncates `ns` -> `µs`                                                                                     
- Pandas 3.0 moved to `µs` as the default                                                                                                             
- PyArrow coerces `ns` -> `µs` for older Parquet versions 

> - `Timestamp(ns)`->`Timestamp(us)`: truncates nanoseconds, but **matches the Delta Lake's protocol**

At least, we should warn the user about the truncation. I can create a follow-up PR/issue for adding a `warn!()` log, WDYT?

roeap:
>For python this has been the default for a very long time now :P I personally prefer it because having to expclitily cast timestamps each time because delta protocol is so limiting is annoying

well, particularly in Python, if we remember the [python zen](https://peps.python.org/pep-0020/): "explicit is better than implicit". But i guess the cat's been out of the bag for a while now. 😆.

> At least, we should warn the user about the truncation. I can create a follow-up PR/issue for adding a warn!() log, WDYT?

I guess there seems to be a consent here that this is acceptable. But at least emitting a warning would be great. To be honest, I am a bit surprised that so many engines do this, to me this is nothing short of data corruption 😆.

ethan-tyler:
Thanks @roeap for pointing this out and I agree with your concerns. As a follow up, I think we should separate Delta compat normalization from lossy coercion policy.  

Something like where the core defaults to lossless only (`timestamp[s|ms] ` -> `timestamp[us]`) and rejects lossy cases. 

Add something like a `TemporalCoercionPolicy` that allows Python to opt into permissive mode for backward compat.

For the lossy paths, we can detect actual precision loss before casting rather than silently truncating.

Thoughts?

ion-elgreco:
@roeap we can start making it configurable and then slowly move from default True to False but that would be a big breaking change deltalake python

PR #4257, opened 2026-03-07
feat: consolidate target_file_size and allow unbounded writes

### LINKED ISSUE #3855: Compacting a delta table consistently undershoots the `target_file_size`, creating an unnecessary extra file

abhiaagarwal:
# Description

I was trying to investigate the issue in 3855, and I believe I've figured it out. The problem is that when we perform a compaction, we create a bin of files where their total sum is > the target file size, and then, when we go to write the file, we also apply the exact same target file size. This means if, say, the size of all the files in your bin is 105MB, you write a 100MB file and a 5MB file, which I would argue is undesired behavior. Additionally, due to compression/rounding, I've observed that the 100MB file might be written as 97MB and 4MB respectively, which then leads to infinite compactions on the same set of file that generates wasteful IO.

Instead, this changes compactions to allow a target_file_size to be unbounded when actually performing the rewrite operation. To do this, I had to replumb a lot of the code since everything makes the assumption of target-size writes. Additionally, I found there are three separate places where a default target size is set, so I consolidated them to just one place.

Note that this is a redo of #3901 on my personal account. 

# Related Issue(s)

Closes #3855 

# Documentation

### DIFF
--- a/crates/core/src/operations/delete.rs
+++ b/crates/core/src/operations/delete.rs
@@ -77,6 +77,7 @@ use crate::operations::CustomExecuteHandler;
 use crate::operations::cdc::CDC_COLUMN_NAME;
 use crate::operations::write::execution::write_exec_plan;
 use crate::protocol::DeltaOperation;
+use crate::table::config::TablePropertiesExt as _;
 use crate::table::state::DeltaTableState;
 
 const SOURCE_COUNT_ID: &str = "delete_source_count";
@@ -504,12 +505,14 @@ async fn execute(
     };
 
     let exec = session.create_physical_plan(&write_plan).await?;
+    let target_file_size = Some(snapshot.table_properties().target_file_size());
     let (mut actions, _) = write_exec_plan(
         session,
         log_store.as_ref(),
         snapshot.table_configuration(),
         exec.clone(),
         Some(operation_id),
+        target_file_size,
         write_cdc,
     )
     .await?;
--- a/crates/core/src/operations/write/execution.rs
+++ b/crates/core/src/operations/write/execution.rs
@@ -560,6 +560,7 @@ pub(crate) async fn write_exec_plan(
     table_config: &TableConfiguration,
     exec: Arc<dyn ExecutionPlan>,
     operation_id: Option<Uuid>,
+    target_file_size: Option<NonZeroU64>,
     write_as_cdc: bool,
 ) -> DeltaResult<(Vec<Action>, WriteExecutionPlanMetrics)> {
     let writer_properties = session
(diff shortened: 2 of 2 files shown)


crates/core/src/operations/write/mod.rs
```diff
@@ -278,7 +280,7 @@ impl WriteBuilder {
     }
 
     /// Specify the target file size for data files written to the delta table.
-    pub fn with_target_file_size(mut self, target_file_size: usize) -> Self {
+    pub fn with_target_file_size(mut self, target_file_size: Option<NonZeroU64>) -> Self {
```
ethan-tyler, line 283: What's the rationale for exposing `Option<NonZeroU64> `on the public builder? My concern is the semantics aren't clear to callers. For example calling nothing vs `with_target_file_size(None)` are different but both look like "not set." If unbounded writing is meant to be public, should we make that explicit like a separate method or enum? If it's only needed for compaction rewrites, could we keep the tri state internal?

crates/core/src/operations/optimize.rs
```diff
@@ -266,7 +267,7 @@ impl<'a> OptimizeBuilder<'a> {
     }
 
     /// Set the target file size
-    pub fn with_target_size(mut self, target: u64) -> Self {
+    pub fn with_target_size(mut self, target: NonZeroU64) -> Self {
```
ethan-tyler, line 270: Is there a reason to push `NonZeroU64` onto callers here? We could validate size > 0 internally and return an error keeping same safety but with simpler call sites.

abhiaagarwal: I think it's worthwhile since it catches the error at construction time, as opposed to runtime.

crates/core/src/operations/write/mod.rs
abhiaagarwal: Good call, I'll make this an enum `TargetSize { Unbounded, Bounded(NonZeroU64 }`; that should preserve the null pointer optimization that gives it an U64 size

crates/core/src/operations/optimize.rs
ethan-tyler: Makes sense, thanks!

crates/core/src/operations/write/mod.rs
ethan-tyler: I think that's the right call, thanks!

Issue #4256, opened 2026-03-07
delta_kernel 0.20 upgrade


rtyler:
### Describe the solution you'd like

Running on the treadmill 

### Priority

None

### Contribution

- [ ] I'm willing to submit a pull request for this feature
- [ ] I can help with testing this feature
- [ ] I can help with documentation for this feature

PR #4258, opened 2026-03-07
chore: upgrade to buoyant-kernel and prepare arrow/datafusion upgrade

rtyler:
This pull request has been a wild-ride, but after much discussion with the team in Slack this change brings a distribution of `delta_kernel` published to crates.io as `buoyant_kernel`   This is with delta-kernel-rs at `d8d9eaa0acddf16ebdb3ba273be45e4a409f92f2` with some custom patches layered on toip.

See #4256

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ repository = "https://github.com/delta-io/delta.rs"
 
 [workspace.dependencies]
 #delta_kernel = { package = "buoyant_kernel", path = "../delta-kernel-rs/kernel", features = [
-delta_kernel = { package = "buoyant_kernel", version = "0.21.100", features = [
+delta_kernel = { package = "buoyant_kernel", version = "0.21.101", features = [
     "arrow-57",
     "default-engine-rustls",
     "internal-api",
--- a/crates/core/tests/read_delta_log_test.rs
+++ b/crates/core/tests/read_delta_log_test.rs
@@ -396,7 +396,7 @@ async fn test_update_incremental_does_not_reread_initial_commit() {
     );
 }
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn test_update_incremental_same_version_checkpoint_refresh_skips_redundant_hint_lookup() {
     let n_commits = 2;
     let tmp_dir = tempfile::tempdir().unwrap();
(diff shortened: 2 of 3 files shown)


rtyler:
The one test failure looks legitimate because we have a test case which uses a
checkpoint on a table with Identity Columns enabled. In previous versions of
kernel/delta-rs the Stats schema matched between our hard-coded structs and the structs inside of kernel.

Presumably with better identity columns support in kernel 0.20.0 there is now a
some additional metadata associated with the identity column stats, e.g.
`delta.identity.start` which are not present in our struct values causing a
failure to percolate out of `coalesce_arrays` inside the default engine

rtyler:
This has been an absolute nightmare to debug.

With this test table, the last batch of data going through the ActionReconcilliation thing is missing the `metadata` values on the `id` field from the identity column. I''m still tracing this through, but it's a total miserable slog.

rtyler:
The journey to understanding what has been going wrong with this one identity column test has been... tough.

I at least understand the problem now!  In delta-kernel-rs there is an optimization which reads the schema from the `_last_checkpoint` file's `checkpointSchema` member which _doesn't exist_ in v2 checkpoints. That alone isn't enough to cause problems, but when that field is not present, the `LastCheckpointHint` becomes a lame :duck: with a `None` checkpoint_schema` member, that causes the machinery _deep_ in the bowels here to infer the schema of stats on the checkpoints from the **first sidecar file** which does not end up matching the stats schema that is _supposed_ to be there.


I [mentioned this in Slack](https://delta-users.slack.com/archives/C013LCAEB98/p1775485743637459) too

> I think there are few overlapping bugs at play here, which are also ambiguous in the protocol.
> 
> * Photon produce(d/s) v2 checkpoint files with checkpointMetadata.sidecareFileSchema which omits the metadata fields from the schema. in its stats. I don't have a v2 checkpoint with column mapping but that would also probably cause problems with [#delta-kernel](https://delta-users.slack.com/archives/C04TRPG3LHZ) and that table as **well**
> * [#delta-kernel](https://delta-users.slack.com/archives/C04TRPG3LHZ) does not utilize any schema defined in _last_checkpoint for v2 checkpoints which causes it to read the schema from the first sidecar file in order to build a stats schema.  This is probably wrong, since there is a metaData action encoded in that v2Checkpoint.nonFileActions array
>
> The problems only percolate up to the surface much further down the call chain in coalesce_arrays when two different sets of actions have their stats compared and on has a stats schema including metadata while the other has an empty metadata causing the data type comparison to fail.


Using another test table that @hntd187 shared with me, it looks like the likely _correct_ fix is for kernel to utilize the `v2Checkpoint.nonFileActions` to determine the schema hint correctly off the `LastCheckpointHint` otherwise there will have to be some substantial rewiring of how schema gets to the stats parsing code.

rtyler:
I was able to create a _very narrow_ test case for delta-kernel-rs and have found that the latest `main` of kernel fixes the issue, so now I'm looking at what the backporting of some specific changes might look like.  I'm debating whether it's better to bring us up to `main` or to bring the fixes _down_ from `main` :thinking:

PR #4259, opened 2026-03-08
fix(warnings): change const to static for extension planners and reduce warnings

khalidmammadov:
# Description
PR aims to reduce compiler warnings reported by clippy. 

Const to Static field change: although this is only used in one place ATM the difference of const and statics can cause inconsistencies later on. Const inlines codes and due to LazuLock will create a new instances of Extention Planners in each site. Static avoid this

Rest of the chnages is more lint related or minor.

I only focused on **core** crate and picked only handfull to simplify review process.

# Related Issue(s)
# Documentation

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -1492,13 +1492,12 @@ impl RawDeltaTable {
 
             match mode {
                 SaveMode::Overwrite => {
-                    let _converted_filters =
+                    let converted_filters =
                         convert_partition_filters(partitions_filters.unwrap_or_default())
                             .map_err(PythonError::from)?;
 
-                    let _state = self.cloned_state()?;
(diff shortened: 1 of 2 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -205,29 +205,29 @@ async fn replay_files(
         files.extend(file);
     }
 
-    if let Some(selection) = file_selection {
-        if selection.missing_file_policy == super::MissingFilePolicy::Error {
```
khalidmammadov, line 209: Here we are just replacing nested if with && and code block shifts to left. Rest is the same

ethan-tyler (approved):
LGTM - thanks @khalidmammadov!

Issue #4239, opened 2026-02-27
EPIC: Remove DeltaTableProvider and complete next provider migration


ethan-tyler:
### Is your feature request related to a problem?

`DeltaTableProvider` is the main legacy seam in DataFusion integration. Write/optimize/merge still depends on it despite next provider primitives being merged. Maintaining two parallel provider paths is a constant source of branching, duplication, and subtle correctness risk.

This Epic tracks all the completed, ongoing and planned work to fully remove `DeltaTableProvider` and complete next provider migration. 

### Describe the solution you'd like

Migrate all operations to `DeltaScanNext` & `FileSelection`, then delete `DeltaTableProvider`.

## Already merged
- [x] #4026 
- [x] #4049 
- [x] #4057 
- [x] #4061 
- [x] #4167 
- [x] #4172
- [x] #4174 
- [x] #4187 
- [x] #4188

## In flight
- [ ] #4119
- [ ] #4134 
- [ ] #4080 
- [ ] #4040
- [ ] #4062 

## Blockers (Tracked as sub issues)
- #3528
- #4115
- #4126
- #4143

## Remaining work
- [ ] Land #4119 & #4134 
- [ ] Migrate merge off DeltaTableProvider
- [ ] Resolve #3528, #4115, #4126
- [ ] Land #4080
- [ ] First transaction pattern seam from #4143
- [ ] Remove `DeltaTableProvider` from public exports
- [ ] Delete impl & dead helpers
- [ ] Port legacy provider tests to next provider parity

### Priority

None

### Contribution

- [ ] I'm willing to submit a pull request for this feature
- [ ] I can help with testing this feature
- [ ] I can help with documentation for this feature

PR #4266, opened 2026-03-09
refactor: migrate merge target scan to DeltaScanNext

ethan-tyler:
# Description
Migrate merge target scan to `DeltaScanNext`. Route the merge predicate into file skipping instead of scan filters, and match rewritten files against full file IDs.

# Related Issue(s)
- #4239

# Documentation

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -1444,12 +1444,7 @@ async fn execute(
         .unwrap()
         .survivors();
 
-    let table_root = snapshot
-        .snapshot()
-        .scan_builder()
-        .build()?
-        .table_root()
-        .clone();
+    let table_root = snapshot.table_configuration().table_root().clone();
 
     for action in snapshot.log_data() {
         let rel_path = action.path();


crates/core/src/operations/merge/mod.rs
```diff
@@ -848,19 +854,18 @@ async fn execute(
...
-        log_store.clone(),
-        scan_config.clone(),
+    let file_column = Arc::new(resolve_file_column_name(
+        snapshot.input_schema().as_ref(),
+        None,
     )?);
 
-    let target_provider = provider_as_source(target_provider);
+    let target_provider = provider_as_source(
+        DeltaScanNext::builder()
```
ion-elgreco, line 863: I don't see parquet pushdown being disabled?

crates/core/src/operations/merge/mod.rs
```diff
@@ -1350,7 +1355,7 @@ async fn execute(
                     .when(col(TARGET_UPDATE_COLUMN).is_null(), lit("update"))
                     .end()?,
             )?
-            .drop_columns(&["__delta_rs_path"])? // WEIRD bug caused by interaction with unnest_columns, has to be dropped otherwise throws schema error
+            .drop_columns(&[file_column.as_str()])? // Avoid unnest schema errors.
```
roeap, line 1358: Did we try to skip this recently? Just wondering since I have seen more than one case where DF just fixed stuff.

However, if we don't need the column, might as well drop it.

crates/core/src/operations/merge/mod.rs
```diff
@@ -1434,12 +1444,20 @@ async fn execute(
...
-            if survivors.contains(action.path().as_ref()) {
-                metrics.num_target_files_removed += 1;
-                actions.push(action.remove_action(true).into());
-            }
+    let table_root = snapshot
+        .snapshot()
+        .scan_builder()
+        .build()?
+        .table_root()
+        .clone();
```
roeap: I think `.snapshot()` already has `table_root`? OR was it `.snapshot().snapshot()`?

crates/core/src/operations/merge/mod.rs
```diff
@@ -1434,12 +1444,20 @@ async fn execute(
...
+
+    for action in snapshot.log_data() {
+        let rel_path = action.path();
+        let rel_path_str = rel_path.as_ref();
+
+        if should_remove_rewritten_file(&survivors, rel_path_str, &table_root)? {
+            metrics.num_target_files_removed += 1;
+            actions.push(action.remove_action(true).into());
         }
     }
```
roeap, line 1457: Any chance we can also do this using `file_views`? `log_data` is another API that only works if we have a materialized log.

We do have to be careful though since we are not using the stored data there. I thunk we need to recover that though, otherwise we'll see some more performance surprises as we continue..

crates/core/src/operations/merge/mod.rs
```diff
@@ -1455,8 +1473,53 @@ async fn execute(
...
+            }
+            warn!(
+                %operation_id,
+                metric_names = ?target_files_skipped_metric_names,
+                derived,
+                "Missing target skipped-file metric; deriving from total-files minus scanned-files"
+            );
+            derived
+        },
+    );
```
roeap, line 1517: I know this is not part of this PR, but there are some questions around the skipped metrics ...

Essentially we may eventually not be able to get this number. Specifically, If we push predicates into the log replay we may ever see all files, so we cannot reliably get this number.

There might be quite some effort in getting accurate stats for this. But its also a very interesting metric.

The actual point - this is also using `log_data`, which ... Hoping we night be able to drop this stat, or collect the total count in some iterator above?

The fact that we can ever have an `impossible_state` here is concerning though, no?

crates/core/src/operations/merge/mod.rs
ethan-tyler: You are correct and it was intentional. The merge predicate now goes through file skipping only predicates and  never reaches parquet pushdown. Both the pushdown regression and early filter regression tests pass on this branch.

I think this is a more precise mechanism than the old global disable but lmk if you think differently.

crates/core/src/operations/merge/mod.rs
ethan-tyler: Yes thank you - it's `snapshot.table_configuration().table_root()`

crates/core/src/operations/merge/mod.rs
ethan-tyler: this is the workaround that @ion-elgreco put in. Only change is the file column is dynamically resolved instead of hardcoded to `__delta_rs_path`. File column isn't needed in the CDC projection path.

crates/core/src/operations/merge/mod.rs
ethan-tyler: Agree with the direction - delete/update are using `file_views()` and merge should move there too. However, I don't think that work belongs in this migration PR as there's meaningful architecture work to get that right. If you agree, I'll pick it up as a follow up.

crates/core/src/operations/merge/mod.rs
ethan-tyler: I agree this is the same underlying concern.

The current fallback is best effort with merge deriving skipped files from `total_files - scanned_files` and that depends on having a fully materialized view of active files. If replay becomes more selective, we may no longer have a stable notion of total files at this layer.

I think the long term direction is to move file accounting into scan/replay ownership, or narrow/drop this metric if we cannot define it robustly there.

`impossible_state` should be treated as a defensive guard and not as an expected condition.

crates/core/src/operations/merge/mod.rs
ion-elgreco: Alright, then its fine!

crates/core/src/operations/merge/mod.rs
ion-elgreco: Just make an issue of this so we don't lose track of it :)

ethan-tyler: I think that's what #4269 is. @roeap is that right? Don't want to duplicate an issue. I really want to work this next.

ion-elgreco:
@roeap shall we merge this?

crates/core/src/operations/merge/mod.rs
ion-elgreco: So what is the decided path forward here?

ethan-tyler: If @roeap has no objection- the path forward is to leave the metric as best effort in this PR. Do the fix in follow up work owned by scan/replay. Use #4269 as part of that prerequisite path because moving toward `file_views()` cleanly is harder while `eager file_views()` still misses stored state reuse.

ethan-tyler: I will track this as a issue :)

ethan-tyler:
> yeet

Send itttt

crates/core/src/operations/merge/mod.rs
roeap: Yes, the issue is a prerequisite for using file_views more.

Issue #4269, opened 2026-03-10
Re-use stored state in EagerSnapshot


roeap:
### What happened?

In [this](https://github.com/delta-io/delta-rs/commit/fd79507cc73218999eb14fb205f8be39c4aba7e5) commit we disable using the stored state (Log) to do a log replay and expose the file data.

This causes us to always re-read the log data from storage rather that leveraging the fact that we already did some log replay. In our migrations we are using the `file_views` more and more and paying a full log replay every time.

### Expected behavior

For now I hope we can just re-enable using existing data in the file views method.

Eventually we need to take a closer look at how we can reuse or chache log replay data while also allowing for certain optimizations to be used (like selectively loading stats). But for now we can just use the already fully loaded state including all stats.

### Operating System

None

### Binding

None

### Relevant logs

```shell

```

rtyler:
I would like to see some instrumentation associated with this issue report. The change linked simply passes things through to the underlying Snapshot. If the `KernelSnapshot` is requiring expensive reloading of data from storage then that's a pretty big issue IMHO and not one resolved by simply reverting the commit.

eeroel:
This may be somewhat related https://github.com/delta-io/delta-rs/issues/4276

PR #4279, opened 2026-03-13
fix: restore incremental update to reuse existing table state

### LINKED ISSUE #4276: [Bug]: update_incremental is not incremental
### What happened?

Starting from `update_incremental` here: https://github.com/delta-io/delta-rs/blob/001db7d7646a2100a27fb3ab0e1e4cd4f52fb4e9/crates/core/src/table/mod.rs#L206

The call chain ends up in `KernelSnapshot::builder_for`, which does not trigger the incremental `try_new_from_impl` path in Delta Kernel, causing the whole log to be replayed https://github.com/delta-io/delta-rs/blob/001db7d7646a2100a27fb3ab0e1e4cd4f52fb4e9/crates/core/src/kernel/snapshot/mod.rs#L83

Some relevant discussion on Slack

### Expected behavior

Incremental update

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

_No response_

### Relevant logs

```shell

```

ethan-tyler:
# Description
Fixes #4276

Incremental update from a cached table state could fall off the reuse path and reread older log state from scratch.

The fix here is to route `DeltaTable::update_incremental` back through `DeltaTableState::update` and add a narrow same version checkpoint refresh when `_last_checkpoint` points at the current version but the state isn't checkpoint backed.

Since there is multiple things going on in this area, I intentionally scoped to the regression only. 

This PR does not touch:
- https://github.com/delta-io/delta-kernel-rs/issues/2103
- #4269

Tests cover the original reread bug and checkpoint refresh behavior

# Related Issue(s)
- #4276

# Documentation

### DIFF
--- a/crates/core/src/table/mod.rs
+++ b/crates/core/src/table/mod.rs
@@ -196,15 +196,31 @@ impl DeltaTable {
         self.update_incremental(None).await
     }
 
-    /// Updates the DeltaTable to the latest version by incrementally applying newer versions.
-    /// It assumes that the table is already updated to the current version `self.version`.
+    /// Updates the DeltaTable by incrementally applying newer versions, optionally bounded by
+    /// `max_version`.
+    ///
+    /// This API is forward-only. Use [`DeltaTable::load_version`] to load an older version.
     pub async fn update_incremental(
         &mut self,
         max_version: Option<i64>,
     ) -> Result<(), DeltaTableError> {
-        self.state = Some(
-            DeltaTableState::try_new(&self.log_store, self.config.clone(), max_version).await?,
-        );
+        let Some(state) = self.state.as_mut() else {
+            self.state = Some(
+                DeltaTableState::try_new(&self.log_store, self.config.clone(), max_version).await?,
+            );
+            return Ok(());
+        };
+
+        let current_version = state.version();
+        if let Some(requested_version) = max_version
+            && requested_version < current_version
+        {
+            return Err(DeltaTableError::generic(format!(
+                "Cannot downgrade via update_incremental from version {current_version} to {requested_version}; use load_version"
+            )));
+        }
+
+        state.update(&self.log_store, max_version).await?;
         Ok(())
     }
 
--- a/crates/core/src/table/state.rs
+++ b/crates/core/src/table/state.rs
@@ -22,6 +22,7 @@ use crate::kernel::{
     ARROW_HANDLER, DataType, EagerSnapshot, LogDataHandler, Metadata, Protocol, TombstoneView,
 };
 use crate::logstore::LogStore;
+use crate::protocol::checkpoints::read_last_checkpoint;
 use crate::{DeltaResult, DeltaTableError};
 
 /// State snapshot currently held by the Delta Table instance.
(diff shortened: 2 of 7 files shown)


crates/core/src/table/mod.rs
```diff
@@ -197,14 +197,29 @@ impl DeltaTable {
...
+        let Some(current_version) = self.state.as_ref().map(|state| state.version()) else {
+            self.state = Some(
+                DeltaTableState::try_new(&self.log_store, self.config.clone(), max_version).await?,
+            );
+            return Ok(());
+        };
+
+        if max_version.is_some_and(|version| version < current_version) {
+            self.state = Some(
+                DeltaTableState::try_new(&self.log_store, self.config.clone(), max_version).await?,
```
ion-elgreco: I think we should raise here, update incremental should now allow you to downgrade, because then it's not update anymore haha

crates/core/src/table/state.rs
```diff
@@ -161,12 +162,49 @@ impl DeltaTableState {
...
+                    log_store,
+                    refreshed_version,
+                    loaded_checkpoint_version,
+                )
+                .await?
+        {
+            tracing::trace!(
+                version = refreshed_version,
+                loaded_checkpoint_version = ?loaded_checkpoint_version,
+                "reloading table state to pick up a newer checkpoint at the current version"
```
ion-elgreco, line 184: Why are we reloading the state again if the state wasnt loaded from a checkpoint? This seems strange

crates/core/src/table/mod.rs
ethan-tyler: Fair question and I see your concern. This was the existing explicit version behavior and for this PR I was focused on restoring incremental forward updates. 

`read_delta_log_test` exercises `update_incremental(Some(...))` and `load_version() `delegates through it.

As a long term design, I agree `update_incremental` should be strictly forward only, however I think that's a separate cleanup where `load_version()` owns the downgrade path and we update tests/callers together. lmk what you think

crates/core/src/table/state.rs
ethan-tyler: This is a narrow workaround for a current same version refresh bug/gap in the snapshot update stack.

If state is already at version N but was loaded without a checkpoint, and a checkpoint for N gets written later, a normal same version update will not adopt it because `Snapshot::update()` and `EagerSnapshot::update()` both return early when `target == current`.

The `_last_checkpoint` check is there to cover that specific case until the underlying same version refresh behavior is fixed lower in the stack. It’s not intended as general checkpoint selection logic in delta-rs.

crates/core/src/table/mod.rs
ion-elgreco: But if you are focusing on restoring incremental forward updates, then for that reason we should not allow this code path

crates/core/src/table/state.rs
ion-elgreco: But is this a bug upstream in kernel or in our snapshot updating logic?

crates/core/src/table/mod.rs
ethan-tyler: That’s reasonable. I just was trying to avoid changing explicit version behavior in the same regression fix, but agree and will make the change.

crates/core/src/table/state.rs
ethan-tyler: I think the gap is in our snapshot update stack and not a kernel issue. `Snapshot::update()` and `EagerSnapshot::update()` both no-op when `target == current` which causes us to never adopt a checkpoint written later for the current version.

We should fix the underlying same version update behavior, but that changes update semantics in the snapshot layer and IMO should be a separate follow up. I opened #4280 to track it.  This PR should be a narrow workaround keeps the scope on the regression. Wdyt?

ion-elgreco: I don't understand tbh why we should care if a checkpoint is written at version the state already is in? We already read all logs and reconciled them, we can just keep reading from that point onwards?

ethan-tyler: We care because the base changed and  not that version N's contents changed. State was built from replay, but now there's a checkpoint for N. We want to rebase onto that so future updates start from a checkpoint backed base instead of replayed state. Without it, same version update is always a no-op and we never adopt the checkpoint.

ethan-tyler: Here's an example:

Long lived reader loads table at version N from replay (no checkpoint exists yet). Later, a checkpoint for N gets written and old log files get cleaned up. Same version update is a no-op today, and the loaded state never rebases onto the checkpoint.

Now when the reader tries to advance to N+1, it's working from a replay base whose underlying log files may have been cleaned up. 

You can see this in the checkpoint_writer regression test

ion-elgreco: Okay thanks for the example!

crates/core/src/table/mod.rs
ion-elgreco: Great, if you make the change we can merge it!

PR #4283, opened 2026-03-15
fix(python): make commit control args keyword-only across mutating APIs

### LINKED ISSUE #4252: feature: make commit control args keyword only across mutating APIs

Mutating APIs allow positional use of `commit_properties` and `post_commithook_properties` with inconsistent ordering across methods. Positional calls can silently map to the wrong type like `CommitProperties` into the post commit hook slot.

Rather than fixing methods one at a time, we should adopt a consistent policy across the API surface.

- Make commit control args keyword only with consistent ordering across all mutating APIs:

`..., *, commit_properties=None, post_commithook_properties=None
`

- Apply to TableAlterer methods, write, update, delete, merge, vacuum, restore, etc.
- Update signatures in table.py, lib.rs, _internal.pyi

_No response_

### Priority

Low - Nice to have

### Additional context

Since this would be a breaking change, I propose adding runtime type validation and DeprecationWarning on positional usage in the PR. Then enforce keyword only in a subsequent release.

### Contribution

bhavanaeh:
# Description

Added `deprecate_positional_commit_args` to `_util.py` as a helper that preserves legacy positional behavior per method, emits a DeprecationWarning, rejects invalid usage, and normalizes to canonical commit_properties-first order.

Wired into all public mutating APIs across `table.py`, `writer/convert_to.py`, and `transaction.py`. `write_deltalake` params reordered to canonical order (already keyword-only, no shim needed). `restore` left untouched (already keyword-only).

Unit tests added in `tests/test_util.py`.
# Related Issue(s)

- closes #4252 

Notes:
- `_internal.pyi stubs are already in canonical order
- `create()` and `vacuum()` handle legacy positional args inline (they had extra trailing params beyond the commit args, so the shared helper is bypassed for those two)
Follow-up PR will make all APIs keyword-only and remove the compatibility path

AI disclosure: I used Claude as a coding assistant to help map out affected methods. I reviewed every modification, ran the full test suite locally, and understand each change made.

### DIFF
--- a/python/tests/test_table_read.py
+++ b/python/tests/test_table_read.py
@@ -1,6 +1,5 @@
 import multiprocessing
 import os
-import tempfile
 from concurrent.futures import Executor, ProcessPoolExecutor, ThreadPoolExecutor
 from datetime import date, datetime, timezone
 from pathlib import Path
@@ -1000,10 +999,10 @@ def test_partitions_filtering_partitioned_table():
 
 
 @pytest.mark.pyarrow
-def test_partitions_date_partitioned_table():
+def test_partitions_date_partitioned_table(tmp_path: Path):
     import pyarrow as pa
 
-    table_path = tempfile.gettempdir() + "/date_partition_table"
+    table_path = tmp_path / "date_partition_table"
     date_partitions = [
         date(2024, 8, 1),
         date(2024, 8, 2),


python/deltalake/table.py
```diff
@@ -773,7 +826,7 @@ def merge(
     def restore(
         self,
         target: int | datetime | str,
-        *,
+        *args,
```
bhavanaeh: please let me know if you find the change here a bit odd 😅  I know I technically made this method _more permissive_ by letting it accept positional args. I was trying to be consistent with the change. Happy to revert if preferred.

bhavanaeh:
@ethan-tyler Thanks for the detailed review! These are great calls that I hadn't fully considered. The shared compatibility helper approach makes a lot of sense, and I'm completely aligned with you on the two-step migration plan. I'll rework this PR along those lines. Appreciate the direction!!

ethan-tyler:
Sorry for the long comment! Wanted to provide a thorough explanation, especially as a first time contributor putting in effort. Didn't want to just say take a different approach without laying out the why and the how.

bhavanaeh:
@ethan-tyler please don't apologize, the detailed explanation is genuinely helpful!!

I've reworked the PR around the shared compatibility helper approach. Introduced `deprecate_positional_commit_args` to _util.py -- hope this is now closer to what you have in mind!

Thanks!

python/deltalake/_util.py
```diff
@@ -18,3 +19,38 @@ def encode_partition_value(val: Any) -> str:
...
+def deprecate_positional_commit_args(
+    method_name: str,
+    args: tuple,
+    commit_properties: Any = None,
+    post_commithook_properties: Any = None,
+    legacy_order: tuple[str, ...] = ("commit_properties", "post_commithook_properties"),
+) -> tuple[Any, Any]:
+    if not args:
+        return commit_properties, post_commithook_properties
+    if len(args) > len(legacy_order):
```
ethan-tyler, line 33: This works when commit args were the last positional params, but create and vacuum both had extra trailing params after those (`raise_if_key_not_exists`, `full`, `keep_versions`). Helper doesn't cover those yet.

python/deltalake/table.py
```diff
@@ -199,6 +199,7 @@ def create(
         description: str | None = None,
         configuration: Mapping[str, str | None] | None = None,
         storage_options: dict[str, str] | None = None,
+        *args,
```
ethan-tyler: `create()` has a compat issue here. Old callers could pass `commit_properties`, `post_commithook_properties`, then `raise_if_key_not_exists` positionally. Now that third arg gets swept into `*args` and the helper rejects it.

python/deltalake/table.py
```diff
@@ -548,8 +554,9 @@ def vacuum(
         retention_hours: int | None = None,
         dry_run: bool = True,
         enforce_retention_duration: bool = True,
-        post_commithook_properties: PostCommitHookProperties | None = None,
+        *args,
```
ethan-tyler: Same thing on `vacuum()` - `full` and `keep_versions` used to come after the commit args positionally. Now they land in `*args` and the helper doesn't know what to do with them.

python/tests/test_util.py
ethan-tyler, line 1: The helper tests are good, I think we also need tests that call the real public methods with old positional patterns.  Thinking of something like `create(..., commit, post, False)` and `vacuum(..., post, commit, True, [7])`. As those are the cases still breaking.

bhavanaeh: good call, added!

bhavanaeh:
@ethan-tyler Thanks for the detailed review! Tried addressing all your comments:

- Replaced the shared helper in `create()` and `vacuum()` with inline logic that handles their extra trailing params 
- Added tests in `test_create.py` and `test_vacuum.py` that call the public methods with the old positional patterns

python/deltalake/table.py
```diff
@@ -240,6 +242,21 @@ def create(
...
+                stacklevel=2,
+            )
+            if len(args) > 3:
+                raise TypeError("create() got unexpected positional arguments")
+            commit_properties, post_commithook_properties, raise_if_key_not_exists = (
+                *args,
+                commit_properties,
+                post_commithook_properties,
+                raise_if_key_not_exists,
+            )[:3]
```
ethan-tyler: Explicit parsing feels safer here than tuple slicing for the deprecated positional path.

```suggestion
        if args:
            warnings.warn(
                "Passing commit arguments positionally to create() is deprecated "
                "and will be removed in a future release. Use keyword arguments instead.",
                DeprecationWarning,
                stacklevel=2,
            )
            if len(args) > 3:
                raise TypeError("create() got unexpected positional arguments")
            if len(args) >= 1:
                if commit_properties is not None:
                    raise TypeError(
                        "create() got multiple values for 'commit_properties'"
                    )
                commit_properties = args[0]
            if len(args) >= 2:
                if post_commithook_properties is not None:
                    raise TypeError(
                        "create() got multiple values for 'post_commithook_properties'"
                    )
                post_commithook_properties = args[1]
            if len(args) == 3:
                raise_if_key_not_exists = args[2]
```

python/deltalake/table.py
```diff
@@ -562,14 +580,30 @@ def vacuum(
...
+            )
+            if len(args) > 4:
+                raise TypeError("vacuum() got unexpected positional arguments")
+            post_commithook_properties, commit_properties, full, keep_versions = (
+                *args,
+                post_commithook_properties,
+                commit_properties,
+                full,
+                keep_versions,
+            )[:4]
```
ethan-tyler: Same here - explicit parsing is a bit safer than tuple slicing for the deprecated positional path.

```suggestion
        if args:
            warnings.warn(
                "Passing commit arguments positionally to vacuum() is deprecated "
                "and will be removed in a future release. Use keyword arguments instead.",
                DeprecationWarning,
                stacklevel=2,
            )
            if len(args) > 4:
                raise TypeError("vacuum() got unexpected positional arguments")
            if len(args) >= 1:
                if post_commithook_properties is not None:
                    raise TypeError(
                        "vacuum() got multiple values for 'post_commithook_properties'"
                    )
                post_commithook_properties = args[0]
            if len(args) >= 2:
                if commit_properties is not None:
                    raise TypeError(
                        "vacuum() got multiple values for 'commit_properties'"
                    )
                commit_properties = args[1]
            if len(args) >= 3:
                full = args[2]
            if len(args) == 4:
                keep_versions = args[3]
```

python/tests/test_create.py
```diff
@@ -165,3 +165,11 @@ def test_delta_table_invalid_path_raises_error(tmp_path: pathlib.Path):
...
 
     assert not nonexistent_path.exists()
+
+
+def test_positional_commit_args_raise_deprecation_warning(tmp_path: pathlib.Path):
+    commit = CommitProperties(custom_metadata={"userName": "John Doe"})
+    with pytest.warns(DeprecationWarning, match="positionally"):
+        DeltaTable.create(
+            tmp_path, schema, "error", None, None, None, None, None, commit, None, False
+        )
```
ethan-tyler, line 175: A small wrapper level regression test here would help cover the bespoke `create()` parsing path and duplicate binding edge case.

```suggestion
def test_create_partial_legacy_positional_args_preserve_defaults(
    tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch
):
    import deltalake.table as table_module

    captured: dict[str, tuple] = {}

    def fake_create(*args):
        captured["args"] = args

    def fake_init(self, *args, **kwargs):
        pass

    monkeypatch.setattr(table_module, "_create_deltalake", fake_create)
    monkeypatch.setattr(DeltaTable, "__init__", fake_init)

    commit = CommitProperties(custom_metadata={"userName": "John Doe"})
    with pytest.warns(DeprecationWarning, match="positionally"):
        DeltaTable.create(
            tmp_path, schema, "error", None, None, None, None, None, commit
        )

    create_args = captured["args"]
    assert create_args[4] is True
    assert create_args[9] is commit
    assert create_args[10] is None


def test_create_positional_and_keyword_commit_conflict_raises(
    tmp_path: pathlib.Path,
):
    commit = CommitProperties(custom_metadata={"userName": "John Doe"})
    with pytest.raises(TypeError, match="multiple values for 'commit_properties'"):
        DeltaTable.create(
            tmp_path,
            schema,
            "error",
            None,
            None,
            None,
            None,
            None,
            commit,
            commit_properties=commit,
        )
```

python/tests/test_vacuum.py
```diff
@@ -321,3 +321,14 @@ def test_issue_3745(tmp_path: pathlib.Path):
...
+
+def test_positional_commit_args_raise_deprecation_warning(
+    tmp_path: pathlib.Path, sample_table: Table
+):
+    write_deltalake(tmp_path, sample_table)
+    dt = DeltaTable(tmp_path)
+
+    commit = CommitProperties(custom_metadata={"userName": "John Doe"})
+    with pytest.warns(DeprecationWarning, match="positionally"):
+        dt.vacuum(None, True, True, None, commit, True, [0])
```
ethan-tyler, line 334: Mirroring the `create()` coverage here would help lock in the `vacuum()` compatibility path and duplicate binding behavior too.
```suggestion
def test_vacuum_partial_legacy_positional_args_preserve_defaults():
    captured: dict[str, tuple] = {}

    class StubTable:
        def vacuum(self, *args):
            captured["args"] = args
            return []

    dt = object.__new__(DeltaTable)
    dt._table = StubTable()

    commit = CommitProperties(custom_metadata={"userName": "John Doe"})
    with pytest.warns(DeprecationWarning, match="positionally"):
        dt.vacuum(None, True, True, None, commit)

    assert captured["args"] == (True, None, True, commit, None, False, None)


def test_vacuum_positional_and_keyword_commit_conflict_raises():
    dt = object.__new__(DeltaTable)
    commit = CommitProperties(custom_metadata={"userName": "John Doe"})

    with pytest.raises(TypeError, match="multiple values for 'commit_properties'"):
        dt.vacuum(None, True, True, None, commit, commit_properties=commit)
```

ethan-tyler (commented):
Thanks for sticking with this :) This is so close and just a couple more things to merge. 

Left suggested changes inline for the custom `create()` and `vacuum()` compat paths

With the suggested changes applied and CI fix for the new `*args` params, we will be in good shape!

python/deltalake/table.py
bhavanaeh: @ethan-tyler exactly what I had in mind originally 😅  — went with tuple slicing to keep the code clean/compact but you're right that explicit parsing is clearer and safer here. Applied!

bhavanaeh:
thank you for being so patient with me 😅 @ethan-tyler  -- I'm learning a lot here, appreciate the direction!

PR #4285, opened 2026-03-15
feat(storage): support certificate_path in storage_options

### LINKED ISSUE #4246: Passing cert file in python storage_options
Should be easy to add custom support for this, we have to Call on the ClientOptions.with_root_certificates if we see they custom config key: "certificate_path" and then do 

```rust
let mut buf = Vec::new();
File::open("my_cert.pem").unwrap()
  .read_to_end(&mut buf).unwrap();
let cert = Certificate::from_pem(&buf).unwrap();
```
and pass that into

```rust
pub fn with_root_certificate(mut self, certificate: Certificate) -> Self {
        self.root_certificates.push(certificate);
        self
    }
```

### Discussed in https://github.com/delta-io/delta-rs/discussions/4198

<sup>Originally posted by **sbonthala-0714** February 13, 2026</sup>
I have a custom https s3 end point.Wanted to check if any one of the below options are possible.

1) Is there a way to skip ssl verification similar to how requests module that can be set in storage_options?

2) If the above option is not possible, is there a way to pass cert path in the storage_options without setting SSL_CERT_FILE env variable.

Currently delta rs is able to connect to custom https s3 endpoint with out setting os environment variable.

I am using python delta lake version 0.25.4</div>

bhavanaeh:
# Description

- Added `CertificateConfig` with `#[derive(DeltaConfig)]` to support `certificate_path` in `storage_options` (tried matching the `LimitConfig` pattern 😅 )
- Also loads from `SSL_CERT_FILE` env var
- Shared `client_options_from_certificate()` helper used by S3, Azure, and GCS backends
- Updated docs in `special_configuration.md`

# Related Issue(s)
Closes #4246

## Test plan
- [x] `cargo build` passes for all affected crates
- [x] Unit tests pass (`cargo test -p deltalake-core --lib -- certificate`)
- [ ] Happy to add an integration test if there's an existing pattern for testing against a TLS endpoint

### DIFF
--- a/crates/aws/src/storage.rs
+++ b/crates/aws/src/storage.rs
@@ -58,15 +58,15 @@ impl ObjectStoreFactory for S3ObjectStoreFactory {
                 builder.with_http_connector(SpawnedReqwestConnector::new(runtime.get_handle()));
         }
 
-        for (key, value) in options.iter() {
-            if let Ok(key) = AmazonS3ConfigKey::from_str(&key.to_ascii_lowercase()) {
-                builder = builder.with_config(key, value.clone());
-            }
-        }
-
         if let Some(ref cert_config) = config.certificate {
             if let Some(ref path) = cert_config.certificate_path {
                 builder = builder.with_client_options(client_options_from_certificate(path)?);
+            }
+        }
+
+        for (key, value) in options.iter() {
+            if let Ok(key) = AmazonS3ConfigKey::from_str(&key.to_ascii_lowercase()) {
+                builder = builder.with_config(key, value.clone());
             }
         }
 


ion-elgreco:
@bhavanaeh we have to implement it similar to LimitConfig with a DeltaConfig derive

bhavanaeh:
@ion-elgreco Refactored, matching the LimitConfig pattern. Thanks for the direction! Kept the shared `client_options_from_certificate()` helper since each backend needs to apply it to its own builder.

crates/aws/src/storage.rs
```diff
@@ -63,6 +64,12 @@ impl ObjectStoreFactory for S3ObjectStoreFactory {
             }
         }
 
+        if let Some(ref cert_config) = config.certificate {
+            if let Some(ref path) = cert_config.certificate_path {
+                builder = builder.with_client_options(client_options_from_certificate(path)?);
+            }
+        }
+
```
ion-elgreco, line 66: Ordering matters here, builder.with_client_options overrides the Clientoptions in the builder, which may have previously been set in builder.with_config(). So this will silently ignore certain client options now.

We first should set the client options on the builder and then call builder.with_config

ion-elgreco (approved):
Thanks! :)

PR #4287, opened 2026-03-16
fix: restore incremental snapshot reuse and same version checkpoint adoption

### LINKED ISSUE #4280: bug: same version snapshot update doesn't adopt checkpoint written later
### What happened?

If state is loaded at version N without a checkpoint and a checkpoint for N is written later, same version update doesn't adopt it. Both `Snapshot::update()` and `EagerSnapshot::update()` return early when `target == current`.

Found while fixing #4276. That PR fix #4279 works around it with a narrow `_last_checkpoint` check at the table/state layer,  but the gap exists lower in the snapshot update stack.

### Expected behavior

Same version update should adopt a checkpoint written later for the current version when the loaded state isn't already backed by that checkpoint

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

1. Load table state at version N (no checkpoint exists for N)
2. Write checkpoint for version N
3. Call same version update
4. Update is a no-op - doesn't rebase onto the checkpoint

No error raised. Visible from control flow and both update paths return early. Regression test covering the workaround was added in #4276.

### Relevant logs

```shell

```

ethan-tyler:
# Description

Fixes two related snapshot refresh regressions:

1. `update_incremental()` stopped reusing loaded state and rebuilt from scratch, causing full log rereads from version 0, unnecessary `_last_checkpoint` lookups, and CI failures on windows via invalid paths like `C:\C:\...\_delta_log`.

2. Same version updates ignored checkpoints written after the snapshot was first loaded. This fixes #4280.

Changes:

- Make `update_incremental()` forward only and reuse loaded state via incremental update. Return `DeltaTableError::VersionDowngrade` when callers try to move backward; `load_version()` remains the downgrade path.
- `Snapshot::update()` now adopts a checkpoint written later for the current version. It no-ops when the snapshot is already checkpoint backed.
- Make `EagerSnapshot::update()` reuse the current snapshot/files when nothing changed.

# Related Issue(s)
- #4280

# Documentation

### DIFF
--- a/crates/core/src/kernel/transaction/mod.rs
+++ b/crates/core/src/kernel/transaction/mod.rs
@@ -300,10 +300,9 @@ impl CommitData {
         if !actions.iter().any(|a| matches!(a, Action::CommitInfo(..))) {
             let mut commit_info = operation.get_commit_info();
             commit_info.timestamp = Some(Utc::now().timestamp_millis());
-            app_metadata.insert(
-                "clientVersion".to_string(),
-                Value::String(format!("delta-rs.{}", crate_version())),
-            );
+            app_metadata
+                .entry("clientVersion".to_string())
+                .or_insert(Value::String(format!("delta-rs.{}", crate_version())));
             app_metadata.extend(commit_info.info);
             commit_info.info = app_metadata.clone();
             // commit info should be the first action to support in-commit timestamps.
@@ -1007,6 +1006,7 @@ mod tests {
         LogStore, StorageConfig, commit_uri_from_version, default_logstore::DefaultLogStore,
     };
     use object_store::{ObjectStore, PutPayload, memory::InMemory};
+    use serde_json::json;
     use url::Url;
 
     #[test]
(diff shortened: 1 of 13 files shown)


crates/core/src/protocol/checkpoints.rs
```diff
@@ -324,11 +302,34 @@ pub async fn cleanup_expired_logs_for(
...
+        let maybe_data = storage.get(&file_path).await;
+        let data = match maybe_data {
+            Ok(data) => data.bytes().await?,
+            Err(ObjectStoreError::NotFound { .. }) => return Ok(None),
+            Err(err) => return Err(err.into()),
+        };
+        Ok(serde_json::from_slice(&data)
+            .inspect_err(|e| warn!("invalid _last_checkpoint JSON: {e}"))
+            .ok())
+    }
```
rtyler, line 331: With the `read_last_checkpoint_version` added into the snapshot/mod I thought that code looked somewhat familiar to me then I remembered this file 😄 

Since the Snapshot change is reading from a kernel interface, I do wonder (aloud) whether this should just go away and these tests should use that function if they need to. I'm loathe to have too much redundant parsing code floating between the two crates

crates/core/src/errors.rs
```diff
@@ -60,6 +60,17 @@ pub enum DeltaTableError {
     #[error("Invalid table version: {0}")]
     InvalidVersion(i64),
 
+    /// Error returned when an operation requests an older version than the currently loaded one.
+    #[error(
+        "Cannot downgrade from version {current_version} to {requested_version}; use load_version"
```
rtyler:
```suggestion
        "Cannot downgrade from version {current_version} to {requested_version}; use DeltaTable.load_version()"
```

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -177,11 +184,58 @@ impl Snapshot {
...
         );
 
-        Ok(Arc::new(Self {
+        let snapshot = Arc::new(Self {
             inner: snapshot,
             schema,
             config: self.config.clone(),
-        }))
+        });
+        if snapshot.version() as u64 == current_version {
```
rtyler, line 192: This line caught my eye, we have some silliness going on in the Snapshot. At least as of v0.19.x the KernelSnapshot.version() returns `u64` which we're converting to `i64` in a lot of places. 😮‍💨 

Chances are once upon a time this was i64 and we codified that, and then somebody came to their senses and made it u64 in kernel 😆  I will fix this separately from this PR.

```suggestion
        if snapshot.version() as Version == current_version {
```

ethan-tyler: I agree - This line could use Version instead of u64, but that would only be cosmetic here. Let do that work in #4293

crates/core/src/protocol/checkpoints.rs
ethan-tyler: Agreed - tracking this with #4294

PR #4289, opened 2026-03-16
fix(warning): reduce warnings in core crate

khalidmammadov:
# Description
This is to reduce warning emitted by cargo and clippy. 
It reduces warnings in core crate from 17->13

# Related Issue(s)
Warning reduction related issues:
https://github.com/delta-io/delta-rs/pull/4275
https://github.com/delta-io/delta-rs/pull/4270
https://github.com/delta-io/delta-rs/pull/4259
https://github.com/delta-io/delta-rs/pull/4282

# Documentation
N/A

### DIFF
--- a/.github/workflows/issue_comments.yml
+++ b/.github/workflows/issue_comments.yml
@@ -7,6 +7,19 @@ permissions:
   issues: write
 
 jobs:
+  # When a comment is created on an issue, we can automatically remove the
+  # awaiting-feedback label to make it more clear which issues need to be
+  # revisited
+  remove-labels:
+    runs-on: ubuntu-slim
+    steps:
+      - name: Remove awaiting-feedback label if present
+        run: |
+          gh issue edit \
+            ${{ github.event.issue_number }} \
+            --remove-label "awaiting-feedback" \
+            -R $${ github.repository }}
+
   issue_assign:
     runs-on: ubuntu-slim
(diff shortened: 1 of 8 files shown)


crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -357,10 +357,12 @@ impl DeltaScan {
             SnapshotWrapper::Snapshot(_) => scan_plan.scan.scan_metadata(engine),
             SnapshotWrapper::EagerSnapshot(esn) => {
                 if let Ok(files) = esn.files() {
+                    #[allow(clippy::unnecessary_to_owned)]
```
hntd187: We shouldn't add these, is there a way to address the warning directly?

khalidmammadov: existing to_vec approach was looking the cleanest to make owned value and clippy suggestion wasn't working, but let me take another look

khalidmammadov: fixed, please check again

khalidmammadov: @hntd187 can you please review?

rtyler:
I think I picked up all of these in #4296, once that is merged I will see what the difference looks like here

PR #4296, opened 2026-03-18
fix: default to unsigned longs for all treatment of table versions

### LINKED ISSUE #4293: Inconsistent i64/u64 usage for defining versions

`delta_kernel::Version` is u64, which we convert to and from i64 in a bunch of places for some reason

we should just use `delta_kernel::Version` 

_No response_

### Priority

None

### Additional context

_No response_

### Contribution

rtyler:
I noticed this in an unrelated pull request that we were doing a fair bit of juggling with i64/u64 that is unnecessary. Table versions can literally *never* be negative numbers. This was at one point treated as u64 and somewhere along the way it became i64.

This change pulls through `delta_kernel::Version` which is just a type alias on u64 and provides some clear consistency on what is a table version and what is not here.

Closes #4293

### DIFF
--- a/crates/aws/Cargo.toml
+++ b/crates/aws/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-aws"
-version = "0.14.0"
+version = "0.15.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -13,7 +13,7 @@ rust-version.workspace = true
 
 [dependencies]
 # path dependencies
-deltalake-core = { version = "0.31.0", path = "../core" , features = ["cloud"]}
+deltalake-core = { version = "0.32.0", path = "../core" , features = ["cloud"]}
 
 # workspace dependencies
 async-trait = { workspace = true }
--- a/crates/azure/Cargo.toml
+++ b/crates/azure/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-azure"
-version = "0.14.0"
+version = "0.15.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -12,7 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-deltalake-core = { version = "0.31.0", path = "../core", features = ["cloud"] }
+deltalake-core = { version = "0.32.0", path = "../core", features = ["cloud"] }
 
 # workspace depenndecies
 bytes = { workspace = true }
@@ -22,7 +22,7 @@ tokio = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
-deltalake-core = { version = "0.31.0", path = "../core", features = [
+deltalake-core = { version = "0.32.0", path = "../core", features = [
     "datafusion",
 ] }
 chrono = { workspace = true }
--- a/crates/catalog-glue/Cargo.toml
+++ b/crates/catalog-glue/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-catalog-glue"
-version = "0.15.0"
+version = "0.16.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
@@ -25,7 +25,7 @@ aws-sdk-glue = { version = "1.120", default-features = false, features = [
     # Use the new default HTTPS client stack (no hyper 0.14).
     "default-https-client",
 ] }
-deltalake-core = { version = "0.31.0", path = "../core" }
+deltalake-core = { version = "0.32.0", path = "../core" }
 thiserror = { workspace = true }
 
 [dev-dependencies]
--- a/crates/catalog-unity/Cargo.toml
+++ b/crates/catalog-unity/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-catalog-unity"
-version = "0.15.0"
+version = "0.16.0"
 authors.workspace = true
 keywords.workspace = true
 readme.workspace = true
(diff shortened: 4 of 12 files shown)


crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -348,9 +345,7 @@ impl Snapshot {
         let start_from = log_root.child(
             format!(
                 "{:020}",
-                limit
-                    .map(|l| (self.version() - l as i64 + 1).max(0))
-                    .unwrap_or(0)
+                limit.map(|l| self.version() - (l as u64) + 1).unwrap_or(0)
```
ethan-tyler: This underflows when `limit >= self.version() + 1`, which is a normal case like version-0 tables, or requesting more history than exists. I think this should clamp to `0` instead of underflowing and something like `saturating_sub`/equivalent would fix it.

crates/core/src/table/mod.rs
```diff
@@ -369,7 +372,7 @@ impl DeltaTable {
         &mut self,
         datetime: DateTime<Utc>,
     ) -> Result<(), DeltaTableError> {
-        let mut min_version: i64 = -1;
+        let mut min_version: Version = 0;
```
ethan-tyler: It looks to me that setting `min_version` to `0` breaks vacuumed tables.  If the earliest retained version is > 0, we never record the lower bound and end up probing from 0, which looks like `NotATable`.

This also affects the binary search below when the timestamp is before the first commit, `pivot == 0` and `pivot - 1` underflows under unsigned arithmetic. 

I think this needs something like `Option<Version>` or an equivalent for first seen initialization and the "before earliest commit" case would benefit from a clean exit that doesn't depend on going negative.

crates/core/src/kernel/snapshot/mod.rs
ethan-tyler: I also see this underflows when `limit >= self.version() + 1`, which is a normal case you'd see with version-0 tables or requesting more history than exists etc. I think the easiest fix would be something like `saturating_sub` or an equivalent

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -44,7 +44,7 @@ pub fn create_spec_partition_values<F: FileAction>(
     action_type: Option<&ScalarValue>,
 ) -> Vec<ScalarValue> {
     let mut spec_partition_values = action_type.cloned().map(|at| vec![at]).unwrap_or_default();
-    spec_partition_values.push(ScalarValue::Int64(Some(spec.version)));
+    spec_partition_values.push(ScalarValue::UInt64(Some(spec.version)));
```
ethan-tyler, line 47: `_commit_version` is emitted as `UInt64` here but the schema in `cdf/mod.rs` (~lines 23-34) still declares it as `Int64`

ethan-tyler (commented):
Thanks @rtyler, the migration is looking great! I left some comments below :)

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -525,15 +523,22 @@ impl Snapshot {
         &self,
         log_store: &dyn LogStore,
         app_id: String,
-    ) -> DeltaResult<Option<i64>> {
+    ) -> DeltaResult<Option<Version>> {
```
ethan-tyler: For table versions the underlying model is already `delta_kernel::Version`. However, app transaction versions are different, where `transaction.version` is still i64 at actions.rs:942. Risk is low but it's a separate semantic change.

python/src/lib.rs
```diff
@@ -1953,7 +1954,7 @@ impl RawDeltaTable {
     /// Get the latest transaction version for the given application ID.
     ///
     /// Returns `None` if the application ID is not found.
-    pub fn transaction_version(&self, app_id: String) -> PyResult<Option<i64>> {
+    pub fn transaction_version(&self, app_id: String) -> PyResult<Option<u64>> {
```
ethan-tyler: Same thing on the python side - `Transaction(...)` still accepts signed i64. Makes the binding internally inconsistent.

crates/core/src/table/mod.rs
```diff
@@ -373,7 +376,7 @@ impl DeltaTable {
         let mut min_version: i64 = -1;
         let log_store = self.log_store();
         let prefix = log_store.log_path();
-        let offset_path = commit_uri_from_version(min_version);
+        let offset_path = commit_uri_from_version(None);
```
ethan-tyler, line 379: This path still depends on an `i64` sentinel (`-1`) via `commit_uri_from_version(None)` and `try_into().unwrap()` and `assert!` later. We can also still panic instead of returning an error. 

Would it be better to keep this in Version end to end and use a separate offset helper rather than encoding a fake -1 through the commit path api?

ethan-tyler (commented):
left a couple of comments thanks!

crates/core/src/table/mod.rs
rtyler: It is perhaps better, I'm not terribly interested in addressing this bit of technical debt right now

crates/core/src/kernel/snapshot/mod.rs
rtyler: I was thinking about this, the [protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers) is a little unclear on whether this is a signed or unsigned long. 

I'm not passionate about this here, so I'll just walk it back because it is ambiguous

crates/core/src/table/mod.rs
ethan-tyler: No worries just wanted get your thoughts 😀

crates/core/src/table/mod.rs
```diff
@@ -392,37 +395,47 @@ impl DeltaTable {
...
+        };
         let mut max_version = match self
             .log_store
-            .get_latest_version(self.version().unwrap_or(min_version))
+            .get_latest_version(self.version().unwrap_or(latest_default_version))
             .await
         {
             Ok(version) => version,
             Err(DeltaTableError::InvalidVersion(_)) => {
+                println!("this shit again");
```
khalidmammadov: did you mean to leave this? :)

PR #4302, opened 2026-03-19
fix: propagate schema_force_view_types through merge write projection

sheganinans:
### Problem                                                                                      

Delta Lake merge operations fail with `DeltaError: Generic DeltaTable error: worker join error   
when driving partition` when a table contains string or binary columns whose aggregate data        
exceeds 2 GB per batch.                                                                            

The scan side of the merge correctly reads target data as `Utf8View`/`BinaryView` when           
`schema_force_view_types` is enabled (the default). These view types use a different memory layout 
that is not subject to the 2 GB offset limit of legacy `Utf8`/`Binary` arrays.                     

However, the merge's **output write projection** casts columns back to `Utf8`/`Binary` (32-bit   
offsets) because `delta_field.data_type().try_into_arrow()` maps Delta's `"string"` type to Arrow  
`Utf8` and `"binary"` to Arrow `Binary`. This cast reconstructs a contiguous offset array,         
re-introducing the 2 GB limit that view types were designed to eliminate.                          

### Fix                                                                                          

Instead of re-deriving Arrow types from the Delta kernel schema (which always gives              
`Utf8`/`Binary`), the write projection now takes its cast targets from the **target scan schema**, 
which already has the correct physical types applied by `DeltaScanConfig` (e.g.                    
`Utf8View`/`BinaryView` when `schema_force_view_types` is on). Dictionary-wrapped partition        
columns and schema-evolution columns fall back to the kernel-derived type.                         

The caller's session is also threaded to `DeltaScanNext::builder()` so that                      
`schema_force_view_types` is authoritative across both the target scan and the write projection,   
rather than only at the write layer.                                                               

### Not a breaking change                                                                        

This change only affects the **in-memory Arrow representation** during merge execution. The      
persisted format is unchanged:                                                                     
- Parquet files are byte-for-byte identical (`Utf8View` and `Utf8` both write as parquet         
`BYTE_ARRAY`)                                                                                      
- The Delta log schema remains `"string"` / `"binary"`                                           
- Readers determine their own Arrow type mapping independently of what the writer used           
internally                                                                                         

### Behavior                                                                                     

- **`schema_force_view_types = true` (default)**: Merge uses `Utf8View`/`BinaryView` in the      
write projection — no 2 GB limit                                                                   
- **`schema_force_view_types = false`**: Legacy `Utf8`/`Binary` behavior preserved, no change    

### Files changed                                                                                

- `crates/core/src/operations/merge/mod.rs` — Write projection derives cast targets from the     
target scan schema; session threaded to scan builders                                              
- `crates/core/tests/command_merge.rs` — `test_merge_respects_schema_force_view_types` asserts   
the physical Arrow types in both enabled and disabled modes

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -1249,10 +1249,7 @@ async fn execute(
                 col_ref
             }
         } else {
-            null_target_column = Some(cast(
-                lit(ScalarValue::Null).alias(name),
-                cast_type.clone(),
-            ));
+            null_target_column = Some(cast(lit(ScalarValue::Null).alias(name), cast_type.clone()));
             Column::new(source_qualifier.clone(), name)
         };
 
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -910,6 +910,7 @@ async fn execute(
         DeltaScanNext::builder()
             .with_eager_snapshot(snapshot.clone())
             .with_log_store(log_store.clone())
+            .with_session(state.clone().into())
             .with_file_column(file_column.as_str())
             .await?,
     );
@@ -918,6 +919,7 @@ async fn execute(
 
     let source_schema = source.schema();
     let target_schema = target.schema();
+    let target_arrow_schema = Arc::new(target_schema.as_arrow().clone());
 
     let join_schema_df = build_join_schema(source_schema, target_schema, &JoinType::Full)?;
 
@@ -968,6 +970,7 @@ async fn execute(
         let mut builder = DeltaScanNext::builder()
             .with_eager_snapshot(snapshot.clone())
             .with_log_store(log_store.clone())
+            .with_session(state.clone().into())
             .with_file_column(file_column.as_str());
 
         if !file_skipping_predicates.is_empty() {
@@ -1227,7 +1230,13 @@ async fn execute(
             None => TableReference::none(),
         };
         let name = delta_field.name();
-        let mut cast_type: DataType = delta_field.data_type().try_into_arrow()?;
+        let kernel_arrow_type = delta_field.data_type().try_into_arrow()?;
+        let mut cast_type: DataType = target_arrow_schema
+            .field_with_name(name)
+            .ok()
+            .filter(|f| !matches!(f.data_type(), DataType::Dictionary(_, _)))
+            .map(|f| f.data_type().clone())
+            .unwrap_or(kernel_arrow_type);
 
         // Receive the correct column reference given that some columns are only in source table
         let column = if let Some(field) = snapshot.schema().field(name) {
@@ -1242,7 +1251,7 @@ async fn execute(
         } else {
             null_target_column = Some(cast(
                 lit(ScalarValue::Null).alias(name),
-                delta_field.data_type().try_into_arrow()?,
+                cast_type.clone(),
             ));
             Column::new(source_qualifier.clone(), name)
         };
--- a/crates/core/tests/command_merge.rs
+++ b/crates/core/tests/command_merge.rs
@@ -6,7 +6,7 @@ use arrow_schema::{DataType, Field, Schema as ArrowSchema};
 use datafusion::common::Column;
 use datafusion::dataframe::DataFrame;
 use datafusion::logical_expr::{Expr, col, lit};
-use datafusion::prelude::SessionContext;
+use datafusion::prelude::{SessionConfig, SessionContext};
 use deltalake_core::kernel::transaction::TransactionError;
 use deltalake_core::kernel::{DataType as DeltaDataType, PrimitiveType, StructField, StructType};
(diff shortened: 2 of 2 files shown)


ion-elgreco (dismissed):
We should reuse the existing casting module, this has already been covered there to passthrough view types in the schema. 

This also just highlights that AI is still not there, a fix is created that just introduces new logic that makes it even more difficult to maintain this project

ethan-tyler (commented):
Thanks for putting this forward @sheganinans - a couple of additional initial thoughts

The session config isn't authoritative across the full merge pipeline. `with_session_state(...)` only applies `schema_force_view_types` at the write projection layer. The target scan path can still use default provider config instead of the caller's session. This is something we've been tightening up in existing code and want new PRs to follow the same pattern.

Also, the test proves merge succeeds but doesn't assert the actual arrow physical types in the output. For a regression fix we should be checking observable type behavior and not just that it doesn't error.

sheganinans:
Thank you both @ion-elgreco @ethan-tyler for the feedback, I'll go through the code and address the issues you brought up. 

What brought me to bring this PR is that this issue is currently causing us issues in our stack, we have pretty large strings (we are downloading SEC files) in our partitions and this caused us to hit the 2GB limit. It seems like this is genuinely a bug, since it seems like there are pre-existing abstractions to address this issue, but they just aren't being applied in the particular branch we are hitting. 

I submitted this PR knowing it will likely need a number of iterations, since this is a pretty deep change, and there are very strict requirements (don't break consumers) so I'll go through and make sure to go through and fix the code based on the feedback. 

We tested this patch internally and it did fix the issue we are encountering, so I thought it would be the right time to submit it upstream, but I also understand that the demands of the project at large are of a different character than the demands we have internally.

So yeah, I'll make the changes based on the feedback, if we are hitting this issue, I'm sure there are others, and the code should meet the quality of the rest of the code, and should reuse the existing logic, and not replicate it's own logic for this single case. Talk soon!

sheganinans:
hopefully it looks better this time around @ion-elgreco @ethan-tyler 
Thanks I'll try to look at it tonight or tomorrow at the latest

crates/core/src/operations/merge/mod.rs
```diff
@@ -859,10 +859,12 @@ async fn execute(
         None,
     )?);
 
+    let state_arc: Arc<SessionState> = Arc::new(state.clone());
```
hntd187: There is no need to create a new Arc here, just clone the original.

crates/core/src/operations/merge/mod.rs
```diff
@@ -921,6 +924,7 @@ async fn execute(
         let mut builder = DeltaScanNext::builder()
             .with_eager_snapshot(snapshot.clone())
             .with_log_store(log_store.clone())
+            .with_session(state_arc.clone())
```
hntd187: same as above

hntd187 (changes requested):
2 changes I'd like to see

sheganinans:
@hntd187 done

crates/core/tests/command_merge.rs
```diff
@@ -236,3 +236,227 @@ async fn test_merge_concurrent_with_overlapping_files() {
...
+                    .set("payload", col("source.payload"))
+            })?
+            .await?;
+
+        assert_eq!(metrics.num_target_rows_updated, 1);
+        assert_eq!(metrics.num_target_rows_inserted, 1);
+        assert_eq!(metrics.num_target_rows_copied, 1);
+        assert_eq!(metrics.num_output_rows, 3);
+
+        // Read back and assert the physical Arrow types are Utf8View / BinaryView.
```
ethan-tyler, line 335: I don't think this validates what the PR changed. These assertions go through `table_provider()` on readback, which already gets view types from the read session config. This can pass even if merge cast rows back to Utf8/Binary before writing.

Can we assert the merge output schema before the writer consumes it, or add a reproducer that fails on pre fix code?

crates/core/src/operations/merge/mod.rs
ethan-tyler, line 1248: This looks incomplete for schema evolution. Changed fields fall back to `source_schema.data_type()` here, which is usually still Utf8/Binary. Similarly, new fields at `mod.rs:1157` fall back to the kernel Arrow type. 

Doesn't that reintroduce the issue for `with_merge_schema(true)` when a string/binary column is added or changed?

ethan-tyler (commented):
thanks @sheganinans for working on the follow ups - sorry for the delay in re-review. I left two comments below. I'd like to see a test that asserts the merge write projection preserves view types and not just the read path.

sheganinans:
@ethan-tyler thanks for the very thorough review!! im busy today so i wont be able to give this attention it deserves at this moment. expect me to push a new commit in the coming days, thanks again!

hntd187 (dismissed):
I'm good, thanks for the updates

hntd187 (approved):

Issue #4316, opened 2026-03-24
[Bug]: Wrong log store registry for file:// schema


khalidmammadov:
### What happened?

Currently for local paths `file://` schema suppesed to use default `DefaultLogStoreFactory` but instead ending up using object store created by `MountFactory` from `crates/mount/src/lib.rs`

This issue only related python due to default handlers registration in Python code startup.
https://github.com/khalidmammadov/delta-rs/blob/main/python/src/lib.rs#L3003

### Expected behavior

By the look of it does not seems to create an issue but semantically incorrect. 

### Operating System

None

### Binding

None

### Relevant logs

```shell

```

khalidmammadov:
take

khalidmammadov:
As mentioned in the PR it's intentional

PR #4319, opened 2026-03-26
fix: activate dbfs storage support explicitly with feature

khalidmammadov:
# Description
This is to fix issue where Python API loads object store handler for `dbfs://` and `file://` schemas and overwritting default `file://` hendler. 

Below is sequence of registration of Store Factories. As you can see `file://` schema gets overwritten during [Python initialisation
](https://github.com/delta-io/delta-rs/blob/main/python/src/lib.rs#L3003)

And causes local file operations to use MountFactory instead of the intended default factory.

| Step | Factory Name | Schemes Registered | Issue Status |
|------|-------------|-------------------|--------------|
| **0** | Initial Registry (DefaultObjectStoreFactory) | `memory://`, `file://` | ✅ Default state |
| 1 | AWS (S3ObjectStoreFactory) | `s3://`, `s3a://` | ✅ Ok |
| 2 | Azure (AzureFactory) | `az://`, `adl://`, `azure://`, `abfs://`, `abfss://` | ✅ Ok|
| 3 | GCP (GcpFactory) | `gs://` | ✅ Ok |
| 4 | HDFS (HdfsFactory) | `hdfs://`, `viewfs://` | ✅ Ok |
| **5** | **Mount (MountFactory)** | **`dbfs://`, `file://`** | ⚠️ Oh no **OVERWRITES `file://`!** |
| 6 | LakeFS (LakeFSObjectStoreFactory) | `lakefs://` | ✅ Ok |
| 7 | Unity Catalog (UnityCatalogFactory) | `uc://` | ✅ Ok |

I am suggestting to fix it by letting users to activate mount schema support explicitly and so they can call it when they run code on Databricks clusters. Rest of the time it will not be needed.

# Related Issue(s)
Closes: https://github.com/delta-io/delta-rs/issues/4316

# Documentation
Added into Storage section

### DIFF
--- a/python/deltalake/__init__.py
+++ b/python/deltalake/__init__.py
@@ -5,11 +5,13 @@ from deltalake._internal import (
     TableFeatures,
     Transaction,
     __version__,
-    register_mount_handlers as _register_mount_handlers,
     rust_core_version,
 )
 from deltalake._internal import (
     init_tracing as _init_tracing,
+)
+from deltalake._internal import (
+    register_mount_handlers as _register_mount_handlers,
 )
 from deltalake._internal import (
     shutdown_tracing as _shutdown_tracing,
@@ -81,8 +83,8 @@ def init_tracing(endpoint: Optional[str] = None) -> None:
 def register_mount_handlers() -> None:
     """
     In order to work with DBFS files system (dbfs://) you need
-    to enable it explicitly by calling this function. 
-    This is required because DBFS file system support is not 
+    to enable it explicitly by calling this function.
+    This is required because DBFS file system support is not
     loaded automatically
     """
     _register_mount_handlers()


khalidmammadov:
> Why is this needed? Mount storage handlers are only in use when you provide the storage option key for mount

I've updated description. Hopefully, it make issue clear.

ion-elgreco:
> > Why is this needed? Mount storage handlers are only in use when you provide the storage option key for mount
> 
> I've updated description. Hopefully, it make issue clear.

if AllowUnsafeRename is not set, it will just fallback to the same behavior as DefaultObjectStoreFactory which is to create LocalFileSystem. See the code below.

https://github.com/delta-io/delta-rs/blob/61ea71b77d3322bec3ddb857685a46562925d9fd/crates/mount/src/lib.rs#L48-L75

The proposed change solves an issue that doesn't exist because mount support is only activated if you set the storage_option for allowing unsafe rename

khalidmammadov:
> > > Why is this needed? Mount storage handlers are only in use when you provide the storage option key for mount
> > 
> > 
> > I've updated description. Hopefully, it make issue clear.
> 
> if AllowUnsafeRename is not set, it will just fallback to the same behavior as DefaultObjectStoreFactory which is to create LocalFileSystem. See the code below.
> 
> https://github.com/delta-io/delta-rs/blob/61ea71b77d3322bec3ddb857685a46562925d9fd/crates/mount/src/lib.rs#L48-L75
> 
> The proposed change solves an issue that doesn't exist because mount support is only activated if you set the storage_option for allowing unsafe rename

Agreed, as I mentioned on the issue "By the look of it does not seems to create an issue but semantically incorrect.".

There is no actual impact but the fact that default handler gets overwritten by mount handler looked incorrect to me and was misleading. And by the look of it this unsafe renaming logic and fall back is mount specific but all local storage access is ending up using this path instead of DefaultObjectStoreFactory hence all local Python DeltaTable access.

Another way to solve this would be, for `file://` schema it should just skip registering handler when AllowUnsafeRename is not set and so default handler remains in place?

I am just trying to avoid overwrite and semantics.

But, if this is intentional and by design then I can close PR.

ion-elgreco:
> > > > Why is this needed? Mount storage handlers are only in use when you provide the storage option key for mount
> > > 
> > > 
> > > I've updated description. Hopefully, it make issue clear.
> > 
> > 
> > if AllowUnsafeRename is not set, it will just fallback to the same behavior as DefaultObjectStoreFactory which is to create LocalFileSystem. See the code below.
> > https://github.com/delta-io/delta-rs/blob/61ea71b77d3322bec3ddb857685a46562925d9fd/crates/mount/src/lib.rs#L48-L75
> > 
> > The proposed change solves an issue that doesn't exist because mount support is only activated if you set the storage_option for allowing unsafe rename
> 
> Agreed, as I mentioned on the issue "By the look of it does not seems to create an issue but semantically incorrect.".
> 
> There is no actual impact but the fact that default handler gets overwritten by mount handler looked incorrect to me and was misleading. And by the look of it this unsafe renaming logic and fall back is mount specific but all local storage access is ending up using this path instead of DefaultObjectStoreFactory hence all local Python DeltaTable access.
> 
> Another way to solve this would be, for `file://` schema it should just skip registering handler when AllowUnsafeRename is not set and so default handler remains in place?
> 
> I am just trying to avoid overwrite and semantics.
> 
> But, if this is intentional and by design then I can close PR.

That wouldn't make much sense because the handlers are set during module init. But this is intentional, so lets keep it :)

PR #4322, opened 2026-03-27
ci: refactor all actions to pin shas rather than tags

rtyler:
This change is being made ahead of an enforcement change at the organization level


.github/workflows/dev_pr.yml
ethan-tyler: Did you mean to leave `actions/labeler@v4.0.2` tag pinned? This runs on `pull_request_target` with write access.

.github/workflows/docs.yml
```diff
@@ -45,7 +48,8 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v5
+      # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
 
       - name: Setup Environment
         uses: ./.github/actions/setup-env
```
ethan-tyler, line 55: Should `./.github/actions/setup-env` be pinned to sha instead of tag?

.github/workflows/build.yml
ethan-tyler, line 1: A few direct refs are unpinned here - `actions/setup-python@v4`, `taiki-e/install-action@cargo-llvm-cov`, `codecov/codecov-action@v4`. Not sure if that was intentional or not.

.github/actions/upload-coverage/action.yml
```diff
@@ -0,0 +1,15 @@
+name: Upload to codecov
+inputs:
+  token:
+    required: true
+    type: string
+runs:
+  using: "composite"
+  steps:
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v4
```
ethan-tyler: `codecov/codecov-action@v4` is still tag pinned here

.github/workflows/docs.yml
rtyler: the call is coming from inside the house! That's a local repo reference, so no need there.

Issue #3637, opened 2025-07-29
Upcoming kernel variant support


rtyler:
My CI testing with the delta-kernel-rs main branch highlights that they recently added Variant support. This ticket is to discuss how we might want to expose that up into the Python layer.

```
error[E0004]: non-exhaustive patterns: `deltalake::DataType::Variant(_)` not covered
   --> python/src/schema.rs:35:11
    |
35  |     match schema_type {
    |           ^^^^^^^^^^^ pattern `deltalake::DataType::Variant(_)` not covered
    |
note: `deltalake::DataType` defined here
   --> /home/tyler/source/github/buoyant-data/delta-kernel-rs/kernel/src/schema/mod.rs:674:1
    |
674 | pub enum DataType {
    | ^^^^^^^^^^^^^^^^^
...
691 |     Variant(Box<StructType>),
    |     ------- not covered
    = note: the matched value is of type `deltalake::DataType`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
50  ~         },
51  +         deltalake::DataType::Variant(_) => todo!()
    |

For more information about this error, try `rustc --explain E0004`.
```

In `python/src/schema.rs` we have some Python wrappers like `MapType` and `ArrayType` which provide help[er methods for using these objects in Python.

I suppose we can add a `VariantType` into this file, but I'm not sure if that should be handled in any special way or exposed in any particular way to work well with the pycapsule interface

ion-elgreco:
I guess so yeah but that arrow type needs to be an extension type, I am not sure how arro3 handles that stuff though 🤔

roeap:
After looking through the current state of implementations a bit, we may decide to just not officially support the variant features just yet and maybe make some contributions to arro3 and or arrow-rs.

Generally I think we should also work with the iceberg projects, to see that we handle variant type the same or at least in compatible ways transparent to end users.

mrucker:
Is there any way to write a Variant column from Python currently? I hoped I would be able to manually create a `Schema` object with a "variant" `Field`, but that didn't work. Looking at schema.rs it doesn't look like it is possible right now.

alamb:
A brief update here

The upcoming arrow release 57.0.0 will contain an initial experimental support for Variant in parquet (including an Arrow extension type and kernels to work with arrays containing variant)
- https://github.com/apache/arrow-rs/issues/7835

You can find more information in these issues:
[...]

PR #4325, opened 2026-03-29
feat: add support for variant type

abhiaagarwal:
# Description

Adds support for the variant type. Parquet, Arrow, Datafusion, and kernel are already pretty well set up for this; this was mostly some light plumbing. Steals some fixtures from the delta-spark repository.

# Related Issue(s)

Work towards #3637 

# Documentation

Marked as draft since there are unclear semantics about feature support; adding the `VariantType` feature caused existing tests to fail. Needs a sanity check.


abhiaagarwal:
Failures are due to kernel issue: https://github.com/delta-io/delta-kernel-rs/issues/2267

crates/core/src/kernel/models/actions.rs
rtyler:
```suggestion
pub(crate) fn contains_timestampntz<'a>(mut fields: impl Iterator<Item = &'a StructField>) -> bool {
```

crates/core/src/kernel/models/actions.rs
```diff
@@ -159,6 +159,23 @@ pub fn contains_timestampntz<'a>(mut fields: impl Iterator<Item = &'a StructFiel
     fields.any(|f| _check_type(f.data_type()))
 }
 
+/// checks if table contains variant in any field including nested fields.
+pub fn contains_variant<'a>(mut fields: impl Iterator<Item = &'a StructField>) -> bool {
```
rtyler:
```suggestion
pub(crate) fn contains_variant<'a>(mut fields: impl Iterator<Item = &'a StructField>) -> bool {
```

I validated this locally, but I don't believe there's a compelling reason to make these two functions visible outside of the crate. The `contains_timestampntz` visibility was missed in a previous review it looks like

python/tests/test_table_read.py
```diff
@@ -1226,6 +1226,21 @@ def test_deletion_vectors_without_files_raises():
...
 
+@pytest.mark.pyarrow
+def test_read_variant_fixture():
+    table_path = "../crates/test/tests/data/spark-variant-checkpoint"
+    dt = DeltaTable(table_path)
+
+    schema = dt.schema()
+    assert schema.fields[1].name == "v"
+    assert schema.fields[1].type.type == "variant"
+    assert dt.protocol().reader_features == ["variantType-preview"]
```
rtyler, line 1242:
```suggestion
    dt = DeltaTable(table_path)
    assert dt.protocol().reader_features == ["variantType-preview"]
    
    found = False
    schema = dt.schema()
    for field in schema.fields:
        if field.name == "v":
            found = True
            assert field.type.type == "variant"
    assert found, f"Failed to locate the variant type in the schema: {schema}"
```

With this change the test still fails below on the column content check. I'm not sure what the appropriate value of the first three rows should be though 🙈

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -725,6 +725,10 @@ impl TableProviderBuilder {
             }
         };
 
+        PROTOCOL
+            .can_read_from_protocol(snapshot.snapshot().protocol())
+            .map_err(|err| DataFusionError::Plan(err.to_string()))?;
+
```
abhiaagarwal, line 391: Unsure whether the logic should live here... it was called in the old table provider. Thoughts? @ethan-tyler @roeap

crates/core/src/kernel/transaction/protocol.rs
```diff
@@ -237,11 +274,18 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     let mut reader_features = HashSet::new();
     reader_features.insert(TableFeature::TimestampWithoutTimezone);
     reader_features.insert(TableFeature::DeletionVectors);
-    // reader_features.insert(TableFeature::ColumnMapping);
+    reader_features.insert(TableFeature::VariantType);
+    reader_features.insert(TableFeature::VariantTypePreview);
+    #[cfg(feature = "datafusion")]
+    {
+        reader_features.insert(TableFeature::ColumnMapping);
+    }
```
abhiaagarwal, line 305: also drive by appears that we _do_ indeed support column mapping

hntd187:
@abhiaagarwal can I help with this? Could you give me push access to your branch?

python/tests/test_table_read.py
hntd187: I don't think there is an expectation that the rows come back in the same order, but there are duplicate rows in the read. From reading with spark, I can see there are duplicate rows for 0 and 1, so this test looks correct, it's just the rows aren't in order. So if you order by ID, then it'd be `[0, 0, 1]`

crates/core/src/delta_datafusion/table_provider.rs
ethan-tyler: I think this should live lower. In the old flow the protocol gate was part of scan construction and not the outer builder. The closest equivalent now is next::DeltaScan::new, which is the shared constructor for the new provider. Keeping it in TableProviderBuilder::build means direct DeltaScan::new(...) call sites bypass the check. I think I am going to move it there and keep TableProviderBuilder focused on snapshot/log_store-specific validation.

abhiaagarwal:
> @abhiaagarwal can I help with this? Could you give me push access to your branch?

Hey Steve, sorry for the late response as the past few weeks have been some of the busiest of my life; just granted you access!

rtyler:
I am going to pick this up and try to bring it across the finish line for 0.32.1 :running_man:

ethan-tyler (approved):
LGTM :)

PR #4326, opened 2026-03-29
fix: preserve compact locality by default

### LINKED ISSUE #4310: [Bug]: Insertion Order not maintained when compacting
### What happened?

After the update to 1.5.0 we started seeing a big increase in querying time on unpartitioned delta tables (size around 20GB).
In my understanding, the issue comes from this [PR](https://github.com/delta-io/delta-rs/pull/4257) (which fixed a problem with compaction, thank you for that ❤️).
Compaction adds the new smaller files to older files that did not reach `target_file_size`. This implies that when you query the latest data, they are spread over many files.
Here is an example of the issue, after updating the library.

Before:
Image

After:
Image

### Expected behavior

I would expect compaction to keep the insertion order, as stated also in the compaction stats.

Image

To maintain the insertion order, the oldest file that doesn't reach `target_file_size` should take data from the second oldest until it reaches `target_file_size`, than the second oldest should take data from the third oldest and so on. Would it be a possibility?

Additionally, I would expect for the files to reach the target size of 100MB, but this doesn't happen. The screenshot shows the files generated from two following compaction runs file sizes written during the compaction.

First run:
Image
Image

Second run:
Image

Image

### Operating System

Linux

### Binding

Python

### Bindings Version

1.5.0

### Steps to reproduce

_No response_

### Relevant logs

```shell

```

ethan-tyler:
Closes #4310

Compact was re-packing files by size, which could mix older and newer file ranges within a partition and merge across large file gaps.

The fix is to preserve partition local file order by default. Use `file_views()` order for bin building, and treat skipped large files as locality barriers so compact doesn't freely re-pack across them.

Changes:
- Compact planning keeps partition local order and stops bins at ordinal gaps and skipped large files
- `file_views()` ordering documented and tested
- Writer output ordering stabilized for deterministic results
- Metrics now report `plannerStrategy`, `preservedStableOrder`, `maxBinSpanFiles` (`preserveInsertionOrder` kept as compat alias)
- Removed placeholder `maxInputDisplacement` metric

Tests cover: `file_views()` ordering, no merge across large file gaps, tail locality after recent appends, repeated compaction overlap, metric serialization


abhiaagarwal:
Thanks, this is 100% the proper fix

crates/core/src/operations/write/writer.rs
```diff
@@ -153,11 +168,8 @@ impl WriterConfig {
         num_indexed_cols: DataSkippingNumIndexedCols,
         stats_columns: Option<Vec<String>>,
     ) -> Self {
-        let writer_properties = writer_properties.unwrap_or_else(|| {
-            WriterProperties::builder()
-                .set_compression(Compression::SNAPPY)
-                .build()
-        });
+        let writer_properties =
+            writer_properties.unwrap_or_else(|| default_writer_properties(false));
```
rtyler, line 172: @ethan-tyler I understand the conservative change in _not_ adding the created_by for this case, but I think this is unnecessary and we can always include the created_by in the `default_writer_properties`

crates/core/src/operations/optimize.rs
```diff
@@ -402,12 +476,25 @@ struct OptimizeInput {
...
+const MAX_OPTIMIZE_TARGET_SIZE: u64 = i64::MAX as u64;
+
+fn optimize_target_size_to_i64(target_size: NonZeroU64) -> Result<i64, DeltaTableError> {
+    i64::try_from(target_size.get()).map_err(|_| {
+        DeltaTableError::Generic(format!(
+            "optimize target_size {} exceeds i64::MAX ({MAX_OPTIMIZE_TARGET_SIZE})",
+            target_size.get()
+        ))
+    })
+}
```
rtyler, line 488: This block confuses me, it seems to be a check to prevent the overflow when downcasting from a `NonZeroU64` to an `i64`, which makes me confused why we don't just use a `NonZeroI64` instead and let the compiler find the issue? 

When possible I want rustc to find issues rather than runtime checks

From a different angle, why isn't `u64::MAX` valid here? There's i64/u64 juggling here and it's not clear to me why we cannot just accept wildly large numbers 😆

python/src/lib.rs
```diff
@@ -146,6 +147,22 @@ fn deletion_vector_schema() -> Arc<arrow::datatypes::Schema> {
...
+        PyValueError::new_err(format!(
+            "target_file_size must be between 1 and {MAX_OPTIMIZE_TARGET_SIZE}"
+        ))
+    })?;
+
+    if target_size.get() > MAX_OPTIMIZE_TARGET_SIZE {
+        return Err(PyValueError::new_err(format!(
+            "target_file_size must be between 1 and {MAX_OPTIMIZE_TARGET_SIZE}"
+        )));
+    }
```
rtyler, line 161: I think on the Python side there is probably going to be some runtime checking needed to take the target size into a `NonZeroI64` to correspond to the comment above about runtime vs compile time checks, but I don't think we should need this as a runtime check if the code is using enforced compile time types.

crates/core/src/operations/write/writer.rs
ethan-tyler: Optimize already sets `created_by` in its explicit default `writer_properties` path. Changing `default_writer_properties` would broaden that to all default write paths and while I agree- will do that as a separate cleanup

crates/core/src/operations/optimize.rs
ethan-tyler: The current limit comes from `DeltaOperation::Optimize.target_size`, which we serialize as an `i64` in the operation metadata. The builder takes a `NonZeroU64` because this is conceptually a byte size but we have to guard the conversion before writing that metadata.

`u64::MAX` is not valid because it would overflow the current `i64` metadata field. We could switch the api to `NonZeroI64` but that would push a metadata constraint into the public type for something that is naturally an unsigned byte count.

If we want this path to stay unsigned end to end, the follow up IMO would be changing `DeltaOperation::Optimize.target_size` to `u64`, which would change the serialized operation metadata we emit.

python/src/lib.rs
ethan-tyler: Same underlying reason as above - this check exists because the rust side still stores `DeltaOperation::Optimize.target_size` as an `i64`. On the python side we need runtime validation since this comes in as a python int over FFI, but if that Rust metadata field moves to `u64` in a follow up, this extra check can go away and we’d just keep the normal python side input/non-zero validation.

PR #4336, opened 2026-04-06
feat: add post_commithook_properties to restore

### LINKED ISSUE #4251: feature: Add post_commithook_properties to DeltaTable.restore

When reviewing #4249 I found that restore supports `commit_properties` but not `post_commithook_properties`. Other mutating APIs expose both. 

Add optional `post_commithook_properties` to restore:
- Public API in table.py
- Binding signature in lib.rs
- Typing stub in _internal.pyi
- Wire into maybe_create_commit_properties

_No response_

### Priority

Low - Nice to have

### Additional context

_No response_

### Contribution

byuly:
# Description
- replacing manual commit pipeline (into_prepared_commit_future + manual write_commit_entry) in restore.rs into standard pipeline, all 3 stages run automatically like other endpoints

- exposed `post_commithook_properties` in the Python binding, public API, and type stub 

- added tests verifying the parameter is accepted with post commimt hook

# Related Issue(s)

closes #4251

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -633,6 +633,7 @@ mod tests {
     use delta_kernel::schema::ArrayType;
     use futures::{StreamExt, TryStreamExt};
     use serde_json::json;
+    use std::ops::Range;
     use url::Url;
 
     use super::*;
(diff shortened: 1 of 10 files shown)


byuly:
tests failing when tables conflict commit. going to be adding restore to read_whole_table enum. I see all other mutating endpoitns go through the conflict checker and this makes more sense than setting max retries to 0 (which would be exact behavior before my changes)

crates/core/src/protocol/mod.rs
```diff
@@ -489,6 +489,7 @@ impl DeltaOperation {
         match self {
             // Predicate is none -> Merge operation had to join full source and target
             Self::Merge { predicate, .. } if predicate.is_none() => true,
+            Self::Restore { .. } => true,
```
ethan-tyler: I think marking restore as `read_whole_table()` looks too weak for the retry path. The conflict checker ignores concurrent `data_change=false` rewrites for whole table reads, but restore computes an exact file set diff from the read snapshot.

That means a concurrent OPTIMIZE/compaction can slip through retry and let restore commit against a rewritten layout that previously would have failed on `VersionAlreadyExists`.

I think restore needs either stricter conflict semantics than generic whole table reads, or it should stay on a no-retry path

byuly: thanks for looking it over!

switching to with_max_retries(0) on the CommitBuilder instead, which fails immediately on any version conflict seems better, it's similar semantics as the old manual commit path that's been removed

PR #4337, opened 2026-04-07
fix: address non-determinism in table provider for filters

tobixdev:
# Description

This PR contains a small fix that addresses non-determinism in the DeltaTableProvider.

The issue occurs under the following circumstances:
- A projection is used
- There is a filter condition that contains at least two columns that are not in the projection

Currently, the additional columns which are used in the filter are stored in a `HashSet`. As a result, the iterator over the HashSet may return the additional columns in any order. This introduces non-determinism in creating the logical schema. As the schema affects the string representation of the query plan (@<column_index>), this is a pain for asserting query plans in downstream projects. 

Running the new test without the fix produces one of two possible query plans:

Query Plan 1:
```text
DeltaScan
  DataSourceExec: file_groups={1 group: [[]]}, projection=[v1], file_type=parquet, predicate=v2@1 = CAST(2 AS Int64) AND v3@2 = CAST(3 AS Int64), pruning_predicate=v2_null_count@2 != row_count@3 AND v2_min@0 <= 2 AND 2 <= v2_max@1 AND v3_null_count@6 != row_count@3 AND v3_min@4 <= 3 AND 3 <= v3_max@5, required_guarantees=[]
```

Query Plan 2:
```text
DeltaScan
  DataSourceExec: file_groups={1 group: [[]]}, projection=[v1], file_type=parquet, predicate=v2@2 = CAST(2 AS Int64) AND v3@1 = CAST(3 AS Int64), pruning_predicate=v2_null_count@2 != row_count@3 AND v2_min@0 <= 2 AND 2 <= v2_max@1 AND v3_null_count@6 != row_count@3 AND v3_min@4 <= 3 AND 3 <= v3_max@5, required_guarantees=[]\n
```

The crux lies in the predicate. `predicate=v2@2 [...]` versus `predicate=v2@1`. By sorting the additional columns the indeterminsm should be fixed.

# Related Issue(s)

I think this is a trivial fix and therefore I did not create an issue

# Documentation

- [Expr::column_refs] returns a `HashSet` (https://docs.rs/datafusion/latest/datafusion/logical_expr/enum.Expr.html#method.column_refs)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ resolver = "3"
 
 [workspace.package]
 authors = ["Qingping Hou <dave2008713@gmail.com>", "R Tyler Croy <rtyler@brokenco.de>"]
-rust-version = "1.88"
+rust-version = "1.91"
 keywords = ["deltalake", "delta", "datalake"]
 readme = "README.md"
 edition = "2024"
--- a/crates/core/src/delta_datafusion/find_files.rs
+++ b/crates/core/src/delta_datafusion/find_files.rs
@@ -528,7 +528,7 @@ pub(crate) async fn scan_files_where_matches(
     // and with the source file path exposed as column.
     let table_source = provider_as_source(
         DeltaScanNext::builder()
-            .with_eager_snapshot(snapshot.clone())
+            .with_snapshot(snapshot.snapshot().clone())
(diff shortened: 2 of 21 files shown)


tschwarzinger:
@ethan-tyler Thanks for pointing me towards that! I'll give it a go tomorrow.

PR #4355, opened 2026-04-10
fix: correct CDF delete metrics

### LINKED ISSUE #4353: [Bug]: Number of deleted rows incorrect when table had "enableChageDataFeed" set to true
### What happened?

When deleting items from a deltatable with "enableChangeDataFeed" set to true the output of the table.delete(predicate) call had "num_deleted_rows" as 0 despite there being matching rows that were correctly deleted.  In the example below, I expected the dictionary to have "num_deleted_rows" as 2 rather than the returned 0.

### Expected behavior

I expected the number of deleted rows to match the number of rows matching the predicate. In the example below, 2 rows. 

### Operating System

Linux

### Binding

Python

### Bindings Version

1.5.0

### Steps to reproduce

```python
"""
Tested with deltalake==1.5.0 on Python 3.10.
"""

import tempfile
from pathlib import Path

import pyarrow as pa

# -- setup: write a small Delta table with Change Data Feed enabled --------
tmpdir = Path(tempfile.mkdtemp())
table_path = str(tmpdir / "test_table")

data = pa.table(
    {
        "id": [1, 2, 3, 4, 5],
        "value": ["a", "b", "c", "d", "e"],
    }
)

write_deltalake(
    table_path,
    data,
    mode="overwrite",
    configuration={"delta.enableChangeDataFeed": "true"},
)

dt = DeltaTable(table_path)
rows_before = dt.to_pyarrow_dataset().count_rows()
assert rows_before == 5, f"Expected 5 rows, got {rows_before}"

# -- delete two rows ------------------------------------------------------
predicate = "id IN (2, 4)"
metrics = dt.delete(predicate)

rows_after = dt.to_pyarrow_dataset().count_rows()
actual_deleted = rows_before - rows_after

print(f"\n--- With enableChangeDataFeed ---")
print(f"Rows before delete : {rows_before}")
print(f"Rows after delete  : {rows_after}")
print(f"Actually deleted   : {actual_deleted}")
print(f"Reported deleted   : {metrics.get('num_deleted_rows', 'key missing')}")
print(f"Full metrics       : {metrics}")

# -- compare: same operation WITHOUT Change Data Feed ----------------------
table_path_no_cdf = str(tmpdir / "test_table_no_cdf")

write_deltalake(table_path_no_cdf, data, mode="overwrite")

dt2 = DeltaTable(table_path_no_cdf)
metrics2 = dt2.delete(predicate)
rows_after2 = dt2.to_pyarrow_dataset().count_rows()
actual_deleted2 = 5 - rows_after2

print(f"\n--- Without enableChangeDataFeed ---")
print(f"Actually deleted : {actual_deleted2}")
print(f"Reported deleted : {metrics2.get('num_deleted_rows', 'key missing')}")
```

### Relevant logs

```shell

```

ethan-tyler:
Fixes #4353

Delete metrics were derived from the final write plan output, which includes CDC delete rows when CDF is enabled. That inflated `num_deleted_rows` and `num_copied_rows` even though the delete itself was correct.

The fix is to count source and rescued rows separately, compute delete metrics from `source_rows` - `rescued_rows` instead of the union output.

Tests cover CDF enabled deletes and the all rows deleted case

### DIFF
--- a/crates/core/src/operations/delete.rs
+++ b/crates/core/src/operations/delete.rs
@@ -69,7 +69,7 @@ use crate::delta_datafusion::{
     Expression, add_actions_partition_mem_table, create_session, resolve_session_state,
     scan_files_where_matches, update_datafusion_session,
 };
-use crate::errors::DeltaResult;
+use crate::errors::{DeltaResult, DeltaTableError};
 use crate::kernel::transaction::{CommitBuilder, CommitProperties, PROTOCOL};
 use crate::kernel::{Action, EagerSnapshot, resolve_snapshot};
 use crate::logstore::{LogStore, LogStoreRef};
@@ -81,6 +81,7 @@ use crate::table::config::TablePropertiesExt as _;
 use crate::table::state::DeltaTableState;
 
 const SOURCE_COUNT_ID: &str = "delete_source_count";
+const RESCUED_COUNT_ID: &str = "delete_rescued_count";
 const SOURCE_COUNT_METRIC: &str = "num_source_rows";
 
 /// Delete Records from the Delta Table.
@@ -314,10 +315,10 @@ impl ExtensionPlanner for DeleteMetricExtensionPlanner {
         _session_state: &SessionState,
     ) -> DataFusionResult<Option<Arc<dyn ExecutionPlan>>> {
         if let Some(metric_observer) = node.as_any().downcast_ref::<MetricObserver>()
-            && metric_observer.id.eq(SOURCE_COUNT_ID)
+            && (metric_observer.id.eq(SOURCE_COUNT_ID) || metric_observer.id.eq(RESCUED_COUNT_ID))
         {
             return Ok(Some(MetricObserverExec::try_new(
-                SOURCE_COUNT_ID.into(),
+                metric_observer.id.clone(),
                 physical_inputs,
                 |batch, metrics| {
                     MetricBuilder::new(metrics)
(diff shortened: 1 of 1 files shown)


crates/core/src/operations/delete.rs
```diff
@@ -519,13 +527,20 @@ async fn execute(
...
     .await?;
 
-    if let Some(source_count) = find_metric_node(SOURCE_COUNT_ID, &exec) {
-        let source_count_metrics = source_count.metrics().unwrap();
-        let read_records = get_metric(&source_count_metrics, SOURCE_COUNT_METRIC);
-        let filter_records = exec.metrics().and_then(|m| m.output_rows()).unwrap_or(0);
-        metrics.num_copied_rows = filter_records;
-        metrics.num_deleted_rows = read_records - filter_records;
-    };
+    let err = || DeltaTableError::Generic("Unable to locate expected metric node".into());
```
rtyler, line 530: cute

rtyler (approved):

keen85:
Hi @ethan-tyler ,
It seems `num_deleted_rows` is also wrongly set to 0 in delta-rs 1.5.0 when deleting a **full hive partition**.
Does this fix cover that case too?

ethan-tyler:
> Hi @ethan-tyler , It seems `num_deleted_rows` is also wrongly set to 0 in delta-rs 1.5.0 when deleting a **full hive partition**. Does this fix cover that case too?

Thanks for calling that out @keen85 - This patch fixes the scan/rewrite metrics path for row level deletes with CDF. A partition only delete takes a different fast path, where it removes matching files directly and returns before num_deleted_rows is populated which makes it still 0.

Opened a separate issue for it. #4379

PR #4359, opened 2026-04-12
fix: refine merge duplicate validation semantics

ethan-tyler:
Previous validation treated every repeated match as ambiguous, even when a row didn't satisfy any WHEN MATCHED predicate. That rejected valid merges, for example two source rows match the same target, but only one actually hits the update/delete clause.

The fix is to classify matched rows by semantic participation instead of final operation shape.

- Matched no ops don't count as duplicates
- Unconditional matched deletes are allowed (deterministic)
- Duplicate conditional matched actions still fail
- Target row identity is now file & row ordinal (avoids cross file collisions)

Tests cover: no-op & update, unconditional delete, conditional delete rejection, cdc behavior. rust and python.

### DIFF
--- a/python/tests/test_merge.py
+++ b/python/tests/test_merge.py
@@ -271,10 +271,183 @@ def test_merge_when_matched_update_duplicates(
             predicate="s.deleted = true",
         )
 
-    with pytest.raises(DeltaError):
+    with pytest.raises(DeltaError, match="duplicate relevant WHEN MATCHED"):
         merger.execute()
 
     dt = DeltaTable(tmp_path)
+    result = (
+        QueryBuilder()
+        .register("tbl", dt)
+        .execute("select * from tbl order by id asc")
+        .read_all()
+    )
+
+    assert dt.version() == 0
+    assert result == sample_table
+
+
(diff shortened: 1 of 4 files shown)


python/tests/test_merge.py
```diff
@@ -286,6 +286,179 @@ def test_merge_when_matched_update_duplicates(
     assert result == sample_table
 
 
+@pytest.mark.parametrize("streaming", (True, False))
+def test_merge_when_matched_update_duplicate_with_noop_source_row_passes(
```
ion-elgreco, line 290: Does this have the same semantics as spark when duplicate source rows match the target? Or is this less constrained since we also include the predicate

ion-elgreco (approved):
Looks good, just left one comment for my own understanding

python/tests/test_merge.py
ethan-tyler: Yes, this is intended to match current spark semantics. We still reject the ambiguous cases where multiple source rows satisfy duplicate relevant matched actions- that’s what `test_merge_when_matched_update_duplicates` and `test_merge_when_matched_delete_conditional_duplicates_fail` are covering.

PR #4364, opened 2026-04-12
fix: `DeltaScan` (Next) schema overrides and file column projection

### LINKED ISSUE #4306: [Bug]: DeltaScan ignores DeltaScanConfig::schema unlike old DeltaTableProvider
### What happened?

With `DeltaTableProvider`, one could provide a custom `DeltaScanConfig::schema` to override the schema exposed to DataFusion. This allowed projecting types over the physical Delta kernel types (e.g., mapping a physical Int64 to a logical UInt64). The provider handled coercion during query execution.

PR #4044, migrated scanning to `DeltaScan` and `KernelScanPlan`. The still public `DeltaScanConfig::schema` is ignored during planning and execution.

### Details: (AI assist)
1. `TableProvider::scan()`: KernelScanPlan::try_new() calls config.table_schema(table_config). This method derives the schema directly from the kernel's table_config via try_into_arrow(), bypassing the config.schema property. Consequently, the resulting ExecutionPlan returns strict physical kernel types, resulting in DataFusion schema mismatch errors if the engine expects the configured logical types.
2. `TableProvider::supports_filters_pushdown()`: Filters are evaluated natively against the kernel's physical TableConfiguration. If DataFusion pushes down an AST predicate typed according to the user's logical config.schema, a physical type mismatch occurs in the Parquet scan.
3. `TableProvider::insert_into()`: The internal DeltaDataSink implementation does not accept DeltaScanConfig. It derives its schema strictly from snapshot.read_schema(), preventing type coercions on the write path.

### Expected behavior

- `Update DeltaScanConfig::schema` (or KernelScanPlan initialization) to respect the provided self.schema as an override.
- Re-introduce a RecordBatch casting step in the execution plan if the physical kernel schema differs from the requested `DeltaScanConfig::schema`, aligning behavior with the legacy provider.

.. or remove the field and point to some helper functions to migrate.

### Operating System

None

### Binding

None

### Bindings Version

0.31.1 or main

### Steps to reproduce

```rust
use datafusion::arrow::array::Int64Array;
use datafusion::arrow::datatypes::{DataType, Field, Schema};
use datafusion::arrow::record_batch::RecordBatch;
use datafusion::catalog::TableProvider;
use datafusion::prelude::SessionContext;
use deltalake::delta_datafusion::DeltaScanConfig;
use deltalake::delta_datafusion::DeltaScanNext;
use deltalake::kernel::{DataType as KernelDataType, PrimitiveType, StructField, StructType};
use deltalake::DeltaTable;
use std::sync::Arc;

async fn main() -> Result<(), Box<dyn std::error::Error>> {
    // 1. Create a physical table with an Int64 column in memory
    let physical_schema = StructType::try_new(vec![StructField::new(
        "id".to_string(),
        KernelDataType::Primitive(PrimitiveType::Long),
        true,
    )])?;
    
    let table = DeltaTable::new_in_memory()
        .create()
        .with_columns(physical_schema.fields().cloned())
        .await?;

    // 2. Write a row so the table has a valid state
    let batch = RecordBatch::try_new(
        Arc::new(Schema::new(vec![Field::new("id", DataType::Int64, true)])),
        vec![Arc::new(Int64Array::from(vec![1]))],
    )?;
    
    let table = deltalake::operations::write::WriteBuilder::new(
        table.log_store(), 
        table.state.clone().map(|s| s.snapshot().clone())
    )
    .with_input_batches(vec![batch])
    .await?;

    // 3. Create a DeltaScanConfig defining the schema as UInt64
    let logical_schema = Arc::new(Schema::new(vec![
        Field::new("id", DataType::UInt64, true),
    ]));
    let config = DeltaScanConfig::default().with_schema(logical_schema.clone());

    let provider = DeltaScanNext::new(table.state.unwrap().snapshot().clone(), config)?;
    let advertised_schema = provider.schema();

    println!("Advertised Schema by TableProvider: {:?}", advertised_schema);

    if advertised_schema.fields()[0].data_type() != &DataType::UInt64 {
        println!("The table provider bypassed the logical schema override and advertised {:?}.", advertised_schema.fields()[0].data_type());
        return Err("Schema mismatch".into());
    }

    ctx.register_table("test_table", Arc::new(provider))?;

    let df = ctx.sql("SELECT id FROM test_table").await?;
    df.collect().await?;

    println!("Success");
    
    Ok(())
}
```

### Relevant logs

```shell

```

### LINKED ISSUE #4307: [Bug]: DeltaScan scan includes file column although it's not projected
### What happened?

Configuring a custom `DeltaScanConfig::file_column_name` allows exposing the source file path of each row in an additional column. If to de-select the the file column (i.e. only `SELECT id FROM table`), it is still present in the returned record batches.

#### Details: (AI assist)
1. Bisection identifies that this bug was introduced in **PR #4036 (commit `28679783`)**, which added the `DeltaScanNext` (originally `DeltaScan` in the `next` module) and the `retain_file_ids` logic.
2. In `DeltaScanNext::scan()`, when a `file_column_name` is provided via `DeltaScanConfig`, `retain_file_ids` is set to `true` purely based on the configuration, ignoring the `projection` argument passed by DataFusion to `TableProvider::scan()`.
3. During the execution of a `SELECT` statement (e.g., `SELECT id FROM test_table`), DataFusion passes down a projection that expects only the `id` field.
4. The underlying `DeltaScanExec` fails to strip out the file column if it wasn't requested in the logical plan projection, unconditionally appending it because `retain_file_ids` evaluates to `true`. 

### Expected behavior

- Update the scan execution to properly respect the projection pushed down by DataFusion.
- If a `file_column_name` is configured, it should only be included in the output `RecordBatch` if it is explicitly part of the selected projection.

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

```rust
async fn get_table_and_provider() -> (DeltaTable, Arc<crate::delta_datafusion::DeltaScanNext>) {
    let physical_schema = StructType::try_new(vec![StructField::new(
        "id".to_string(),
        DataType::Primitive(PrimitiveType::Long),
        true,
    )])
    .unwrap();

    let table = DeltaTable::new_in_memory()
        .create()
        .with_columns(physical_schema.fields().cloned())
        .await
        .unwrap();

    let batch = RecordBatch::try_new(
        Arc::new(ArrowSchema::new(vec![ArrowField::new(
            "id",
            ArrowDataType::Int64,
            true,
        )])),
        vec![Arc::new(Int64Array::from(vec![1]))],
    )
    .unwrap();

    let table = crate::operations::write::WriteBuilder::new(
        table.log_store(),
        table.state.clone().map(|s| s.snapshot().clone()),
    )
    .with_input_batches(vec![batch])
    .await
    .unwrap();

    let logical_schema = Arc::new(ArrowSchema::new(vec![ArrowField::new(
        "id",
        ArrowDataType::Timestamp(TimeUnit::Millisecond, None),
        true,
    )]));
    let config = DeltaScanConfig::default().with_schema(logical_schema.clone());

    let provider = crate::delta_datafusion::DeltaScanNext::new(
        table.state.as_ref().unwrap().snapshot().clone(),
        config,
    )
    .unwrap()
    .with_log_store(table.log_store());

    (table, Arc::new(provider))
}

#[tokio::test]
async fn test_delta_scan_config_file_column_projection() {
    let (table, _) = get_table_and_provider().await;

    let logical_schema = Arc::new(ArrowSchema::new(vec![ArrowField::new(
        "id",
        ArrowDataType::Timestamp(TimeUnit::Millisecond, None),
        true,
    )]));
    let config = DeltaScanConfig::default()
        .with_schema(logical_schema.clone())
        .with_file_column_name("_file");

    let provider = crate::delta_datafusion::DeltaScanNext::new(
        table.state.as_ref().unwrap().snapshot().clone(),
        config,
    )
    .unwrap()
    .with_log_store(table.log_store());

    register_table(&ctx, "test_table", &table, Arc::new(provider));

    let df = ctx.sql("SELECT id FROM test_table").await.unwrap();
    let res = df.collect().await;
    assert!(res.is_ok(), "Expected OK, but got: {:?}", res);

    let batches = res.unwrap();
    assert_eq!(
        batches[0].schema().fields().len(),
        1,
        "Expected exactly 1 field after projection, but got: {:?}",
        batches[0]
            .schema()
            .fields()
            .iter()
            .map(|f| f.name())
            .collect::<Vec<_>>()
    );
}
```

### Relevant logs

```shell

```

ahirner:
# Description

- preserve `DeltaScanConfig::schema` through delta scan planning and execution
- __defensively__ avoid filter pushdown when schema overrides introduce logical/physical type mismatches
- fix `file` column projection so selected and excluded projections behave correctly

# Related Issue(s)

- closes #4306
- closes #4307

I can split the PR but think that reviewing is more efficient in one go.


crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -333,13 +339,51 @@ impl ExecutionPlan for DeltaScanExec {
...
+                filter,
+                &self.scan_plan.result_schema,
+                &self.input.schema(),
+            )
+        }) {
+            return Ok(FilterDescription::all_unsupported(
+                &parent_filters,
+                &self.children(),
+            ));
+        }
```
ahirner, line 353: Intuitively, @adriangb's [PhysicalExprAdapter](https://github.com/apache/datafusion/pull/18998) should help to adapt and thus push down those filters. Alas, I don't know enough to legitimately thread it through `deltalake` specifics.

This can and probably should be done in a follow up. Maybe it also requires to discuss what more parts to use from the (parquet) machinery in df.

adriangb: Not sure if I was tagged in purpose but I’m happy to help if there’s a specific question.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -112,7 +115,7 @@ pub(super) async fn execution_plan(
                 vec![file_rows],
                 Arc::new(transforms),
                 Arc::new(dvs),
-                config.retain_file_id().then_some(file_id_field),
+                projected_file_id_column.map(|name| file_id_field(Some(name))),
```
ethan-tyler, line 118: This makes file-id emission projection aware for the metadata path, but `DeltaScanMetaExec` is still deriving its schema / plan properties from `scan_plan.output_schema`. If `projected_file_id_column` is `None` then exec emits batches without the file column while reporting a schema that includes it.

I think this needs the same effective schema handling `DeltaScanExec` has now.

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -502,7 +543,11 @@ impl Stream for DeltaScanStream {
 
 impl RecordBatchStream for DeltaScanStream {
     fn schema(&self) -> SchemaRef {
-        Arc::clone(&self.scan_plan.output_schema)
+        if self.file_id_column.is_some() {
```
ethan-tyler: `schema()` is projection aware now however the empty batch fallback should use the same effective schema. If it still uses the full output schema, empty partitions will produce batches with a different column set than non empty ones.

ethan-tyler (commented):
Thanks @ahirner for working on this! I think there's a schema mismatch in the metadata only path. `DeltaScanExec` is projection aware for file column emission, but `DeltaScanMetaExec` still uses `scan_plan.output_schema` for plan properties while conditionally dropping the file id column from batches. While the plan says the column is there but the batch doesn't have it. 

Once we lock this consistency down, it should be ready to merge. Thanks!

ahirner:
Great catches @ethan-tyler, thanks!

Let me know if there is something more and what to squash, to also sign then.

ethan-tyler:
Thanks @ahirner! I will do a squash merge if you can just ensure all that commits have DCO sign off and we should be good to go.

ethan-tyler (approved):
Thank you!!

PR #4370, opened 2026-04-13
feat: nanosecond timestamps support, gated by a Cargo feature

itamarst:
Add optional support for nanosecond timestamps.


crates/core/src/delta_datafusion/engine/expressions/to_datafusion.rs
```diff
@@ -81,6 +81,10 @@ pub(crate) fn to_datafusion_scalar(scalar: &Scalar) -> DFResult<ScalarValue> {
         Scalar::Timestamp(value) => {
             ScalarValue::TimestampMicrosecond(Some(*value), Some("UTC".into()))
         }
+        #[cfg(feature = "nanosecond-timestamps")]
+        Scalar::TimestampNanos(value) => {
+            ScalarValue::TimestampNanosecond(Some(*value), Some("UTC".into()))
```
hntd187, line 86: Should this be UTC by default? The other timestamp granularities don't specify a TZ at all.

crates/core/src/kernel/models/actions.rs
```diff
@@ -146,17 +146,31 @@ impl MetadataExt for Metadata {
...
 
-/// checks if table contains timestamp_ntz in any field including nested fields.
-pub fn contains_timestampntz<'a>(mut fields: impl Iterator<Item = &'a StructField>) -> bool {
-    fn _check_type(dtype: &DataType) -> bool {
-        match dtype {
-            &DataType::TIMESTAMP_NTZ => true,
-            DataType::Array(inner) => _check_type(inner.element_type()),
-            DataType::Struct(inner) => inner.fields().any(|f| _check_type(f.data_type())),
+/// checks if table contains a datatype in any field including nested fields.
+fn contains_datatype<'a>(
```
hntd187, line 150: We have like 3-4 versions of these check types in schema functions. I kind of wish we could collate these into something more generic. Not saying you need to change anything just thinking outloud.

crates/core/src/kernel/schema/cast/mod.rs
```diff
@@ -224,17 +224,24 @@ pub fn cast_record_batch(
...
         DataType::Date64 => Some(DataType::Date32),
         DataType::Timestamp(TimeUnit::Second, tz)
-        | DataType::Timestamp(TimeUnit::Millisecond, tz)
-        | DataType::Timestamp(TimeUnit::Nanosecond, tz) => {
+        | DataType::Timestamp(TimeUnit::Millisecond, tz) => {
             Some(DataType::Timestamp(TimeUnit::Microsecond, tz.clone()))
         }
+        #[cfg(not(feature = "nanosecond-timestamps"))]
+        DataType::Timestamp(TimeUnit::Nanosecond, tz) => {
+            Some(DataType::Timestamp(TimeUnit::Microsecond, tz.clone()))
```
hntd187, line 239: Typo here? you specify microseconds

crates/core/src/kernel/snapshot/iterators/scan_row.rs
```diff
@@ -331,6 +330,15 @@ pub(crate) fn parse_partitions(
                             }))
                             .with_timezone("UTC"),
                         ) as ArrayRef,
+                        #[cfg(feature = "nanosecond-timestamps")]
+                        PrimitiveType::TimestampNanos => Arc::new(
+                            TimestampNanosecondArray::from_iter(values.iter().map(|v| match v {
+                                Scalar::TimestampNanos(t) => Some(*t),
+                                Scalar::Null(_) => None,
+                                _ => panic!("unexpected scalar type"),
```
hntd187, line 338: No panic please, is this error specifically difficult to propagate up?

crates/core/src/kernel/scalars.rs
```diff
@@ -76,6 +76,11 @@ impl ScalarExt for Scalar {
             Self::Float(f) => f.to_string(),
             Self::Double(d) => d.to_string(),
             Self::Boolean(b) => if *b { "true" } else { "false" }.to_string(),
+            #[cfg(feature = "nanosecond-timestamps")]
+            Self::TimestampNanos(ts) => {
```
hntd187, line 80: I know the code below this does the same thing, but we probably shouldn't just assume UTC here (I wish we could)

hntd187 (dismissed):
This generally LGTM, I assume we won't be merging this as the kernel PR gels, so we'll expect changes

crates/core/src/delta_datafusion/engine/expressions/to_datafusion.rs
itamarst: `Scalar::Timestamp` in microseconds, right above, also has UTC, and that's what this is modeled on: timestamps that have (UTC) timezone.

The case below is specifically without timezone; that will eventually get its own nanosecond primitive type too.

crates/core/src/kernel/schema/cast/mod.rs
itamarst: No, that's deliberate, it's reimplementing the current `main` behavior in the `not(feature = "nanosecond-timestamps")` case.

crates/core/src/kernel/snapshot/iterators/scan_row.rs
itamarst: This is just emulating the current behavior of the other existing datatypes, all of whom panic.

crates/core/src/kernel/scalars.rs
itamarst: My impression from something I read on delta-kernel-rs is that only UTC is supported in actual storage? And if you have another timezone that's just indication for how to display, not of what gets stored? PROTOCOL.md says for the standard microseconds timestamp "When this is stored in a parquet file, its `isAdjustedToUTC` must be set to true."

itamarst:
I found a few more places where tests could be added so will do those today/tomorrow.

itamarst:
Actually not just tests, some missing logic. Will hopefully push an update later today.

itamarst:
OK, updated.

itamarst:
I've pushed some updates to match the latest RFC (always ISO8601 timestamp formatting, also enable timestampNtz table feature if there are nanosecond timestamps).

rtyler (dismissed):
I have done some integration testing and fixing to make this code safe to land in main. I think it's sufficient for experimentation, but I reserve the right to rip this out if the ecosystem moves in a different direction on this data type 😆 

I should also note that our Spark integration tests **explicitly** drop the `timestamp_ns` type from their test fixtures so we should not be claiming any portability in this support

hntd187 (approved):
LGTM

PR #4371, opened 2026-04-14
fix: map Delta `void` type to `pyarrow.null()` in schema conversion

### LINKED ISSUE #1947: python DeltaTable.schema().to_pyarrow() does not support data type `void`

codebygarrysingh:
## Summary

Fixes #1947

Delta tables can contain columns with `void` data type — typically columns created via schema evolution that were never populated. When calling `DeltaTable.schema().to_pyarrow()` on such a table, the conversion currently raises:

```
Exception: Schema error: Invalid data type for Arrow: void
```

### Root Cause

The Arrow type mapping in the Rust kernel does not handle the `void` Delta type. PyArrow's equivalent for a typeless / null column is `pa.null()` (Arrow `NullType`).

### Fix

This PR adds regression tests that:

1. **`test_void_type_maps_to_pyarrow_null`** — verifies a flat schema with a `void` column converts cleanly, with `void → pa.null()`
2. **`test_void_type_in_struct_field`** — verifies `void` nested inside a struct field also converts correctly

The Rust-side fix should map `DataType::Void` (or equivalent) to `arrow_schema::DataType::Null` in the Arrow type conversion logic (likely in `crates/core/src/kernel/schema.rs`).

### Test

```python
schema_json = '{"type":"struct","fields":[{"name":"empty_col","type":"void","nullable":true,"metadata":{}}]}'
schema = Schema.from_json(schema_json)
pa_schema = schema.to_arrow()
assert pa_schema.field("empty_col").type == pa.null()  # was raising before fix
```

### References

- Issue: #1947
- PyArrow null type docs: https://arrow.apache.org/docs/python/generated/pyarrow.null.html

### DIFF
--- a/python/tests/test_schema.py
+++ b/python/tests/test_schema.py
@@ -101,6 +101,94 @@ def test_primitive_delta_types():
 
 
 @pytest.mark.pyarrow
+def test_void_type_maps_to_pyarrow_null():
+    """
+    Regression test for: https://github.com/delta-io/delta-rs/issues/1947
+
+    Delta tables can contain columns with `void` data type (e.g. columns that
+    were created but never populated). When converting a Delta schema to PyArrow
+    via `DeltaTable.schema().to_pyarrow()`, `void` should map to `pyarrow.null()`
+    rather than raising a `Schema error: Invalid data type for Arrow: void`.
+
+    PyArrow's `null` type is the correct semantic equivalent:
+    https://arrow.apache.org/docs/python/generated/pyarrow.null.html
+    """
+    import pyarrow as pa
+
+    # Construct a schema containing a void-typed column via JSON
+    # (void columns appear in Delta tables that have been schema-evolved
+    # or where a column was added but never written to)
+    schema_json = json.dumps({
+        "type": "struct",
+        "fields": [
+            {
+                "name": "id",
+                "type": "long",
+                "nullable": False,
+                "metadata": {},
+            },
+            {
+                "name": "empty_col",
+                "type": "void",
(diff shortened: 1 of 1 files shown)


ethan-tyler (changes requested):
This looks incomplete. If this is wip / tests first that's fine, but moving to draft until the intent is clearer. Let me know what you're going for, thanks!

ion-elgreco:
@ethan-tyler we also have this PR: https://github.com/delta-io/delta-rs/pull/4240 which is halted due to no upstream support in kernel

hntd187:
This PR has been open for several months with no movement, plus it's incomplete. If you want to re-open this submit a new PR.

PR #4375, opened 2026-04-15
perf(core): wire kernel skip_stats for no-predicate file listing

fvaleye:
# Description
Wires `ScanBuilder::with_skip_stats` from delta-kernel-rs into delta-rs so callers opening a table for write-only or metadata-only workflows can skip per-file stats parsing during materialization via `DeltaTableBuilder::with_skip_stats(true)` or `DeltaTableConfig { skip_stats: true, .. }`.

The kernel side https://github.com/delta-io/delta-kernel-rs/pull/1738 added `ScanBuilder::with_skip_stats(bool)`: when enabled, log replay skips `parse_json` and emits null stats for every file.

Python exposure will land in a follow-up PR.

## When it fires
The flag is only wired when the call site has `predicate.is_none()`. Calls that pass a predicate (predicated `delete` / `update`, `find_files`, DataFusion table provider scans, and any other query path) continue to parse stats unconditionally, so stats-based data skipping still works for those.

### Known behavior: predicated queries lose pruning after `skip_stats=true`
A deliberate trade-off worth calling out because it is surprising if undocumented. A `DeltaTable` opened with `skip_stats=true` materializes its file list with a null `stats_parsed` column. A subsequent call to `files(Some(predicate))` on the same snapshot goes through `files_from` with `skip_stats=false`, and stats parsing is re-enabled in principle. But the cached `existing_data` already has null stats, and the kernel has no raw JSON to re-parse. `DataSkippingFilter` receives all-null stats, treats every file as "cannot prune", and no files are pruned.

Query answers stay correct, but stats-based data skipping is disabled for the lifetime of this `DeltaTable` instance. Partition filtering is unaffected. Use `skip_stats=true` only for workflows that will not run predicated queries on the same instance: vacuum, filesystem check, append-only writers, bulk metadata inspection. For query workloads, leave the default.

The constraint is documented on `DeltaTableConfig.skip_stats` and on `DeltaTableBuilder::with_skip_stats`. 

To warn users, I added a runtime `tracing::warn!` that also fires on every predicated call against a `skip_stats=true` snapshot, so users running a predicated query on such a snapshot can see the cause in their logs. I'm open to your suggestions. delta-kernel-rs has no tracing for this case.

## Benchmark
I measured end-to-end on a local Delta table (benchmark kept as a local, not committed): one commit with N `Add` actions (N = 1k / 10k / 50k below), a 20-column schema, `~300 bytes` realistic stats JSON per file, and one parquet checkpoint.

| Files  | skip_stats=false | skip_stats=true | Speedup |
|-------:|-----------------:|----------------:|--------:|
|  1,000 |  4.93 ms         |  4.62 ms        | ~1.07x  |
| 10,000 | 19.72 ms         | 14.47 ms        | ~1.36x  |
| 50,000 | 81.28 ms         | 54.19 ms        | ~1.50x  |

The speedup climbs with wider schemas and larger stats blobs, and with longer `_delta_log` histories where more log files have to be parsed. It also climbs on object storage: the kernel's column projection pushes down to the parquet reader and skips downloading the stats column bytes entirely. On S3, it is a real network-transfer saving on top of the CPU saving measured above.

### DIFF
--- a/crates/core/src/kernel/snapshot/mod.rs
+++ b/crates/core/src/kernel/snapshot/mod.rs
@@ -370,7 +370,14 @@ impl Snapshot {
         engine: Arc<dyn Engine>,
         predicate: Option<PredicateRef>,
     ) -> SendableRBStream {
-        let scan = match self.scan_builder().with_predicate(predicate).build() {
+        self.warn_if_skip_stats_with_predicate(&predicate);
+        let skip_stats = predicate.is_none() && self.config.skip_stats;
+        let scan = match self
+            .scan_builder()
+            .with_predicate(predicate)
+            .with_skip_stats(skip_stats)
+            .build()
+        {
             Ok(scan) => scan,
             Err(err) => return Box::pin(once(ready(Err(err)))),
         };
@@ -379,9 +386,20 @@ impl Snapshot {
             .scan_metadata(engine)
             .map(|d| Ok(rb_from_scan_meta(d?)?));
 
-        match ScanRowOutStream::try_new(self.inner.clone(), stream) {
+        match ScanRowOutStream::try_new(self.inner.clone(), stream, skip_stats) {
             Ok(s) => s.boxed(),
             Err(err) => Box::pin(once(ready(Err(err)))),
+        }
+    }
+
+    fn warn_if_skip_stats_with_predicate(&self, predicate: &Option<PredicateRef>) {
+        if self.config.skip_stats && predicate.is_some() {
+            tracing::warn!(
+                "`DeltaTable` was opened with `skip_stats=true`, but this query has \
+                 a predicate. Every file in the table will be scanned. To avoid \
+                 this, open a separate `DeltaTable` without `skip_stats=true` for \
+                 query workloads."
+            );
         }
     }
 
@@ -393,7 +411,14 @@ impl Snapshot {
         existing_data: Box<T>,
         existing_predicate: Option<PredicateRef>,
     ) -> SendableRBStream {
-        let scan = match self.scan_builder().with_predicate(predicate).build() {
+        self.warn_if_skip_stats_with_predicate(&predicate);
+        let skip_stats = predicate.is_none() && self.config.skip_stats;
+        let scan = match self
+            .scan_builder()
+            .with_predicate(predicate)
+            .with_skip_stats(skip_stats)
+            .build()
+        {
             Ok(scan) => scan,
             Err(err) => return Box::pin(once(ready(Err(err)))),
         };
@@ -402,7 +427,7 @@ impl Snapshot {
             .scan_metadata_from(engine, existing_version, existing_data, existing_predicate)
             .map(|d| Ok(rb_from_scan_meta(d?)?));
 
-        match ScanRowOutStream::try_new(self.inner.clone(), stream) {
+        match ScanRowOutStream::try_new(self.inner.clone(), stream, skip_stats) {
             Ok(s) => s.boxed(),
             Err(err) => Box::pin(once(ready(Err(err)))),
         }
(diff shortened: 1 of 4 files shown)


crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -370,7 +370,14 @@ impl Snapshot {
         engine: Arc<dyn Engine>,
         predicate: Option<PredicateRef>,
     ) -> SendableRBStream {
-        let scan = match self.scan_builder().with_predicate(predicate).build() {
+        self.warn_if_skip_stats_with_predicate(&predicate);
+        let skip_stats = predicate.is_none() && self.config.skip_stats;
+        let scan = match self
+            .scan_builder()
+            .with_predicate(predicate)
+            .with_skip_stats(skip_stats)
```
rtyler, line 402: I like the `warn!` used here. My understanding reading this code is that if `predicate.is_some()` there's not a possible way for the `skip_stats` to be `true`. While the warning is helpful, what I believe will actually happen here is that the predicate will be used, but stats will **not** be skipped whether the user specified it or not.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -379,12 +386,23 @@ impl Snapshot {
...
             Err(err) => Box::pin(once(ready(Err(err)))),
         }
     }
 
+    fn warn_if_skip_stats_with_predicate(&self, predicate: &Option<PredicateRef>) {
+        if self.config.skip_stats && predicate.is_some() {
+            tracing::warn!(
+                "`DeltaTable` was opened with `skip_stats=true`, but this query has \
+                 a predicate. Every file in the table will be scanned. To avoid \
+                 this, open a separate `DeltaTable` without `skip_stats=true` for \
```
rtyler, line 424: My understanding of the behavior here is that the stats will be used, so the warning is that we are _not_ going to skip stats even though the user thinks we will

rtyler (commented):
I'm stoked to see these changes land, I want to make sure I am understanding some of the warn logic before we proceed

crates/core/src/kernel/snapshot/mod.rs
fvaleye: Thanks for the feedback. There is no warning in place in `delta-kernel-rs`, but I wanted to warn `delta-rs` users of the side effects of this configuration.
Yes, when `predicate.is_some()`, the gate makes `skip_stats=false`, so the kernel is asked to parse stats.

The thing the warn captures is one layer down:                                                             `DeltaTableBuilder::with_skip_stats(true).load()` populates `MaterializedFiles.batches` with a null      `stats_parsed` column. Later, `files(Some(predicate))` passes that cached `existing_data` to the kernel via `scan_metadata_from`. The kernel is told to parse stats, but the source data is null: the original `stats` JSON was never read off the checkpoint at materialization time. `DataSkippingFilter` receives all-null stats, treats every file as "cannot prune", and no files are pruned. The predicate runs, and answers stay correct, but pruning is lost.
We can let the documentation play a role in informing users of the consequences, but I prefer to keep this warn in place per call because it could lead to degraded performance on follow-up calls!

crates/core/src/kernel/snapshot/mod.rs
fvaleye: In other words: kernel is asked to parse stats here, exactly as you read, but it is asked to parse a cached existing_data whose stats_parsed column was already null at materialization time. I added more details [here](https://github.com/delta-io/delta-rs/pull/4375#discussion_r3086664237).
I'm happy to change the warning or documentation if it's not clear enough for you.

rtyler (approved):

Issue #4237, opened 2026-02-27
[Feature]: Support for Writing to Delta Tables with Identity Columns in delta-rs


MammadTavakoli:
### Describe the solution you'd like

Hi team,

I am trying to write data to a Delta Lake table that contains an `IDENTITY` column (generated always/by default as identity). However, when using **delta-rs**, the write operation fails.

From what I understand, delta-rs currently does not support writing to tables that contain Identity Columns.

My questions are:

1. Is there any current workaround to write to Delta tables that contain Identity Columns using delta-rs?
2. Is support for Identity Columns planned in the roadmap?
3. If not supported, what is the recommended architectural approach?

   * Should the identity column be removed from the schema before writing?
   * Or should IDs be generated manually in the application layer?

For context:

* Delta table version: (please fill your version)
* delta-rs version: (please fill your version)
* Storage backend: (e.g., S3, local, ADLS, etc.)

Any guidance would be appreciated.

Thanks.

### Priority

None

### Contribution

- [x] I'm willing to submit a pull request for this feature
- [x] I can help with testing this feature
- [x] I can help with documentation for this feature

ion-elgreco:
The delta protocol describes what is required to implement this, I would start with that.

ethan-tyler:
Thanks for the interest @MammadTavakoli! This is a duplicate issue of #1976, however it has not had any updates in while. Unless @hntd187 has reservations, I can assign it to you and we can port this discussion there.

Issue #1976, opened 2023-12-15
Implement Identity Columns table feature


r3stl355:
# Description
Feature definition: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#identity-columns

r3stl355:
take

r3stl355:
I'll use this as a design/planning/progress board. 

Extra points not explicitly mentioned in the Delta protocol definition, from the Databricks documentation (https://learn.microsoft.com/en-us/azure/databricks/delta/generated-columns#--use-identity-columns-in-delta-lake)

- Identity column must be defined at the table creating time, it cannot be added to an existing table:
- Declaring an identity column on a Delta table disables concurrent transactions on the table

Proposed implementation steps:

1. Define a structure to hold an Identity column properties
2. Extend the table metadata to store Identity column related properties
3. Allow adding an Identity column definition to a schema on table create
4. On write, generate the Identity column values if explicit values are not provided/allowed. Fail if generated values overflow
5. [TBD] If explicit Identity values are provided (protocol doesn't have answers to these):
	- Do we need to validate that those values are not in the table already? Assume this is a YES.
	- Do these values need to conform to the current state/definition of the Identity column, i.e. 
		- they are higher than `delta.identity.highWaterMark`? Assume this is a YES.
		- they respect the `delta.identity.step`? Assume it's a NO.
6. Extend the write commit to record `delta.identity.highWaterMark`
7. Detect and fail concurrent writes on tables with an Identity column

ion-elgreco:
On point 5, my understanding is you indeed have to check for duplicates during an append or merge, since the column values need to be unique.

r3stl355:
I've done some research and updated the execution plan above. Delta-io implementation of GENERATED columns is incomplete/flaky at the moment (I was unable to make it run due to some obscure errors - actually SQL and DeltaTable API methods show different errors despite using the same `spark` instance), therefore the behavior/outcome not explicitly defined in the protocol was test on Databricks

ethan-tyler:
@MammadTavakoli — honestly there's no good workaround in delta-rs today. If you need identity columns now, I'd use Spark/Databricks for those tables. Any workaround would have to live upstream of delta-rs in your application logic.

Let me know if you have questions or want guidance on building this out. Thanks!

ethan-tyler:
@ion-elgreco - can you please assign @MammadTavakoli? he's not showing up as an option to assign for me.

[...]

PR #4376, opened 2026-04-15
feat: add identity column support for write operations

naikaayush:
# Description
adds identity column support to the write path.

## What's included
- extract identity column  definitions from field metadata                
- enable in protocol and set writer version to v7
- `with_identity_columns()` injects a `row_number()` window function and computes `base + row_number * step`
- update the hwm (high water mark) in schema metadata after write
- support `allowExplicitInsert` semantics
  - false - user-provided values are rejected (GENERATED ALWAYS);
  - true - values are only generated if the column is absent from the input

## Known limitations
- Overflow in value generation and HWM update uses unchecked arithmetic in some paths (xfail tests document each path)

# Related Issue(s)
- #1976 (closed as supported for reads, this PR adds write-path support) does this need a new issue?
- #4237

# Documentation
- [Delta Protocol: Identity Columns](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#identity-columns)


hntd187:
There are a number of places where what the code does here does not align with what the spec clearly lays out. Was AI utilized to create this PR?

naikaayush:
> There are a number of places where what the code does here does not align with what the spec clearly lays out. Was AI utilized to create this PR?

apologies for the hasty pr, the allowExplicitInsert was picked up from my implementation in my fork (we have a query planner that does this for us, so we didn’t need it in our fork).

i just pushed the fix to be in line with the protocol.

i would consider the step=0 part as a miss from my end, i did put it forward in the description, but im pushing a fix for it now

ai was used for generating tests and the scaffolding structs (definitely validated them)
i mostly followed the pattern from generated_columns to implement this pr

ethan-tyler:
Identity columns are an unsuspecting complex feature and I appreciate the ambition as one of your first contributions to the project.

That said, this pr  has a number of gaps that are direct deviations from the delta protocol/docs and the xfail cases calls out several of those. For a feature at this level of complexity, that leaves too much concern to proceed.

More broadly, even when using ai tooling, we still expect contributors to fully understand and be able to justify the implementation during review. Code reviews create shared understanding and long term project value. When a pr is mostly an "ai produced pass through" without that depth of ownership, it ends up being more work for maintainers without the same upside for the contributor or the codebase.

I think the better path forward would be:
- Keep this pr as draft for now
- Pick up a smaller “good first issue” to get familiar with the codebase and contribution workflow
- Revisit identity columns as a set of smaller, focused PRs that are easier to reason about and review

naikaayush:
Thank you @ethan-tyler for the elaborate response, really appreciate it. I understand where you're coming from.

I also worked on default columns in our fork, it follows this approach very similarly, but is simpler since there are no moving parts. Would that be something simpler as a contribution?

I'm fully on board with the AI usage take.

ion-elgreco:
> Thank you @ethan-tyler for the elaborate response, really appreciate it. I understand where you're coming from.
> 
> I also worked on default columns in our fork, it follows this approach very similarly, but is simpler since there are no moving parts. Would that be something simpler as a contribution?
> 
> I'm fully on board with the AI usage take.

If default columns is a simpler implementation, then lets start with that

hntd187:
This PR hasn't moved in a few months and the author hasn't responded. If they still want to contribute this they can re-open a PR.

PR #4380, opened 2026-04-16
fix: refine next provider scan contract

ethan-tyler:
stacked PR on #4378 - Will rebase this PR once that merges 

Only unique commits: 
- 01985340037014a7dcc23f943ce208d4daf7b642
- 946b95dd6b21382db0becdd8a9871e2ed81d0c95

This is another follow up to next provider scan contract that closes remaining gaps in direct scans and DML file matching

Changes:
- Separate provider schema from scan/output schema in ProjectedScanContract
- Make file id retention explicit for projection and filter queries
- Derive kernel projection names from computed scan schema instead of ad hoc reconstruction
- Validate direct reads up front, register log store for fresh sessions
- Allow column mapping tables through the direct provider path
- Refactor `find_files_scan` / `scan_files_where_matches` around a shared matching files seed
- Normalize matched file ids against table root before mapping back to Add actions
- Pass `log_store` explicitly through delete/update matching scans

### DIFF
--- a/crates/core/src/delta_datafusion/bench_support.rs
+++ b/crates/core/src/delta_datafusion/bench_support.rs
@@ -81,7 +81,7 @@ pub async fn scan_files_where_matches(
     predicate: Expr,
 ) -> DataFusionResult<Option<MatchedFilesScan>> {
     prepare_session(session, &log_store)?;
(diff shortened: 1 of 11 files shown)


ethan-tyler:
Superseded by #4387

PR #4383, opened 2026-04-17
refactor: separate planning from execution

ethan-tyler:
Stacked on #4382. Will rebase once that merges.

Only unique commits: 
- b2b9291d
- b14f4293ae9ed67f172c9257b929bb08e7c0acb9

This PR isolates the planning seam before we swap overwrite / replaceWhere onto the matched file scan contract.

Adds write/plan.rs and splits write into boundaries:
- `prepare_write()`: source normalization, generated columns, schema evolution, exact predicate resolution, sink options
- `plan_overwrite_rewrite()`: overwrite rewrite planning
- `write_execution_plan_v2()`: sink execution (unchanged)

Introduces typed planner state with `PreparedWrite`, `SchemaDelta`, `WriteExecOptions`, `OverwritePlan`. Schema/protocol actions assembled before sink execution instead of fused in write/mod.rs.

In the next PR are replacing the legacy rewrite substrate with a local rewrite planning change instead of another broad semantic rewrite.

Tests cover the new planner seam and replaceWhere no match validation.

PR #4385, opened 2026-04-17
refactor: migrate overwrite rewrites to matched file scans

ethan-tyler:
Stacked on #4383  Will rebase once that merges

Only unique commit: 
- 1f89215ad98176b514bf2a3043cff96e31466683
- 3fe469e8816fb1f5453b5bf2abd4e5b75af7db49
- f49dd4dfe44c1dd28122dbe2ce18a25bb049e801

Moves `overwrite` / `replaceWhere` onto `scan_files_where_matches()`. Write no longer depends on `find_files()` -> `Add list` -> `DeltaTableProvider::with_files()`

Replaces `prepare_predicate_actions` / `execute_non_empty_expr` with typed rewrite planning in write/plan.rs. Matched files stay typed until commit assembly instead of lowering to raw actions during planning

Rescue rewrites built from the matched file scan and unioned with validated inserts. Partition only rewrites skip the rescue scan. CDC handling stays within the rewrite/sink boundary. Insert only validation threaded through via an internal write marker column dropped before final output

`replaceWhere` validation is now explicit across all outcomes with no matched files, partition only, and full rescue rewrites

### DIFF
--- a/crates/core/src/delta_datafusion/bench_support.rs
+++ b/crates/core/src/delta_datafusion/bench_support.rs
@@ -81,7 +81,7 @@ pub async fn scan_files_where_matches(
     predicate: Expr,
 ) -> DataFusionResult<Option<MatchedFilesScan>> {
     prepare_session(session, &log_store)?;
-    super::scan_files_where_matches(session, snapshot, predicate)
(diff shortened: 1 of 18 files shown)


ethan-tyler:
Superseded by #4387

Issue #4015, opened 2025-12-27
Limit stats parsing to stats required to resolve a query


roeap:
### Is your feature request related to a problem?

Currently stats parsing is a source of some redundant work as delta kernel internally parses the stats field as well as us since the parsed stats are (not yet) exposed for engines to consume as well. This can be a significant performance overhead we want to avoid.

### Describe the solution you'd like

Make stats parsing more selective to only the fields (if any) required to execute a given query.

### Priority

High - Important for my use case

### Contribution

- [x] I'm willing to submit a pull request for this feature
- [ ] I can help with testing this feature
- [ ] I can help with documentation for this feature

aditanase:
Quick comment on this as it's a huge blocker for us for migrating to more recent versions of delta and delta-kernel (how the stats are loaded/parsed and used degraded considerably in the last 6-9 months):
- some writers produce both JSON stats and parsed stats
- for some of our tables the JSON is HUGE (sometimes 75% of volume of the parquet checkpoints)
- the real gain is from not loading it at all, especially if the stats are already parsed (this used to existing in older versions of delta-rs) - so I would advocate for exposing the parsed stats if they are available
- there is special consideration for the `num_rows` statistic which is really useful when implementing push down limit

ethan-tyler:
@aditanase thank you for the feedback and really appreciate hearing from what matters to delta-rs users! Agreed on the direction.

I am actively working on this and wanted to share a couple of my thoughts. I think the important thing I've found is that parsing less json is only part of the win. For checkpoints that have both `stats` and `stats_parsed`, query paths should prefer `stats_parsed` and skip materializing raw `stats` json unless something explicitly needs exact add action data.

The feature shape I'm targeting:
- Query/planning paths use a `FileStatsMaterialization` policy
- Prefer checkpoint `stats_parsed` when available
- Raw json `stats` omitted from query batches by default
- json falls back when parsed stats aren't there
- Compat/add action paths can still preserve or lazily recover exact raw stats
- `numRecords` stays as the cheap baseline for row count planning and future limit pushdown
- `skip_stats` still avoids min/max/null count parsing

My goal here isn't to just reduce json parse cpu. For checkpoints with both `stats` and `stats_parsed`, the hot path shouldn't carry the raw `stats` column at all, while compat paths that need exact add action stats keep working.

PR #4403, opened 2026-04-23
refactor: introduce selective stats projection primitives

ethan-tyler:
First pr in the  _Selective Stats Stack_ ™.  This adds the kernel level primitives for selective stats parsing without wiring them through snapshot materialization or query execution yet.

Changes:
- `StatsProjection` abstraction for describing which stats a scan needs
- Parsed scan row schema retains raw stats while appending `stats_parsed`
- Scan row parsing and `LogicalFileView` helpers work with selectively projected stats
- Error/plumbing for distinguishing raw stats backed compared to projection backed views

I am landing this first so the follow up PRs can focus on snapshot wiring and query path behavior without mixing in this foundational schema/parsing refactors.

Tests: unit coverage in the touched kernel modules.

Part of #4015 - **but does not close it yet!** & builds on @fvaleye's work (would love your thoughts)


ethan-tyler:
20 failed test wooohoo that's a record ... moving to draft for now

crates/core/src/kernel/arrow/engine_ext.rs
```diff
@@ -23,7 +23,7 @@ use delta_kernel::transforms::SchemaTransform;
 use delta_kernel::{DeltaResult, ExpressionEvaluator};
 
 use crate::errors::{DeltaResult as DeltaResultLocal, DeltaTableError};
-use crate::kernel::SCAN_ROW_ARROW_SCHEMA;
+use crate::kernel::{SCAN_ROW_ARROW_SCHEMA, StatsProjection};
```
hntd187, line 27: Looks like you forgot to check a file in?

ethan-tyler: ya ya - This is what I get for trying to split my work up into _smaller_ PRs and it got messy because my split was made across semantic pieces and not across independently compiling boundaries. I am getting this cleaned up.

rtyler (approved):

PR #4406, opened 2026-04-25
fix: release GIL for DeltaTable update calls

### LINKED ISSUE #4402: [Bug]: update_incremental() and transaction_version() hold the GIL for the entire S3 round-trip
### What happened?

Several DeltaTable methods hold the Python GIL for the full duration of their S3 I/O, blocking every other Python thread (asyncio event loops, thread-pool workers, etc.) until the round-trip completes. PyO3 bindings appear to call Tokio `block_on()` without `py.allow_threads()` for these paths.

Measured against an S3-backed Delta table with `deltalake==1.5.0`:

| Call | Wall time | GIL release ratio* |
|---|---:|---:|
| `DeltaTable(uri)` | ~290 ms | 87% (released) |
| `pl.scan_delta(dt).collect()` (full scan) | varies | 97% (released) |
| `dt.update_incremental()` | ~290 ms | **<2% (held)** |
| `dt.transaction_version(app_id)` | ~115 ms | **<5% (held)** |

*Ratio = Python counter-thread tick rate during the call ÷ baseline tick rate. ~100% = GIL released, ~0% = GIL held.

Thread-scaling confirms the serialisation: running `update_incremental()` from N parallel threads achieves only ~1.8x throughput at N=4 vs the ideal 4x.

The `DeltaTable` constructor and the polars read path already release the GIL — only the table-update and app-transaction lookup paths still hold it. In practice this forces users who poll many tables concurrently to spawn subprocesses just to keep their event loops responsive.

Reproduction script (short Python probe with a background counter thread) available on request.

### Expected behavior

update_incremental() and transaction_version() should release the GIL around their S3 I/O — i.e. wrap the Tokio block_on()/blocking work in py.allow_threads(), matching what DeltaTable() already does.

### Operating System

Linux

### Binding

Python

### Bindings Version

1.5.0

### Steps to reproduce

Minimal probe — counter thread increments while the GIL is available; ratio of during-call ticks to baseline ticks reveals whether the call holds it.

```python
import threading
import time
import deltalake

URI = "s3://your-bucket/some-delta-table"  # any S3-backed Delta table

counter = {"n": 0}
stop = threading.Event()

def tick():
    while not stop.is_set():
        counter["n"] += 1

threading.Thread(target=tick, daemon=True).start()
time.sleep(0.5)

def measure(label, fn):
    fn(); fn()  # warm
    base = counter["n"]; time.sleep(0.05)
    base_rate = (counter["n"] - base) / 0.05
    before = counter["n"]; t0 = time.perf_counter()
    for _ in range(10):
        fn()
    elapsed = time.perf_counter() - t0
    rate = (counter["n"] - before) / elapsed
    print(f"{label}: {elapsed*1000/10:6.1f}ms/call  ratio {rate/base_rate:6.1%}")

measure("DeltaTable(uri)         ", lambda: deltalake.DeltaTable(URI))
dt = deltalake.DeltaTable(URI)
measure("dt.update_incremental() ", lambda: dt.update_incremental())
measure("dt.transaction_version()", lambda: dt.transaction_version("any-app-id"))

stop.set()
```

A ratio near 100% means the GIL was released for the duration of the call (other Python threads ran freely); a ratio near 0% means the GIL was held end-to-end.

### Relevant logs

```shell
deltalake 1.5.0
target: s3://<redacted>  (S3)

DeltaTable(uri)              :  279.0ms/call  ratio 90.45%   -> released
dt.update_incremental()      :  322.3ms/call  ratio  1.29%   -> HELD
dt.transaction_version(...)  :  110.8ms/call  ratio  4.22%   -> HELD

Thread-scaling for dt.update_incremental() (10 iterations × N threads):
   1 thread :  5.95s   1.75 calls/s   1.00x
   2 threads:  8.80s   2.27 calls/s   1.35x   (would be ~2.0x if GIL released)
   4 threads: 13.45s   2.97 calls/s   1.77x   (would be ~4.0x if GIL released)
```

ethan-tyler:
Fixes #4402

`update_incremental()` and `transaction_version()` held the GIL while waiting on rust async work, blocking unrelated Python threads during s3 I/o.

The fix is to wrap both paths in `py.detach()`, matching the pattern used by other binding methods. Add s3/localstackregression coverage for GIL release and concurrent table access.

Scoped to the two api's reported in #4402. Other `rt().block_on()` call sites can/should be looked at separately

### DIFF
--- a/.github/scripts/docs_changelog.sh
+++ b/.github/scripts/docs_changelog.sh
@@ -9,7 +9,7 @@ output="$(
   /^[[:space:]]*##[[:space:]]*\[rust-v/ {
     s|^[[:space:]]*##[[:space:]]*\[(rust-v)([0-9]+)\.([0-9]+)\.([0-9]+)\]\([^)]*\)[[:space:]]*\(([0-9-]+)\)|- [\1\2.\3.\4]('"$BASE_URL"'#rust-v\2\3\4-\5)|p
   }
-  ' "$FILE" | head -n 5
+  ' "$FILE" | sed -n '1,5p'
 )"
 
 # Fail CI if nothing was produced
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -1298,18 +1298,25 @@ impl RawDeltaTable {
             .collect())
     }
 
-    pub fn update_incremental(&self) -> PyResult<()> {
-        #[allow(clippy::await_holding_lock)]
-        #[allow(deprecated)]
-        Ok(rt()
-            .block_on(async {
+    /// Incrementally update the table snapshot to the latest committed version.
+    ///
+    /// Releases the Python GIL while the async table update runs.
+    pub fn update_incremental(&self, py: Python) -> PyResult<()> {
+        py.detach(|| {
+            // `DeltaTable::update_incremental` mutates the cached table in place, and the
+            // guard must live until the async update finishes.
+            #[allow(clippy::await_holding_lock)]
+            #[allow(deprecated)]
+            rt().block_on(async {
                 let mut table = self
                     ._table
                     .lock()
                     .map_err(|e| DeltaTableError::Generic(e.to_string()))?;
                 (*table).update_incremental(None).await
             })
-            .map_err(PythonError::from)?)
+            .map_err(PythonError::from)
+            .map_err(PyErr::from)
+        })
     }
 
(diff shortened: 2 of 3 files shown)


ion-elgreco (approved):
I don't think we need those tests tbh :)

ethan-tyler:
> I don't think we need those tests tbh :)

That's fair :)

PR #4408, opened 2026-04-26
fix: exclude partition columns from add stats schema

### LINKED ISSUE #4398: [Bug]: Clarification (or regression) `dataSkippingNumIndexedCols`
### What happened?

WIth the upgrade to 1.5.1, we see a different convention for `get_add_actions()` that affects the meaning of `dataSkippingNumIndexedCols` when partitioning is used.

With:
```python
import polars as pl

root = "./tmp/delta"

df = pl.DataFrame({"p": [10, 10, 20, 20], "a": [1, 2, 3, 4]})

df.write_delta(
    root,
    delta_write_options={
        "partition_by": "p",
        "configuration": {"delta.dataSkippingNumIndexedCols": "1"},
    },
)

dt = DeltaTable(root)
print(pl.DataFrame(dt.get_add_actions()))
```

Before (1.5.0):
```
shape: (2, 8)
┌─────────────────────────────────┬────────────┬───────────────────┬─────────────┬────────────┬───────────┬───────────┬───────────┐
│ path                            ┆ size_bytes ┆ modification_time ┆ num_records ┆ null_count ┆ min       ┆ max       ┆ partition │
│ ---                             ┆ ---        ┆ ---               ┆ ---         ┆ ---        ┆ ---       ┆ ---       ┆ ---       │
│ str                             ┆ i64        ┆ i64               ┆ i64         ┆ struct[1]  ┆ struct[1] ┆ struct[1] ┆ struct[1] │
╞═════════════════════════════════╪════════════╪═══════════════════╪═════════════╪════════════╪═══════════╪═══════════╪═══════════╡
│ p=10/part-00000-cce43df7-72cb-… ┆ 494        ┆ 1776845457687     ┆ 2           ┆ {0}        ┆ {1}       ┆ {2}       ┆ {10}      │
│ p=20/part-00000-753c9624-8fe4-… ┆ 494        ┆ 1776845457688     ┆ 2           ┆ {0}        ┆ {3}       ┆ {4}       ┆ {20}      │
└─────────────────────────────────┴────────────┴───────────────────┴─────────────┴────────────┴───────────┴───────────┴───────────┘
```

After (1.5.1):
```
shape: (2, 8)
┌─────────────────────────────────┬────────────┬───────────────────┬─────────────┬────────────┬───────────┬───────────┬───────────┐
│ path                            ┆ size_bytes ┆ modification_time ┆ num_records ┆ null_count ┆ min       ┆ max       ┆ partition │
│ ---                             ┆ ---        ┆ ---               ┆ ---         ┆ ---        ┆ ---       ┆ ---       ┆ ---       │
│ str                             ┆ i64        ┆ i64               ┆ i64         ┆ struct[1]  ┆ struct[1] ┆ struct[1] ┆ struct[1] │
╞═════════════════════════════════╪════════════╪═══════════════════╪═════════════╪════════════╪═══════════╪═══════════╪═══════════╡
│ p=20/part-00000-da7635c1-174b-… ┆ 494        ┆ 1776845476390     ┆ 2           ┆ {null}     ┆ {null}    ┆ {null}    ┆ {20}      │
│ p=10/part-00000-f3ee074c-a8c5-… ┆ 494        ┆ 1776845476390     ┆ 2           ┆ {null}     ┆ {null}    ┆ {null}    ┆ {10}      │
└─────────────────────────────────┴────────────┴───────────────────┴─────────────┴────────────┴───────────┴───────────┴───────────┘
```

If we drop `dataSkippingNumIndexedCols`, it looks like the partition column was excludd prior and is included now.

### Expected behavior

We noticed this in our CI. It looks like the scope of `get_add_actions` changed: it excluded the partition columns prior and includes them now. As an implication, `dataSkippingNumIndexedCols` has a different meaning.

Is this intentional or a bug? Apologies if I missed a PR update, and I could not find anythin rightaway in the docs.

### Operating System

Linux

### Binding

Python

### Bindings Version

_No response_

### Steps to reproduce

See above.

### Relevant logs

```shell

```

ethan-tyler:
Fixes #4398

`buoyant_kernel` changed parsed stats schema to use `physical_schema()`, which includes partition columns on partitioned tables. However, writers don't store partition column stats in `Add.stats`, those values live in `Add.partitionValues`. This broke stats parsing for partitioned tables in python-v1.5.1.

The fix here is to restore the invariant that partition columns are excluded from the stats schema. Share the corrected derivation between the legacy add actions path and the datafusion next replay path. Preserve the existing stats schema naming behavior while excluding partition columns.

This pr is scoped to the shared read path and does not change `dataset_partitions` pruning in `python/src/lib.rs`. This is related but separate.

As always you know I bring the tests, I've included rust and python regression for partitioned tables with `dataSkippingNumIndexedCols` and `dataSkippingStatsColumns`

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
@@ -30,6 +30,7 @@ use delta_kernel::{
         state::{DvInfo, ScanFile},
     },
     schema::{DataType, Schema, StructField},
+    table_configuration::TableConfiguration,
 };
 use futures::Stream;
 use itertools::Itertools;
@@ -41,7 +42,8 @@ use crate::{
     delta_datafusion::{DeltaScanConfig, engine::to_datafusion_scalar},
     kernel::{
         LogicalFileView, ReceiverStreamBuilder, Scan, StructDataExt,
-        arrow::engine_ext::stats_schema, parse_stats_column_with_schema,
+        arrow::engine_ext::{stats_inputs, stats_schema},
+        parse_stats_column_with_schema,
     },
 };
 
@@ -73,10 +75,10 @@ fn extract_predicate_columns(scan: &KernelScan) -> Option<HashSet<String>> {
 /// Create a stats schema containing only columns needed for predicate evaluation
 ///
 /// Returns:
-/// - If no predicate: Schema with just `numRecords`
-/// - If predicate exists: Schema with `numRecords` + stats for referenced columns only
+/// No predicate returns a schema with just `numRecords`.
+/// A predicate returns `numRecords` plus stats for referenced columns.
 fn create_minimal_stats_schema(
-    scan: &KernelScan,
+    table_configuration: &TableConfiguration,
     predicate_columns: Option<&HashSet<String>>,
 ) -> DeltaResult<Arc<Schema>> {
     let minimal_schema = || {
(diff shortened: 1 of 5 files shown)


ethan-tyler:
I'd like to land #4403 first and rebase this one

rtyler (approved):

PR #4410, opened 2026-04-27
fix(catalog-unity): skip non-Unity keys in try_with_options

### LINKED ISSUE #4225: [Bug]: UnityCatalogFactory::parse_url_opts only allows unity-specific options
### What happened?

PR https://github.com/delta-io/delta-rs/pull/3769 added the ability to pass storage options to `UnityCatalogFactory::execute_uc_future` in response to  https://github.com/delta-io/delta-rs/issues/3757. However, the way this function is used in `UnityCatalogFactory::parse_url_opts` means that any non-unity options, such as for example `aws_region` or `timeout` that should be passed to `DeltaTableBuilder` are rejected and cannot be used to configure the catalog.

In more details:
- Create a `DeltaTableBuilder` with a unity catalog URL (`uc://`). Use `DeltaTableBuilder::with_storage_options` to set a bunch of unity-specific config options (`databricks_host`, `unity_client_id`, `unity_client_secret`, etc.). I also set `object_store` options, e.g.,  `aws_region` and `timeout`.
- Call `builder.load()`, which internally creates a `UnityCatalogFactory` and calls [parse_url_opts](https://github.com/delta-io/delta-rs/blob/fef97b8334382517b9cc17c120b01f7fdc0be79e/crates/catalog-unity/src/lib.rs#L870) on it. This in turn invokes `UnityCatalogBuilder::get_uc_location_and_token` [here](https://github.com/delta-io/delta-rs/blob/fef97b8334382517b9cc17c120b01f7fdc0be79e/crates/catalog-unity/src/lib.rs#L876C13-L876C59).
- `UnityCatalogBuilder::get_uc_location_and_token` fails if it encounters any storage options that are not unity catalog options.

The practical implication of this is that I am not able to open any tables that require non-default storage backend options, `aws_regions` being the most common one.

### Expected behavior

I suppose UnityCatalogBuilder::get_uc_location_and_token should simply ignore any options it doesn't understand, so that they can be forwarded to the new DeltaTableBuilder created here https://github.com/delta-io/delta-rs/blob/fef97b8334382517b9cc17c120b01f7fdc0be79e/crates/catalog-unity/src/lib.rs#L885

### Operating System

macOS

### Binding

Rust

### Bindings Version

0.30.2

### Steps to reproduce

_No response_

### Relevant logs

```shell

```

1fanwang:
## Problem

`UnityCatalogFactory::parse_url_opts` forwards the full `StorageConfig.raw` map into `UnityCatalogBuilder::get_uc_location_and_token`, which calls `try_with_options`. Since `try_with_options` rejected any key that didn't parse as a `UnityCatalogConfigKey`, mixing `aws_region` / `timeout` / etc. with unity options caused the whole `parse_url_opts` call to error — even though those non-unity keys are destined for the `DeltaTableBuilder` that `parse_url_opts` builds right after, via `with_storage_options`.

## Fix

Skip non-Unity keys in `try_with_options` instead of failing. `try_with_option` (singular) stays strict so single-key callers still surface typos. Matches `from_env`, which already filters by `UNITY_`/`DATABRICKS_` prefix rather than rejecting unknown keys.

## Tests

- Added `test_try_with_options_skips_unknown_keys` covering the mixed-map case from #4225.
- Added `test_try_with_option_rejects_unknown_key` documenting the singular form's strict behavior.
- Existing 13 unity unit tests pass unchanged (`cargo test -p deltalake-catalog-unity --lib`: 15 passed).

Fixes #4225

### DIFF
--- a/crates/catalog-unity/src/lib.rs
+++ b/crates/catalog-unity/src/lib.rs
@@ -437,12 +437,20 @@ impl UnityCatalogBuilder {
         Ok(self)
     }
 
-    /// Hydrate builder from key value pairs
+    /// Hydrate builder from key value pairs.
+    ///
+    /// Keys that are not recognised as Unity Catalog options are skipped, so
+    /// callers can pass a mixed config map that also contains object store
+    /// options (e.g. `aws_region`, `timeout`) destined for downstream
+    /// consumers. Use [`Self::try_with_option`] to fail on unknown keys.
     pub fn try_with_options<I: IntoIterator<Item = (impl AsRef<str>, impl Into<String>)>>(
         mut self,
         options: I,
     ) -> DataCatalogResult<Self> {
         for (key, value) in options {
+            if UnityCatalogConfigKey::from_str(key.as_ref()).is_err() {
+                continue;
+            }
             self = self.try_with_option(key, value)?;
         }
         Ok(self)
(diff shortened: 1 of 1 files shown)


ethan-tyler:
Thanks @1fanwang! - this looks good, can you sign off with DCO and we should be good to merge.

1fanwang:
Done — pushed 7d11460 with DCO sign-off.

ethan-tyler (approved):

PR #4411, opened 2026-04-27
feat: support writes to name-mapped column mapping tables

fksegundo:
## what changed

  Refs #930.

  This PR adds support for writing to existing Delta tables that use
  `delta.columnMapping.mode = name`.

  The write path now resolves logical columns to their column mapping physical
  names before writing Parquet files. Partition values and data skipping stats are
  also written with physical keys, while table scans and DataFusion queries keep
  using logical column names.

  `SchemaMode::Merge` is supported for top-level fields on existing name-mapped
  tables. Existing column mapping metadata is preserved, and newly added
  top-level fields receive column mapping IDs and generated physical names.

  ## supported behavior

  - appends to existing name-mapped tables;
  - top-level schema evolution with `SchemaMode::Merge`;
  - logical-name reads and DataFusion filters after writes;
  - physical Parquet field names for mapped data columns;
  - physical partition values and physical data skipping stats;
  - `replaceWhere`, partitioned overwrite, update, delete, and non-evolving
    merge rewrites;
  - randomized partition prefixes by default for partitioned column-mapped
    writes;
  - explicit opt-in for Hive-style partition path preservation with
    `with_preserve_column_mapping_hive_style_partitions(true)`;
  - `ADD FEATURE columnMapping` continues to work as a protocol-only feature
    addition. It does not activate column mapping unless table metadata also sets
    `delta.columnMapping.mode`.

  ## guardrails

  The following paths are intentionally rejected instead of being partially
  enabled:

  - `columnMapping.mode = id`;
  - nested column-mapped writes;
  - creating column-mapped tables through `WriteBuilder`;
  - `ADD COLUMN` on column-mapped tables outside write schema merge;
  - enabling or changing column mapping through
    `SET TBLPROPERTIES delta.columnMapping.mode`;
  - CDC writes on column-mapped tables;
  - `OPTIMIZE` rewrites on column-mapped tables;
  - `MERGE` with schema evolution on column-mapped tables.

  These paths return a consistent unsupported column mapping write error without
  adding a new public `DeltaTableError` variant.

  ## validation

  Passed locally:

  ```bash
  cargo check -p deltalake-core --features datafusion
  cargo test -p deltalake-core --features datafusion --lib column_mapping -- --nocapture
  cargo test -p deltalake-core --features datafusion --lib add_feature -- --nocapture
  cargo test -p deltalake-core --features datafusion --lib test_add_column_rejects_column_mapping_table -- --nocapture
  cargo test -p deltalake-core --features datafusion --lib test_set_tbl_properties_rejects_column_mapping_mode -- --nocapture
  cargo test -p deltalake-core --features datafusion --lib test_column_mapping_name_table_rejects_invalid_random_prefix_length -- --nocapture
  cargo fmt --check
```
  Python validation after rebuilding the local binding with CPython 3.10:
```bash
  uv run --no-sync pytest tests/test_alter.py -q -k "add_feature_variations or add_features" --maxfail=1
  uv run --no-sync pytest tests/test_lakefs.py -q -m "lakefs and integration" -k add_features --maxfail=1
```
  Results:
```
  tests/test_alter.py: 17 passed
  tests/test_lakefs.py::test_add_features: 1 passed
```
  I also ran the standalone lab against the local checkout. It covers regular
  partitioned tables, freshly name-mapped partitioned tables, evolved name-mapped
  tables with Hive-style partition path preservation enabled, and report-style
  filtered reads.

  Latest lab result:
```
  regular: rows=3000
  column_mapping: rows=3000
  evolved_column_mapping: rows=6000
  validation passed
```
  Lab repository: [https://github.com/fksegundo/delta-column-mapping-lab](https://github.com/fksegundo/delta-column-mapping-lab)

  ## current limitations

  This PR is limited to top-level columnMapping.mode = name writes. Nested
  physical-name rewrites, columnMapping.mode = id, CDC output files,
  OPTIMIZE, and schema-evolving MERGE remain follow-up work.

  ## disclosure

  I used Codex for code navigation, patch drafting, and test scaffolding. I
  reviewed the changes and validated them with the Rust tests, Python tests, and
  standalone lab run listed above.

### DIFF
--- a/crates/core/src/errors.rs
+++ b/crates/core/src/errors.rs
@@ -242,3 +242,14 @@ impl DeltaTableError {
         Self::Generic(msg.to_string())
     }
 }
+
+pub(crate) fn unsupported_column_mapping_write(
+    feature: impl AsRef<str>,
+    reason: impl AsRef<str>,
+) -> DeltaTableError {
+    DeltaTableError::Generic(format!(
+        "Unsupported column mapping write for {}: {}",
+        feature.as_ref(),
+        reason.as_ref()
+    ))
+}
--- a/crates/core/src/kernel/transaction/protocol.rs
+++ b/crates/core/src/kernel/transaction/protocol.rs
@@ -237,7 +237,7 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     let mut reader_features = HashSet::new();
     reader_features.insert(TableFeature::TimestampWithoutTimezone);
     reader_features.insert(TableFeature::DeletionVectors);
-    // reader_features.insert(TableFeature::ColumnMapping);
+    reader_features.insert(TableFeature::ColumnMapping);
 
     let mut writer_features = HashSet::new();
     writer_features.insert(TableFeature::AppendOnly);
@@ -250,7 +250,7 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
         writer_features.insert(TableFeature::GeneratedColumns);
     }
     writer_features.insert(TableFeature::DeletionVectors);
-    // writer_features.insert(TableFeature::ColumnMapping);
+    writer_features.insert(TableFeature::ColumnMapping);
     // writer_features.insert(TableFeature::IdentityColumns);
 
     ProtocolChecker::new(reader_features, writer_features)
--- a/crates/core/src/operations/add_column.rs
+++ b/crates/core/src/operations/add_column.rs
@@ -3,10 +3,12 @@
 use std::sync::Arc;
 
 use delta_kernel::schema::StructType;
+use delta_kernel::table_features::ColumnMappingMode;
 use futures::future::BoxFuture;
 use itertools::Itertools;
 
 use super::{CustomExecuteHandler, Operation};
+use crate::errors::unsupported_column_mapping_write;
 use crate::kernel::schema::merge_delta_struct;
 use crate::kernel::transaction::{CommitBuilder, CommitProperties};
 use crate::kernel::{
@@ -100,6 +102,13 @@ impl std::future::IntoFuture for AddColumnBuilder {
                 ));
             }
 
+            if snapshot.table_configuration().column_mapping_mode() != ColumnMappingMode::None {
+                return Err(unsupported_column_mapping_write(
+                    "ADD COLUMN",
+                    "adding column mapping metadata outside write schema merge is not supported yet",
+                ));
+            }
+
             let table_schema = snapshot.schema();
             let new_table_schema = merge_delta_struct(table_schema.as_ref(), fields_right)?;
 
@@ -136,5 +145,45 @@ impl std::future::IntoFuture for AddColumnBuilder {
                 commit.snapshot(),
             ))
         })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::kernel::{DataType, PrimitiveType, StructField};
+
+    #[tokio::test]
+    async fn test_add_column_rejects_column_mapping_table() -> crate::DeltaResult<()> {
+        let temp_dir = tempfile::tempdir()?;
+        let fixture_source = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("../test/tests/data/table_with_column_mapping");
+        fs_extra::dir::copy(&fixture_source, temp_dir.path(), &Default::default())
+            .map_err(crate::DeltaTableError::generic)?;
+        let table_url = url::Url::from_directory_path(
+            temp_dir
+                .path()
+                .join("table_with_column_mapping")
+                .canonicalize()?,
+        )
+        .unwrap();
+        let table = crate::open_table(table_url).await?;
+
+        let err = table
+            .add_columns()
+            .with_fields([StructField::new(
+                "new_column",
+                DataType::Primitive(PrimitiveType::String),
+                true,
+            )])
+            .await
+            .expect_err("ADD COLUMN should be rejected for column-mapped tables");
+
+        assert!(
+            err.to_string()
+                .contains("Unsupported column mapping write for ADD COLUMN:"),
+            "unexpected error: {err}"
+        );
+
+        Ok(())
     }
 }
--- a/crates/core/src/operations/add_feature.rs
+++ b/crates/core/src/operations/add_feature.rs
@@ -8,7 +8,6 @@ use itertools::Itertools;
 
 use super::{CustomExecuteHandler, Operation};
 use crate::DeltaTable;
-use crate::errors::unsupported_column_mapping_write;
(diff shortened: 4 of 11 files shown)


fksegundo:
Context behind this PR:

  I work with Delta tables where some downstream workflows inspect or query the
  Parquet layout directly, including DuckDB-based reads over the files. While
  evaluating delta-rs for those workflows, I found a compatibility gap with
  tables that use `columnMapping.mode = name`, especially after schema evolution.

  The table shape I wanted to cover is:

  - the table was created before column mapping;
  - column mapping was enabled later;
  - existing columns, including partition columns, kept their original physical
    names;
  - newly evolved data columns received generated `col-*` physical names;
  - existing Hive-style partition directories remained part of the table layout.

  This PR keeps Spark-style randomized prefixes as the default for partitioned
  column-mapped writes. It also adds an explicit delta-rs opt-in,
  `with_preserve_column_mapping_hive_style_partitions(true)`, for integrations
  that intentionally need to keep writing into Hive-style partition directories
  when the partition physical names are still equal to the logical names.

fksegundo:
Implementation note for reviewers:

This draft intentionally supports only a bounded column mapping surface for this milestone:

- `columnMapping.mode = name` writes are supported for top-level schemas.
- Appends and `SchemaMode::Merge` assign/preserve column mapping metadata for top-level fields.
- Partition values and stats are written with physical names.
- Partitioned column-mapped writes use random file prefixes by default, matching Spark/Databricks behavior.
- A delta-rs write-builder opt-in, `with_preserve_column_mapping_hive_style_partitions(true)`, keeps Hive-style partition paths for integrations that intentionally depend on that layout.
- `columnMapping.mode = id`, nested column-mapped writes, CDC file writes, `OPTIMIZE`, and `MERGE` with schema evolution are explicitly rejected with consistent unsupported errors until those paths are implemented end to end.

Motivation/context: I work with Delta tables that evolved from normal schemas to `columnMapping.mode = name`, and I often inspect/query the underlying Parquet files directly with tools such as DuckDB. I hit this compatibility gap while evaluating delta-rs for a PoC, especially around schema evolution where existing columns keep their original physical names while newly added columns receive generated `col-*` physical names.

I also published a standalone lab that exercises the integration behavior and report-style filtered queries: https://github.com/fksegundo/delta-column-mapping-lab

ethan-tyler:
@fksegundo - Not sure if this is still a wip, but if it's ready for review can you please mark it as such. Thanks!

fksegundo:
@ethan-tyler  Sorry for the delay , i'm pushing one more validation pass and will mark ready for review shortly.

ethan-tyler:
No worries, I saw you were merging in updates from main so it looked like you were treating this as ready for review.

fksegundo:
I pushed the final validation pass, rewrote the branch into a linear signed-off history,

fksegundo:
I checked the Python CI failures and found that the new guard on `ADD FEATURE columnMapping` was too broad.

  `add_feature(TableFeatures.ColumnMapping)` previously only added the table feature to the protocol. It does not set `delta.columnMapping.mode` and does not by itself activate column mapping semantics. The effective column mapping mode
  still remains `None` unless the table metadata has `delta.columnMapping.mode`.

  I’m going to restore that behavior and keep the guard on the actual activation path, `SET TBLPROPERTIES delta.columnMapping.mode`, because that path would require schema metadata migration.

fksegundo:
Follow-up on the Python CI failure: I restored the previous `ADD FEATURE columnMapping` behavior and pushed the fix.

  `add_feature(TableFeatures.ColumnMapping)` is now treated as a protocol-only feature addition again. It does not set `delta.columnMapping.mode`, so it does not activate column mapping semantics by itself. The guard remains on `SET
  TBLPROPERTIES delta.columnMapping.mode`, which is the path that would require schema metadata migration.

  I rebuilt the local Python binding with CPython 3.10 and validated the previously failing cases locally:

  ```bash
  uv run --no-sync pytest tests/test_alter.py -q -k "add_feature_variations or add_features" --maxfail=1
  uv run --no-sync pytest tests/test_lakefs.py -q -m "lakefs and integration" -k add_features --maxfail=1

  Results:

  tests/test_alter.py: 17 passed
  tests/test_lakefs.py::test_add_features: 1 passed

ethan-tyler (changes requested):
Thanks @fksegundo for your interest in delta-rs and I appreciate the effort here, column mapping writer support is a feature that would be appreciated. 

I took a first pass at this and unfortunately found several protocol correctness issues that I don't think we can work through in a pr of this size. A couple examples:

- Parquet field ids aren't being written for column mapped columns. Physical names alone aren't sufficient because delta column mapping requires field ids for protocol compliance across readers

- `ColumnMapping` is enabled globally but not every write path is column mapping aware. Legacy paths like `RecordBatchWriter` / `JsonWriter` still write logical schemas and partition values. Those either need to reject column mapped tables or be updated

- Column mapping metadata (`delta.columnMapping.id`, `physicalName`) can come in from user provided schemas during schema merge. The table implementation needs to own and validate those values.

Given these initial concerns touching protocol correctness and the size of the pr, I don't think I'd have the bandwidth to coach and closely review this into a mergeable state as one pr.

I'd recommend:
- Breaking this into smaller focused prs
- Revisiting the delta column mapping protocol spec to make sure we are preserving the full invariant
- Starting with a narrower surface like append only on existing name mode tables and expanding from there

Happy to review smaller chunks, but blocking on protocol correctness for now

fksegundo:
@ethan-tyler  Thanks for the review, that makes sense.

  I agree this PR is too large to review and merge safely as-is, especially with the protocol correctness concerns you pointed out.

  The reason it grew this way is that my initial goal was to validate whether delta-rs could support a PoC for a reporting workflow I’m working on, where we need to read and write existing `columnMapping.mode = name` tables. I used this
  branch to explore the end-to-end behavior, including schema evolution, partition handling, stats, and rewrite paths.

  Regardless of this PR’s outcome, I’d still like to contribute this support upstream. I’ll close this PR and split the work into smaller focused PRs, starting with a narrower append-only path for existing name-mapped tables and
  revisiting the column mapping protocol requirements first.

  Before I start splitting this up, do you think append-only support for existing `columnMapping.mode = name` tables is the right first slice, or would you prefer a protocol/guardrails PR first?

ethan-tyler:
thanks for understanding @fksegundo. I'd start with a protocol/guardrails pr before enabling any writer support. The main thing I think is important here is a clear safety boundary, where every unsupported column mapped write surface rejects before writing files. 

After that, a narrow/well scoped append only on existing name mode tables is a good first slice. Schema evolution can come separately once id assignment, maxColumnId, and reserved metadata handling are reviewed on their own. Thank you!

fksegundo:
Thanks, that helps a lot.
I’ll close this PR and come back with a smaller first PR focused on the safety boundary before enabling writer support.

PR #4420, opened 2026-05-01
feat: on Python, nanosecond timestamp support is compiled by default, disabled by default, and can be enabled at runtime

itamarst:
This will allow Python users to access nanosecond timestamps, without changing default behavior, by making it a runtime setting and leaving it disabled by default.


itamarst:
For review, the main risk is that behavior changes somehow even if nanosecond timestamps are left disabled at runtime.

PR #4421, opened 2026-05-01
refactor: selective stats materialization policy

ethan-tyler:
Adds `StatsProjection` and `FileStatsMaterialization` enabling scan paths to choose: full stats, `numRecords` only, predicate scoped stats, no stats, or raw json preservation for compat.

Scan row materialization prefers existing `stats_parsed` from checkpoints and falls back to raw json when needed. Query oriented paths can omit raw json while compat paths (`Snapshot::files`, `file_views`, eager materialization) still preserve it for add action round trips.

This is another foundational PR and is **not** the full raw json memory fix. My plan is to land the next follow up after `EagerSnapshot` removal and will wire query paths to omit raw json by default.

part of #4015

### DIFF
--- a/crates/core/src/kernel/snapshot/iterators/scan_row.rs
+++ b/crates/core/src/kernel/snapshot/iterators/scan_row.rs
@@ -43,19 +43,16 @@ pin_project! {
 }
 
 impl<S> ScanRowOutStream<S> {
-    pub fn try_new(
+    pub fn try_new_with_materialization(
         snapshot: Arc<KernelSnapshot>,
         stream: S,
-        skip_stats: bool,
+        stats_materialization: FileStatsMaterialization,
     ) -> DeltaResult<Self> {
-        let stats_schema = snapshot.stats_schema()?;
+        let stats_schema = stats_materialization
+            .stats_projection()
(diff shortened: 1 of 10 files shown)


crates/core/src/kernel/snapshot/iterators/scan_row.rs
```diff
@@ -25,34 +27,45 @@ use crate::kernel::ARROW_HANDLER;
...
+use crate::kernel::snapshot::stats_projection::{
+    FIELD_PARTITION_VALUES_PARSED, FIELD_STATS, FIELD_STATS_PARSED, FileStatsMaterialization,
+    StatsSourcePolicy,
+};
 use crate::{DeltaResult, DeltaTableError};
 
 pin_project! {
     pub(crate) struct ScanRowOutStream<S> {
         stats_schema: KernelSchemaRef,
+        stats_arrow_schema: ArrowSchemaRef,
```
rtyler, line 41: This is a little peculiar to me since in theory the `KernelSchemaRef` is directly translatable to/from Arrow. I presume this is a convenience thing

rtyler (approved):

PR #4424, opened 2026-05-03
fix(core): reject unsupported column mapping writes

fksegundo:
## What changed

This PR adds defensive guardrails for Delta tables with `columnMapping.mode` enabled.

Unsupported write and metadata mutation paths now fail fast with a clear error before planning rewrites, writing data files, or committing unsafe metadata changes.

Covered paths:

- `WRITE`
- `CREATE TABLE` with column mapping mode or reserved column mapping metadata
- `ADD COLUMN`
- `SET TBLPROPERTIES delta.columnMapping.mode`
- `UPDATE`
- `DELETE`
- `MERGE`
- `OPTIMIZE`
- legacy `RecordBatchWriter`
- legacy `JsonWriter`

This PR does not enable column-mapped writes and does not add `ColumnMapping` to the global writer protocol support set.

## Why

Column-mapped table support needs to be introduced in smaller, reviewable steps. Before adding append/write support, unsupported paths should have a clear safety boundary so they cannot proceed into logical-schema write code or metadata updates that are not column-mapping aware.

## Validation

```bash
cargo fmt -- --check
make check-rust
make check-python
make develop
make unit-test
cargo test -p deltalake-core --features datafusion --test column_mapping_guardrails -- --nocapture
cargo test -p deltalake-core --features datafusion --lib add_feature -- --nocapture
```

## AI disclosure

I used Codex to help inspect the affected write paths and draft the guardrail tests. I reviewed the resulting changes and validated them locally with the commands above.

### DIFF
--- a/crates/core/src/writer/json.rs
+++ b/crates/core/src/writer/json.rs
@@ -7,7 +7,6 @@ use arrow::record_batch::*;
 use bytes::Bytes;
 use delta_kernel::engine::arrow_conversion::TryIntoArrow as _;
 use delta_kernel::expressions::Scalar;
-use delta_kernel::table_features::ColumnMappingMode;
 use indexmap::IndexMap;
 use itertools::Itertools;
 use object_store::path::Path;
@@ -22,9 +21,9 @@ use super::utils::{
     arrow_schema_without_partitions, next_data_path, record_batch_from_message,
     record_batch_without_partitions,
 };
-use super::{DeltaWriter, DeltaWriterError, WriteMode};
+use super::{DeltaWriter, DeltaWriterError, WriteMode, ensure_legacy_writer_supports_table};
 use crate::DeltaTable;
-use crate::errors::{DeltaTableError, unsupported_column_mapping_write};
+use crate::errors::DeltaTableError;
 use crate::kernel::{Add, PartitionsExt, scalars::ScalarExt};
 use crate::logstore::ObjectStoreRetryExt;
 use crate::parquet_utils::default_writer_properties;
@@ -196,6 +195,8 @@ impl JsonWriter {
             .with_storage_options(storage_options.unwrap_or_default())
             .load()
             .await?;
+        ensure_legacy_writer_supports_table(&table, "JsonWriter")?;
+
         // Initialize writer properties for the underlying arrow writer
         let writer_properties = default_writer_properties(parquet::basic::Compression::SNAPPY);
 
@@ -210,14 +211,7 @@ impl JsonWriter {
 
     /// Creates a JsonWriter to write to the given table
     pub fn for_table(table: &DeltaTable) -> Result<JsonWriter, DeltaTableError> {
-        if table
-            .snapshot()?
-            .table_config()
-            .column_mapping_mode
-            .is_some_and(|mode| mode != ColumnMappingMode::None)
-        {
-            return Err(unsupported_column_mapping_write("JsonWriter"));
-        }
+        ensure_legacy_writer_supports_table(table, "JsonWriter")?;
 
         // Initialize an arrow schema ref from the delta table schema
         let metadata = table.snapshot()?.metadata();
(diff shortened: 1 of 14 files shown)


fksegundo:
Added a small follow-up for the failing optimize tests.

  `RecordBatchWriter` and `JsonWriter` were rejecting any table with the `delta.columnMapping.mode` property, including tables where the mode is set
  to `none`. The deletion vector optimize fixture uses that property with `none`, so the test setup was failing before reaching `OPTIMIZE`.

  The guard now only rejects parsed column mapping modes other than `None`.

crates/core/tests/column_mapping_guardrails.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    let err =
+        JsonWriter::for_table(&table).expect_err("json writer should reject column-mapped tables");
+    assert_unsupported_column_mapping_write(&err, "JsonWriter");
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn column_mapping_guardrails_dml_rejects_before_files() -> TestResult {
```
hntd187, line 144: Can we gate this single test that uses datafusion instead of the entire test module?

fksegundo: Updated.
I removed the module-level gate and gated only the DataFusion-dependent tests instead, so the remaining guardrail tests can still run without the datafusion feature enabled.

crates/core/src/writer/json.rs
```diff
@@ -209,6 +210,15 @@ impl JsonWriter {
 
     /// Creates a JsonWriter to write to the given table
     pub fn for_table(table: &DeltaTable) -> Result<JsonWriter, DeltaTableError> {
+        if table
```
ethan-tyler: `for_table()` is guarded but `try_new()` also loads an existing table and can still construct a writer for a column mapped target. Could we add the same `column_mapping_mode` check there?

crates/core/src/writer/record_batch.rs
```diff
@@ -121,6 +122,15 @@ impl RecordBatchWriter {
 
     /// Creates a [`RecordBatchWriter`] to write data to provided Delta Table
     pub fn for_table(table: &DeltaTable) -> Result<Self, DeltaTableError> {
+        if table
```
ethan-tyler: Same thing as above in json.rs

ethan-tyler (commented):
Thanks for splitting this into a focused PR @fksegundo! Aligned with the approach - 

Left a couple inline comments. But basically `for_table()` rejects column mapped tables,
but the URI based constructors can still create writers against existing column mapped targets.

There are other write surfaces outside this diff (DataFusion `insert_into`, etc.) that I think should get a closer look, but I am not holding this PR if we agree the scope here is operation builders and legacy writer apis.

ethan-tyler:
Non-blocking follow ups and direction: - 
- Look into DataFusion `insert_into` / `DeltaDataSink` as additional write entry point outside this diff
- Decide whether explicit `delta.columnMapping.mode = "none"` should be allowed during create
- Reuse metadata validation for `ADD COLUMN` and schema evolution
- Consider renaming `unsupported_column_mapping_write` to something that is operation neutral

fksegundo:
Thanks for pointing this out. I added the guard to JsonWriter::try_new(), since that constructor already loads the table asynchronously before creating the writer.

  I also looked at doing the same for RecordBatchWriter::try_new(), but that path is different: it is a synchronous constructor that currently builds an unloaded table. Guarding it there would require loading
  table state from storage inside a sync API, or adding a blocking async bridge, which feels riskier than I would like for this PR.

  I think the cleaner follow-up would be to make RecordBatchWriter::try_new() async, or add an async equivalent, in a separate change. Given your note that this is not blocking, I left
  RecordBatchWriter::try_new() unchanged here and kept this PR scoped to the safe guardrails.

ethan-tyler (approved):
Thanks @fksegundo!

That makes sense and I am fine with leaving `RecordBatchWriter::try_new()` out of his pr.

I think `RecordBatchWriter` would be a good one to add to your next prs either as an async checked constructor or routing callers toward a checked path for existing tables. Thanks for working on this and a solid foundation pr. Looking forward to the next ones!

PR #4428, opened 2026-05-04
fix: reduce duplicate validation memory

### LINKED ISSUE #4422: [Bug]: Python 1.5.1 MERGE regression: small source batch uses substantially more RSS than 1.5.0
### What happened?

After upgrading the Python package from `deltalake==1.5.0` to `deltalake==1.5.1`, a Delta `MERGE` workload started using substantially more memory. In the original environment this eventually resulted in the process being killed by the OS.

I reduced this to a local synthetic repro with no production data. The workload is:

- Local Delta table
- String key column similar to a typical row key
- Large target table
- Small source update batch
- Merge predicate: `source.ROWKEY = target.ROWKEY`
- Unique source keys
- `streamed_exec=True`

Smaller local repro:

| deltalake | pyarrow | target rows | source rows | max RSS increase | end RSS |
|---|---:|---:|---:|---:|---:|
| `1.5.0` | `24.0.0` | 300k | 5k | `+126 MB` | `303 MB` |
| `1.5.1` | `24.0.0` | 300k | 5k | `+651 MB` | `825 MB` |

Larger local repro:

| deltalake | pyarrow | target rows | source rows | max RSS increase | end RSS |
|---|---:|---:|---:|---:|---:|
| `1.5.0` | `24.0.0` | 1M | 10k | `+294 MB` | `537 MB` |
| `1.5.1` | `24.0.0` | 1M | 10k | `+1,225 MB` | `1,442 MB` |

The compare from `python-v1.5.0` to `python-v1.5.1` is:

https://github.com/delta-io/delta-rs/compare/python-v1.5.0...python-v1.5.1

The most suspicious change I found is:

https://github.com/delta-io/delta-rs/pull/4333

That PR says duplicate-match validation adds marker columns:

- `__delta_rs_match_cardinality_class`
- `__delta_rs_target_row_index`

It also says this validation state is kept in memory per validation stream:

```text
target_row_state: HashMap<u64, (i32, i32)>
```

There is a follow-up merge validation semantics change here:

https://github.com/delta-io/delta-rs/pull/4359

The release also includes the DataFusion/Arrow/object_store upgrade:

https://github.com/delta-io/delta-rs/pull/4346

The memory jump follows `deltalake==1.5.1` in the local repro matrix.

### Expected behavior

I expected `deltalake==1.5.1` merge memory usage to remain in the same general range as `deltalake==1.5.0` for the same workload, especially with `streamed_exec=True`.

For this repro, `deltalake==1.5.0` uses roughly `+126 MB` to `+294 MB` max RSS depending on table size. I would expect `deltalake==1.5.1` to stay close to that range, or at least not require several times more RSS for the same small source batch and merge predicate.

If the additional memory is expected because of duplicate-match validation, I would expect a way to bound/spill that state or a documented mitigation for merge workloads where source keys are already unique.

### Operating System

macOS

### Binding

Python

### Bindings Version

1.5.1

### Steps to reproduce

1. Save this script as `delta_merge_memory_repro.py`.

```python
from __future__ import annotations

import argparse
import gc
import os
import shutil
import tempfile
import threading
import time
from pathlib import Path

import psutil
import pyarrow as pa


def make_target_table(row_count: int) -> pa.Table:
    return pa.table(
        {
            "ROWKEY": [f"key-{i:010d}" for i in range(row_count)],
            "rowversion": list(range(row_count)),
            "amount": [i % 1000 for i in range(row_count)],
            "payload": [f"payload-{i % 1000:04d}" for i in range(row_count)],
        }
    )


def make_source_table(*, target_rows: int, source_rows: int) -> pa.Table:
    start = max(0, target_rows - source_rows)
    keys = [f"key-{i:010d}" for i in range(start, target_rows)]
    return pa.table(
        {
            "ROWKEY": keys,
            "rowversion": list(range(target_rows, target_rows + source_rows)),
            "amount": [10_000 + (i % 1000) for i in range(source_rows)],
            "payload": [f"updated-{i % 1000:04d}" for i in range(source_rows)],
        }
    )


def sample_peak_rss(*, stop: threading.Event, samples: list[float]) -> None:
    while not stop.is_set():
        samples.append(process.memory_info().rss / 1024 / 1024)
        time.sleep(0.02)


def main() -> None:
    parser = argparse.ArgumentParser()
    parser.add_argument("--target-rows", type=int, default=300_000)
    parser.add_argument("--source-rows", type=int, default=5_000)
    parser.add_argument("--target-file-size", type=int, default=800_000)
    args = parser.parse_args()

    table_path = Path(tempfile.mkdtemp(prefix="delta-merge-memory-repro-"))

    try:
        write_deltalake(
            str(table_path),
            make_target_table(args.target_rows),
            mode="overwrite",
            target_file_size=args.target_file_size,
        )

        gc.collect()
        source = make_source_table(
            target_rows=args.target_rows,
            source_rows=args.source_rows,
        )

        before_rss_mb = process.memory_info().rss / 1024 / 1024
        samples: list[float] = []
        stop = threading.Event()
        sampler = threading.Thread(
            target=sample_peak_rss,
            kwargs={"stop": stop, "samples": samples},
            daemon=True,
        )

        sampler.start()
        started = time.perf_counter()
        metrics = (
            DeltaTable(str(table_path))
            .merge(
                source=source,
                predicate="source.ROWKEY = target.ROWKEY",
                source_alias="source",
                target_alias="target",
                streamed_exec=True,
            )
            .when_matched_update(
                {
                    "rowversion": "source.rowversion",
                    "amount": "source.amount",
                    "payload": "source.payload",
                }
            )
            .when_not_matched_insert(
                {
                    "ROWKEY": "source.ROWKEY",
                    "rowversion": "source.rowversion",
                    "amount": "source.amount",
                    "payload": "source.payload",
                }
            )
            .execute()
        )
        duration_seconds = time.perf_counter() - started

        stop.set()
        sampler.join(timeout=1)
        after_rss_mb = process.memory_info().rss / 1024 / 1024
        peak_rss_mb = max(samples or [before_rss_mb])

        print(
            {
                "target_rows": args.target_rows,
                "source_rows": args.source_rows,
                "before_rss_mb": round(before_rss_mb, 1),
                "peak_rss_mb": round(peak_rss_mb, 1),
                "after_rss_mb": round(after_rss_mb, 1),
                "peak_rss_delta_mb": round(peak_rss_mb - before_rss_mb, 1),
                "duration_seconds": round(duration_seconds, 3),
                "metrics": metrics,
            }
        )
    finally:
        shutil.rmtree(table_path, ignore_errors=True)


if __name__ == "__main__":
    main()
```

2. Run the small comparison.

```bash
uv run --no-project --with deltalake==1.5.0 --with pyarrow==24.0.0 --with psutil python delta_merge_memory_repro.py --target-rows 300000 --source-rows 5000 --target-file-size 800000
uv run --no-project --with deltalake==1.5.1 --with pyarrow==24.0.0 --with psutil python delta_merge_memory_repro.py --target-rows 300000 --source-rows 5000 --target-file-size 800000
```

3. Optionally run the larger confirmation.

```bash
uv run --no-project --with deltalake==1.5.0 --with pyarrow==24.0.0 --with psutil python delta_merge_memory_repro.py --target-rows 1000000 --source-rows 10000 --target-file-size 1200000
uv run --no-project --with deltalake==1.5.1 --with pyarrow==24.0.0 --with psutil python delta_merge_memory_repro.py --target-rows 1000000 --source-rows 10000 --target-file-size 1200000
```

### Relevant logs

```shell

```

ethan-tyler:
Fixes #4422.

Duplicate match validation introduced a buffered `row_number()` window over the join output, which held the full target side in memory.

The patch here is to compute target row ordinals inside `DeltaScanStream` instead. Keep validation on one streaming path, retain state only for matched rows that can produce duplicate errors. Preserve streamed source behavior for single use `RecordBatchReader` inputs.

Memory improvement (manual repro):
- 300k target / 5k source: ~109 MB (was ~861 MB on 1.5.1)
- 1M target / 10k source: ~243 MB

Tests cover scan ordinals, validation, plan shape, and python streamed source regression.

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
@@ -11,9 +11,9 @@ use std::task::{Context, Poll};
 
 use arrow::array::{RecordBatch, StringArray};
 use arrow::compute::filter_record_batch;
-use arrow::datatypes::{SchemaRef, UInt16Type};
+use arrow::datatypes::{DataType, Field, Schema, SchemaRef, UInt16Type};
 use arrow_array::StringViewArray;
-use arrow_array::{Array, BooleanArray};
+use arrow_array::{Array, ArrayRef, BooleanArray, UInt64Array};
 use dashmap::DashMap;
 use datafusion::common::config::ConfigOptions;
 use datafusion::common::error::{DataFusionError, Result};
@@ -118,6 +118,8 @@ pub struct DeltaScanExec {
     input_file_id_column: String,
     /// User-visible file-id column name when projected in the output.
     file_id_column: Option<String>,
+    /// Row index column name visible in projected output.
+    row_index_column: Option<String>,
     /// plan properties
     properties: Arc<PlanProperties>,
     /// Aggregated partition column statistics
@@ -134,6 +136,9 @@ impl DisplayAs for DeltaScanExec {
                 write!(f, "DeltaScanExec")?;
                 if let Some(file_id_column) = &self.file_id_column {
                     write!(f, ": file_id_column={file_id_column}")?;
+                }
+                if let Some(row_index_column) = &self.row_index_column {
+                    write!(f, ": row_index_column={row_index_column}")?;
                 }
                 Ok(())
             }
@@ -155,7 +160,17 @@ impl DeltaScanExec {
             .contract
             .retain_file_id
             .then(|| scan_plan.contract.file_id_field.name().to_owned());
-        let output_schema = scan_plan.effective_schema(file_id_column.is_some());
+        let row_index_column = scan_plan.contract.retain_row_index.then(|| {
+            scan_plan
+                .contract
+                .row_index_field
+                .as_ref()
+                .expect("row index field")
+                .name()
+                .to_owned()
+        });
+        let output_schema =
+            scan_plan.effective_schema(file_id_column.is_some(), row_index_column.is_some());
         let properties = Arc::new(PlanProperties::new(
             EquivalenceProperties::new(output_schema),
             input.properties().partitioning.clone(),
@@ -171,6 +186,7 @@ impl DeltaScanExec {
             metrics,
             input_file_id_column,
             file_id_column,
+            row_index_column,
             properties,
         }
     }
@@ -279,6 +295,12 @@ impl ExecutionPlan for DeltaScanExec {
         target_partitions: usize,
         config: &ConfigOptions,
     ) -> Result<Option<Arc<dyn ExecutionPlan>>> {
+        if self.row_index_column.is_some() {
+            // DeltaScanStream stores row ordinal counters per execution partition. Repartitioning
+            // can split a file's rows across streams and break ordinal contiguity.
+            return Ok(None);
+        }
+
         if let Some(input) = self.input.repartitioned(target_partitions, config)? {
             Ok(Some(Arc::new(Self {
                 input,
@@ -303,6 +325,8 @@ impl ExecutionPlan for DeltaScanExec {
             selection_vectors: Arc::clone(&self.selection_vectors),
             input_file_id_column: self.input_file_id_column.clone(),
             file_id_column: self.file_id_column.clone(),
+            row_index_column: self.row_index_column.clone(),
+            row_index_by_file: HashMap::new(),
             pending: VecDeque::new(),
             schema_adapter: super::SchemaAdapter::new(Arc::clone(
                 &self.scan_plan.contract.result_schema,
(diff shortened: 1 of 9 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -118,6 +118,8 @@ pub struct DeltaScanExec {
     input_file_id_column: String,
     /// User-visible file-id column name when projected in the output.
     file_id_column: Option<String>,
+    /// Row index column name visible in projected output.
+    row_index_column: Option<String>,
```
rtyler, line 122: There seems to be a lot of API surface area change to accommodate this new column to track, which as far as i can tell is always `TARGET_ROW_ORDINAL_IN_FILE_COLUMN`.

I recognize that there are other fixed name columns brought into this struct like the `file_id_column` but I would like to register my annoyance with the court on this general unnecessary API cruft, 😒 

Why `DeltaScanExec` needs these as properties is silly

rtyler (approved):

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
ethan-tyler: Annoyance received and taking note! I'll look into this more! Thank you for your patience @rtyler

PR #4432, opened 2026-05-06
test: port DeltaTableProvider tests to next provider

ethan-tyler:
Part of #4239 

Ports tests from `DeltaTableProvider` construction to next provider keeping the follow up deletion pr more focused

### DIFF
--- a/python/src/datafusion.rs
+++ b/python/src/datafusion.rs
@@ -30,7 +30,7 @@ pub(crate) struct LazyTableProvider {
 }
 
 impl LazyTableProvider {
-    /// Build a DeltaTableProvider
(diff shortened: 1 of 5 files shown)


python/src/datafusion.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) struct LazyTableProvider {
 }
 
 impl LazyTableProvider {
-    /// Build a DeltaTableProvider
+    /// Build a LazyTableProvider
```
ethan-tyler, line 33: this was a stale comment that incorrectly referred to `DeltaTableProvider`

ion-elgreco (approved):

Issue #3551, opened 2025-06-22
Mapped network drives in Windows no longer work


riziles:
# Environment

**Delta-rs version**: 1.0.2

**Binding**: Python

**Environment**:
- **Cloud provider**: Unknown
- **OS**: Windows

***
# Bug

**What happened**: Mapped network drives in Windows no longer work. Error message:
```
Unable to convert URL "file:///path/to/my/folder/_delta_log/_last_checkpoint" to filesystem path
```

**What you expected to happen**: Write delta table

**How to reproduce it**:

```python
import arro3.core
from deltalake import write_deltalake

arr1 = arro3.core.Array([1, 2, 3], type=arro3.core.DataType.int64())
tbl = arro3.core.Table.from_pydict({"a": arr1})

dpath = "s:/path/to/my/folder"

write_deltalake(
    dpath,
    tbl,
    mode="overwrite",
)
```

**More details**: Prior to v1.0, the above example worked fine. After upgrading, it no longer works. I've tried setting the `MOUNT_ALLOW_UNSAFE_RENAME` to `True` (and `true`), but that did not help. I've also tried passing in the path as a URI, but that did not help either. Is this just no longer possible, or is there some configuration parameter that I am missing?

rtyler:
This is going to be tricky to fix because many of us do not develop on Windows, let alone network drives! 😬 

I don't have a good way to move this forward 🤔

riziles:
Ughh. I was hoping there was just some configuration piece I was missing. I've spent countless hours trying to figure it out, but it seems to be happening inside the Rust code, and I don't have a way to debug. If you think of anything I should try, please let me know.

tbezemer:
We are also experiencing this issue. It's blocking for us - we cannot update our polars and deltalake versions because of this. It also happens with non-mapped network paths (e.g. `\\HOSTNAME\data\my_folder`)

tbezemer:
I wanted to try and contribute a fix for this, so I cloned the repo on a Windows laptop that I have lying around. However, I cannot even get the unit tests to run correctly. It seems things are thoroughly broken on Windows. The test in `crates/core/tests/dat.rs` fails with a similar error message:
`The system cannot find the path specified.`. Something is off with paths on windows.

riziles:
I believe the problem is in https://github.com/delta-io/delta-kernel-rs , not this repo.

tbezemer:
@riziles Thanks! I'll look into it and see if I can contribute a fix.

[...]

PR #4436, opened 2026-05-09
fix: support Windows unc paths

lachlancahill:
<type>fix: <description>

**Description**

Reading and writing Delta tables from Windows UNC paths (`\\server\share\path\to\table`) currently fails. delta-rs canonicalises the path and hands the resulting URL to `object_store::parse_url_opts`. This PR fixes that end-to-end without requiring any change in `apache/arrow-rs-object-store` or `delta-io/delta-kernel-rs`.

**Related Issue(s)**

Resolves issues:

- [delta-io/delta-kernel-rs#1482](https://github.com/delta-io/delta-kernel-rs/issues/1482)
- [delta-io/delta-rs#3551](https://github.com/delta-io/delta-rs/issues/3551)

Underlying cause upstream is apache/arrow-rs-object-store#715 though I don't think the fix belongs there as it seems likely that it would break existing code which expects the current behaviour dropping the host. Arguably it should fail loudly on UNC paths and refer downstream devs to LocalFileSystem::new_with_prefix to make these issues easier to debug in the future, but that's a separate discussion.

**What's Changed:**

A small Windows-only branch in `crates/core/src/logstore/mod.rs::logstore_for`, a cross-platform helper, and a `delta-unc` URL scheme registration:

- **`DELTA_UNC_SCHEME` / `DELTA_UNC_PREFIX_KEY`** — constants for the `"delta-unc"` URL scheme and its `unc_prefix` query parameter. Using a non-`file` scheme prevents the kernel from calling `Url::to_file_path()` (which rejects paths without a drive letter on Windows). The query parameter stores the original `\\server\share` prefix so that `to_uri` can reconstruct native UNC filesystem paths.
- **`split_unc_url(&Url) -> Option<(PathBuf, Url)>`** — detects `file://server/share/...` URLs, peels off `(server, share)`, and returns `(\\server\share PathBuf, delta-unc:///table_path/?unc_prefix=... Url)`. Cross-platform (so it can be unit tested on any host); only consumed by Windows code.
- **`logstore_for`** — on Windows, when a UNC URL is detected, builds `LocalFileSystem::new_with_prefix(\\server\share)` directly and forwards the share-relative `delta-unc:///` URL to `logstore_with`. Falls through to the existing factory lookup in every other case (non-Windows, non-UNC, or non-`file` schemes).
- **`to_uri`** — handles `delta-unc` scheme by extracting the UNC prefix from the query parameter and reconstructing native filesystem paths (e.g. `\\server\share\table\file.parquet`). This ensures downstream consumers (Polars, PyArrow) receive valid Windows paths.
- **`extract_unc_prefix`** — small helper to read the `unc_prefix` query parameter from a `delta-unc` URL.
- **`logstore_factories`** — registers `delta-unc://` on Windows, pointing to `DefaultLogStoreFactory`, so that `logstore_with` can dispatch the rewritten URL.
- **7 unit tests** for `split_unc_url` covering: extraction of server+share with prefix round-trip, nested paths, share root, percent-decoded share names, and rejection of non-`file` schemes etc. 

**Underlying problem:**  

UNC Paths break because:

1. **`object_store` silently drops the host.** `object_store::path::Path::from_absolute_path` calls `Url::from_file_path(...).path()` — the URL path component excludes the host by definition, so `\\server\share\table` becomes just `share/table` with the server lost. This is the root cause.

2. **The kernel rejects share-relative `file://` URLs on Windows.** Even after fixing (1) by constructing `LocalFileSystem::new_with_prefix(\\server\share)` and rewriting the URL to a share-relative `file:///table/`, the kernel's `resolve_uri_type` calls `Url::to_file_path()` on every `file://` URL. On Windows, `file:///table/` has no drive letter and no host, so `to_file_path()` returns `Err` and the kernel emits `Invalid table location: file:///table/.`

This PR solves both in one pass:

- Problem (1): bypass `object_store::parse_url_opts` for UNC inputs and construct `LocalFileSystem::new_with_prefix` directly.
- Problem (2): use the `delta-unc` URL scheme for the share-relative URL so the kernel treats it as a generic URL (skipping `to_file_path()` validation) while preserving the correct path math.

**Why `delta-unc://`**

The kernel validates `file://` URLs by calling `Url::to_file_path()`, which on Windows requires either a drive letter (`file:///C:/...`) or a UNC host (`file://server/...`). A share-relative `file:///table/` has neither, so validation fails.

Any non-`file` scheme bypasses this: the kernel's `resolve_uri_type` returns non-file URLs as `UriType::Url(url)` with no filesystem validation. The URL is then used purely for path computation (joining `_delta_log/...` etc.), and the engine's object store handles actual I/O.

The path math is identical regardless of scheme — `delta-unc:///table/` and `file:///table/` have the same path component `/table/`, so `PrefixStore` decoration and `LocalFileSystem` resolution work the same way. The scheme change is purely to satisfy the kernel's validation gate.

**Why this fix lives here (not upstream)**

**Why not `object_store`**

`object_store` already exposes the right primitive: `LocalFileSystem::new_with_prefix(r"\\server\share")` works correctly today. delta-rs simply isn't using it because `object_store::parse_url_opts` doesn't dispatch to it for UNC inputs. The connector-side fix is to detect UNC at the engine factory and instantiate `LocalFileSystem` with the prefix, bypassing the broken parse path for this one case.

Considered three upstream options and ruled them out for this PR:

1. **Reject UNC loudly** — doesn't fix the user-visible bug.
2. **Preserve UNC info in `Path`** — large blast radius across every `ObjectStore` impl; unlikely to land quickly.
3. **Documentation only** — helpful but not sufficient.

**Why not `delta-kernel-rs`**

The kernel never sees a UNC URL because it's rewritten before it gets there. URI interpretation is already a connector responsibility — pushing platform-specific URI surgery into the kernel concentrates Windows-specific complexity in a cross-platform crate that is supposed to be store-agnostic.

A small kernel patch (making `resolve_uri_type` fall through to `UriType::Url` when `to_file_path()` fails on Windows) would also solve problem (2) and is documented in `PROPOSED_KERNEL_RS_PATCH.md` for future consideration. The `delta-unc` scheme approach used here avoids the need for that patch.

**Behaviour**

| Input on Windows | Before | After |
|---|---|---|
| `\\server\share\table` | `Kernel error: Invalid table location` | Reads/writes correctly via `LocalFileSystem` rooted at `\\server\share` |
| `C:\path\to\table` | Works | Works (no path change) |
| `file:///C:/path/to/table` | Works | Works (no path change) |
| `s3://bucket/key`, `memory://` | Works | Works (factory fallthrough) |

On non-Windows hosts, everything compiles to exactly the same code as before the PR.

**End to End testing performed:** 

Built a `deltalake` python wheel with these changes on both Windows and MacOS, and confirmed basic read/write functionality via polars now works for network drives + local on windows, and still works for local on macos. 

```python 
import polars as pl

# TEST_PATH = r"\\DESKTOPBESTTOP\TestShare\basic_partitioned" # Test shared drive on windows
# TEST_PATH = r"C:\TestShare\basic_partitioned" # Test local path on windows
TEST_PATH = r"./test_deltatable"  # Test local path on UNIX

df = pl.read_delta(TEST_PATH, version=None)
print(df)
```

### DIFF
--- a/crates/core/src/logstore/factories.rs
+++ b/crates/core/src/logstore/factories.rs
@@ -153,6 +153,11 @@ pub fn logstore_factories() -> LogStoreFactoryRegistry {
                 Url::parse("file://").unwrap(),
                 Arc::new(DefaultLogStoreFactory::default()),
             );
+            #[cfg(target_os = "windows")]
+            registry.insert(
+                Url::parse(&format!("{}://", super::DELTA_UNC_SCHEME)).unwrap(),
+                Arc::new(DefaultLogStoreFactory::default()),
+            );
             registry
         })
         .clone()
--- a/crates/core/src/logstore/mod.rs
+++ b/crates/core/src/logstore/mod.rs
@@ -960,18 +960,6 @@ pub(crate) mod tests {
     }
 
     #[test]
-    fn split_unc_url_rejects_localhost() {
-        let location = Url::parse("file://localhost/path/to/table/").unwrap();
-        assert!(split_unc_url(&location).is_none());
-    }
-
-    #[test]
-    fn split_unc_url_rejects_host_without_share() {
-        let location = Url::parse("file://server/").unwrap();
-        assert!(split_unc_url(&location).is_none());
-    }
-
-    #[test]
     fn logstore_with_memory_and_rt() {
         let location = Url::parse("memory:///table").unwrap();
         let store = logstore_for(
--- a/crates/core/src/logstore/mod.rs
+++ b/crates/core/src/logstore/mod.rs
@@ -210,8 +210,17 @@ pub fn logstore_for(location: &Url, storage_config: StorageConfig) -> DeltaResul
     Err(DeltaTableError::InvalidTableLocation(location.to_string()))
 }
 
+/// URL scheme used internally for UNC-backed tables. The kernel validates `file://` URLs by
+/// calling `Url::to_file_path()`, which rejects share-relative paths on Windows (no drive
+/// letter). Using a non-file scheme makes the kernel treat it as a generic URL, skipping
+/// filesystem validation while preserving correct path math for the prefixed `LocalFileSystem`.
+#[cfg(any(target_os = "windows", test))]
+pub(crate) const DELTA_UNC_SCHEME: &str = "delta-unc";
+
 /// Split a `file://server/share/...` URL produced from a Windows UNC path into
-/// `(\\server\share PathBuf, share-relative file:/// URL)`.
+/// `(\\server\share PathBuf, share-relative delta-unc:/// URL)`.
+///
+/// The returned URL uses the `delta-unc` scheme so the kernel skips `to_file_path()` validation.
 ///
 /// Returns `None` for URLs that are not UNC-shaped (any non-`file` scheme, no host, the
 /// `localhost` sentinel, or a missing share segment). Defined cross-platform so it can be unit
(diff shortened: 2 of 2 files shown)


lachlancahill:
AI Disclosure: I used Claude Code (Opus 4.7 and 4.6) across `delta-io/delta-rs`, `delta-io/delta-kernel-rs` and `apache/arrow-rs-object-store` repos to help diagnose the issue, determine the best place for the fix, and draft the code changes.

rtyler:
This change generally looks fine to me, I have cleaned things up for merge in #4447

Issue #4448, opened 2026-05-12
`DeltaTable.merge()` does a type mismatch between source and target raise an error


ethan-tyler:
### Discussed in https://github.com/delta-io/delta-rs/discussions/4369

<div type='discussions-op-text'>

<sup>Originally posted by **loic-bellinger** April 13, 2026</sup>
## Context

The documentation states:

> `write_deltalake` will raise `ValueError` if the schema of the data passed to it differs from the existing table's schema.

https://delta-io.github.io/delta-rs/usage/writing/

## Observed behavior with `DeltaTable.merge()`

When merging data whose column type **differs** from the target table schema (e.g. `Int64` source into a `String` target column), **no error is raised**. The data is silently cast to match the target schema (seems similiar to that [issue](https://github.com/delta-io/delta-rs/issues/2642)).

Minimal repro :

```python
import polars as pl
from deltalake import write_deltalake, DeltaTable

# 1. Create table with 'value' as String
write_deltalake(
    "/tmp/t",
    pl.DataFrame({"id": [1, 2], "value": ["a", "b"]}),
    mode="overwrite",
)

# 2. Merge with 'value' as Int64 - different type, no error raised
(
    DeltaTable("/tmp/t")
    .merge(
        pl.DataFrame({"id": [2, 3], "value": [99, 100]}),  # Int64 source
        predicate="source.id = target.id",
        source_alias="source",
        target_alias="target",
        # error_on_type_mismatch=True  ← default, but has no effect here
    )
    .when_matched_update_all()
    .when_not_matched_insert_all()
    .execute()
)

# 3. Schema unchanged (String), Int64 silently coerced → "99", "100"
print(DeltaTable("/tmp/t").schema())  # value: string
print(pl.read_delta("/tmp/t"))
# shape: (3, 2)
# ┌─────┬───────┐
# │ id  ┆ value │
# │ --- ┆ ---   │
# │ i64 ┆ str   │  ← still String
# ╞═════╪═══════╡
# │ 1   ┆ a     │
# │ 2   ┆ 99    │  ← Int64 silently cast to String
# │ 3   ┆ 100   │  ← idem
# └─────┴───────┘
```

Note: `error_on_type_mismatch=True` is the default in `.merge()` but does **not** prevent this silent coercion.

## Additional finding

An error **is** raised only when the cast is **impossible** (e.g. `"abc"` into an `Int64` column):

```
Exception: Cast error: Cannot cast string 'abc' to value of Int64 type
```

[...]

PR #4454, opened 2026-05-13
fix: restore merge safe cast behavior

ethan-tyler:
The merge `safe_cast` flag was being threaded into planning but ignored during execution, which caused `error_on_type_mismatch=False` to quietly behave like the strict path. This wires the flag through allowing merge write projections and CDF projections to actually use safe casts when the caller asks for them  and keeps strict casts as the default.

The behavior lines up with what the docs describe for merge and update. I also cleaned up the type mismatch docs in both places since they were a little ambiguous about which knob controls what. Added rust and python regression coverage for strict casts, safe casts, CDF output, schema merge interactions, overflow, and NOT NULL handling so this path doesn't silently regress again.

part of #4448

### DIFF
--- a/crates/core/src/operations/merge/mod.rs
+++ b/crates/core/src/operations/merge/mod.rs
@@ -47,7 +47,7 @@ use datafusion::logical_expr::build_join_schema;
 use datafusion::logical_expr::simplify::SimplifyContext;
 use datafusion::logical_expr::utils::split_conjunction_owned;
 use datafusion::logical_expr::{
-    Expr, JoinType, col, conditional_expressions::CaseBuilder, lit, when,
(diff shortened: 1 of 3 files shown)


crates/core/src/operations/merge/mod.rs
```diff
@@ -138,6 +136,14 @@ const SOURCE_COUNT_ID: &str = "merge_source_count";
 const TARGET_COUNT_ID: &str = "merge_target_count";
 const OUTPUT_COUNT_ID: &str = "merge_output_count";
 
+fn merge_cast(expr: Expr, data_type: DataType, safe_cast: bool) -> Expr {
```
rtyler, line 139: This is so wittle it could be cute to add a little `#[inline]` onto it, but i'm not sure if that would ever do anything from a performance standpoint 😆

rtyler (approved):

PR #4455, opened 2026-05-13
fix: normalize commit metadata handling

### LINKED ISSUE #4443: [Feature]: Support passing datatypes other than string in custom_metadata in CommitProperties from python

I am trying to create commits in the delta transaction log by hand and I am using DeltaTable.create_write_transaction for this and it works well. But the table history shown by databricks "DESCRIBE HISTORY" doesn't show the all the metadata as if they commit was made from databricks / delta internal scala code. I understand that is probably normal because the exact commit metadata isn't standardized.

But I thought as a user I should be able to use custom_metadata of CommitProperties, to fill as much of this data as possible. However, this currently with a TypeError if the value of the entry in custom_metadata isn't a string.

I think the problem is at the python->rust boundary, because the PyCommitProperties objects has a HashMap<String, String> member, it could be a HashMap<String, PyObject> maybe, and support any type of data in the value side?

Specifically, I want to add metadata using "custom_metadata" of CommitProperties, but I would like to pass metadata like this:

```
custom_metadata={
    "readVersion": 15,
    "operationParameters": {
        "numAddedRows": 123,
        ...
    },
    "my_string_key": "my_string_value",
}
```

This currently fails with TypeError because not all values are strings. If the custom_metadata only contains "my_string_key" then it works, but "readVersion" does not work because it is an int, and operationParameters does not work because it's value is a dict. I could json encode the dict, but then it gets double-json-encoded and ends up as a string in the transaction log. readVersion probably could be set as a string, but still the json that ends up in the transaction log would be different

json-encoding values myself, but it doesn't work

### Priority

Medium - Would be helpful

### Additional context

_No response_

### Contribution

ethan-tyler:
Cleans up a few rough edges in how python commit metadata flows through to the commit info action. This fix makes `CommitProperties.custom_metadata` accept arbitrary json typed values instead of being restricted to strings, which matches what the protocol allows and what users were already trying to pass in.

Reserved commit info keys are promoted into their typed fields on `CommitInfo` rather than being left in the custom metadata bag, this makes sure we don't end up emitting duplicate json keys when a user passes one through. Generated commit fields are preserved on the way out, but `clientVersion` stays user overridable since there are legitimate reasons to want to set it explicitly. Also threads `app_transactions` through `create_write_transaction`, which was being dropped on that path.

Closes #4443

### DIFF
--- a/crates/core/src/kernel/transaction/mod.rs
+++ b/crates/core/src/kernel/transaction/mod.rs
@@ -74,6 +74,7 @@
 //!       └───────────────────────────────┘
 //!</pre>
 use std::collections::HashMap;
+use std::str::FromStr;
 use std::sync::Arc;
 
 use bytes::Bytes;
@@ -93,11 +94,13 @@ use serde::{Deserialize, Serialize};
 
 use self::conflict_checker::{TransactionInfo, WinningCommitSummary};
 use crate::errors::DeltaTableError;
-use crate::kernel::{Action, CommitInfo, EagerSnapshot, Metadata, Protocol, Transaction, Version};
+use crate::kernel::{
+    Action, CommitInfo, EagerSnapshot, IsolationLevel, Metadata, Protocol, Transaction, Version,
+};
 use crate::logstore::ObjectStoreRef;
 use crate::logstore::{CommitOrBytes, LogStoreRef};
 use crate::operations::CustomExecuteHandler;
-use crate::protocol::DeltaOperation;
+use crate::protocol::{DeltaOperation, operation_parameter_value};
 use crate::protocol::{cleanup_expired_logs_for, create_checkpoint_for};
 use crate::table::config::TablePropertiesExt as _;
 use crate::table::state::DeltaTableState;
@@ -115,6 +118,21 @@ mod state;
 
 const DELTA_LOG_FOLDER: &str = "_delta_log";
 pub(crate) const DEFAULT_RETRIES: usize = 15;
+// These keys map to typed CommitInfo fields used by common Spark compatible writers. They are not
+// required by the protocol, but leaving them in flattened metadata can produce duplicate JSON keys.
+// Keep this list aligned with validate_reserved_commit_metadata in python/src/lib.rs.
+const RESERVED_COMMIT_INFO_KEYS: &[&str] = &[
+    "timestamp",
+    "userId",
+    "userName",
+    "operation",
+    "operationParameters",
+    "readVersion",
+    "isolationLevel",
+    "isBlindAppend",
+    "engineInfo",
+    "userMetadata",
+];
 
 #[derive(Default, Debug, PartialEq, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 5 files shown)


beranekba-Rational:
I tested this branch and it works well, it fixes #4443

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -290,6 +305,139 @@ pub struct CommitData {
...
+}
+
+fn log_unexpected_reserved_metadata_type(key: &str, value: &Value, expected: &str) {
+    debug!(
+        key,
+        expected,
+        actual = json_value_kind(value),
+        "Ignoring reserved commit metadata key with unexpected type"
+    );
+}
```
rtyler, line 412: I understand the draw of this kind of refactoring _however_ the downside to keep in mind when doing something like this is that `debug!()` is a macro that is going to take the line information from _here_ not from the caller. 

I think this reduces the utility of the debug message. (non-blocking feedback)

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -115,6 +118,18 @@ mod state;
 
 const DELTA_LOG_FOLDER: &str = "_delta_log";
 pub(crate) const DEFAULT_RETRIES: usize = 15;
+const RESERVED_COMMIT_INFO_KEYS: &[&str] = &[
```
rtyler, line 124: I have always had a strong distaste for [the Commit Provenance](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#commit-provenance-information) information which is "optional" but **not really** for almost all writers which rhyme with Fark.

Please include some commentary here on what is dictating these fields, since it's not the protocol, how will future maintainers decide whether or not a field should be listed here (or not!)?

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -290,6 +305,139 @@ pub struct CommitData {
     pub app_transactions: Vec<Transaction>,
 }
 
+fn normalize_reserved_commit_metadata(
```
rtyler, line 318: Even though this is an internal function, it would benefit from some rustdoc with more context, like why this function should exist, and probably a link to the originating issue

crates/core/src/kernel/transaction/mod.rs
```diff
@@ -290,6 +305,139 @@ pub struct CommitData {
...
+        Value::Array(_) => "array",
+        Value::Object(_) => "object",
+    }
+}
+
+fn strip_reserved_commit_info_keys(metadata: &mut HashMap<String, Value>) {
+    for key in RESERVED_COMMIT_INFO_KEYS {
+        metadata.remove(*key);
+    }
+}
```
rtyler: This function is only used once and does not have a corresponding unit test as far as I can tell. 

Non-blocking feedback, but IMHO one of the best reasons to have small functions like this is to have tests for them, otherwise 🤷

python/src/lib.rs
```diff
@@ -2733,13 +2721,116 @@ impl From<&PyTransaction> for Transaction {
...
+
+    serde_json::from_str(&value_json).map_err(|err| {
+        PyValueError::new_err(format!(
+            "CommitProperties.custom_metadata value for '{key}' must be valid JSON: {err}"
+        ))
+    })
+}
+
+fn validate_reserved_commit_metadata(key: &str, value: &Value) -> PyResult<()> {
+    match key {
```
rtyler, line 2787: This _feels_ redundant with the list of things in `kernel/transaction/mod.rs`. Since this is in Rust is there a strong reason not to keep all the type validation logic inside of `normalize_reserved_commit_metadata`? 

If this kind of refactoring is tedious/impossible/you-just-don't-wanna-do-it then I think some additional commentary on `RESERVED_COMMIT_BLAH` can reference this function as something to consider updating when that list is modified, just leave a sign post for future generations of sleep-deprived maintainers

ethan-tyler: good point, tracking this here #4457

crates/core/src/kernel/transaction/mod.rs
ethan-tyler: Removed the helper and inlined the loop

crates/core/src/kernel/transaction/mod.rs
ethan-tyler: Added rustdoc with the context for why this exists

crates/core/src/kernel/transaction/mod.rs
ethan-tyler: This makes sense, however I kept the helper for now. The structured fields (key, expected, actual) IMO are the important context here and inlining at each branch would duplicate the same log shape. If the file/line pointing at the helper turns out to be a problem in practice more than happy to inline later.

crates/core/src/kernel/transaction/mod.rs
ethan-tyler: Added a comment explaining these are fark compatible `CommitInfo` provenance fields (not protocol
required)and they exist to avoid duplicate flattened json keys, and must stay aligned with the python validator.

rtyler (approved):

PR #4465, opened 2026-05-19
feat: add generic OpenDAL storage backend (deltalake-opendal)

kszucs:
# Description

Adds a generic `deltalake-opendal` crate that plugs any [OpenDAL](https://opendal.apache.org/) service into delta-rs's storage registry — including the HuggingFace Hub.

- A single pair of generic factories (`OpendalObjectStoreFactory` / `OpendalLogStoreFactory`) parameterized by an `OpendalAdapter`; the built-in `GenericAdapter` covers bucket-rooted services with no per-service code.
- Each enabled service registers under an unambiguous `opendal+<service>://` scheme, plus its bare `<service>://` scheme when that doesn't collide with a native delta backend (so `hf://`, `gcs://`, … work directly, while `opendal+s3://` / `opendal+memory://` never clobber the native backends).
- Services are opt-in via `opendal-<service>` Cargo features; config is passed through `storage_options` using the `opendal.<key>` convention, forwarded verbatim to the operator.

# Usage

```python

storage_options = {
    "opendal.repo_type": "datasets",
    "opendal.repo_id": "my-org/my-dataset",
    "opendal.token": "hf_...",
}
write_deltalake("hf:///my_table", table, storage_options=storage_options)
dt = DeltaTable("hf:///my_table", storage_options=storage_options)
```

# Notes

- Supersedes the earlier `deltalake-hf` crate. The Parquet content-defined-chunking change is dropped from this PR and will be proposed separately.
- Docs: `docs/integrations/object-storage/opendal.md`.


.github/workflows/ci_hf_cdc.yml
```diff
@@ -0,0 +1,53 @@
...
+env:
+  SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
+
+jobs:
+  hf-cdc-python:
+    name: HuggingFace and CDC Python tests
+    runs-on: ubuntu-latest
+    # Skip entirely when HF secrets are not available (e.g. PRs from forks).
+    if: ${{ secrets.HF_TOKEN != '' }}
```
kszucs: Integration tests must be run against live huggingface so a user and a public bucket should be configured for running them.

crates/core/src/table/config.rs
```diff
@@ -119,6 +119,21 @@ pub enum TableProperty {
 
     /// 'classic' for classic Delta Lake checkpoints. 'v2' for v2 checkpoints.
     CheckpointPolicy,
+
+    /// true to enable Parquet content-defined chunking when writing data files.
+    /// CDC produces row groups whose boundaries are deterministic functions of the
+    /// data content, which improves dedupe and incremental upload efficiency on
+    /// content-addressable stores (e.g. HuggingFace Hub / Xet).
+    ParquetContentDefinedChunkingEnabled,
```
kszucs: Probably this is not the right place for these options since these are not part of the spec, trying to move them for `format.options`.

adamreeve:
This seems like a good use case for more flexible Parquet format configuration options, rather than delta-rs having to know about the CDC options itself. Eg. something like what was proposed in #3794 (although that was complicated by also dealing with read paths that go through Datafusion).

rtyler:
@kszucs thanks for the proposal! There are kind of two things joined together in this pull request that I will try to pull into distinct parts:

## Content-defined Chunking

Is this something specifically required for integration with the HuggingFace Hub? As @adamreeve alludes to we have had some previous discussions around how to wire parquet write/read options through in a generally useful/meaningful way It's not clear to me from the PR whether the CDC support is required, or desirable for other reasons

## HF/OpenDAL

Generally speaking I am not in favor of bringing in dependencies which require contributors and maintainers to sign up for external services. If HuggingFace Hub doesn't have a mock service or container that can be added to run tests against, this code cannot be enabled in this project.  Developers being able to locally run tests for the project is very important and I do not want to accept any maintenance burden for something I cannot actually test.

_However_ this is not the first time the orbit of OpenDAL and delta-rs have crossed paths. I am curious if the newly added crate were converted to a `deltalake-opendal` crate if that could wire through enough supporting infrastructure to allow any OponDAL supported service to exist? I think that might open the door to a lot of flexibility and might defuse the local development question around HuggingFace Hub specifically, since we would defer to OpenDAL

kszucs:
> @kszucs thanks for the proposal! There are kind of two things joined together in this pull request that I will try to pull into distinct parts:
Yes, these are orthogonal. 
> 
> ## Content-defined Chunking
> Is this something specifically required for integration with the HuggingFace Hub? 

CDC is a parquet writer feature present in arrow-cpp and its bindings (including pyarrow) and arrow-rs + datafusion and soon in iceberg. The feature itself is primarily used by Hugging Face but open to the ecosystem for reuse, see a prototype parquet page store https://github.com/apache/arrow-rs/pull/9661

> As @adamreeve alludes to we have had some previous discussions around how to wire parquet write/read options through in a generally useful/meaningful way It's not clear to me from the PR whether the CDC support is required, or desirable for other reasons

Sure, I'm open to whatever approach. I primarily wanted to open a PoC here for further discussions, going to chime in there.

> 
> ## HF/OpenDAL
> Generally speaking I am not in favor of bringing in dependencies which require contributors and maintainers to sign up for external services. If HuggingFace Hub doesn't have a mock service or container that can be added to run tests against, this code cannot be enabled in this project. Developers being able to locally run tests for the project is very important and I do not want to accept any maintenance burden for something I cannot actually test.

I can understand this decision , sadly (as of today) HF doesn't have a local testing shim, but this may change in the future. 

> 
> _However_ this is not the first time the orbit of OpenDAL and delta-rs have crossed paths. I am curious if the newly added crate were converted to a `deltalake-opendal` crate if that could wire through enough supporting infrastructure to allow any OponDAL supported service to exist? I think that might open the door to a lot of flexibility and might defuse the local development question around HuggingFace Hub specifically, since we would defer to OpenDAL

Sounds doable, essentially having an `object_store_opendal` based storage backend like the one in this PR but without hardcoding the HF operator. Also cc @Xuanwo

ion-elgreco:
The thing is with opendal not each integration has the same level of api support, so this is important to check (just fyi :))

rtyler (dismissed):
I am liking the OpenDAL generic crate. I am going to land this, but we will need to discuss what gets wired through into the Python wheels since that dependency may bloat up our wheel size or surface area in a way that is not as beneficial for all Python users.

kszucs:
Thanks @rtyler for the fixes! The auto-merge has failed so I updated the missing signoff in my commit messages as well as updated the HF related examples in the readme. Also rebased on top of the latest upstream.

kszucs:
The python wheel size has increased by 7.7MB on arm64 mac including 17 opendal backends. We could slim it down by disabling redundant ones like opendal-s3 or have an optional opendal extra for the python wheel.

ion-elgreco:
> The python wheel size has increased by 7.7MB on arm64 mac including 17 opendal backends. We could slim it down by disabling redundant ones like opendal-s3 or have an optional opendal extra for the python wheel.

Whats the size diff on disk unpacked in release mode?

ion-elgreco (approved):
Yeet

PR #4409, opened 2026-04-26
fix: listing catalogs from Unity

hjohnss6:
# Description
- Adds more types of catalogs.
- Adds a new type of ProvisioningState.
- Uses macro for formatting the ProvisioningState.

This should align with the Databricks API.

## Side notes (out of scope)
There are potentially more work that could be done to align the data model with the Databricks API.
For instance, `securable_type` is an enum, but treated as string in data model.

Some of the CatalogTypes are not really queriable (Foreign for instance). But this problem is not really introduced by this change.

# Related Issue(s)
- closes #4360 

# Documentation

- [Databricks API](https://docs.databricks.com/api/azure/workspace/catalogs/list)
- [Unity Catalog OSS](https://docs.unitycatalog.io/swagger-docs/#/Catalogs/)

PR #4468, opened 2026-05-20
Description  Update the deltalake-core version in other crates to get latest updates from recent releases.

jijosg:
# Description
Update the deltalake-core version in other crates to get latest updates from recent releases.

### DIFF
--- a/crates/aws/Cargo.toml
+++ b/crates/aws/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deltalake-aws"
-version = "0.15.0"
+version = "0.15.1"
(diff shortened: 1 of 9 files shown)


rtyler:
Were you seeing any particular errors? Cargo will already resolve the latest core with these version ranges from my understanding

jijosg:
Yeah, I realized that later. I was only interested in updating the catalog-unity crate to get #4409 to be part of deltalake-catalog-unity release.

rtyler:
@jijosg aha, I can help with that. I will close this with a follow up PR that just bumps the unity crate

PR #4469, opened 2026-05-20
fix: keep delta materialized filters above parquet

### LINKED ISSUE #4467: [Bug]: QueryBuilder: JOIN + WHERE-IN involving a partition column returns 0 rows
### What happened?

When two partitioned Delta tables are joined on the partition column and a `WHERE col IN (...)` filter is applied to that same column, `QueryBuilder.execute` returns 0 rows. The data is unchanged from a query that returns the expected rows — only the SQL shape changes. The bug only fires when at least one additional string column also appears in both the `JOIN ON` and the `WHERE IN`. Removing the partitioning, removing the `WHERE`, or wrapping the string join keys in `CAST(... AS VARCHAR)` each independently restores the rows.

The partition column is materialised as Arrow `Utf8View` after a Delta read (via `arro3.core`).

### **Observed behaviour**

With `partition_by=["scenario"]` on both sides, joining on `(scenario, price_zone, idx)` and filtering on `(scenario, price_zone)`:

- `SELECT COUNT(*) FROM left l JOIN right r ON ...`  →  **10,000** (expected)
- `SELECT COUNT(*) FROM left l JOIN right r ON ... WHERE l.scenario IN (...) AND l.price_zone IN (...)`  →  **0**
- `SELECT COUNT(*) FROM left l JOIN right r ON CAST(...) = CAST(...) AND ... WHERE ... IN (...)`  →  **10,000**

The `WHERE … IN (…)` lists every distinct value in each column, so the filter is a no-op at the row level.

Bisection results:

| `partition_by` | second string column in JOIN+WHERE | repro? |
|---|---|---|
| yes | yes | **yes** |
| yes | no | no |
| no | yes | no |
| no | no | no |

Writer properties (dictionary encoding, chunk-level statistics) were also bisected and are **not** part of the trigger.

### Expected behavior

The middle query should return the same number of rows as the other two.

### Operating System

Windows

### Binding

Python

### Bindings Version

1.5.1

### Steps to reproduce

```python
"""Repro: JOIN + WHERE-IN involving a Delta partition column returns 0 rows."""
from __future__ import annotations

import shutil
import tempfile
from pathlib import Path

import pyarrow as pa
from deltalake import QueryBuilder

SCENARIOS = [f"s{i}" for i in range(10)]
ZONES = [f"z{i}" for i in range(5)]
N = 200
EXPECTED = len(SCENARIOS) * len(ZONES) * N


def make_table(offset: int) -> pa.Table:
    s, z, idx = [], [], []
    for sc in SCENARIOS:
        for zn in ZONES:
            s += [sc] * N
            z += [zn] * N
            idx += list(range(N))
    return pa.table(
        {
            "scenario": pa.array(s, type=pa.string()),
            "price_zone": pa.array(z, type=pa.string()),
            "idx": pa.array(idx, type=pa.int64()),
            "val": pa.array(range(offset, offset + len(s)), type=pa.int64()),
        }
    )


def run(*, partition: bool) -> tuple[int, int, int]:
    tmp = Path(tempfile.mkdtemp())
    try:
        kw = {"partition_by": ["scenario"]} if partition else {}
        write_deltalake(str(tmp / "left"), make_table(0), **kw)
        write_deltalake(str(tmp / "right"), make_table(10**6), **kw)

        qb = QueryBuilder()
        qb = qb.register("left", DeltaTable(str(tmp / "left")))
        qb = qb.register("right", DeltaTable(str(tmp / "right")))

        def n(sql: str) -> int:
            return int(qb.execute(sql).read_all()["n"][0].as_py())

        s_in = ", ".join(repr(s) for s in SCENARIOS)
        z_in = ", ".join(repr(z) for z in ZONES)
        j = (
            "left l JOIN right r "
            "ON l.scenario = r.scenario AND l.price_zone = r.price_zone AND l.idx = r.idx"
        )
        jc = (
            "left l JOIN right r "
            "ON CAST(l.scenario AS VARCHAR) = CAST(r.scenario AS VARCHAR) "
            "AND CAST(l.price_zone AS VARCHAR) = CAST(r.price_zone AS VARCHAR) "
            "AND l.idx = r.idx"
        )
        w = f"WHERE l.scenario IN ({s_in}) AND l.price_zone IN ({z_in})"
        return (
            n(f"SELECT COUNT(*) n FROM {j}"),
            n(f"SELECT COUNT(*) n FROM {j} {w}"),
            n(f"SELECT COUNT(*) n FROM {jc} {w}"),
        )
    finally:
        shutil.rmtree(tmp, ignore_errors=True)


def main() -> None:
    print(f"expected per cell: {EXPECTED:,}\n")
    print(f"{'config':28}  {'JOIN':>6}  {'+WHERE':>6}  {'+CAST+WHERE':>11}")
    print(f"{'-' * 28}  {'-' * 6}  {'-' * 6}  {'-' * 11}")
    for partition in (True, False):
        a, b, c = run(partition=partition)
        tag = "partition_by=['scenario']" if partition else "no partitioning"
        print(f"{tag:28}  {a:>6,d}  {b:>6,d}  {c:>11,d}")


if __name__ == "__main__":
    main()
```

### Relevant logs

```shell
expected per cell: 10,000

config                          JOIN  +WHERE  +CAST+WHERE
----------------------------  ------  ------  -----------
partition_by=['scenario']     10,000       0       10,000   <-- WHERE collapses to 0
no partitioning               10,000  10,000       10,000   <-- baseline
```

ethan-tyler:
Fixes #4467.

`DefaultPhysicalExprAdapterFactory` rewrites nullable columns missing from the child schema to null literals. This is correct behavior for schema evolution but wrong for Delta materialized columns (partition values, file id, row index). Those columns are injected above the parquet child. Under join dynamic filters, rewriting a partition column filter against the parquet child and turns it into a null predicate and drops valid rows.

The fix here is to skip parquet child filter adaptation when a parent filter references a delta materialized column. Keep those filters bound to `DeltaScanExec` output.

### DIFF
--- a/crates/core/tests/it_datafusion/integration_datafusion.rs
+++ b/crates/core/tests/it_datafusion/integration_datafusion.rs
@@ -417,19 +417,25 @@ mod local {
         Ok(())
     }
 
-    const ISSUE_4467_SCENARIO_COUNT: usize = 3;
-    const ISSUE_4467_ZONE_COUNT: usize = 2;
+    const ISSUE_4467_SCENARIOS: &[&str] = &["s0", "s1", "s2"];
+    const ISSUE_4467_ZONES: &[&str] = &["z0", "z1"];
     const ISSUE_4467_ROWS_PER_ZONE: usize = 5;
-    const ISSUE_4467_FILTERED_SCENARIO_COUNT: usize = 2;
-    const ISSUE_4467_FILTERED_ZONE_COUNT: usize = 1;
-    const ISSUE_4467_FILTERED_IDX_COUNT: usize = 3;
-    const ISSUE_4467_FILTERED_EXPECTED_ROWS: i64 = (ISSUE_4467_FILTERED_SCENARIO_COUNT
-        * ISSUE_4467_FILTERED_ZONE_COUNT
-        * ISSUE_4467_ROWS_PER_ZONE) as i64;
+    const ISSUE_4467_FILTERED_SCENARIOS: &[&str] = &["s0", "s1"];
+    const ISSUE_4467_FILTERED_ZONES: &[&str] = &["z0"];
+    const ISSUE_4467_FILTERED_IDXS: &[i64] = &[0, 1, 2];
 
-    fn issue_4467_quoted_values(prefix: &str, count: usize) -> String {
-        (0..count)
-            .map(|value| format!("'{prefix}{value}'"))
+    fn issue_4467_quoted_values(values: &[&str]) -> String {
+        values
+            .iter()
+            .map(|value| format!("'{value}'"))
+            .collect::<Vec<_>>()
+            .join(", ")
+    }
+
+    fn issue_4467_int_values(values: &[i64]) -> String {
+        values
+            .iter()
+            .map(|value| value.to_string())
             .collect::<Vec<_>>()
             .join(", ")
     }
@@ -440,11 +446,11 @@ mod local {
         let mut idx = vec![];
         let mut values = vec![];
 
-        for scenario in 0..ISSUE_4467_SCENARIO_COUNT {
-            for zone in 0..ISSUE_4467_ZONE_COUNT {
+        for scenario in ISSUE_4467_SCENARIOS {
+            for zone in ISSUE_4467_ZONES {
                 for row_idx in 0..ISSUE_4467_ROWS_PER_ZONE {
-                    scenarios.push(format!("s{scenario}"));
-                    zones.push(format!("z{zone}"));
+                    scenarios.push((*scenario).to_string());
+                    zones.push((*zone).to_string());
                     idx.push(row_idx as i64);
                     values.push(offset + values.len() as i64);
                 }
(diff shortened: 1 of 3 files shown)


crates/core/tests/it_datafusion/integration_datafusion.rs
```diff
@@ -417,6 +417,150 @@ mod local {
         Ok(())
     }
 
+    const ISSUE_4467_SCENARIO_COUNT: usize = 3;
+    const ISSUE_4467_ZONE_COUNT: usize = 2;
+    const ISSUE_4467_ROWS_PER_ZONE: usize = 5;
+    const ISSUE_4467_FILTERED_SCENARIO_COUNT: usize = 2;
```
rtyler: What in the heck is this about 😆 😭 

Might be time for us to introduce [rstest](https://github.com/la10736/rstest)?  This is kinda wacky

ethan-tyler: lol 😆 what can I say- I am a wacky guy but fair point, the count/prefix style obscures my intent. These
are just fixture dimensions and not a parameter matrix. I don't think rstest is needed here. Will it help if I make the fixture explicit with named scenario/zone/idx value lists and derive expected counts from those?

rtyler (approved):

PR #4472, opened 2026-05-23
feat(python): propagate read_version to write conflict checking

### LINKED ISSUE #4417: [Feature]: Expose read snapshot to high-level Python operations (merge, update, delete)

Cross-engine read-modify-write patterns (e.g. read via DuckDB or Polars, write via delta-rs) cannot achieve snapshot isolation today because the high-level Python operations always commit against HEAD-at-execution-time, not against a user-specified read version. The Rust core already tracks read_version in its commit/conflict-check machinery, but the Python binding does not surface it — so a concurrent writer that lands a commit between the user's read and merge can silently invalidate the assumptions the merge was built on, without triggering a conflict.

This is the standard "lost update" problem for Delta. Single-engine merges (delta-rs reading and writing) are safe because the operation pins its own snapshot internally. The moment the read happens in another engine, that guarantee is gone.

Allow the high-level write operations to accept an optional read_version: int (or an opaque snapshot handle obtained from DeltaTable.version() / DeltaTable.snapshot()), which is threaded through to the commit's conflict check:

```
dt = DeltaTable("s3://bucket/table")
read_version = dt.version()

# ... read with DuckDB / Polars / Spark, build the change set ...

dt.merge(
    source=changes,
    predicate="t.id = s.id",
    source_alias="s",
    target_alias="t",
    read_version=read_version,        # NEW

# Same for:
dt.update(predicate=..., updates=..., read_version=read_version)
dt.delete(predicate=..., read_version=read_version)
write_deltalake(dt, df, mode="overwrite", predicate=..., read_version=read_version)
```
If a commit has landed at a version > read_version that conflicts under Delta's isolation rules, the operation should fail with a CommitFailedError / concurrency exception rather than silently committing on top.

Pin the read to a specific version in the reading engine, capture that version, and hope nothing has committed by the time the merge runs. DuckDB's Delta extension already supports pinning via ATTACH ... (TYPE delta, VERSION ...):

```
dt = DeltaTable(TABLE)
current_version = dt.version()
con.sql(f""" ATTACH '{TABLE}' AS scada (TYPE delta, VERSION {current_version}); """)
result = con.sql(f""" SELECT * FROM scada limit 10 """).arrow()

# ... build change set from `result` ...

# PROBLEM: this merge does not know about `current_version`. If another writer
# committed between the ATTACH above and this execute(), the merge commits
# on top of the new HEAD anyway — no conflict, no error, silently wrong.
dt.merge(
    source=changes,
    predicate="t.id = s.id",
    source_alias="s",
    target_alias="t",
```
The reader side is solved (DuckDB pins cleanly). The writer side is the gap: there is no way to tell merge / update / delete "this plan is only valid against version N, abort if HEAD has moved into a conflicting state." Only the Rust core can close that window properly, since it already does so for single-engine merges.

Surfacing read_version on the Python operations would make this safe by reusing the conflict-check logic the Rust core already implements, and would make delta-rs a first-class writer for multi-engine pipelines.

### Priority

None

### Additional context

_No response_

### Contribution

felipemchdev:
Closes #4417

Motivation

In cross-engine read-modify-write flows (for example: read in DuckDB/Polars/Spark, write in delta-rs), callers may want write-side conflict validation to operate against the snapshot originally used to build the change set.

While investigating #4417, I found that DeltaTable(version=N) already propagates pinned OCC semantics for operations like merge, update, and delete. So this PR is not introducing an entirely new transaction model or broad new OCC semantics.

Instead, this change focuses on:

exposing explicit read_version ergonomics consistently across the high-level Python write APIs
making snapshot-pinned behavior more explicit/documented at the API level
enabling pinned OCC validation for write_deltalake(uri, ...) flows where there is currently no explicit way to propagate the original read snapshot without reconstructing a pinned DeltaTable handle manually

The implementation intentionally reuses the existing delta-rs optimistic concurrency/conflict-checking behavior.

Approach

This PR adds optional read_version: int | None = None to:

DeltaTable.merge(...)
DeltaTable.update(...)
DeltaTable.delete(...)
write_deltalake(...)

read_version is threaded through the existing PyO3 bindings into the current Rust operation builders/conflict-checking paths.

Implementation details
operations clone the table and load the requested version on the clone before building the transaction
the original Python DeltaTable state is not mutated by snapshot pinning
log_store handling was adjusted to preserve pinned-table semantics in the write path
merge snapshots are pinned at builder creation time rather than execute() time

This change intentionally reuses existing delta-rs OCC semantics rather than introducing new conflict resolution behavior.

Tests

Validated with targeted repros and broader regression coverage:

stale read conflict scenarios
merge delayed execute / builder snapshot pinning
sequential read_version flows
invalid read_version handling
concurrent append conflict scenarios
metadata preservation
regression coverage for legacy behavior (read_version=None)

Additional verification included comparisons against upstream/main to validate existing OCC behavior for version-pinned DeltaTable handles.

cargo check -p deltalake-python passed.

The broader Python suite also passed aside from the existing Windows forkserver incompatibility in multiprocessing-related tests.

### DIFF
--- a/python/deltalake/_internal.pyi
+++ b/python/deltalake/_internal.pyi
@@ -213,6 +213,7 @@ class RawDeltaTable:
         writer_properties: WriterProperties | None,
         commit_properties: CommitProperties | None,
         post_commithook_properties: PostCommitHookProperties | None,
+        read_version: int | None = None,
     ) -> str: ...
     def repair(
         self,
@@ -228,6 +229,7 @@ class RawDeltaTable:
         safe_cast: bool,
         commit_properties: CommitProperties | None,
         post_commithook_properties: PostCommitHookProperties | None,
+        read_version: int | None = None,
     ) -> str: ...
     def create_merge_builder(
         self,
@@ -244,6 +246,7 @@ class RawDeltaTable:
         streamed_exec: bool,
         max_spill_size: int | None,
         max_temp_directory_size: int | None,
+        read_version: int | None = None,
     ) -> PyMergeBuilder: ...
     def merge_execute(self, merge_builder: PyMergeBuilder) -> str: ...
     def get_active_partitions(
@@ -295,6 +298,7 @@ class RawDeltaTable:
         writer_properties: WriterProperties | None,
         commit_properties: CommitProperties | None,
         post_commithook_properties: PostCommitHookProperties | None,
+        read_version: int | None = None,
     ) -> None: ...
 
 def rust_core_version() -> str: ...
@@ -329,6 +333,7 @@ def write_to_deltalake(
     writer_properties: WriterProperties | None,
     commit_properties: CommitProperties | None,
     post_commithook_properties: PostCommitHookProperties | None,
+    read_version: int | None = None,
 ) -> None: ...
 def convert_to_deltalake(
     uri: str,
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -707,6 +707,7 @@ class DeltaTable:
         *args: Any,
         commit_properties: CommitProperties | None = None,
         post_commithook_properties: PostCommitHookProperties | None = None,
+        read_version: int | None = None,
     ) -> dict[str, Any]:
         """`UPDATE` records in the Delta Table that matches an optional predicate. Either updates or new_values needs
         to be passed for it to execute.
@@ -719,6 +720,7 @@ class DeltaTable:
             error_on_type_mismatch: specify if update returns an error when update expressions fail to cast to target column types :default = True
             commit_properties: properties of the transaction commit. If None, default values are used.
             post_commithook_properties: properties for the post commit hook. If None, default values are used.
+            read_version: version of the table that was read to build the update. Will fail if the table has been updated since then in a way that conflicts with the update.
         Returns:
             the metrics from update
 
@@ -807,7 +809,10 @@ class DeltaTable:
             safe_cast=not error_on_type_mismatch,
             commit_properties=commit_properties,
             post_commithook_properties=post_commithook_properties,
+            read_version=read_version,
         )
+        if read_version is not None:
+            self.update_incremental()
         return json.loads(metrics)
 
     @property
@@ -847,6 +852,7 @@ class DeltaTable:
         *args: Any,
         commit_properties: CommitProperties | None = None,
         post_commithook_properties: PostCommitHookProperties | None = None,
+        read_version: int | None = None,
     ) -> TableMerger:
         """Pass the source data which you want to merge on the target delta table, providing a
         predicate in SQL query like format.
@@ -872,6 +878,7 @@ class DeltaTable:
             max_temp_directory_size: The maximum disk space for temporary spill files. If not specified, uses DataFusion's default.
             commit_properties: properties for the commit. If None, default values are used.
             post_commithook_properties: properties for the post commit hook. If None, default values are used.
+            read_version: version of the table that was read to build the merge. The commit fails with :class:`~deltalake.exceptions.CommitFailedError` if a concurrent writer committed a conflicting change after this version.
 
         Returns:
             TableMerger: TableMerger Object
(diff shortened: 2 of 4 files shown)


ion-elgreco:
@felipemchdev I don't think you nees this because you can do pass version directly in Delta table and then those actions will be done against that version

felipemchdev:
> @felipemchdev I don't think you nees this because you can do pass version directly in Delta table and then those actions will be done against that version

Interesting, I genuinely did not realize the version-pinned DeltaTable path already propagated into the merge OCC flow like this.

I was mostly approaching the issue from the explicit `read_version` API angle for cross-engine read/modify/write flows, especially since the original issue described the Python mutation APIs as always validating against HEAD-at-execution-time.

From the example which @djouallah made at the issue page, it looks like constructing the table with `version=vB` already carries the pinned snapshot into the conflict checker path internally.

So the PR may be solving something narrower than I originally thought, or making the semantics more explicit/consistent across the high-level APIs rather than introducing entirely new behavior.

I'll dig a bit more into the existing merge/write paths and compare the exact behavior against the scenarios from #4417.

felipemchdev:
So... i did a deeper pass comparing upstream/main vs this branch.

It turns out merge/update/delete already inherit pinned OCC semantics when operating on a DeltaTable(version=N) handle. I had initially assumed the Python mutation APIs always validated against HEAD-at-execution-time, based on the original issue framing.

So for table-handle operations, this PR is mostly providing:

* explicit API consistency
* ergonomics around read_version propagation
* clearer/stabler semantics across the high-level write APIs

The one case that does appear to introduce genuinely new behavior is:

write_deltalake(uri, ..., read_version=...)

since URI-based writes currently reload HEAD state internally and have no explicit way to propagate stale-read validation semantics without reconstructing a pinned DeltaTable object manually.

I also verified the behavior directly against upstream/main with targeted repros for:

* stale snapshot conflicts
* merge builder timing
* write_deltalake(table) vs write_deltalake(uri)
* explicit read_version flows

I'll probably narrow the PR framing/scope accordingly so the semantic delta is more accurate and easier to review.

ion-elgreco:
> So... i did a deeper pass comparing upstream/main vs this branch.
> 
> It turns out merge/update/delete already inherit pinned OCC semantics when operating on a DeltaTable(version=N) handle. I had initially assumed the Python mutation APIs always validated against HEAD-at-execution-time, based on the original issue framing.
> 
> So for table-handle operations, this PR is mostly providing:
> 
> * explicit API consistency
> * ergonomics around read_version propagation
> * clearer/stabler semantics across the high-level write APIs
> 
> The one case that does appear to introduce genuinely new behavior is:
> 
> write_deltalake(uri, ..., read_version=...)
> 
> since URI-based writes currently reload HEAD state internally and have no explicit way to propagate stale-read validation semantics without reconstructing a pinned DeltaTable object manually.
> 
> I also verified the behavior directly against upstream/main with targeted repros for:
> 
> * stale snapshot conflicts
> * merge builder timing
> * write_deltalake(table) vs write_deltalake(uri)
> * explicit read_version flows
> 
> I'll probably narrow the PR framing/scope accordingly so the semantic delta is more accurate and easier to review.
> 

I don't believe the ergonomics are worth it. Load_at_version or Delta table(version=) are the semantics, we don't want another loading pattern, it gets confusing

felipemchdev:
> I don't believe the ergonomics are worth it. Load_at_version or Delta table(version=) are the semantics, we don't want another loading pattern, it gets confusing

Yeah, that's fair.

I agree "DeltaTable(version=...)" / "load_as_version()" already cover most of the semantics I thought were missing.

The URI-based "write_deltalake(...)" case still felt useful to me, but I understand the concern about introducing another snapshot-loading pattern into the API.

I'll rethink whether this is better handled through docs/tests around the existing behavior instead.

felipemchdev:
Closing the PR , since the issue was closed. Will made another PR in the future just for "write_deltalake(...)" case.

PR #4473, opened 2026-05-23
fix: preserve merge row ordinals across scan planning

### LINKED ISSUE #4471: [Bug]: Bug: deltalake 1.6.0 false-positive MERGE error "matched a target row with multiple source rows" on clean deduplicated data
### Describe the bug
There is a critical regression in `deltalake 1.6.0`'s `MERGE` validation logic (driven by the underlying Apache DataFusion engine). When running a conditional merge over an existing single-file table partition, the engine falsely triggers an ambiguous match error, even when the source dataframe and target table are perfectly unique and deduplicated on the join keys. 

This script executes perfectly on `deltalake==1.5.1` / `pyarrow==18.1.0` but reliably fails on `deltalake==1.6.0` / `pyarrow==24.0.0`.

### Environment
* **OS:** Linux / Ubuntu
* **Python Version:** 3.12
* **Working Stack:** `deltalake==1.5.1`, `pyarrow==18.1.0`, `pandas==3.0.3`
* **Broken Stack:** `deltalake==1.6.0`, `pyarrow==24.0.0`, `pandas==3.0.3`

### Minimal Reproducible Example (`repro.py`)
```python
import os
from pathlib import Path
import shutil
import pandas as pd

DB_PATH = Path("/tmp/repro_delta_bug")
if DB_PATH.exists():
    if DB_PATH.is_dir():
        shutil.rmtree(DB_PATH)
    else:
        os.remove(DB_PATH)

SENTINEL_TIME = "2261-12-31T23:59:59"
TS_V1 = "2026-05-13T11:27:43"
TS_V2 = "2026-05-13T11:31:38"

print("--- Step 1: Initializing Table with V1 State ---")
num_rows = 25000 

df_v1 = pd.DataFrame({
    "CustomerId": [f"C-{i}" for i in range(num_rows)],
    "MetricValue": [100.0] * num_rows,
    "ingestion_datetime": pd.to_datetime([TS_V1] * num_rows),
    "valid_from_dt": pd.to_datetime([TS_V1] * num_rows),
    "valid_to_dt": pd.to_datetime([SENTINEL_TIME] * num_rows)
})

write_deltalake(DB_PATH, df_v1, mode="overwrite")
dt = DeltaTable(DB_PATH)
print(f"Initialized table with {len(df_v1)} records.")

print("\n--- Step 2: Running a subsequent Merge on unique clean data ---")
df_v2 = df_v1.copy()
df_v2["ingestion_datetime"] = pd.to_datetime(TS_V2)
df_v2["valid_from_dt"] = pd.to_datetime(TS_V2)

# Alter values for a slice of the dataset to force an SCD2 mutation state
df_v2.loc[df_v2.index % 2 == 0, "MetricValue"] = 200.0

predicate_v2 = (
    "target.CustomerId = source.CustomerId "
    f"AND target.valid_to_dt = arrow_cast('{SENTINEL_TIME}', 'Timestamp(Microsecond, None)') "
    f"AND target.ingestion_datetime < arrow_cast('{TS_V2}', 'Timestamp(Microsecond, None)')"
)

print(f"Source uniqueness verification: {df_v2.duplicated(subset=['CustomerId']).sum()} duplicate rows found.")

dt.merge(
    source=df_v2,
    predicate=predicate_v2,
    source_alias="source",
    target_alias="target"
).when_matched_update(
    updates={"valid_to_dt": f"arrow_cast('{TS_V2}', 'Timestamp(Microsecond, None)')"},
    predicate="target.MetricValue IS DISTINCT FROM source.MetricValue"
).execute()

print("Merge completed successfully.")

### Expected behavior

The merge should complete successfully because the source payload and target table current records are completely unique on the join/business key column (`CustomerId`) - SCD Type 2 lookup pattern.

### Operating System

Linux

### Binding

Python

### Bindings Version

1.6.0

### Steps to reproduce

For deltalake 1.6.0 the execution fails inside Step 2 throwing:
_internal.DeltaError: Generic DeltaTable error: External error: Generic DeltaTable error: MERGE matched a target row with multiple source rows that satisfy duplicate relevant WHEN MATCHED clauses

Note, on deltalake 1.5.1 it terminates without issues.

### Relevant logs

```shell

```

### LINKED ISSUE #4475: [Bug]: MERGE raises "multiple source rows matched" when target partition file exceeds DataFusion's batch size (8,192 rows)
### What happened?

`DeltaTable.merge().when_matched_update_all().when_not_matched_insert_all().execute()` raises a `DeltaError` when the target partition file contains more than 8,192 rows. The source data has zero duplicate keys, every (datetime, scenario, price_zone) triple is unique, so each target row should match at most one source row. The error does not occur at 8,192 rows; it fires at exactly 8,193. 8,192 is DataFusion's compiled-in default target_batch_size, suggesting the bug is triggered when a parquet file is read in more than one batch during the merge.

Reducing `WriterProperties(max_row_group_size=N)` shifts the failure threshold, and the number of partitions in the target table appears to influence where exactly it triggers. I could not find a good way to reproduce the exact threshold, so I left it out, but thought it worth mentioning.

### Expected behavior

Both cases should succeed and produce identical results.

### Operating System

Windows

### Binding

Python

### Bindings Version

1.6.0

### Steps to reproduce

```python
import sys
import tempfile
import traceback
from pathlib import Path

import deltalake
import numpy as np
import pyarrow as pa
import pyarrow.compute as pc
from deltalake._internal import DeltaError

_rng = np.random.default_rng(0)
DATAFUSION_BATCH_SIZE = 8_192


def _make_tables(n: int) -> tuple[pa.Table, pa.Table]:
    """Target: two partitions (A, B), n unique rows each.
    Source: partition A only, same keys, values all 1.0.
    Zero duplicate keys in either table.
    """
    start_us = 1_704_067_200 * 1_000_000  # 2024-01-01 00:00:00 UTC in microseconds
    dts = pa.array(
        np.arange(start_us, start_us + n * 3_600_000_000, 3_600_000_000, dtype=np.int64),
        type=pa.timestamp("us"),
    )
    target = pa.table({
        "datetime":   pa.chunked_array([dts, dts]).combine_chunks(),
        "scenario":   pa.array(["A"] * n + ["B"] * n),
        "price_zone": pa.array(["zone_0"] * n * 2),
        "value":      pa.array(_rng.random(n * 2)),
    })
    source = pa.table({
        "datetime":   dts,
        "scenario":   pa.array(["A"] * n),
        "price_zone": pa.array(["zone_0"] * n),
        "value":      pa.array(np.ones(n)),
    })
    return target, source


def _run_merge(n: int) -> None:
    target, source = _make_tables(n)
    with tempfile.TemporaryDirectory() as tmp:
        table_path = str(Path(tmp) / "tbl")
        write_deltalake(table_path, target, partition_by=["scenario"])
        dt = DeltaTable(table_path)
        dt.merge(
            source=source,
            predicate=(
                "target.datetime = source.datetime"
                " AND target.scenario = source.scenario"
                " AND target.price_zone = source.price_zone"
            ),
            source_alias="source",
            target_alias="target",
        dt.update_incremental()

        result = dt.to_pyarrow_table()
        if len(result) != n * 2:
            msg = f"Expected {n * 2} rows, got {len(result)}"
            raise AssertionError(msg)
        a_values = result.filter(pc.equal(result["scenario"], "A"))["value"]
        if not pc.all(pc.equal(a_values, 1.0)).as_py():
            msg = "Partition A was not correctly updated"
            raise AssertionError(msg)


def main() -> None:
    print(f"deltalake {deltalake.__version__}\n")
    failed = False
    for n in [DATAFUSION_BATCH_SIZE, DATAFUSION_BATCH_SIZE + 1]:
        print(f"--- rows/partition = {n} ---")
        try:
            _run_merge(n)
            print("  OK\n")
        except DeltaError:
            print("  FAIL — expected merge to succeed, got:\n")
            traceback.print_exc(file=sys.stdout)
            failed = True
            print()
    sys.exit(1 if failed else 0)


if __name__ == "__main__":
    main()
```

### Relevant logs

```shell
deltalake 1.6.0

--- rows/partition = 8192 ---
  OK

--- rows/partition = 8193 ---
  FAIL — expected merge to succeed, got:

  File "c:\repos\mt-fumo-blob\reproduce_merge_bug copy.py", line 117, in main
    _run_merge(n)
  File "c:\repos\mt-fumo-blob\reproduce_merge_bug copy.py", line 93, in _run_merge
                                                              ^^^^^^^^^
  File "c:\repos\mt-fumo-blob\.venv\Lib\site-packages\deltalake\table.py", line 1862, in execute
    metrics = self._table.merge_execute(self._builder)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_internal.DeltaError: Generic DeltaTable error: External error: Generic DeltaTable error: MERGE matched a target row with multiple source rows that satisfy duplicate relevant WHEN MATCHED clauses (file: file:///C:/Users/nicsc/AppData/Local/Temp/tmprsen3k6d/tbl/scenario=A/part-00000-093a13f9-36c1-4d8a-aa0f-127191e83a01-c000.snappy.parquet, row ordinal in file: 1)
```

ethan-tyler:
Fixes #4471.

Retained row index scans synthesize row ordinals with per stream counters, which only works if all rows for a file flow through one stream. DataFusion can insert a repartition below `DeltaScanExec`
and split a file across partitions, causing colliding ordinals and duplicate match errors or dropped target rows.

To fix this we retain row index `DeltaScanExec` plans and declare `SinglePartition` input distribution which blocks DataFusion from repartition underneath them. Also adds a runtime guard for directly constructed plans that bypass the optimizer.

we can rework this once we have physical row identities from the source

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
@@ -21,8 +21,8 @@ use datafusion::common::{
     ColumnStatistics, HashMap, internal_datafusion_err, internal_err, plan_err,
 };
 use datafusion::execution::{RecordBatchStream, SendableRecordBatchStream, TaskContext};
-use datafusion::physical_expr::EquivalenceProperties;
 use datafusion::physical_expr::utils::collect_columns;
+use datafusion::physical_expr::{Distribution, EquivalenceProperties};
 use datafusion::physical_plan::execution_plan::{CardinalityEffect, PlanProperties};
 use datafusion::physical_plan::filter_pushdown::{FilterDescription, FilterPushdownPhase};
 use datafusion::physical_plan::metrics::{BaselineMetrics, ExecutionPlanMetricsSet, MetricsSet};
@@ -272,6 +272,15 @@ impl ExecutionPlan for DeltaScanExec {
         vec![&self.input]
     }
 
+    fn required_input_distribution(&self) -> Vec<Distribution> {
+        if self.scan_plan.contract.retained_row_index_field().is_some() {
+            // Retained row indexes depend on one stream seeing each file's rows.
+            vec![Distribution::SinglePartition]
+        } else {
+            vec![Distribution::UnspecifiedDistribution]
+        }
+    }
+
     // TODO: setting this will fail certain tests, but why
     // fn maintains_input_order(&self) -> Vec<bool> {
     //     vec![true]
@@ -300,8 +309,8 @@ impl ExecutionPlan for DeltaScanExec {
         config: &ConfigOptions,
     ) -> Result<Option<Arc<dyn ExecutionPlan>>> {
         if self.scan_plan.contract.retained_row_index_field().is_some() {
-            // DeltaScanStream stores row ordinal counters per execution partition. Repartitioning
-            // can split a file's rows across streams and break ordinal contiguity.
+            // Each DeltaScanStream keeps row ordinal counters for one execution partition.
+            // Repartitioning can split one file across streams and break ordinal contiguity.
             return Ok(None);
         }
 
@@ -320,6 +329,17 @@ impl ExecutionPlan for DeltaScanExec {
         partition: usize,
         context: Arc<TaskContext>,
     ) -> Result<SendableRecordBatchStream> {
+        // Normal planning enforces this through EnforceDistribution. Keep this check for
+        // callers that build DeltaScanExec directly or replace its child plan.
+        if self.scan_plan.contract.retained_row_index_field().is_some() {
+            let input_partition_count = self.input.properties().partitioning.partition_count();
+            if input_partition_count > 1 {
+                return plan_err!(
+                    "DeltaScanExec retained row indexes require a single input partition, got {input_partition_count}"
+                );
+            }
+        }
+
         Ok(Box::pin(DeltaScanStream {
             scan_plan: Arc::clone(&self.scan_plan),
             kernel_type: Arc::clone(self.scan_plan.scan.logical_schema()).into(),
@@ -747,9 +767,11 @@ mod tests {
         physical_expr::expressions::{
             BinaryExpr, Column, DynamicFilterPhysicalExpr, lit as physical_lit,
         },
+        physical_expr::{Distribution, Partitioning},
         physical_plan::{
             PhysicalExpr, collect, collect_partitioned,
             filter_pushdown::{FilterPushdownPhase, PushedDown},
+            repartition::RepartitionExec,
         },
         prelude::{col, lit},
         scalar::ScalarValue,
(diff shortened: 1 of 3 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -272,6 +272,15 @@ impl ExecutionPlan for DeltaScanExec {
         vec![&self.input]
     }
 
+    fn required_input_distribution(&self) -> Vec<Distribution> {
+        if self.scan_plan.contract.retained_row_index_field().is_some() {
+            // Retained row indexes depend on one stream seeing each file's rows.
+            vec![Distribution::SinglePartition]
```
ion-elgreco, line 278: What are the performance implications, since datafusion planner seemed to have wanted to repartition a file into multiple partitions before

ethan-tyler: Yes that's the biggest trade off here but required for correctness. I will fire up @delta-force-bot to do some benchmarking.

Good news is that datafusion is working hard on getting the primitives to enable physical row identities. https://github.com/apache/datafusion/issues/20135

rtyler (approved):

PR #4479, opened 2026-05-25
fix: simplify expressions when conflict checking

### LINKED ISSUE #4478: [Bug]: Conflict checking crashes when using a predicate with a timestamp expression
### What happened?

When conflict checking a concurrent operation that involves an expression with a timestamp typed column, an error is returned:
```
unexpected error: Transaction failed: Failed to commit transaction: Error evaluating predicate: Generic DeltaTable error: Internal error: arrow_cast should have been simplified to cast.
```

This looks very similar to #4093, except the fix for that only handled the delete operation and not the conflict checking.

### Expected behavior

No error except when there is a legitimate conflict.

### Operating System

Linux

### Binding

Rust

### Bindings Version

main

### Steps to reproduce

Repro as a test case that can be added to `crates/core/src/operations/delete.rs`. This fails as the error message is different to what is expected:

```rust
const START_TIMESTAMP_US: i64 = 1_704_067_200 * 1_000_000; // 2024-01-01 00:00:00 UTC
const DAY_US: i64 = 60 * 60 * 24 * 1_000_000;

async fn setup_timestamp_table() -> DeltaResult<(DeltaTable, Arc<Schema>)> {
    use arrow::array::TimestampMicrosecondArray;

    let table: DeltaTable = DeltaTable::new_in_memory()
        .create()
        .with_column(
            "timestamp",
            DeltaDataType::Primitive(PrimitiveType::Timestamp),
            true,
            None,
        )
        .await?;

    let arrow_schema = Arc::new(Schema::new(vec![Field::new(
        "timestamp",
        DataType::Timestamp(arrow_schema::TimeUnit::Microsecond, Some("UTC".into())),
        true,
    )]));

    let batch = RecordBatch::try_new(
        arrow_schema.clone(),
        vec![Arc::new(
            TimestampMicrosecondArray::from(vec![
                START_TIMESTAMP_US,
                START_TIMESTAMP_US + DAY_US,
                START_TIMESTAMP_US + 2 * DAY_US,
            ])
            .with_timezone("UTC"),
        )],
    )?;
    let table = table.write(vec![batch]).await?;

    Ok((table, arrow_schema))
}

/// Ensure conflict checking works when a delete operation with a timestamp predicate follows
/// a concurrent write operation. This requires arrow_cast expressions to be simplified.
#[tokio::test]
async fn test_concurrent_delete_with_timestamp_predicate_after_write() -> DeltaResult<()> {
    use arrow::array::TimestampMicrosecondArray;
    let (table, arrow_schema) = setup_timestamp_table().await?;

    // Clone the table to make two concurrent operations
    let table_for_op1 = table.clone();
    let table_for_op2 = table;

    // Successful commit: adds new data
    let new_batch = RecordBatch::try_new(
        arrow_schema.clone(),
        vec![Arc::new(
            TimestampMicrosecondArray::from(vec![
                START_TIMESTAMP_US + 3 * DAY_US,
                START_TIMESTAMP_US + 4 * DAY_US,
            ])
            .with_timezone("UTC"),
        )],
    )?;

    let table_after_append = table_for_op1
        .write(vec![new_batch])
        .await
        .expect("Failed to append new data");
    assert_eq!(table_after_append.version(), Some(2));

    // Concurrent commit that deletes a range of data.
    let result = table_for_op2
        .delete()
        .with_predicate("timestamp >= '2024-01-02T00:00:00Z'")
        .await;

    let err = result.expect_err(
        "expected concurrent delete with a timestamp predicate to fail conflict checking",
    );
    assert!(
        err.to_string()
            .contains("concurrent transactions added new data"),
        "unexpected error: {err}",
    );
    Ok(())
}
```

### Relevant logs

```shell

```

adamreeve:
# Description

Uses `Expression::resolve` when evaluating predicates during conflict resolution to fix `arrow_cast should have been simplified to cast` errors.

Both `files_matching_predicate` and `predicate_matches` needed to be updated to handle different code paths for the current operation and for previous operations (via `check_for_added_files_that_should_have_been_read_by_current_txn` and `check_for_deleted_files_against_current_txn_read_files`).

# Related Issue(s)

- closes #4478

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -266,9 +266,12 @@ pub(crate) fn files_matching_predicate<'a>(
     if let Some(Some(predicate)) =
         (!filters.is_empty()).then_some(conjunction(filters.iter().cloned()))
     {
-        let expr = SessionContext::new()
-            .create_physical_expr(predicate, &log_data.read_schema().to_dfschema()?)?;
-        let pruning_predicate = PruningPredicate::try_new(expr, log_data.read_schema())?;
+        let session: SessionContext = create_session().into();
+        let schema = log_data.read_schema();
+        let df_schema = Arc::new(schema.clone().to_dfschema()?);
+        let resolved = Expression::from(predicate).resolve(&session.state(), df_schema.clone())?;
+        let expr = session.create_physical_expr(resolved, &df_schema)?;
+        let pruning_predicate = PruningPredicate::try_new(expr, schema)?;
         let mask = pruning_predicate.prune(&log_data)?;
 
         Ok(Either::Left(log_data.into_iter().zip(mask).filter_map(
--- a/crates/core/src/kernel/transaction/state.rs
+++ b/crates/core/src/kernel/transaction/state.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::sync::Arc;
 
 use arrow_array::{ArrayRef, BooleanArray};
 use arrow_schema::{DataType as ArrowDataType, SchemaRef as ArrowSchemaRef};
@@ -8,7 +9,9 @@ use datafusion::execution::context::SessionContext;
 use datafusion::logical_expr::Expr;
 use datafusion::physical_optimizer::pruning::{PruningPredicate, PruningStatistics};
 
-use crate::delta_datafusion::{get_null_of_arrow_type, to_correct_scalar_value};
+use crate::delta_datafusion::{
+    Expression, create_session, get_null_of_arrow_type, to_correct_scalar_value,
+};
 use crate::errors::DeltaResult;
 use crate::kernel::{Add, EagerSnapshot};
 
@@ -85,9 +88,10 @@ impl<'a> AddContainer<'a> {
     /// so evaluating expressions is inexact. However, excluded files are guaranteed (for a correct log)
     /// to not contain matches by the predicate expression.
     pub fn predicate_matches(&self, predicate: Expr) -> DeltaResult<impl Iterator<Item = &Add>> {
-        //let expr = logical_expr_to_physical_expr(predicate, &self.schema);
-        let expr = SessionContext::new()
-            .create_physical_expr(predicate, &self.schema.clone().to_dfschema()?)?;
+        let session: SessionContext = create_session().into();
+        let df_schema = Arc::new(self.schema.clone().to_dfschema()?);
+        let resolved = Expression::from(predicate).resolve(&session.state(), df_schema.clone())?;
+        let expr = session.create_physical_expr(resolved, df_schema.as_ref())?;
         let pruning_predicate = PruningPredicate::try_new(expr, self.schema.clone())?;
         Ok(self
             .inner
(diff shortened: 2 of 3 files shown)


crates/core/src/delta_datafusion/mod.rs
```diff
@@ -266,9 +266,12 @@ pub(crate) fn files_matching_predicate<'a>(
     if let Some(Some(predicate)) =
         (!filters.is_empty()).then_some(conjunction(filters.iter().cloned()))
     {
-        let expr = SessionContext::new()
-            .create_physical_expr(predicate, &log_data.read_schema().to_dfschema()?)?;
-        let pruning_predicate = PruningPredicate::try_new(expr, log_data.read_schema())?;
+        let session: SessionContext = create_session().into();
```
adamreeve, line 269: I've also changed `SessionContext::new()` to `create_session()` here and below, which isn't strictly needed but seemed like the right thing to do to match conventions elsewhere. Let me know if you'd rather I keep this as it was.

ethan-tyler:
Will review today

crates/core/src/operations/delete.rs
```diff
@@ -2641,4 +2641,122 @@ mod tests {
...
+        )?;
+        let table = table.write(vec![batch]).await?;
+
+        Ok((table, arrow_schema))
+    }
+
+    /// Ensure conflict checking works when a delete operation with a timestamp predicate follows
+    /// a concurrent write operation. This requires arrow_cast expressions to be simplified.
+    #[tokio::test]
+    async fn test_concurrent_delete_with_timestamp_predicate_after_write() -> DeltaResult<()> {
```
ethan-tyler, line 2686: I like this test and conflict case a lot. Make me think a similar test the concurrent append is outside the delete predicate and the delete commits successfully. This would prove the simplification preserves pruning and doesn't turn timestamp predicates into false positive conflicts.

crates/core/src/kernel/transaction/state.rs
```diff
@@ -85,9 +88,10 @@ impl<'a> AddContainer<'a> {
     /// so evaluating expressions is inexact. However, excluded files are guaranteed (for a correct log)
     /// to not contain matches by the predicate expression.
     pub fn predicate_matches(&self, predicate: Expr) -> DeltaResult<impl Iterator<Item = &Add>> {
-        //let expr = logical_expr_to_physical_expr(predicate, &self.schema);
-        let expr = SessionContext::new()
-            .create_physical_expr(predicate, &self.schema.clone().to_dfschema()?)?;
+        let session: SessionContext = create_session().into();
```
ethan-tyler, line 91: I like this and it also mirrors the fix in `files_matching_predicate`. Something to think about for longer term maintenance is maybe extracting the repeated sequence into a helper or documenting how it relates to `table_provider::simplify_expr` allowing future conflict checking code to not accidentally skip simplification.

ethan-tyler (approved):
Thanks @adamreeve ! This looks good and appreciate the work, left a couple of thinking out loud thoughts.

crates/core/src/operations/delete.rs
adamreeve: Good point, I'll follow up and add that as an extra test case.

crates/core/src/kernel/transaction/state.rs
adamreeve: Ah yeah I used `Expression::resolve` rather than `table_provider::simplify_expr` to match the behaviour of the `Delete` evaluation. It looks like `resolve` does a bit more than `simplify_expr` does, like calling `coerce_predicate_literals` for example, so this consistency seemed important, but I haven't looked into where exactly that difference would matter.

PR #4483, opened 2026-05-26
feat: remove S3DynamoDbLogStore and prepare for 1.0.0

rtyler:
This pull request changes the versions of all our crates to 1.0 in preparation
for the biggest most awesome release of 2026. Only four years late!

The hallmark change that I'm wanting to make here is the removal of
S3DynamoDbLogStore which has outlived its useful life as the project moves
towards coordinate commits v2


crates/aws/src/lib.rs
```diff
@@ -59,18 +35,13 @@ impl LogStoreFactory for S3LogStoreFactory {
     ) -> DeltaResult<Arc<dyn LogStore>> {
         let s3_options = self.with_env_s3(&options.raw.clone());
         let s3_options = S3StorageOptions::from_map(&s3_options)?;
+
         if s3_options.locking_provider.as_deref() == Some("dynamodb") {
-            debug!(
-                "S3LogStoreFactory has been asked to create a LogStore with the dynamodb locking provider"
+            error!(
```
ethan-tyler, line 40: Should we hard error here instead of only logging?

rtyler: Because this is going to happen with such a major version change, my inclination is honestly not to panic on the presence of the environment variable.  In practical terms for my own deployment, that would require a fairly substantial coordination of deployments to do the code and infra upgrade simultaneously in a way that would inhibit adoption. That's why I went with the "neutering" approach rather than "explosion" approach here :smile:

ethan-tyler (approved):
image

PR #4485, opened 2026-05-26
refactor(optimize): use DataFusion scan for compact reads instead of raw ParquetObjectReader

corwinjoy:
## Summary

- The compact path in `OptimizeBuilder` previously used `ParquetObjectReader` + `ParquetRecordBatchStreamBuilder` directly, bypassing DataFusion entirely
- Replaced with `DeltaScanNext` + a transient `SessionContext`, matching the pattern already used by the z-order path
- Removes the raw parquet reader dependency from the optimize hot path — DataFusion's scan handles file access, projection, and schema evolution uniformly across both compact and z-order

**Why this matters:** the raw reader cannot participate in DataFusion session-level features (encryption factories, custom object stores, pushdown). Routing all reads through `DeltaScanNext` ensures those features apply consistently.

## Test plan

- [ ] All 16 existing `command_optimize` tests pass unchanged
- [ ] z-order tests (`test_zorder_*`) unaffected
- [ ] Compact produces identical row counts and data values before/after

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/core/src/delta_datafusion/mod.rs
+++ b/crates/core/src/delta_datafusion/mod.rs
@@ -64,6 +64,7 @@ pub use self::session::{
     DeltaParserOptions, DeltaRuntimeEnvBuilder, DeltaSessionConfig, DeltaSessionContext,
     create_session, create_session_state_with_spill_config,
 };
+pub(crate) use self::table_provider::next::FileSelection;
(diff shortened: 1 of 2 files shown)


corwinjoy:
Work in progress. Ignore.

PR #4487, opened 2026-05-27
feat(cdf): prune load_cdf reads by partition predicates

osmant:
# Description

Adds partition pruning to `load_cdf` reads when the pushed filter contains partition-only predicates.

Before this change, CDF reads over partitioned tables still planned file groups from non-matching partitions. This change extracts partition-only conjuncts from the DataFusion filter, evaluates them against CDF file action partition values, and skips non-matching `Add`, `AddCDCFile`, and `Remove` actions before building the parquet scans.

The pruning path is conservative:
- pruning setup and evaluation failures fail open
- mixed partition/data predicates only use the partition-only portion for pruning
- uncoercible partition values keep the file rather than pruning incorrectly
- only referenced partition columns are included in the pruning schema
- constant partition-only predicates are handled, including actions without readable partition metadata

Row-level correctness still comes from the existing `FilterExec` in `DeltaCdfTableProvider::scan`; this change only avoids planning files that can be ruled out by partition values.

AI assistance was used while developing and reviewing this change. I reviewed the generated code, checked the edge cases, and ran the local validation listed below.

## Examples

A CDF read of a **partitioned** table with a filter on a **partition column** now skips files in non-matching partitions instead of opening every file and discarding rows afterward. The examples below use a table partitioned by `birthday`.

### How it's triggered

The predicate just needs to reach the CDF scan — both user-facing paths do this:

```python
# Python: the `predicate` is pushed into the CDF table provider
reader = dt.load_cdf(starting_version=0, predicate="birthday = '2023-12-25'")
```

```rust
// Rust: any DataFusion filter over the registered CDF provider
ctx.register_table("cdf", Arc::new(DeltaCdfTableProvider::try_new(
    table.scan_cdf().with_starting_version(0),
)?))?;
let df = ctx.sql("SELECT * FROM cdf WHERE birthday = DATE '2023-12-25'").await?;
```

### What changes — `birthday = '2023-12-25'`

The version range contains data across `birthday = 2023-12-22 … 2023-12-29`.

**Before** — files from *every* partition are opened; a row-level `FilterExec` discards the non-matching rows only after the parquet is read:

```text
UnionExec
  FilterExec: birthday = 2023-12-25
    DataSourceExec: file_groups={5 groups: [ …22/, …23/, …24/, …29/ ]}   # AddCDCFile — all partitions
  FilterExec: birthday = 2023-12-25
    DataSourceExec: file_groups={4 groups: [ …22/, …23/, …24/, …25/ ]}   # Add — all partitions
  FilterExec: birthday = 2023-12-25
    DataSourceExec: file_groups={0 groups: []}                           # Remove
```

**After** — only the matching partition's files are planned; the rest are dropped *before* any parquet is opened:

```text
UnionExec
  FilterExec: birthday = 2023-12-25
    DataSourceExec: file_groups={0 groups: []}                           # AddCDCFile — none in 2023-12-25
  FilterExec: birthday = 2023-12-25
    DataSourceExec: file_groups={1 group:  [ birthday=2023-12-25/part-00007…, -00008…, -00009… ]}   # Add
  FilterExec: birthday = 2023-12-25
    DataSourceExec: file_groups={0 groups: []}                           # Remove
```

The `FilterExec` is untouched, so the returned rows are identical — the change only avoids opening files that partition values rule out.

### When it applies

| Filter | Pruning | Result |
|---|---|---|
| `birthday = '2023-12-25'` (partition column) | ✅ scans only that partition | same rows, fewer files opened |
| `id = 5` (data column, not partitioned) | ❌ no change — all files planned | correct via the existing `FilterExec` |
| `birthday = '2023-12-25' AND id > 100` | ✅ partition part prunes; `id > 100` left to `FilterExec` | same rows, fewer files opened |
| `birthday = '2023-12-25'`, unrelated partition value malformed | ✅ still prunes on `birthday` | unreadable/unrelated values never force-keep a file |

Pruning is strictly conservative: it never changes which rows are returned, only which files are read. Uncoercible or missing partition values keep the file, mixed predicates use only the partition-only portion, and setup/evaluation failures fall back to scanning everything.

# Related Issue(s)

None.

# Documentation

No documentation changes. This is an internal planning optimization for CDF reads.

Validation run locally:

```sh
cargo fmt -- --check
cd python
make check-rust
make check-python
make develop
make unit-test
make build-docs
cargo test -p deltalake-core --features datafusion --lib cdf
cargo test -p deltalake-core --features datafusion --lib find_files
cargo test -p deltalake-core --features datafusion --lib delete_partition_only
```


hntd187:
When would you realistically do this? This looks like a solution looking for a problem.

osmant:
The scenario we ran into is multi-tenant CDC processing.

We are an ISV, and we have many tenants. Tenant data is stored in shared source databases/tables, with `tenant_id` as the separator. We then mirror that data into Delta tables using CDC.

From there, the downstream processing is tenant-specific. Different tenants can have different custom fields enabled, so after the shared Bronze/landing Delta table is created, each tenant’s data may need to be normalized and transformed differently into its own Silver/Gold tables.

As a result, we do not always want to process the full CDF for the shared table. A tenant-specific pipeline naturally wants to read only the CDF records for that tenant, for example:

```python
dt.load_cdf(
    starting_version=last_processed_version,
    ending_version=current_version,
    predicate="tenant_id = 'tenant_a'",
)
```

Today that returns the correct rows, but the CDF scan can still plan/read files for other tenant partitions in the version range and discard them later via the row-level filter. With many tenants and object storage, that means a tenant-specific incremental job can pay for unrelated tenants’ change files.

This is also something we previously observed in our Spark-based processing. Before Delta Lake 3.3.0, our Spark CDF reads had the same kind of inefficiency: even when we only needed one tenant’s changes, the read path still scanned much more data than necessary. After moving to a Delta Lake version that included delta-io/delta#3127, those Spark CDF reads became noticeably more efficient for us.

That Spark change was the inspiration for this PR. This delta-rs change is narrower: it only uses partition-only predicate terms to avoid planning files whose partition values cannot match. The existing `FilterExec` still enforces row-level correctness.

rtyler (commented):
From my perspective this looks reasonable to me. @hntd187 and I also discussed the topic a bit in Slack privately. I will let him be the ✔️ if he approves on the PR.

hntd187 (approved):

ethan-tyler:
@osmant  - curious to know what ISV this feature is for where you have multi tenets using delta tables.

osmant:
@ethan-tyler , It is "beqom"

PR #4489, opened 2026-05-27
feat: add public snapshot file selection api

ethan-tyler:
Public replacement for selected file scans that previously depended on internal provider hooks.

Selected inputs identify files only: partition values, stats, DVs, column mapping, tags, and active status are resolved from the scan snapshot.

Changes:
- Public `FileSelection` and `MissingSelectedFilePolicy`
- `with_files`, `with_file_paths`, and `with_file_selection` on `DeltaScanNext` and `TableProviderBuilder`
- Selected files resolved against snapshot before pruning
- Add actions treated as path identity only
- Url credentials stripped from file selections before serialization
- Absolute paths outside table root rejected with redacted errors
- Empty selections short circuit without replaying scan metadata
- Reader protocol validation moved into `DeltaScanNext::new`

Non empty selections need a metadata planning pass that allow missing files to be validated before query pruning. 

`MissingSelectedFilePolicy::Error` is the default; `Ignore` is for maintenance rewrites where removed files may be skipped.


crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -54,111 +53,182 @@ use crate::delta_datafusion::DeltaScanConfig;
...
+/// Selections with paths are resolved against snapshot metadata before the data scan. Each scan
+/// with a file selection requires that metadata pass.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct FileSelection {
+    paths: Vec<String>,
+    missing_file_policy: MissingSelectedFilePolicy,
+}
+
+#[derive(Clone, Debug)]
+struct ResolvedFileSelection {
```
rtyler, line 112: Thanks for adding the rustdoc for the `FileSelection`, with the APIs changing to return this Resolved version, por que no rustdoc on this new struct? The members refer to `file_ids` which I think would be very useful to make abundantly clear _what_ those IDs actually are and who produces them

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -54,111 +53,182 @@ use crate::delta_datafusion::DeltaScanConfig;
...
-                // Add.path is URI-decoded once, while scan file IDs are keyed by the
-                // object-store/raw path form. Route through Path so literal '%' bytes in
-                // partition directories are escaped back to the canonical file-id form.
-                .map(|add| String::from(Path::from(add.path))),
-            table_root_url,
-        )
+fn add_path_for_selection(add: Add) -> String {
+    let path = add.path;
+    if Url::parse(&path).is_ok() {
+        sanitize_selection_path(path)
```
rtyler: This interaction with the sanitize code is a bit goofy and I'm not sure why.  This line parses the `String` as a `Url`, and then the function immediately `Url::parses` it again 😕  If the thingie cannot be parsed as a URL, the `strip_unparsed_url_sensitive_parts` does a ton of other stuff that makes it seem like it's _also_ trying to parse the thing like it's a Url too 😵‍💫 

I think I understand the desire to normalize Add actions here, but there's a call tree 4-5 calls deep which are not unit tested here, that are doing a  _lot_ of futzing with strings in a way that I am not seeing the utility of compared to the code that was here before.

crates/core/src/delta_datafusion/table_provider.rs
```diff
@@ -348,6 +350,25 @@ impl TableProviderBuilder {
         self
     }
 
+    /// Restrict reads to Add action paths. File metadata comes from the selected snapshot.
+    pub fn with_files(self, files: impl IntoIterator<Item = Add>) -> Self {
```
rtyler: Since this is introducing a new public API, `with_adds` is IMO more suitable considering what this is accepting as parameters

crates/core/src/delta_datafusion/table_provider/next/mod.rs
```diff
@@ -186,6 +252,12 @@ pub(crate) fn canonical_table_root_identity(root: &url::Url) -> url::Url {
     root
 }
 
+fn table_root_identity(root: &url::Url) -> url::Url {
+    let mut root = ensure_table_root_url(&normalize_table_url(root));
+    strip_url_sensitive_parts(&mut root);
+    root
+}
```
rtyler: I do not see any benefit to this logic being pulled out into a separate function as opposed to simply being inlined in its caller `canonical_blahblah`.  Generally speaking, I think separate functions for logic like this _can_ be useful but only if they're getting called from multiple locations *and* they have N+1 unit tests for them.  

Every new function introduces adds some more cognitive load for any reader of the code to build the mental map of how interactions flow together. for example, if there are a lot of small functions that are all _close together in the same file_ that ends up being more aesthetically pleasing to me. If those same small functions are spread across multiple modules in different files, I find myself getting frustrated "now wtf does _this_ do? <jump> And that ties into _that_?"

I don't think [there's an ideal line length](https://tidyfirst.substack.com/p/how-long-should-functions-be) and I know you love a little function @ethan-tyler  😄 , my request is that small functions have a higher burden of proof in order to exist

rtyler (changes requested):

rtyler (approved):

PR #4493, opened 2026-05-30
feat: column mapping physical exec

ion-elgreco:
# Description
Add column mapping write support through datafusion physical execution node that rewrites columns and nested column names and attaches the metadata from te physical kernel schema.

Schema evolution, and optimize is not supported yet. Will do that in followups

### DIFF
--- a/crates/core/src/delta_datafusion/column_mapping.rs
+++ b/crates/core/src/delta_datafusion/column_mapping.rs
@@ -10,6 +10,7 @@
 //! survive), making the rewrite effectively zero-copy.
 
 use std::any::Any;
+use std::collections::HashMap;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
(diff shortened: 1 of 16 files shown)


ethan-tyler:
reviewing this today

crates/core/src/kernel/transaction/protocol.rs
```diff
@@ -319,9 +319,9 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
         writer_features.insert(TableFeature::Invariants);
         writer_features.insert(TableFeature::CheckConstraints);
         writer_features.insert(TableFeature::GeneratedColumns);
+        writer_features.insert(TableFeature::ColumnMapping);
```
ethan-tyler, line 322: This adds `ColumnMapping` to writer features but not every datafusion write path goes through the mapping rewrite.

crates/core/src/operations/write/mod.rs
```diff
@@ -406,8 +406,12 @@ impl WriteBuilder {
...
         match &self.snapshot {
             Some(snapshot) => {
-                if snapshot.table_configuration().column_mapping_mode() != ColumnMappingMode::None {
-                    return Err(unsupported_column_mapping_write("WRITE"));
+                if snapshot.table_configuration().column_mapping_mode() != ColumnMappingMode::None
+                    && self.schema_mode.is_some()
+                {
+                    return Err(DeltaTableError::Generic(
+                        "Schema evolution on column-mapped tables is not yet supported".to_string(),
+                    ));
```
ethan-tyler, line 414: This gates `WriteBuilder` schema evolution, but merge still exposes `with_merge_schema(true)`. That path can build replacement metadata without mapping ids, physical names, or `maxColumnId`.

crates/core/src/operations/merge/mod.rs
```diff
@@ -1827,10 +1825,6 @@ impl std::future::IntoFuture for MergeBuilder {
         Box::pin(async move {
             let snapshot =
                 resolve_snapshot(&this.log_store, this.snapshot.clone(), true, None).await?;
-            if snapshot.table_configuration().column_mapping_mode() != ColumnMappingMode::None {
-                return Err(unsupported_column_mapping_write("MERGE"));
-            }
-
```
ethan-tyler, line 1833: Removing this enables more than simple matched/not matched rewrites. `MergeBuilder` supports `with_merge_schema(true)` and that path isn't column mapping aware.

crates/core/src/delta_datafusion/column_mapping.rs
```diff
@@ -0,0 +1,552 @@
...
+            None => field.as_ref().clone(),
+        })
+        .collect()
+}
+
+/// Build the physical Arrow field for a single logical field: rename to the physical name,
+/// attach the `field_id` metadata, and recurse into nested types — all while preserving the
+/// input Arrow data type.
+fn physical_field(field: &Field, kernel_field: &StructField, mode: ColumnMappingMode) -> Field {
+    let mut metadata = field.metadata().clone();
```
ethan-tyler: Top level field ids come from the table schema but nested list/map fields below preserve collar arrow metadata. Column mapping ids and field ids imo should be table owned and not from input batches.

crates/core/src/operations/write/execution.rs
```diff
@@ -277,6 +279,62 @@ struct WriteSinkConfig {
...
+
+    /// Translate logical partition column names to their physical names.
+    fn physical_partition_columns(&self, partition_columns: &[String]) -> Vec<String> {
+        partition_columns
+            .iter()
+            .map(|name| {
+                self.logical_schema
+                    .field(name)
+                    .map(|field| field.physical_name(self.mode).to_string())
+                    .unwrap_or_else(|| name.clone())
```
ethan-tyler: I am not sure if this should fallback silently to the logical partition name if the field is missing from the table schema. On a mapped table, that writes the wrong `partitionValues` key.

crates/core/src/delta_datafusion/column_mapping.rs
```diff
@@ -0,0 +1,552 @@
...
+/// into its (possibly nested) value type.
+fn rebuild_element(
+    element: &Arc<Field>,
+    element_kernel: &KernelDataType,
+    mode: ColumnMappingMode,
+) -> Arc<Field> {
+    let data_type = physical_data_type(element.data_type(), element_kernel, mode);
+    Arc::new(
+        Field::new(element.name(), data_type, element.is_nullable())
+            .with_metadata(element.metadata().clone()),
```
ethan-tyler: Similarly to my comment above, preserving list element metadata here can carry stale `PARQUET:field_id` into the writer. I think this should come from the table schema's mapping and not the input batch.

ethan-tyler (commented):
I think the approach is strong and good. I did find some concerns and left those comments below.

crates/core/src/kernel/transaction/protocol.rs
ion-elgreco: Fair point, forgot about that datafusion sink one :*(

Did you manage to get an upstream change in datafusion regarding the stream we discussed last time on that PR?

crates/core/src/operations/write/execution.rs
ion-elgreco: Yeah this should raise, its probably unreachable unless there is a silent bug in kernel

ethan-tyler (approved):

PR #4504, opened 2026-06-02
fix: correct format for timestampNtz statistics

### LINKED ISSUE #4503: [Bug]: timestampNtz statistics are formatted as UTC
### What happened?

When writing data to a table that uses the `timestampNtz` type, the statistics in JSON log files are formatted as UTC timestamps, for example `2026-06-01T09:00:00Z`.

### Expected behavior

`timestampNtz` statistics should be formatted without specifying a timezone.

I checked Delta PROTOCOL.md and it doesn't actually specify how statistics should be formatted as far as I can tell. See the [section on per-file statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics).

There is however a section on [partition value serialization](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization) which allows the use of ISO8601 for `timestamp` but not `timestampNtz`.

### Operating System

Linux

### Binding

Python

### Bindings Version

1.6.0

### Steps to reproduce

1. Write data:
```python
import datetime
import deltalake
import pyarrow as pa


start = datetime.datetime(2026, 6, 1, 9, 0, 0, tzinfo=datetime.UTC)
batch = pa.Table.from_pydict({
    'ts': pa.array([start + datetime.timedelta(seconds=i) for i in range(10)], type=pa.timestamp('us', 'UTC')),
    'ts_ntz': pa.array([start + datetime.timedelta(seconds=i) for i in range(10)], type=pa.timestamp('us', None)),
})

deltalake.write_deltalake("./data", batch)
```

2. Check stats:
```bash
jq 'add.stats // empty | fromjson' < data/_delta_log/00000000000000000000.json
```

This outputs:
```
{
  "numRecords": 10,
  "minValues": {
    "ts_ntz": "2026-06-01T09:00:00Z",
    "ts": "2026-06-01T09:00:00Z"
  },
  "maxValues": {
    "ts_ntz": "2026-06-01T09:00:09Z",
    "ts": "2026-06-01T09:00:09Z"
  },
  "nullCount": {
    "ts": 0,
    "ts_ntz": 0
  }
```

adamreeve:
# Description

Changes how min and max-value statistics are formatted for `timestampNtz` columns so they aren't formatted as UTC timestamps. Previously they weren't treated any different to `timestamp` typed colums and used the ISO8601 format with a "Z" suffix.

# Related Issue(s)

- closes #4503

### DIFF
--- a/crates/core/src/writer/stats.rs
+++ b/crates/core/src/writer/stats.rs
@@ -266,6 +266,7 @@ enum StatsScalar {
     Float64(f64),
     Date(chrono::NaiveDate),
     Timestamp(chrono::NaiveDateTime),
+    TimestampNtz(chrono::NaiveDateTime),
     // We are serializing to f64 later and the ordering should be the same
     // Scale is stored to handle scale=0 serialization correctly
     Decimal { value: f64, scale: i32 },
@@ -308,10 +309,13 @@ impl StatsScalar {
             }
             (Statistics::Int32(v), _) => Ok(Self::Int32(get_stat!(v))),
             // Int64 can be timestamp, decimal, or integer
-            (Statistics::Int64(v), Some(LogicalType::Timestamp { unit, .. })) => {
-                // For now, we assume timestamps are adjusted to UTC. Non-UTC timestamps
-                // are behind a feature gate in Delta:
-                // https://github.com/delta-io/delta/blob/master/PROTOCOL.md#timestamp-without-timezone-timestampntz
+            (
+                Statistics::Int64(v),
+                Some(LogicalType::Timestamp {
+                    is_adjusted_to_u_t_c,
+                    unit,
+                }),
+            ) => {
                 let v = get_stat!(v);
                 let timestamp = match unit {
                     TimeUnit::MILLIS => chrono::DateTime::from_timestamp_millis(v),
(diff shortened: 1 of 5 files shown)


crates/core/src/writer/stats.rs
```diff
@@ -451,6 +459,9 @@ impl From<StatsScalar> for serde_json::Value {
             StatsScalar::Timestamp(v) => {
                 serde_json::Value::from(v.format("%Y-%m-%dT%H:%M:%S%.fZ").to_string())
             }
+            StatsScalar::TimestampNtz(v) => {
+                serde_json::Value::from(v.format("%Y-%m-%d %H:%M:%S%.f").to_string())
```
adamreeve, line 463: I'm not 100% sure this is the right format to use. Spark actually uses a "T" separator rather than a space between the date and time components but this matches the format used in the [partition value serialization](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization) section of PROTOCOL.md. 

As far as I can tell, the exact format for file statistics isn't specified anywhere. Maybe it should be?

I guess the main concern with changing this format is whether it will affect use of these statistics for pruning files during queries. I've manually tested that queries with both `delta-rs` and `pyspark` (version 4.1.1 with `delta-spark` 4.2.0) still correctly prune files based on predicates with a `timestampNtz` column (by using strace to check which Parquet files are opened). I just noticed there is a `test_files_scanned` unit test and that does seem to cover the `timestampNtz` case but not `timestamp`, so I'll fix that.

ion-elgreco (dismissed):

crates/core/tests/it_datafusion/integration_datafusion.rs
```diff
@@ -1173,7 +1203,6 @@ mod local {
             // TODO: Float and decimal partitions are not supported by the writer
             // binary fails since arrow does not implement a natural order
             // The current Datafusion pruning implementation does not work for binary columns since they do not have a natural order. See #1214
-            // Timestamp and date are disabled since the hive path contains illegal Windows values. see #1215
```
adamreeve, line 1176: Looks like this was fixed in #1661.

rtyler (dismissed):
Looks reasonable to me, I'm not planning on backporting any nano timestamp things to 0.32.4 FYI

ion-elgreco (approved):

PR #4505, opened 2026-06-02
feat: add column mapping create support

ion-elgreco:
# Description
Pending cherry picking the changes of upstream pr https://github.com/delta-io/delta-kernel-rs/pull/2671 onto Buoyant kernel :)

### DIFF
--- a/crates/core/src/operations/create.rs
+++ b/crates/core/src/operations/create.rs
@@ -5,6 +5,9 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use delta_kernel::schema::{ColumnMetadataKey, MetadataValue};
+use delta_kernel::table_features::{
+    ColumnMappingMode, assign_column_mapping_metadata, find_max_column_id_in_schema,
+};
 use futures::TryStreamExt as _;
 use futures::future::BoxFuture;
 use serde_json::Value;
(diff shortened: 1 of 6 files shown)


ethan-tyler:
reviewing today

crates/core/tests/it/column_mapping_guardrails.rs
```diff
@@ -389,12 +425,89 @@ async fn column_mapping_guardrails_create_rejects_activation_and_reserved_metada
...
+            .build()?
+            .create()
+            .with_columns([
+                StructField::nullable("Company Very Short", DataType::STRING),
+                StructField::nullable("Super Name", DataType::STRING),
+            ])
+            .with_configuration([("delta.columnMapping.mode", Some(mode))])
+            .await?;
+
+        let table = table.write(vec![column_mapping_batch()]).await?;
```
ethan-tyler, line 455: The first write path (`WriteBuilder` creating the table and writing in one transaction) writes with `snapshot: None`, which causes the column mapping rewrite to be skipped and get logical names.

The committed metadata has generated physical names which also causes a corrupt table at version 0.

This test doesn't catch it because it creates the table first, then writes against the existing snapshot.

crates/core/src/operations/create.rs
```diff
@@ -351,6 +344,29 @@ impl CreateBuilder {
...
+                DeltaTableError::Generic(format!("Invalid delta.columnMapping.mode: {mode}"))
+            })?,
+            None => ColumnMappingMode::None,
+        };
+        let schema = if matches!(
+            column_mapping_mode,
+            ColumnMappingMode::Name | ColumnMappingMode::Id
+        ) {
+            let mut max_id = find_max_column_id_in_schema(&schema).unwrap_or(0);
+            let mapped = assign_column_mapping_metadata(&schema, &mut max_id, false)?;
```
ethan-tyler, line 360: Create generates fresh physical names but caller provided file actions are committed as is.

`CreateBuilder::with_actions `and `create_table_with_add_actions` can pass `Action::Add` into this create. After this line, the table schema expects generated `col-<uuid>` names but the Add action may still use logical or unrelated physical names.

ion-elgreco: Will try to get back to this later in July!

ethan-tyler:
Just @ me when this is ready @rtyler

ethan-tyler (dismissed):

ethan-tyler (approved):

PR #4509, opened 2026-06-02
chore: be a little more exhaustive in our kernel expression matching

rtyler:
regrettably these non-exhaustive match expressions are released into the
world which means that the next release of buoyant_kernel needs to bump
into the 100s and cannot be safely incorporated into the currently released 0.32.x crates


rtyler:
@ion-elgreco et al, this updates `main`to a branch of buoyant_kernel that I have not yet pushed to crates.io but it _does_ enable the functionality associated with #4505

ion-elgreco (approved):

PR #4510, opened 2026-06-02
feat: update schema mismatch error message

### LINKED ISSUE #1652: Improve schemaMismatch error display
# Description

**Use Case**
Currently when there is a schema mismatch, you get a huge stack trace, displaying the full arrow and delta schema together. This quickly becomes a find Waldo situation and it's easier to just display the symmetric difference first.

**Related Issue(s)**
https://github.com/delta-io/delta-rs/issues/1648#issuecomment-1728341099

JustinRush80:
# Description
Updates the Schema Mismatch message to return the symmetric difference instead of the two full schema definition

# Related Issue(s)

- closes #1652

# Documentation
I wrote all the tests myself and used AI for very specific task such as making my solution more idiomatic rust , Lifetime annotation which I am still learning, and reducing code while still passing the tests.

### DIFF
--- a/crates/core/src/kernel/schema/cast/mod.rs
+++ b/crates/core/src/kernel/schema/cast/mod.rs
@@ -7,8 +7,10 @@ use arrow_array::{
 };
 use arrow_cast::{CastOptions, cast_with_options};
 use arrow_schema::{
-    ArrowError, DataType, FieldRef, Fields, Schema, SchemaRef as ArrowSchemaRef, TimeUnit,
+    ArrowError, DataType, Field, FieldRef, Fields, Schema, SchemaRef as ArrowSchemaRef, TimeUnit,
 };
+use std::collections::HashSet;
+use std::ops::Deref;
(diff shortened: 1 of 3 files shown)


crates/core/src/kernel/schema/cast/mod.rs
```diff
@@ -345,6 +346,26 @@ pub fn normalize_for_delta(schema: &ArrowSchemaRef) -> ArrowSchemaRef {
     }
 }
 
+pub fn symmetric_differences<'a>(
```
rtyler:
```suggestion
pub(crate) fn symmetric_differences<'a>(
```

I don't believe this needs to be [a public API](https://docs.rs/deltalake-core/latest/deltalake_core/kernel/schema/cast/index.html), but should be visible across the crate.

ion-elgreco (dismissed):

ion-elgreco (approved):

PR #4511, opened 2026-06-03
Refactor Parquet writers

adamreeve:
Refactor Parquet writers as a base to build more customizable Parquet options and support encryption.


adamreeve:
Ugh sorry, made this against the wrong repo.

PR #4513, opened 2026-06-04
feat: allow switching nanosecond timestamps on at runtime in Python

adamreeve:
Follow up to adding nanosecond timestamp support (#4370). Allows use of this experimental feature from Python without turning it on by default, and also allows testing via Python.

Supersedes #4420

### DIFF
--- a/python/deltalake/writer/_conversion.py
+++ b/python/deltalake/writer/_conversion.py
@@ -9,7 +9,9 @@ def _convert_arro3_schema_to_delta(
     existing_schema: Arro3Schema | None = None,
 ) -> Arro3Schema:
     """Convert a arro3 schema to a schema compatible with Delta Lake. Converts unsigned to signed equivalent, and
-    converts all timestamps to `us` timestamps. Also handles null column types by converting them to match
+    converts all timestamps to `us` timestamps, except nanosecond timestamps when the experimental nanosecond
+    timestamp feature is enabled.
+    Also handles null column types by converting them to match
     corresponding fields in the existing table schema.
 
     Args:
@@ -19,6 +21,7 @@ def _convert_arro3_schema_to_delta(
     Returns:
         Arro3Schema: Delta-compatible schema with converted types.
     """
+    from deltalake import _nanosecond_timestamps_enabled
 
     def dtype_to_delta_dtype(
         dtype: DataType, field_name: str | None = None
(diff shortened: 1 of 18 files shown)


python/tests/test_cdf.py
```diff
@@ -503,10 +503,12 @@ def test_delete_unpartitioned_cdf(tmp_path, sample_data_pyarrow: "pa.Table"):
...
         )
     )
-    cdc_data = pq.read_table(cdc_path)
+    # Ensure the CDC data is read using the expected schema
+    # (eg. standard binary/string types rather than binary/string_view).
+    cdc_data = pq.read_table(cdc_path, schema=expected_data.schema)
 
     assert os.path.exists(cdc_path), "_change_data doesn't exist"
-    assert cdc_data.to_pydict() == expected_data.to_pydict()
+    assert cdc_data == expected_data
```
adamreeve, line 511: Using `to_pydict` caused an [error when Pandas wasn't installed](https://github.com/delta-io/delta-rs/actions/runs/25940061211/job/76256672471):

```
ValueError: Nanosecond resolution temporal type 3 is not safely convertible to microseconds to convert to datetime.datetime. Install pandas to return as Timestamp with nanosecond support or access the .value attribute.
```

It looks like this was needed because otherwise if you compared the tables directly you got errors about the types not matching, due to pyarrow reading some string typed columns as `string_view` whereas the expected data used `string` (and similarly for binary columns). I've worked around this by providing the desired arrow schema to use when reading the Parquet CDC data.

python/deltalake/writer/_conversion.py
```diff
@@ -51,7 +54,7 @@ def dtype_to_delta_dtype(
         elif DataType.is_timestamp(dtype):
             if dtype.tz is None:
                 return DataType.timestamp("us")
-            elif dtype.time_unit == "ns":
+            elif dtype.time_unit == "ns" and _nanosecond_timestamps_enabled():
```
adamreeve, line 57: This change wasn't strictly needed for the normal write path because the schema gets normalized to microseconds on the rust side if nanos isn't enabled, but the previous behaviour caused a problem if doing a merge that tries to add a nanosecond timestamp column. I've added a new regression test to cover that scenario in `python/tests/test_merge.py`.

python/tests/test_merge.py
```diff
@@ -3398,3 +3398,77 @@ def test_merge_with_spill_config(tmp_path: pathlib.Path):
...
+            source_alias="s",
+            target_alias="t",
+            merge_schema=True,
+        ).when_not_matched_insert_all().execute()
+
+    if enable_nanos:
+        if not _NANOSECOND_TIMESTAMPS:
+            pytest.skip("Rust library not built with nanosecond-timestamps enabled")
+        # This currently errors because a merge doesn't automatically add
+        # the nanosecond timestamps feature to the table.
```
adamreeve, line 3446: I should fix this but it seems out of scope for this PR.

rtyler (approved):

PR #4525, opened 2026-06-10
chore: upgrade to datafusion 54

laverem:
## Description

Bumps the workspace from DataFusion 53.1.0 to 54.0.0. arrow (58) and
object_store (0.13.2) already match DataFusion 54's requirements, so this is a
pure DataFusion bump.

### Rust API changes

**Dropped `as_any` (now `Any` supertraits with inherent `downcast_ref`/`is`):**
`ExecutionPlan`, `TableProvider`, `SchemaProvider`, `ScalarUDFImpl`, `DataSink`,
`CatalogProvider`, `CatalogProviderList`, `DataSource`, `FileSource`. Call sites
that did `plan.as_any().downcast_ref::<T>()` on an `Arc<dyn ExecutionPlan>` now
call `plan.downcast_ref::<T>()` directly (this includes the next-scan test
helpers).

**Other API changes:**
- `Session` / `ContextProvider`: implement the new higher-order-function and
  extension-type registry methods
- `partition_statistics()` now returns `Arc<Statistics>`
- `PruningStatistics::row_counts()` drops the `column` arg
- `Cast` moved to its `FieldRef` field
- `PartitionedFile`: fill in the new fields
- `SimplifyContext`: use the builder

### Python / FFI

The exported FFI `TableProvider` is now DataFusion 54.x, so the runtime guard
(`REQUIRED_DATAFUSION_PY_MAJOR`) requires a datafusion-python major of 54.
datafusion-python has no 54.x wheel published yet, so the
`test_datafusion_table_provider` integration test is gated on `==54` and skips
otherwise (the install pin stays at the latest available `53.*` until a 54
wheel ships). The version-mismatch unit tests assert the guard rejects a 53
wheel with a `datafusion==54` message.

### Verification

`cargo fmt --check`, `cargo clippy --workspace --all-features`,
`cargo build --workspace --all-features`, and
`cargo test -p deltalake-core --all-features` all pass.

Generated-by: Claude (Anthropic), with human review and testing

### DIFF
--- a/python/src/writer.rs
+++ b/python/src/writer.rs
@@ -31,7 +31,10 @@ pub fn to_lazy_table(
     )?))
 }
 pub struct ReaderWrapper {
-    reader: Mutex<Box<dyn RecordBatchReader + Send + 'static>>,
+    /// The underlying reader. `None` once it has been handed off to a fresh
+    /// generator by [`ArrowStreamBatchGenerator::reset_state`], since a
+    /// one-shot stream can only be consumed once.
+    reader: Mutex<Option<Box<dyn RecordBatchReader + Send + 'static>>>,
 }
 
 impl fmt::Debug for ReaderWrapper {
@@ -61,7 +64,7 @@ impl ArrowStreamBatchGenerator {
     pub fn new(array_stream: Box<dyn RecordBatchReader + Send + 'static>) -> Self {
         Self {
             array_stream: ReaderWrapper {
-                reader: Mutex::new(array_stream),
+                reader: Mutex::new(Some(array_stream)),
             },
         }
     }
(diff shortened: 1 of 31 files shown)


python/src/writer.rs
```diff
@@ -91,8 +102,19 @@ impl LazyBatchGenerator for ArrowStreamBatchGenerator {
         }
     }
 
+    /// DataFusion 54's `LazyMemoryExec::execute` calls `reset_state` on every
+    /// execution (including the first) to obtain a fresh stream. Since the
+    /// underlying reader is one-shot, we hand it off to the new generator the
+    /// first time and leave an [`ExhaustedStreamGenerator`] behind so any later
+    /// re-execution fails loudly rather than silently yielding no rows.
```
ethan-tyler, line 109: Nice ... This reset state handoff is what I would have done for DF's `LazyMemoryExec::execute` behavior. However, I would appreciate a test that proves first execution succeeds after `reset_state()'.

ethan-tyler (commented):
Thanks @laverem! The rust upgrade looks good to me.

I left one comment that is nice to have but not blocking. The tricky thing with DataFusion upgrades is that `datafusion-python` usually lags the main rust release by a couple weeks while they do their upgrade causing the next version’s wheels aren’t available right away. That trips up a bunch of the python code and we usually gate that away until the wheels are released.

ethan-tyler (approved):
Approving - If you want to add those follow ups, feel free to open a new PR and I'll get those in too.

laverem:
@ethan-tyler thank you for the review! I've opened https://github.com/delta-io/delta-rs/pull/4530

PR #4530, opened 2026-06-11
test: prove first execution succeeds after reset_state

laverem:
Follow-up to the DataFusion 54 upgrade (#4525). Adds unit tests for
`ArrowStreamBatchGenerator` in python/src/writer.rs covering DF54's
`LazyMemoryExec::execute` behavior, which calls `reset_state` on the first
execution:

- first `reset_state` hands the reader to a fresh generator that yields the
  original batch (the regression that was breaking the first write)
- a second `reset_state` is exhausted and errors loudly rather than silently
  yielding no rows

AI disclosure: tests were AI-generated and human verified locally with
`cargo test --lib writer::tests`.

### DIFF
--- a/python/src/writer.rs
+++ b/python/src/writer.rs
@@ -31,9 +31,8 @@ pub fn to_lazy_table(
     )?))
 }
 pub struct ReaderWrapper {
-    /// The underlying reader. `None` once it has been handed off to a fresh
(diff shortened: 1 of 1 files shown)


ethan-tyler (commented):
Thanks for the follow up!! I hate to be nit picky but the AI comments are overly verbose and the datafusion version is irrelevant.

PR #4532, opened 2026-06-12
fix: coerce inner field name when writing fixed size lists from Rust

### LINKED ISSUE #4531: [Bug]: Cannot write FixedSizeList data from Rust when using the default "item" inner field name
### What happened?

Basically the same issue as #3566, but because that was fixed in the Python wrapper, you still get this error when writing with Rust:

```
Error: ArrowError(InvalidArgumentError("column types must match schema types, expected List(Int32, field: 'element') but found List(Int32) at column index 0"), Some(""))
```

### Expected behavior

You can write Arrow fixed-size list data and it's stored as a Delta Lake `List` column.

### Operating System

Linux

### Binding

Rust

### Bindings Version

20bdf3c2c62416923fe327bd0e41976627f143ec

### Steps to reproduce

I'll make a PR that includes a unit test to reproduce this.

### Relevant logs

```shell

```

adamreeve:
Fixes #4531

### DIFF
--- a/crates/core/src/kernel/schema/cast/mod.rs
+++ b/crates/core/src/kernel/schema/cast/mod.rs
@@ -278,7 +278,18 @@ fn normalize_datatype(dt: &DataType) -> Option<DataType> {
         }
         DataType::List(inner) => normalize_field(inner).map(DataType::List),
         DataType::FixedSizeList(inner, size) => {
-            normalize_field(inner).map(|normalized| DataType::FixedSizeList(normalized, *size))
+            let normalized_inner = if inner.name() != "element" {
+                let inner = inner.as_ref().clone().with_name("element");
+                let dt = normalize_datatype(inner.data_type());
+                let inner = match dt {
+                    Some(dt) => inner.with_data_type(dt),
+                    None => inner,
+                };
+                Some(Arc::new(inner))
+            } else {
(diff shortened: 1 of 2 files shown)


crates/core/src/kernel/schema/cast/mod.rs
```diff
@@ -278,7 +278,18 @@ fn normalize_datatype(dt: &DataType) -> Option<DataType> {
         }
         DataType::List(inner) => normalize_field(inner).map(DataType::List),
         DataType::FixedSizeList(inner, size) => {
-            normalize_field(inner).map(|normalized| DataType::FixedSizeList(normalized, *size))
+            let normalized_inner = if inner.name() != "element" {
```
adamreeve, line 281: This duplicates logic in the Python `_convert_arro3_schema_to_delta` function. Should that Python function exist? Can we do all casting in Rust instead to avoid issues like this?

ion-elgreco: _convert_arro3_schema_to_delta predates the rust opinionated casting logic when I refactored everything to arro3 and previously it was in pyarrow. We can remove it as long as the rust casting is guaranteed to happen the same. In python we also cast unsigned to signed.

rtyler (approved):

PR #4553, opened 2026-06-19
fix: relax column nullability under SchemaMode::Merge

### LINKED ISSUE #3688: Allow changing a column non-nullable to nullable in `SchemaMode::Merge` mode.
# Environment

**Delta-rs version**: `0.27.0`

**Binding**:  I am using the Rust library

***
# Bug

**What happened**:
`delta-rs` does not allow making a column non-nullable to nullable with `SchemaMode::Merge`.

`delta-rs` should allow making a column non-nullable to nullable. Nullable to non-nullable should not be permitted with a concept of a default value (Not in the scope of this issue).

**How to reproduce it**:
```rust
#[tokio::test]
async fn null_to_non_null_column() {
    let schema_without_null_column = vec![StructField::new("id", DeltaDataType::INTEGER, false)];

    let non_null_batch = RecordBatch::try_from_iter_with_nullable(vec![(
        "id",
        Arc::new(Int32Array::from(vec![1, 2, 3])) as ArrayRef,
        false,
    )]);

    let null_batch = RecordBatch::try_from_iter_with_nullable(vec![(
        "id",
        Arc::new(Int32Array::from(vec![Some(1), None, Some(3)])) as ArrayRef,
        true,
    )]);

    let ops = DeltaOps::try_from_uri("memory:///").await.unwrap();
    let table = ops
        .create()
        .with_table_name("my_table")
        .with_save_mode(SaveMode::Ignore)
        .with_columns(schema_without_null_column)
        .await
        .unwrap();

    DeltaOps::from(table.clone())
        .write(non_null_batch)
        .with_schema_mode(SchemaMode::Merge)
        .with_save_mode(SaveMode::Append)
        .await
        .unwrap();

    DeltaOps::from(table)
        .write(null_batch)
        .with_schema_mode(SchemaMode::Merge)
        .with_save_mode(SaveMode::Append)
        .await
        .unwrap();
}
```

The above code raises `InvalidData { violations: ["Non-nullable column violation for id, found 1 null values"] }`.

**More details**:

https://github.com/delta-io/delta-rs/blob/ef9e077402013bedf44fccf872f7d7d4fd9157ab/crates/core/src/kernel/schema/cast/merge_schema.rs#L258

https://github.com/apache/arrow-rs/blob/6c1b96f877cbab99ae4dfcbfd503590323bf83cb/arrow-schema/src/field.rs#L825C9-L825C40

adityavaish:
# Description

Writing nullable data into a previously non-nullable column with `SchemaMode::Merge` did not relax the column to nullable. The relaxation was never committed to the table metadata, and any actual `null` values were rejected at write time by the `NOT NULL` data-validation check. This affected both non-empty batches (containing nulls) and empty/zero-row batches that only change nullability.

## Root cause

Two independent gaps:

1. **Schema metadata not updated.** `try_cast_schema` only checks data-type castability and *ignores nullability*, so a nullability-only difference (nullable source field → non-nullable table column) produced `schema_drift = false`. Because the `SchemaMode::Merge` metadata update is gated on `schema_drift`, the merged (relaxed) schema was never written, so the column stayed non-nullable.
2. **Data validation rejected the nulls.** `validation_predicates` derives `IS NOT NULL` checks from the **pre-write** snapshot (columns non-nullable in the table but nullable in the source — exactly the columns a merge write relaxes). These checks rejected the very nulls the merge was supposed to allow.

## Fix

- Add `has_nullability_relaxation(from_fields, to_fields)` (recursing into structs) in `schema_evolution`, and in `write/plan.rs` mark `schema_drift = true` under `SchemaMode::Merge` when a relaxation is detected, so the merged nullable schema is committed.
- Thread a `relax_nullability` flag through `write_execution_plan_v2` into `validation_predicates`; under `SchemaMode::Merge` the relaxation-column `NOT NULL` checks are skipped (the column is being relaxed). Invariants, check constraints, and generated-column validations are unaffected.

The change is scoped to `SchemaMode::Merge`: default and `Overwrite` schema modes keep their existing behavior (nulls into a non-nullable column are still rejected).

## Behavior

```rust
// table has non-nullable column `id`
let table = DeltaOps(table)
    .write(vec![nullable_id_batch])      // id typed nullable, may contain nulls
    .with_schema_mode(SchemaMode::Merge)
    .with_save_mode(SaveMode::Append)
    .await?;
// `id` is now nullable and the null rows are written
```

## Testing

- Added two regression tests in `operations::write::tests`:
  - `test_merge_schema_relaxes_nullability_with_null_batch` (non-empty batch with null values)
  - `test_merge_schema_relaxes_nullability_with_empty_batch` (nullability-only change, zero rows)
- `cargo test -p deltalake-core --features datafusion --lib` — **1018 passed, 0 failed, 2 ignored**.
- `cargo fmt --check` and `cargo clippy -p deltalake-core --features datafusion --tests` are clean (no new warnings).

# Related Issue(s)

- closes #3688

# Documentation

- Delta protocol — schema / `metaData` and column nullability: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#change-metadata

---

> [!NOTE]
> **AI assistance disclaimer:** This change was implemented with the help of an AI coding assistant (GitHub Copilot CLI). All changes were made under human oversight and review: the root-cause analysis, design, and scope decisions were reviewed by a human before submission, the failing scenario was reproduced first, and the full test suite, formatting, and lints were run and verified to pass. Feedback is welcome and will be addressed with the same human-in-the-loop process.

### DIFF
--- a/crates/core/src/delta_datafusion/data_validation.rs
+++ b/crates/core/src/delta_datafusion/data_validation.rs
@@ -259,29 +259,38 @@ pub(crate) fn validation_predicates(
     session: &dyn Session,
     source_schema: &DFSchema,
     table_configuration: &TableConfiguration,
+    relax_nullability: bool,
 ) -> Result<Vec<Expr>> {
     // find all columns that are non-nullable in the table schema but nullable
     // in the source schema and add IS NOT NULL checks for them.
-    let table_schema: Schema = table_configuration
-        .logical_schema()
-        .as_ref()
-        .try_into_arrow()?;
-    let non_nullable_table: HashSet<_> = collect_non_nullable_fields(&table_schema)
-        .into_iter()
-        .collect();
-    let non_nullable_source: HashSet<_> = collect_non_nullable_fields(source_schema.as_arrow())
-        .into_iter()
-        .collect();
-    let mut validations: Vec<_> = non_nullable_table
-        .difference(&non_nullable_source)
-        .map(|col| {
-            to_datafusion_expr(
-                &Expression::Column(col.clone()),
-                &delta_kernel::schema::DataType::BOOLEAN,
-            )
-        })
-        .map_ok(|e| e.is_not_null())
-        .try_collect()?;
+    //
+    // Under a schema-merge write (`relax_nullability`), these are exactly the
+    // columns being relaxed from non-nullable to nullable, so the NOT NULL
+    // constraint must not be enforced for them.
+    let mut validations: Vec<Expr> = if relax_nullability {
+        Vec::new()
+    } else {
+        let table_schema: Schema = table_configuration
+            .logical_schema()
+            .as_ref()
+            .try_into_arrow()?;
+        let non_nullable_table: HashSet<_> = collect_non_nullable_fields(&table_schema)
+            .into_iter()
+            .collect();
+        let non_nullable_source: HashSet<_> = collect_non_nullable_fields(source_schema.as_arrow())
+            .into_iter()
+            .collect();
+        non_nullable_table
+            .difference(&non_nullable_source)
+            .map(|col| {
+                to_datafusion_expr(
+                    &Expression::Column(col.clone()),
+                    &delta_kernel::schema::DataType::BOOLEAN,
+                )
+            })
+            .map_ok(|e| e.is_not_null())
+            .try_collect()?
+    };
 
     if table_configuration.is_feature_enabled(&TableFeature::Invariants) {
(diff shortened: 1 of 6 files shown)


hntd187:
I don’t know if I agree with this. I would never want this to happen in real practice I’d almost always describe the data in this case as being incorrect and validation correct for catching it.

Not to mention how implicit this is one single null changes the entire column to be nullable in the future?

Issue #4518, opened 2026-06-09
[Bug]: New DeltaTableProvider eats too much memory on larger tables


rtyler:
### What happened?

<img width="894" height="414" alt="Image" src="https://github.com/user-attachments/assets/8636c364-042f-429f-98a2-b6b271314ec5" />

This ☝️ is the upgrade from 0.32.1 to 0.32.4 which incorporated some changes of note. In Slack @ethan-tyler hypothesized that this is something about the long URLs in datafusion planning.

Some 🦜 generated analysis:

```
    PRIMARY SUSPECT — EagerSnapshot now materializes FULL stats and preserves raw JSON, resident in memory

    In crates/core/src/kernel/snapshot/mod.rs, the snapshot materialization path was changed:

    - Old: materialize_files_with_engine called files_with_engine(...) → with_skip_stats(...), parsing a minimal stats schema (just numRecords, plus predicate columns).
    - New: it calls files_with_engine_preserving_raw(...) → FileStatsMode::FullPreserveRaw → StatsProjection::full() wrapped in FileStatsMaterialization::compatibility(...).
```

### Expected behavior

Not so much memory! 

### Operating System

None

### Binding

Rust

### Bindings Version

0.34.3

### Relevant logs

```shell

```

ethan-tyler:
Opened child investigation issues for the additional DeltaScanNext planning-memory areas found during the #4518 analysis:

- #4520: metadata replay retains high per-file planning memory
- #4521: repeated object path data during file planning
- #4522: per-file DataFusion statistics planning overhead
- #4523: one transform expression per file during scan planning
- #4524: deletion-vector planning overhead not yet measured

#4519 tracks the first concrete fix for the full-url `file_id` retention path.

ethan-tyler:
Had #AI organize my additional thoughts/ideas, things to look into more and open those related issues. Let's see how #4518 does and go from there.

ethan-tyler:
@rtyler did this get resolved for you?

rtyler:
I'm guessing it did. The exact apples-to-apples comparison with the screenshot above I won't have time for another week or so to get to. The integration testing I did over the weekend did not produce excess memory usage compared to the previous release I was running 🤷

PR #4555, opened 2026-06-20
fix: add DV-backed tables to scan-planning probe matrix (#4524)

### LINKED ISSUE #4524: [Bug]: DeltaScanNext memory investigation does not measure deletion-vector planning overhead
## What happened?

Parent issue: #4518
Related first fix: #4519

The #4518 scan-planning matrix wires deletion-vector counters, but the synthetic cases did not generate valid DV sidecars or inline DV metadata. All DV counters stayed at zero, so the current investigation cannot say whether DVs contribute meaningful planning memory.

This leaves a gap for DV-heavy tables. DeltaScanNext keeps per-file selection vectors and DV metadata in the scan path, so DV planning should be measured separately before making memory claims for tables with deletion vectors.

## Expected behavior

The scan-planning memory investigation should include a DV-backed table shape that exercises valid deletion-vector metadata and reports DV count, mask bytes, and RSS deltas across planning phases.

## Operating System

Linux

## Binding

Rust

## Bindings Version

`0.32.4` / current development branch

## Steps to reproduce

Extend the #4518 scan-planning probe matrix with valid inline or sidecar deletion vectors, then compare DV-backed and non-DV-backed tables at the same file count.

## Relevant logs

```shell
Current matrix DV counters: 0
DV overhead: not measured
```

sandy-sachin7:
Closes #4524

**Problem:** The #4518 scan-planning probe matrix wired deletion-vector counters, but all synthetic cases used dummy inline DVs (base85 "AAAA") that failed to decode. DV counters stayed at zero, so DV planning overhead was not measured.

**Fix:** Replace the inline-DV approach with the real `WithDvSmall` fixture (sidecar DVs, cardinality 2). Two new unit tests in `exec_meta.rs`:

1. `test_dv_probe_reports_non_zero_dv_metrics` — metadata-only scan of DV-backed table, asserts `dv_count > 0` and `total_mask_bytes > 0`, logs probe diagnostics.
2. `test_dv_probe_short_mask_counter_fires` — executes the plan and verifies the `dv_short_mask_padded_files_total` counter is registered after `DeltaScanMetaStream::new` creates it.

Also removes the unused `make_test_add_with_dv` / `make_test_add_with_short_dv` helpers from `test_utils/mod.rs` that were left over from the failed inline-DV attempt.

**Verification:**
- All 11 scan-planning unit tests pass (9 existing + 2 new)
- `cargo clippy --features datafusion -p deltalake-core --tests --no-deps` — no new warnings
- `cargo fmt` — clean


crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -1171,4 +1171,89 @@ mod tests {
...
+    // ------------------------------------------------------------------
+    // DV scan-planning probe — metrics for deletion-vector planning overhead
+    // ------------------------------------------------------------------
+
+    /// Return total DV mask bytes and DV count for a metadata-only scan.
+    fn probe_dv_metrics(exec: &DeltaScanMetaExec) -> (usize, usize) {
+        let total_mask_bytes: usize = exec
+            .selection_vectors
+            .iter()
+            .map(|entry| entry.value().len() * size_of::<bool>())
```
ethan-tyler, line 1218: This function name is confusing, because it counts logical mask entries but `Vec<bool>` stores bits.  Because it doesn't report memory you'd need to adjust this for full #4524 completion.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -1171,4 +1171,89 @@ mod tests {
...
+        // Execute partition 0 and collect results.
+        let task_ctx = state.task_ctx();
+        let stream = exec.execute(0, task_ctx)?;
+        let batches: Vec<RecordBatch> = stream.try_collect().await?;
+        assert_eq!(batches.len(), 1, "expected one batch from WithDvSmall");
+        let row_count = batches.iter().map(|b| b.num_rows()).sum::<usize>();
+        assert_eq!(row_count, 8, "WithDvSmall: 10 rows minus 2 deleted = 8");
+
+        // Now the counter should be registered.
+        let metric_str = format!("{:?}", exec.metrics());
```
ethan-tyler, line 1285: This tests metric registration but not counter behavior. `WithDvSmall` has a full length mask, which means `padded_rows` stays 0 and this counter should not increment.

ethan-tyler (commented):
Thanks @sandy-sachin7! This is a good regression check, however I don't think it closes #4524. It doesn't wire into the actual scan planning probe matrix or harness. If you want to keep it as is that's fine, we can adjust the description and merge pending the couple of comments I added below.

PR #4561, opened 2026-06-22
fix: enforce non-null map values and list elements in DataValidationExec

### LINKED ISSUE #3690: `delta-rs` does not check nullability of Map values while writing, and fails at query time
# Environment

**Delta-rs version**: `0.27.0`

**Binding**: I am using the Rust library.

***
# Bug

**What happened**:

Assume I have a column called `map`
The schema is,
```
"map": StructField { name: "map", data_type: Map(MapType { type_name: "map", key_type: Primitive(String), value_type: Primitive(String), value_contains_null: false }), nullable: false, metadata: {} }
```

`delta-rs` does not raise any error if I try to ingest `{"a": null}`. It ingests successfully.

Now, at the query time, I will get an error like,
```
ArrowError(InvalidArgumentError("Found unmasked nulls for non-nullable StructArray field \"value\""), None)
```

`delta-rs` should block writing maps with `null` values if the schema does not allow it.

**How to reproduce it**:
There are some helper functions used below, which I have skipped, but the code should be self-explanatory.

```rust
#[tokio::test]
async fn null_to_non_null_column_map() {
    let schema_without_null_column = vec![
        StructField::new("id", DeltaDataType::INTEGER, false),
        StructField::new(
            "map",
            DeltaDataType::Map(Box::new(MapType::new(
                DeltaDataType::STRING,
                DeltaDataType::STRING,
                false,
            ))),
            false,
        ),
    ];

    let non_null_batch = RecordBatch::try_from_iter_with_nullable(vec![
        (
            "id",
            Arc::new(Int32Array::from(vec![1, 2])) as ArrayRef,
            false,
        ),
        (
            "map",
            create_string_map_array(
                &[
                    &Some(HashMap::from([("a".to_string(), "b".to_string())])),
                    &Some(HashMap::from([("a".to_string(), "b".to_string())])),
                ],
                false,
            )
            .unwrap() as ArrayRef,
            false,
        ),
    ])
    .unwrap();

    let null_batch = RecordBatch::try_from_iter_with_nullable(vec![
        (
            "id",
            Arc::new(Int32Array::from(vec![Some(1), Some(1)])) as ArrayRef,
            true,
        ),
        (
            "map",
            create_string_map_array_optional_value(&[
                &Some(HashMap::from([("a".to_string(), None)])),
                &Some(HashMap::from([("a".to_string(), None)])),
            ])
            .unwrap() as ArrayRef,
            false,
        ),
    ])
    .unwrap();

    let ops = DeltaOps::try_from_uri("memory:///").await.unwrap();
    let mut table = ops
        .create()
        .with_table_name("my_table")
        .with_save_mode(SaveMode::ErrorIfExists)
        .with_columns(schema_without_null_column)
        .await
        .unwrap();

    table = DeltaOps::from(table.clone())
        .write(vec![non_null_batch.clone()])
        .with_schema_mode(SchemaMode::Merge)
        .with_save_mode(SaveMode::Append)
        .await
        .unwrap();

    let schema = table.get_schema().unwrap();
    println!("schema: {:?}\ndata: {:?}", schema, non_null_batch);

    // https://github.com/delta-io/delta-rs/blob/ef9e077402013bedf44fccf872f7d7d4fd9157ab/crates/core/src/delta_datafusion/mod.rs#L1425
    // delta does not check for nullability of nested columns.
    table = DeltaOps::from(table)
        .write(vec![null_batch.clone()])
        .with_schema_mode(SchemaMode::Merge)
        .with_save_mode(SaveMode::Append)
        .await
        .unwrap();

    let schema = table.get_schema().unwrap();
    println!("schema: {:?}\ndata: {:?}", schema, null_batch);

    ctx.register_table("simple_table", Arc::new(table.clone()))
        .unwrap();
    let df = ctx.sql("SELECT * FROM simple_table;").await.unwrap();
    df.show().await.unwrap();
}
```

**More details**:

raphaelroshan:
# Which issue does this PR close?

Closes #3690.

# Rationale for this change

Non-nullable map values and list elements weren't actually being enforced on write. `validation_predicates` generated a `col IS NOT NULL` check for every non-nullable field. That's the right check for scalar fields (one value per row), but it doesn't work for collections:

- For a map value, `field("value").is_not_null()` lowers to `get_field(map, "value")` — a lookup of the entry whose *key* is `"value"` — so it never looked at the actual values.
- List elements weren't handled at all.

As a result, a null in a non-nullable map value (or list element) wrote successfully and only blew up later at read time with `Found unmasked nulls for non-nullable StructArray field "value"`, which is the failure reported in #3690.

The distinction that matters is scalar vs. collection: a struct field is a single value per row, so a scalar `IS NOT NULL` is correct; a map value or list element is a collection per row, so it needs an "are any of the elements null?" check instead.

# What changes are included in this PR?

- Validation now chooses the predicate based on the field kind:
  - scalar / nested struct field → `col IS NOT NULL` (unchanged)
  - list element → `array_length(col) IS NOT DISTINCT FROM array_length(array_compact(col))` — compacting drops nulls, so an unchanged length means no null elements
  - map value → the same check applied to `map_values(col)`
- `collect_non_nullable_fields` now also records non-nullable list elements (previously only structs and maps were traversed).
- Collection-element checks are derived from the table schema rather than the table-vs-source difference. The source's declared element nullability is exactly what's unreliable here — #3690's repro declares the value non-nullable while the data contains nulls — so trusting it would skip the check. Scalar fields keep the existing source-difference optimization.

# Are these changes tested?

Yes. New unit tests in `data_validation.rs` cover the #3690 repro (null map value is rejected), a valid map (accepted — the case that originally surfaced the `get_field` bug), null/valid/empty/all-null lists, and non-nullable nested struct fields (rejected/accepted).

# Are there any user-facing changes?

Writes that previously succeeded with a null in a non-nullable map value or list element now fail with a validation error at write time, rather than producing a table that errors on read. No public API changes.

One known limitation: a non-nullable field nested *past* a list or map element (e.g. a required field inside a list of structs) still can't be expressed as a per-element check, so those paths are skipped — same behavior as before this PR.


ion-elgreco:
@raphaelroshan validation of data should not occur in the write layer, we do this work in a datafusion node `DataValidationExec`. See here since we already do non_nullable validation https://github.com/delta-io/delta-rs/blob/0ec5175f37ffc3b3482254e806574a1c5bb46a18/crates/core/src/delta_datafusion/data_validation.rs#L269-L272, a fix should be there

raphaelroshan:
Thanks @ion-elgreco for the feedback - I checked again and it seems like the listed problem in https://github.com/delta-io/delta-rs/issues/3690 is fixed, I could help extend it for list checking if that helps (didnt see lists being handled). Will close this otherwise

ion-elgreco:
@raphaelroshan feel free to just create test cases then and link the issue to be closed once we merge the tests pr

raphaelroshan:
@ion-elgreco I realised while adding tests that it works for nested structs, but not for lists and maps, so I added back my fixes :"), and test it more clearly so its clear that everything works properly

raphaelroshan:
@ion-elgreco , it's ready for review, do you mind taking a look when you have time?

ethan-tyler (dismissed):

Issue #4528, opened 2026-06-11
[Bug]: Optimize dropping partition columns when partition_filters contains a subset of filters


anon-corpo-slave:
### What happened?

On a table partitioned by multiple columns, running optimize.compact(...) or optimize.z_order(...) with partition_filters that reference only some of the partition columns merges files across the unreferenced partition columns. The output Add.partitionValues keeps only the filtered columns — the rest become null, records from different partitions are merged into one file, and the file is written to a flattened hive path. It's silent data corruption.

The rule observed: the filter must reference every partition column, or none — any strict subset drops the unreferenced ones. 

### Expected behavior

partition_filters is a file-selection filter only. Compaction/z-order must merge files within a single full partition tuple and preserve all partition-column values in the output, regardless of which columns the filter references.

### Operating System

None

### Binding

None

### Steps to reproduce

```python
import tempfile, pyarrow as pa
from deltalake import DeltaTable, write_deltalake

path = tempfile.mkdtemp() + "/tbl"
def batch(part_a, s=0, n=5):
    return pa.table({
        "part_a": pa.array([part_a] * n, pa.int64()),   # partition col that gets dropped
        "part_b": pa.array(["x"] * n, pa.string()),      # the filtered partition col
        "val":    pa.array(list(range(s, s + n)), pa.int64()),
    })

write_deltalake(path, batch(1), partition_by=["part_a", "part_b"], mode="overwrite")
for a in (1, 2, 3):
    for k in range(3):
        write_deltalake(path, batch(a, 100 * k), partition_by=["part_a", "part_b"], mode="append")

# filter references ONLY part_b (a subset of [part_a, part_b])
DeltaTable(path).optimize.compact(partition_filters=[("part_b", "=", "x")])

dt = DeltaTable(path)
adds = dt.get_add_actions(flatten=True).to_pandas()
print("part_a=NULL files:", int(adds["partition.part_a"].isnull().sum()))
print(sorted({u.split("/tbl/")[-1].rsplit("/", 1)[0] for u in dt.file_uris()}))
# Buggy:    part_a=NULL files: 1 ; ['part_b=x']
# Expected: part_a=NULL files: 0 ; ['part_a=1/part_b=x', 'part_a=2/part_b=x', 'part_a=3/part_b=x']
```

### Relevant logs

```shell

```

PR #4564, opened 2026-06-24
feat(core): expose `LogicalFileView::partition_values_map` as public API

### LINKED ISSUE #4563: [Feature]: Expose `LogicalFileView::partition_values_map()` as public API

`LogicalFileView::partition_values()`  is narrowed to the partition columns referenced by the scan predicate (the selective-stats / data-skipping projection). Connectors that enumerate files via `Snapshot::file_views(predicate)` and need each file's complete partition tuple therefore get incomplete results whenever the predicate references only a subset of partition columns. The same method returns different values for the same file depending on the scan predicate, and there is no public accessor that returns the full partition map.

Make the existing `LogicalFileView::partition_values_map()` public. It already returns the complete `HashMap<String, Option<String>>` from the raw `partitionValues` column and is not narrowed by the predicate; it is simply not public today. This is a minimal, additive visibility change with no behavioral impact on existing callers.

All current ways to obtain the complete partition map are unsatisfactory: `add_action().partition_values` is `#[deprecated(since = "0.31.0")]` and also builds a full `Add` (including `stats()`) per file; `remove_action(false).partition_values` is not deprecated but semantically wrong for a read (it fabricates a `deletion_timestamp`); and `full_partition_values()` (added in #4556) is both `pub(crate)` and `#[cfg(feature = "datafusion")]`, so it's unreachable for connectors built with `default-features = false`.

### Priority

### Additional context

This is the read-path analogue of #4528 / #4556, which fixed the optimize/write path by reading the raw `Add.partitionValues` map in table-metadata order. The read-path narrowing of `partition_values()` is unchanged by that PR, so external consumers of `file_views` still have no complete-partition accessor. We hit this in an engine that reads Delta tables via `file_views` with partition predicates: a filter on a subset of partition columns (e.g. `year=2020/month=*` on a `year/month/day` table) dropped the unreferenced columns. We're carrying a one-line fork patch (`fn -> pub fn`) plus a regression test in the meantime.

### Contribution

BoazC-MSFT:
# Description

`LogicalFileView::partition_values()` returns only the partition columns referenced by the scan predicate - the selective-stats / data-skipping projection narrows `partitionValues_parsed` to the predicate's columns. Connectors that enumerate files via `Snapshot::file_views(predicate)` and also read each file's partition values therefore lose the unreferenced partition columns whenever the predicate touches a strict subset of them (e.g. a `year = 2020` filter on a table partitioned by `year/month/day` drops `month` and `day`). The same method returns different values for the same file depending on the scan predicate, and there is no public accessor for the complete set.

The complete map already exists internally as `LogicalFileView::partition_values_map()`, which reads the raw `partitionValues` and is not narrowed by the predicate; it is simply private. This change makes it `pub`.

This is the read-path analogue of the optimize/write-path fix in #4556: that PR reconstructs the full partition identity from the raw `Add.partitionValues` map, but `partition_values()` on the read path remains narrowed, leaving external `file_views` consumers without a complete-partition accessor. The remaining alternatives are all unsatisfactory - `add_action().partition_values` is `#[deprecated(since = "0.31.0")]` (and builds a full `Add`, including stats, per file), `remove_action(false).partition_values` is semantically wrong for a read, and `full_partition_values()` is `pub(crate)` and gated behind the `datafusion` feature, so it is unreachable for connectors built with `default-features = false`.

# Related Issue(s)
- closes #4563
- Related: #4528 / #4556 (same root cause on the optimize/write path)

# Documentation
No new documentation. The existing rustdoc on `partition_values_map()` already documents that it preserves all partition columns even when `partitionValues_parsed` is narrowed for data skipping.

### DIFF
--- a/crates/core/src/kernel/snapshot/iterators.rs
+++ b/crates/core/src/kernel/snapshot/iterators.rs
@@ -216,7 +216,7 @@ impl LogicalFileView {
     ///
     /// This preserves all partition columns even when `partitionValues_parsed` was narrowed to the
     /// predicate-referenced subset for data skipping.
-    fn partition_values_map(&self) -> HashMap<String, Option<String>> {
+    pub fn partition_values_map(&self) -> HashMap<String, Option<String>> {
         self.raw_partition_values()
             .filter(|partitions| partitions.is_valid(self.index))
             .and_then(|partitions| collect_string_map(&partitions.value(self.index)))
--- a/crates/core/src/kernel/snapshot/mod.rs
+++ b/crates/core/src/kernel/snapshot/mod.rs
@@ -2602,6 +2602,63 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_partition_values_map_preserves_all_columns_under_narrowing_predicate()
+    -> TestResult {
+        use std::collections::BTreeSet;
+
+        // A predicate that references only a subset of the partition columns (year + month, but
+        // not day) makes the scan narrow `partition_values()` to the referenced columns for data
+        // skipping. `partition_values_map()` must still return every partition column so connectors
+        // can reconstruct the full partition identity of each file.
+        let base = TestTables::Delta0_8_0Partitioned
+            .table_builder()?
+            .build_storage()?;
+        let snapshot = Snapshot::try_new(base.as_ref(), Default::default(), None).await?;
+
+        let predicate = Arc::new(to_kernel_predicate(
+            &[
+                PartitionFilter::try_from(("year", "=", "2020"))?,
+                PartitionFilter::try_from(("month", "=", "2"))?,
+            ],
+            snapshot.schema().as_ref(),
+        )?);
+
+        let views: Vec<_> = snapshot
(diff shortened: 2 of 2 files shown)


ethan-tyler (approved):
Thanks @BoazC-MSFT - this is a reasonable direction. Make sure you add your DCO sign off on the commits and we should be good to merge.

BoazC-MSFT:
> Thanks @BoazC-MSFT - this is a reasonable direction. Make sure you add your DCO sign off on the commits and we should be good to merge.

Thanks @ethan-tyler. Happy to sign off - what's the preferred way given you've merged main in?

Also, what needs to be done so that this change is included in a coming 0.32.x coming release?

ethan-tyler:
Of course! The easiest way to do it is to include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Once we merge it will be included in the next rust release which should be pretty soon.

PR #4567, opened 2026-06-25
refactor: reduce internal EagerSnapshot usage

ethan-tyler:
This removes internal `EagerSnapshot` dependencies where the code can use `Snapshot`, `Arc<Snapshot>`, active add APIs, tombstone APIs, `DeltaScanNext`, or `FileSelection` directly. Public compatibility stays intact.

### DIFF
--- a/crates/core/src/operations/write/plan.rs
+++ b/crates/core/src/operations/write/plan.rs
@@ -34,8 +34,8 @@ use crate::delta_datafusion::{
 use crate::errors::{DeltaResult, DeltaTableError};
 use crate::kernel::schema::cast::{merge_arrow_schema, normalize_for_delta};
 use crate::kernel::{
-    Action, ActiveAddOptions, Add, AddStatsPolicy, DeletionVectorDescriptor, EagerSnapshot,
-    Metadata, ProtocolExt as _, Remove, StructType, StructTypeExt,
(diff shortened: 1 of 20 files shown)


crates/core/src/operations/write/plan.rs
```diff
@@ -1001,7 +999,13 @@ mod tests {
             .unwrap();
 
         let prepared_append = prepare_write(WritePreparationInput {
-            snapshot: Some(table_with_snapshot.snapshot().unwrap().snapshot()),
+            snapshot: Some(
```
hntd187: Now with even more `.snapshot()` !!

hntd187 (commented):
I'll be honest, I think this may shift the usage instead of actually reducing it. There are a lot of places where there was no `.snapshot()` and now there is a call and a lot of places where there are now more `.snapshot()` on top of the prior call sites. 

I think this needs a bit more re-thinking in general and I think we should just rip the band-aid off. Why did we create different kinds of snapshots in the first place? Because the implementation details are different right? If that is the case is this not something we can collate into a trait and just pass a single "Snapshot" around instead of having to do this snapshot().snapshot().snapshot() things, which seems really ugly. If the different kinds of snapshots are so different they cannot be represented in a single trait then let's discuss more.

ethan-tyler:
going to keep this as a draft for now with it likely being superseded. A new issue was created to track a more robust snapshot design and future state here #4584.

PR #4568, opened 2026-06-26
refactor: consolidate parquet read and write paths behind traits

corwinjoy:
# Description
Right now, there are several duplicated read and write paths for creating parquet files within delta-rs. The purpose of this PR is to consolidate these paths behind a more abstract interface in two parts:
1. A basic interface that needs only the parquet library.
2. An extended interface using datafusion and supporting advanced features such as predicates and projections.

There are a few reasons for doing this:
1. Code maintenance. Having (mostly) a single path for reads and writes helps reduce maintenance and makes it easier to control behavior around how underlying reads and writes are handled.
2. Consolidating these paths makes it a lot easier to attach advanced parquet properties to reads and writes. Examples would include future encryption support and content defined chunking as discussed in https://github.com/delta-io/delta-rs/pull/4465
3. I believe that having the simplified read and write paths will also help with delta-rs kernel integration.

This PR follows from the ideas discussed at the deltabricks summit and the slack channel at:
https://delta-users.slack.com/archives/C013LCAEB98/p1781571891482589

AI notes: This PR was created with (and reviewed with) the help of AI. However, the design and individual pieces were reviewed and tested by me. I believe them to be correct and an improvement to the project.

## Behavior changes in the legacy writers

The `JsonWriter`/`RecordBatchWriter` rewrite (now streaming through the consolidated write path) intentionally changes a few observable behaviors — flagging them explicitly for reviewers:

1. **All-or-nothing flush windows.** Batches stream to storage as they are written, and a flush window commits atomically: an IO error on any write discards *every* batch buffered since the last flush (the aborted uploads are invisible to vacuum; finalized files are left for vacuum). The old writers buffered in memory and lost only the failing batch. Documented on both writers.
2. **`JsonWriter` no longer quarantines bad records.** A write containing malformed JSON records now rejects the whole call and writes nothing, instead of writing the good records and returning `PartialParquetWrite` with the skipped values. (The error enum was already crate-private, so this is a behavior change, not an API break.)
3. **`write_partition` + `MergeSchema`:** unpartitioned widening now updates the writer's schema, and a later `flush_and_commit` commits the metadata evolution; widening on a partitioned table is rejected (fixes the class of bug in #3783 where files could carry columns absent from table metadata).
4. **Counter semantics:** `buffered_record_batch_count` counts write calls (previously one per partition touched per call), and `buffer_len` approximates encoded parquet size written this window (previously in-memory buffer bytes). Documented on the methods.


corwinjoy:
## Key changes

- Relocate the already-DataFusion-free `DeltaWriter` / `PartitionWriter` out of the `#[cfg(feature = "datafusion")]`-gated `operations/write` module into non-gated `datafile::writer` — the modern writer is now usable without DataFusion.
- Route the modern write path (`execution.rs`) through the basic `DeltaDataWriter` / `write_plan`, deleting ~135 lines of bespoke mpsc fan-in while preserving the scan/write metric split.
- Re-route legacy `RecordBatchWriter` / `JsonWriter` through the shared `DeltaWriter` sink, removing their duplicated partitioning / parquet / stats logic. The writers now **stream** each batch into the sink (encoding incrementally), and a schema-widening `MergeSchema` write auto-flushes the current file and continues under the widened schema. `MergeSchema` schema evolution is preserved at the legacy layer; the `JsonWriter` per-record parquet-encode quarantine is **not** (see Behavior changes).
- Read traits: basic `KernelDataReader` / `KernelDataFileReader` are stubs for a follow-up kernel-backed implementation; the DataFusion read extension wraps the existing `DeltaScanNext`.
- `ReaderProperties` centralizes parquet read-option construction across all `ParquetSource` sites.

## Behavior changes

  - **`JsonWriter` bad-record quarantine removed.** This only ever fired on a record that decodes to valid Arrow but *fails to parquet-encode* — an extremely rare case. On such a `ParquetError` the old `write()` fell back to a record-by-record re-encode (`quarantine_failed_parquet_rows`) to isolate the bad rows, skipped them, and reported them via  `DeltaWriterError::PartialParquetWrite { skipped_values }` — a lazy on-failure fallback, not a double-encode, and the flush ended in an error either way.

    The new trait waist streams batches into the shared sink and no longer exposes a per-batch parquet error to hook into, so keeping this would mean either always encoding twice or pushing a "lenient write" mode into the file-tier writer — not worth it for so rare a case (the blocker is the missing hook, not performance). A bad record now fails the whole flush instead of being skipped; JSON decode and schema-mismatch errors are still reported per `write`. The one thing lost is **diagnostics** — `skipped_values` told you *which* records were bad — recoverable by writing one batch per `flush`. The `PartialParquetWrite` variant and quarantine helpers are removed.

- **`operations::write::writer` path deprecated.** The writer types moved to `datafile::writer`; `operations::write::writer` remains as a deprecated re-export for out-of-tree back-compat.

corwinjoy:
In addition, a new write benchmark was added to test and tune the performance of the new streaming write path.

## Write performance vs `main`

Divan benchmark (`cargo bench -p delta-benchmarks --bench write`, **median of 25 samples**), `main` (`0ec5175f`) vs this branch, both `--release`, run on an idle machine; only the write + commit is timed. Allocation totals from Divan's `AllocProfiler`. (Measured on the final branch, including the two write-path micro-optimizations noted below.)

**High-level `DeltaTable::write` (DataFusion path)** — unchanged within run-to-run noise (this path is not altered by the writer changes):

| scenario (few large batches) | main | this PR | Δ time |
|---|---|---|---|
| 1M, unpartitioned | 217 ms | 221 ms | +2% |
| 1M, 32 partitions | 93 ms | 98 ms | +5% |
| 10M, unpartitioned | 1.531 s | 1.543 s | +1% |
| 10M, 32 partitions | 441 ms | 456 ms | +3% |

**`RecordBatchWriter` (direct parquet path)** — faster and lower-allocation on **7 of 8** scenarios; the lone regression is the single smallest unpartitioned case:

| scenario | main | this PR | Δ time | main alloc | PR alloc |
|---|---|---|---|---|---|
| 1M, unpart, few large | 161 ms | 175 ms | +9%\* | 72.3 MB | 65.8 MB |
| 1M, unpart, many small | 164 ms | 157 ms | −5% | 72.5 MB | 66.1 MB |
| 1M, 32 part, few large | 395 ms | 363 ms | −8% | 176.9 MB | 149.9 MB |
| 1M, 32 part, many small | 460 ms | 444 ms | −3% | 183.4 MB | 156.7 MB |
| 10M, unpart, few large | 1.592 s | 1.409 s | −11% | 909.6 MB | 660.6 MB |
| 10M, unpart, many small | 16.06 s | 1.452 s | **−91%** | **33.6 GB** | **664 MB** |
| 10M, 32 part, few large | 3.604 s | 3.40 s | −6% | 1.474 GB | 1.403 GB |
| 10M, 32 part, many small | 3.669 s | 3.50 s | −5% | 1.545 GB | 1.476 GB |

\* By fastest-sample (noise-robust on these sub-200 ms writes) this is ~+1% (154 ms vs 153 ms); the median carries run-to-run variance.

Takeaways:
- **The direct path is faster and far leaner across the board.** Streaming each batch into the sink removes the old double-buffer (raw batches *and* encoded output held at once), cutting allocations ~8–15% in normal cases and ~50× for many small batches.
- **The old buffer-then-encode-at-flush path collapses on many small batches:** 10M unpartitioned rows churned **33.6 GB** of allocations over **16 s**; streaming holds that to **664 MB / 1.45 s**.
- **The lone residual cost is the smallest unpartitioned single-file write (1M rows, few large batches): ~+9% median / ~+1% fastest.** It vanishes as data grows (−11% at 10M) and is absent for partitioned and many-small writes. Root-caused below.

### Small-write regression: root cause and the trade-off

Profiling the 1M-unpartitioned case (phase timing + `perf`) traced the gap to a deliberate difference in how the encoded parquet reaches storage:

- **`main`'s legacy `RecordBatchWriter`** encodes into an **in-memory buffer** (`ArrowWriter<ShareableBuffer>`, synchronous) and uploads the whole buffer once at `flush`. It also copies that growing buffer on every `write()` for crash recovery (`buffer.to_vec()`).
- **This PR's shared sink** encodes through an **`AsyncArrowWriter` that streams parquet to object storage as it goes**, via a multipart `ParquetObjectWriter`/`BufWriter`.

The `perf` profile shows the write time is dominated by the *identical* parquet encode (SNAPPY compression, string-dictionary interning, RLE), and the sink's dispatch code has negligible self-time. The residual delta at 1M is the **fixed cost of the async streaming-upload path** — multipart buffer setup, async flushing during the write, and multipart finalize at close.

That same design is **why the PR is faster (−11%) and far leaner at 10M**: it never materializes the whole encoded file in memory and skips `main`'s O(file-size) per-`write()` recovery copy (which is what makes `main` churn 33.6 GB / 16 s on the 10M many-small case).

**Trade-off chosen — keep the streaming-to-storage sink.** The overhead is small (a few ms, ~+1% by fastest), only visible on the smallest single-file writes, and is the direct enabler of the large-write throughput and memory wins — the data-lake common case and the point of the consolidation. We deliberately do **not** special-case an in-memory path for small writes, since that would re-introduce the memory cost the streaming design removes. Two minor optimizations were applied along the way (an unpartitioned fast path that skips the per-batch partition split/projection, and writing whole batches instead of 1024-row slices when no target file size is set — both with identical output), which is what flipped the 1M many-small and partitioned cases to net-faster.

corwinjoy:
In reviewing this I would suggest beginning with `crates/core/src/datafile/mod.rs` which implements the trait we discussed and then `crates/core/src/datafile/datafusion_ext.rs` which is the datafusion extension. Much of the remaining PR is refactoring code to use these traits + tests.

corwinjoy:
@rtyler @roeap @adamreeve

crates/core/src/datafile/reader.rs
```diff
@@ -0,0 +1,293 @@
...
+        Err(err) => {
+            let failing: BatchFuture = Box::pin(async move { Err(err) });
+            futures::stream::once(std::future::ready(failing)).boxed()
+        }
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Placeholders for the future kernel-backed (full-fidelity) reader.
+// ---------------------------------------------------------------------------
```
corwinjoy, line 196: I left this as an outline for where a future delta kernel implementation could go.

crates/core/src/datafile/reader.rs
```diff
@@ -0,0 +1,293 @@
...
+        if let Some(size) = size {
+            reader = reader.with_file_size(size);
+        }
+        let stream = ParquetRecordBatchStreamBuilder::new(reader)
+            .await?
+            .build()?;
+        Ok(results_to_future_stream(stream))
+    }
+}
+
```
corwinjoy, line 76: The current code doesn't have a pure parquet reader but I felt it made sense to add for two reasons.
1. To prove out the design and make sure the traits were correct.
2. There is utility in having a fast read path for delta tables in addition to a fast write path.

crates/core/src/writer/json.rs
```diff
@@ -319,160 +202,60 @@ impl DeltaWriter<Vec<Value>> for JsonWriter {
...
-                    ParquetError::EOF(msg) => ParquetError::EOF(msg.to_owned()),
-                    ParquetError::External(err) => ParquetError::General(err.to_string()),
-                    ParquetError::IndexOutOfBound(u, v) => {
-                        ParquetError::IndexOutOfBound(u.to_owned(), v.to_owned())
-                    }
-                    ParquetError::NYI(msg) => ParquetError::NYI(msg.to_owned()),
-                    // ParquetError is non exhaustive, so have a fallback
-                    e => ParquetError::General(e.to_string()),
-                },
-                skipped_values: partial_writes,
```
corwinjoy, line 362: We tried to see if there was any reason for the good,bad row split in the JSON writer and couldn't find any good explanation for this. It was present in the original version of JSON writing added in [github.com/delta-io/delta-rs/pull/573](https://github.com/delta-io/delta-rs/pull/573)
As mentioned in the PR description, with the more abstract streaming / trait design it made sense to report this error at the batch level instead. If a finer level of detail is needed, the batch size can be set to one. Also, it is hard to visualize cases where the Arrow encoding succeeds but the parquet encoding does not.

corwinjoy:
In addition, writing under schema changes takes some careful logic here. In the process, I believe this PR fixes some latent bugs. From Claude:

---

  ### Schema evolution: behaviour vs. the previous (in-memory) writer

  The legacy `RecordBatchWriter` / `JsonWriter` move from an in-memory buffer
  (`ArrowWriter<ShareableBuffer>`, encoded at `flush`) to a streaming sink
  (`AsyncArrowWriter` → multipart object-store upload). This changes how
  schema-evolving (`MergeSchema`) writes produce data files, and along the way
  fixes two latent bugs in the old path.

  **Unchanged (observable behaviour)**

  - `MergeSchema` + partition columns → unsupported error.
  - `Default` write whose schema doesn't match → `SchemaMismatch`.
  - `MergeSchema` that changes an existing column's *type* → rejected.
  - After evolution the table schema gains the new column; old rows read it back as null.

  **Changed — physical file layout (read-equivalent).** A `MergeSchema` write that
  widens the schema now **seals the current file and rolls a new one** under the
  widened schema, instead of buffering everything and rewriting it into a single
  file at flush. A flush window can therefore emit files with different *physical*
  schemas: older files omit the new column (read back as null via schema-on-read),
  newer files include it. The committed table schema and query results are
  identical — only the on-disk layout differs.

  **Changed — error model.** A failed batch write aborts the whole flush window
  (the in-progress multipart upload can't be rolled back) rather than recovering an
  in-memory buffer. A failed *log commit* retains the staged files so it can be
  retried without re-uploading.

  **Bugs fixed in the old writer**

  1. **Data loss on schema evolution.** The old writer encoded the widened batch
     through an `ArrowWriter` built with the *pre-widening* schema; arrow-rs
     silently accepts the wider batch but writes only the original columns, so the
     **new column's data was dropped** — the table schema evolved while the column
     read back all-null. (The old tests checked only metadata.) The streaming
     writer encodes under the *merged* schema, preserving the data;
     `test_merge_schema_preserves_new_column_data` reads the data back to lock this
     in.
  2. **Dropped evolution.** A `should_evolve` flag, overwritten on every write,
     gated metadata evolution — so a `MergeSchema` widening followed by a `Default`
     write silently failed to evolve the table schema. Evolution is now keyed on the
     schema diff itself.

  **Justification.** Streaming removes the dual in-memory buffer (raw batches *and*
  encoded output), the source of the large-write memory/throughput wins in the
  benchmark table; the resulting mixed-file layout is read-equivalent under Delta's
  schema-on-read; and the rewrite eliminates the silent data-loss bug above.

ion-elgreco:
@corwinjoy I glanced over it a bit, two thoughts that come to mind:

- We could perhaps first have a PR introducing the trait and only applying a trait impl for our main writer that is used by all our operations. The json writer and such are older writers that are sparsely used, I think those are better to be targeted as followups
- I also suggest also prompting claude to "tighten the comments and docs" opus 4.8 especially tends to be overly aggressive with it's 'what' comments

corwinjoy:
> @corwinjoy I glanced over it a bit, two thoughts that come to mind:
> 
> * We could perhaps first have a PR introducing the trait and only applying a trait impl for our main writer that is used by all our operations. The json writer and such are older writers that are sparsely used, I think those are better to be targeted as followups
> * I also suggest also prompting claude to "tighten the comments and docs" opus 4.8 especially tends to be overly aggressive with it's 'what' comments

Thanks @ion-elgreco. For #1, I  would like to simplify this, but I am not sure if splitting this PR into pieces makes this easier or harder to review, since the point is to consolidate the logic. I think I will hold off for now and see what @roeap  and @rtyler say.

The tip for #2, is very helpful thanks! I've struggled a bit with getting Claude to have more succinct comments, and that request seems to have improved things.

ethan-tyler:
I'm going to take a look at this week.

@corwinjoy This is obviously pretty important functionality, I know AI can sometimes miss some nuanced things and just want to make sure we're covering our bases.

crates/core/src/datafile/writer.rs
```diff
@@ -219,56 +251,81 @@ impl DeltaWriter {
...
 
-        match self.partition_writers.get_mut(&partition_key) {
-            Some(writer) => {
-                writer.write(&record_batch).await?;
-            }
-            None => {
-                let prefix_override = match self.config.random_prefix_length {
-                    Some(length) => Some(Path::parse(random_prefix(length))?),
-                    None => None,
-                };
```
corwinjoy, line 250: The prefix_override logic got moved as part of the refactor. Might be worth adding an explicit test for this logic.

---

The prefix logic that main's write_partition computed inline at operations/write/writer.rs:247 now lives in build_partition_writer at
  `crates/core/src/datafile/writer.rs:266-268`:
```
  let prefix_override = match self.config.random_prefix_length {
      Some(length) => Some(Path::parse(random_prefix(length))?),
      None => None,
  };
```

corwinjoy:
Additional commits.
1. Added fixes for potential errors identified by Claude Fable. The most important gap was better `abort()` handling.
2. Improve bad record reporting for the JSON writer.

corwinjoy:
> I'm going to take a look at this week.
> 
> @corwinjoy This is obviously pretty important functionality, I know AI can sometimes miss some nuanced things and just want to make sure we're covering our bases.

Sounds good and thanks! I've tried my best to follow the design outlined by @rtyler and @roeap and review carefully but there is no substitute for someone with deeper expertise with the original code. Plus, it's definitely worth being careful here.

crates/core/src/writer/json.rs
corwinjoy: Amended. We added a later commit to improve the failed record reporting here.

corwinjoy:
Following the suggestion from @ion-elgreco, Claude has offered to split up this PR. However, I am holding off for now since I think it makes sense to see the whole picture first. Then, if folks think it helps, I am happy to split this into pieces if that makes life easier.

From Claude:

---

  ## Proposed split
  
  The cut line is: PR 1 = everything the main DeltaOps write/read path uses; PR 2 = the legacy JsonWriter/RecordBatchWriter rewrite and everything only it consumes.

  Following up on the suggestion to break this PR up: after looking at where the risk
  actually concentrates, I think a two-way split works, with one sharpening of the
  original suggestion — the follow-up shouldn't be "add trait impls to the JSON writer
  later" but **defer the entire legacy-writer rewrite**. The line counts and review
  findings both support cutting there: of the ~2,900 inserted lines, roughly 2,100 are
  the `JsonWriter`/`RecordBatchWriter` rewrite, and nearly all of the behavioral
  changes found in review were in that piece. The datafile module and main-writer
  migration are mostly code motion plus additive API.

  ### PR 1 — datafile module, traits, main-path consolidation (~1,600 lines)

  * `datafile/mod.rs` — the traits and stream plumbing: `DataFileWriter`,
    `DataFileReader`, `DeltaDataWriter`, `DeltaDataReader`, `ReadOptions`,
    `BatchFuture` / `RecordBatchFutureStream`, `results_to_future_stream`.
  * `datafile/writer.rs` — the move of `DeltaWriter` / `PartitionWriter` /
    `WriterConfig` / `PartitionWriterConfig` from `operations/write/writer.rs`, plus
    `write_batches_timed`, the `write_all` impl, `LazyArrowWriter`, and the partition
    routing dedup (`build_partition_writer`, `write_keyed`, `write_unpartitioned`).
    The two `target_file_size = None` fixes (0-row guard, row-group slicing) ride
    here — inert correctness guards with no legacy dependency.
  * `datafile/reader.rs`, `properties.rs`, `datafusion_ext.rs` —
    `ParquetTableReader` / `ParquetFileReader`, `ReaderProperties`,
    `DataFusionDataReader`, `write_plan` / `scan`, `coalesce_provider_scan`,
    `projection_indices`.
  * Wiring: `operations/write/execution.rs` (`write_streams`), `load.rs`,
    `load_cdf.rs`, `scan/mod.rs`, `data_sink.rs`, `lib.rs`, and the benchmarks crate.
  * `writer/{json,record_batch,mod}.rs` **stay at main untouched** — the old buffered
    writers keep working exactly as they do today, so PR 1 changes no behavior for
    anyone. (One seam to double-check during extraction: `datafile/writer.rs` imports
    the pre-existing `divide_by_partition_values` from `writer/record_batch.rs`.)

  ### PR 2 — legacy writers over the streaming window (~2,100 lines)

  * `writer/window.rs` (entire new file): `WriteWindow` and `SinkFactory` — `write`,
    `write_partition`, `seal`, `drain`, `stage`, `committed`, `abort`,
    `buffered_size`, `schema_evolved`, `set_target_file_size`,
    `set_writer_properties`.
  * `writer/json.rs` rewrite: `from_parts` / `read_stats_config`, the streaming
    `write_with_mode`, `invalid_records_error`, the new `reset` / `buffer_len` /
    `with_target_file_size` semantics, and the deletions (old `DataArrowWriter`,
    `quarantine_failed_parquet_rows`).
  * `writer/record_batch.rs` rewrite: `from_parts`, the two-phase `write_with_mode`,
    `conform_to_schema`, `non_nullable_violation_error`, the new `write_partition`
    (MergeSchema widening), `flush` / `flush_and_commit` via `drain` / `stage` /
    `committed`, and deletion of the old in-memory `PartitionWriter`.
  * `writer/mod.rs`: removal of `DeltaWriterError::PartialParquetWrite`,
    `INVALID_PREVIEW_CAP`.
  * The abort machinery in `datafile/writer.rs` — `LazyArrowWriter::abort`,
    `PartitionWriter::abort`, `DeltaWriter::abort` / `abort_detached`, and the
    `rolled_bytes` size tracking. These are only consumed by `WriteWindow`, and
    deferring them keeps PR 1 free of dead `pub(crate)` code.
  * The bad-record identification commit (d45fb141) rides here.

  ### Why this cut

  PR 1 delivers the stated goals of the refactor (single main-path read/write
  implementation, seams for encryption and kernel work) with essentially no
  compatibility risk. PR 2 isolates every user-visible semantic change — streaming
  uploads during `write()`, all-or-nothing flush windows, `reset()` aborting
  multipart uploads, `buffer_len` semantics, MergeSchema behavior — into one focused
  review with its own release notes, and it reverts cleanly if a downstream ingest
  pipeline hits a behavior change. The legacy writers are the sparsely-used surface,
  so deferring them costs little.

  One note: the "~2,900 inserted lines" figure includes the review-fix commits already on the branch; adjust if you post this after further changes.

corwinjoy:
Further review indicated that there were still some subtle bugs that happen with schema changes under partitions. I have pushed a commit to fix this and error early on schema mismatch so that the existing batch window is not affected. (See https://github.com/delta-io/delta-rs/pull/4568/commits/eded719322feb91fd17ddf5c73fc93e49e15faa3) This does introduce a modest behavior change in this PR.  I've also had Claude add tests to more thoroughly check what happens under schema changes. Here is the description from Claude (I have not verified the behavior it reports for main):

---

  ## Behavior note: `write_partition` + `MergeSchema` on partitioned tables

  This PR changes what happens when a caller invokes
  `RecordBatchWriter::write_partition(batch, values, WriteMode::MergeSchema)` with a
  batch that would **widen the schema** of a **partitioned** table:

  * **On `main`**: the call silently succeeds. The per-partition writer merges the
    schema and writes the parquet file with the extra column, but `flush_and_commit`
    never emits a `Metadata` action (only `write_with_mode` sets the evolution flag).
    The result is a committed data file whose schema disagrees with the table schema —
    the extra column's data is silently dropped on read. This is the known #3783
    data-mismatch behavior.

  * **With this PR**: the call fails immediately with
    `Generic("Merging Schemas with partition columns present is currently unsupported")`
    — the same error `write_with_mode` and `flush_and_commit` already raise for this
    case. The check lives in `WriteWindow` (rejecting any schema widening when
    partition columns are present) and runs **before any window state changes or
    bytes are uploaded**, so previously buffered batches remain flushable and the
    writer is left fully usable.

  Non-widening `MergeSchema` calls on partitioned tables (e.g. a batch that matches
  the schema once its partition columns are stripped) continue to succeed, and
  schema evolution on **unpartitioned** tables is unaffected.

  In short: a call pattern that previously appeared to work but silently produced
  inconsistent table state now surfaces a clear, early error. Pinned by
  `test_write_partition_merge_schema_partitioned_rejected_early`.

corwinjoy:
@ethan-tyler Any thoughts?

corwinjoy:
For context, this is part of a chain where we eventually hope to add support for columnar encryption and better support for extended Parquet read and write properties. See: https://github.com/delta-io/delta-rs/issues/3300#issuecomment-5186366569

PR #4569, opened 2026-06-26
docs: move feature table to docs

plaindocs:
README is less cluttered, feature table is more visible in the docs.


keen85:
I think, cloumn mapping is supported now for read and write

rtyler (approved):

PR #4570, opened 2026-06-26
docs: misc fixes and one added example

plaindocs:
# Description

- various small formatting fixes
- new example for table  restore


thomasfrederikhoeck:
As mentioned on #4569 Column Mapping is working for both read and Write I think - would be good to have correct i Readme (if it is correct)

plaindocs:
> As mentioned on #4569 Column Mapping is working for both read and Write I think - would be good to have correct i Readme (if it is correct)

I'll pick that up in a future PR.

rtyler (approved):

PR #4571, opened 2026-06-27
fix: gate python sftp dependency on windows

### LINKED ISSUE #4565: [Bug]: Windows builds broken due to openssh
### What happened?

Windows builds fails https://github.com/delta-io/delta-rs/actions/runs/28101273672/job/83203692991

### Expected behavior

.

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

_No response_

### Relevant logs

```shell

```

ethan-tyler:
closes #4565

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -16,6 +16,35 @@ env:
   RUSTC_WRAPPER: "sccache"
 
 jobs:
+  check-windows-python-dependencies:
+    name: Python Windows dependency graph
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: .
+    steps:
+      # v6.0.2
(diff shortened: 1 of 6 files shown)


.github/workflows/python_build.yml
```diff
@@ -16,6 +16,35 @@ env:
   RUSTC_WRAPPER: "sccache"
 
 jobs:
+  check-windows-python-dependencies:
```
ion-elgreco, line 19: Honestly I think its better we make windows and mac builds required again before merging to main  so this doesn't come as a surprise when releasing

ethan-tyler: I restored the builds

ion-elgreco: We don't need this job then anymore, right?

rtyler (approved):

PR #4585, opened 2026-07-08
feat: true DNF filters and SQL predicates for the file listing APIs [wip]

### LINKED ISSUE #1923: Partition predicate typing inconsistencies, and no support for OR in partition predicates
# Environment

**Delta-rs version**: 0.13.0

**Binding**: python

**Environment**:
- **Cloud provider**: n/a
- **OS**: ubuntu 22.04
- **Other**: testing locally via jupyter notebook

***
# Bug

**What happened**:
There are inconsistencies between the python type-hints, docstring comments, and rust implementation for partition predicates. 

In `table.py`, we see `partitions: Optional[List[Tuple[str, str, Any]]]` (https://github.com/delta-io/delta-rs/blob/main/python/deltalake/table.py#L704). There is also a docstring comment here referencing a `DeltaTable.files_by_partitions`, which seems to be a deprecated method, so I'm going to assume that is an out of date comment.

In `_internal.pyi`, we see `Optional[FilterType]` (https://github.com/delta-io/delta-rs/blob/main/python/deltalake/_internal.pyi#L86), which is defined as a `DNF` or `Conjuction` (https://github.com/delta-io/delta-rs/blob/main/python/deltalake/_internal.pyi#L717-L720).

In `lib.rs`, we see `PartitionFilterValue` (https://github.com/delta-io/delta-rs/blob/main/python/src/lib.rs#L655), which is defined as a string, or a vector of strings (https://github.com/delta-io/delta-rs/blob/main/python/src/lib.rs#L60-L63).

To my understanding, these three, which should all be defining the same partition filter type, all have different typing. The one in `table.py` and `lib.rs` are relatively congruent with each other, if we treat `Any` as `Union[str,List[str]]`, which sort of makes sense.

The bigger difference is with `_internal.pyi`, which presents a whole different form of filter. This DNF style would allow for `OR` in the predicate filter, but this seems to be impossible in the python or rust typing. Would it be difficult to add support for `OR` in partition predicates via this DNF style or another?

At the minimum, in the above code's current standing, I am not sure what the correct typing for predicate filters are. If the python one is incorrect, that should be fixed to provide accurate type hints and allow for correct static type checking when using python bindings.

Perhaps this should be raised in a Feature Request, but I'm also wondering why there isn't support for `OR` in predicate filters?

FrankPortman:
# Description

The tuple partition filters on `file_uris()`, `partitions()`, `to_pyarrow_dataset()`, `to_pyarrow_table()` and `to_pandas()` only ever supported AND, while the docstrings claimed full DNF, and the `file_uris` docstring was wiped at runtime by a leftover `__doc__ = ""` besides. This closes the gap, building on the kernel file skipping from #3669.

One new parameter, `file_pruning_predicate`, on the listing APIs. It takes three forms:

```python
# SQL string, same dialect delete/update/merge already take
dt.file_uris(file_pruning_predicate="year = 2021 OR (year = 2020 AND month >= 10)")

# flat list of tuples: one conjunction (AND), same meaning as the old params
dt.file_uris(file_pruning_predicate=[("year", "=", "2020"), ("month", ">=", "10")])

# list of lists: true DNF, an OR across the inner AND groups
dt.file_uris(file_pruning_predicate=[
    [("year", "=", "2021")],
    [("year", "=", "2020"), ("month", ">=", "10")],
])
```

The old `partition_filters`/`partitions` params keep working and are documented as deprecated (no runtime warning yet).

Both forms turn into a kernel `Predicate` and prune during log replay. Any column is allowed, not just partition columns:

- Partition columns prune exactly.
- Other columns prune on file-level min/max stats: files that provably contain no matching row are dropped, anything that might match is kept, files without stats for a referenced column are always kept. A complete superset, never a miss. How much actually prunes depends on layout (partitioning, z-ordering); documented on `file_uris` and in the user guide.

Following roeap's suggestion in #1923, tuples translate directly to kernel predicates (`literal/conjunction/dnf_to_kernel_predicate` in `partitions.rs`); the old `PartitionFilter` translation is a shim over that path. The SQL path (`parse_sql_predicate_to_kernel`) reuses the delete/update parse/coerce/simplify pipeline. The last two uses of the deprecated `EagerSnapshot::file_views_by_partitions` are removed.

Bugs fixed while in here:

- `coerce_predicate_literals` didn't coerce IN lists: `year IN (2020)` handed the kernel an Int64 literal against an Int32 column.
- `DeltaTable.partitions([("year", "=", 2021)])` failed on the int while the same filter worked on `file_uris`.
- The docstring rot above, three dead `files_by_partitions` references, an op list missing four of the eight supported ops.

## The stack

Follow-ups are built, not filed. They live on my fork because a PR's base must be a branch in the target repo; each retargets to main after this merges.

- **delta-io/delta-rs#4585** — this PR
  - **FrankPortman/delta-rs#1** — drop `file_pruning_predicate` from `to_pandas`/`to_pyarrow_table`, where `filters` already prunes files via fragment stats and then filters rows exactly. Lands with this PR before a release, or closes.
  - **FrankPortman/delta-rs#3** — delete `PartitionFilter`, add a typed `FilterOp`. Standalone; operationParameters serialization byte-identical.
  - **FrankPortman/delta-rs#2** — QueryBuilder path for `to_pandas`/`to_pyarrow_table` (ion's suggestion). Reads columnMapping and DV tables today; adds a second read path until the chain below completes.
    - **FrankPortman/delta-rs#4** — deprecate the pyarrow fallbacks; `to_pyarrow_dataset` becomes the explicit pyarrow interop surface.
      - **FrankPortman/delta-rs#5** — dataframe methods run only through QueryBuilder. Removes released API (`filesystem=`, `Expression` filters), so it ships a release after FrankPortman/delta-rs#4's warnings unless maintainers prefer one breaking release.

Preferred sequencing: this PR + #1 + #3 before the next release, #2 + #4 with it or next, #5 one release later. Every piece is separable; #1 keeps the pruning param off the dataframe methods even if the chain stalls.

## Decision table

Open rows are defaults with defensible alternatives; object per row and I'll change it here.

| # | Boundary | Status |
|---|---|---|
| 1 | Param count | **Resolved (review):** single `file_pruning_predicate`, SQL string or tuples |
| 2 | Both params passed | **Moot:** one param. Old deprecated params error if combined with the new one |
| 3 | Column scope | Any column, no flag. roeap called data-column skipping the point of kernel predicates (#1923); rename carries the disambiguation. Open to restricting if blocking |
| 4 | `get_active_partitions` validation | Non-partition-column error removed (error, now results) |
| 5 | Superset semantics | Data columns prune to a documented stats-based superset |
| 6 | Flat-list back-compat | Kept, flat means one conjunction |
| 7 | `''` for null | Kept, documented; SQL `IS NULL` is the unambiguous spelling |
| 8 | Unconvertible SQL (`LIKE`, functions) | Hard error listing what is supported |
| 9 | Dataframe-method pruning | **Resolved by the stack:** FrankPortman/delta-rs#1 now, #2/#4/#5 as the direction |
| 10 | Empty inner conjunction `[[]]` | Error |
| 11 | optimize / create_write_transaction | Out of scope; migrated off deprecated APIs only |
| 12 | `PartitionFilter` | **Built:** deleted in FrankPortman/delta-rs#3 |
| 13 | Session for SQL parsing | Throwaway `SessionContext::new()` per call, same as `load_cdf` |
| 14 | Op typing | **Built:** typed `FilterOp` in FrankPortman/delta-rs#3 |
| 15 | Core API shape | New `get_active_add_actions_by_predicate(Option<PredicateRef>)` next to the `_by_partitions` variant |

## Tradeoffs kept

- IN is lowered in two places (tuple path in `partitions.rs`, SQL path in `to_kernel.rs`); a parity test asserts identical predicates, so drift fails CI.
- `dataset_partitions` keeps its two-pass shape; single-pass touches the pyarrow fragment plumbing and belongs in its own change.
- Null handling differs between forms for legacy reasons: tuple `("col", "=", "")` means IS NULL, SQL `col = ''` means empty string. Documented.
- Ordering ops on string columns compare lexicographically (`("month", ">=", 12)` matches `"4"`). Documented on `file_uris`.

# Related Issue(s)

- closes #1923
- follow up to #1316 and #3669

# Documentation

- `docs/usage/querying-delta-tables.md`: selecting files with a pruning predicate, all three forms, exact vs superset rules, layout note
- `docs/usage/working-with-partitions.md`: OR across partitions in both forms
- `docs/how-delta-lake-works/delta-lake-file-skipping.md`: connects the concept page to the API
- API reference regenerates from the rewritten docstrings

### DIFF
--- a/crates/core/src/kernel/schema/partitions.rs
+++ b/crates/core/src/kernel/schema/partitions.rs
@@ -183,8 +183,149 @@ impl TryFrom<&str> for DeltaTablePartition {
     }
 }
 
-#[allow(unused)] // TODO: remove once we use this in kernel log replay
-pub(crate) fn to_kernel_predicate(
+/// The value of a `(column, op, value)` filter literal: a single partition-value
+/// encoded string, or a set of them for `in` / `not in`.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum FilterValue<'a> {
+    /// A single encoded value, compared with one of `=`, `!=`, `<`, `<=`, `>`, `>=`.
+    Scalar(&'a str),
+    /// A set of encoded values, compared with `in` or `not in`.
+    Set(Vec<&'a str>),
+}
+
+/// A `(column, op, value)` comparison, mirroring the tuple filters accepted by
+/// the Python bindings.
+pub type FilterLiteral<'a> = (&'a str, &'a str, FilterValue<'a>);
+
+/// Translate a single filter literal into a kernel [`Predicate`].
+///
+/// The raw value is parsed against the schema type of `column`. A null scalar
+/// under `=` / `!=` becomes an IS [NOT] NULL check: in SQL NULL compares equal
+/// to nothing, itself included, but these filters have always allowed equality
+/// against the null partition value.
+pub fn literal_to_kernel_predicate(
+    literal: &FilterLiteral<'_>,
+    table_schema: &StructType,
+) -> DeltaResult<Predicate> {
+    let (column, op, value) = literal;
+    let op_matches_value = match value {
+        FilterValue::Scalar(_) => matches!(*op, "=" | "!=" | "<" | "<=" | ">" | ">="),
+        FilterValue::Set(_) => matches!(*op, "in" | "not in"),
+    };
+    if column.is_empty() || !op_matches_value {
+        return Err(invalid_filter_error(literal));
+    }
+    let Some(field) = table_schema.field(column) else {
+        return Err(DeltaTableError::SchemaMismatch {
+            msg: format!("Field '{column}' is not a root table field."),
+        });
+    };
+    let Some(dt) = field.data_type().as_primitive_opt() else {
+        return Err(DeltaTableError::SchemaMismatch {
+            msg: format!("Field '{}' is not a primitive type", field.name()),
+        });
+    };
+
+    let col = Expression::column([field.name()]);
+    Ok(match (*op, value) {
+        ("=", FilterValue::Scalar(raw)) => {
+            let scalar = dt.parse_scalar(raw)?;
+            if scalar.is_null() {
+                col.is_null()
+            } else {
+                col.eq(scalar)
+            }
+        }
+        ("!=", FilterValue::Scalar(raw)) => {
+            let scalar = dt.parse_scalar(raw)?;
+            if scalar.is_null() {
+                col.is_not_null()
+            } else {
+                col.ne(scalar)
+            }
+        }
+        ("<", FilterValue::Scalar(raw)) => col.lt(dt.parse_scalar(raw)?),
+        ("<=", FilterValue::Scalar(raw)) => col.le(dt.parse_scalar(raw)?),
+        (">", FilterValue::Scalar(raw)) => col.gt(dt.parse_scalar(raw)?),
+        (">=", FilterValue::Scalar(raw)) => col.ge(dt.parse_scalar(raw)?),
+        (op @ ("in" | "not in"), FilterValue::Set(raws)) => {
+            let values = raws
+                .iter()
+                .map(|v| dt.parse_scalar(v))
+                .collect::<Result<Vec<_>, _>>()?;
+            let (term, junction): (Box<dyn Fn(Scalar) -> Predicate>, _) = if op == "in" {
+                (Box::new(|v| col.clone().eq(v)), JunctionPredicateOp::Or)
+            } else {
+                (Box::new(|v| col.clone().ne(v)), JunctionPredicateOp::And)
+            };
+            let predicates = values.into_iter().map(term).collect::<Vec<_>>();
+            Predicate::junction(junction, predicates)
+        }
+        _ => unreachable!("op/value shapes checked above"),
+    })
+}
+
+fn invalid_filter_error(literal: &FilterLiteral<'_>) -> DeltaTableError {
+    let (column, op, value) = literal;
+    let value = match value {
+        FilterValue::Scalar(v) => format!("{v:?}"),
+        FilterValue::Set(vs) => format!("{vs:?}"),
+    };
+    DeltaTableError::InvalidPartitionFilter {
+        partition_filter: format!("({column:?}, {op:?}, {value})"),
+    }
+}
+
+/// Translate a conjunction (AND) of filter literals into a kernel [`Predicate`].
+///
+/// Errors on an empty conjunction: an empty AND is vacuously true and would
+/// silently match every file.
+pub fn conjunction_to_kernel_predicate(
+    literals: &[FilterLiteral<'_>],
+    table_schema: &StructType,
+) -> DeltaResult<Predicate> {
+    if literals.is_empty() {
+        return Err(DeltaTableError::Generic(
+            "empty conjunction in filter; pass no filter to match all files".to_string(),
+        ));
+    }
+    let mut predicates = literals
+        .iter()
+        .map(|literal| literal_to_kernel_predicate(literal, table_schema))
+        .collect::<DeltaResult<Vec<_>>>()?;
+    Ok(match predicates.len() {
+        1 => predicates.pop().unwrap(),
+        _ => Predicate::junction(JunctionPredicateOp::And, predicates),
+    })
+}
+
+/// Translate filters in disjunctive normal form -- an OR across conjunctions
+/// (AND groups) of `(column, op, value)` literals -- into a kernel [`Predicate`].
+pub fn dnf_to_kernel_predicate(
+    dnf: &[Vec<FilterLiteral<'_>>],
+    table_schema: &StructType,
+) -> DeltaResult<Predicate> {
+    if dnf.is_empty() {
+        return Err(DeltaTableError::Generic(
+            "empty filter; pass no filter to match all files".to_string(),
+        ));
+    }
+    let mut groups = dnf
+        .iter()
+        .map(|conjunction| conjunction_to_kernel_predicate(conjunction, table_schema))
+        .collect::<DeltaResult<Vec<_>>>()?;
+    Ok(match groups.len() {
+        1 => groups.pop().unwrap(),
+        _ => Predicate::junction(JunctionPredicateOp::Or, groups),
+    })
+}
+
+/// Translate a conjunction of [`PartitionFilter`]s into a kernel [`Predicate`].
+///
+/// Unlike [`dnf_to_kernel_predicate`], an empty slice yields an empty AND
+/// junction, which is vacuously true: callers treat "no filters" as "match
+/// every file".
+pub fn to_kernel_predicate(
     filters: &[PartitionFilter],
     table_schema: &StructType,
 ) -> DeltaResult<Predicate> {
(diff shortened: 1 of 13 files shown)


FrankPortman:
@ion-elgreco friendly bump here - i am happy to start pulling out any simple or uncontroversial commits into other PRs so we can have a smaller blast radius to discuss

there are unfortunately a lot of choices to be made on edge-case behavior, parameters, etc.

ion-elgreco:
> @ion-elgreco friendly bump here - i am happy to start pulling out any simple or uncontroversial commits into other PRs so we can have a smaller blast radius to discuss
> 
> there are unfortunately a lot of choices to be made on edge-case behavior, parameters, etc.

I am unfortunately extremely busy at the moment, perhaps someone else can have a look at it, I will remove my assignment for now

docs/usage/querying-delta-tables.md
```diff
@@ -34,6 +34,47 @@ pyarrow.Table
...
+are then evaluated in disjunctive normal form, an OR across the inner AND
+groups -- or pass a SQL predicate string instead:
+
+``` python
+>>> dt.to_pandas(
+...     partitions=[
+...         [("year", "=", "2020"), ("month", "=", "2")],
+...         [("year", "=", "2021"), ("month", "=", "12")],
+...     ],
+... )
```
ion-elgreco: Can we not deprecate partitions now? Since predicate does partition + stats pruning.

I am also thinking why should we even go through pyarrow for pandas, we can also do QueryBuilder -> Arro3 -> Pandas

Predicate should probably be named differently then otherwise without reading docs people might assume it does full filter

python/src/lib.rs
```diff
@@ -293,6 +301,42 @@ impl RawDeltaTable {
...
+        &self,
+        partition_filters: Option<Vec<PyFilterConjunction>>,
+        predicate: Option<String>,
+    ) -> PyResult<Option<PredicateRef>> {
+        let partition_filters = partition_filters.filter(|filters| !filters.is_empty());
+        let resolved = match (partition_filters, predicate) {
+            (Some(_), Some(_)) => {
+                return Err(PyValueError::new_err(
+                    "partition_filters and predicate are mutually exclusive; pass only one",
+                ));
```
ion-elgreco: Why would we validate in python if we already validating here? Just doing it here should be enough

crates/core/src/kernel/schema/partitions.rs
```diff
@@ -210,51 +226,139 @@ fn filter_to_kernel_predicate(
         });
     };
 
-    let column = Expression::column([field.name()]);
-    Ok(match &filter.value {
-        // NOTE: In SQL NULL is not equal to anything, including itself. However when specifying partition filters
```
ion-elgreco, line 215: I would not remove this note tbh

crates/core/src/kernel/schema/partitions.rs
```diff
@@ -183,25 +183,41 @@ impl TryFrom<&str> for DeltaTablePartition {
...
+    Scalar(&'a str),
+    /// A set of encoded values, compared with `in` or `not in`.
+    Set(Vec<&'a str>),
 }
 
-fn filter_to_kernel_predicate(
-    filter: &PartitionFilter,
+/// A `(column, op, value)` comparison, mirroring the tuple filters accepted by
+/// the Python bindings.
+pub type FilterLiteral<'a> = (&'a str, &'a str, FilterValue<'a>);
```
ion-elgreco, line 198: Why did we have to do this and not keep partitionFilter? Structurally the code is the same.

ion-elgreco (changes requested):
Using your decision table as a starting point:

1. if we keep param count to 1, then we dont have to validate if both params are passed. So I suggest that
2. ..
3. Agreed
4. Agreed
5. I think the alternative sounds better here, but in general this can be a footgun. Terminology is key here to not give the false idea of a predicate doing a proper full predicate, its only a file_pruning_predicate
6. Agreed
7.  Keep it yeah
8. Agreed, should error
9. This can be resolved by going through QueryBuilder, both for pyarrow table and to pandas
10. fine by me
12. Agreed, do you have a followup ready?
13. I guess thats fine
14. Agreed, works well in a followup
15. makes sense

Left some other notes for you to look at as well!

FrankPortman:
Pushed the changes from your review and rebased on main:
   - One parameter now, `file_pruning_predicate`, taking either a SQL string or tuple filters (flat list = AND, list of lists = OR of ANDs). The old tuple params still work and are documented as deprecated; no runtime warning yet to keep the next release quiet but I can add one. Or we can even consider hard-switching since this new param has been renamed.
   - Restored the null equality note.
  - On data columns in the pruning predicate: kept as is for now. The rename should fix disambiguation, and in #1923 @roeap  called data-column skipping the main benefit of adopting kernel predicates here. Still open to restricting it if it blocks you or another reviewer.

I have not yet pushed a large change or series of changes to this repo (only 1-2 self contained shorter PRs before), so I am **experimenting** with the following flow. Let me know if this makes reviewing harder and if it's easier to just try squashing these as new commits onto this same PR, etc. But I figure they al tell separate stories and can be reviewed and land separately (although hopefully, without much delay between).

Rather than file follow-up issues I built the follow-ups, so the full picture is reviewable. Everything is stacked on this branch in my fork (a PR's base has to live in the target repo, so they can't point at delta-io until this PR merges):

```
delta-io#4585   DNF + file_pruning_predicate on the listing APIs   <- this PR
    |
    |-- FrankPortman/delta-rs#1   drop the pruning param from to_pandas/to_pyarrow_table
    |                    optional, but time-bound. We should merger with this
    |                    PR and release together if accepted., or gets closed.
    |
    |-- FrankPortman/delta-rs#3   delete PartitionFilter, add a typed FilterOp
    |                    Standalone cleanup, no Python-visible change.
    |                    Lands any time after this PR.
    |
    '-- FrankPortman/delta-rs#2   QueryBuilder path for to_pandas/to_pyarrow_table
          |              your suggestion from this review. Reads columnMapping
          |              and DV tables today. Trade off is a second read path.
          |              Unless 4 and 5 are also accepted.
          |
          '-- FrankPortman/delta-rs#4   deprecate the pyarrow fallbacks on the two
                |              dataframe methods. to_pyarrow_dataset becomes
                |              the explicit pyarrow interop surface
                |
                '-- FrankPortman/delta-rs#5   dataframe methods run ONLY through
                                     QueryBuilder. If #1 lands first, this gets simpler.
```

Clickable links in the order displayed in the ASCII drawing:
- **delta-io/delta-rs#4585**
- **FrankPortman/delta-rs#1**
- **FrankPortman/delta-rs#3**
- **FrankPortman/delta-rs#2**
- **FrankPortman/delta-rs#4**
- **FrankPortman/delta-rs#5**

Note I only reviewed these high level for now but directionally they look correct for the intended goal. I'd make them "ready to land" up to and including PR description as we peel things off into PRs against this repo.

There are some release order and bundling concerns here but let me know what would make this easier to review and land.

crates/core/src/kernel/schema/partitions.rs
FrankPortman: It can't represent DNF since it is a flat (column, op, value) with no nesting, so OR was structurally impossible without changing the public type.

Also I was going off of the guidance re @roeap in #1923: "the idea would be to skip PartitionFilter and immediately translate DNF to Predicate."

FrankPortman/delta-rs#3 deletes the type entirely which I think is a nice cleanup win.

crates/core/src/kernel/schema/partitions.rs
FrankPortman: restored

python/src/lib.rs
FrankPortman: moot now with one param, resolving

docs/usage/querying-delta-tables.md
FrankPortman: Did 3 things:
* single param named `file_pruning_predicate`
* old tuple params deprecated in docs (runtime warning can follow - I am not sure what this repo's policy is on that stuff)
* the QueryBuilder read path is built as FrankPortman/delta-rs#2, with some optional followups in FrankPortman/delta-rs#4 and FrankPortman/delta-rs#5

ion-elgreco: We should keep both first with the old one marked as deprecated and then after couple minor releases we con drop the param

FrankPortman: by mark do you mean with a specific deprecation runtime warning or something like that? or is doc enough

crates/core/src/kernel/schema/partitions.rs
ion-elgreco: Got it!

docs/usage/querying-delta-tables.md
ion-elgreco: Yes a runtime warning when that value is set and then we kust pass it to the file_pruning_predicate of the pyo3 signature

ion-elgreco:
Stacked PRs are fine to me, generally when making additional changes in a PR it's easier if those are separate commits so I can review those changes individually from the last review

docs/usage/querying-delta-tables.md
FrankPortman: Done and pushed as a new commit here! Didn't change anything in the downstream PRs yet, but one at a time :D.

FrankPortman:
@ion-elgreco updated the PR with the formal deprecation as a new commit.

ion-elgreco (approved):

ion-elgreco:
Thanks @FrankPortman, could you resolve the spell check if it causes it to not pass the merge

FrankPortman:
@ion-elgreco nice! thank you! stay tuned for followups :D.

PR #4592, opened 2026-07-10
feat(catalog-unity): add custom token credential support to UnityCatalogBuilder

### LINKED ISSUE #4587: [Feature]: Possibility to supply own CredentialProvider and TokenCredential when building Unity Catalog.

We can not use a custom implementation of `TokenCredential` to authenticate Unity Catalog requests, despite the `TokenCredential` trait being public.

We would like to be able to define our own implementation of `TokenCredential` and then be able to set the `CredentialProvider` when building the Unity Catalog with this custom `TokenCredential`.

The alternative would be to use our own token provider to fetch the token and then use `.bearer_token()` when building the Unity Catalog with that token, but then it will expire instead of being refreshed because we are not using the `CredentialProvider` enum that contains the cache and expiry logic.

### Priority

### Additional context

_No response_

### Contribution

11alieberman:
# Description
Allow callers to provide a custom TokenCredential implementation when building UnityCatalog, without changing the public API surface.

- add optional `token_credential` field to `UnityCatalogBuilder`
- resolve custom credential through `get_credential_provider` with expected precedence
- add test to assert `CredentialProvider::TokenCredential` is selected

# Related Issue(s)
- closes https://github.com/delta-io/delta-rs/issues/4587

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>
Signed-off-by: Alec Lieberman <alec.lieberman@gmail.com>

### DIFF
--- a/crates/catalog-unity/src/lib.rs
+++ b/crates/catalog-unity/src/lib.rs
@@ -1353,6 +1353,14 @@ mod tests {
             "Credential provider should be resolved from custom credential"
         );
 
+        assert!(
+            matches!(
+                credential_provider.as_ref(),
+                Some(crate::credential::CredentialProvider::TokenCredential(_, _))
+            ),
+            "Expected TokenCredential variant, got BearerToken"
+        );
+
         if let Some(crate::credential::CredentialProvider::TokenCredential(_, token_credential)) =
             credential_provider
         {
@@ -1364,8 +1372,6 @@ mod tests {
                 token.token, "test_credential_token",
                 "Fetched token should match the test credential token"
             );
-        } else {
-            panic!("Expected TokenCredential variant, got BearerToken");
         }
     }
 }
--- a/crates/catalog-unity/src/lib.rs
+++ b/crates/catalog-unity/src/lib.rs
@@ -23,7 +23,8 @@ use tracing::instrument;
 use typed_builder::TypedBuilder;
 
 use crate::credential::{
-    AzureCliCredential, ClientSecretOAuthProvider, CredentialProvider, WorkspaceOAuthProvider,
+    AzureCliCredential, ClientSecretOAuthProvider, CredentialProvider, TokenCredential,
+    WorkspaceOAuthProvider,
 };
 use crate::models::{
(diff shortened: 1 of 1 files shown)


crates/catalog-unity/src/lib.rs
```diff
@@ -1302,4 +1317,55 @@ mod tests {
...
+            let token = token_credential
+                .fetch_token(&builder_mut.client_options.client().unwrap())
+                .await
+                .unwrap();
+            assert_eq!(
+                token.token, "test_credential_token",
+                "Fetched token should match the test credential token"
+            );
+        } else {
+            panic!("Expected TokenCredential variant, got BearerToken");
```
hntd187: No panics in test code

hntd187 (changes requested):
Remove the panic from the tests, also out of curiosity, can you say more about what your use case is for this?

11alieberman:
> Remove the panic from the tests, also out of curiosity, can you say more about what your use case is for this?

We have a custom way of obtaining an azure TokenCredential from the Azure rust SDK. I want to create an implementation of the TokenCredential for the Unity Catalog builder that uses the Azure TokenCredential to retrieve a token that the Credential provider can use.

Something like

```
#[derive(Debug)]
struct AzureTokenCredential {
    token_credential: Arc<dyn TokenCredential>,
}
```

and then

```
impl deltalake_catalog_unity::credential::TokenCredential for AzureTokenCredential {

}
```

Where we use `self.token_credential.get_token` to return the `TemporaryToken` and then we get the caching and refresh mechanisms from the `CredentialProvider`

crates/catalog-unity/src/lib.rs
11alieberman: It has been removed.

hntd187:
@11alieberman you'll have to resolve the DCO before we can merge, if you click through there should be directions on how to sign your commits

11alieberman:
> @11alieberman you'll have to resolve the DCO before we can merge, if you click through there should be directions on how to sign your commits

I have followed the instructions to resolve that. I think it should be good now. Sorry about that.

hntd187 (approved):

PR #4596, opened 2026-07-13
fix: filter files also when_not_matched_by_source

### LINKED ISSUE #4535: [Bug]: MERGE with WHEN NOT MATCHED BY SOURCE ignores target partition predicate and scans all partitions
### What happened?

Adding a `when_not_matched_by_source_update` clause disables target partition pruning, even when both the merge predicate and the clause predicate restrict the operation to the same partition.

Without `when_not_matched_by_source_update`, delta-rs scans only the targeted partition. With the clause added, it scans all target partitions:

- `small_without_not_matched_by_source`: 1 file scanned, 2 skipped
- `small_with_not_matched_by_source`: 3 files scanned, 0 skipped

Increasing only the size of the non-target partitions also substantially increases execution time, even though the merge predicate and the `when_not_matched_by_source_update` predicate both restrict the relevant target rows to `2026-01-02`. In the reproduction below, execution time increases from 6 ms to 686 ms, with `scan_time_ms` increasing from 1 ms to 680 ms.

This matters for workflows where `WHEN NOT MATCHED BY SOURCE` is used to deactivate or delete records within a specific partition. On tables with many or large partitions, scanning all partitions can make these merges unnecessarily slow.

This behavior has already been flagged in discussion #3477. The reply states that `WHEN NOT MATCHED BY SOURCE` triggers a full scan, and points to this implementation comment:

    // It's only worth trying to create an early filter where there are no
    // `when_not_matched_source` operators, since that implies a full scan

This explains the current implementation, but not whether a full scan is required by the merge semantics. I may be missing an implementation constraint, but it seems that the clause’s own target-only partition predicate could still be used to build the target subset filter.

### Expected behavior

I expected delta-rs to scan only the partition selected by:

    t.date = '2026-01-02'

The merge predicate restricts matching to that partition, and the `when_not_matched_by_source_update` predicate independently restricts the unmatched-by-source action to the same partition.

Therefore, partitions outside this predicate should not need to be scanned. I would expect:

- the same partition pruning with and without `when_not_matched_by_source_update`;
- `small_with_not_matched_by_source ` and `large_with_not_matched_by_source` to have similar execution times, because the source data and targeted partition are identical and only the size of non-target partitions changes.

The resulting data is correct, so this appears to be a partition-pruning or performance issue rather than a correctness issue.

### Operating System

macOS

### Binding

Python

### Bindings Version

1.6.0

### Steps to reproduce

```python
import shutil

import polars as pl

path = "delta-rs-merge-pruning-repro"
partition = "2026-01-02"

def create_table(extra_rows_per_other_partition: int = 0):
    shutil.rmtree(path, ignore_errors=True)

    target = pl.DataFrame({
        "date": ["2026-01-01"] * 2 + ["2026-01-02"] * 3 + ["2026-01-03"] * 2,
        "id": [1, 2, 1, 2, 3, 1, 2],
        "val": [10, 20, 100, 200, 300, 1000, 2000],
        "keep": [True] * 7,
    })

    if extra_rows_per_other_partition:
        irrelevant = pl.DataFrame({
            "date": (
                ["2026-01-01"] * extra_rows_per_other_partition
                + ["2026-01-03"] * extra_rows_per_other_partition
            ),
            "id": (
                list(range(10_000, 10_000 + extra_rows_per_other_partition))
                + list(range(20_000, 20_000 + extra_rows_per_other_partition))
            ),
            "val": [1] * (2 * extra_rows_per_other_partition),
            "keep": [True] * (2 * extra_rows_per_other_partition),
        })
        target = pl.concat([target, irrelevant])

    write_deltalake(
        path,
        target,
        mode="overwrite",
        partition_by=["date"],
    )


source = pl.DataFrame({
    "date": [partition, partition],
    "id": [1, 2],
    "val": [101, 202],
})


def run_merge(with_not_matched_by_source: bool):
    merger = (
        DeltaTable(path)
        .merge(
            source=source.to_arrow(),
            source_alias="s",
            target_alias="t",
            predicate=(
                "t.date = s.date "
                "AND t.id = s.id "
                f"AND t.date = '{partition}'"
            ),
        )
        .when_matched_update({"val": "s.val"})
    )

    if with_not_matched_by_source:
        merger = merger.when_not_matched_by_source_update(
            predicate=f"t.date = '{partition}'",
            updates={"keep": "false"},
        )

    return merger.execute()


runs = []

create_table()
runs.append({
    "run": "small_without_not_matched_by_source",
    **run_merge(with_not_matched_by_source=False),
})

create_table()
runs.append({
    "run": "small_with_not_matched_by_source",
    **run_merge(with_not_matched_by_source=True),
})

create_table(extra_rows_per_other_partition=10_000_000)
runs.append({
    "run": "large_with_not_matched_by_source",
    **run_merge(with_not_matched_by_source=True),
})

result = pl.DataFrame(runs).select(
    "run",
    "num_target_files_added",
    "num_target_files_scanned",
    "num_target_files_skipped_during_scan",
    "execution_time_ms",
    "scan_time_ms",
)

for row in result.iter_rows(named=True):
    print(f"{row['run']}:")
    print(f"  files_added: {row['num_target_files_added']}")
    print(f"  files_scanned: {row['num_target_files_scanned']}")
    print(f"  files_skipped: {row['num_target_files_skipped_during_scan']}")
    print(f"  execution_time_ms: {row['execution_time_ms']}")
    print(f"  scan_time_ms: {row['scan_time_ms']}")
    print()

shutil.rmtree(path, ignore_errors=True)
```

### Relevant logs

```shell
small_without_not_matched_by_source:
  files_added: 1
  files_scanned: 1
  files_skipped: 2
  execution_time_ms: 6
  scan_time_ms: 1

small_with_not_matched_by_source:
  files_added: 1
  files_scanned: 3
  files_skipped: 0
  execution_time_ms: 6
  scan_time_ms: 1

large_with_not_matched_by_source:
  files_added: 1
  files_scanned: 3
  files_skipped: 0
  execution_time_ms: 686
  scan_time_ms: 680
```

MedHeikelBouzayene:
# Description
Fixes partition pruning for MERGE operations that include a when_not_matched_by_source_update clause.

# Related Issue(s)
- closes #4535 

# Documentation


hntd187:
Sorry, but how does just removing this check fix this?

MedHeikelBouzayene:
@hntd187 That check skip early filtering, and force full merge. By removing, we enable early filtering even when  we use when_not_matched_by_source_update. Check with the example mentioned in the issue

rtyler:
At one point I had set Mistral loose on this bug too, and it identified the same segment of code. I approved the workflow so the test failures could be made evident. Tests simply don't pass and functionality breaks if that check is removed.

ion-elgreco:
We need a full scan unfortunately if its when_not_matched_by_source, I think we can simply close this pr

MedHeikelBouzayene:
@ion-elgreco can you elaborate more why full scan is needed when when_not_matched_by_source is applied?
as shown in the example in the issue, full scan is not required

ion-elgreco:
> @ion-elgreco can you elaborate more why full scan is needed when when_not_matched_by_source is applied? as shown in the example in the issue, full scan is not required

A when not matched by source means we need to act (delete/update) on the rows that don't match with source, if we prune the files early it means they will not show up in the dataframe and we will not be able to act on these rows that are not matched by source.

MedHeikelBouzayene:
> > @ion-elgreco can you elaborate more why full scan is needed when when_not_matched_by_source is applied? as shown in the example in the issue, full scan is not required
> 
> A when not matched by source means we need to act (delete/update) on the rows that don't match with source, if we prune the files early it means they will not show up in the dataframe and we will not be able to act on these rows that are not matched by source.

that makes sense. I think the issue should be closed too?

PR #4598, opened 2026-07-14
feat: added TableFeature::V2Checkpoint to the reader and writer features

### LINKED ISSUE #4597: [Feature]: Expose TableFeature::V2Checkpoint in kernel/transactions/protocol.rs

This is more of a minor update than a full blown feature. This PR is to expose the V2Checkpoint kernel feature in delta-rs. This is specific to reader functionality.

This should be self contained (going to run the tests next). The underlying feature is already tested inside the kernel: https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/checkpoint/checkpoint_shape.rs

```txt
diff --git a/crates/core/src/kernel/transaction/protocol.rs b/crates/core/src/kernel/transaction/protocol.rs
index fa5cc958..a052350a 100644
--- a/crates/core/src/kernel/transaction/protocol.rs
+++ b/crates/core/src/kernel/transaction/protocol.rs
@@ -302,6 +302,7 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     reader_features.insert(TableFeature::DeletionVectors);
     reader_features.insert(TableFeature::VariantType);
     reader_features.insert(TableFeature::VariantTypePreview);
+    reader_features.insert(TableFeature::V2Checkpoint);
     #[cfg(feature = "nanosecond-timestamps")]
     reader_features.insert(TableFeature::TimestampNanos);
@@ -316,6 +317,7 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     writer_features.insert(TableFeature::TimestampNanos);
     writer_features.insert(TableFeature::VariantType);
     writer_features.insert(TableFeature::VariantTypePreview);
+    writer_features.insert(TableFeature::V2Checkpoint);
     {
         writer_features.insert(TableFeature::ChangeDataFeed);
```

I can't. Just looking at the kernel features and exposing the reader/writer.

### Priority

Medium - Would be helpful

### Additional context

I have been running compatibility tests between delta-spark and delta-rs and ran into this issue with V2Checkpoint.

### Contribution

newfront:
# Description
The changes made here add TableFeature::V2Checkpoint to the reader_ and writer_ features in core/src/kernel/transactions/protocol.rs. 

Overview: https://github.com/delta-io/delta-rs/issues/4597

# Related Issue(s)
Closes #4597

# Documentation
Happy to add documentation where it would be applicable for this.

# Tests
I modified the test cases that were using this table feature as 'unsupported' and passed the baton over to TypeWidening. 

- given this is my first PR, if there are better / alternative tests I could add please let me know and I'll be happy to update this.

### DIFF
--- a/crates/core/src/kernel/transaction/protocol.rs
+++ b/crates/core/src/kernel/transaction/protocol.rs
@@ -302,6 +302,7 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     reader_features.insert(TableFeature::DeletionVectors);
     reader_features.insert(TableFeature::VariantType);
     reader_features.insert(TableFeature::VariantTypePreview);
+    reader_features.insert(TableFeature::V2Checkpoint);
     #[cfg(feature = "nanosecond-timestamps")]
     reader_features.insert(TableFeature::TimestampNanos);
     #[cfg(feature = "datafusion")]
--- a/crates/core/src/kernel/transaction/protocol.rs
+++ b/crates/core/src/kernel/transaction/protocol.rs
@@ -317,6 +317,7 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     writer_features.insert(TableFeature::TimestampNanos);
     writer_features.insert(TableFeature::VariantType);
     writer_features.insert(TableFeature::VariantTypePreview);
+    writer_features.insert(TableFeature::V2Checkpoint);
     #[cfg(feature = "datafusion")]
     {
         writer_features.insert(TableFeature::ChangeDataFeed);
--- a/crates/core/src/delta_datafusion/table_provider/next/mod.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/mod.rs
@@ -1099,8 +1099,8 @@ mod tests {
             .with_columns(schema.fields().cloned())
             .with_actions(vec![Action::Protocol(
                 ProtocolInner::new(3, 7)
-                    .append_reader_features([TableFeature::V2Checkpoint])
-                    .append_writer_features([TableFeature::V2Checkpoint])
+                    .append_reader_features([TableFeature::TypeWidening])
+                    .append_writer_features([TableFeature::TypeWidening])
                     .as_kernel(),
             )])
             .await


newfront:
Is there a simple fix for the draft_pr action? I manually converted this PR to draft, but it isn't helping.

hntd187:
Hmm, why do you change some features to type widening instead of v2 checkpoints? Also we can't just add table features unless we're sure they supported. I'm pretty sure V2Checkpoints are via kernel, but it wouldn't surprise me if we hadn't enabled or integrated them yet meaning adding this would actually cause more problems.

newfront:
> Hmm, why do you change some features to type widening instead of v2 checkpoints? Also we can't just add table features unless we're sure they supported. I'm pretty sure V2Checkpoints are via kernel, but it wouldn't surprise me if we hadn't enabled or integrated them yet meaning adding this would actually cause more problems. 

So I have a patched build that works with V2Checkpoints and via the corresponding delta-spark based table and this was all that was needed to enable the feature. It is part of the kernel without needing to opt in to a specific feature.

As for the test, the original test was testing for the absence of V2Checkpoint, so the thought was to continue to use said test for another feature that currently hasn't been brought up to delta-rs.

hntd187:
I'd prefer separating the two if we are going to change the test to do something else. I'm also confused, are you saying that this change is to make your patched version work?

newfront:
> I'd prefer separating the two.
check. That is fine with me. 

>  I'm also confused, are you saying that this change is to make your patched version work?

I ran into an issue with the Unity Catalog Console (new project with embedded delta-rs datafusion for live query). 

```
Error: build table provider for coffee-orders: Error during planning: Unsupported table features required: [V2Checkpoint]
```
After the patch (enabling the feature), I can now read the following table: 

uc-console

crates/core/src/kernel/transaction/protocol.rs
```diff
@@ -316,6 +317,7 @@ pub static INSTANCE: LazyLock<ProtocolChecker> = LazyLock::new(|| {
     writer_features.insert(TableFeature::TimestampNanos);
     writer_features.insert(TableFeature::VariantType);
     writer_features.insert(TableFeature::VariantTypePreview);
+    writer_features.insert(TableFeature::V2Checkpoint);
```
hntd187: We don't support writing v2checkpoints yet, so let's remove this.

hntd187 (changes requested):

hntd187:
Not sure what happened here, but I think you broke your commit history.

crates/core/src/kernel/transaction/protocol.rs
ion-elgreco: But our checkpoints are written with kernel, the kernel docs says it supports v2 writing

hntd187: I stand corrected, apologies. We should be fine here then until the commits are cleaned up

newfront:
> Not sure what happened here, but I think you broke your commit history.

This was a mistake. The branch state got corrupted by an agent.

crates/core/src/kernel/transaction/protocol.rs
newfront: I'll add the writer_features back. Having this be full support was what I was hoping for. The work I'm doing right now is full cycle testing between delta-spark and delta-rs, handoff to the different engines.

hntd187 (approved):

hntd187 (approved):

PR #4599, opened 2026-07-15
feat: support RemoveAction together with AddAction from Python

### LINKED ISSUE #4480: [Feature]: Support committing AddAction and RemoveAction together in Python
## Feature Request

Expose a Python transaction API that can commit `AddAction` and `RemoveAction` together in a single Delta transaction.

## Use Case

I am building a Python service that ingests CSV logs into Delta Lake tables.

For append-only writes, `DeltaTable.create_write_transaction()` works well because I can write Parquet files myself and commit the generated `AddAction`s.

However, for custom compaction / file rewrite workflows, I need to atomically commit both:

- `AddAction`s for newly compacted Parquet files
- `RemoveAction`s for old small files being replaced

Currently the Python API appears to accept only `AddAction`s:

```python
DeltaTable.create_write_transaction(
    actions: list[AddAction],
    ...
)
```

This is enough for append-only writes, but not for compaction. If old files cannot be removed in the same transaction, queries will read both old and new files and produce duplicate data.

## Proposed Solution

Expose a more general Python transaction API, or extend the existing one, to allow committing both add and remove actions atomically:

```python
table.create_transaction(
    actions=[
        AddAction(...),
        RemoveAction(...),
    ],
    operation="OPTIMIZE",
)
```

The exact API name is not important. The key requirement is the ability to commit add and remove file actions atomically through delta-rs.

## Alternatives

- `DeltaTable.optimize.compact()` works for built-in compaction, but does not support custom rewrite pipelines.
- Manually writing `_delta_log` JSON files is unsafe because it bypasses delta-rs transaction handling, conflict checks, protocol validation, and commit hooks.

_No response_

### Priority

None

### Additional context

_No response_

### Contribution

phofl:
# Description
Add support for RemoveAction together with AddAction

# Related Issue(s)
- closes #4480

# Documentation


phofl:
Do we need to write tests for the rust code specifically as well? The python side tests call into it, so it's covered from python

djouallah:
I would love to have this functionality

python/deltalake/table.py
```diff
@@ -1364,8 +1367,12 @@ def create_write_transaction(
         if not isinstance(schema, DeltaSchema):
             schema = DeltaSchema.from_arrow(schema)
 
+        add_actions = [a for a in actions if isinstance(a, AddAction)]
+        remove_actions = [a for a in actions if isinstance(a, RemoveAction)]
+
         self._table.create_write_transaction(
-            actions,
+            add_actions,
+            remove_actions,
```
ion-elgreco, line 1375: We shouldn't do the splitting in python, can you create an enum with a pyfromobject derive, like this:

```
#[derive(FromPyObject)]
pub enum PyFileAction {
    Add(PyAddAction),
    Remove(PyRemoveAction),
}

```

python/deltalake/transaction.py
```diff
@@ -22,6 +22,17 @@ class AddAction:
...
 
+@dataclass
+class RemoveAction:
+    path: str
+    data_change: bool
+    deletion_timestamp: int | None
+    extended_file_metadata: bool | None
+    partition_values: Mapping[str, str | None] | None
+    size: int | None
+    tags: Mapping[str, str | None] | None
```
ion-elgreco, line 33: Instead of extending it this way, I suggest we refactor so that we just expose the pyClasss to python and add a type hint to _inernal.pyi and then  exposed through transaction module

python/src/lib.rs
```diff
@@ -3107,11 +3143,23 @@ fn create_table_with_add_actions(
...
+        // iteration order wins" semantics. If an add and a remove target the same path
+        // within this single commit, ordering removes first ensures the remove wins (the
+        // file is correctly treated as not live), matching the semantics of a
+        // compaction/rewrite that both adds and removes the same path atomically.
+        let actions = remove_actions
+            .iter()
+            .map(|remove| Action::Remove(remove.into()))
+            .chain(add_actions.iter().map(|add| Action::Add(add.into())))
+            .collect::<Vec<_>>();
+
```
ion-elgreco, line 3157: That comment can be tightened

ion-elgreco (changes requested):
Sorry for the delay, been super busy lately. I left some comments I would like to see resolved and then we can merge it

seokjin0414:
@phofl since this has been quiet for a while, I've opened #4713 picking up #4480 (which I'm assigned to). It folds in @ion-elgreco's review feedback from this PR (the `PyFileAction` enum / pyclass approach) and rejects a same-path add+remove pair in one commit instead of ordering it, since the protocol doesn't define a winner inside a single commit. Happy to close mine in favor of this one if you're coming back to it.

PR #4601, opened 2026-07-16
feat: cdf feed respects in commit timestamps now

hntd187:
# Description
Previously CDF loads wouldn't interact with in-commit timestamps at all, but in-commit timestamps are more correct to use for version filtering than the standard commit timestamp. This protocol change I believe landed around the time we originally implemented CDF. 

Keep in mind, delta-rs still doesn't write in-commit timestamps (correct me if I'm wrong) this only modifies the read path to use them if they are available. We otherwise don't check for in-commit timestamps being enabled or any table features, we just check if they exist and use them if they do. 

# Documentation

https://github.com/delta-io/delta/issues/2532


crates/core/src/operations/load_cdf.rs
```diff
@@ -274,14 +278,26 @@ impl CdfLoadBuilder {
...
+                    }
+                    Some(Action::CommitInfo(CommitInfo {
+                        timestamp: Some(t), ..
+                    })) if starting_timestamp.timestamp_millis() > *t
+                        || *t > ending_timestamp.timestamp_millis() =>
+                    {
+                        log::debug!("Version: {version} skipped, due to commit timestamp");
+                        continue;
+                    }
+                    _ => {}
```
ethan-tyler, line 300: Should we choose/check one timestamp once? It looks like we are checking both `inCommitTimestamp` & `timestamp` independently in several places and I worry it allows them to contradict each other. Maybe a resolver would keep this consistent?

crates/core/src/operations/load_cdf.rs
```diff
@@ -147,16 +147,20 @@ impl CdfLoadBuilder {
         self
     }
 
+    #[inline]
+    fn timestamp_in_range(action: &Action, ts: DateTime<Utc>) -> bool {
+        matches!(action, Action::CommitInfo(CommitInfo { in_commit_timestamp: Some(t), .. }) if ts.timestamp_millis() < *t)
```
ethan-tyler: Should we use an inclusive comparison for earliest version selection? A commit whose timestamp equals `starting_timestamp` should belong in the cdc result but < can advance start to the next version and drop the boundary commit.

hntd187: Yes good idea, I thought to do this then ...just forgot

crates/core/src/operations/load_cdf.rs
hntd187: Well the timestamp field in the commit is technically the fallback to in commit timestamp so the precedence should be to check ICT first then timestamp second. Could you elaborate what you mean by contradicting each other?

ethan-tyler: I mean the two values can place the same commit on opposite sides of the requested range. The current match can fall through to the legacy timestamp when ICT is present but inside the range and allows the fallback value to override ICT

hntd187: Well we use one or the other, I suppose there is a possibility the user turns the feature on and off in the middle of a version range, but I can't imagine this would really cause a lot of issues.

ethan-tyler: Yeah makes sense, I agree that would be unlikely

ethan-tyler (approved):

PR #4603, opened 2026-07-17
fix: unify reserved commit metadata validation

### LINKED ISSUE #4457: Unify reserved commit metadata validation between Python and core
Python and core maintain the same reserved commit info key contract in two places:
- python/src/lib.rs (`validate_reserved_commit_metadata`)
- kernel/transaction/mod.rs (`RESERVED_COMMIT_INFO_KEYS` + `normalize`)

Python raises `ValueError` early, core logs and drops. `Duplicatedkey` list can create drift risk and a future change to commit provenance fields could update one path but not the other.

Options I see going foward with this are:
1. Keep the split, add tests asserting both lists stay aligned
2. Move reserved keys into a shared rust structure both paths use
3. Move more validation into core, expose structured errors to python

This will be closed when:
- Python and core agree on reserved key set
- Python still raises clear `ValueErrors` for invalid input
- Rust callers keep current compat behavior
- No duplicate commitInfo json keys emitted

adityavaish:
# Description

This PR removes the duplicated reserved `commitInfo` metadata-key contract between Rust core and the Python binding.

- Introduces `ReservedCommitInfoKey` in core as the shared classification for typed `CommitInfo` fields.
- Refactors core normalization to use that classification when promoting valid reserved values and removing reserved keys from flattened metadata, preventing duplicate JSON keys.
- Refactors Python custom-metadata validation to classify keys through core while preserving the existing early `ValueError` behavior and messages.
- Keeps Rust caller compatibility unchanged: valid reserved values are promoted, while invalid values remain tolerated and dropped by core.
- Marks the public enum non-exhaustive and safely rejects future unhandled reserved variants in Python.
- Expands Rust and Python regression coverage for key round-tripping, duplicate-free serialization, valid reserved values, and invalid value types.

# Related Issue(s)

- Closes #4457

# Documentation

No user-facing documentation changes are required. This centralizes an internal metadata contract without changing the supported Python API or Rust transaction behavior.

# Validation

Passed locally:

- `cargo clippy` (existing unrelated warnings only)
- `cargo fmt -- --check`
- `cargo test -p deltalake-core kernel::transaction::tests:: --lib`
- `cargo check -p deltalake-python`
- `ruff format --check --diff .`
- `ruff check .`
- `mypy`
- Targeted `python/tests/test_commit_properties.py` coverage: 18 passed
- `mkdocs build`

The full Python unit suite was also attempted on Windows. It reached unrelated existing platform assumptions around unsupported `forkserver` multiprocessing and `/C:/...` versus `C:/...` path normalization in table-read, deletion-vector, and vacuum tests. The metadata tests changed by this PR pass independently.

# AI usage disclosure

This contribution was generated with help from GitHub Copilot under human oversight. Copilot assisted with issue triage, implementation drafting, test expansion, review, and PR preparation. The author directed the approach, reviewed the resulting changes and validation output, and retains responsibility for the contribution. No third-party copyrighted material was introduced.


hntd187:
I fail to see a reason why this is necessary? This looks like a solution searching for a problem. This doesn’t make anything easier if commit info reserved fields change in the future. (In fairness it also doesn’t make it more difficult.)

What are you trying to solve with this change?

PR #4611, opened 2026-07-21
feat: support optimized sorting via Datafusion when Parquet files are ordered

adamreeve:


adamreeve:
Sorry, wrong repo.

In case anyone reads this and is interested, I'm just experimenting with an idea to make sorted reads faster when Parquet files are already written in sorted order, but I'm not sure whether it will be feasible to upstream this to delta-rs. I may make an issue to discuss this.

Issue #4614, opened 2026-07-24
Merge with `max_spill_size` set deadlocks indefinitely (FairSpillPool + RepartitionExec); unbounded completes in seconds


Gilbert09:
## Environment

- `deltalake` 1.6.1 (bundles DataFusion 54.0.0)
- Linux aarch64, Python 3.13
- Table storage: S3

## Call shape

```python
dt.merge(
    source,
    predicate="...",  # includes a per-partition predicate to scope the target
    streamed_exec=True,
).when_matched_update_all().when_not_matched_insert_all().execute()
```

with `max_spill_size = 6 GiB` and `max_temp_directory_size = 48 GiB` configured, merging a small source into a target partition whose merge working set exceeds `max_spill_size`.

## Symptom

`execute()` hangs indefinitely:

- CPU drops to ~0, disk I/O drops to ~0
- The spill directory freezes at ~557 MB — far below both `max_spill_size` and `max_temp_directory_size`
- No error is ever raised; the call never returns

Off-CPU profiling confirms the hang is a parked-thread deadlock, not slow progress: all native `deltalake._internal` threads sit in `__arm64_sys_futex` for the entire duration.

## Root cause

`create_session_state_with_spill_config` (crates/core/src/delta_datafusion/session.rs) builds a bounded DataFusion `FairSpillPool` when `max_spill_size` is set. Merge compiles into a `JoinType::Full` hash join (crates/core/src/operations/merge/mod.rs), and with the default `target_partitions = num_cpus` DataFusion inserts a parallel `RepartitionExec(Hash)`. The combination deadlocks:

1. Under the bounded pool, a spillable operator stalls waiting to spill.
2. Its consumer stops draining a `RepartitionExec` output channel.
3. The non-empty channel pins the global backpressure gate in `distributor_channels.rs`, parking all producers.
4. The hash-join build-side barrier never sees all partitions report.
5. Every tokio worker parks; the query can never make progress.

The memory bound is the sole trigger: with `max_spill_size` unset (unbounded pool), `try_grow` never fails, the channels drain, and the identical merge completes in ~8 seconds.

This is a known open DataFusion problem: apache/datafusion#17334, apache/datafusion#20715, apache/datafusion#22090 are all open, and the fix PR apache/datafusion#22092 was closed without merging. No released delta-rs or DataFusion version fixes it — delta-rs main and 1.6.2 still pin DataFusion 54.

## Proposed fix

[...]

PR #4615, opened 2026-07-24
feat(python): expose target_partitions on merge to control execution parallelism

Gilbert09:
# Description

Adds an optional `target_partitions` parameter to the Python `DeltaTable.merge()` API, controlling the number of partitions DataFusion uses to execute the merge (`datafusion.execution.target_partitions`). When not set, behavior is unchanged (DataFusion defaults to the number of CPU cores).

The motivating problem: a bounded-memory merge (`max_spill_size` set, which installs a `FairSpillPool`) can deadlock when the plan contains a partitioned hash join. `RepartitionExec` backpressure and the spilling join can end up waiting on each other, so the merge hangs (#4614; upstream DataFusion issues apache/datafusion#17334, apache/datafusion#20715, apache/datafusion#22090). Setting `target_partitions=1` removes the `RepartitionExec` from the plan entirely, which avoids that deadlock while keeping the memory bound.

The knob is a generic parallelism control rather than a deadlock-specific workaround — it is also useful to reduce the merge's CPU/memory footprint on shared machines.

A note on scope, from working through the review: the deadlock in #4614 does not reproduce reliably across machines or at small scales, so this PR does not include a test that reproduces it (a hanging merge is also unsuitable for CI). `target_partitions=1` is primarily a way to remove the `RepartitionExec`; it does not by itself force the merge to spill to disk (under single-partition execution the hash-join build side is not disk-spillable — it either fits within the pool or errors). The included test therefore verifies that the parameter is plumbed correctly and that a merge runs to correct results under a genuinely constraining memory pool, rather than asserting a specific spill or deadlock outcome.

Changes:

- Adds `create_session_state_with_execution_config(max_spill_size, max_temp_directory_size, target_partitions)` in `crates/core/src/delta_datafusion/session.rs`. The existing `create_session_state_with_spill_config(max_spill_size, max_temp_directory_size)` keeps its original two-argument signature and delegates to it, so this is not a breaking source change; in-tree callers such as `optimize` are unchanged.
- The Python bindings (`create_merge_builder` / `PyMergeBuilder`, plus the `compact_optimize` / `z_order_optimize` bindings that share the same session helper) thread the new parameter through the new function; only `DeltaTable.merge()` exposes it publicly for now.
- Rust unit tests asserting that the built session state reports the configured `target_partitions`, and that the two-argument wrapper still works. A Python merge test running `target_partitions=1` with a constraining `max_spill_size` over a 120k-row merge and checking the results.

Testing: `cargo check` for the core (`--features datafusion`) and Python binding crates passes; the new Rust unit tests pass (`cargo test -p deltalake-core --features datafusion --lib delta_datafusion::session`); and after `maturin develop` the reworked `test_merge_with_target_partitions` passes. `cargo fmt` / `ruff format` / `ruff check` are clean on the edited files. Commits are signed off (DCO).

# Related Issue(s)

- helps with #4614

# Documentation

- https://datafusion.apache.org/user-guide/configs.html (`datafusion.execution.target_partitions`)
- apache/datafusion#17334, apache/datafusion#20715, apache/datafusion#22090


crates/core/src/delta_datafusion/session.rs
```diff
@@ -25,21 +25,24 @@ pub fn create_session() -> DeltaSessionContext {
...
+/// of running out of memory.
 ///
 /// # Arguments
 /// * `max_spill_size` – Maximum bytes kept in memory before spilling. `None` uses DataFusion's default (unbounded) pool.
 /// * `max_temp_directory_size` – Maximum disk space for temporary spill files. `None` uses DataFusion's default disk manager.
+/// * `target_partitions` – Number of partitions DataFusion uses for query execution. `None` uses DataFusion's default (number of CPU cores).
 pub fn create_session_state_with_spill_config(
     max_spill_size: Option<usize>,
     max_temp_directory_size: Option<u64>,
+    target_partitions: Option<usize>,
```
ethan-tyler, line 60: I am not sure why we need to make a breaking change here? The same capability can be introduced while preserving downstream source compatibility

python/tests/test_merge.py
```diff
@@ -3546,3 +3546,75 @@ def do_merge():
...
+        },
+    )
+
+    dt.merge(
+        source=source_table,
+        predicate="t.id = s.id",
+        source_alias="s",
+        target_alias="t",
+        streamed_exec=streaming,
+        max_spill_size=100 * 1024 * 1024,
```
ethan-tyler: This only establishes a 100 MiB memory bound and it does not itself cause spilling.

ethan-tyler (commented):
Thanks @Gilbert09 - while I personally was not able to reproduce the issue, your strategy is directionally sound. Left a couple of comments and you also need to sign off your commits with DCO. After that we should be good to merge.

crates/core/src/delta_datafusion/session.rs
Gilbert09: Good call — reverted the signature change in d82bdfd. `create_session_state_with_spill_config` keeps its original two-argument form, and the `target_partitions` control now lives in a new `create_session_state_with_execution_config(max_spill_size, max_temp_directory_size, target_partitions)` that the old function delegates to. `optimize` and the other in-tree callers are unchanged, so there's no longer a breaking source change.

python/tests/test_merge.py
Gilbert09: You're right. Reworked the test in d82bdfd: it now runs a 120k-row merge under a genuinely constraining 16 MiB `FairSpillPool` (the same order as the join's working set) and asserts correct results, instead of the 100 MiB bound over three rows that was never reached.

While reworking it I also couldn't reproduce a hard deadlock at these scales (matching your experience), and I found that `target_partitions=1` doesn't itself trigger disk spilling here — under single-partition execution the hash-join build side isn't disk-spillable, so it either fits or errors (disk-on vs disk-off made no difference). Its real effect is removing the `RepartitionExec`, so I've framed the test comment and the PR description around that rather than claiming it forces spilling. Happy to adjust further if you'd prefer a different framing.

PR #4617, opened 2026-07-26
fix: when merging on non-null field withe mergeSchema enabled, the pr…

### LINKED ISSUE #4527: [Bug] MERGE with merge_schema=True wrongly fails NOT NULL check on rows not matched by the source
### What happened?

On DeltaTable.merge(), with merge_schema=True, delta-rs validates an intermediate merge projection - the outer-join batch in which source.* is NULL for target rows unmatched by the source - against the table's NOT-NULL column invariants. Those rows legitimately have a NULL merge-key column INSIDE the plan, but the final output would resolve them from the target side. The validator fires before that resolution and spuriously rejects the merge. Proof of false positive: with the key column made NULLABLE (T2), the very same merge commits and the final table contains ZERO null document_id values.

Tested on delta-rs 1.5.1 and 1.6.0.

Repro script:

```
import shutil
import uuid

import pyarrow as pa
import deltalake

print("deltalake", deltalake.__version__, "| pyarrow", pa.__version__)

IDS = [str(uuid.uuid4()) for _ in range(100)]


def make_target(path, key_nullable):
    schema = pa.schema([
        pa.field("document_id", pa.string(), nullable=key_nullable),
        pa.field("file_path", pa.string()),
        pa.field("thumbnail_path", pa.string()),   # all-null at create (like gold)
        pa.field("preview_count", pa.int64()),     # all-null at create
        pa.field("row_id", pa.int64()),
    ])
    t = pa.table({
        "document_id": pa.array(IDS),
        "file_path": pa.array(["s3://x/%d.pdf" % i for i in range(100)]),
        "thumbnail_path": pa.nulls(100, pa.string()),
        "preview_count": pa.nulls(100, pa.int64()),
        "row_id": pa.array(range(100), pa.int64()),
    }).cast(schema)
    shutil.rmtree(path, ignore_errors=True)
    write_deltalake(path, t, mode="append")
    return DeltaTable(path)


def subset_source():
    sub = IDS[:68]
    return pa.table({
        "document_id": pa.array(sub),
        "thumbnail_path": pa.array(["thumb/%s.png" % i for i in sub]),
        "preview_count": pa.array([4] * 68, pa.int64()),
        "row_id": pa.array(range(1000, 1068), pa.int64()),
    })


def run_merge(dt, merge_schema):
    src = subset_source()
    op = dt.merge(source=src, source_alias="source", target_alias="target",
                  predicate="source.document_id = target.document_id",
                  merge_schema=merge_schema)
    updates = {c: "source.%s" % c for c in src.schema.names}
    op = op.when_matched_update(updates=updates,
                                predicate="(source.thumbnail_path IS DISTINCT FROM target.thumbnail_path)")
    op = op.when_not_matched_insert_all()
    return op.execute()


for label, key_nullable, merge_schema in [("T1 nonnull-key, merge_schema=True", False, True),
                                          ("T2 NULLABLE-key, merge_schema=True", True, True),
                                          ("T3 nonnull-key, merge_schema=False", False, False)]:
    path = "/tmp/repro_delta_" + label.split()[0].lower()
    dt = make_target(path, key_nullable)
    try:
        m = run_merge(dt, merge_schema)
        out = DeltaTable(path).to_pyarrow_table()
        d = out.column("document_id").to_pylist()
        nulls = [i for i, x in enumerate(d) if not x]
        print("%s: MERGE OK | rows=%d null_document_id=%d" % (label, out.num_rows, len(nulls)))
        if nulls:
            fp = out.column("file_path").to_pylist()
            tp = out.column("thumbnail_path").to_pylist()
            matched_fp = {("s3://x/%d.pdf" % i) for i in range(68)}
            nulled_were_matched = sum(1 for i in nulls if fp[i] in matched_fp)
            print("   -> CORRUPTION VISIBLE: %d rows lost document_id; of them %d were MATCHED rows, %d UNMATCHED"
                  % (len(nulls), nulled_were_matched, len(nulls) - nulled_were_matched))
            print("   sample nulled row: file_path=%s thumbnail=%s" % (fp[nulls[0]], tp[nulls[0]]))
    except Exception as e:
        print("%s: MERGE FAILED -> %s: %s" % (label, type(e).__name__, str(e).splitlines()[0][:120]))
```

Result:
```
deltalake 1.6.0 | pyarrow 23.0.1
T1 nonnull-key, merge_schema=True: MERGE FAILED -> DeltaError: Generic DeltaTable error: External error: Invalid data found: 32 rows failed validation check.
T2 NULLABLE-key, merge_schema=True: MERGE OK | rows=100 null_document_id=0
T3 nonnull-key, merge_schema=False: MERGE OK | rows=100 null_document_id=0
```

### Expected behavior

All three merge operations should succeed with zero null key field rows.

### Operating System

Linux

### Binding

Python

### Bindings Version

_No response_

### Steps to reproduce

Run the repro code above.

### Relevant logs

```shell

```

MedHeikelBouzayene:
# Description
This fixes the problem when we merge on a non-nullable columns while mergeSchema is enabled. The bug happens when some rows are existing in the target delta table and not in the source table.

.P.S: From my point of view, the field should be defined nullable or not nullable only from the delta table, and not from both target table and source table.

# Related Issue(s)
- closes #4527


MedHeikelBouzayene:
Why exactly unity catalog test is failing?

MedHeikelBouzayene:
@ion-elgreco could you please re-run the test, and any idea why unity catalog test was failing?

MedHeikelBouzayene:
@ion-elgreco sorry for many requests, I did fix the issues in last run and added some tests to increase test coverage, could you please re-run

crates/core/src/operations/merge/mod.rs
```diff
@@ -2051,6 +2052,57 @@ mod tests {
         assert_batches_sorted_eq!(&expected, &actual);
     }
 
+    #[tokio::test]
+    async fn test_merge_non_null() {
```
JustinRush80, line 2082: For this test, it will be better to have the non-null column variant to be included in the not match clauses/ match clauses since those are the columns we are conducting the merge schema on. For example, if id column was included in the when_not_matched_insert does it nullability changes.

MedHeikelBouzayene: Hello @JustinRush80, I did update the test to verify that nullability of that field didn't change, also did add assert to make sure.

crates/core/src/kernel/schema/cast/merge_schema.rs
```diff
@@ -146,7 +146,7 @@ pub(crate) fn merge_arrow_field(
             Ok(ArrowField::new(
                 right.name(),
                 Dictionary(key_type.clone(), Box::new(batch_type.clone())),
-                left.is_nullable() || right.is_nullable(),
+                left.is_nullable(),
```
JustinRush80, line 149: Of course @ion-elgreco and other reviewers are the expert here. But I believe it not safe to assume "the field should be defined nullable or not nullable only from the delta table" since merge_schema.rs is also used in the schema overwrite mode for the Write operation. This change wouldn't allow the schema overwrite mode to completely change the preexisting delta table schema.

MedHeikelBouzayene: @JustinRush80 The problem in the existing implementation, once the field is nullable(from source or target delta table) it will be will always nullable and couldn't change it also. So here at least, the nullability of fields is coming from the target delta table. I am also still waiting for other reviewers opinion about that

MedHeikelBouzayene:
Hello @ion-elgreco any update on this PR

rtyler (approved):

PR #4624, opened 2026-07-27
fix: percent encode spaces in action paths

### LINKED ISSUE #4304: [Bug]: Commits don't %-encode spaces in AddFile / RemoveFile paths
### What happened?

During some testing, I noticed path-encoding isn’t handled in accordance with the delta-lake spec, because spaces in AddFile action paths don't get %-encoded.

The problem occurs during the serialization of AddFile actions into json-files during the commit-process (and presumably affects other actions such as RemoveFile, from a cursory glance at the other code).

The PreCommit::into_prepared_commit_future implementation takes the stored data field (of type CommitData) and then executes the get_bytes method on it. Within get_bytes, we create the json files to write into the new _delta_log entry by using serde_json::to_string:

```
pub fn get_bytes(&self) -> Result<bytes::Bytes, TransactionError> {
      let mut jsons = Vec::<String>::new();
      for action in &self.actions {
          let json = serde_json::to_string(action)
              .map_err(|e| TransactionError::SerializeLogJson { json_err: e })?;
          jsons.push(json);
      }
      Ok(bytes::Bytes::from(jsons.join("\n")))
  }
```
In the file `src/kernel/models/actions.rs`, the serialization of the path is specified to use “serde_path.“ Here’s the relevant snippet of the Add action:
```
[derive(Serialize, Deserialize, Debug, Clone, Default)]
/// Defines an add action
pub struct Add {
    /// A relative path to a data file from the root of the table or an absolute path to a file
    /// that should be added to the table. The path is a URI as specified by
    /// [RFC 2396 URI Generic Syntax], which needs to be decoded to get the data file path.
    ///
    /// [RFC 2396 URI Generic Syntax]: https://www.ietf.org/rfc/rfc2396.txt
    #[serde(with = "serde_path")]
    pub path: String,
```
The ultimate bug is in the serde_path implementation, which encodes only the characters that are present in this set:
```
const INVALID: &AsciiSet = &CONTROLS
        // The delimiter we are reserving for internal hierarchy
        // .add(DELIMITER_BYTE)
        // Characters AWS recommends avoiding for object keys
        // https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
        .add(b'\\')
        .add(b'{')
        .add(b'^')
        .add(b'}')
        .add(b'%')
        .add(b'`')
        .add(b']')
        .add(b'"')
        .add(b'>')
        .add(b'[')
        // .add(b'~')
        .add(b'<')
        .add(b'#')
        .add(b'|')
        // Characters Google Cloud Storage recommends avoiding for object names
        // https://cloud.google.com/storage/docs/naming-objects
        .add(b'\r')
        .add(b'\n')
        .add(b'*')
        .add(b'?');
```
Notably, the CONTROLS set [explicitly omits the space character](https://docs.rs/percent-encoding/2.3.2/percent_encoding/constant.CONTROLS.html):
> The set of 0x00 to 0x1F (C0 controls), and 0x7F (DEL).
> Note that this includes the newline and tab characters, but not the space 0x20.
> https://url.spec.whatwg.org/#c0-control-percent-encode-set   

However, [according to the delta-lake protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#how-to-url-encode-keys-and-string-values ), spaces should be encoded, since they’re not in the list of non-reserved characters.

Practically, if there's another writer into the table, where the other writer _is_ encoding spaces, this could multiply the number of distinct directories that exist in a table, which makes it harder to inspect the table through simple listing commands, etc., especially for tables that embed partition values into the directory paths. 

Some implementations, like Spark, may return an error upon encountering spaces that are not encoded, as [illustrated elsewhere](https://github.com/delta-io/delta-rs/issues/2978#issuecomment-2461003888).

Such inconsistency can also cause deleted data to be read, as follows:
- deltalake_core writes an AddFile for the file F, which does not encode spaces
- A different writer creates a RemoveFile for the file F, which does encode spaces
- During Action Reconciliation, some implementation (call it IMPL) encounters these actions and does not decode their paths. When interacting with a fully valid table, such decoding should be unnecessary during Action Reconciliation (I think?), and so it would be reasonable, as a performance optimization, for IMPL to avoid such decoding. Consequently, IMPL compares the paths in the AddFile and RemoveFile, finds that they differ, and does not remove F from the resulting snapshot. 

From what I can tell, delta-kernel-rs instantiates the relevant qualities of IMPL. That is, delta-kernel-rs's action-reconciliation implementation does not perform decoding of AddFile / RemoveFile paths. delta-kernel-rs compares FileActionKeys to determine, amongst other things, whether a given RemoveFile corresponds to a given AddFile. FileActionKeys contain paths, which are extracted within FileActionDeduplicator::extract_file_action via GetData::get_str:
```
  // Try to extract an add action by the required path column
  if let Some(path) = getters[self.add_path_index].get_str(i, "add.path")? {
      let dv_unique_id = self.extract_dv_unique_id(i, getters, self.add_dv_start_index)?;
      return Ok(Some((FileActionKey::new(path, dv_unique_id), true)));
  }
```
The GetData::get_str trait documentation does not require URI decoding of its inputs, which makes sense since get_str may be used on arbitrary string data, not just %-encoded data.

Consequently, the fix is easy, but it may not be backwards compatible w.r.t. existing tables that have Parquet file paths with spaces in them, depending upon how exactly the user interacts with their table(s). For example, if an existing table were only ever modified by deltalake_core, and a user upgraded their deltalake_core version to one that fixes this problem, this would result in a table that uses an inconsistent approach to space encoding, for files added/removed before vs. after the upgrade. Then, if the user relies upon an action-reconciliation implementation that matches directly on the path, as in the case of delta-kernel-rs (and I think deltalake_core via delta-kernel-rs), they could read already deleted data.

I can see three ways forward in such a case:

1. Fix the issue and gate it behind a SemVer break (is that sufficient?)
2. Possibly in addition to 1, ensure that deltalake_core / delta-kernel-rs action-reconciliation can handle inconsistent encoding
3. Possibly in addition to 1 (but not 2), offer a migration tool that re-writes and normalizes the relevant paths into a new _delta_log checkpoint file

### Expected behavior

Encode spaces as %20 in AddFiles (and other _delta_log actions as appropriate)
Note that RFC 2396 uses %20 to encode spaces, not '+':

https://www.ietf.org/rfc/rfc2396.txt

See the section titled 'Escaped Encoding'

### Operating System

Linux

### Binding

Rust

### Bindings Version

0.31.1

### Steps to reproduce

I didn’t setup a shareable end-to-end reproduction case that is mediated by CommitBuilder, but this demonstrates the serialization bug:
```
use deltalake_core::kernel::models::Add;
use std::collections::HashMap;

fn main() {
    const PATH: &str = "/path segment with spaces/path*segment*with*currently*invalid*character";

    let add = Add {
        path: PATH.to_string(),
        partition_values: HashMap::new(),
        size: 100,
        modification_time: 1700000000000,
        data_change: true,
        stats: None,
        tags: None,
        deletion_vector: None,
        base_row_id: None,
        clustering_provider: None,
    };

    let json = serde_json::to_string_pretty(&add).expect("Failed to serialize Add action");

    println!("=== Serialized Add action (JSON) ===");
    println!("{json}");

    // Extract the path from the serialized JSON
    let parsed: serde_json::Value = serde_json::from_str(&json).unwrap();
    let serialized_path = parsed["path"].as_str().unwrap();

    println!("\n=== Analysis ===");
    println!("Original path:   {PATH}");
    println!("Serialized path: {serialized_path}");

    assert!(serialized_path.contains(' '));
    assert!(! serialized_path.contains("%20"));
    assert!(serialized_path.contains("%2A"));
}
```
```
Cargo.toml:

[package]
name = "delta-rs-encoding-bug-repro"
version = "0.1.0"
edition = "2024"

[dependencies]
deltalake-core = "0.31.1"
serde_json = "1"
```

### Relevant logs

```shell
N/A
```

SreeramaYeshwanthGowd:
# Description

Action paths are URIs per the protocol, but `serde_path` encoded them with an object store set whose `CONTROLS` omits the space, so spaces reached the log unencoded. The path set stays looser than the partition value one so `/` and `=` stay literal.

# Related Issue(s)

- closes #4304

### DIFF
--- a/crates/core/src/kernel/models/actions.rs
+++ b/crates/core/src/kernel/models/actions.rs
@@ -1231,7 +1231,8 @@ impl FromStr for IsolationLevel {
 pub(crate) mod serde_path {
     use std::str::Utf8Error;
 
-    use percent_encoding::{AsciiSet, CONTROLS, percent_decode_str, percent_encode};
+    use percent_encoding::percent_decode_str;
+    use percent_encoding_rfc3986::{AsciiSet, CONTROLS, utf8_percent_encode};
     use serde::{self, Deserialize, Deserializer, Serialize, Serializer};
 
     pub fn deserialize<'de, D>(deserializer: D) -> Result<String, D::Error>
@@ -1255,9 +1256,14 @@ pub(crate) mod serde_path {
     pub const _DELIMITER_BYTE: u8 = _DELIMITER.as_bytes()[0];
 
     /// Characters we want to encode.
+    ///
+    /// Paths are URIs, so the hierarchy delimiter stays literal. Partition values are single
+    /// strings and go through the stricter set in [`crate::kernel::scalars`] instead.
     const INVALID: &AsciiSet = &CONTROLS
         // The delimiter we are reserving for internal hierarchy
         // .add(DELIMITER_BYTE)
+        // RFC 2396 excludes the space from URIs and CONTROLS does not cover it, see #4304
+        .add(b' ')
         // Characters AWS recommends avoiding for object keys
         // https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
         .add(b'\\')
@@ -1282,9 +1288,10 @@ pub(crate) mod serde_path {
         .add(b'?');
 
     fn encode_path(path: &str) -> String {
-        percent_encode(path.as_bytes(), INVALID).to_string()
+        utf8_percent_encode(path, INVALID).to_string()
     }
 
+    // rfc3986's percent_decode_str is fallible, so decoding stays on the url spec crate.
     pub fn decode_path(path: &str) -> Result<String, Utf8Error> {
         Ok(percent_decode_str(path).decode_utf8()?.to_string())
     }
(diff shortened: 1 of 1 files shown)


SreeramaYeshwanthGowd:
On why this is a correctness fix rather than conformance tidying. Deleting a Spark written file whose log path is city=New%20York/f.parquet currently emits a remove carrying a literal space, which never matches the add. Kernel dedups on the raw log string, FileActionKey is built straight off get_str(i, "add.path") with no decoding, so the tombstone is dropped. The new test_serialize_round_trips_spark_written_path is that reproducer, and its literal is byte exact from dat/v0.0.3/reader_tests/generated/multi_partitioned_2. Reverting the single space entry fails 3 of the 5 new tests.

Upgrade disclaimer, @ion-elgreco, since you asked for one on the issue. An add written before this change with a literal space will not be suppressed by a remove written after it. Normal writes are not exposed, since scalars.rs already percent encodes the space in partition values and serde_path then double encodes it to %2520. The realistic sources are convert_to_delta and external writers that do not encode.

On consolidating to one set, RFC3986_PART encodes the slash, the equals and the colon, so reusing it for a path gives year%3D2021%2Ff.parquet, an action pointing at an object that does not exist. test_serialize_path_preserves_hive_partition_delimiters locks that in. Worth noting the kernel encoder for add.path already includes the space, in partition/hive.rs, so this converges the two rather than diverging.

rtyler (approved):
This looks sensible to me and all the integration tests are passing. I will defer to @ion-elgreco for the merge however

PR #4629, opened 2026-07-30
fix(unity): refresh Unity Catalog's temporary object store credentials

### LINKED ISSUE #4628: [Bug]: Unity Catalog uc:// tables fail with ExpiredToken after ~1 hour
### What happened?

Unity Catalog vends temporary S3 credentials that expire after about an hour. The uc:// object-store path fetches them once when the table is opened and passes them to the S3 store as static credentials, so nothing ever refreshes them.

### Expected behavior

tokens should have refreshed internally

### Operating System

macOS

### Binding

Rust

### Bindings Version

v0.32.3 / main branch

### Steps to reproduce

1. Open an S3-backed table via uc://<catalog>.<schema>.<table>.
2. Keep the handle alive and idle for over an hour.
3. Read from it -> ExpiredToken.

### Relevant logs

```shell
ERROR read_commit_entry{ version=13 path=_delta_log/00000000000000000013.json}: deltalake_core::logstore: failed to read commit entry error=Generic S3 error: Error performing GET https://../_delta_log/00000000000000000013.json in 478.848433ms - Server returned non-2xx status code: 400 Bad Request: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>ExpiredToken</Code><Message>The provided token has expired.</Message>
```

swanandx:
Unity Catalog vends temporary S3 tokens (~1h). The uc:// object-store path fetched them once at table open and baked them in as static credentials, so long-lived stores started failing with ExpiredToken after an hour.

# Description

Carry an optional CloudCredentialProvider on StorageConfig and have the AWS,
Azure and GCP factories prefer it over credentials resolved from the raw
options. catalog-unity builds one for the cloud hosting the table, which
re-vends from Unity Catalog shortly before the cached token expires, with a
floor on the re-vend rate so a stale expiration_time cannot turn into one
catalog request per object store operation.

This also fixes GCP, where the vended oauth token was passed as
google_application_credentials, which object_store reads as a path to a
service account file rather than as a bearer token.

# Related Issue(s)

- closes #4628 

# Documentation

### DIFF
--- a/crates/aws/src/storage.rs
+++ b/crates/aws/src/storage.rs
@@ -15,8 +15,8 @@ use deltalake_core::logstore::object_store::{
     Result as ObjectStoreResult, path::Path,
 };
 use deltalake_core::logstore::{
-    ObjectStoreFactory, ObjectStoreRef, StorageConfig, client_options_from_certificate,
-    config::str_is_truthy,
+    CloudCredentialProvider, ObjectStoreFactory, ObjectStoreRef, StorageConfig,
+    client_options_from_certificate, config::str_is_truthy,
 };
 use deltalake_core::{DeltaResult, DeltaTableError};
 use futures::Future;
@@ -69,7 +69,9 @@ impl ObjectStoreFactory for S3ObjectStoreFactory {
         }
 
         let s3_options = S3StorageOptions::from_map(&options)?;
-        if let Some(ref sdk_config) = s3_options.sdk_config {
+        if let Some(CloudCredentialProvider::Aws(credentials)) = &config.credentials {
+            builder = builder.with_credentials(credentials.clone());
+        } else if let Some(ref sdk_config) = s3_options.sdk_config {
             builder =
                 builder.with_credentials(Arc::new(AWSForObjectStore::new(sdk_config.clone())));
         }
--- a/crates/azure/src/lib.rs
+++ b/crates/azure/src/lib.rs
@@ -3,8 +3,9 @@ use std::str::FromStr;
 use std::sync::Arc;
 
 use deltalake_core::logstore::{
-    LogStore, LogStoreFactory, ObjectStoreFactory, ObjectStoreRef, StorageConfig,
-    client_options_from_certificate, default_logstore, logstore_factories, object_store_factories,
+    CloudCredentialProvider, LogStore, LogStoreFactory, ObjectStoreFactory, ObjectStoreRef,
+    StorageConfig, client_options_from_certificate, default_logstore, logstore_factories,
+    object_store_factories,
 };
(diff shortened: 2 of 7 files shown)


hntd187:
So, what about other cloud providers? Why does AWS only get this? 

Why not just recreate the table object? If the tokens are expired you have to refresh the table state anyways? Table objects aren't meant to be long lived objects you keep around, as their token expiration implies they are meant to be short lived objects for their use and removal later. 

I'll be honest, I don't see a reason to ever merge this if it's only going to be AWS specific. Even if it wasn't this is a lot of code to account for what I assume is very much not the normal loop for most users.

swanandx:
> So, what about other cloud providers? Why does AWS only get this?
> 

that is what i tried with and wanted to keep scope small in initial PR, I can extend it to Azure and GCP.

> Why not just recreate the table object? If the tokens are expired you have to refresh the table state anyways? Table objects aren't meant to be long lived objects you keep around, as their token expiration implies they are meant to be short lived objects for their use and removal later.
> 
token refresh is existing behavior for normal `s3://` / other parts. `uc://` was the odd one here, so trying to make it consistent. Also a single operation on big table _may_ last more than token, in that case it won't make sense to recreate table object.


> I'll be honest, I don't see a reason to ever merge this if it's only going to be AWS specific. Even if it wasn't this is a lot of code to account for what I assume is very much not the normal loop for most users.
>

about 300 of the added lines are tests. The provider is roughly 100 lines and the deltalake-aws change is 22. Most of the rest is removing duplication that was already there, the uc:// URI parsing and the READ_WRITE to READ fallback both existed twice.

Happy to do the generic version if you're open to the approach in principle. and open to suggestions for other approaches! 

thanks!

hntd187:
I'm not open to entirely re-writing and re-implementing a bunch of things we already had in addition to testing that looks to be also redundant but added because of the new path here. You're more than welcome to go for a generic approach that works for all cloud providers, I'll switch this to draft in the mean time.

swanandx:
> I'm not open to entirely re-writing and re-implementing a bunch of things we already had in addition to testing that looks to be also redundant but added because of the new path here. You're more than welcome to go for a generic approach that works for all cloud providers, I'll switch this to draft in the mean time.

thank you so much for feedback. i pushed the generic version implementation.

hntd187:
Thanks but as I mentioned before I am not interested in everything being rewritten by AI, so thanks, but we won't be merging this.

swanandx:
> Thanks but as I mentioned before I am not interested in everything being rewritten by AI, so thanks, but we won't be merging this.

fair enough. I just submitted whatever workaround I had in my fork to get around #4628 . I'm happy as long as the reported issue gets resolved in upstream by one way or another :)

thanks for taking a look anyways, appreciate it!

PR #4631, opened 2026-08-05
python: add Delta merge continuous aggregation example

dylangriffinshub:
## What
Adds `python/examples/merge_streaming_aggregation.py`: a runnable, dependency-light example of a Delta Lake continuous-aggregation pipeline built on `DeltaTable.merge`.

## Why
The example demonstrates the pattern behind a Spark-based continuous aggregation framework using Delta merge logic:
- Idempotent event ingestion: `ingest_events` MERGEs micro-batches into a raw Delta table keyed on a stable `event_id`, so re-delivered or late events never duplicate.
- Incremental rollup: `aggregate` MERGEs only the new micro-batch into a curated table keyed on the aggregate key, accumulating counts/sums on match and inserting new keys, so metrics stay accurate at scale without a full-table recompute or double counting.

It runs locally with no Spark cluster (pandas + pyarrow) and the same `write_deltalake` / `DeltaTable.merge` calls run unchanged on Databricks against DBFS.

## Verification
```
pip install deltalake pyarrow pandas
python python/examples/merge_streaming_aggregation.py
```
Prints the curated rollup and `MERGE_STREAMING_AGGREGATION_DONE`; re-delivering batch 1 leaves the raw count at 10 (idempotent) and the rollup unchanged.

## Checklist
- [x] Example runs end to end
- [x] No new public API surface

### DIFF
--- a/python/examples/merge_streaming_aggregation.py
+++ b/python/examples/merge_streaming_aggregation.py
@@ -1,0 +1,168 @@
+"""Delta Lake continuous aggregation pipeline with idempotent Delta MERGE.
+
+Demonstrates the pattern behind a "Spark-based continuous aggregation
+framework using Delta merge logic":
+
+1. Ingest event data in batches (the same code path works for a streaming
+   micro-batch: call `ingest_events` once per micro-batch).
+2. Upsert events into the raw Delta table with MERGE keyed on a stable event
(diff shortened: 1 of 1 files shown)


hntd187:
Thanks, but people can generate their own AI examples, we don't need to check them in here.

PR #4632, opened 2026-08-06
fix(core): do not panic serializing out of range Scalar::Date

manjunathbhaskar:
## Summary

`Scalar::Date` wraps an `i32` days since epoch value that is read verbatim from a DATE typed partition value or a min/max stat in the transaction log or a checkpoint parquet file, with no range validation against the Delta protocol. A corrupted or maliciously crafted table can set this to any `i32`, including values whose implied timestamp (`days * 86400` seconds) falls outside the range chrono's `DateTime` can represent (roughly +/-262,000 years).

Both `Scalar::serialize()` and `Scalar::to_json()` called `DateTime::from_timestamp(...).unwrap()` on this value, which panics on `None`. This is a remotely triggerable denial of service for any code path that formats a DATE partition value coming from an untrusted or corrupted table, for example computing a hive partition path or displaying add action stats.

## Fix

Both call sites now match on the `Option` returned by `DateTime::from_timestamp` and fall back to a plain numeric representation (`invalid-date(days=...)`) instead of panicking. Exact formatting for absurd out of range dates is not load bearing, only that the process degrades gracefully instead of crashing.

## Test plan

Added regression tests covering `i32::MAX`, `i32::MIN`, and their halves for both `serialize()` and `to_json()`:

```
test kernel::scalars::tests::test_scalar_serialize_date_out_of_range_does_not_panic ... ok
test kernel::scalars::tests::test_scalar_to_json_date_out_of_range_does_not_panic ... ok
```

Full `kernel::scalars` module: 35 passed, 0 failed.

### DIFF
--- a/crates/core/src/kernel/scalars.rs
+++ b/crates/core/src/kernel/scalars.rs
@@ -92,8 +92,18 @@ impl ScalarExt for Scalar {
                 ts.format("%Y-%m-%d %H:%M:%S%.6f").to_string()
             }
             Self::Date(days) => {
-                let date = DateTime::from_timestamp(*days as i64 * 24 * 3600, 0).unwrap();
-                date.format("%Y-%m-%d").to_string()
+                // `days` is read verbatim from a DATE-typed partition value or
+                // min/max stat in the transaction log / checkpoint parquet, with
+                // no range validation against the Delta protocol. A corrupted or
+                // maliciously crafted table can set this to any i32, including
+                // values whose implied timestamp falls outside the range chrono's
+                // `DateTime` can represent (roughly +/-262,000 years). Fall back to
+                // a raw numeric representation instead of panicking so that a
+                // single bad DATE value can't crash the process.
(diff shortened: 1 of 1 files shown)


ion-elgreco:
Theoretically it's possible  to have date that falls out of that range, but In this case a panic is better than corrupting the output. Now the serialized value doesn't represent the value, but this will break a couple things downstream as encoding and partition values info.

manjunathbhaskar:
Yeah, fair. I traced it properly after your comment: serialize() backs serialize_encoded(), which is what feeds encode_partition_value(), and convert_to_delta puts serialize() straight into partition_values. So that fallback string would end up as an actual partition directory name and get written into the log. Worse than the panic, agreed.

Parking it. If it's ever worth revisiting, the place to catch this is at parse time so an out of range Scalar::Date never gets constructed, which keeps the serialized output honest and still avoids the crash. Not going to push that unless you think it's worth having.

ion-elgreco:
We can revisit once a user runs into this in a real pipeline

PR #4633, opened 2026-08-06
fix(core): prevent path traversal via percent-encoded dot-segments in to_uri

### LINKED ISSUE #4637: [Bug]: to_uri can resolve outside the table root when add.path contains a percent-encoded ".."
### What happened?

to_uri (crates/core/src/logstore/mod.rs) builds a file's URI by joining the table root with a path taken from add.path or remove.path. object_store's Path treats a percent-encoded ".." as just another segment and accepts it, but Url::join actually decodes and resolves it. So a bad add.path can make to_uri point somewhere outside the table.

If the root URL ends in a slash, this walks straight past the table root.

If it doesn't, Url::join treats the last path segment as a filename and replaces it. That can land you inside a completely different table if the name happens to overlap, e.g. root s3://bucket/tenant-a plus path %2e%2e/tenant-a-evil/x.json gives s3://bucket/tenant-a-evil/x.json.

The second one isn't just a corner case either. LogStoreConfig's Deserialize impl parses the root with a raw Url::parse, so it never goes through the step that normally adds the trailing slash.

### Expected behavior

to_uri should stay under the table root it was given. A path out of the log shouldn't be able to resolve into another table's storage.

### Operating System

macOS

### Binding

Rust

### Bindings Version

1.0.0 (main, fd7e9691)

### Steps to reproduce

No bucket needed, this is all string handling:

```rust
use deltalake_core::logstore::to_uri;
use object_store::path::Path;
use url::Url;

fn main() {
    let root = Url::parse("s3://bucket/tenant-a").unwrap();
    let evil = Path::parse("%2e%2e/tenant-a-evil/orders_table/_delta_log/00000000000000000000.json").unwrap();
    println!("{}", to_uri(&root, &evil));
}
```

### Relevant logs

s3://bucket/tenant-a-evil/orders_table/_delta_log/00000000000000000000.json

manjunathbhaskar:
Fixes #4637

to_uri was joining the table root with a path straight out of the log and never checking where it ended up. object_store's Path is fine with a percent-encoded "..", Url::join resolves it, and the result can point outside the table. With no trailing slash on the root it can also end up inside a neighbouring table whose name starts the same way.

The join result is now only used if it's still under the root, compared against a prefix that always ends in a slash so a name like tenant-a-evil doesn't pass for tenant-a. Otherwise it falls back to concatenating, which is what the file branch above already does.

Added a test per case. Both fail without the change.

Ran the deltalake-core lib tests and clippy with the CI feature set.

### DIFF
--- a/crates/core/src/logstore/mod.rs
+++ b/crates/core/src/logstore/mod.rs
@@ -687,16 +687,28 @@ pub fn to_uri(root: &Url, location: &Path) -> String {
             if location.as_ref().is_empty() || location.as_ref() == "/" {
                 root.as_ref().to_string()
             } else {
+                // Comparing against a bare root string is not enough: without a
+                // trailing slash, a sibling table like "tenant-a-evil" also
+                // passes `starts_with("tenant-a")`. Force the boundary here
+                // instead of relying on root always being normalized upstream.
+                let root_str = root.as_str();
+                let root_prefix = if root_str.ends_with('/') {
+                    root_str.to_string()
+                } else {
+                    format!("{root_str}/")
+                };
                 match root.join(location.as_ref()) {
-                    // `Url::join` applies WHATWG dot-segment normalization to
-                    // percent-encoded "." / ".." segments (e.g. "%2e%2e"), even though
-                    // `object_store::path::Path` treats them as opaque, non-special path
-                    // segments. `location` comes from an `add.path` / `remove.path` value
-                    // read verbatim from the transaction log or a checkpoint parquet file,
-                    // so a corrupted or maliciously crafted table could otherwise make this
-                    // resolve outside the table's own storage prefix. Only trust the joined
-                    // result when it is still rooted under `root`.
-                    Ok(joined) if joined.as_str().starts_with(root.as_str()) => joined.to_string(),
+                    // Url::join normalizes percent-encoded ".." even though
+                    // object_store::path::Path treats it as an ordinary
+                    // segment, so a bad add.path can otherwise walk outside
+                    // the table root. Only trust the result if it's still
+                    // under root.
+                    Ok(joined)
+                        if joined.as_str() == root_str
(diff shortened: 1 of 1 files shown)


ion-elgreco:
The description just screams AI here, fine if you use it, but please tighten and condense the message and create an issue first with an MRE.

manjunathbhaskar:
Opened #4637 with a repro, and cut the description down.

Ran into a second case while writing that up. If the table root has no trailing slash, the same bad path doesn't walk above the root, it replaces the last segment and can land in a neighbouring table with a similar name. LogStoreConfig deserializes the root with a raw Url::parse so the trailing slash isn't guaranteed. Covered both with tests.

hntd187:
Where in the code is such a situation possible? Did you encounter this in a normally created table or is this a speculative problem to a problem that _might_ exist. Your issue doesn't actually point to a place where this code could actually be possible.

manjunathbhaskar:
It's not a normal write, you'd need a log with a bad add.path in it. But nothing checks that path against the table root. get_file_uris runs each add's path through to_uri, and for an s3 or azure or gcs root to_uri uses Url::join, which folds %2e%2e into .. and climbs out of the table. file:// just concatenates so it stays put.

You can see it without a bucket. Add a commit with path %2e%2e/%2e%2e/other-table/f.parquet, and get_file_uris on a local table carries the %2e%2e straight through into the resolved path, nothing strips it. The escape itself is the Url::join step for object store roots, which is what the test in the PR covers: s3://bucket/table/ plus that path comes out as s3://bucket/other-table/f.parquet on main.

So yeah it needs a crafted log. If that's outside what you all defend against that's fair, I'm just flagging that nothing bounds where the path lands right now.

hntd187:
Thanks, I don’t think it’s necessarily to guard against this. No normal user is intentionally creating bad or corrupt add paths. 

We can revisit this when a user actually encounters it.

PR #4635, opened 2026-08-06
fix: speed up full vacuum listing on multi-level partitions (#4634)

### LINKED ISSUE #4634: [Bug]: Vacuum is slow on scanning datasets with small file problem
### What happened?

Given a partitioned dataset with small files problem (millions or tens of millions of files), scanning for files when trying to execute vacuum, takes lots of time because of storage pagination mechanism, which is sequential. 

### Expected behavior

Use more parallelism based on defined partitioning columns for better performance.

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

Given a partitioned dataset with millions of files (ie: 50 million files), the file scanning takes lots of time because of storage pagination which is sequential. In the case of these partitioned datasets, for the ones partitioned by more than one column we have some options. On AWS the max number of URIs returned per page is 1000 thus leading to tens of thousands of pages listed one after another. 

In the case of a partitioned dataset, where the number of partitioning columns is `n > 1`, we can use prefixes and delimiters to achieve a greater parallelism when scanning the files in the dataset. 

Given a `50_000_000` files dataset partitioned like `/a=[1..5_000]/b=[1..10_000]` we could parallelise by first partitioning column. If we choose a pool of `256` threads,  this will lead to about **200x** better performance when retrieving pages in parallel. 

### Rea Life Example

I have multiple production dataset with tens of millions of files spread in two Hive partition columns. I'll give the one of my findings on such a dataset: first partition has a cardinality of approx `5_000` and the second partition columns has a cardinality of approx `10_000`. Current approach does listing sequentially the `5_000 * 10_000` object, page by page with `1_000` objects per page and is very slow.

### Relevant logs

```shell

```

andrei-ionescu:
# Description

Full-mode vacuum used a single sequential object-store list over the table root. On large multi-level partitioned tables that is correct but serial, so wall time is dominated by `LIST` round-trips.

This change switches multi-level tables (`n > 1` partition columns) to a parallel listings:

- expand only to depth `n-1` with `list_with_delimiter`
- list each resulting prefix in parallel
- leave the last partition column unexpanded so files under it are returned in bulk (avoids millions of tiny per-leaf `LIST` calls when the final column has high fan-out)

Single-level and unpartitioned tables keep the flat `list(None)` path, which is cheaper when expanding the only column would create many sparse leaves.

`LIST` concurrency defaults to 10 and can be overridden with `DELTARS_VACUUM_LIST_CONCURRENCY` but from my tests `10` gives some of the best results. 

# Benchmarks

### Summary

In short, this approach is 🚀 **`8 times`** 🚀 faster.

Here is a quick comparison between the old way of scanning and the parallel way in regards to duration:

|                       | fastest | slowest | median  | mean    | samples | iters |
| --------------------- | ------- | ------- | ------- | ------- | ------- | ----- |
| dry_run_flat (original)     | 17.88 m | 19.6 m  | 18.74 m | 18.74 m | 2       | 2     |
| dry_run_parallel (new) | 2.256 m | 2.297 m | 2.277 m | 2.277 m | 2       | 2     |

### By store latency

Here is a comparison with different storage latencies:

| store latency | dry_run_parallel | dry_run_flat | times faster | time reduction |
|---------------|----------|-----------------|--------------|----------------|
| 1 s           | 13.45 m  | 1.79 h          | 7.98 x       | 87.47 %        |
| 500 ms        | 7.938 m  | 59.01 m         | 7.43 x       | 86.54 %        |
| 300 ms        | 5.479 m  | 39.4 m          | 7.19 x       | 86.09 %        |
| 200 ms        | 4.23 m   | 30.23 m         | 7.15 x       | 86.01 %        |
| 100 ms        | 2.35 m   | 19.91 m         | 8.47 x       | 88.19 %        |
| 50 ms         | 2.099 m  | 13.54 m         | 6.45 x       | 84.50 %        |
| 10 ms         | 2.073 m  | 8.706 m         | 4.20 x       | 76.19 %        |
| 1 ms          | 2.107 m  | 7.75 m          | 3.68 x       | 72.81 %        |

### CPU load

Here is another comparison between the two modes this time in terms of CPU load:

|      | dry_run_flat | dry_run_parallel |
|--------|--------|--------|
| Wall time | ~1181 s (~19.7 min) | ~149 s (~2.5 min) |
| Total CPU time (all threads summed) | ~403 s (~6.7 min) | ~546 s (~9.1 min) |
| Avg CPU load | ~0.34 cores busy | ~3.67 cores busy |
| Avg as % of one core | ~34% | ~367% |
| Equivalent machine utilization (if 32 cores) | ~1% of all cores | ~12% of all cores |
| Threads | 171 | 227 |
| CPU time | ~403 s | ~546 s |
| Avg cores busy | 0.34 | 3.69 |
| Median | 0.33 | 3.85 |
| p95 | 0.37 | 4.81 |
| Max | 1.55 | 6.42 |
| Seconds with ≥ 0.5 cores | 1.3% | 99.3% |
| Seconds with ≥ 1.0 cores | 0.8% | 96.6% |
| Seconds with ≥ 2.0 cores | 0% | 89.9% |
| Seconds with ≥ 4.0 cores | 0% | 22.3% |

### Details

The fixture (aka the big 5.5 million partition dataset) was created once and multiple dry-run vacuums were executed on top of it to gather some valuable numbers.

I did use this command to generate it and it took at least `1h`:

```bash
cargo run --release -p delta-benchmarks -- generate-vacuum-fixture \
  --out ./crates/benchmarks/data/vacuum_bench/d1095_g5001 \
  --days 1095 \
  --groups 5001 \
  --force
```

Then I used the following command with a synthetic latency of `100ms` trying to mimic an object store.

```bash
cargo bench -p delta-benchmarks --bench vacuum -- \
--fixture ./crates/benchmarks/data/vacuum_bench/d1095_g5001 \
--list-latency-ms 100 \
--sample-count 2
```

# Real Life Example

I have multiple production dataset with tens of millions of files spread in two Hive partition columns. I'll give the one of my findings on such a dataset: first partition has a cardinality of approx `5000` and the second partition columns has a cardinality of approx `10000`. Current approach does listing sequentially the `5000 * 10000` object, page by page with `1000` objects per page and is very slow. 

# Related Issue(s)

- closes #4634

# Documentation

### DIFF
--- a/crates/benchmarks/src/vacuum.rs
+++ b/crates/benchmarks/src/vacuum.rs
@@ -229,7 +229,7 @@ pub async fn run_vacuum_full_dry_run(
         builder = builder.with_scan_concurrency(n);
     }
     if mode == VacuumScanMode::Flat {
-        builder = builder.disable_parallel_scan();
+        builder = builder.parallel_scan(false);
     }
 
     builder.await
--- a/crates/benchmarks/src/vacuum.rs
+++ b/crates/benchmarks/src/vacuum.rs
@@ -1,0 +1,338 @@
+//! Full-mode vacuum listing benchmarks.
+//!
+//! Fixture strategy: generate a multi-level partitioned Delta table **once**,
+//! then dry-run full vacuum many times (parallel vs flat). Dry-run does not
+//! delete, so the same table can be reused across samples.
+
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use chrono::Duration as ChronoDuration;
+use deltalake_core::arrow;
+use deltalake_core::kernel::{DataType, PrimitiveType, StructField, StructType};
+use deltalake_core::operations::vacuum::{VacuumMetrics, VacuumMode};
+use deltalake_core::protocol::SaveMode;
+use deltalake_core::{DeltaResult, DeltaTable, DeltaTableBuilder, DeltaTableError};
+use object_store::local::LocalFileSystem;
+use url::Url;
+
+use crate::latency_store::LatencyStore;
+
+/// Parameters for the multi-level vacuum fixture.
+///
+/// Layout: partitions `date` (string YYYY-MM-DD) and `group` (string).
+#[derive(Debug, Clone, Copy)]
+pub struct VacuumFixtureParams {
+    /// Number of distinct `date` partition values (one commit per day).
+    pub days: usize,
+    /// Number of distinct `group` values per day (`0..groups`).
+    pub groups: usize,
+    /// Rows written per (date, group) partition.
+    pub rows_per_partition: usize,
+}
+
+impl Default for VacuumFixtureParams {
+    fn default() -> Self {
+        // ~30 * 500 = 15k leaf partitions — enough to exercise listing without
+        // multi-hour generation.
+        Self {
+            days: 30,
+            groups: 500,
+            rows_per_partition: 1,
+        }
+    }
+}
+
+/// Whether full-mode vacuum uses the parallel multi-level scan.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VacuumScanMode {
+    /// Default: parallel prefix expansion when the table has >1 partition cols.
+    Parallel,
+    /// Force flat `list(None)` via [`VacuumBuilder::disable_parallel_scan`].
+    Flat,
+}
+
+impl VacuumScanMode {
+    pub fn name(self) -> &'static str {
+        match self {
+            Self::Parallel => "parallel",
+            Self::Flat => "flat",
+        }
+    }
+}
+
+impl std::fmt::Display for VacuumScanMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.name())
+    }
+}
+
+/// Marker file written next to a generated fixture for resume/skip.
+const FIXTURE_META: &str = "vacuum_fixture_meta.txt";
+
+/// True if `dir` already looks like a complete vacuum fixture.
+pub fn fixture_exists(dir: &Path) -> bool {
+    dir.join(FIXTURE_META).is_file() && dir.join("_delta_log").is_dir()
+}
+
+/// Generate (or skip if present) a multi-level partitioned table at `dir`.
+///
+/// This is partitioned by `date` and `group`.
+/// One commit per day containing all groups for that day.
+///
+/// Depending on the size of dataset it may take a long time (hours).
+/// For example: for 1095 days (`date`) and 5001 groups (`group`)
+/// it will take 1 to 2 hours depending on the machine it's executed.
+pub async fn generate_vacuum_fixture(
+    dir: &Path,
+    params: &VacuumFixtureParams,
+    force: bool,
+) -> DeltaResult<PathBuf> {
+    if !force && fixture_exists(dir) {
+        return Ok(fs::canonicalize(dir).unwrap_or_else(|_| dir.to_path_buf()));
+    }
+
+    if dir.exists() {
+        fs::remove_dir_all(dir).map_err(|e| {
+            DeltaTableError::generic(format!(
+                "failed to clear fixture dir {}: {e}",
+                dir.display()
+            ))
+        })?;
+    }
+    fs::create_dir_all(dir).map_err(|e| {
+        DeltaTableError::generic(format!(
+            "failed to create fixture dir {}: {e}",
+            dir.display()
+        ))
+    })?;
+
+    // Url::from_directory_path requires an absolute path; --out is often relative.
+    let dir = fs::canonicalize(dir)
+        .map_err(|e| DeltaTableError::generic(format!("canonicalize {}: {e}", dir.display())))?;
+
+    let table_url = Url::from_directory_path(&dir)
+        .map_err(|_| DeltaTableError::generic(format!("invalid table path: {}", dir.display())))?;
+
+    let schema = StructType::try_new(vec![
+        StructField::new("date", DataType::Primitive(PrimitiveType::String), false),
+        StructField::new("group", DataType::Primitive(PrimitiveType::String), false),
+        StructField::new("id", DataType::Primitive(PrimitiveType::Long), false),
+        StructField::new("value", DataType::Primitive(PrimitiveType::String), true),
+    ])?;
+
+    let mut table = DeltaTable::try_from_url(table_url)
+        .await?
+        .create()
+        .with_columns(schema.fields().cloned())
+        .with_partition_columns(["date", "group"])
+        .await?;
+
+    let start_day = days_ago_ymd(params.days.saturating_sub(1) as i64);
+    let mut global_id: i64 = 0;
+
+    for day_idx in 0..params.days {
+        let date = add_days(&start_day, day_idx as i64);
+        let batch = day_batch(
+            &date,
+            params.groups,
+            params.rows_per_partition,
+            &mut global_id,
+        )?;
+        table = table
+            .write(vec![batch])
+            .with_save_mode(SaveMode::Append)
+            .with_partition_columns(["date", "group"])
+            .await?;
+
+        if day_idx == 0 || day_idx + 1 == params.days || (day_idx + 1) % 10 == 0 {
+            eprintln!(
+                "vacuum fixture: committed day {}/{} ({date})",
+                day_idx + 1,
+                params.days
+            );
+        }
+    }
+
+    let meta = format!(
+        "days={}\ngroups={}\nrows_per_partition={}\npartitions_approx={}\n",
+        params.days,
+        params.groups,
+        params.rows_per_partition,
+        params.days.saturating_mul(params.groups)
+    );
+    fs::write(dir.join(FIXTURE_META), meta)
+        .map_err(|e| DeltaTableError::generic(format!("failed to write fixture meta: {e}")))?;
+
+    Ok(dir)
+}
+
+/// Open a previously generated fixture from the local filesystem.
+pub async fn open_vacuum_fixture(dir: &Path) -> DeltaResult<DeltaTable> {
+    open_vacuum_fixture_with_list_latency(dir, Duration::ZERO).await
+}
+
+/// Open a fixture, optionally wrapping the store so every LIST sleeps
+/// `list_latency` (simulates cloud LIST RTT on local disk).
+pub async fn open_vacuum_fixture_with_list_latency(
+    dir: &Path,
+    list_latency: Duration,
+) -> DeltaResult<DeltaTable> {
+    if !fixture_exists(dir) {
+        return Err(DeltaTableError::generic(format!(
+            "vacuum fixture not found at {} — run generate first",
+            dir.display()
+        )));
+    }
+    let dir = fs::canonicalize(dir)
+        .map_err(|e| DeltaTableError::generic(format!("canonicalize {}: {e}", dir.display())))?;
+    let table_url = Url::from_directory_path(&dir)
+        .map_err(|_| DeltaTableError::generic(format!("invalid table path: {}", dir.display())))?;
+
+    if list_latency.is_zero() {
+        return DeltaTableBuilder::from_url(table_url)?.load().await;
+    }
+
+    // Root-level local FS (not prefix-stripped). Table location in the URL
+    // selects the table path; LatencyStore only delays LIST-family calls.
+    let store = Arc::new(LatencyStore::new(
+        Arc::new(LocalFileSystem::new()),
+        list_latency,
+    ));
+    DeltaTableBuilder::from_url(table_url.clone())?
+        .with_storage_backend(store, table_url)
+        .load()
+        .await
+}
+
+/// Dry-run full vacuum once
+///
+/// Note: Clones `table` so the shared fixture can be reused across iterations.
+pub async fn run_vacuum_full_dry_run(
+    table: &DeltaTable,
+    mode: VacuumScanMode,
+    scan_concurrency: Option<usize>,
+) -> DeltaResult<(DeltaTable, VacuumMetrics)> {
+    let mut builder = table
+        .clone()
+        .vacuum()
+        .with_mode(VacuumMode::Full)
+        .with_dry_run(true)
+        .with_enforce_retention_duration(false)
+        // Listing-cost bench: retention 0. Live add files stay referenced so the
+        // delete set should be empty on a clean fixture (no planted orphans).
+        .with_retention_period(ChronoDuration::seconds(0));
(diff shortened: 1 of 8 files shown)


hntd187:
Can you describe a situation where this actually happens or at least describe the latency you experienced prior to this change and what you experience now? Also, any tests you want to add?

andrei-ionescu:
@hntd187 I described the problem in the linked ticket. If more than that is needed, or if I should copy text from there into this PR I can do it.

hntd187:
You did not describe how it actually happens, I'm trying to understand if this is a problem users actually encounter in real practice. What is the series of events that leads to a table with 50 million individual small files to read?

andrei-ionescu:
@hntd187 It did happen to me. I have a dataset  — production dataset — with tens of millions of files spread in 2 partition columns.

andrei-ionescu:
@hntd187 I added a "Real Life Example" chapter trying to explain the scale and the datasets that I work with.

hntd187:
Again, I am not asking what you currently have at the end I understand that. I am asking how you get to that point. Do you really create a table with 10s of millions of files and never run an optimize or vacuum through that process?

andrei-ionescu:
@hntd187 Short answer: high cardinality on partitions + tremendous number of writers. The problem is that optimise does not compact cross partitions. So, even with optimise, which we do run, constantly, since I have `10K * 5K = 50 million` partitions we still have 50 millions files. 

I'm sorry I didn't understand your questions and maybe I didn't answer as expected.

hntd187:
Not a problem, I don't have a test table to try something like this, have you ran this on any of the tables you have to get a better idea of how much it really helps? I'm sure it'll big a good boost, I'm just curious just how much it would be.

andrei-ionescu:
I'll come with some number soon.

rtyler:
> Again, I am not asking what you currently have at the end I understand that. I am asking how you get to that point. Do you really create a table with 10s of millions of files and never run an optimize or vacuum through that process?

It is probably helpful for our working relationship that @hntd187 has never seen some of the tables I have produced :laughing: 

Anywhoo

@andrei-ionescu I have been reviewing the code but GitHub is being GitHub right now so I cannot leave line-by-line comments. Based on the stated goal that you have, there is a very high likelihood that rate limits by the object store are going to cause premature exits to the vacuum operation. Additionally, there is some `Vec` `collect()` operations which are likely to lead to a lot of memory use as the prefixes are scanned before operations are started.

For the first one, there is a configurable [RetryConfig](https://docs.rs/object_store/latest/object_store/struct.RetryConfig.html) which does get used in some other places. I have had to make sure that it is always set in my environments but it's not very intuitive (IMHO). I briefly considered maybe [LimitStore](https://docs.rs/object_store/latest/object_store/limit/struct.LimitStore.html) would be a good thing to wrap in here, but i'm not sure if that's more trouble than its worth.

Since it appears that you're using a LLM to help drive this code, I would encourage restructuring the interaction between the newly added functions to return streams rather than `Vec` of data for vacuuming, that will help keep memory consumption down for larger operations.  With Devstral I got decent results asking

> The recent change issues a `collect` and for large file listing  will consume a lot of memory. Consider the options and propose a  change to make it operate on a stream rather than Vecs

andrei-ionescu:
Thank you @rtyler for your feedback. It is very valuable. I'll implement them as you suggested. Greatly appreciated and thanks for bearing with me.

Regarding @hntd187's questions... because we are in fact a platform, our customers are the ones that are creating these tables plus all kinds of other tables, and these small file problem datasets are occurring in a rate that sometimes it is bothering us, so we try to improve the data maintenance processes to be as fast as possible. I'll be able to get some numbers next week.

andrei-ionescu:
Hello @rtyler and @hntd187 and thank you for your valuable feedback. It really did help me have a better solution and better numbers.

I did a bit of a refactor to switch from Vecs and collects to something better and I think it did pay off. 

Regarding `RetryConfig` I'm basing it on the default RetryConfig set on the whole object store and now with a lower limit of parallelisation it should be good. 

Please have a look and let me know what you think.

Greatly appreciated 🙏

andrei-ionescu:
Here is another full output of a new run of the vacuum benchmark:

```bash
cargo bench -p delta-benchmarks --bench vacuum -- \
--fixture ~/github/delta-rs/crates/benchmarks/data/vacuum_bench/d1095_g5001 \
--list-latency-ms 100 \
--sample-count 2
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.71s
     Running benches/vacuum.rs (target/release/deps/vacuum-27a59b6d6222321d)
Timer precision: 41 ns
vacuum                    fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ full_dry_run_flat      19.5 m        │ 20.13 m       │ 19.82 m       │ 19.82 m       │ 2       │ 2
│                         max alloc:    │               │               │               │         │
│                           4050503     │ 4050344       │ 4050423       │ 4050423       │         │
│                           59.33 KB    │ 53.63 KB      │ 56.48 KB      │ 56.48 KB      │         │
│                         alloc:        │               │               │               │         │
│                           18108370    │ 18108153      │ 18108261      │ 18108261      │         │
│                           1.867 GB    │ 1.867 GB      │ 1.867 GB      │ 1.867 GB      │         │
│                         dealloc:      │               │               │               │         │
│                           30488381    │ 30488323      │ 30488352      │ 30488352      │         │
│                           4.452 GB    │ 4.452 GB      │ 4.452 GB      │ 4.452 GB      │         │
│                         grow:         │               │               │               │         │
│                           11015746    │ 11015735      │ 11015740      │ 11015740      │         │
│                           444.1 MB    │ 444.1 MB      │ 444.1 MB      │ 444.1 MB      │         │
│                         shrink:       │               │               │               │         │
│                           11148       │ 11148         │ 11148         │ 11148         │         │
│                           89.18 KB    │ 89.18 KB      │ 89.18 KB      │ 89.18 KB      │         │
╰─ full_dry_run_parallel  2.306 m       │ 2.315 m       │ 2.311 m       │ 2.311 m       │ 2       │ 2
                          max alloc:    │               │               │               │         │
                            4050420     │ 4050412       │ 4050416       │ 4050416       │         │
                            53.63 KB    │ 53.63 KB      │ 53.63 KB      │ 53.63 KB      │         │
                          alloc:        │               │               │               │         │
                            18130136    │ 18130136      │ 18130136      │ 18130136      │         │
                            1.978 GB    │ 1.978 GB      │ 1.978 GB      │ 1.978 GB      │         │
                          dealloc:      │               │               │               │         │
                            30510217    │ 30510222      │ 30510219      │ 30510219      │         │
                            4.563 GB    │ 4.563 GB      │ 4.563 GB      │ 4.563 GB      │         │
                          grow:         │               │               │               │         │
                            11022273    │ 11022273      │ 11022273      │ 11022273      │         │
                            444.3 MB    │ 444.3 MB      │ 444.3 MB      │ 444.3 MB      │         │
                          shrink:       │               │               │               │         │
                            11148       │ 11148         │ 11148         │ 11148         │         │
                            89.18 KB    │ 89.18 KB      │ 89.18 KB      │ 89.18 KB      │         │
```

hntd187:
So if I understand the metrics correctly from your benchmarks it would appear the runtime was reduced from 19-20m to about 2 minutes with no real difference in any memory footprint, is that correct? If so that is fantastic to see, let me have sometime to look over the PR since I haven't reviewed since you changed things.

crates/benchmarks/src/vacuum.rs
```diff
@@ -0,0 +1,335 @@
...
+    Some(ymd_to_unix_days(y, m, d).saturating_mul(86_400))
+}
+
+fn format_utc_ymd(unix_secs: i64) -> String {
+    let days = unix_secs.div_euclid(86_400);
+    let (y, m, d) = unix_days_to_ymd(days);
+    format!("{y:04}-{m:02}-{d:02}")
+}
+
+// Civil date conversion (Howard Hinnant algorithms), UTC day number since 1970-01-01.
```
hntd187: Is there such a thing as uncivil date conversion?

crates/core/src/operations/vacuum.rs
```diff
@@ -177,6 +210,34 @@ impl VacuumBuilder {
...
+        self
+    }
+
+    /// Disable the parallel multi-level partition scan used by full-mode vacuum.
+    ///
+    /// By default, tables with more than one partition column use hierarchical
+    /// prefix expansion and concurrent leaf LIST calls. Calling this forces the
+    /// flat `list(None)` scan path instead (useful for comparison, debugging, or
+    /// working around store rate limits).
+    pub fn disable_parallel_scan(mut self) -> Self {
```
hntd187: I'd prefer this instead be `with_parallel_scan(mut self, parallel_scan: bool) -> Self`

crates/core/src/operations/vacuum.rs
```diff
@@ -336,61 +397,117 @@ impl VacuumBuilder {
...
+                    })
+                    .try_flatten_unordered(scan_concurrency)
+                    .try_for_each(|(path, size)| {
+                        files_to_delete.push(path);
+                        file_sizes.push(size);
+                        ready(Ok::<_, DeltaTableError>(()))
+                    })
+                    .await?;
+
+                    for (path, size) in intermediate_orphans.lock().unwrap().drain(..) {
```
hntd187: deal with this error instead of unwrap

crates/core/src/operations/vacuum.rs
```diff
@@ -605,6 +722,222 @@ fn ok_to_delete(
...
+    debug!(
+        "The file {:?} is an untracked stale orphan and will be vacuumed in full mode",
+        &obj_meta.location
+    );
+    Ok(Some((obj_meta.location.clone(), obj_meta.size as i64)))
+}
+
+/// Root prefixes that should not be expanded during full-mode partition walks.
+fn is_skippable_root_prefix(path: &Path) -> bool {
+    let path_name = path.to_string();
```
hntd187: Being pedantic, but you probably don't need to reallocate this as a String would using `.as_str()` or something similar be better?

crates/benchmarks/src/vacuum.rs
andrei-ionescu: The function is self explanatory... I removed the comment since the other functions do not have comments either.

andrei-ionescu:
Another set of vacuum benchmark run, 3 samples/iterations this time:

```bash
cargo bench -p delta-benchmarks --bench vacuum -- \
--fixture ~/github/delta-rs/crates/benchmarks/data/vacuum_bench/d1095_g5001 \
--list-latency-ms 100 \
--sample-count 3
   Compiling delta-benchmarks v0.0.1 (/Users/aionescu/github/delta-rs/crates/benchmarks)
    Finished `bench` profile [optimized + debuginfo] target(s) in 31.41s
     Running benches/vacuum.rs (target/release/deps/vacuum-27a59b6d6222321d)
Timer precision: 41 ns
vacuum                    fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ full_dry_run_flat      19.58 m       │ 20.05 m       │ 19.82 m       │ 19.82 m       │ 3       │ 3
│                         max alloc:    │               │               │               │         │
│                           4050504     │ 4050343       │ 4050349       │ 4050398       │         │
│                           59.33 KB    │ 53.63 KB      │ 53.63 KB      │ 55.53 KB      │         │
│                         alloc:        │               │               │               │         │
│                           18108370    │ 18108153      │ 18108153      │ 18108225      │         │
│                           1.867 GB    │ 1.867 GB      │ 1.867 GB      │ 1.867 GB      │         │
│                         dealloc:      │               │               │               │         │
│                           30488381    │ 30488325      │ 30488319      │ 30488341      │         │
│                           4.452 GB    │ 4.452 GB      │ 4.452 GB      │ 4.452 GB      │         │
│                         grow:         │               │               │               │         │
│                           11015746    │ 11015735      │ 11015735      │ 11015738      │         │
│                           444.1 MB    │ 444.1 MB      │ 444.1 MB      │ 444.1 MB      │         │
│                         shrink:       │               │               │               │         │
│                           11148       │ 11148         │ 11148         │ 11148         │         │
│                           89.18 KB    │ 89.18 KB      │ 89.18 KB      │ 89.18 KB      │         │
╰─ full_dry_run_parallel  2.319 m       │ 2.376 m       │ 2.372 m       │ 2.356 m       │ 3       │ 3
                          max alloc:    │               │               │               │         │
                            4050420     │ 4050411       │ 4050444       │ 4050425       │         │
                            53.63 KB    │ 53.63 KB      │ 53.63 KB      │ 53.63 KB      │         │
                          alloc:        │               │               │               │         │
                            18130136    │ 18130136      │ 18130136      │ 18130136      │         │
                            1.978 GB    │ 1.978 GB      │ 1.978 GB      │ 1.978 GB      │         │
                          dealloc:      │               │               │               │         │
                            30510219    │ 30510222      │ 30510187      │ 30510209      │         │
                            4.563 GB    │ 4.563 GB      │ 4.563 GB      │ 4.563 GB      │         │
                          grow:         │               │               │               │         │
                            11022273    │ 11022273      │ 11022273      │ 11022273      │         │
                            444.3 MB    │ 444.3 MB      │ 444.3 MB      │ 444.3 MB      │         │
                          shrink:       │               │               │               │         │
                            11148       │ 11148         │ 11148         │ 11148         │         │
                            89.18 KB    │ 89.18 KB      │ 89.18 KB      │ 89.18 KB      │         │
```

I'll try also to measure the processor load during these benchmarks.

andrei-ionescu:
@hntd187, I added another table with CPU load comparison between the vacuum modes in the PR description.

hntd187 (dismissed):
few changes I'd like

crates/core/src/operations/vacuum.rs
andrei-ionescu: Implemented

crates/core/src/operations/vacuum.rs
andrei-ionescu: Switched the approach a little bit to support these errors

crates/core/src/operations/vacuum.rs
andrei-ionescu: Implemented

andrei-ionescu:
@hntd187 Your suggestions were very valuable. I implemented all of them in a way or another. Thanks for the feedback. I re-requested your review.

andrei-ionescu:
Added [By store latency](#by-store-latency) section with metrics on different storage latencies.

@hntd187, @rtyler could you have another round of review?

Greatly appreciated!

andrei-ionescu:
@hntd187, @rtyler, @roeap Can you have another look on this one? Thanks.

rtyler (approved):

andrei-ionescu:
@rtyler just for my curiosity, what does the `ai` label mean and in what cases is it added? 

If is due to the branch name that I did give — `ai.parallel_vacuum` — the `ai` prefix comes from my name: `A`ndrei `I`onescu, so `ai`. 😄 

Thanks

PR #4639, opened 2026-08-09
feat: deletion vector support in cdf

hntd187:
# Description
This PR adds support for deletion vectors inside CDF, resolving proper feeds if deletion vectors have marked specific rows for deletion. 

I'm opening this PR in draft to start because I'd like feedback on this early and I don't particularly like parts of the implementation. My extend pairs portion which deals with when a version has a remove followed immediately by an add for a single file path. This single edge case path adds a lot of extra code to resolve them and there are places in scan_utils where I have "kinda the same, but slightly" different shapes of code that does the exact same thing, just depends on what data I at the time to accomplish it. I'd actually like to reduce this PR down, I think my code (and some AI suggestions) are a bit too verbose. 

# Performance
This PR uses kernel as much as it can in some places so we don't actually do any DV parsing ourselves and we don't actually read any DVs ourselves. We defer all of that to kernel, but the reason I mention this under performance is because while this is the only aspect that uses kernel, I think moving it out of kernel would ultimately make this faster. 

For very small simple tables performance is about on par with kernel from memory usage and time perspective. Ours is slightly faster here, but it's not something a human would ever notice. Now, for larger tables, kernel REALLY starts to drag on cdf reads. I have observed 10gb local tables taking upwards of 20 minutes to read in kernel while consuming approximately 60gb of memory. On the flip side ours does this same read in about 14 seconds while consuming about 15gb of memory at the peak, but 9-10gb through most of the read. I am speculating, but the memory spike might be because of using kernel to do the DV parsing and reading internally and could likely be improved if we did this ourselves.


ion-elgreco:
Is kernel's declarative plans perhaps already available on their cdf reads?

hntd187:
They are not. They're still a work in progress.

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -251,6 +313,163 @@ pub fn prune_specs_by_partition<F: FileAction>(
...
+        let tree_map = read_dv_treemap(Some(dv), &engine, &table_root_url)?;
+        return Ok(access_plan_for_selection(
+            tree_map,
+            &file_action.path(),
+            file_action.size()? as u64,
+            cache,
+            metrics,
+        )
+        .await
+        .ok());
```
ethan-tyler: I believe this converts all metadata read errors into `none` which the caller interprets as the file having no access plan and scans the complete file. I worry a transient read failure or malformed path can reintroduce dv rows into the cdc result.

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -251,6 +313,163 @@ pub fn prune_specs_by_partition<F: FileAction>(
...
+) -> DeltaResult<Option<ParquetAccessPlan>> {
+    if let Some(dv) = file_action.deletion_vector() {
+        let tree_map = read_dv_treemap(Some(dv), &engine, &table_root_url)?;
+        return Ok(access_plan_for_selection(
+            tree_map,
+            &file_action.path(),
+            file_action.size()? as u64,
+            cache,
+            metrics,
+        )
```
ethan-tyler: `access_plan_for_selection()` is treating marked row numbers as rows to include but they are marked from tree_map which are the file's DVs?

crates/core/src/operations/load_cdf.rs
```diff
@@ -375,7 +386,7 @@ impl CdfLoadBuilder {
             remove_files = prune_specs_by_partition(remove_files, partition_pruning)?;
         }
 
-        Ok((change_files, add_files, remove_files))
+        Ok((change_files, add_files, remove_files, resolved_pairs))
```
ethan-tyler, line 389: `resolved_pairs` bypasses `prune_specs_by_partition()`. Should we be pruning pairs before `extend_groups_with_pairs()`?

hntd187: good point, I think I missed this

crates/core/src/delta_datafusion/cdf/mod.rs
```diff
@@ -40,12 +40,19 @@ pub(crate) static ADD_PARTITION_SCHEMA: LazyLock<Vec<Field>> = LazyLock::new(||
         ),
     ]
 });
+#[derive(Debug)]
+pub(crate) struct ResolvedPair {
```
rtyler, line 49: I think I understand the need for this struct as a convenience to carry mappings of DV descriptors to the corresponding file `Add` action. That's an assumption which I hope is right 😄 

Please add some rustdoc that explains what this struct  is for and the members, that will help IDE/editor integrations explain what this is when it pops up in the function calls below

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -94,35 +104,87 @@ pub fn create_spec_partition_values<F: FileAction>(
...
+            .map(|part| map_action_to_scalar(&pair.add, part, schema.clone()))
+            .collect::<DeltaResult<Vec<ScalarValue>>>()?;
+
+        push_pair_selection(
+            deletes,
+            "delete",
+            remove_groups,
+            &pair,
+            &table_partition_values,
+            Arc::clone(&cache),
```
rtyler, line 137: Line 128 above uses `.clone()` and these calls to `push_pair_selection` use `Arc::clone` which as far as I can tell is functionally identical in how it's being used here.

I don't _think_ this matters, it just struck me as bizarre.

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -251,6 +313,163 @@ pub fn prune_specs_by_partition<F: FileAction>(
...
+            storage_type,
+            self.path_or_inline_dv,
+            self.offset,
+            self.size_in_bytes,
+            self.cardinality,
+        )?)
+    }
+}
+
+fn row_group_access(
```
rtyler, line 343: I have the same feedback as above about rustdoc for these newly added utility functions. I'm sure _you_ know what they do and why they exist, but as a filthy casual I have to re-read undocumented functions every time I encounter them to understand why they exist 😄 


Many of these utility functions would benefit from unit tests as well. Right now there is one "integration" test here in `test_reading_cdf_with_dv` which will make many of these utilities brittle to maintenance over time.

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -251,6 +313,163 @@ pub fn prune_specs_by_partition<F: FileAction>(
...
+
+    if marked.peek().is_none_or(|&idx| idx >= rg_end) {
+        return if keep_marked {
+            RowGroupAccess::Skip
+        } else {
+            RowGroupAccess::Scan
+        };
+    }
+
+    let mut selectors = Vec::new();
```
rtyler, line 350: Elsewhere this change introduces `Vec::with_capacity` presumably as a hand-wavy optimization. In this function isn't the potential size of `selectors` known by `rg_end`?

crates/core/src/delta_datafusion/cdf/scan_utils.rs
hntd187: I believe so, I can't tell if this is an issue or if you are just asking a clarifying question

crates/core/src/delta_datafusion/cdf/scan_utils.rs
hntd187: Good point, let me resolve that

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -251,6 +313,163 @@ pub fn prune_specs_by_partition<F: FileAction>(
...
+    keep_marked: bool,
+) -> RowGroupAccess {
+    let rg_end = rg_start + num_rows as u64;
+
+    if marked.peek().is_none_or(|&idx| idx >= rg_end) {
+        return if keep_marked {
+            RowGroupAccess::Skip
+        } else {
+            RowGroupAccess::Scan
+        };
```
rtyler: This `return` block syntax feels _very_ Ruby, take that how you will 😆 😈 

You don't have to change this if you don't want to, but I found reading this code I had to work harder to piece together the conditional logic for when Skip vs. Scan would happen.with the `is_none_or` here. This may be a case where the logic needs some inline comments to help explain

crates/core/src/delta_datafusion/cdf/scan_utils.rs
```diff
@@ -251,6 +313,163 @@ pub fn prune_specs_by_partition<F: FileAction>(
...
+
+fn read_dv_treemap(
+    dv: Option<crate::kernel::DeletionVectorDescriptor>,
+    engine: &Arc<dyn Engine>,
+    table_root_url: &Url,
+) -> DeltaResult<roaring::RoaringTreemap> {
+    match dv {
+        Some(dv) => {
+            let kernel_dv: DeletionVectorDescriptor = dv.try_into()?;
+            Ok(kernel_dv.read(engine.storage_handler(), table_root_url)?)
```
rtyler, line 391: Ah yay, a kernel I/O call that's totally blocking and synchronous 🤗 😬

crates/core/src/operations/load_cdf.rs
```diff
@@ -180,6 +183,7 @@ impl CdfLoadBuilder {
         Vec<CdcDataSpec<AddCDCFile>>,
         Vec<CdcDataSpec<Add>>,
         Vec<CdcDataSpec<Remove>>,
+        Vec<ResolvedPair>,
```
rtyler, line 186: Girl do we love a giant function in the Datafusion code 😬

crates/core/src/operations/load_cdf.rs
```diff
@@ -190,15 +194,16 @@ impl CdfLoadBuilder {
             self.calculate_earliest_version(snapshot).await?
         };
 
-        let mut change_files: Vec<CdcDataSpec<AddCDCFile>> = vec![];
```
rtyler, line 193: I have really mixed feelings about pre-allocation of Vecs like this with an arbitrary size. In this case, these variables are being potentially being returned early in a number of cases without actually containing anything.

I think the preallocation works well in other places where we know we have N entries ahead of time that the `Vec` **must** grow to support, but in this function there are a lot of ways in which these could be allocation and then never used, e.g. `cdc_actions.empty()`, `remove_actions.empty()`, etc.

crates/core/src/operations/load_cdf.rs
```diff
@@ -270,8 +275,9 @@ impl CdfLoadBuilder {
             let version_actions: Vec<Action> = get_actions(version, &snapshot_bytes?)?;
 
             let mut ts = 0;
-            let mut cdc_actions = vec![];
-
+            let mut cdc_actions = Vec::with_capacity(1024);
+            let mut add_actions = Vec::with_capacity(1024);
+            let mut remove_actions = Vec::with_capacity(1024);
             if self.starting_timestamp.is_some() || self.ending_timestamp.is_some() {
```
rtyler: 🤪

crates/core/src/operations/load_cdf.rs
hntd187: I looked into this and resolved pairs don't actually need to go through prune specs, they are created only to augment existing add/remove groups and then be discarded. The add/remove pairing stuff is very much bolted on to the existing flow, but it's somehow not the worst it could have been (or at least where I started)

crates/core/src/delta_datafusion/cdf/scan_utils.rs
hntd187: I've been trying to use `Arc::clone` instead of just straight `.clone()` for clarity sake looks like I missed one

crates/core/src/delta_datafusion/cdf/scan_utils.rs
hntd187: If it's any consolation, I've seen it account for less than 1% of runtime in most larger tables.

crates/core/src/operations/load_cdf.rs
hntd187: I think I did this in anticipation of kernel causing some allocation pressure so I wanted to try and allocate as much upfront, but I think these are probably not necessary now, at least not necessary as a performance step.

crates/core/src/operations/load_cdf.rs
hntd187: Same as above, I can back these out

crates/core/src/delta_datafusion/cdf/scan_utils.rs
hntd187: clippy actually complains about not lifting the return out so apparently this is the "correct" way to do this, but I also had it written the same way you had thought

crates/core/src/delta_datafusion/cdf/scan_utils.rs
hntd187: No, we don't know ahead of time what the deletion vector indexes will be for a given row group ahead of time. I suppose we could infer this, but I've just universally removed all the `with_capacity` if we ever run into allocation issues I can revisit.

crates/core/src/delta_datafusion/cdf/scan_utils.rs
ethan-tyler: My concern is that `create_file_scan_plan()` passes an action’s raw dv into the same helper while using `keep_marked = true`. For an unpaired dv - cdf should emit the rows that were logically live before removal, which should be physical rows excluding the raw dv. The current selection emits only the dv rows instead.

Unless `create_file_scan_plan()` is guaranteed never to receive an unpaired Add/Remove carrying a dv, I think the two paths need explicit  semantics- something like `IncludeMarked` and `ExcludeMarked`

hntd187: Correct, the paired actions are separated out prior to this ever being hit.

rtyler (approved):

PR #4650, opened 2026-08-18
chore: upgrade our build to 1.94.1 which is now required

rtyler:
The AWS crates bumped their MSRV and thus so do we -_-


hntd187 (approved):
I'm about it

PR #4658, opened 2026-08-21
fix: vacuum percent-encoded partition paths

### LINKED ISSUE #4646: [Bug]: vacuum reports success but deletes nothing, when a partition directory name is percent-escaped (regression in 1.5.0)
### What happened?

*The investigation and write-up below were done with Claude Code. I hit the bug on our production table, I've run the repro myself, and I can defend what's here.*

Thank you for delta-rs — Delta Lake is the storage layer for essentially our whole system, so your work is genuinely load-bearing for us.

**The symptom.** `vacuum(dry_run=False)` deletes nothing, returns the superseded file as deleted, and commits `VACUUM START` / `VACUUM END` with `numDeletedFiles` counting it.

It shows up whenever the partition *directory name* contains a character in object_store's `INVALID` set. Every character escaped when that name is built becomes a `%`, and `%` is itself in the set, so any escaped partition value triggers it — a `Timestamp` unavoidably, since the rendered value always contains a space and colons. A few unescaped characters trigger it directly: `"a~b"` gives `part=a~b`, with no `%` anywhere, and still fails. The column type is not the trigger — in the repro below a `String` partition fails on `"a b"` and succeeds on `"ab"`.

It works in 1.4.2 and fails from 1.5.0 onward — I ran the repro against each. The difference looks like #4227, which moved the lite path from a storage listing to building the delete list from tombstone paths. Still reproduces on `main` at `fd7e9691`, so it isn't covered by #4624.

These two things look like they combine:

**1. The `add` and `remove` paths reach object-store keys by different routes.** Both views decode first, so both start from the real on-disk directory name — `LogicalFileView::path()` and `TombstoneView::path()` (`kernel/snapshot/iterators/tombstones.rs:29`) each end in `percent_decode_str`. They then diverge. For an `add`, `object_store_path()` (`kernel/snapshot/iterators.rs:140`) calls `Path::parse`, which validates without re-encoding, so the key matches the file on disk. That is why reads work. For a `remove`, `operations/vacuum.rs:329` (and `:620`) calls `Path::from`, which does encode, and object_store's `INVALID` set contains `%`. The key lands back on the log's escaped form, where no object exists.

**2. A `NotFound` while deleting is counted as a deletion.** `operations/vacuum.rs:529` maps `Err(Error::NotFound { .. })` to `Ok(path)`, so the miss reaches the returned list and `numDeletedFiles`. This one predates 1.5.0 and was harmless until the first started producing misses.

To check that the encoding really is the cause rather than something else, I planted a decoy file at the double-encoded path: vacuum deleted the decoy and left the real superseded file.

On existing issues: this looks like the same inconsistent-path-encoding root cause as #4053 (closed by #4056), at a call site I don't think #4056 touched. It's easy to mistake for #3911, so to save you checking — the repro passes `dry_run=False`, and the `"ab"` control case does delete.

<details>
<summary>Also noticed while investigating: <code>full=True</code> behaviour</summary>

`full=True` still deletes the file, since it works from a storage listing. On a table with escaped partition names it reports each file twice, once per code path, and commits `numDeletedFiles: 2` for a single file. Its dedupe and recent-tombstone guards (`vacuum.rs:345`, `:365`) also compare listed paths against a set built by the same `Path::from` at `:620`, so as far as I can tell the protection added by #4300 doesn't apply there and a file with a recent tombstone falls through to the physical-age branch. Happy to split this out into its own issue if you'd rather keep them separate.

</details>

### Expected behavior

A `remove` action's path should resolve to the same object-store key an `add` action's path would.

Separately, a `NotFound` while deleting probably shouldn't count as a deleted file — reporting it as one means an operator believes space was reclaimed and time travel severed when neither happened.

On the shape of a fix, entirely your call: a change at the call site would do it, though `TombstoneView` has no `object_store_path()` of its own, so the two paths could drift apart again.

### Operating System

Linux

### Binding

Python

### Bindings Version

1.6.2

### Steps to reproduce

```python
import tempfile
from datetime import datetime
from pathlib import Path

import pyarrow as pa

# Only the third case avoids percent-escaping. The type is irrelevant: the
# String "a b" fails exactly as the Timestamp does.
for value in (datetime(2026, 8, 13), "a b", "ab"):
    table = Path(tempfile.mkdtemp()) / "t"
    literal = value.isoformat() if isinstance(value, datetime) else value
    for x in (1, 2):  # the second write supersedes the first in the same partition
        write_deltalake(
            table,
            pa.table({"part": [value], "x": [x]}),
            mode="overwrite",
            predicate=f"part = '{literal}'",
            partition_by=["part"],
        )

    part_dir = next(p for p in table.iterdir() if p.name.startswith("part="))
    before = len(list(part_dir.glob("*.parquet")))
    reported = DeltaTable(table).vacuum(
        retention_hours=0, dry_run=False, enforce_retention_duration=False
    )
    after = len(list(part_dir.glob("*.parquet")))

    print(f"{part_dir.name}")
    print(f"  vacuum reported deleting {len(reported)}, actually deleted {before - after}")
    print(f"  reported path: {reported[0]}")
```

### Relevant logs

```shell
part=2026-08-13%2000%3A00%3A00.000000
  vacuum reported deleting 1, actually deleted 0
  reported path: /tmp/tmp2pfefxfg/t/part=2026-08-13%252000%253A00%253A00.000000/part-00000-9c9bd228-….parquet
part=a%20b
  vacuum reported deleting 1, actually deleted 0
  reported path: /tmp/tmpdme8wrx9/t/part=a%2520b/part-00000-2ad7e9ae-….parquet
part=ab
  vacuum reported deleting 1, actually deleted 1
  reported path: part=ab/part-00000-559362dd-….parquet
```

The returned path is absolute in the two failing cases and relative in the working one — the absolute form is the `NotFound` error's path, which is the second point above showing through.

mameikagou:
# Description

Vacuum rebuilt decoded tombstone paths with `Path::from`, which re-encoded
percent-escaped partition names. It could therefore report a deletion while
leaving the original file in place.

This constructs tombstone object-store paths the same way as active files and
uses them in both lite and full vacuum modes. A local regression test verifies
that lite vacuum deletes a superseded file under an escaped partition path.

# Related Issue(s)

- Closes #4646

# Documentation

Not applicable.

# Testing

- Regression fails with the old vacuum path conversion and passes with this fix
- `cargo check -p deltalake-core --lib --no-default-features --features rustls`
- `git diff --check`

# AI assistance

OpenAI Codex helped draft the patch and tests. I reviewed the path semantics and
verified the regression and final diff.

### DIFF
--- a/crates/core/src/kernel/snapshot/iterators/tombstones.rs
+++ b/crates/core/src/kernel/snapshot/iterators/tombstones.rs
@@ -5,7 +5,7 @@ use arrow::{
     datatypes::Int64Type,
 };
 use delta_kernel::{actions::Remove, schema::ToSchema};
-use object_store::path::Path;
+use object_store::path::{Path, PathPart};
 use percent_encoding::percent_decode_str;
 
 use crate::kernel::snapshot::iterators::get_string_value;
@@ -38,10 +38,10 @@ impl TombstoneView {
     /// Returns an object store path using the same decoded representation as logical files.
     pub(crate) fn object_store_path(&self) -> Path {
         let path = self.path();
-        match Path::parse(path.as_ref()) {
-            Ok(path) => path,
-            Err(_) => Path::from(path.as_ref()),
-        }
+        path.as_ref()
+            .split('/')
+            .map(|segment| PathPart::parse(segment).unwrap_or_else(|_| PathPart::from(segment)))
+            .collect()
     }
 
     /// Returns the deletion timestamp (milliseconds since epoch), if recorded.
@@ -94,8 +94,7 @@ mod tests {
     use delta_kernel::scan::scan_row_schema;
     use std::sync::Arc;
 
-    #[test]
-    fn object_store_path_matches_logical_file_view() {
+    fn tombstone_view(raw_path: &str) -> TombstoneView {
         let kernel_schema = Remove::to_schema();
         let schema: Schema = (&kernel_schema).try_into_arrow().unwrap();
         let mut columns: Vec<ArrayRef> = schema
@@ -103,14 +102,19 @@ mod tests {
             .iter()
             .map(|field| new_null_array(field.data_type(), 1))
             .collect();
-        let raw_path = "part=a%20b/file.parquet";
         columns[schema.index_of("path").unwrap()] =
             Arc::new(StringArray::from(vec![Some(raw_path)]));
         columns[schema.index_of("dataChange").unwrap()] =
             Arc::new(BooleanArray::from(vec![Some(true)]));
 
         let batch = RecordBatch::try_new(Arc::new(schema), columns).unwrap();
-        let view = TombstoneView::new(batch, 0);
+        TombstoneView::new(batch, 0)
+    }
+
+    #[test]
+    fn object_store_path_matches_logical_file_view() {
+        let raw_path = "part=a%20b/file.parquet";
+        let view = tombstone_view(raw_path);
 
         assert_eq!(view.path(), "part=a b/file.parquet");
         let expected = Path::parse(view.path().as_ref()).unwrap();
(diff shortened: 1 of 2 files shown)


hntd187:
It seems inefficient to reparse every path? Shouldn’t this be resolved in path itself instead of reparsing every path?

mameikagou:
Good question. This does not add another conversion in the lite path: the old code built one `Path` per consumed tombstone with `Path::from`; the patch replaces that with one `Path::parse` so percent encoding is preserved. It also matches the existing `LogicalFileView::object_store_path()` boundary while keeping `path()` as the decoded logical Delta path.

Resolving in `TombstoneView::new` would make the view eager and parse unexpired tombstones that lite vacuum never consumes. There is one real duplicate in full mode: an expired tombstone is parsed while building the path sets and again when building the delete list. If that is the inefficiency you mean, I can carry the parsed `Path` forward there instead.

I also pushed the Rust 2024 import-order formatting fix in `b7c2c31`.

hntd187:
@mameikagou I asked you, not your AI, please respect our time and respond yourself. The error condition will parse the string twice, does it not?

mameikagou:
> @mameikagou I asked you, not your AI, please respect our time and respond yourself. The error condition will parse the string twice, does it not?

Yes，u are right，sorry about that . I will change this  PR.

mameikagou:
You were right about the error path doing a second pass over the string. I changed it to validate per segment and encode only the invalid segment, and full vacuum now carries the computed `Path` forward instead of rebuilding it.

I added coverage for mixed valid/invalid segments and preserving existing percent encoding. The focused tombstone tests and both lite/full vacuum regressions pass.

Issue #4584, opened 2026-07-08
Make Snapshot the table state authority and retire EagerSnapshot compatibility


ethan-tyler:
## Summary
Make `Snapshot` the single authority for interpreted table state at a table version.

Make `EagerSnapshot` a compatibility adapter. It may preserve existing Rust, Python, serde, transaction, and table-state call shapes while replacements land. Add new table-state capabilities to `Snapshot` or to the appropriate scan/planning module instead.

This issue is to track this work separately from the DeltaTableProvider/provider migration epic.

## Background

We currently have two snapshot concepts with overlapping responsibilities:

- `Snapshot`, which represents table version, schema, metadata, protocol, and lazy replay backed table state access.
- `EagerSnapshot`, which remains a materialized compatibility surface for older Rust, Python, serde, transaction, and table state paths.

That split has made the table-state model hard to reason about. Some code paths depend on eager implementation details such as `log_data()` and materialized active-file batches. Other code paths need lazy snapshots created with `require_files=false` or Python `without_files=True` to remain valid table-state objects.

The key invariant is:

> A lazy snapshot is a complete snapshot. Missing materialized files means "not cached," not "empty table."

Do not introduce a broad public `SnapshotLike` trait. `EagerSnapshot` is a compatibility adapter around the canonical snapshot state. A broad trait would either mirror `Snapshot` or hide materialization behavior behind another abstraction.

## Design Direction

`Snapshot` owns delta-rs table-state semantics for:

- table identity and version;
- protocol, metadata, schema, table properties, and resolved table configuration;
- lazy replay of active add actions;
- lazy replay of tombstones;
- transaction conflict read-set construction;
- explicit materialization and cache policy;
- typed semantic errors for replay-backed table-state capabilities.

Delta Kernel remains responsible for protocol-correct replay primitives, checkpoint interpretation, action schemas, and kernel-level scan construction. `Snapshot` may delegate to kernel components, but it owns the delta-rs interface and compatibility contract.

`DeltaScanNext` and `FileSelection` continue to own selected-file scan semantics. Selected files are scan intent, not snapshot identity. `Snapshot` supplies table truth: active membership, protocol, metadata, deletion vectors, column mapping, stats, partition values, and paths.

## Goals

The target state where internal table state consumers work with `Snapshot` directly. The migration should reduce wrapper hops, reduce materialized log dependencies, and move eager compatibility to the edge.

In the end state:

[...]

PR #4661, opened 2026-08-23
refactor: use Snapshot for file discovery

ethan-tyler:
# Description

Use `Snapshot` as the internal state boundary for file discovery in delete, update, and overwrite planning. Each operation unwraps `EagerSnapshot` once at its compatibility boundary and passes `&Snapshot` through the internal discovery path.

Partition predicate discovery reads active adds through Snapshot replay APIs. Lazy snapshots can discover files without materializing file state, and the discovery path reuses the replayed batches for predicate evaluation and `Add` reconstruction.

Preserve complete `Add` actions when converting `LogicalFileView`, including tags, deletion vectors, row tracking fields, and clustering metadata. Existing `EagerSnapshot` entry points remain as compatibility wrappers.

# Related Issue(s)

Part of #4584

PR description drafted and completed code review with Codex

### DIFF
--- a/crates/core/src/kernel/snapshot/iterators.rs
+++ b/crates/core/src/kernel/snapshot/iterators.rs
@@ -37,6 +37,10 @@ const FIELD_NAME_SIZE: &str = "size";
 const FIELD_NAME_MODIFICATION_TIME: &str = "modificationTime";
 const FIELD_NAME_FILE_CONSTANT_VALUES: &str = "fileConstantValues";
 const FIELD_NAME_RAW_PARTITION_VALUES: &str = "partitionValues";
+const FIELD_NAME_TAGS: &str = "tags";
+const FIELD_NAME_BASE_ROW_ID: &str = "baseRowId";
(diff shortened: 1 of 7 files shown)


crates/core/src/kernel/snapshot/iterators.rs
```diff
@@ -202,13 +206,44 @@ impl LogicalFileView {
...
+            .file_constant_values()?
+            .column_by_name(field_name)?
+            .as_map_opt()?;
+        values
+            .is_valid(self.index)
+            .then(|| collect_string_map(&values.value(self.index)))
+            .flatten()
+    }
+
+    fn file_constant_i64(&self, field_name: &str) -> Option<i64> {
```
ion-elgreco, line 227: Do we have test coverage for these new file constant fns?

ion-elgreco (dismissed):
Lgtm, just have a look at my only comment about test coverage

rtyler (approved):
I agree with @ion-elgreco about some test coverage but it's not a blocking requirement for me here since the changes already have a good set of new tests introduced. I'm happy to incorporate more tests at any time 😄

rtyler (approved):

PR #4666, opened 2026-08-24
refactor(python): drop file_pruning_predicate from the dataframe read methods

FrankPortman:
# Description

Removes `file_pruning_predicate` from `to_pandas` and `to_pyarrow_table` before it ships in a release. 

On these two methods `filters` already prunes files: every dataset fragment carries its partition values and min/max statistics as a guarantee expression, so files that cannot match are skipped before any data is read, and the surviving rows are then filtered exactly. 

A pruning parameter there is a second way to pass the same tuples with strictly weaker semantics (whole files instead of rows). The parameter stays where nothing else does the job: `file_uris`, `partitions`, and `to_pyarrow_dataset`.

The parameter landed in #4585 AFTER python-v1.6.3 cut, so it has never been released and no user can be calling it unless building themselves from Github.

We should land this before the next Python release or just close if it's not important. Doesn't seem super worth a deprecation cycle to land later but of course it costs nothing to leave this up for a bit so I defer to y'all. 

# Related Issue(s)

- follow up to #4585

# Documentation

`docs/usage/querying-delta-tables.md` and the file-skipping page now describe `filters` as the pruning-and-filtering path on the dataframe methods.

### DIFF
--- a/docs/how-delta-lake-works/delta-lake-file-skipping.md
+++ b/docs/how-delta-lake-works/delta-lake-file-skipping.md
@@ -74,22 +74,23 @@ Polars can use the Delta table metadata to skip the file that does not contain d
 
 ## File skipping with deltalake
 
-The same skipping is exposed directly on `DeltaTable`: file-listing and read
-methods accept a SQL `predicate` (or tuple filters) and prune files against the
-transaction log before any data is read.
+The same skipping is exposed directly on `DeltaTable`: the file-listing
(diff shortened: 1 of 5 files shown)


ion-elgreco (dismissed):

FrankPortman:
Added a commit to re-trigger CI but that may have backfired on me

ion-elgreco (approved):

PR #4667, opened 2026-08-24
feat(python): route to_pyarrow_table and to_pandas through QueryBuilder

FrankPortman:
# Description

Routes `to_pyarrow_table` and `to_pandas` through QueryBuilder (DataFusion) when the call is expressible as SQL: a column projection plus tuple `filters`, rendered as a `SELECT` with a `WHERE` built from the tuples. 

Suggested by @ion-elgreco in the #4585 review ("why should we even go through pyarrow for pandas, we can also do QueryBuilder -> Arro3 -> Pandas").

The immediate win is capability: the DataFusion path has no columnMapping or deletionVectors guard, so tables with those features become readable through `to_pandas`/`to_pyarrow_table` today (tested against the Spark-written column-mapped fixture). Calls that need pyarrow specifics keep the existing dataset path unchanged: `Expression` filters, a custom `filesystem`, `file_pruning_predicate`, or filter values with no SQL literal. Results are cast to the table's logical arrow schema, and equivalence with the pyarrow path is asserted in tests.

Two things to weigh:

- Row order on the QueryBuilder path follows query execution, not log order, and is not deterministic across calls. One existing test is now sorted.
- This adds a second read path. It pays for itself only as step one of converging the dataframe methods on one engine, with `to_pyarrow_dataset` remaining the explicit pyarrow interop surface. 
- Steps two and three are built and stacked on this branch (deprecate the pyarrow fallbacks, then remove them a release later per a typical deprecation facing, but I completely defer to y'all on when/if we can just rip the bandaid off). 

FrankPortman/delta-rs#4 (deprecate the pyarrow fallbacks), then FrankPortman/delta-rs#5 (remove them) are stacked on this, and IMHO this PR is not worth landing in isolation so if this is not a useful direction, I can just close all 3.

Happy to unstack and just put up a larger PR if that makes sense.

# Related Issue(s)

- follow up to #4585

# Documentation

Covered in the stacked follow-ups. This PR changes no documented behavior for supported calls.

### DIFF
--- a/python/deltalake/table.py
+++ b/python/deltalake/table.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import json
+import math
 import warnings
 from collections.abc import Iterable, Mapping
 from dataclasses import dataclass
-from datetime import datetime, timedelta, timezone
+from datetime import date, datetime, timedelta, timezone
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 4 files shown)


FrankPortman:
@rtyler 

Before you go deeper on this one, I am pretty sure I found three places where the two engines return different rows for the same call once tuple filters route through DataFusion:

* `filters=[("s","=","")]` matches empty strings on the pyarrow path but rewrites to `IS NULL` on the SQL path
*  `NOT IN` drops `null` rows where pyarrow keeps them
* duplicate names in `columns=` are a planning error in DataFusion but legal (and documented) on the pyarrow path.

None currently have specific unit tests for them.

 I see a few ways (not strictly mutually exclusive, I think) to shape this and would like your input:

1. **Narrow the router to capability only.** QueryBuilder handles just the tables the pyarrow scan errors on today (columnMapping, deletion vectors); every other read stays true with the existing pyarrow. FrankPortman/delta-rs#4 + FrankPortman/delta-rs#5 flips everyone in one breaking release where the semantics are documented decisions. This is potentially weird however since "that path used to not work at all" is not necessarily a strong argument in favor of "that path now works, but with different semantics for certain edge cases". So, sitting on the followups may bite users.

2. **Wide router, fall back on anything divergent.** DataFusion serves simple reads immediately, but the router refuses every construct where the engines disagree (`""` values, `None` in sets, `not in`, duplicate columns) and sends those to pyarrow. Cost: the fallback list is invisible at the call site, grows every time we find a new divergence, and is only as complete as our testing.

3. **Declare the semantics and take the break now.** Decide the contract is SQL semantics (three-valued `NOT IN`, `""` is an empty string, duplicate projections error), apply it uniformly, ship as a breaking change. Most honest end state and it shrinks FrankPortman/delta-rs#5 to argument removal, but it skips the deprecation cycle. This might include changes to the QB/DataFusion contract if we are taking an "honest look" at the semantics, which might be cumbersome.

4. **Fold this PR into the stack.** Close it, move the routing and SQL rendering into FrankPortman/delta-rs#5, and the switch ships after FrankPortman/delta-rs#4's warning release. No coexisting engines ever, but columnMapping/DV reads stay broken until then.

5. **Emulate pyarrow bug for bug in SQL.** Total DataFusion coverage with masks like `(col NOT IN (...) OR col IS NULL)` to reproduce pyarrow's null matching. Listed for completeness: it signs us up to mirror every pyarrow accident indefinitely, including ones nobody has found yet. I would not pick it.

FrankPortman:
For the record, here is what I think the cleanest end state is, independent of how we sequence it. I am just unsure how many deprecation hoops we need to jump through to get there.

1. `to_pandas` / `to_pyarrow_table`: thin wrappers over the kernel-backed DataFusion provider. columnMapping/DV/row-tracking work for free. Other details: 
    - NOT IN follows SQL null handling: rows where the column is NULL don't match
    -  `""` is an empty string
    - duplicate projections error
    - Tuple filters survive only as sugar that we can compile into this semantics
    - `filesystem=` goes away, replaced by `storage_options`.

2. Streaming and interop: a `RecordBatchReader` (what `QueryBuilder` already returns) with a direct entrypoint that takes columns + predicate. Query fixed at creation but lazy to compute. Covers DuckDB, polars, pandas, etc glue without materializing and without pyarrow scanner semantics.

3. `to_pyarrow_dataset`: kept as a legacy escape hatch with documented protocol limits. Its unique value is post-handoff pushdown for pyarrow-native scanners, but duckdb/daft use kernel directly and I believe even polars skips pyarrow now. We could even consider deleting `to_pyarrow_dataset` outright eventually and providing clear documentation for all migration paths (all the constructs listed in 1 and 2).

With the following minimal Rust changes:

- A direct read entrypoint (something like `RawDeltaTable.read(columns, predicate)`) that registers the provider and builds the plan directly. Tuples go through the kernel predicate machinery #4585 built, strings through the existing `parse_predicate_expression`. This deletes the tuple-to-SQL string rendering entirely, which is the jankiest part of #4667 and where some of the divergences are from.
- If we want to, make row-ordering deterministic. IMHO the old path was only ordered by accident, but we can fix that as well. Should probably document as a breaking change at minimum.

What I don't have a strong opinion on is the deprecation saga, e.g. whether `filesystem=` and `Expression` filters need the full couple-of-releases warning cycle, whether the row-order change counts as breaking on its own, and so on.

ion-elgreco:
To ease the migration since datafusion behaves differently than pyarrow, we can add an engine param, which takes "pyarrow" or "datafusion", then we keep pyarrow as default, mark it deprcated and simply note that these are behaviour changes users need to account for when migrating

ion-elgreco:
> * A direct read entrypoint (something like `RawDeltaTable.read(columns, predicate)`)

This would be a nice feature, I would name it  `DeltaTable.scan()` though since its lazy

FrankPortman:
@ion-elgreco good idea - my brain has been working in monorepo + "`HEAD` is king" for the last few years, so I forget what nice deprecations or alternative paths in standalone packages look like :). 

Adding the `DeltaTable.scan()` PR right now after which I can re-work this stack.

FrankPortman:
Alright @ion-elgreco PTAL when you can: https://github.com/delta-io/delta-rs/pull/4675

FrankPortman:
Subsumed by https://github.com/delta-io/delta-rs/pull/4676

PR #4670, opened 2026-08-24
chore(core): tighten the filter literal internals

FrankPortman:
Small cleanups behind #4665: `literal_to_predicate_string` drops to `pub(crate)` (one internal caller, renders commit metadata), debug_asserts a validated literal shape instead of silently rendering an op/value mismatch into operationParameters, and the `FilterOp` FromStr error now names the expected operators.


FrankPortman:
@rtyler some small cleanups I found when doing a pass on #4665. noticed it merged when i had tried to push to that branch :D

ion-elgreco (approved):

PR #4675, opened 2026-08-26
feat(python): add DeltaTable.scan for lazy engine-native reads

FrankPortman:
# Description

Adds `DeltaTable.scan(columns=None, predicate=None)`, a lazy typed read over the kernel-backed DataFusion table provider that returns an arro3 `RecordBatchReader` (the same reader type `QueryBuilder.execute` returns). 

Tables with column mapping or deletion vectors become readable without hand-writing `QueryBuilder` SQL, and the reader streams batch-by-batch into anything arrow-native (pyarrow, duckdb, polars). 

The `predicate` is a row filter executed by the engine and returns exactly the matching rows. 

Internally it reuses the `QueryBuilder` session setup and the `DeltaScanNext` provider, going through the DataFusion `DataFrame` API with no SQL string assembly.

Came up in the #4667 review. The datafusion engine work in #4667 can build on this as the read entrypoint.


python/deltalake/table.py
```diff
@@ -562,6 +562,44 @@ def load_cdf(
...
+
+        Parameters:
+            columns: Column names to project. If None, all columns are read.
+            predicate: SQL predicate evaluated per row by the engine; the
+                result contains exactly the matching rows. This is unlike
+                `file_pruning_predicate` on the file listing APIs, which only
+                skips whole files. Follows DataFusion SQL semantics
+                (three-valued NULL logic, SQL type coercion), which differ
+                from pyarrow dataset filters in edge cases; take care when
+                using this as a drop-in replacement.
```
ion-elgreco: Tighten the docstring here a bit, sql predicate according to datafusion sql is enough. Good to verify which sql dialiect datafusion the way we have configured it follows and mention that.

ion-elgreco (commented):
small nit regarding docstring, rest lgtm!

python/deltalake/table.py
FrankPortman: took out some of the specifics and mentions of `pyarrow` and pushed a new commit. LMK if you want it even terser.

ion-elgreco (dismissed):

ion-elgreco:
Thanks! Nice addition:)

FrankPortman:
Trying to kick off the Windows CI again

ion-elgreco (approved):

PR #4677, opened 2026-08-28
feat(writer): optionally roll files on row-group boundaries

djouallah:
# Description

A row group cannot span files, so when `target_file_size` is reached mid-group, `PartitionWriter` closes the file with the buffered partial group — every file's **last row group is truncated** wherever the byte target lands. Readers like Vertipaq  that map one row group to one in-memory segment degrade on those runt tails. Measured on a 142M-row table (6M-row groups, 128 MB target): 4 of 25 groups truncated, one to 1.25M rows.

This adds an **opt-in** group-aligned roll: batch slices never straddle a row-group boundary, and the byte check only fires when the writer sits exactly on one (`in_progress_rows() == 0`). Every row group is then a full `max_row_group_row_count` — the single end-of-data remainder excepted — and a file may overshoot `target_file_size` by at most one row group.

- **Off by default**: behavior is unchanged, existing tests (incl. `test_unflushed_row_group_size`) untouched.
- Enable via `DELTARS_ROLL_ON_ROW_GROUP_BOUNDARY=1` (same pattern as the existing `DELTARS_*` knobs in this file) or `PartitionWriterConfig::with_roll_on_row_group_boundary`.
- Falls back to the legacy roll when row groups aren't bounded by row count alone (`max_row_group_bytes` set, or no row-count bound).
- Unit test included. Happy to thread a `write_deltalake` kwarg instead of the env default if you'd prefer.

**Disclosure:** this PR was authored with AI assistance (Claude); I reviewed and validated it.

# Related Issue(s)

# Documentation

Doc comments on `PartitionWriterConfig::with_roll_on_row_group_boundary` cover the semantics and the overshoot trade.

### DIFF
--- a/crates/core/src/operations/write/writer.rs
+++ b/crates/core/src/operations/write/writer.rs
@@ -68,6 +68,15 @@ fn get_max_concurrency_tasks() -> usize {
     })
 }
 
+fn roll_on_row_group_boundary_default() -> bool {
+    static ROLL_ON_ROW_GROUP_BOUNDARY: OnceLock<bool> = OnceLock::new();
+    *ROLL_ON_ROW_GROUP_BOUNDARY.get_or_init(|| {
+        std::env::var("DELTARS_ROLL_ON_ROW_GROUP_BOUNDARY")
+            .map(|s| matches!(s.to_ascii_lowercase().as_str(), "1" | "true" | "yes"))
+            .unwrap_or(false)
+    })
+}
+
 /// Upload a parquet file to object store and return metadata for creating an Add action
 #[instrument(skip(arrow_writer), fields(rows = 0, size = 0))]
 async fn upload_parquet_file(
@@ -331,6 +340,10 @@ pub struct PartitionWriterConfig {
     write_batch_size: usize,
     /// Concurrency level for writing to object store
     max_concurrency_tasks: usize,
+    /// Defer the `target_file_size` roll until the current row group is complete, so no
+    /// file ends in a truncated row group. See
+    /// [`PartitionWriterConfig::with_roll_on_row_group_boundary`].
+    roll_on_row_group_boundary: bool,
 }
 
 impl PartitionWriterConfig {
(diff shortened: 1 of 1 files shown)


ethan-tyler:
image

djouallah:
what are you talking about , i have been here since 2022 :)

ethan-tyler:
> what are you talking about , i have been here since 2022 :)

oh I know but per GitHub first time contributing 

image

djouallah:
delta_rs can be a little bit scary for contribution :)

ethan-tyler:
Well welcome, stay awhile make yourself at home :)

ion-elgreco:
Would undershooting be better here or do we then have problems with optimize again

djouallah:
> Would undershooting be better here or do we then have problems with optimize again

in my use case, both are fine as long as it does not cut the row group before completion.

crates/core/src/operations/write/writer.rs
```diff
@@ -493,15 +555,26 @@ impl PartitionWriter {
...
-            let length = usize::min(self.config.write_batch_size, max_offset - offset);
+        let mut offset = 0;
+        while offset < max_offset {
+            let mut length = usize::min(self.config.write_batch_size, max_offset - offset);
+            // Never let a slice straddle a row-group boundary: the roll below may only fire
+            // when the writer sits exactly on one (no rows buffered in an open group).
+            let boundary = self.rows_to_row_group_boundary();
+            if let Some(to_boundary) = boundary {
+                length = usize::min(length, to_boundary);
+            }
```
ethan-tyler, line 571: if  `write_batch_size` ever equal 0, it would cause the loop to spin without bound. We previously would panic on such case (which prob was not the best practice).  I think we should reject 0 when constructing `WriterConfig` or `PartitionWriterConfig` or at least return an error before entering this loop.

crates/core/src/operations/write/writer.rs
```diff
@@ -732,6 +806,72 @@ mod tests {
...
+        ]));
+        let batch = RecordBatch::try_new(schema, vec![base_str, base_int]).unwrap();
+
+        let object_store = DeltaTableBuilder::from_url(url::Url::parse("memory:///").unwrap())
+            .unwrap()
+            .build_storage()
+            .unwrap()
+            .object_store(None);
+        let properties = WriterProperties::builder()
+            .set_max_row_group_row_count(Some(1024))
```
ethan-tyler, line 833: This test uses 1024 for both `max_row_group_row_count` and the default `PartitionWriter` write batch size. The exisiting path already pushes bounded slices under those inputs. I was able to confirm this by changing `.with_roll_on_row_group_boundary(true)` to false and the test still passes.

ethan-tyler (commented):
I am aligned, I think overshooting is fine and fits the current meaning of `target_file_size` and requires less prediction. Left a couple of comments.

ethan-tyler:
Getting blessed by them fable tokens 😄

ethan-tyler (approved):
Thanks for the changes, looks good.  Only non blocking nit is the verbosity of the comments.

ethan-tyler:
@djouallah You need DCO sign off on the last commit

djouallah:
thanks @ethan-tyler for the review , really appreciate it

Issue #4171, opened 2026-02-06
[Feature]: Kernel based table scans serialization / wire format


adragomir:
### Is your feature request related to a problem?

ATM, the newly introduced kernel scans, which by default are what is returned from all the `open_table...` functions are not serializable. 
Serialization of physical plan is needed in the Datafusion context for example for distributed execution. Projects like [datafusion-distributed](https://github.com/datafusion-contrib/datafusion-distributed/) rely on serialized plans sent over to worker nodes. 

### Describe the solution you'd like

A new physical codec, that can recreate a `DeltaScanExec` physical plan over the wire - the equivalent of the currently existing `DeltaPhysicalCodec`, which can serialize the old flavor of physical plans, `DeltaScan`. 

### Describe alternatives you've considered

For the moment, what we'll probably do in our usecases, is re-expose the functionality that can create the "old" flavor of plans, logical and physical as those are serializable. 
For the actual implementation, maybe with judicious serde annotations, this can be solved simply, like it is for the current codec, but there are a lot of live objects in there. 

### Priority

High - Important for my use case

### Contribution

- [x] I'm willing to submit a pull request for this feature
- [x] I can help with testing this feature
- [ ] I can help with documentation for this feature

ion-elgreco:
@adragomir feel free to open a PR to add these serialization using datafusion for our new plans

ethan-tyler:
Hey @adragomir - are you still working on this? I just deprecated the legacy codec surface and we ready for the new one. Just lmk and happy to take this one if needed.

jijosg:
+1 on this feature request to be able to work with delta tables in [Datafusion ballista](https://github.com/milenkovicm/ballista_delta).

Nordalf:
Ran into exactly this wiring deltalake 0.32.4 into Ballista 53 (scheduler→executor physical serde). Concrete findings that may help scope the work:

- The **logical** side already round-trips: `DeltaLogicalCodec` serializes the kernel provider (`DeltaScanNext`) as JSON, so client→scheduler works out of the box.
- The **physical** side can't be solved downstream today: `DeltaScanExec` has private fields + `pub(crate) fn new`, and `KernelScanPlan` holds a live kernel `Scan` — so an external `PhysicalExtensionCodec` is impossible without crate changes.
[...]

PR #4679, opened 2026-08-29
feat(datafusion): DeltaScanExecCodec — wire codec for kernel-based scans

Nordalf:
# Description

Kernel-based scans (`DeltaScanExec`, `DeltaScanMetaExec`) hold live kernel state with no wire form, so they cannot cross the physical plan-serde boundary of distributed engines (Ballista, datafusion-distributed, custom schedulers). This PR makes them serializable by capturing and replaying the request that produced them, rather than trying to serialize the live state:

- `ScanReplay` — captures (provider, projection, filters, limit) at plan time in `DeltaScan::scan()`, attached to both scan exec shapes; survives `with_new_children`/`repartitioned` (propagated through `with_new_input`).
- `DeltaScanExecCodec` (`PhysicalExtensionCodec`) — encodes the replay (the provider is already fully serde, with the snapshot embedding the materialized file list; filters travel as a datafusion-proto `LogicalExprList`) and re-runs `scan()` on decode. Metadata replays from the embedded snapshot via `scan_metadata_seeded` (no log-store round trip); deletion-vector and data IO resolve through the receiving side's `RuntimeEnv` from the decode `TaskContext`. The framework-decoded child is intentionally discarded — the replay rebuilds an identical parquet plan deterministically.
- Sync/async bridge: decode delegates via `block_in_place` inside a multi-thread tokio runtime, falls back to a local executor outside one, and errors loudly on a current-thread runtime (would deadlock).

Verified end-to-end in Apache DataFusion Ballista (2-executor cluster over MinIO): distributed SELECT w/ predicate pushdown, joins, time travel, and — the case that has no workaround without this codec — **partitioned tables**, whose partition values arrive via per-file transforms, including the metadata-only `DeltaScanMetaExec` COUNT path. `cargo test --lib --features datafusion next`: no regressions vs main (identical pass/fail set).

Happy to adjust the wire format (e.g. proto instead of serde_json) or the replay-capture placement to maintainer preference.

# Related Issue(s)

- addresses #4171

# Documentation

Module docs on `table_provider/next/codec.rs` describe the replay design and its IO characteristics.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -48,6 +48,7 @@ datafusion = { workspace = true, optional = true }
 datafusion-datasource = { workspace = true, optional = true }
 datafusion-physical-expr-adapter = { workspace = true, optional = true }
 datafusion-proto = { workspace = true, optional = true }
+prost = { version = "0.14.1", optional = true }
 
 # serde
 serde = { workspace = true, features = ["derive"] }
@@ -121,6 +122,7 @@ datafusion = [
     "datafusion-datasource",
     "datafusion-physical-expr-adapter",
     "datafusion-proto",
+    "dep:prost",
 ]
 datafusion-ext = ["datafusion"]
 json = ["parquet/json"]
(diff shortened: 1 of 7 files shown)


Nordalf:
Closing, since it's not ready for a PR. Was a big too quick to the trigger. Ignore.

PR #4681, opened 2026-08-29
fix(datafusion): mark num_rows inexact when deletion vectors are present

Nordalf:
# Description

When deletion vectors are present, the inner parquet plan's row counts do not account for deleted rows. Reporting them as exact lets DataFusion's `AggregateStatistics` optimizer short-circuit `COUNT(*)` with inflated counts (deleted rows counted). This marks `num_rows` as inexact in `statistics_from_inputs` whenever the scan carries selection vectors, so the aggregate executes instead of answering from statistics.

# Related Issue(s)

—

# Documentation

Inline comment at the guard.

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/mod.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/mod.rs
@@ -2587,6 +2587,74 @@ mod tests {
         }])
     }
 
+    /// Regression test for the COUNT(*) deletion-vector bug: the inner parquet
+    /// plan's row counts do not account for deleted rows, so a scan carrying
+    /// selection vectors must report `num_rows` as INEXACT — an exact value lets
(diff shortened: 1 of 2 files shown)


ion-elgreco:
Lgtm, can you add a test @Nordalf ?

Nordalf:
> Lgtm, can you add a test @Nordalf ?

Two tests have been added. I do consider moving this fix into `map_statistics` but I think I'll do that as a follow-up PR. I'm testing if column statistics are inflated too: null_count counts nulls in deleted rows, and if exact min/max can be values that only exist in deleted rows. 

EDIT 1:
We're good. The answers to those two questions become unreachable because both (`count(col)` and `MIN/MAX`) "rely" on `num_rows`

rtyler (approved):

PR #4690, opened 2026-09-01
fix: truncate timestamp min and max to millisecond precision

### LINKED ISSUE #4659: [Bug]: Delta-RS does not truncate timestamp statistics to ms per Delta spec
### What happened?

Hello from the delta-kernel-rs community!

The Delta Protocol (1) says that timestamp statistic values (in the AddFile action) must be truncated down to milliseconds.

Delta-RS does not follow this. Your writer reads (2) min/max values from Parquet metadata and converts timestamps according to their original unit (ms, us,ns) without reducing their precision.

Its custom serializer then uses Chrono’s %.f formatter, preserving the available fractional digits (3).

Consequently, a microsecond timestamp such as .298677Z is written unchanged instead of being truncated to the protocol-required .298Z.

Kernel RS had a similar problem. We fixed it here: https://github.com/delta-io/delta-kernel-rs/commit/51ce3981882a37c2ab7df0e32928b3df803a975b

One _really weird_ thing that can happen is: Using Apache Spark, with `spark.sql.legacy.timeParserPolicy = LEGACY`, Apache Spark's timestamp reader will (incorrectly, buggily) treat the decimal digits as # millieconds. so `.12345` is read as `12,345ms` insted of `123.45ms` or `0.12345s`.

1: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#user-content-fn-1-d4685645e4f15dcb3d12eef5ab499f5e

2: https://github.com/delta-io/delta-rs/blob/cee6f9ea9d1deab2f70d23221abb5386b60f3a9a/crates/core/src/writer/stats.rs#L315-L334

3: https://github.com/delta-io/delta-rs/blob/cee6f9ea9d1deab2f70d23221abb5386b60f3a9a/crates/core/src/writer/stats.rs#L448-L462

### Expected behavior

Truncate timestamp values to milliseconds in the delta log. You also have to account for this during data skipping on reads, btw.

### Operating System

None

### Binding

None

### Bindings Version

_No response_

### Steps to reproduce

_No response_

### Relevant logs

```shell

```

adamreeve:
# Description

Truncates the min and max statistics for timestamps to millisecond precision, for compliance with the Delta Protocol.

# Related Issue(s)

- closes #4659

# Documentation

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics

### DIFF
--- a/crates/core/src/writer/stats.rs
+++ b/crates/core/src/writer/stats.rs
@@ -6,12 +6,13 @@ use std::{
     ops::AddAssign,
 };
 
+use chrono::Timelike;
 use delta_kernel::expressions::Scalar;
 use delta_kernel::table_properties::DataSkippingNumIndexedCols;
 use indexmap::IndexMap;
 use itertools::Itertools;
 use parquet::basic::Type;
-use parquet::basic::{ConvertedType, DecimalType, IntType, LogicalType, TimestampType};
+use parquet::basic::{ConvertedType, DecimalType, LogicalType};
 use parquet::file::metadata::ParquetMetaData;
 use parquet::schema::types::{ColumnDescriptor, SchemaDescriptor};
 use parquet::{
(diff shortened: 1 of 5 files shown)


crates/core/src/writer/stats.rs
```diff
@@ -455,10 +456,23 @@ impl From<StatsScalar> for serde_json::Value {
             StatsScalar::Float64(v) => serde_json::Value::from(v),
             StatsScalar::Date(v) => serde_json::Value::from(v.format("%Y-%m-%d").to_string()),
             StatsScalar::Timestamp(v) => {
-                serde_json::Value::from(v.format("%Y-%m-%dT%H:%M:%S%.fZ").to_string())
+                // Timestamps must be truncated to millisecond precision when writing
+                // to JSON for compatibility with the Delta Protocol.
+                let format = if v.nanosecond() == 0 {
+                    // Omit the sub-second component if it is zero
```
adamreeve, line 462: I did this for consistency with the previous behaviour and to match expected values in tests. It also saves a few bytes in the JSON when they're not needed. But we could just always use "%.3f".

ethan-tyler:
I'll take a look at this tomorrow

rtyler (approved):

PR #4692, opened 2026-09-02
fix: preserve row order in deletion vector scans

ethan-tyler:
# Description

`DeltaScanExec` consumes deletion vectors as per file keep masks in physical Parquet row order. DataFusion could filter, limit, split, or replace the child scan before mask application. Those rewrites could apply mask entries to the wrong rows.

This PR blocks predicate, limit, fetch, and repartition pushdown for scans with deletion vectors. It locks scans to whole file groups and requires a single input partition.

`DeltaScanExec` now validates child rewrites against the planned file identities. It rejects filtered or fetched sources, byte-ranged files, unsupported nodes, and missing, duplicate, or remapped files. Each execution stream receives its own mask copy, which keeps repeated and concurrent executions independent.

Deletion vector scans also report conservative cardinality and statistics because the Parquet child describes physical rows before deletion vector filtering.

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
@@ -49,14 +49,14 @@ use crate::kernel::arrow::engine_ext::ExpressionEvaluatorExt;
 const DELTA_MATERIALIZED_PUSHDOWN_SENTINEL: &str =
     "__delta_rs_unpushable_delta_materialized_filter";
 
-/// DeltaScanExec uses these inputs to apply deletion vectors during a physical scan.
+/// Deletion-vector inputs for [`DeltaScanExec`].
 #[derive(Clone, Debug)]
 pub(super) enum DvExecutionState {
-    None,
+    NotPresent,
     Sequential {
-        /// The planner captures these masks. Each execution stream copies them before use.
+        /// Keep masks shared by plan clones. Each execution copies and consumes them.
         selection_vectors: Arc<HashMap<String, Vec<bool>>>,
-        /// The validator compares child files against this whole file ownership map.
+        /// Expected object store and path for each file ID in the child plan.
         physical_file_identities: Arc<super::PhysicalFileIdentityMap>,
     },
 }
@@ -68,7 +68,7 @@ impl DvExecutionState {
 
     fn selection_vectors(&self) -> Option<&HashMap<String, Vec<bool>>> {
         match self {
-            Self::None => None,
+            Self::NotPresent => None,
             Self::Sequential {
                 selection_vectors, ..
             } => Some(selection_vectors),
@@ -77,7 +77,7 @@ impl DvExecutionState {
 
     fn physical_file_identities(&self) -> Option<&super::PhysicalFileIdentityMap> {
         match self {
-            Self::None => None,
+            Self::NotPresent => None,
             Self::Sequential {
                 physical_file_identities,
                 ..
@@ -561,7 +561,8 @@ impl ExecutionPlan for DeltaScanExec {
             input: self.input.execute(partition, context)?,
             baseline_metrics: BaselineMetrics::new(&self.metrics, partition),
             transforms: Arc::clone(&self.transforms),
-            // Each execution owns its cursors and copies the masks that the planner captured.
+            // Each stream consumes its masks as it reads batches. Deep-copy the masks
+            // once per execution to isolate repeated or concurrent scans.
             selection_vectors: self
                 .dv_state
                 .selection_vectors()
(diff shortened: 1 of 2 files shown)


ryzhyk:
Thanks for the fix @ethan-tyler ! Our customer just ran into this. It's a scary bug, causes rows to disappear from the table. I hope this can be reviewed and merged soon.

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
```diff
@@ -194,7 +202,7 @@ pub(super) async fn replay_deletion_vectors(
     let dv_stream = stream.dv_stream.build();
     // Only files with `dv_info.has_vector()` spawn tasks, so every item should carry a DV.
     // Guard with a typed error (instead of panic) in case that invariant drifts.
-    let dvs: DashMap<_, _> = dv_stream
+    let dvs: HashMap<_, _> = dv_stream
```
rtyler, line 205: Is this a drive-by change for some other reason? I couldn't piece together why this switch off `DashMap` was made here.

`exec_meta` still has some use of Dashmap so it's not as if the dependency is being dropped here.

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -394,7 +561,12 @@ impl ExecutionPlan for DeltaScanExec {
             input: self.input.execute(partition, context)?,
             baseline_metrics: BaselineMetrics::new(&self.metrics, partition),
             transforms: Arc::clone(&self.transforms),
-            selection_vectors: Arc::clone(&self.selection_vectors),
+            // Each execution owns its cursors and copies the masks that the planner captured.
+            selection_vectors: self
+                .dv_state
+                .selection_vectors()
+                .cloned()
```
rtyler, line 568: From a memory standpoint I believe the switch from `DashMap` to `HashMap` results in each of these `selection_vectors()` being deep-cloned, since the `DeltaScanStream` requires a full `HashMap<String, Vec<bool>>` now.

Is this a one time copy ever? Or will this end up copying the selection vectors repeatedly at runtime? I'm not sure how much during a datafusion operation `execute` would be called

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
```diff
@@ -47,6 +49,43 @@ use crate::kernel::arrow::engine_ext::ExpressionEvaluatorExt;
 const DELTA_MATERIALIZED_PUSHDOWN_SENTINEL: &str =
     "__delta_rs_unpushable_delta_materialized_filter";
 
+/// DeltaScanExec uses these inputs to apply deletion vectors during a physical scan.
+#[derive(Clone, Debug)]
+pub(super) enum DvExecutionState {
+    None,
```
rtyler, line 55: Using `None` as an arm of an enum is awkawrd and could be confusing. I would recommend something like `NotPresent` or analogous here

ryzhyk:
I wonder if a proper fix for the issue is in the works. As expected, this PR throws away a lot of performance (since it drops predicate pushdown or paritioning), which is certainly better than returning incorrect results, but I expect the real fix should be able to keep the optimizations.

ethan-tyler:
> I wonder if a proper fix for the issue is in the works. As expected, this PR throws away a lot of performance (since it drops predicate pushdown or paritioning), which is certainly better than returning incorrect results, but I expect the real fix should be able to keep the optimizations.

Yes @ryzhyk  - this is a correctness fix/patch. With the new `file_row_index` in datafusion 55 we will be able to reenable the optimized plans safely. I am stacking that pr on top of this and we should be able to merge all of it together.

ryzhyk:
> > I wonder if a proper fix for the issue is in the works. As expected, this PR throws away a lot of performance (since it drops predicate pushdown or paritioning), which is certainly better than returning incorrect results, but I expect the real fix should be able to keep the optimizations.
> 
> Yes @ryzhyk - this is a correctness fix/patch. With the new `file_row_index` in datafusion 55 we will be able to reenable the optimized plans safely. I am stacking that pr on top of this and we should be able to merge all of it together.

Sounds great, thanks!

crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs
ethan-tyler: yes this change is incidental to this fix, I restored DashMap

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
ethan-tyler: Renamed it to `DvExecutionState::NotPresent` and updated the matches and constructors

crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs
ethan-tyler: Ok so there's two parts to this

1) Yes - , `.cloned()` deep copies the map and its masks once for each dv stream created by `execute()`. However it doesn't copy them for each batch, but executing the same plan again makes another copy and concurrent executions each hold their own copy. The current dv path requires one input partition and it makes one full copy per execution of that scan.


2) Sharing that mutable state lets one execution consume masks another execution needs and this private copy is part of the repeatability fix.  Yes - this fix retains that cost. However we will remove it in my "physical row position" follow up, where I plan to replace this with a shared immutable dv index and it will remove this per execution copy.

rtyler (approved):

PR #4693, opened 2026-09-03
fix: keep non-skippable columns out of the kernel scan predicate

adampolomski:
# Description
A filter on a Map / List / whole-Struct column fails.

Test case demonstrating the problem:
```
use std::io::Cursor;
use std::sync::Arc;

use arrow_json::ReaderBuilder;
use deltalake::delta_datafusion::{DeltaScanConfigBuilder, DeltaScanNext};
use deltalake::kernel::{DataType, MapType, StructField};
use deltalake::operations::create::CreateBuilder;
use tempfile::TempDir;

#[tokio::test]
async fn predicate_on_non_primitive_column_does_not_fail_the_scan() {
    let tmp_dir = TempDir::new().unwrap();
    let table_dir = tmp_dir.path().join("map_predicate");
    std::fs::create_dir(&table_dir).unwrap();

        .with_location(table_dir.to_str().unwrap())
        .with_columns(vec![
            StructField::nullable("id", DataType::INTEGER),
            StructField::nullable("m", MapType::new(DataType::STRING, DataType::STRING, true)),
        ])
        .await
        .unwrap();

    let batch = ReaderBuilder::new(table.snapshot().unwrap().snapshot().arrow_schema())
        .build(Cursor::new(
            br#"{"id": 1, "m": {"k": "v"}}
               {"id": 2, "m": null}"#
                .to_vec(),
        ))
        .unwrap()
        .next()
        .unwrap()
        .unwrap();
    let table = table.write(vec![batch]).await.unwrap();

    let snapshot = table.snapshot().unwrap().snapshot();
    let scan_config = DeltaScanConfigBuilder::new()
        .with_schema(snapshot.arrow_schema())
        .build(snapshot)
        .unwrap();
    let ctx = datafusion::prelude::SessionContext::new();
    ctx.register_table(
        "t",
        Arc::new(DeltaScanNext::new(snapshot.clone(), scan_config).unwrap()),
    )
    .unwrap();

    let count = ctx
        .sql(&"SELECT id FROM t WHERE m IS NOT NULL".to_string())
        .await
        .unwrap()
        .count()
        .await
        .unwrap();
    assert_eq!(count, 1);
}
```

Result:
> External error: Kernel error: Predicate references unknown column: m

Expected result:
Query should return a single row.

Affected version: 0.32.4

# Related Issue(s)
N/A

# Documentation
N/A

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
@@ -591,30 +591,36 @@ fn process_predicate<'a>(
 
     // Try to convert the expression into a kernel predicate
     if let Ok(kernel_predicate) = to_delta_predicate(expr) {
-        let (pushdown, parquet_predicate) = if only_partition_refs {
-            // All references are to partition columns so the kernel
-            // scan can fully handle the predicate and return exact results
-            (TableProviderFilterPushDown::Exact, None)
-        } else if any_partition_refs {
-            // Some references are to partition columns, so the kernel
-            // scan can only handle the predicate on best effort. Since the
-            // parquet scan cannot reference partition columns, we do not
-            // push down any predicate to parquet
-            (TableProviderFilterPushDown::Inexact, None)
-        } else {
-            // For non-partition predicates we can *attempt* Parquet pushdown, but it is not a
-            // correctness boundary (it may be partially applied or skipped). Keep this Inexact so
-            // DataFusion retains a post-scan Filter.
-            (
-                TableProviderFilterPushDown::Inexact,
-                parquet_pushdown_enabled.then_some(expr),
-            )
-        };
-        return ProcessedPredicate {
-            pushdown,
-            kernel_predicate: Some(kernel_predicate),
-            parquet_predicate,
-        };
+        if is_data_skipping_eligible(&kernel_predicate, config) {
+            let (pushdown, parquet_predicate) = if only_partition_refs {
+                // All references are to partition columns so the kernel
+                // scan can fully handle the predicate and return exact results
+                (TableProviderFilterPushDown::Exact, None)
+            } else if any_partition_refs {
+                // Some references are to partition columns, so the kernel
+                // scan can only handle the predicate on best effort. Since the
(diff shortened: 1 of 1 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -661,6 +667,27 @@ fn process_predicate<'a>(
...
+        .zip(types)
+        .filter(|(_, data_type)| matches!(data_type, KernelDataType::Primitive(_)))
+        .map(|(name, _)| fold(name))
+        .collect();
+
+    predicate
+        .references()
+        .iter()
+        .all(|reference| skippable.contains(&fold(reference)))
+}
```
ethan-tyler: This is likely to add noticeable planning overhead for wide tables or queries with many conjuncts.

I think a better way to implement this is to compute the folded primitive leaf index once per planning operation and then pass it into process_predicate. Or you could resolve only the paths referenced by each predicate.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
```diff
@@ -661,6 +667,27 @@ fn process_predicate<'a>(
     }
 }
 
+/// Whether kernel data skipping can resolve every column the predicate references.
```
ethan-tyler: We should ensure we are distinguishing valid nonskippable references from invalid column references.

Are you able to validate references against the logical table schema independently of data skipping eligibility?

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
adampolomski: I took the latter path. In majority of the cases it should be constant time. `field_ignore_case` falls back to linear scan when case-insensitive lookup is required.

crates/core/src/delta_datafusion/table_provider/next/scan/plan.rs
adampolomski: Invalid column references should now be handled exactly as they used to. It was not my intention to change that behavior in any way.

ethan-tyler (approved):

PR #4697, opened 2026-09-06
chore: add regression test for percent encoding as part of #4684

rtyler:
This came up in an automated review on some changes with #4684 that modified some percent encoding handling. Those are the bane of my existence.


ethan-tyler (approved):

ethan-tyler:
@rtyler - this is ready

PR #4700, opened 2026-09-07
fix(datafusion): apply deletion vectors by physical row position

ethan-tyler:
Deletion-vector visibility now uses `(compact file identity, absolute physical Parquet row position)`. A complete selected-file registry is captured before DV loading, and one immutable `Arc<DeletionVectorIndex>` owns the validated masks across executions. Missing expected vectors, unknown identities, inconsistent counts and invalid positions fail closed. Payload and identity are filtered together before Kernel transformations.

This PR retains DV predicate, file-layout, partition/repartition and LIMIT/fetch containment. MERGE keeps its one-based surviving-row ordinal; migrating that ordinal is a separate follow-up. No public metadata functions, decoder-level DV selection, durable RowTracking or physical serialization codec are added.

The private reader contract checks hidden fields, bound stores/readers/adapters, footer counts and coordinate metadata. A store-qualified adapter uses the existing session metadata cache, sharing one retention allowance across stores and concurrent scans. Snapshot visibility is separate from reusable file metadata. Metadata-only exact counts rely on validated log/DV protocol metadata; they are not presented as footer-verified reads. Non-DV files without statistics remain readable.

### Stack and review range

Depends on open #4692 at `42ef7613dc6a9f40a6a65b9475cb24dd28ad7b1a`. Landing order is #4692 → this correctness follow-up → optimizer follow-up. Suitable upstream stacking branches are unavailable, so this draft targets `main` and GitHub's displayed diff is cumulative.

Review the immutable [correctness-base → PR1 diff](https://github.com/ethan-tyler/delta-rs/compare/42ef7613dc6a9f40a6a65b9475cb24dd28ad7b1a...fbe01a93a9e647ded3f9fbc247bb03ad43ff8f51). The optimizer follow-up is being qualified separately; this PR does not claim to eliminate every ordering restriction discussed in #4115.

### Local qualification

Source `fbe01a93a9e647ded3f9fbc247bb03ad43ff8f51`; incremental parent `42ef7613dc6a9f40a6a65b9475cb24dd28ad7b1a`. Rust 1.94.1, `aarch64-apple-darwin`, `datafusion` feature, dev/test profile.

[...]

PR #4703, opened 2026-09-08
refactor: apply deletion vectors by physical row position

ethan-tyler:
# Description
So I reviewed the rouge agent's work in #4700 and pulled out all the good stuff and polished it up 

This PR builds on #4692 and replaces consuming deletion masks with lookups keyed by file id and parquet row position. 

This keeps deletion filtering consistent across batch boundaries, repeated execution, and cancellation. Preserve the surviving row ordinals used by MERGE.

Validate file metadata, reader settings, and footer row counts before applying deletion vectors. Keep the physical row position column internal to the scan and share cached Parquet metadata by object store and path. Existing restrictions on predicate pushdown and repartitioning remain in place.

Add coverage for reader compatibility, child plan replacement, cancellation, and surviving row ordinals across batches and files.

I am planning on another follow up PR that will restore predicate pushdown and parallel scans to capture the expected performance impact.

# Related Issue(s)

# Documentation

### DIFF
--- a/crates/core/src/delta_datafusion/table_provider/next/scan/metadata_cache.rs
+++ b/crates/core/src/delta_datafusion/table_provider/next/scan/metadata_cache.rs
@@ -1,0 +1,188 @@
+//! Store-qualified views of the session cache. The backing cache owns the single
+//! retention allowance, including entries from concurrent, independently planned scans.
+
+use std::{
+    any::Any,
+    sync::{Arc, Weak},
+    time::Duration,
+};
+
+use datafusion::{
+    common::{HashMap, Result, TableReference},
+    execution::cache::{
+        Cache, CacheEntryInfo,
+        cache_manager::{CachedFileMetadataEntry, FileMetadata, FileMetadataCache},
+    },
+};
+use object_store::{ObjectMeta, ObjectStore, path::Path};
+
+#[derive(Debug)]
+pub(super) struct StoreMetadataCache {
+    backing: Arc<FileMetadataCache>,
+    store: Arc<dyn ObjectStore>,
+    object: ObjectMeta,
+}
+
+struct TaggedMetadata {
+    // Weak retains the allocation identity, preventing address reuse without
+    // retaining an unregistered store's data for the cache entry's lifetime.
+    store: Weak<dyn ObjectStore>,
+    value: Arc<dyn FileMetadata>,
+    object: ObjectMeta,
+    footer: Option<std::result::Result<super::reader::FooterEvidence, String>>,
+}
+
+impl FileMetadata for TaggedMetadata {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+    fn memory_size(&self) -> usize {
+        self.value.memory_size()
+            + std::mem::size_of::<Self>()
+            + self.object.location.as_ref().len()
+            + self.object.e_tag.as_ref().map_or(0, String::len)
+            + self.object.version.as_ref().map_or(0, String::len)
+            + self
+                .footer
+                .as_ref()
+                .and_then(|proof| proof.as_ref().err())
+                .map_or(0, String::len)
+    }
+    fn extra_info(&self) -> HashMap<String, String> {
+        self.value.extra_info()
+    }
+}
+
+impl StoreMetadataCache {
+    pub(super) fn new(
+        backing: Arc<FileMetadataCache>,
+        store: Arc<dyn ObjectStore>,
+        object: ObjectMeta,
+    ) -> Self {
+        Self {
+            backing,
+            store,
+            object,
+        }
+    }
+
+    pub(super) fn footer_evidence(
+        &self,
+        metadata: &Arc<parquet::file::metadata::ParquetMetaData>,
+    ) -> std::result::Result<super::reader::FooterEvidence, String> {
+        if let Some(entry) = self.backing.get(&self.key(&self.object.location))
+            && let Some(tagged) = entry.file_metadata.as_any().downcast_ref::<TaggedMetadata>()
+            && Weak::ptr_eq(&Arc::downgrade(&self.store), &tagged.store)
+            && tagged.object == self.object
+            && let Some(cached) = tagged.value.as_any().downcast_ref::<datafusion::datasource::physical_plan::parquet::metadata::CachedParquetMetaData>()
+            && Arc::ptr_eq(cached.parquet_metadata(), metadata)
+            && let Some(evidence) = &tagged.footer
+        {
+            return evidence.clone();
+        }
+        super::reader::validate_footer(metadata)
+    }
+
+    fn key(&self, path: &Path) -> Path {
+        // The store allocation and canonical path qualify each key. Tagged values
+        // distinguish these entries from bare-path entries and verify full identity.
+        let address = Arc::as_ptr(&self.store) as *const () as usize;
+        Path::from(format!("__delta_rs_metadata/{address:x}/{}", path.as_ref()))
+    }
+
+    fn untag(&self, entry: CachedFileMetadataEntry) -> Option<CachedFileMetadataEntry> {
+        let tagged = entry
+            .file_metadata
+            .as_any()
+            .downcast_ref::<TaggedMetadata>()?;
+        if !Weak::ptr_eq(&tagged.store, &Arc::downgrade(&self.store))
+            || tagged.object != self.object
+        {
+            return None;
+        }
+        Some(CachedFileMetadataEntry::new(
+            tagged.object.clone(),
+            Arc::clone(&tagged.value),
+        ))
+    }
+}
+
+impl Cache<Path, CachedFileMetadataEntry> for StoreMetadataCache {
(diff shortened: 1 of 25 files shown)


crates/core/src/delta_datafusion/table_provider/next/scan/metadata_cache.rs
```diff
@@ -0,0 +1,187 @@
...
+            backing,
+            store,
+            object,
+        }
+    }
+
+    pub(super) fn footer_evidence(
+        &self,
+        metadata: &Arc<parquet::file::metadata::ParquetMetaData>,
+    ) -> std::result::Result<super::reader::FooterEvidence, String> {
```
rtyler, line 71: This is **gnarly**, I cannot sensibly work through this without spending an unnecessary amount of time.

I think this function should be expanded out to not smash everything into a 10 armed conditional

crates/core/src/delta_datafusion/table_provider/next/scan/metadata_cache.rs
```diff
@@ -0,0 +1,187 @@
+//! Parquet metadata cache keyed by object store and path.
+//! All scans share the session cache's capacity limit.
+
```
rtyler, line 3: I don't understand why this is necessary for a Datafusion-based bit of code. My understanding was that the `datafusion::datasource::physical_plan::parquet::metadata` structs cache parquet metadata already within Datafusion

crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs
```diff
@@ -78,25 +72,53 @@ pin_project! {
...
+    for (selected_id, file) in files.iter().enumerate() {
+        if !file.expected_dv.has_vector() {
+            continue;
+        }
+        let engine = engine.clone();
+        let descriptor = file.expected_dv.clone();
+        let table_root = table_root.clone();
+        let tx = stream.tx();
+        stream.spawn_blocking(move || {
+            let keep_mask = descriptor.get_selection_vector(engine.as_ref(), &table_root)?;
```
rtyler, line 102: If this error `?` happens, my understanding is that the spawn_blocking operation will abort here and effectively drop anything pending in this stream.

rtyler (changes requested):
Mentioned in Slack, but if you're gonna have 🦜 write some code, have it write a lot more rustdoc for these newly added structs and functions to help others understand what the hizzle is happening here

rtyler:
This is a :parrot: review which was subsequently reviewed by me

## Code Review — High & Medium Findings

---

### [HIGH] Stale `file_id_idx` after `remove_column(position_idx)` may remove wrong column or panic

**File:** `crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs:729-773`

`file_id_idx` is computed from the original batch at the top of `batch_project`. Inside `batch_project_single_file`, when deletion vectors are present, `filtered.remove_column(position_idx)` removes the physical-position column *before* `batch.remove_column(file_id_idx)` removes the file-id column. If `position_idx < file_id_idx`, removing the position column shifts every subsequent column left by one — making `file_id_idx` point to the wrong column (or panic if file-id was last).

**Current behaviour:** DataFusion currently emits virtual columns (PHYSICAL_POSITION_COLUMN) *after* partition columns (file_id), so the schema is `[data..., file_id, position]` and the code works today. But this ordering is an undocumented internal DataFusion detail with no assertion guarding it.

**Suggested fix:** Remove the higher-indexed column first to avoid index shifting, or look up both column positions immediately before the removals rather than reusing a stale index:

```rust
// Remove higher index first to keep the lower index stable.
let position_idx = batch.schema().index_of(super::PHYSICAL_POSITION_COLUMN)?;
// file_id_idx was computed before any mutation; adjust if position comes first
let adjusted_file_id_idx = if position_idx < file_id_idx {
    file_id_idx - 1
} else {
    file_id_idx
};
filtered.remove_column(position_idx);
filtered.remove_column(adjusted_file_id_idx);
```

Or add a `debug_assert!(position_idx > file_id_idx, "position column must follow file-id column")` to lock in the ordering assumption at test time.

---

### [HIGH] `append_row_index` assigns 1-based ordinals — contract with MERGE is undocumented

**File:** `crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs:832-836`

```rust
let values = if row_count == 0 {
    Vec::new()
} else {
    ((*next_row_index + 1)..=end).collect()  // 1-based: first row gets ordinal 1
};
```

For the first batch from a file (`next_row_index = 0`, `row_count = N`, `end = N`), the range is `(1..=N)` — ordinals `[1, 2, ..., N]`. These are **1-based**.

The PR description says *"Preserve the surviving row ordinals used by MERGE."* DV filtering uses 0-based physical positions (from `PHYSICAL_POSITION_COLUMN`). If MERGE correlates surviving rows back to physical positions via these ordinals, an off-by-one means MERGE acts on the wrong rows. If 1-based is intentional (SQL `ROW_NUMBER()` semantics), that should be documented and verified against what MERGE actually consumes.

**Suggested fix:** Add a comment explaining the intended contract, or change to 0-based and verify against the MERGE consumer:

```rust
// Row-index ordinals are 0-based: ordinal 0 = first surviving row of this file.
// MERGE uses these to correlate surviving rows; must be consistent with physical positions.
let values = (*next_row_index..end).collect();
```

---

### [MEDIUM] `ScanFileStream` returns unhelpful error when DV file has no descriptor in scan metadata

**File:** `crates/core/src/delta_datafusion/table_provider/next/scan/replay.rs` (inside `poll_next`)

```rust
let Some(cardinality) = cardinality else {
    return Poll::Ready(Some(Err(DeltaTableError::generic(
        "Selected DV file has no descriptor",
    ))));
};
```

`dv_info.has_vector()` and `deletion_vector_descriptor()` come from different sources. If they disagree (kernel version mismatch, corrupted metadata), the user gets a generic error with no file path and no actionable detail.

**Suggested fix:**

```rust
let Some(cardinality) = cardinality else {
    return Poll::Ready(Some(Err(DeltaTableError::generic(format!(
        "Deletion vector file '{}' has DV info but no descriptor in scan metadata",
        super::super::redact_url_for_error(&file.file_url)
    )))));
};
```

---

### [MEDIUM] Inconsistent short-mask handling between scan execution and the `deletion_vectors()` API

**Files:**
- Scan execution: `exec.rs:761` — `mask.get(index).copied().unwrap_or(true)` (short mask silently extended)
- Public API: `mod.rs:414-441` — `normalize_dv_keep_mask_for_api` pads to `numRecords`; a mask longer than `numRecords` is an **error**

These two paths handle short DV masks in fundamentally different ways. A future maintainer wiring masks from the API into scan state (or vice versa) will be surprised by the asymmetry. Neither site documents the divergence.

**Suggested fix:** Add a comment at both sites explaining the intentional difference:

```rust
// Scan execution resolves each row by physical position with `.unwrap_or(true)`;
// short masks are valid per the Delta spec — kernel stops at the last deleted row.
// The deletion_vectors() API pads masks to numRecords for external consumers;
// see normalize_dv_keep_mask_for_api in mod.rs.
```

---

### [MEDIUM] `validate_dv_parquet_metadata` silently accepts files with >32,767 row groups when ordinals are absent

**File:** `crates/core/src/delta_datafusion/table_provider/next/scan/mod.rs:457-462`

```rust
if group.ordinal() != i16::try_from(group_index).ok() {
```

When `group_index >= 32768`, `i16::try_from(group_index).ok()` returns `None`. If `group.ordinal()` is also `None` (ordinal not stored in footer), the comparison `None != None` is `false` — the check **passes silently** despite an unvalidated row group. Files with this many row groups are rare but valid Parquet.

**Suggested fix:** Reject `None` ordinals explicitly:

```rust
let ordinal = group.ordinal().ok_or_else(|| DataFusionError::Plan(format!(
    "Parquet row-group {group_index} has no ordinal for file id '{id}'"
)))?;
let expected = i16::try_from(group_index).map_err(|_| DataFusionError::Plan(format!(
    "Parquet file '{id}' has too many row groups ({group_index})"
)))?;
if ordinal != expected {
    return plan_err!("Parquet row-group ordinal mismatch for file id '{id}'");
}
```

---

### [MEDIUM] `validate_dv_child_topology` uses pointer identity to track plan nodes, which breaks after optimizer rewrites

**File:** `crates/core/src/delta_datafusion/table_provider/next/scan/exec.rs` (`validate_dv_child_topology`)

The validator uses pointer address identity (via `HashSet<usize>`) to find `DataSourceExec` leaf nodes in the child plan tree, then verifies every expected file identity appears in those leaves. If DataFusion's optimizer replaces a `DataSourceExec` with a new `Arc` wrapping identical content (which happens after `with_new_children`), the pointer changes and validation incorrectly reports "DV child omits selected files". Conversely, an `Arc` clone shared across two plan paths would make the validator believe fewer distinct nodes are present.

DataFusion makes no guarantee that plan nodes retain pointer identity across optimization passes.

**Suggested fix:** Enumerate file identities from the `FileScanConfig` directly (comparing store URL + path + size) rather than relying on pointer identity to distinguish nodes.

ethan-tyler:
@rtyler damn that's brutal 😩

Issue #4549, opened 2026-06-15
Catalog Managed Commits Support


elazarlachkar:
### What happened?

Opening a Delta table that has the [catalogManaged protocol feature](https://docs.delta.io/delta-catalog-managed-tables/) raises an error with no Python-level workaround:

```
deltalake.exceptions.DeltaError: Kernel error: Generic delta kernel error:
Catalog-managed table requires max_catalog_version to be set.
Use with_max_catalog_version() when loading a catalog-managed table.
```
The underlying Rust kernel (delta-kernel-rs) exposes `with_max_catalog_version()` to handle these tables, but the Python bindings provide no way to set this value, making catalog-managed tables completely inaccessible from Python.

### Expected behavior

`DeltaTable(...)` should accept a `max_catalog_version` parameter that is forwarded to the kernel.

### Operating System

Linux

### Binding

Python

### Bindings Version

`>=1.5.1`

### Steps to reproduce

1. Create a Delta table in with the `catalogManaged` table feature:
```
CREATE TABLE test_catalog_commits (
  id BIGINT,
  value DECIMAL(10,2)
)
TBLPROPERTIES ('delta.feature.catalogManaged' = 'supported');
```

2. Add some data to generate commits:
```
INSERT INTO test_catalog_commits VALUES (1, 100.00);
```

3. Attempt to open it:
```
from deltalake import DeltaTable
dt = DeltaTable("abfss://container@account.dfs.core.windows.net/path/to/table")
```

Observe the DeltaError above.

hntd187:
This is unsupported currently and won't work in delta-rs. I'm not sure when we'll be supporting this as there is a lot of surrounding pieces that also need to support ccv2 for this to function.

elazarlachkar:
Thanks for the answer @hntd187, I'll be happy to get updates once you get to it 🙏🏼 
In the meantime I suggest changing the error a bit (something like "Catalog managed tables are currently unsupported"), as the current error made me search for where I can set `max_catalog_version`.

abhishekverma-ray:
Hi delta-rs maintainers 👋

[...]

PR #4710, opened 2026-09-10
refactor: scope log store writes per operation and introduce committer trait

ion-elgreco:
# Description
Every writing operation now runs inside an operation scope owned by the log store, instead of threading an operation id through core and relying on a CustomExecuteHandler that callers had to attach by hand.


ethan-tyler:
image

crates/lakefs/src/transaction.rs
```diff
@@ -0,0 +1,776 @@
...
+        let published = if dirty { self.publish().await } else { Ok(()) };
+        // The branch is deleted whether or not publishing succeeded, so a failed merge never
+        // leaves a stale transaction branch behind.
+        let deleted = self
+            .client
+            .delete_branch(&self.location.repo, &self.branch)
+            .await;
+        self.closed.store(true, Ordering::SeqCst);
+        published.and(deleted)
+    }
```
ethan-tyler, line 381: If `delete_branch` fails, this is still marking the transaction closed. The scope then calls abort but cannot retry the delete and leaves the hidden branch. Would it be better to mark it closed only after deletion succeeds and test a failed delete followed by a retry?

crates/core/src/logstore/scope.rs
```diff
@@ -0,0 +1,971 @@
...
+    async fn put_opts(
+        &self,
+        location: &Path,
+        payload: PutPayload,
+        opts: PutOptions,
+    ) -> ObjectStoreResult<PutResult> {
+        self.state.ensure_open()?;
+        self.state.mark_dirty();
+        self.inner.put_opts(location, payload, opts).await
+    }
```
ethan-tyler, line 166: Can `finish` run while this `put_opts` is pending? It could merge and delete the branch before the write completes. Could it wait for writes already in flight, including multipart uploads?

crates/core/src/operations/mod.rs
```diff
@@ -260,63 +283,6 @@ impl DeltaTable {
     }
 }
 
-/// Hook for embedding custom behavior into the lifecycle of a Delta operation.
-///
-/// Implementors can run arbitrary async code around an operation's execution and its post-commit
-/// hook, e.g. to integrate external transaction coordination, metrics, or cleanup. Each callback
-/// receives the operation's [`LogStoreRef`] and a unique `operation_id`.
-#[async_trait]
-pub trait CustomExecuteHandler: Send + Sync {
```
ethan-tyler, line 269: Why are we breaking and removing `CustomExecuteHandler` and builder methods? I think some audience on how callers using those callbacks should migrate would be good.

ethan-tyler (commented):
eh could be worse 😉

abhishekverma-ray:
Hi @ion-elgreco, thanks for this PR. The `Committer` trait looks like a good base for Unity Catalog managed tables (#4549), which I'd like to help with once this lands. I've put my notes on #4549 so this review stays focused.

Two small questions about the trait, whenever you have time:

1. Should a catalog committer return `PayloadKind::Bytes` to skip writing the tmp commit file?
2. Is `OperationContext` meant only for writes, with catalog read state kept on `LogStore`?

I also opened #4781, a small change that shows a clearer "not supported yet" error for catalog-managed tables. It touches the snapshot loading code, so if it gets in your way, I'm happy to rebase it or wait.

No rush. I'll hold off on anything bigger until this is merged.

crates/lakefs/src/transaction.rs
ion-elgreco: I will add retries in the lakefs client however I think we should still mark it as closed, but not bubble up the error to the user. A failed delete here is fine, we can add a warning trace, because an old tx- branch has no effect and shouldn't error out the code, since the commit completed.

crates/core/src/operations/mod.rs
ion-elgreco: The customExecuteHandler was a funky setup to handle custom executions for lakefs but I don't think anyone else uses this api. I can add small description on how to migrate

crates/core/src/logstore/scope.rs
ion-elgreco: All write paths currently await before finish is called. In theory it could still happen if a new code path is introduced where writing is happening in the background but I wouldn't expect that to happen since writing needs to clear first before committing.

rtyler:
We just did 1.0 and here comes @ion-elgreco wanting to do a 2.0.

I have only started looking at this change at a high level but this will require a major API version bump in order to proceed.

ion-elgreco:
@rtyler I was hoping that we could have merged this before 1.0.0 ^^

PR #4712, opened 2026-09-11
chore: upgrade to the released buoyant_kernel 0.28

rtyler:
Comically enough my previous [change](3c5aed533c136f4b05fb25d3dd8097555833a41a) didn't actually update the tree to 0.28 because my Git repository replication from an internal system to GitHub for buoyant-data/delta-kernel-rs failed. I will wait until this is merged to correct that so as not to break every PR's CI.

This commit properly upgrades the tree to 0.28.0 with a couple notable changes:

* A couple `Interval*` types were added into delta-kernel-rs which I cannot find sufficient details on how those should be managed at the protocol level, so they're marked as unimplemented :unamused: 
* Checkpoint creation and other table mutation operations will fail (correctly) when the table has unsupported writer features. Our v2 checkpoint test table has `identityColumns` which isn't actually supported.


ion-elgreco:
Interval types got introduced here: https://github.com/delta-io/delta/pull/7078, as in it's still an RFC ^^

crates/core/src/kernel/models/actions.rs
```diff
@@ -621,6 +623,30 @@ impl ProtocolInner {
...
         }
+
+        // If the minWriterVersion has been set to 7 then there are additional writer features
+        // which must be applied
+        if self.min_writer_version >= 7
+            && let Some(features) = writer_features_for_version(self.min_writer_version)
+        {
+            self = self.append_writer_features(features);
+            // When setting the minWriterVersion the minReaderVersion must necessarily come up to
+            // version 2 if it's not already there
```
ion-elgreco, line 634: Why is that?

crates/core/src/kernel/models/actions.rs
```diff
@@ -1316,6 +1342,50 @@ pub(crate) mod serde_path {
...
+pub(crate) fn writer_features_for_version(
+    writer_version: i32,
+) -> Option<Vec<delta_kernel::table_features::TableFeature>> {
+    if writer_version >= 7 {
+        // For writer version 7+, include common features
+        Some(vec![
+            delta_kernel::table_features::TableFeature::ColumnMapping,
+            delta_kernel::table_features::TableFeature::Invariants,
+            delta_kernel::table_features::TableFeature::AppendOnly,
+        ])
```
ion-elgreco, line 1354: This is quite a behavior change, isn't it? How could one set a table to V7 but without column mapping now

crates/core/src/kernel/models/actions.rs
rtyler: Writer features require at least minReaderVersion 2 on the table to even exist,

crates/core/src/kernel/models/actions.rs
rtyler: This was based on an incorrect reading of the protocol document. Basically all tables v6+ must support column mapping but the _table features_ themselves don't necessarily need to exist

ion-elgreco: Ah oops! Did they update the protocol to make it more clear?

ion-elgreco (approved):

PR #4713, opened 2026-09-13
feat(python): support RemoveAction in create_write_transaction

### LINKED ISSUE #4480: [Feature]: Support committing AddAction and RemoveAction together in Python

seokjin0414:
# Description

`DeltaTable.create_write_transaction` now accepts `RemoveAction`s alongside `AddAction`s, so a caller that rewrites files outside of delta-rs (custom compaction, external file rewrite) can commit the new files and the tombstones of the replaced files in one atomic transaction.

- `RemoveAction` is a pyclass exposed through `deltalake._internal` and re-exported from `deltalake.transaction`, following the `Transaction` pattern.
- The binding extracts the list through a `#[derive(FromPyObject)] enum PyFileAction { Remove, Add }`, so there is no `isinstance` splitting on the Python side and anything that is neither raises `TypeError`.
- `deletionTimestamp` is required and `extendedFileMetadata` is derived from whether `size` and `partition_values` were supplied, so the written tombstones are protocol-conformant.
- Adding and removing the same path in one transaction is rejected with `ValueError`. The protocol only requires the reconciled add/remove sets to be disjoint and does not define a precedence inside a single commit (delta-kernel resolves it first-seen-wins, delta-spark's in-memory replay is last-wins within a version), so rather than pinning an engine-specific winner the ambiguous input is refused. A file rewrite never needs to add and remove the same path anyway.
- `RemoveAction`s are rejected in `mode="overwrite"`, since overwrite already derives its removes from the partition filters.
- The commit still goes through `CommitBuilder`, so conflict checking is unchanged and `delta.appendOnly` tables keep rejecting removes with `data_change=True` while allowing `data_change=False` (compaction).

This picks up #4480 and folds in the review feedback from #4599 (`PyFileAction` enum, pyclass exposure through `_internal` / `transaction`).

# Related Issue(s)

- closes #4480

# Documentation

`docs/api/transaction.md` now lists `RemoveAction`, and `create_write_transaction` has a docstring describing the remove semantics.


ion-elgreco (approved):
Thanks! 

A potential follow up is to also make AddAction a PyClass in pyo3 layer instead of dataclass in python

PR #4715, opened 2026-09-14
perf: improve efficiency of batch partition splitting and preserve input order

adamreeve:
Splitting record batches by partition values currently sorts full arrays of partition column values. But the cardinality of partition values is usually much lower than the number of rows in a batch being written, so this is quite inefficient. Instead, group the row indices into batches based on their partition values. 

The previous approach could also result in writing data out of order compared to the input batch, due to using an unstable sort. Delta-rs generally doesn't guarantee that write order is preserved and Delta tables don't have any concept of data ordering, but keeping the order comes for free with this performance improvement and it's a useful property for us (allows us to often avoid sorting on reads).

Benchmark results on my machine from the RecordBatchWriter benchmarks that write partitioned data, including extra 1024 partition cases to stress test this more:

| Benchmark | Time before | Time after | Speed change | Max alloc before | Max alloc after |
| --- | --- | --- | --- | --- | --- |
| rbw/1M/part32/few-large | 123 ms | 104 ms | 18.9% | 22 MB | 21 MB |
| rbw/1M/part32/many-small | 136 ms | 121 ms | 12.8% | 21 MB | 20 MB |
| rbw/10M/part32/few-large | 1.18 s | 854 ms | 38.5% | 92 MB | 67 MB |
| rbw/10M/part32/many-small | 1.15 s | 990 ms | 16.5% | 31 MB | 30 MB |
| rbw/1M/part1024/few-large | 292 ms | 258 ms | 13.3% | 148 MB | 145 MB |
| rbw/1M/part1024/many-small | 541 ms | 531 ms | 1.9% | 147 MB | 146 MB |
| rbw/10M/part1024/few-large | 1.76 s | 1.29 s | 36.8% | 420 MB | 395 MB |
| rbw/10M/part1024/many-small | 5.2 s | 4.84 s | 7.4% | 347 MB | 347 MB |

Credit goes to @toms972 for originally implementing this and I've made some tweaks.

A small amount of AI has been used, mostly for writing new unit tests. The non-test changes started out AI generated but have been nearly all rewritten.

The first commit moves this code out to its own module as it's not only used by the RecordBatchWriter and there were quite a few test cases for it. It might help to look at the commits separately when reviewing this to follow what has changed.


crates/core/src/writer/partition_split.rs
```diff
@@ -0,0 +1,521 @@
...
+        .convert_columns(arrays)
+        .map_err(|e| DeltaTableError::generic(format!("failed to convert columns: {e}")))?;
+    let row_count = u32::try_from(rows.num_rows()).map_err(|_| {
+        DeltaTableError::generic(format!(
+            "cannot partition a batch of {} rows: row indices are u32",
+            rows.num_rows()
+        ))
+    })?;
+
+    let mut rows_by_key: HashMap<Row<'_>, Vec<u32>> = HashMap::new();
```
ion-elgreco: I ran the bench with foldhash as the hasher and this speeded it up as well by 8-13% 

```suggestion
    let mut rows_by_key: HashMap<Row<'_>, Vec<u32>, foldhash::fast::RandomState> =
        HashMap::default();
```

ion-elgreco (commented):
Nice improvement @toms972 @adamreeve! Thanks for structuring the commits, that made reviewing it easier :)

crates/core/src/writer/partition_split.rs
adamreeve: Nice, I've added that change and found it also gave a similar speed up on my machine. I had to add foldhash as a new dependency but it was already pulled in transitively via hashbrown.

ion-elgreco (approved):

Issue #4183, opened 2026-02-09
[Bug]: nullable=True is not respected if there is a constraint on the field


LoicRaillon:
### What happened?

deltalake = "1.4.2"
polars = "1.38.0"

```python
import datetime as dt
from pathlib import Path

import deltalake as dl
import polars as pl


folder = Path("mre/weather")
folder.mkdir(parents=True, exist_ok=True)


table = dl.DeltaTable.create(
    table_uri=folder,
    schema=dl.Schema(
        [
            dl.Field("timestamp", dl.schema.PrimitiveType("timestamp"), nullable=False),
            dl.Field("irradiance", dl.schema.PrimitiveType("float"), nullable=True),
        ]
    ),
    mode="ignore",
    name="weather",
)

table.alter.add_constraint({"irradiance_ge_0": "irradiance >= 0"})

df = pl.DataFrame(
    data={
        "timestamp": [
            dt.datetime(2023, 3, 7, tzinfo=dt.timezone.utc),
            dt.datetime(2024, 3, 6, tzinfo=dt.timezone.utc),
        ],
        "irradiance": [
            200.0,
            None,
        ],
    },
    schema={
        "timestamp": pl.Datetime(time_zone="UTC"),
        "irradiance": pl.Float32,
    },
)

df.write_delta(target=folder, mode="append")
```

```bash
DeltaError: Generic DeltaTable error: External error: Invalid data found: 1 rows failed validation check.
Preview of invalid data:

+-----------+------------+
| timestamp | irradiance |
+-----------+------------+
+-----------+------------+
```

### Expected behavior

The nullable field with constraints must accept null values.

### Operating System

Linux

### Binding

Python

### Bindings Version

3.11

### Relevant logs

```shell

```

ion-elgreco:
Is this a bug tho? The protocol is unclear on what to do with this case. What is delta-spark behavior here?

@hntd187 do you know?

rtyler:
I think the check constraint should not allow the second row mentioned. The protocol is not going to help us here unfortunately since it states that the constraint must be "SQL" but doesn't specify what flavor and heavily implies Spark SQL at the time of the writing :smile:

In this case, `NULL >= 0` is going to cause the check constraint fail the validation.

That said, there _is_ a bug here in that the example output shared shows no rows failed validation, when clearly one of them should be present in the little ascii table

ernestprovo23:
hey rtyler — picking this up but want to lock in the intended semantics
[...]

PR #4721, opened 2026-09-14
fix: include null constraint violations in invalid data preview

mahesh-desu:
# Description

When a Delta **CHECK constraint** expression evaluates to `NULL` (e.g. `"high price" >= 0` against a `NULL` value), the write correctly fails with `N rows failed validation check`, but the `Preview of invalid data` table rendered with **zero rows** — the offending NULL row was counted but never shown.

Root cause: `DataValidationStream::poll_next` counted `Some(false) | None` as invalid (`crates/core/src/delta_datafusion/data_validation.rs:741`), but built the preview with `arrow::compute::not(mask)`. Since `NOT(NULL) = NULL` and `filter_record_batch` treats NULL mask entries as false, rows whose check result was NULL were dropped from the preview.

Fix: build a single `BooleanArray` invalid mask (`true` for `Some(false) | None`) and use it for both `true_count()` and `filter_record_batch`, so the count and the preview can no longer disagree.

Semantics note: this PR **keeps fail-on-NULL** behavior (a check expression evaluating to `NULL` is a violation), per maintainer direction in #4183 from @rtyler and @hntd187. Only the preview rendering is fixed; the error message wording and the 5-row preview cap are unchanged.

# Related Issue(s)

- Refs #4183

# Documentation

- No user-facing documentation changes: the error message format is unchanged; the preview now simply contains the rows it always claimed to show.

# Testing

- `cargo test -p deltalake-core --features datafusion --lib data_validation::tests` — **51/51 pass** (includes `test_validation_detailed_error_message`, which asserts exact preview lines for non-NULL violations).
- New Rust unit test `test_validation_check_constraint_includes_null_row_in_preview`: fails before the fix (preview missing the NULL row), passes after.
- New Python end-to-end test `test_constraint_null_row_preview` reproduces the issue's scenario through the bindings (constraint `'"high price" >= 0'` + append with a `NULL` value) and asserts the error preview contains the offending row: `pytest tests/test_constraint.py` — **7/7 pass**.
- `pytest tests/test_alter.py` — **75/75 pass**.
- `cargo fmt --check` clean; `cargo clippy -p deltalake-core --features datafusion` clean for the changed file; `ruff format`/`ruff check` clean. All commits DCO-signed.


python/tests/test_constraint.py
```diff
@@ -158,3 +158,30 @@ def test_add_multiple_constraint(tmp_path, sample_table: Table):
...
+    dt.alter.add_constraint({"check_price": '"high price" >= 0'})
+
+    invalid = Table.from_pydict(
+        {
+            "id": Array(["null-row"], DataType.string()),
+            "high price": Array([None], DataType.int64()),
+        },
+    )
+
+    with pytest.raises(DeltaError) as exc_info:
```
ion-elgreco, line 190: Small nit, when doing test assertions I prefer the assert is on the full message. Partial in asserts can cause it to drift and hides the exact returned error.

ion-elgreco (changes requested):
Lgtm! Can you resolve the small nit regarding the test, then we can merge!

python/tests/test_constraint.py
mahesh-desu: Fixed in 41986a5: the assertion now compares the full error message (`str(exc_info.value) == expected`) instead of partial `in` checks.

ion-elgreco (approved):
Thanks!

mahesh-desu:
Hi @ion-elgreco — could you please approve and run the workflows for the latest push? I pushed an empty commit (6720e60) to retrigger CI after the flaky Windows failure (test_concurrency_local PermissionDenied on _delta_log, unrelated to this change — Ubuntu was green with 155 passed and got cancelled via fail-fast). All pull_request workflows are now sitting at action_required with 0 jobs. Thanks!

Issue #4613, opened 2026-07-23
[Bug]: RESTORE silently fails to revert deletion-vector changes


isaacparker0:
### What happened?

Restoring a table to a version before a deletion-vector DELETE does not bring the deleted rows back. The RESTORE commit succeeds but contains no add/remove actions, and the table still reads the post-delete data.

This appears to be a regression in >=0.31.0:

-   0.28.1 through 0.30.2: restore on a DV table fails loudly with `Transaction { source: UnsupportedReaderFeatures([DeletionVectors]) }` (`UnsupportedTableFeatures` on 0.30).
-   0.31.0 through current `main`: restore succeeds silently without reverting DV changes.

### Expected behavior

Restore should revert the deletion-vector change or fail loudly like <=0.30 did, rather than silently succeeding with unrestored data.

### Operating System

Linux

### Binding

Rust

### Bindings Version

0.32.4

### Steps to reproduce

Using the`crates/test/tests/data/table-with-dv-small` fixture in this repo, whose history is:

- version 0 writes 10 rows (`value` 0–9)
-  version 1 deletes values 0 and 9 via a deletion vector

```rust
let table = deltalake::open_table(table_url).await?;

let (table, stream) = DeltaOps(table).load().await?;
// Reads 8 rows (values 1..=8), correct.

let (table, metrics) = DeltaOps(table).restore().with_version_to_restore(0).await?;
// RestoreMetrics { num_removed_file: 0, num_restored_file: 0 }

let (_table, stream) = DeltaOps(table).load().await?;
// still reads 8 rows, should be 10.
```

### Relevant logs

```shell

```

mahesh-desu:
take

mahesh-desu:
Update: this no longer reproduces on current `main`. Using the in-repo `crates/test/tests/data/table-with-dv-small` fixture (v1 deletes rows 0 and 9 via a deletion vector):

- restore to version 0 now reports `num_restored_file: 1, num_removed_file: 1` and reads return all 10 rows (previously `0/0` and 8 rows)
- restore by datetime behaves the same

Looks fixed by the Snapshot-based restore planning changes (`d1039f9`, with DV-identity coverage in `978f2bf`). I'll open a small PR adding fixture-based regression tests for both restore paths so this can't silently regress again — good to close this once that lands?

mahesh-desu:
PR up: https://github.com/delta-io/delta-rs/pull/4722 — fixture-based regression tests for both restore paths.

[...]

PR #4722, opened 2026-09-14
test: cover restore of deletion-vector deletions

mahesh-desu:
# Description

Issue #4613 reported that `RESTORE` silently failed to revert a deletion-vector (DV) `DELETE`: the restore commit succeeded but contained no add/remove actions, so the table still read the post-delete rows. The behavior is already fixed on `main` (Snapshot-based restore planning, `d1039f9`, with DV-identity coverage in `978f2bf`), but nothing pinned the user-visible behavior end-to-end.

This PR adds fixture-based regression tests that restore across the in-repo `table-with-dv-small` fixture (its v1 deletes rows 0 and 9 via a UUID-relative deletion vector):

- `test_restore_reverts_deletion_vector_delete` — restore by version
- `test_restore_by_datetime_reverts_deletion_vector_delete` — restore by datetime

Both assert `num_restored_file == 1` / `num_removed_file == 1`, that reads return all 10 rows, and that a freshly opened handle sees the restored data too. A restore implementation that ignores DV identity between versions plans zero file changes (`0/0` metrics, 8 rows) — the reported bug — and fails these tests. Tests-only change; no production code touched.

# Related Issue(s)

- Refs #4613

# Documentation

- None.

# Testing

- `cargo test -p deltalake-core --features datafusion --lib restore::tests` — **6 passed** (4 pre-existing + 2 new).
- `cargo fmt --check` clean; `cargo clippy -p deltalake-core --features datafusion --lib` clean for the changed file.
- Tests are gated `#[cfg(feature = "datafusion")]` (DV reads require the DataFusion engine; the legacy pyarrow reader does not support `deletionVectors`).
- All commits DCO-signed.


mahesh-desu:
Hi @roeap @rtyler @hntd187 @ion-elgreco  — could you please check this PR and approve if it looks good? Tests-only regression coverage for RESTORE with deletion vectors (refs #4613). Thanks!

PR #4723, opened 2026-09-14
test: cover void type support in schema conversion and reads

mahesh-desu:
# Description

Adds regression coverage for `void` data-type handling in the Python bindings. No production code changes.

`DeltaTable.schema().to_pyarrow()` used to raise `Schema error: Invalid data type for Arrow: void` for tables containing void columns (#1947). The blocker was upstream: the kernel now supports void (delta-io/delta-kernel-rs#1858, merged 2026-06-04):

- `void` parses in Delta metadata schemas and converts to Arrow `Null`
- void columns are stripped from the physical Parquet write schema, with write-time validation for placements that cannot be written (void inside Array/Map, all-void structs, all-void tables)
- reads keep void visible in the logical schema and synthesize it as all-null

delta-rs inherits this behavior, so this PR locks it in with tests:

- `python/tests/test_schema.py`: `void` primitive/field/schema Arrow conversion, `from_arrow` round-trips, `to_json`, `pa.schema(...)` interop, and nested `struct<a: integer, v: void>` conversion.
- `python/tests/test_table_read.py`: writes a table with a `pa.null()` column, asserts the metadata schema preserves `type: "void"`, that `schema().to_arrow()` maps it to `pa.null()` (the original failure), and that `to_pyarrow_table()` / `to_pyarrow_dataset()` read it back as an all-null column.

Note: the `Schema.to_pyarrow()` from the original report is now `Schema.to_arrow()`.

The earlier PRs #4240 (draft) and #4390 are superseded — they were written before kernel void support landed and work around its absence.

# Related Issue(s)

- related delta-io/delta-rs#1947 — regression coverage for the reported failure; the behavior itself is provided by the kernel upgrade already on `main`.

# Documentation

- Delta protocol primitive types: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types
- Kernel void support: https://github.com/delta-io/delta-kernel-rs/pull/1858


ion-elgreco (approved):
Thanks!

mahesh-desu:
Friendly ping @ion-elgreco @rtyler — CI here is red due to the upstream smithy-rs breakage (aws-smithy-types 1.7.0 vs aws-smithy-json 0.63.0, smithy-lang/smithy-rs#4853, tracked in #4726). All jobs fail in `make develop`/clippy before any tests run, so it's unrelated to this PR's changes.

Would you be able to land a fix on main (e.g. pin `aws-smithy-types = "=1.6.3"`) and re-run the checks here so this can merge? I verified locally that the pin resolves `aws-smithy-types 1.6.3` + `aws-smithy-json 0.63.0` and `cargo check -p deltalake-aws -p deltalake-catalog-glue` passes with it. Happy to open that as a separate PR if preferred.

ion-elgreco:
@mahesh-desu if you could create a fix pr for that, that woud be helpful!

mahesh-desu:
@ion-elgreco #4728 (the `aws-smithy-types` pin) is now merged, so the CI failures here should be resolved. Could you please re-run the failed checks so this can merge? Happy to rebase onto latest `main` first if you prefer.

PR #4725, opened 2026-09-15
feat(python): allow passing datafusion session config in query builder

### LINKED ISSUE #3516: Allow passing `DeltaSessionConfig` in `PyQueryBuilder`
# Description
```python
dt = DeltaTable("./my_table")
qb = QueryBuilder().register("tbl", dt)
```

When initialising the `QueryBuilder`, allow passing [`SessionConfig`](https://github.com/apache/datafusion-python/blob/0cc9b0a513e784597bc6e35f883fefa9e2d3210b/python/datafusion/context.py#L81).

[Currently, the default configurations are used.](https://github.com/delta-io/delta-rs/blob/2ffaacc9e1d892aa2a9a0d773efb416d9adb8bf5/python/src/query.rs#L28)

**Use Case**
I want to be able to set [DataFusion configurations](https://datafusion.apache.org/user-guide/configs.html) mentioned here when initialising the `QueryBuilder`.

**Related Issue(s)**

ArulJerald:
# Description

`QueryBuilder` always built its DataFusion `SessionContext` from the defaults deltalake tunes for Delta tables, so python users had no way to set datafusion options such as `datafusion.execution.batch_size`.

this adds an optional `session_config` argument:

```py
QueryBuilder({"datafusion.execution.batch_size": "1024"})
```

keys are the dotted names from the datafusion config guide. a dict of strings matches the convention already used for config-ish params elsewhere in the python bindings (e.g. `storage_options`). passing nothing behaves exactly as before.

two behaviours worth calling out:

- values are applied as **overrides on top of** the delta defaults rather than replacing them, so any setting not listed keeps its tuned value
- unknown or malformed keys raise a `DeltaError` instead of being silently ignored

**why the layering lives in the core crate:** it is implemented as `DeltaSessionContext::new_with_overrides` in `crates/core`, and `python/src/query.rs` just calls it. `DeltaSessionContext`'s session construction (`new_with_config_and_runtime`) is private, so keeping this python-side would have meant duplicating the planner and runtime wiring — a second copy that drifts the next time the core defaults change. happy to move it into the bindings instead if you would prefer a smaller core surface.

**testing:** `python/tests/test_query.py` is new — there were no `QueryBuilder` tests before. the tests read back `information_schema.df_settings` to confirm a setting actually reached the datafusion session. the interesting assertion checks a key i did *not* override: `datafusion.sql_parser.enable_ident_normalization` must still read `false`. deltalake sets it `false` while datafusion's own default is `true`, so it can only pass if the overrides layered on top of the delta defaults instead of replacing them. two rust unit tests in `crates/core/src/delta_datafusion/session.rs` cover the same layering property and the unknown-key error.

# Related Issue(s)

- closes #3516

# Documentation

- [datafusion configuration settings](https://datafusion.apache.org/user-guide/configs.html) — the list of keys accepted by `session_config`


crates/core/src/delta_datafusion/session.rs
```diff
@@ -356,6 +357,18 @@ impl DeltaSessionContext {
         Self::new_with_config_and_runtime(config, runtime_env)
     }
 
+    /// Create a DeltaSessionContext with the tuned defaults, applying `overrides` on top.
+    pub fn new_with_overrides(overrides: &HashMap<String, String>) -> DeltaResult<Self> {
```
ion-elgreco, line 361:
```
session_overrides
```

ion-elgreco (changes requested):
Looks good! Could you rename the param for clarity! Then we are good to go

crates/core/src/delta_datafusion/session.rs
ArulJerald: Sure ! Will do that. Thank you for your comment !!

ArulJerald:
@ion-elgreco  - I have renamed the fn name as you mentioned

ion-elgreco (approved):

ion-elgreco:
Thanks @ArulJerald !

Issue #4726, opened 2026-09-15
[Bug]: deltalake-aws does not compile due to DocumentObject mismatched types


rtyler:
### What happened?

Hey thanks for stopping by! If you saw a build issue with the `deltalake` or `deltalake-aws` crates, it is due to a semantic version issue upstream in the [AWS SDK for Rust](https://github.com/smithy-lang/smithy-rs/issues/4853) dependencies.

We know about the issue and you can (like and) Subscribe to this issue for updates.

### Expected behavior

`cargo build` to mostly work 😄 

### Operating System

None

### Binding

Rust

### Bindings Version

0.32.4

### Steps to reproduce

```
cargo new --bin awesomething && cd awesomething && cargo add deltalake --features s3 && cargo build
```

### Relevant logs

```shell

```

rtyler:
The upstream issue was resolved by AWS yanking the broken crates

PR #4728, opened 2026-09-15
fix: pin aws-smithy-types to 1.6.3 to unblock builds

mahesh-desu:
# Description

aws-smithy-types 1.7.0 made breaking changes to `Document` that are
incompatible with the released aws-smithy-json 0.63.0 pulled in via
aws-config. Since aws-smithy-json 0.63.0 declares aws-smithy-types ^1.6.1,
Cargo unifies on a combination that does not compile, breaking all builds:

    error[E0308]: mismatched types
      aws-smithy-json-0.63.0/src/codec/deserializer.rs:707:37
      expected `DocumentObject`, found `HashMap<String, Document>`
    error[E0004]: non-exhaustive patterns: `&_` not covered
      aws-smithy-json-0.63.0/src/serialize.rs:36:15

This pins `aws-smithy-types = "=1.6.3"` in the workspace and adds it as a
direct dependency of deltalake-aws and deltalake-catalog-glue (the crates
that pull in aws-config), with cargo-machete ignores. Cargo then backtracks
the other AWS crates to a compilable set while aws-smithy-json 0.63.0 stays
on aws-smithy-types 1.6.3.

Verification:

- `cargo tree -i aws-smithy-types` shows only 1.6.3, no 1.7.0
- `cargo check -p deltalake-aws -p deltalake-catalog-glue` passes (previously
  failed while compiling aws-smithy-json 0.63.0)

The pin can be dropped once 1.7.0 is yanked or compatible AWS SDK releases
are published.

# Related Issue(s)

- #4726
- smithy-lang/smithy-rs#4853

# Documentation

- https://github.com/smithy-lang/smithy-rs/issues/4853


Cargo.toml
```diff
@@ -71,6 +71,9 @@ futures = { version = "0.3" }
 tokio = { version = "1" }
 typed-builder = { version = "0.23.0" }
 
+# aws sdk
+aws-smithy-types = { version = "=1.6.3" }
```
rtyler, line 75: Thanks for submitting the PR, this pin should only exist in `crates/aws/Cargo.toml` since it's not needed at the workspace level. The diff I mentioned [here](https://github.com/smithy-lang/smithy-rs/issues/4853#issuecomment-5673623450) was against that file 😄

rtyler (dismissed):

Cargo.toml
mahesh-desu: Fixed in 934cc2e: the pin now lives only in `crates/aws/Cargo.toml` as `aws-smithy-types = "=1.6.3"`, with the workspace-level entry and the catalog-glue addition removed.

Verified `cargo tree` shows only `aws-smithy-types 1.6.3` across the workspace (including `-p deltalake-catalog-glue`) and `cargo check -p deltalake-aws -p deltalake-catalog-glue` passes. Note: the types pin alone keeps `aws-smithy-runtime-api` at 1.16.0, so I left it unpinned — say the word if you want the explicit `=1.16.0` too.

ion-elgreco (approved):

PR #4729, opened 2026-09-15
fix: preserve nested schema overrides in parquet read schema

### LINKED ISSUE #4678: [Bug]: DeltaScanNext widens nested schema overrides for Parquet reads

mahesh-desu:
# Description

`DeltaScanNext` lost nested leaf narrowing from `DeltaScanConfig::schema`
overrides when building `parquet_read_schema`: it was derived from the wide
kernel physical schema, so the Parquet reader read every sibling leaf under
each selected struct root.

When an override is configured, the read schema is now narrowed to the
override's selected leaves first. The override carries logical names while
the physical schema carries physical (column-mapped) names, so selection
paths are translated through the kernel logical schema, and the usual
physical type conversions (view types, partition dictionaries) still apply.
Nesting that cannot be narrowed safely is passed through unchanged.

Verified locally: new `test_parquet_read_schema_preserves_nested_schema_override`
fails before and passes after the fix; all 111 scan tests and all 357
`delta_datafusion` tests pass; clippy and rustfmt clean.

Note: repo-wide CI builds are currently blocked by #4726 until #4728 lands;
this change is confined to `deltalake-core` scan planning.

# Related Issue(s)

- Closes #4678

# Documentation

- https://github.com/delta-io/delta-rs/issues/4678


ion-elgreco:
@ethan-tyler could you take a look at this PR since you have done a lot of changes in this area? :)

ethan-tyler:
thanks @mahesh-desu for taking a stab at this - I unfortunately need to close this one as it's duplicative of #4685. To help as a contributor I did see starting at line 391 - iterating over `selection_fields` puts the reader’s children in override order, while kernel’s transforms still expect the original order.

PR #4730, opened 2026-09-15
fix: chunk metadata-only scan batches within session batch size

### LINKED ISSUE #4727: [Bug]: Metadata only scans can overflow Arrow byte array offset for partition columns
### What happened?

We have a Delta table with some largish string Partition values and Parquet files with a largeish number of rows. The product of the partition value length * the number of rows in one file is larger than the maximum value representable by an int32.

Various queries on this table can cause an error like:
```
thread 'tokio-rt-worker' (674362) panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-array-59.2.0/src/builder/generic_bytes_builder.rs:87:57:
byte array offset overflow
Query failed:
External error: Join Error
caused by
External error: task 590 panicked with message "byte array offset overflow"
```

It looks like the queries that fail are ones that only need table metadata and so use `DeltaScanMetaExec`, which generates Arrow RecordBatches with the number of rows equal to the number of rows in a Parquet file.

Updating this to generate multiple smaller batches per file (respecting Datafusion's `batch_size` in the session config) might be one way to fix it. In addition, the partition values could probably be de-duplicated when creating the `Utf8View` array, so equal values point to the same bytes.

I can implement a fix for this.

### Expected behavior

Don't panic.

### Operating System

Linux

### Binding

Python

### Bindings Version

1.6.3

### Steps to reproduce

```python
from pathlib import Path
import tempfile

from deltalake import DeltaTable, write_deltalake, QueryBuilder
import pyarrow as pa


def input_data_stream(num_rows, group_val):
    schema = pa.schema([("group", pa.string()), ("i", pa.int64())])

    batch_size = min(1_000_000, num_rows)
    assert num_rows % batch_size == 0

    def batches():
        batch = pa.record_batch(
            [
                pa.array([group_val] * batch_size, pa.string()),
                pa.array(range(batch_size), pa.int64()),
            ],
            schema=schema,
        )
        for _ in range(num_rows // batch_size):
            yield batch

    return pa.RecordBatchReader.from_batches(schema, batches())

 
with tempfile.TemporaryDirectory() as temp_dir:
    table_path = Path(temp_dir) / "delta_table"

    num_rows = 43_000_000
    partition_value_length = 50

    # The error is triggered if the group values length overflows int32
    assert num_rows * partition_value_length > (2 ** 31 - 1)

    group_val = "x" * partition_value_length

    initial_data = input_data_stream(num_rows, group_val)
    write_deltalake(
        table_path,
        initial_data,
        partition_by=["group"],
        mode="overwrite",
        target_file_size=1024*1024*1024)

    # Query can cause a crash
    try:
        print("\nQuerying table")
        table = DeltaTable(table_path)
        qb = QueryBuilder().register("t", table)
        # Query that has to read data, this works:
        result_arrow = qb.execute("SELECT group, MIN(i) FROM t GROUP BY group").read_all()
        print("First query succeeded")
        # Metadata only query, this crashes:
        result_arrow = qb.execute("SELECT DISTINCT group FROM t").read_all()
        print("Second query succeeded")
    except Exception as exc:
        print("Query failed:")
        print(exc)

    # Overwrite can also cause a crash when a predicate is specified
    try:
        print("Overwriting with predicate")
        overwrite_data = input_data_stream(10, group_val)
        write_deltalake(
            table_path,
            overwrite_data,
            partition_by=["group"],
            mode="overwrite",
            predicate=f"group = '{group_val}'")
        print("Overwrite completed successfully")
    except Exception as exc:
        print("Overwrite failed:")
        print(exc)
```

### Relevant logs

```shell

```

mahesh-desu:
# Description

`DeltaScanMetaExec` synthesized one Arrow batch per file using the file's
full row count. For metadata-only queries over tables with large string
partition values, value bytes times rows can exceed the i32 Arrow offset
limit and the query panics with `byte array offset overflow`.

The stream now splits each file into bounded chunks: the session
`batch_size`, reduced when a single-row probe of the materialized bytes
shows that width would overflow. Masking (deletion vectors), transform
evaluation, and finalization run per chunk exactly as they did per file,
so rows, file-id handling, and metrics are unchanged apart from batch
boundaries.

Verified locally: new
`test_meta_scan_chunks_batches_within_session_batch_size` materializes
100 B x 25 M rows (past the i32 limit), panics without the fix with the
exact reported error, and streams bounded chunks with it. All 111 scan
tests and all 357 `delta_datafusion` tests pass; clippy and rustfmt clean.

Note: repo-wide CI builds are currently blocked by #4726 until #4728
lands; this change is confined to `deltalake-core` scan execution.

# Related Issue(s)

- Closes #4727

# Documentation

- https://github.com/delta-io/delta-rs/issues/4727


crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -493,16 +551,46 @@ impl Stream for DeltaScanMetaStream {
...
+                self.current = Some((file_id, row_count, 0, chunk_rows));
+            }
+            let Some((file_id, total_rows, offset, chunk_rows)) = self.current.clone() else {
+                unreachable!("cursor is set just above");
+            };
+            let len = chunk_rows.min(total_rows.saturating_sub(offset));
+            let batch = match self.batch_project_chunk(&file_id, offset, len) {
+                Ok(batch) => batch,
+                // Drop the file like the old whole-file path did: the error
+                // is yielded once and the stream advances.
```
adamreeve: This comment will live long after this change, so referring to the old code path could be confusing for future code readers (this is a pretty common issue with AI generated comments).
```suggestion
                // Drop the file: the error is yielded once and the stream advances.
```

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -493,16 +551,46 @@ impl Stream for DeltaScanMetaStream {
     type Item = Result<RecordBatch>;
 
     fn poll_next(mut self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let poll = if let Some((file, row_count)) = self.input.pop_front() {
-            Poll::Ready(Some(self.batch_project(file, row_count)))
-        } else {
-            Poll::Ready(None)
+        let poll = loop {
```
adamreeve: This loop doesn't actually loop, every branch breaks. Can this be refactored to remove the loop? And ideally the `unreachable` below too?

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -822,6 +910,121 @@ mod tests {
...
 
+    /// A file whose partition-value bytes times row count exceeds the i32
+    /// Arrow offset limit must stream in bounded chunks instead of
+    /// materializing one giant batch (#4727).
+    ///
+    /// NOTE: the partition value is deliberately short so the partition
+    /// directory stays within Windows path limits; the row count alone
+    /// pushes the materialized bytes past the limit.
+    #[tokio::test]
+    async fn test_meta_scan_chunks_batches_within_session_batch_size() -> TestResult {
```
adamreeve, line 946: This test takes 13 seconds to run on my machine, even though the rows are "virtual". I don't think it's necessary to have a test case that would trigger the overflow previously. This test could instead use a much smaller number of rows and configure a small batch size on the session, and check the batches are all under the configured limit.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -426,13 +419,78 @@ impl DeltaScanMetaStream {
...
+    ) -> Result<RecordBatch> {
+        let _timer = self.baseline_metrics.elapsed_compute().timer();
+
+        let batch = RecordBatch::try_new_with_options(
+            EMPTY_SCHEMA.clone(),
+            vec![],
+            &RecordBatchOptions::new().with_row_count(Some(len)),
+        )?;
+
+        let batch = if let Some(selection) = self.selection_vectors.get(file_id) {
```
adamreeve, line 488: It would be good to have a unit test that verifies the generated batch lengths are correct when selection vectors are present.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -426,13 +419,78 @@ impl DeltaScanMetaStream {
...
+                .map_err(|e| DataFusionError::External(Box::new(e)))
+        } else {
+            Ok(batch)
+        }
+    }
+
+    /// Rows per emitted batch for one file: the session `batch_size`, reduced
+    /// when a single row's materialized bytes would overflow Arrow's 32-bit
+    /// offsets at that width. Partition values repeat per row, so probing a
+    /// single row measures the steady-state per-row cost exactly.
```
adamreeve: Reducing the batch size by probing the bytes per row feels a bit unnecessarily complex to me. The default batch size is 8,192, so rows would need to be > 262,143 bytes to cause an overflow. But I guess someone might decide to configure a much larger batch size, so maybe this is necessary 🤷

adamreeve (commented):
Thanks for taking on this issue @mahesh-desu! I've left some suggestions

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
mahesh-desu: Applied as suggested in 3c4c323 — the referenced whole-file path is gone in this diff, so keeping the reference would rot. The operative behavior (yield once, advance) is preserved in the new wording.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
mahesh-desu: Loop removed in 3c4c323 via a labeled block — verified every path through the old body breaks (4 break sites, no continue), so it never iterated, and all exits still funnel through record_poll. Kept the unreachable: the let-else needs a diverging else and it documents the cursor-is-Some invariant; dropping it would need a larger restructure for no behavior gain.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
mahesh-desu: Shrunk in 3c4c323: 10k virtual rows with batch size 100 (batch_size is read at execute time from the task context, so the scan still uses the default session). The two integration tests now finish in under a second combined, and the small case still fails pre-fix via a single over-size batch. One coverage note: the small version guards the chunking invariant but can no longer reproduce the original i32 overflow, so I extracted the byte-cap math into capped_chunk_rows with a unit test pinning the boundary (2048 * 262144 == cap < i32::MAX < 8192 * 262144).

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
mahesh-desu: Added in 3c4c323 as test_meta_scan_chunk_lengths_with_selection_vectors: real 10-row file masked with a short 8-entry keep-mask, executed at batch size 3, asserting exact per-chunk lengths [2,3,2,1] and kept total 8. Covers the offset/end slicing plus trailing-pad interplay across chunk boundaries, which the existing apply_selection_vector unit tests do not reach. Note the repro partition value is uniform, so it pins lengths/totals only; which-rows correctness stays with the filter unit tests.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
mahesh-desu: Keeping the probing, with reasoning: the 256KB/row threshold holds only at the default batch size of 8192, but batch_size is user-configurable — at 1M rows per batch the overflow threshold drops to ~2KB/row, reachable with wide partition columns, and the failure mode is an Arrow panic. Cost is one 1-row probe per file, and the per-row cost is steady-state exact since partition values repeat per row. The 512MiB cap also keeps a 4x margin under the 2GiB i32 limit for downstream consumers. The byte-cap math is now unit-tested (test_capped_chunk_rows) so the branch is pinned cheaply. If the concern is the doc wording rather than the mechanism, happy to simplify the comment — let me know.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
```diff
@@ -822,6 +919,208 @@ mod tests {
...
+                DataType::LargeUtf8 => {
+                    let arr = parts.as_string::<i64>();
+                    assert_eq!(arr.value(0), value);
+                    assert_eq!(arr.value(batch.num_rows() - 1), value);
+                }
+                DataType::Utf8View => {
+                    let arr = parts.as_string_view();
+                    assert_eq!(arr.value(0), value);
+                    assert_eq!(arr.value(batch.num_rows() - 1), value);
+                }
```
adamreeve: The coverage report showed that these lines aren't hit, the partition is always a dictionary column. So these can be removed.

adamreeve (dismissed):

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
mahesh-desu: Kept the probe: at the 8192 default the overflow needs ~256KB per row, but batch_size is user-configurable and the per-row budget shrinks in step with it (down to ~2KB at 1M rows per batch), so the Arrow panic stays reachable. Reworked the chunk_rows_for_file doc in 711f796 to say the cap only binds for wide values at large batch sizes.

crates/core/src/delta_datafusion/table_provider/next/scan/exec_meta.rs
mahesh-desu: Removed in 711f796: the test now downcasts to the dictionary array directly instead of carrying non-dictionary fallback arms.

adamreeve (approved):

PR #4734, opened 2026-09-17
fix(python): treat naive load_as_version datetime as UTC

jx2lee:
# Description
Fixes `DeltaTable.load_as_version()` so timezone-naive `datetime` inputs are interpreted as UTC, matching existing API docstring.

Previously, naive datetimes were converted with `astimezone(timezone.utc)`, which treats them as local time and can select different table versions depending on host timezone. This now attaches UTC with `replace(tzinfo=timezone.utc)`.

Adds a regression test that checks the same naive datetime under `UTC`, `Asia/Seoul`, and `America/Los_Angeles`.

# Related Issue(s)

# Documentation

N/A. This fixes behavior to match existing `load_as_version()` docstring


ion-elgreco (dismissed):

jx2lee:
@ion-elgreco Since Ruff formatting failed, I re-pushed the changes. Could you please re-review?

ion-elgreco (approved):

PR #4737, opened 2026-09-17
feat(python): expose Delta table property constants

### LINKED ISSUE #2074: ConfigKey Enum Python
# Description
Add equivalent of ConfigurationKey in Python as helper to set config settings

**Use Case**

**Related Issue(s)**

jx2lee:
# Description
Adds `TableProperty` enum containing well-known Delta table property and exports it allow users to configure tables without repeating string keys.

```py
write_deltalake(
      path,
      data,
      configuration={TableProperty.APPEND_ONLY: "true"},
)
```

# Related Issue(s)
- closes #2074

# Documentation
- Added `TableProperty` to the Python writer API reference.
- Updated the `configuration` parameter documentation to reference `TableProperty`.


python/deltalake/table.py
```diff
@@ -178,6 +179,48 @@ def __str__(self) -> str:
         )
 
 
+class TableProperty(str, Enum):
```
ion-elgreco: Can you change it into?

```suggestion
class TableProperty(StrEnum):
```

jx2lee: @ion-elgreco  Thanks for suggestion !  I'd to use `StrEnum`, but since this python sdk supports Python 3.10+ (`requires-python = ">=3.10"` in [pyproject.toml](https://github.com/delta-io/delta-rs/blob/main/python/pyproject.toml#L10), using it directly would break 3.10 users.

Keep the current mixinㅡas-is make sens, wdyt?

ion-elgreco: 3.10 is almost EoL (end of october), so its fine if you change the min version to 3.11 and the abi3 to 3.11+

jx2lee: Sure. Fixed it and modified min py version to 3.11. also modified workflows.

rtyler (dismissed):

ion-elgreco:
@jx2lee could you resolve the conflicts? Then we are good to merge

jx2lee:
> @jx2lee could you resolve the conflicts? Then we are good to merge

@ion-elgreco Sorry to late. succeeded to rebase upstream-main and pytest. Please to review, again. 🙇🏽

ion-elgreco (dismissed):

jx2lee:
@ion-elgreco @rtyler  My changes broke CI (pyspark), because not pinning pandas version in pyproject.toml. Please re-review, Thanks !

python/pyproject.toml
```diff
@@ -115,6 +114,7 @@ lakefs = ["lakefs==0.8.0"]
 pyspark = [
     "pyspark==4.0.1",
     "delta-spark",
+    "pandas<3",
```
jx2lee, line 117: Pinning pandas under v3.

ion-elgreco (approved):

jx2lee:
@ion-elgreco Could you update the required Python 3.10 status checks to their Python 3.11 equivalents? Changed workflow now reports the Python 3.11 check names, but the branch protection rule still requires the old Python 3.10 status checks. 

image

ion-elgreco:
@rtyler can only edit this

PR #4738, opened 2026-09-17
fix: remove DeltaTableConfig

### LINKED ISSUE #2134: Consistently use `DeltaTableConfig` to pass around execution configuration.
# Description

Aside from the table properties, there are several configuration options that control the behaviour of delts-rs during execution. Some specific to e.g. arrow/parquet (like `write_batch_size` or `WriterProperties`) other specific to delta (like `mergeSchema` to control schema evolution behaviour). To ensure consistent treatment of these properties, we should consolidate these on a dedicated struct - `DeltaTableConfig` seems like a good candidate.

**Use Case**

* ensure all options are always honoured and provide a clear interface how and when thwse are specified.

rtyler:
This is pending the merge of #4712 and should not be reviewed until that pull request has been merged and this has been rebased.

- closes https://github.com/delta-io/delta-rs/issues/2134


hntd187:
Even though this doesn't build I'd feel compelled to approve this just because it's a massive delete.

rtyler:
> Even though this doesn't build I'd feel compelled to approve this just because it's a massive delete.

The build failures came after some rebasing that I need to correct

python/deltalake/table.py
```diff
@@ -200,9 +194,6 @@ def __init__(
         table_uri: str | Path | os.PathLike[str],
         version: int | None = None,
         storage_options: dict[str, str] | None = None,
-        without_files: bool = False,
-        log_buffer_size: int | None = None,
-        skip_stats: bool = False,
```
ion-elgreco, line 205: We cannot immediately remove these, that would break things. We should throw a deprecation warning first.

crates/core/src/kernel/snapshot/mod.rs
```diff
@@ -429,13 +416,8 @@ impl Snapshot {
...
 
-    /// Get the table config which is loaded with of the snapshot
-    pub fn load_config(&self) -> &DeltaTableConfig {
-        &self.config
-    }
-
     fn materialization_mode(&self) -> SnapshotMaterializationMode {
-        SnapshotMaterializationMode::from_require_files(self.config.require_files)
+        SnapshotMaterializationMode::Eager
     }
```
ion-elgreco, line 421: So our snapshots are not lazy anymore?

python/deltalake/table.py
rtyler: I mentioned this to @ion-elgreco in Slack, but I'm going to put a Python-level deprecation into `main` and then cut that into 1.6.5 so that this can come down after it

ion-elgreco (dismissed):

crates/core/src/kernel/snapshot/mod.rs
rtyler: @ion-elgreco I am still working on verifying or refuting my hypothesis that we're still very lazy (huehuehue) because of improvements in our kernel interoperation these days.

I would like to land this and pinky-promise I will address this concern in `a follow up before we release 1.0.0.  I am worrying about changes that are going to get lots of conflicts the longer this sits open

rtyler: @ion-elgreco  I added some commentary from my discussion with @ethan-tyler on the subject and I think this should be ready to merge. It is a good thing that you called this out and gave me the opportunity to learn a bit more here :smiling_imp: 

I think this is the appropriate default behavior (setting aside require_files=false ends up breaking a number of operations in goofy ways today). 

For users that want to perform only metadata operations the `Snapshot` construction directly should be the path, and I think some prior work we were exploring with @seokjin0414  (I believe) could be a more ergonomic way for folks that previously were trying to use `require_files=false` as a mechanism for append only writes (like oxbow was).

I think it is useful/important for our implementation to assume that a `DeltaTable` has loaded context (i.e. eager materialization) because otherwise we would need to perform a lot of changes of existing APIs to make them take mutable references in order to allow lazy materialization up and down the stack.

All that said, I would like to merge this now and release 1.0.0 :smile:

ethan-tyler (approved):

ion-elgreco (approved):

crates/core/src/kernel/snapshot/mod.rs
BoazC-MSFT: > setting aside require_files=false ends up breaking a number of operations in goofy ways today

@rtyler, can you please shed some light on which operations break today (before 1.0.0)? We have some functionality that relied on `require_files=false` and would like to assess the risk of staying vs. moving forward to 1.0.0

Thanks!

PR #4739, opened 2026-09-17
feat: support directory partitioning in convert_to_delta

### LINKED ISSUE #1828: Support non-hive partitioning schema for parquet files
# Description
Currently `DeltaTablePartition::try_from` only supports hive partitioning. If a parquet path has non-hive parititioning, `DeltaTablePartition::try_from` will throw an error. It would be nice to support some non-hive partitioning schemes such as the directory scheme like `./partitioned/2000/part-0.parquet`. User will need to provide the column name when they use this scheme. 
**Use Case**
This feature would make ConvertToDelta operation applicable to non-hive parquet tables.
**Related Issue(s)**
https://github.com/delta-io/delta-rs/issues/1041, https://github.com/delta-io/delta-rs/issues/1682, https://github.com/delta-io/delta-rs/pull/1686

akashjainn:
# Description

`ConvertToDeltaBuilder` accepted a `PartitionStrategy` but never read it: `execute` always split every directory on `=`, so a Parquet table laid out as `2020/1/part-0.parquet` could not be converted. This adds the directory strategy the issue asks for.

- `PartitionStrategy::Directory` treats every directory in a file path as a bare partition value and maps the directories in order onto the fields passed to `with_partition_schema`. `"directory"` parses from a string the same way `"hive"` does.
- `partition_schema` is now an `IndexMap` instead of a `HashMap`, so the order the caller passes the fields in is the order used for the mapping. The partition columns of the created table follow that order too, where before they came out of a `HashSet`.
- A file whose depth does not match the number of partition columns fails with a new `PartitionDepthMismatch` error naming the path, the expected depth and the found depth, instead of a generic missing-schema error.
- The value parsing (null marker, percent decoding, primitive parsing) is shared by both strategies through one `parse_partition_value` function. Hive behaviour is unchanged.
- Python: `convert_to_deltalake` accepts `partition_strategy="directory"`, the type stubs and docstring are updated, and the guard rejects anything other than `hive` or `directory`.

Tests:

- Rust: a directory-partitioned table written into a tempdir converts with the expected paths, schema, partition values and partition column order; a file with too few directories fails with the depth error; a directory strategy without a partition schema fails; `PartitionStrategy` parsing is case-insensitive and rejects unknown names. The existing hive tests pass unchanged.
- Python: a `pyarrow` dataset written with directory partitioning converts and reports the partition column and the value-only file paths; an unknown strategy raises `ValueError`.

AI assistance was used while writing this change. Every line was reviewed by me, and the Rust and Python tests listed above were run locally by me.

# Related Issue(s)

- closes #1828

# Documentation

The `convert_to_deltalake` docstring describes both strategies with an example path each.


ion-elgreco (approved):
Thankss!

PR #4740, opened 2026-09-18
fix(python): honor encoding in default column properties

jx2lee:
# Description
`default_column_properties.encoding` was accepted by the Python API but ignored when converting to Rust `WriterProperties`.

Apply the default encoding and disable dictionary encoding, matching existing per-column behavior. Adds a regression test that verifies resulting Parquet metadata.

# Related Issue(s)
- Follow-up to #3737
- Related to #3319

# Documentation
N/A


python/tests/test_writerproperties.py
```diff
@@ -131,6 +131,30 @@ def test_write_with_writerproperties_encoding(
...
+        tmp_path,
+        pa.table({"value": [1, 2, 3]}),
+        writer_properties=writer_properties
+    )
+
+    parquet_path = DeltaTable(tmp_path).file_uris()[0]
+    encodings = pq.read_metadata(parquet_path).row_group(0).column(0).encodings
+
+    assert "DELTA_BINARY_PACKED" in encodings
+    assert "RLE_DICTIONARY" not in encodings
```
ion-elgreco, line 155: Could you do exact asserts here instead of `in`?

jx2lee: @ion-elgreco Fixed it to do exact asserts

ion-elgreco (approved):

ion-elgreco:
Thankss!

PR #4743, opened 2026-09-19
feat: add opt-in sort_columns to compact()

### LINKED ISSUE #4724: [Enhancement]: Add opt-in sort columns to compact() and clarify preservedStableOrder semantics
### What happened?

`DeltaTable.optimize.compact()` changes the physical row order of Parquet files
whose rows were already sorted by a timestamp column.

The reproducer below writes monotonically increasing timestamp batches to a
Delta table partitioned by calendar month. Before compaction:

- Every input batch is sorted by `dt`.
- The timestamp ranges of consecutive batches do not overlap.
- Every active Parquet file is internally sorted by `dt`.
- Reading the active files in partition and `max.dt` order produces a globally
  sorted sequence.

After calling `table.optimize.compact()`, some compacted Parquet files are no
longer internally sorted. Consequently, the final assertion fails.

This is especially surprising because the returned compaction metrics report:

```text
preserveInsertionOrder: True
preservedStableOrder: True
plannerStrategy: preserveLocality
```

The reproducer explicitly reads each Parquet file in physical row order through
a DataFusion external table. It does not depend on the unspecified file
iteration order of a regular Delta table scan.

The example only imports the Python standard library and public APIs from
`deltalake`. It does not depend on Polars, Pandas, or PyArrow.

### Expected behavior

Physical row-order preservation is not currently documented as part of the
Delta Lake compaction contract. However, the delta-rs metrics
`preserveInsertionOrder` and `preservedStableOrder` can reasonably be
interpreted as implying a stronger ordering guarantee than they currently
provide.

I would therefore like the metric semantics to be clarified and an opt-in
lexicographic sorting option to be considered.

### Operating System

- macOS 15.6.1
- Apple Silicon (`arm64`)

### Binding

Python

### Bindings Version

Tested Python binding versions:

- `0.25.5`
- `1.6.3`

Python version: `3.14.7`

### Affected versions

I reproduced this behavior with the following Python binding versions:

| deltalake version | Result | Repetitions |
| ----------------- | ------ | ----------- |
| `0.25.5` | Final post-compaction sortedness assertion failed | 2/2 |
| `1.6.3` | Final post-compaction sortedness assertion failed | 2/2 |

For every tested version:

1. Every generated input batch was sorted.
2. Every source Parquet file was internally sorted.
3. Reading the active files in partition and `max.dt` order was globally
   sorted before compaction.
4. `DeltaTable.optimize.compact()` completed successfully.
5. The final `assert is_sorted(after_compaction)` failed.

No earlier assertion failed. This indicates that the behavior is not limited to
a recent release and can be reproduced from `0.25.5` through `1.6.3`.

All versions were tested in isolated environments on:

- Python `3.14.7`
- macOS `15.6.1`
- Apple Silicon (`arm64`)

The `QueryBuilder` result API differs between the tested releases:
`0.25.5` exposes `QueryResult.fetchall()`, while the `1.x` releases expose
`read_all()`. The reproducer contains a small compatibility adapter for this API
difference. The generated data, partitioning, file-ordering logic, compaction
operation, and sortedness assertions remain identical across all versions.

### Steps to reproduce

Install the Python binding:

```shell
python -m pip install deltalake==1.6.3
```

Run the following standalone script:

```python
import tempfile
from collections.abc import Sequence
from datetime import UTC, datetime, timedelta
from itertools import pairwise
from pathlib import Path
from typing import Any, cast

from deltalake import DeltaTable, QueryBuilder, WriterProperties, write_deltalake

DT = "dt"
YEAR_MONTH = "ym"
PARTITION_KEYS = (YEAR_MONTH,)


def materialize_query(result: Any) -> Any:  # noqa: ANN401
    """Materialize query results across the delta-rs 0.25 and 1.x APIs."""
    if hasattr(result, "read_all"):
        return result.read_all()

    batches = result.fetchall()
    return batches[0] if len(batches) == 1 else batches


def query_column_values(result: Any, column: str) -> list[datetime]:  # noqa: ANN401
    """Convert one query-result column to standard Python datetime values."""
    materialized = materialize_query(result)
    if hasattr(materialized, "column"):
        return cast("list[datetime]", materialized.column(column).to_pylist())

    values: list[datetime] = []
    for batch in materialized:
        values.extend(batch.column(column).to_pylist())
    return values


def create_batch(start: datetime, days: int) -> Any:  # noqa: ANN401
    """Create sorted timestamp rows through delta-rs' DataFusion query API."""
    timestamps = [start + timedelta(days=offset) for offset in range(days)]
    values = ", ".join(
        f"(timestamp '{timestamp:%Y-%m-%d %H:%M:%S} UTC', '{timestamp:%Y%m}')" for timestamp in timestamps
    )
    # All SQL literals come from datetimes generated inside this reproducer.
    query = f"SELECT * FROM (VALUES {values}) AS batch({DT}, {YEAR_MONTH}) ORDER BY {DT}"  # noqa: S608
    return materialize_query(QueryBuilder().execute(query))


def write_delta_partitioned(start: datetime, days: int, target: Path) -> None:
    """Append sorted records to a Delta table partitioned by calendar month."""
    write_deltalake(
        target,
        create_batch(start, days),
        mode="append",
        partition_by=list(PARTITION_KEYS),
        writer_properties=WriterProperties(compression="ZSTD"),
    )


def read_parquet_timestamps(uri: str, table_name: str) -> list[datetime]:
    """Read one Parquet file in physical row order through DataFusion SQL."""
    escaped_uri = uri.replace("'", "''")
    query = QueryBuilder()
    # The URI and table name are generated locally rather than supplied by a user.
    create_sql = f"CREATE EXTERNAL TABLE {table_name} STORED AS PARQUET LOCATION '{escaped_uri}'"
    materialize_query(query.execute(create_sql))
    result = query.execute(f"SELECT {DT} FROM {table_name}")  # noqa: S608
    return query_column_values(result, DT)


def scan_delta_sorted(table: DeltaTable) -> list[datetime]:
    """Read active Delta files in ascending partition and maximum timestamp order."""
    actions = table.get_add_actions(flatten=True)
    paths = actions.column("path").to_pylist()
    partitions = actions.column(f"partition.{YEAR_MONTH}").to_pylist()
    maximums = actions.column(f"max.{DT}").to_pylist()
    record_counts = actions.column("num_records").to_pylist()
    null_counts = actions.column(f"null_count.{DT}").to_pylist()

    rows = zip(partitions, maximums, paths, record_counts, null_counts, strict=True)
    sorted_paths = [
        path
        for _, _, path, _, _ in sorted(
            (row for row in rows if row[3] > 0 and row[4] == 0),
            key=lambda row: (row[0], row[1]),
        )
    ]

    table_uri = table.table_uri.rstrip("/")
    timestamps: list[datetime] = []
    for index, path in enumerate(sorted_paths):
        timestamps.extend(read_parquet_timestamps(f"{table_uri}/{path}", f"file_{index}"))
    return timestamps


def is_sorted(values: Sequence[datetime]) -> bool:
    """Return whether timestamps are monotonically non-decreasing."""
    return all(previous <= current for previous, current in pairwise(values))


with tempfile.TemporaryDirectory() as temp_dir:
    dst_path = Path(temp_dir)
    stop = 360
    step = 40
    expected: list[datetime] = []

    for offset in range(0, stop, step):
        start = datetime(2020, 1, 1, tzinfo=UTC) + timedelta(days=offset)
        batch = [start + timedelta(days=day) for day in range(step)]
        # Every generated batch is already sorted by timestamp.
        assert is_sorted(batch)

        expected.extend(batch)
        # Each new batch starts after the previously written batch ends.
        write_delta_partitioned(start, step, dst_path)

    assert is_sorted(expected)

    table = DeltaTable(dst_path)
    for index, uri in enumerate(table.file_uris()):
        assert is_sorted(read_parquet_timestamps(uri, f"source_{index}"))

    before_compaction = scan_delta_sorted(table)
    assert is_sorted(before_compaction)

    table.optimize.compact()
    after_compaction = scan_delta_sorted(table)
    # This assertion currently fails because compact() does not preserve row order.
    assert is_sorted(after_compaction)
```

The script consistently reaches only the final assertion and terminates with:

```text
  File "compact_partitioned_delta.py", line ..., in <module>
    assert is_sorted(after_compaction)
AssertionError
```

I reproduced the same post-compaction assertion failure in five consecutive
runs.

Example compaction metrics:

```text
{
    'numFilesAdded': 8,
    'numFilesRemoved': 16,
    'partitionsOptimized': 8,
    'numBatches': 16,
    'totalConsideredFiles': 20,
    'totalFilesSkipped': 4,
    'preserveInsertionOrder': True,
    'plannerStrategy': 'preserveLocality',
    'preservedStableOrder': True,
    'maxBinSpanFiles': 2
}
```

Additional inspection of the resulting files showed:

```text
active files after compact: 12
internally unsorted files after compact: 8
timestamp inversions after compact: 8
```

### Related issues

- #4310
- #4326
- #3538

#4326 appears to preserve file locality and stabilize writer output ordering,
but the reproducer above indicates that it does not preserve monotonic row order
inside the compacted Parquet files.

#### Scope of this delta-rs request

Based on the Spark implementation, this issue is not claiming that delta-rs
differs from an existing Delta Lake protocol guarantee. Physical row order is
not currently part of the documented plain-compaction contract.

Instead, this request has two parts:

1. Clarify the meaning of the delta-rs metrics
   `preserveInsertionOrder` and `preservedStableOrder`.

   In the reproducer, both metrics can be `True` even though the compacted
   Parquet files are no longer monotonic by `dt`. If these metrics refer only to
   file-planning locality or deterministic processing order, that distinction
   should be documented explicitly.

2. Add an opt-in lexicographic sorting capability to `compact()`.

   For example:

   ```python
   table.optimize.compact(sort_columns=["dt"])
   ```

   The existing behavior can remain the default. When `sort_columns` is
   specified, each compacted output file would be physically sorted by those
   columns before it is written.

Possible initial semantics could be:

- Sorting is opt-in.
- Sorting is performed within each compaction bin or partition.
- Each output Parquet file is sorted by the requested columns.
- No global ordering guarantee is made across partitions.
- Cross-file ordering within one partition may be considered separately.
- Z-Ordering remains a distinct optimization for multidimensional data
  locality.
- The API can later be aligned with the table-level sort-order proposal in
  [delta#5308](https://github.com/delta-io/delta/issues/5308).

This would address time-series, append-only log, CDC, and ordered range-scan use
cases without incorrectly treating Z-Ordering as lexicographic sorting.

_No response_

### Priority

Low - Nice to have

### Comparison with Apache Spark Delta Lake

I reviewed the Apache Spark implementation in the official
[delta-io/delta](https://github.com/delta-io/delta) repository.

The official implementation does not currently guarantee lexicographic or
monotonic row ordering during ordinary `OPTIMIZE` compaction either.

#### Plain compaction implementation

The default Spark Delta optimization strategy prepares files for bin packing by
sorting them by file size:

- [`OptimizeTableStrategy.prepareFilesPerPartition`](https://github.com/delta-io/delta/blob/dbc6b2e9f98dd3509cd03704b1570b397398f3e6/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala#L49-L56)

```scala
def prepareFilesPerPartition(
    inputFiles: Seq[AddFile]
): Seq[AddFile] = inputFiles.sortBy(_.size)
```

The prepared files are then grouped into bins according to their cumulative
file size:

- [`OptimizeExecutor.groupFilesIntoBins`](https://github.com/delta-io/delta/blob/dbc6b2e9f98dd3509cd03704b1570b397398f3e6/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala#L399-L448)

For ordinary compaction, each bin is written through either `coalesce(1)` or
`repartition(1)`:

- [`OptimizeExecutor.runOptimizeBinJob`](https://github.com/delta-io/delta/blob/dbc6b2e9f98dd3509cd03704b1570b397398f3e6/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala#L528-L548)

```scala
if (useRepartition) {
  input.repartition(numPartitions = 1)
} else {
  input.coalesce(numPartitions = 1)
}
```

There is no sort operation on a user data column in this compaction path.
Consequently, the Spark implementation also does not define plain compaction as
preserving an existing timestamp or lexicographic order.

The official documentation describes compaction as coalescing small files into
larger files and balancing their sizes. It does not document preservation of
physical row order:

- [Delta Lake optimization documentation: Compaction](https://github.com/delta-io/delta/blob/dbc6b2e9f98dd3509cd03704b1570b397398f3e6/docs/src/content/docs/optimizations-oss/index.mdx#L13-L79)

In particular, the documentation states that bin packing aims to produce
evenly balanced files with respect to their size on disk. The documented
correctness guarantee is that queries return the same logical results before
and after `OPTIMIZE`, not that physical row order is preserved.

#### Z-Ordering is not lexicographic sorting

The official documentation defines Z-Ordering as a technique for colocating
related information to improve data skipping:

- [Delta Lake optimization documentation: Z-Ordering](https://github.com/delta-io/delta/blob/dbc6b2e9f98dd3509cd03704b1570b397398f3e6/docs/src/content/docs/optimizations-oss/index.mdx#L120-L174)

Its implementation computes a synthetic space-filling-curve key, repartitions
by range using that key, and can optionally sort within each Spark partition by
the synthetic key:

- [`SpaceFillingCurveClustering.cluster`](https://github.com/delta-io/delta/blob/dbc6b2e9f98dd3509cd03704b1570b397398f3e6/spark/src/main/scala/org/apache/spark/sql/delta/skipping/MultiDimClustering.scala#L64-L99)

```scala
val mdcCol = getClusteringExpression(cols, numRanges)

df.withColumn(repartitionKeyColName, mdcCol)
  .repartitionByRange(
    approxNumPartitions,
    col(repartitionKeyColName),
  )

if (sortWithinFiles) {
  repartitionedDf =
    repartitionedDf.sortWithinPartitions(repartitionKeyColName)
}
```

This improves multidimensional locality, but it is not a general guarantee of
lexicographic ordering by the original columns. Liquid Clustering has a similar
data-locality goal and also does not provide deterministic lexicographic order.

#### Related official Delta Lake proposal

The official Delta Lake repository already has an open cross-engine feature
request for deterministic file-level sorting:

- [delta#5308: Table-level Sort Order Specification for OPTIMIZE operations](https://github.com/delta-io/delta/issues/5308)

That proposal explicitly identifies the following current limitations:

- Plain `OPTIMIZE` does not preserve a user-defined sort order.
- Z-ORDER does not guarantee lexicographic ordering.
- Liquid Clustering does not guarantee lexicographic ordering.
- Users currently need to implement custom partition rewrites to maintain
  sorted files.

It proposes an opt-in table-level sort order with semantics such as:

- Sort each output file by user-selected columns.
- Apply sorting within partition boundaries.
- Do not require a global ordering guarantee across all files.
- Preserve the current unsorted behavior for tables without a sort-order
  configuration.
- Allow implementations such as Spark Delta and delta-rs to support the same
  optimization intent.

The proposal also references the earlier delta-rs proof of concept:

- [delta-rs#3539: Guarantee global sorting in OPTIMIZE output](https://github.com/delta-io/delta-rs/pull/3539)

### Contribution

mahesh-desu:
# Description

Adds an opt-in `sort_columns` option to `DeltaTable.optimize.compact()` / `OptimizeBuilder::with_sort_columns`. When provided, each compacted output file is physically sorted by the requested columns (ascending, nulls first). When omitted, behaviour is byte-for-byte unchanged. Unknown columns are rejected with a clear error, and combining `sort_columns` with `OptimizeType::ZOrder` is rejected.

To avoid a breaking change, the existing public `create_merge_plan` signature is left unchanged; the new implementation lives in the additive `create_merge_plan_with_sort`, and `create_merge_plan` delegates with `None`.

# Related Issue(s)

- closes #4724

# Changes

- Rust core (`crates/core/src/operations/optimize.rs`): `sort_columns` on `OptimizeInput` + `OptimizeBuilder::with_sort_columns`; `create_merge_plan_with_sort`; shared, fallible `column_expr`; validation via `validate_column_path`; explicit rejection of Z-order + sort.
- Python binding: `compact(sort_columns=[...])` in `python/src/lib.rs`, `python/deltalake/table.py`, `python/deltalake/_internal.pyi`.
- Tests:
  - Rust unit tests for validation, unknown-column error, empty path, and Z-order rejection.
  - Rust integration tests in `crates/core/tests/it_datafusion/command_optimize.rs`: sorted output, multi-column sort, preserved default order, partition-local sort, and nested path (`meta.field_a`).
  - Python tests in `python/tests/test_optimize.py`.
- Docs: note that enabling sorting coalesces a bin into a single sorted partition (memory/parallelism trade-off).

# Verification

- `cargo check -p deltalake-core --features datafusion --lib`
- `cargo test -p deltalake-core --features datafusion --lib optimize`
- `cargo test -p deltalake-core --profile test --lib` (test module compiles)
- Python: built the extension with maturin and ran `pytest tests/test_optimize.py -k sort_columns` → 2 passed.

The integration-test binaries require the gitignored `dat/` fixtures, which are not present in my sandbox; the new integration tests were written against, and reuse, the existing helpers in `command_optimize.rs`.

# Notes

- `sort_columns` is not currently persisted in the committed operation metadata; persisting it would require extending the `DeltaOperation::Optimize` protocol action and was left out deliberately.
- Follow-up (separate change): compaction metrics still report `preserveInsertionOrder: true` when `sort_columns` reorders rows; a dedicated metric would avoid the ambiguity.
- An independent multi-axis review was run on this change; no Critical/Required findings remained after the follow-up fixes.

# Documentation

Documentation is limited to rustdoc on `with_sort_columns` / Python `compact()` docstrings. No external docs changes.


mahesh-desu:
Review/context note for reviewers:

**Independent review.** This branch was reviewed across five axes (correctness, readability, architecture, security, performance) and the follow-up commit addresses all required findings. The original change broke the public `create_merge_plan` signature; that is now fixed by keeping the existing signature and adding the additive `create_merge_plan_with_sort`, so downstream Rust users are unaffected. The heavy Parquet-reading tests were moved into `crates/core/tests/it_datafusion/command_optimize.rs` to reuse the existing helpers instead of duplicating them.

**Tests run locally**
- `cargo check -p deltalake-core --features datafusion --lib`
- `cargo test -p deltalake-core --features datafusion --lib optimize`
- Python: `pytest tests/test_optimize.py -k sort_columns` → 2 passed (extension built with maturin).

The integration-test binaries need the gitignored `dat/` fixtures, which I don't have locally; the new integration tests reuse the existing helpers and should run in CI.

**Known follow-ups (not in this PR), happy to pick up if wanted**
1. `sort_columns` is not persisted in the committed `OPTIMIZE` operation metadata — doing so would require extending `DeltaOperation::Optimize`.
2. The compaction metrics still report `preserveInsertionOrder: true` / `preservedStableOrder: true` when sorting reorders rows; a dedicated metric (or setting them false) would remove the ambiguity.
3. `sort_columns` currently accepts partition columns; Z-order rejects them. Should sort do the same, or is the no-op acceptable?

Design questions: do you prefer `create_merge_plan_with_sort(..)` as a separate entry point, or should this instead go through an options/params struct on the existing builder path?

PR #4745, opened 2026-09-19
feat: expose create_add from ParquetMetaData for readers

### LINKED ISSUE #3194: Expose a create_add function to the stats writer that accepts ParquetMetaData
# Description

Currently I have a committer that has read a Parquet file that it is then trying to add to a Delta Lake table. However the current `create_add` function really only applies to writers of the Parquet file, as it uses the `format::FileMetaData` and readers have the `file:metadata::FileMetaData`. 

On our fork we have a function that allows a reader to create an Add: https://github.com/delta-io/delta-rs/commit/1a2dd95f492aa71bbfe8d45a0260dc1d4aab9542

**Use Case**

Allows backfilling of Parquet files into Delta Lake tables.

**Related Issue(s)**

mahesh-desu:
# Description

Exposes a public constructor for `Add` actions from a read Parquet file's `ParquetMetaData`, so consumers can register/backfill pre-existing Parquet files into a Delta table without rewriting the data.

`create_add_from_parquet_metadata` delegates to the existing internal `create_add`, accepts an optional `modification_time` (defaulting to the current time) so callers can mirror the source file's mtime as `convert_to_delta` does, and takes `stats_columns: Option<&[String]>`. It is re-exported from `deltalake_core::writer`.

# Related Issue(s)

- closes #3194

# Changes

- `crates/core/src/writer/stats.rs`: public `create_add_from_parquet_metadata`; private `create_add_with_time`; the internal `create_add` keeps its existing signature and delegates with `None`, so existing callers (e.g. `datafile/writer.rs`) are unaffected.
- `crates/core/src/writer/mod.rs`: re-export the new function.
- `crates/core/src/lib.rs`: `pub use delta_kernel;` so callers can name `Scalar` / `DataSkippingNumIndexedCols`.
- Tests: concrete statistics assertions (`num_records`, min/max, null count), partition-column exclusion, caller-supplied `modification_time`, and a compile-checked doctest.

# Verification

- `cargo test -p deltalake-core --features datafusion --lib writer::stats` → 6 passed.
- `cargo test -p deltalake-core --doc writer::stats` → 1 passed.
- `cargo check -p deltalake-core --features datafusion --lib`.

# Notes

- Follow-up (separate change): `operations::convert_to_delta` still hand-rolls an equivalent `Add` construction and could be unified onto this constructor; `indexmap` is used in the public signature but is not yet re-exported.
- An independent multi-axis review was run; no Critical/Required findings remained after the follow-up fixes.

# Documentation

Rustdoc (including a runnable example) documents the new function. No external docs changes.


mahesh-desu:
Review/context note for reviewers:

**Independent review.** Reviewed across five axes. The follow-up commit addresses the required findings: the test now asserts concrete statistics (row count, min/max, null count) plus a partition-column-exclusion case, and the wrapper now takes `modification_time: Option<i64>` (`None` = current time) so callers can mirror the source file's mtime like `convert_to_delta`, instead of always stamping `now()`.

**Tests run locally**
- `cargo test -p deltalake-core --features datafusion --lib writer::stats` → 6 passed.
- `cargo test -p deltalake-core --doc writer::stats` → 1 passed (the example is a real, compile-checked doctest).

**Design questions / follow-ups**
1. Naming: I used `create_add_from_parquet_metadata`. The original issue references a fork proof-of-concept named `create_add_from_read` — prefer one over the other?
2. `pub use delta_kernel;` was added so callers can name `Scalar`/`DataSkippingNumIndexedCols`; `indexmap` is also in the public signature but not re-exported yet. Should I re-export `indexmap` too?
3. `operations::convert_to_delta` still hand-rolls an equivalent `Add` construction (it sets the same fields). Now that this constructor can express the mtime, it could be unified onto it — left out of this PR to keep it focused, happy to do it as a follow-up.

PR #4746, opened 2026-09-20
feat: add DeltaTable::history_stream to read commit history lazily

### LINKED ISSUE #3908: [Feature]: `history()` api that returns a BoxStream

Continuing from here https://github.com/delta-io/delta-rs/issues/3753#issuecomment-3454912824

We had an issue open for have a non collecting iterator for `history` and although now the api returns a iterator, it still collects all the items from `commit_infos()`  and then returns an iterator on the collected Vec. 

We actually want to get the direct stream from commit_infos itself because we want to break early and prevent unnecessary object store get calls

We know history API takes `limit` but our break condition is based on the commit timestamp

It would help to have a streaming iterator like so:

```rust
    pub async fn history_stream(
        &self,
        limit: Option<usize>,
    ) -> DeltaResult<BoxStream<'_, DeltaResult<Option<CommitInfo>>>> {
        self.snapshot()?
            .snapshot()
            .snapshot()
            .commit_infos(&self.log_store(), limit)
            .await
    }
```

_No response_

### Priority

### Additional context

_No response_

### Contribution

ShayanGho:
# Description
`DeltaTable::history()` returns an iterator, but it collects the whole
`commit_infos` stream first, so every selected commit file is fetched before
the caller sees the first entry. On a 10-commit table I recorded 10 GETs on
`_delta_log/*.json` before calling `next()` once.

This adds `DeltaTable::history_stream(limit)`, which returns
`BoxStream<'_, DeltaResult<CommitInfo>>` and reads commit files only as the
stream is consumed. Callers can stop early, for example on a timestamp.

- Returns the stream directly, with errors yielded through it, matching
  existing `DeltaTable` streaming APIs. Commits without a `commitInfo`
  action are skipped, as in `history()`.
- `history()` is now a collecting wrapper around the stream. Its signature
  and behaviour are unchanged, so the Python binding is untouched.
- New test uses the existing `InstrumentedStore`: building the stream reads
  nothing, and consuming one entry with `log_buffer_size = 1` reads only
  version 9.

Existing limit semantics and buffered read-ahead are preserved. Log-file
metadata is still listed and sorted before the first entry is yielded.

AI disclosure: I used Claude and Codex for code exploration, implementation
guidance, review, and documentation drafting. I applied the changes and ran
the tests myself.

# Related Issue(s)
- closes #3908

# Documentation
Added a `history_stream` example to `docs/usage/examining-table.md`.


ion-elgreco:
The Python code is untouched from what I see, I am reviewing on my phone so I can't verify if the returned value wraps the stream correctly?

ShayanGho:
> The Python code is untouched from what I see, I am reviewing on my phone so I can't verify if the returned value wraps the stream correctly?

The Python binding still calls table.history(limit).await, which now collects history_stream(limit) internally and returns the same iterator as before. The binding then serializes those entries as before, so Python’s return type and eager behavior are unchanged.
This PR exposes streaming to Rust callers only; it doesn’t expose the stream as a Python iterator. Were you expecting a Python streaming API in this PR as well?

rtyler:
@ShayanGho for the Python use-case, until the Python binding invokes the stream function and starts yielding entries I don't believe this change accomplishes your stated goal.

@ion-elgreco since we're about to cut 1.0, I'm open to making `history_stream` just the default `history` API since the signatures are largely the same and it's beneficial IMHO to change this now rather than incorporate both of these public functions into `DeltaTable`

ShayanGho:
> @ShayanGho for the Python use-case, until the Python binding invokes the stream function and starts yielding entries I don't believe this change accomplishes your stated goal.
> 
> @ion-elgreco since we're about to cut 1.0, I'm open to making `history_stream` just the default `history` API since the signatures are largely the same and it's beneficial IMHO to change this now rather than incorporate both of these public functions into `DeltaTable`

@rtyler Thanks, I scoped this to the Rust API based on #3908’s example, so Python currently gets no streaming benefit. I can change the existing Rust 'history()' to return the stream instead of adding a second public method. I'll make that change once you and @ion-elgreco are aligned on it.

For Python, would a separate follow-up be acceptable? Making history() lazy changes its return type from a list, breaking usage such as dt.history(1)[0]. We’d also need to handle the Rust stream’s lifetime across Python iteration. I’m happy to take that on next, or include it here if you’d prefer both changes together for 1.0.

PR #4747, opened 2026-09-20
fix: only treat a directory as a partition when the name is followed by =

Rodrigo-Palma:
# Description

`is_hidden_directory` is what keeps vacuum from deleting paths that are not table data. Partition directories are deliberately excluded from it, so that they can be vacuumed, and the doc comment states the rule as:

> Names of the form partitionCol=[value] are partition directories, and should be deleted even if they'd normally be hidden.

The check only compared the prefix:

```rust
.any(|partition_column| path_name.starts_with(partition_column))
```

So any hidden directory whose name merely *starts with* a partition column name was classified as a partition directory, and lost the protection that the leading `.` or `_` is meant to give it.

With a partition column named `_date`, an unrelated `_dates_backup/` directory matches and becomes a deletion candidate for vacuum, even though it is not a partition of the table.

The comment in this function is taken from Delta Spark, where the equivalent check compares against the column name plus the separator ([`DeltaTableUtils.isHiddenDirectory`](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala)):

```scala
!partitionColumnNames.exists(c => pathName.startsWith(c ++ "="))
```

This change matches on the separator as well, so the code enforces the rule the comment describes.

Reaching it needs a partition column whose name starts with `.` or `_` plus a hidden directory sharing that prefix, so the blast radius is narrow, but the consequence is deleting a directory that is not part of the table.

# Related Issue(s)

None.

# Tests

Added `test_hidden_directory_matches_partition_dirs_not_mere_prefixes`, which covers both sides of the rule: a real `_date=2024-01-01/` partition stays vacuumable, and an unrelated `_dates_backup/` stays hidden. It fails on `main` and passes with this change.

`cargo test -p deltalake-core --features datafusion --lib vacuum` is green (34 passed).


adamreeve (dismissed):
Thanks @Rodrigo-Palma, this makes sense to me and I like that the implementation doesn't need to build a new string like the Java version :smile:.

adamreeve:
The Python format check has failed, you should be able to fix this by running `make format` in the `python` directory. CONTRIBUTING.md has a helpful list of checks that should pass before making a PR: https://github.com/delta-io/delta-rs/blob/main/CONTRIBUTING.md#making-a-pull-request-pr

Rodrigo-Palma:
Thanks, done in 41de658.

One note on where it came from: the file the check rejects is `python/tests/test_writerproperties.py`, which is not part of this PR. It arrived on the branch with the merge of `main` in c7d7d47, and it is unformatted on `main` itself as of 90b904e, so `ruff format --check` fails on any branch that merges main right now. The push to main went green because `python_build.yml` only runs on `pull_request`, so the format check never ran on it.

`ruff format` touches that one file and nothing else (1 reformatted, 79 unchanged, matching the CI output), so I have included it here to get this green. Happy to drop the commit if you would rather fix main separately and keep this PR to the vacuum change.

adamreeve:
Ah, it looks like the formatting issue snuck in to #4740, I didn't notice that it was unrelated to this change. I think it's a small enough fix that it's OK to include here to get the tests to run.

adamreeve (approved):

PR #4757, opened 2026-09-21
fix: preserve exact int64 decimal statistics at scale zero

shoemoney:
# Description

Preserve exact INT64 decimal statistics when the scale is zero. Converting `999999999999999999` to `f64` rounds it to `1000000000000000000`, making a minimum larger than the actual value. Keeping the raw `i64` through JSON serialization produces exact minimum and maximum bounds, including negative values and values beyond `2^53`.

Nonzero-scale INT64 and fixed-length-byte decimal conversions retain their existing behavior. Tests cover both bounds, exact JSON integers, nonzero scales, and a real decimal writer with multiple row groups followed by `Add.get_stats()`.

Validation with the repository's Rust 1.94.1 toolchain:
- The exact-bound and writer-to-Add regression tests fail against the previous implementation and pass with the fix.
- `cargo test -p deltalake-core --lib`: 539 passed, 1 ignored.
- `cargo fmt --all --check`: passed.
- `make setup-dat` and `cargo clippy -p deltalake-core --lib --tests`: passed; existing warnings remain.

Remote CI results are separate from these local checks; the previous head’s test workflows required maintainer approval. The earlier `draft_pr` failure was a `Resource not accessible by integration` error in the draft-conversion workflow, not a source test failure.

AI assistance was used to prepare the code, tests, and description; the change was source-reviewed and tested locally.

# Related Issue(s)

No linked issue; the regression tests reproduce the precision loss.

# Documentation

No public API changes.


adamreeve (approved):
Thanks for the fix @shoemoney! This looks reasonable to me. The stats are only aggregated from a single file and the scale will always be equal, so all comparisons in the aggregation will be on int64 values. And `impl From<StatsScalar> for serde_json::Value` already writes decimals with zero scale as i64 JSON values so that doesn't change.

I'm concerned that this is only a partial fix though. Non-zero scale decimals will have the same problem. Maybe for the non-zero scale case we need to adjust the float value somehow so that it's <= the true minimum for the min stat, and >= the true maximum for the max stat. That could be done as a follow up though and might need more discussion.

PR #4758, opened 2026-09-21
refactor: convert to delta commits its own operation + stats collection opt out

### LINKED ISSUE #2046: Convert to delta operation should add a CONVERT operation to the commit log
Currently our convert to delta operation reuses the create operation to create the table, this works but is not entirely correct. Other engines simply write the commit log as operation="CONVERT".

ion-elgreco:
# Description
We still go through the create builder but only to derive the actions, committing is done now with a CONVERT operation. Now its 1:1 with delta-spark.

- closes https://github.com/delta-io/delta-rs/issues/2046


crates/core/src/operations/convert_to_delta.rs
```diff
@@ -1037,6 +1085,84 @@ mod tests {
...
+            .with_partition_schema(vec![schema_field("year", PrimitiveType::Integer, true)])
+            .await
+            .expect("Failed to convert to Delta table");
+
+        let commit_info = table
+            .history(None)
+            .await
+            .expect("Failed to read the commit log")
+            .next()
+            .expect("The commit log should hold one entry");
```
rtyler, line 1105: There is a `last_commit` function on `DeltaTable` that exists in our test builds for just this kind of purpose 😄

crates/core/src/operations/convert_to_delta.rs
```diff
@@ -1037,6 +1085,84 @@ mod tests {
...
+        let parameters = commit_info
+            .operation_parameters
+            .expect("The commit should record operation parameters");
+        assert_eq!(
+            parameters.get("collectStats"),
+            Some(&serde_json::json!("false"))
```
rtyler: 😆 why does this use `serde_json`

python/tests/test_convert_to_delta.py
```diff
@@ -45,6 +45,33 @@ def test_local_convert_to_delta(
...
+@pytest.mark.pyarrow
+def test_convert_without_stats(tmp_path: pathlib.Path, sample_data_pyarrow: "pa.Table"):
+    import pyarrow as pa
+    import pyarrow.dataset as ds
+
+    ds.write_dataset(
+        sample_data_pyarrow,
+        tmp_path,
+        format="parquet",
+        existing_data_behavior="overwrite_or_ignore",
```
rtyler, line 61: What is the point of the redundant test here? There's already rust test coverage doing this exact same check.

ion-elgreco: Almost all our python tests do mostly cover the same behavior as the rust test coverage. I would say best to keep it in

crates/core/src/operations/convert_to_delta.rs
ion-elgreco: Ha wasn't aware of this api ^^

crates/core/src/operations/convert_to_delta.rs
ion-elgreco: Asserting on the full params instead now

rtyler (approved):

